PR #615, opened 2016-06-10
Fix bug 1548597 / 76142 (InnoDB tablespace import fails when importin…

laurynas-biveinis:
…g table w/ different data directory)

Importing a transportable tablespace will fail if its table definition
has data directory specified while the donor table did not have one,
and vice versa. Fix by relaxing the checks in fil_node_open_file() and
row_import::match_schema() to allow this particular mismatch while
still requiring that the rest of the flags match. Do not actually
patch the tablespace header with the new flag value as InnoDB is
already making assumptions elsewhere that the data directory flag bits
may mismatch. Also fix tablespace import diagnostics that in the case
of flags mismatch printed number of columns, not flags. Add a
testcase.

http://jenkins.percona.com/job/percona-server-5.6-param/1189/


laurynas-biveinis:
@kbauskar, if you happen to be familiar with transportable tablespaces, your second opinion would be very appreciated. Specifically, this fix makes no attempt to re-write imported tablespace flags with the actual data directory flag value, but leaves it mismatched as-is. Based on my understanding of code (see fil_open_single_file_tablespace), this is correct design, but maybe I understood it wrong.

mysql-test/suite/innodb/t/bug-76142.test
```diff
@@ -0,0 +1,41 @@
+--source include/have_innodb.inc
+
+--echo #
+--echo # Bug 76142: InnoDB tablespace import fails when importing table w/ different data directory
+--echo #
+
+--let $MYSQLD_DATADIR = `SELECT @@datadir`
+
```
percona-ysorokin, line 9: Add

> --let $CURRENT_DB = `SELECT DATABASE()`

and then use ".../$CURRENT_DB/..." in --copy-file / --move_file paths instead of ".../test/..."

percona-ysorokin:
LGTM with minor MTR test comments

storage/innobase/fil/fil0fil.cc
```diff
@@ -822,12 +822,19 @@ fil_node_open_file(
...
 		}
 
-		if (UNIV_UNLIKELY(space->flags != flags)) {
+		/* Validate the flags but do not compare the data directory
+		flag, in case this tablespace was relocated. */
+		const unsigned relevant_space_flags
+			= space->flags & ~FSP_FLAGS_MASK_DATA_DIR;
+		const unsigned relevant_flags
+			= flags & ~FSP_FLAGS_MASK_DATA_DIR;
+		if (UNIV_UNLIKELY(relevant_space_flags != relevant_flags)) {
```
kbauskar: I would expect a warning if flag mis-match on data-dir part so that user know we incompatibility we are trying allowing.

storage/innobase/row/row0import.cc
```diff
@@ -1335,12 +1335,15 @@ row_import::match_schema(
 	THD*		thd) UNIV_NOTHROW
 {
 	/* Do some simple checks. */
+	const unsigned relevant_flags = m_flags & ~DICT_TF_MASK_DATA_DIR;
+	const unsigned relevant_table_flags
+		= m_table->flags & ~DICT_TF_MASK_DATA_DIR;
 
-	if (m_flags != m_table->flags) {
+	if (relevant_flags != relevant_table_flags) {
```
kbauskar, line 1342: I would expect a warning if flag mis-match on data-dir part so that user know we incompatibility we are trying allowing.

kbauskar:
LGTM.
I don't see any technical flaw as we are ignore flag during important and tablespace open.
I would propose we warn user about it though.

I am not sure what motivated this but may be it would create issues with 5.8 once all this information is moved to data-dictionary. Can be ignore for now though.

storage/innobase/row/row0import.cc
laurynas-biveinis, line 1342: I am not sure - because the user has no means of actually fixing this warning short of hex editor. A warning that's unfixable would be error log spam

laurynas-biveinis:
Please see my inline comment re. warning. As for 5.8, we would adjust the fix as needed during the porting process.

kbauskar:
LGTM.

laurynas-biveinis:
MTR comments addressed, http://jenkins.percona.com/job/percona-server-5.6-param/1207/

PR #617, opened 2016-06-10
Fix bug 1591202 (client and libmysqlclient VIO drops connection if si…

laurynas-biveinis:
…gnal received during read())

On read()/write() syscall returning EINTR, 5.5 clients will retry only
a couple of times before returning failure. This is different from
both lower and higher versions where client EINTR is retried
indefinitely. It appears that this is a regression caused by
THREAD_SAFE_CLIENT symbol removal cleanup, which effectively disabled
code protected by this symbol whereas it should have been
enabled. Fixed accordingly.

http://jenkins.percona.com/job/percona-server-5.5-param/1247/

### DIFF
--- a/sql/net_serv.cc
+++ b/sql/net_serv.cc
@@ -678,13 +678,13 @@ net_real_write(NET *net,const uchar *packet, size_t len)
 		  my_progname);
 #endif /* EXTRA_DEBUG */
       }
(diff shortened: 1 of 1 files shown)


percona-ysorokin:
LGTM

PR #618, opened 2016-06-10
Bug1591202 5.6

laurynas-biveinis:


percona-ysorokin:
5.6 has another reference to `THREAD_SAFE_CLIENT`

> static my_bool
> net_should_retry(NET _net, uint *retry_count MY_ATTRIBUTE((unused)))
> {
>  my_bool retry;
> #if !defined(MYSQL_SERVER) && defined(THREAD_SAFE_CLIENT)
>  /_
>    In the thread safe client library, interrupted I/O operations
>    are always retried.  Otherwise, its either a timeout or a
>    unrecoverable error.
>  _/
>  retry= vio_should_retry(net->vio);
> #else
>  /_
>    In the non-thread safe client library, or in the server,
>    interrupted I/O operations are retried up to a limit.
>    In this scenario, pthread_kill can be used to wake up
>    (interrupt) threads waiting for I/O.
>  _/
>  retry= vio_should_retry(net->vio) && ((_retry_count)++ < net->retry_count);
> #endif
>  return retry;
> }

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.6-param/1234/

percona-ysorokin:
LGTM

PR #629, opened 2016-06-19
5.7 version of user filtering for audit log

gl-sergei:
Jenkins: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/218/

### DIFF
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -31,6 +31,7 @@
 #include "logger.h"
 #include "buffer.h"
 #include "audit_handler.h"
+#include "filter.h"
 
 #define PLUGIN_VERSION 0x0002
 
@@ -60,6 +61,8 @@ char *audit_log_syslog_ident;
 char default_audit_log_syslog_ident[] = "percona-audit";
 ulong audit_log_syslog_facility= 0;
 ulong audit_log_syslog_priority= 0;
+static char *audit_log_exclude_accounts= NULL;
+static char *audit_log_include_accounts= NULL;
 
 static int audit_log_syslog_facility_codes[]=
   { LOG_USER,   LOG_AUTHPRIV, LOG_CRON,   LOG_DAEMON, LOG_FTP,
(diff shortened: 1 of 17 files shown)


gl-sergei:
Jenkins: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/224/
Merge with some changes due to slightly different API and additional PSI instrumentation.
Commit 033c825 is an example of how we can use `security_context_service` here to filter against accounts from `mysql.user`. I am not sure if we must use it. Oracle's plugin does.

gl-sergei:
oops, forgot to move new test case into `plugin/audit_log` dir

plugin/audit_log/audit_log.c
```diff
@@ -787,6 +802,42 @@ int is_event_class_allowed_by_policy(mysql_event_class_t class,
...
+              audit_log_exclude_accounts == NULL);
+
+  if (event_class == MYSQL_AUDIT_CONNECTION_CLASS)
+  {
+    LEX_STRING priv_user, priv_host;
+    MYSQL_SECURITY_CONTEXT ctx;
+    my_svc_bool ret;
+
+    ret= thd_get_security_context(thd, &ctx);
+    DBUG_ASSERT(!ret);
```
laurynas-biveinis: IMHO any potential security context API failure handling must be much more defensive. If any of those calls fails, further processing should be aborted with a my_message

gl-sergei: but can it really fail? `thd_get_security_context` is just a `thd->security_context()` and in most places in server it used like this:

```
...
thd->security_context()->host_or_ip().str,
thd->security_context()->priv_user().str,
...
```

laurynas-biveinis: We don't know how its implementation is going to change in the future

gl-sergei:
I just updated the last commit with error handling.

laurynas-biveinis:
LGTM

laurynas-biveinis:
This also needs a 5.7 blueprint

gl-sergei:
I realised it couple of days ago and created this: https://blueprints.launchpad.net/percona-server/+spec/audit-filter-users-5.7

PR #634, opened 2016-06-29
Fix bug 1597143 (InnoDB: Failing assertion: log_sys->buf == ut_align(…

laurynas-biveinis:
…log_sys->buf_ptr, srv_log_write_ahead_size))

In order to support innodb_flush_method=ALL_O_DIRECT, the log I/O
buffers were aligned to innodb_log_write_ahead_size. This missed that
the variable is dynamic. Fix by aligning to the maximum possible
innodb_log_write_ahead_size instead, and replace ut_align in the
failing asserts-which checks not only alignment but a particular
address-with a modulo-based alignment check.

http://jenkins.percona.com/job/mysql-5.7-param/228/


percona-ysorokin:
LGTM

PR #649, opened 2016-07-09
Bug 1435099: Audit Log events do not report default DB

gl-sergei:
Blueprint
https://blueprints.launchpad.net/percona-server/+spec/audit-log-default-db

Add field `DB` for `MYSQL_AUDIT_GENERAL_STATUS` records.

### DIFF
--- a/mysql-test/t/audit_log_default_db.test
+++ b/mysql-test/t/audit_log_default_db.test
@@ -1,0 +1,96 @@
+# test correctness of default_db field
+
+--source include/not_embedded.inc
+
+let $MYSQLD_DATADIR= `select @@datadir`;
+let $log_file=$MYSQLD_DATADIR/test_audit.log;
+
+set names utf8;
+
+CREATE USER 'user1'@'%' IDENTIFIED BY '111';
+CREATE USER 'user2'@'%' IDENTIFIED BY '111';
+
+CREATE DATABASE db1;
+CREATE DATABASE db2;
+CREATE DATABASE `ąžąžąžą`;
+
+CREATE TABLE db1.t (a VARCHAR(100));
+CREATE TABLE db2.t (a VARCHAR(100));
+CREATE TABLE ąžąžąžą.t (a VARCHAR(100)) charset=utf8;
+INSERT INTO db1.t VALUES ('db1');
+INSERT INTO db2.t VALUES ('db2');
+INSERT INTO ąžąžąžą.t VALUES ('ąžąžąžą');
+
+GRANT ALL PRIVILEGES ON db1.* TO 'user1'@'%';
+
+GRANT ALL PRIVILEGES ON db2.* TO 'user1'@'%';
+GRANT ALL PRIVILEGES ON db2.* TO 'user2'@'%';
+
+GRANT ALL PRIVILEGES ON ąžąžąžą.* TO 'user1'@'%';
+GRANT ALL PRIVILEGES ON ąžąžąžą.* TO 'user2'@'%';
+
+# truncate audit log
+UNINSTALL PLUGIN audit_log;
+--remove_file $log_file
+
+connect (test,localhost,user1,111,db2,);
+connection test;
+SELECT * FROM t;
+
+connect (root,localhost,root,,,);
+connection root;
+INSTALL PLUGIN audit_log SONAME 'audit_log.so';
+disconnect root;
+--source include/wait_until_disconnected.inc
+connection test;
+
+SELECT * FROM t;
+use 	`db1`;
+--error ER_DBACCESS_DENIED_ERROR
+change_user user2,111,db1;
+disconnect test;
+--source include/wait_until_disconnected.inc
+--replace_result $MASTER_MYPORT MASTER_MYPORT $MASTER_MYSOCK MASTER_MYSOCK
+--error ER_DBACCESS_DENIED_ERROR
+connect (test,localhost,user2,111,db1,);
+--source include/wait_until_disconnected.inc
+--replace_result $MASTER_MYPORT MASTER_MYPORT $MASTER_MYSOCK MASTER_MYSOCK
+--error ER_ACCESS_DENIED_ERROR
+connect (test,localhost,user2,112,db2,);
+--source include/wait_until_disconnected.inc
+--replace_result $MASTER_MYPORT MASTER_MYPORT $MASTER_MYSOCK MASTER_MYSOCK
+--error ER_ACCESS_DENIED_ERROR
+connect (test,localhost,user3,111,db2,);
+--source include/wait_until_disconnected.inc
+connect (test,localhost,user2,111,db2,);
+connection test;
+set names utf8;
+SELECT * FROM t;
+--error ER_DBACCESS_DENIED_ERROR
+use    	`db1`;
+SELECT * FROM t;
+change_user user1,111,db1;
+set names utf8;
+SELECT * FROM t;
+use `db2`;
+SELECT * FROM t;
+use ąžąžąžą;
+SELECT * FROM t;
+disconnect test;
+--source include/wait_until_disconnected.inc
+
+connection default;
+
+--source include/count_sessions.inc
+--exec $MYSQL --user=user1 --password=111 test -e "use db1; SELECT * FROM t;"
+--source include/wait_until_count_sessions.inc
+
+--source audit_log_echo.inc
+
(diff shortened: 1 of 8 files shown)


gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1227/

mysql-test/t/audit_log_default_db.test
```diff
@@ -0,0 +1,74 @@
...
+SELECT * FROM t;
+use `db2`;
+SELECT * FROM t;
+disconnect test;
+--source include/wait_until_disconnected.inc
+
+connection default;
+
+--exec $MYSQL --user=user1 --password=111 test -e "use db1; SELECT * FROM t;"
+# is there better way to check that server reset connection initiated by mysql cli?
```
laurynas-biveinis: Wrapping exec with include/count_connections and include/wait_until_connection_count should do this

mysql-test/t/audit_log_echo.inc
```diff
@@ -0,0 +1,26 @@
+# Echo the contents of hte audit log (in CSV format)
```
laurynas-biveinis: s/hte/the

plugin/audit_log/audit_log.c
```diff
@@ -785,6 +796,39 @@ int is_event_class_allowed_by_policy(unsigned int class,
   return (class_mask[policy] & (1 << class)) != 0;
 }
 
+static const CHARSET_INFO *charset_info= &my_charset_latin1;
```
laurynas-biveinis: Is the charset of strings passed to plugin callbacks specified anywhere?

plugin/audit_log/audit_log.c
```diff
@@ -785,6 +796,39 @@ int is_event_class_allowed_by_policy(unsigned int class,
...
+    {
+      while (*str && !(*str == '*' && str[1] == '/'))
+        str++;
+    }
+    else
+      str++;
+  }
+
+  *len= 0;
+  while (str[*len] && my_isvar(charset_info, str[*len]))
```
laurynas-biveinis: Specifically, for this? Would 
use 'ąžąžąžą' 
work?

laurynas-biveinis:
No testcase for the install plugin with an open parallel connection?

plugin/audit_log/audit_log.c
gl-sergei: this patch works OK, here is what I see in the log:

```
"Query","<ID>","<DATETIME>","change_db","<CONN_ID>",0,"use `ąžąžąžą`","user1[user1] @ localhost []","localhost","","","ąžąžąžą"
```

I can't `use ąžąžąžą`, I need to `use `ąžąžąžą``

What doesn't seem to work is disconnect event... and maybe connect as well. Looks like MySQL passing database wrong.

```
"Quit","<ID>","<DATETIME>","<CONN_ID>",0,"user1","user1","","","localhost","","ä…å¾ä…å¾ä…å¾ä…"
```

In this line database printed from the `mysql_event_connection`. Why is this happening?

gl-sergei: Hmm... it is even different on Linux and OSX!

```
-"Quit","<ID>","<DATETIME>","<CONN_ID>",0,"user1","user1","","","localhost","","ä…å¾ä…å¾ä…å¾ä…"
+"Quit","<ID>","<DATETIME>","<CONN_ID>",0,"user1","user1","","","localhost","","Ä…Å¾Ä…Å¾Ä…Å¾Ä…"
```

laurynas-biveinis: Probably raw UTF-8 bytes? The Linux and OSX difference probably has to do with case-sensitive vs case-insensitive filesystem

gl-sergei: I found what was wrong! `mtr` communicating with server in `latin1` by default. It was the source of these funny issues.
When I `set names utf8` I can `use ąžąžąžą` without backticks. And it was an issue for plugin. Solution is to use `event_general->general_charset` when parsing the query.
Database name in the connection event comes in `utf8mb3`.
So, there is another bug! Query is written to log in client charset while other fields in `utf8`.

gl-sergei:
-   fixed typo
-   changed test to use utf8
-   added test for database 'ąžąžąžą'
-   surrounded `--exec mysql` with `--source include/count_sessions.inc/--source include/wait_until_count_sessions.inc`
-   added test for `INSTALL PLUGIN`

jenkins build: http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1231/

mysql-test/t/audit_log_default_db.test
```diff
@@ -0,0 +1,96 @@
...
+GRANT ALL PRIVILEGES ON db1.* TO 'user1'@'%';
+
+GRANT ALL PRIVILEGES ON db2.* TO 'user1'@'%';
+GRANT ALL PRIVILEGES ON db2.* TO 'user2'@'%';
+
+GRANT ALL PRIVILEGES ON ąžąžąžą.* TO 'user1'@'%';
+GRANT ALL PRIVILEGES ON ąžąžąžą.* TO 'user2'@'%';
+
+# truncate audit log
+UNINSTALL PLUGIN audit_log;
```
laurynas-biveinis, line 33: This gives "plugin is busy and will be uninstalled on shutdown"

mysql-test/t/audit_log_default_db.test
```diff
@@ -0,0 +1,96 @@
...
+--remove_file $log_file
+
+connect (test,localhost,user1,111,db2,);
+connection test;
+SELECT * FROM t;
+
+connect (root,localhost,root,,,);
+connection root;
+INSTALL PLUGIN audit_log SONAME 'audit_log.so';
```
laurynas-biveinis, line 42: ...thus this is probably a no-op WRT plugin loading and testing existing connections?

plugin/audit_log/audit_log.c
```diff
@@ -785,6 +796,38 @@ int is_event_class_allowed_by_policy(unsigned int class,
   return (class_mask[policy] & (1 << class)) != 0;
 }
 
+static
+const char *next_word(const char *str, size_t *len,
+                      struct charset_info_st *charset)
```
laurynas-biveinis: const arg?

laurynas-biveinis:
Something else about encodings, in e.g. XML the header is <?xml version=\"1.0\" encoding=\"UTF-8\"?>. If the server encoding is different, are the strings transcoded before logging?

mysql-test/t/audit_log_default_db.test
gl-sergei, line 42: I disagree about INSTALL/UNINSTALL. Here is what you see in the log:

```
"Query","<ID>","<DATETIME>","install_plugin","<CONN_ID>",0,"INSTALL PLUGIN audit_log SONAME 'audit_log.so'","root[root] @ localhost []","localhost","","",""
"Quit","<ID>","<DATETIME>","<CONN_ID>",0,"root","root","","","localhost","","test"
"Query","<ID>","<DATETIME>","select","<CONN_ID>",0,"SELECT * FROM t","user1[user1] @ localhost []","localhost","","",""
"Query","<ID>","<DATETIME>","change_db","<CONN_ID>",0,"use  `db1`","user1[user1] @ localhost []","localhost","","","db1"
```

 Specifically, the query `SELECT * FROM t` doesn't report default database, but connection `test` connected with database `db1`. I also see the `Audit` record indicating that plugin is loaded. I had to grep it out, because it contains server startup options.

gl-sergei:
> Something else about encodings, in e.g. XML the header is <?xml version=\"1.0\" encoding=\"UTF-8\"?>. If the server encoding is different, are the strings transcoded before logging?

See my comment above:

> So, there is another bug! Query is written to log in client charset while other fields in utf8.

laurynas-biveinis:
Ah, OK, it's the same thing we are talking about :) Have you logged the bug?

gl-sergei:
- `s/struct charset_info_st *charset/const struct charset_info_st *charset/`
- filed bug https://bugs.launchpad.net/percona-server/+bug/1602986

mysql-test/t/audit_log_default_db.test
laurynas-biveinis, line 42: OK

laurynas-biveinis:
LGTM

PR #651, opened 2016-07-11
Fix .gitignore. spec files shouldn't be listed 

EvgeniyPatlan:
https://bugs.launchpad.net/percona-server/+bug/1600051

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,6 @@
 *.sbr
 *.so
 *.so.*
-*.spec
 *.user
 *.vcproj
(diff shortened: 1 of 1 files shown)


tplavcic:
LGTM
Here's my comment from the bug report:
The upstream is using ".spec.in" files in their packaging directory which are spec templates from which they generate ".spec" files with cmake. So I don't think it is an upstream issue. Since we are not using those templates currently we can remove it, but if we start using it at some point we should re-add this.

PR #656, opened 2016-07-12
Bug1592290 5.7

laurynas-biveinis:
http://jenkins.percona.com/job/mysql-5.7-param/233/
http://jenkins.percona.com/job/mysql-5.7-param/248/

### DIFF
--- a/storage/innobase/btr/btr0sea.cc
+++ b/storage/innobase/btr/btr0sea.cc
@@ -79,6 +79,13 @@ byte		btr_sea_pad2[64];
 /** The adaptive hash index */
 btr_search_sys_t*	btr_search_sys;
 
+/** The size of the constantly allocated AHI hash tables. Written to under
+protection of all the AHI latches, read without latches. */
+ulint		btr_search_sys_constant_mem;
+/** The size of the memory heaps attached to the AHI hash tables. Accessed
+atomically. */
+ulint		btr_search_sys_variable_mem;
+
 /** If the number of records on the page divided by this parameter
 would have been successfully accessed using a hash index, the index
 is then built on the page, assuming the global limit has been reached */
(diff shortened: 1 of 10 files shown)


storage/innobase/btr/btr0sea.cc
```diff
@@ -101,6 +108,32 @@ btr_search_get_n_fields(
 	return(n_fields + (n_bytes > 0 ? 1 : 0));
 }
```
percona-ysorokin, line 110: May be better to use a macro turning into nothing in release mode
something like

``` c++
#ifdef UNIV_DEBUG
  #define btr_search_mem_accounting_validate() btr_search_mem_accounting_validate_impl()
#else
  #define btr_search_mem_accounting_validate()
#endif
static void btr_search_mem_accounting_validate_impl(void) { ... }
```

storage/innobase/btr/btr0sea.cc
```diff
@@ -1720,7 +1805,15 @@ btr_search_update_hash_on_delete(btr_cur_t* cursor)
 	if (block->index) {
 		ut_a(block->index == index);
 
+		ulint old_ht_heap_size = mem_heap_get_size(table->heap);
 		if (ha_search_and_delete_if_found(table, fold, rec)) {
```
percona-ysorokin, line 1811: The same

``` c++
ut_d(os_rmb);
ut_ad(btr_search_sys_variable_mem >= old_ht_heap_size);
```

as in `btr_search_drop_page_hash_index()` ?

percona-ysorokin:
LGTM, comments are minor

storage/innobase/btr/btr0sea.cc
laurynas-biveinis, line 110: This does not look very clean, the function is static to this file and has empty body in release build. The compiler should be able to optimise the call away completely.

laurynas-biveinis:
http://jenkins.percona.com/job/mysql-5.7-param/250/

PR #661, opened 2016-07-13
Documented workaround for LP Bug #1520267 (5.5)

hrvojem:


doc/source/management/pam_plugin.rst
```diff
@@ -77,6 +77,14 @@ Known issues
 
 Default mysql stack size is not enough to handle ``pam_ecryptfs`` module. Workaround is to increase the |MySQL| stack size by setting the `thread-stack <https://dev.mysql.com/doc/refman/5.5/en/server-system-variables.html#sysvar_thread_stack>`_ variable to at least ``512KB`` or by increasing the old value by ``256KB``.
 
+PAM authentication can fail with ``ENOMEM`` error even when there is enough memory available. Current workaround is to set `vm.overcommit_memory <https://www.kernel.org/doc/Documentation/vm/overcommit-accounting>`_ to ``1``:
```
laurynas-biveinis: "ENOMEM" is not what the user will see, according to the bug report the user will find something like 

mysqld: pam_unix(mysqld:account): Fork failed: Cannot allocate memory

in the system log.

Also, as the bug reporter notes, setting vm.overcommit_memory is "not without risks", please check with him if something else needs to be said here

PR #666, opened 2016-07-14
Fix bug 79610 / 1525407 (Failed DROP DATABASE due FK constraint on ma…

laurynas-biveinis:
…ster breaks slave)

If DROP DATABASE fails to delete any of the tables, it is binlogged as
DROP TABLE t1, t2, ... for the tables for which drop succeeded. If slave
has different schema so that it references any of the dropped tables,
replication will break.

Fix by running foreign key checks on all the DROP DATABASE tables before
any table is actually dropped. If any FK constraint would fail, do not
convert DROP DATABASE to DROP TABLE in the binary log and abort
immediately.

To implement the fix,
- add a new handlerton method get_parent_fk_list, that returns the list
  of foreign keys referencing the given table;
- implement this method for InnoDB, make it share code with
  ha_innobase::get_parent_foreign_key_list handler method;
- call it from a new function check_drop_database_foreign_keys, which is
  called from mysql_rm_table_no_locks.

Add a testcase to rpl.rpl_foreign_key_innodb, and adjust
binlog.binlog_database as needed.

http://jenkins.percona.com/job/percona-server-5.5-param/1253/ + http://jenkins.percona.com/job/percona-server-5.5-param/1254/

### DIFF
--- a/mysql-test/extra/rpl_tests/rpl_foreign_key.test
+++ b/mysql-test/extra/rpl_tests/rpl_foreign_key.test
@@ -59,4 +59,54 @@ select count(*) from t1 /* must be zero */;
 connection master;
 drop table t2,t1;
 
+--echo #
+--echo # Bug 79610: Failed DROP DATABASE due FK constraint on master breaks slave
+--echo #
+
+SET foreign_key_checks=1;
+CREATE DATABASE `db2`;
+
+USE `db2`;
+
+create table a1(f1 int);
+
+eval CREATE TABLE `table0` (`ID` bigint(20) primary key) ENGINE=$engine_type;
+eval CREATE TABLE `table1` (`ID` bigint(20) primary key) ENGINE=$engine_type;
+
+create database db1;
+use db1;
+
+eval CREATE TABLE `table2` ( `ID` bigint(20) NOT NULL AUTO_INCREMENT,
+                             `DIVISION_ID` bigint(20) DEFAULT NULL,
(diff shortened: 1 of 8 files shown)


mysql-test/extra/rpl_tests/rpl_foreign_key.test
```diff
@@ -59,4 +59,55 @@ select count(*) from t1 /* must be zero */;
...
+--echo #
+--echo # Bug 79610: Failed DROP DATABASE due FK constraint on master breaks slave
+--echo #
+
+SET foreign_key_checks=1;
+CREATE DATABASE `db2`;
+
+USE `db2`;
+
+create table a1(f1 int);
```
percona-ysorokin, line 71: a1 and a2 are not used

storage/innobase/handler/ha_innodb.cc
```diff
@@ -10057,6 +10071,73 @@ get_foreign_key_info(
...
+
+@return error code or zero for success */
+static
+int
+innobase_get_parent_fk_list(
+	THD*			thd,
+	const char*		path,
+	List<FOREIGN_KEY_INFO>*	f_key_list)
+{
+	ut_a(strlen(path) < 1000);
```
percona-ysorokin: FN_REFLEN instead of 1000?

percona-ysorokin:
LGTM, minor comments only

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.5-param/1260/

percona-ysorokin:
LGTM

PR #669, opened 2016-07-14
Fix bug 1603073 (ssl_ca and ssl_crl test failures)

laurynas-biveinis:
5.5.50 upstream merge has introduced a new testcase, ssl_ca, which has
to be adjusted for the recent TLSv1.2 changes.  Make the testcase
accept both YaSSL and OpenSSL whitelisted default ciphers.

http://jenkins.percona.com/job/percona-server-5.5-param/1257/


percona-ysorokin:
LGTM

PR #674, opened 2016-07-15
Fix bug 1590454 (Percona Server 5.5/5.6 cannot be built with '-DMYSQL…

laurynas-biveinis:
…_MAINTAINER_MODE=ON')

The warnings reported in the bug all come from buf_block_get_frame
changed from
to
by the XtraDB innodb_corrupt_table_action feature.

This makes compiler consider 0 as a possible return value in its
callers, causing warnings when this clashes with nonnull
annotations. The warnings are correct in that, if "block" were NULL
already, the server would crash with NULL dereference before the
buf_block_get_frame call, and so returning 0 instead of block->frame
is redundant.

Fix by introducing buf_nonnull_block_get_frame macro/function, which
only returns block->frame.

At the same time fix btr_get_size() returning 0 instead of
ULINT_UNDEFINED for corrupted indexes.

http://jenkins.percona.com/job/percona-server-5.5-param/1259/

### DIFF
--- a/storage/innobase/btr/btr0btr.c
+++ b/storage/innobase/btr/btr0btr.c
@@ -76,7 +76,7 @@ btr_corruption_report(
 			       buf_block_get_zip_size(block),
 			       BUF_PAGE_PRINT_NO_CRASH);
 	}
-	buf_page_print(buf_block_get_frame(block), 0, 0);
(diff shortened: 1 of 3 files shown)


percona-ysorokin:
LGTM

PR #685, opened 2016-07-20
Fix bug 1604462 (main.mysqlbinlog_debug fails with a LeakSanitizer er…

laurynas-biveinis:
…ror)

Make sure dump_remote_log_entries function frees ev object on error
exits.

http://jenkins.percona.com/job/percona-server-5.6-param/1245/

### DIFF
--- a/client/mysqlbinlog.cc
+++ b/client/mysqlbinlog.cc
@@ -2638,6 +2638,7 @@ static Exit_status dump_remote_log_entries(PRINT_EVENT_INFO *print_event_info,
                                        MYF(MY_WME))))
           {
             error("Could not create log file '%s'", log_file_name);
+            reset_temp_buf_and_delete(ev);
             DBUG_RETURN(ERROR_STOP);
           }
           DBUG_EXECUTE_IF("simulate_result_file_write_error_for_FD_event",
@@ -2646,6 +2647,7 @@ static Exit_status dump_remote_log_entries(PRINT_EVENT_INFO *print_event_info,
                         BIN_LOG_HEADER_SIZE, MYF(MY_NABP)))
           {
             error("Could not write into log file '%s'", log_file_name);
+            reset_temp_buf_and_delete(ev);
             DBUG_RETURN(ERROR_STOP);
           }
           /*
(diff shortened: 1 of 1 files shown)


client/mysqlbinlog.cc
```diff
@@ -2699,7 +2701,10 @@ static Exit_status dump_remote_log_entries(PRINT_EVENT_INFO *print_event_info,
       File file;
 
       if ((file= load_processor.prepare_new_file_for_old_format(le,fname)) < 0)
+      {
+        reset_temp_buf_and_delete(ev);
         DBUG_RETURN(ERROR_STOP);
```
percona-ysorokin, line 2706: I believe that `DBUG_RETURN(retval);` at line 2692 is OK without `reset_temp_buf_and_delete(ev);` but please double check me.

``` c++
       if (raw_mode)
       {
         DBUG_EXECUTE_IF("simulate_result_file_write_error",
                         DBUG_SET("+d,simulate_fwrite_error"););
         if (my_fwrite(result_file, net->read_pos + 1 , len - 1, MYF(MY_NABP)))
         {
           error("Could not write into log file '%s'", log_file_name);
           retval= ERROR_STOP;
         }
         if (ev)
         {
           ev->temp_buf=0;
           delete ev;
         }
       }
       else
       {
         retval= process_event(print_event_info, ev, old_off, logname);
       }

       if (retval != OK_CONTINUE)
         DBUG_RETURN(retval);
```

`ev` mut be deleted either directly by

``` c++
         if (ev)
         {
           ev->temp_buf=0;
           delete ev;
         }
```

or by `process_event()`

percona-ysorokin:
LGTM, if agree with the reasonings

client/mysqlbinlog.cc
laurynas-biveinis, line 2706: Sorry, which one DBUG_RETURN(retval) is OK without reset_temp_buf_and_delete(ev)? Line 2692 is neither

percona-ysorokin, line 2706: The one at the end of my first code block in the previous comment

laurynas-biveinis, line 2706: Yes, it will be either deleted at the end of if (raw_mode), either by process_event call

PR #687, opened 2016-07-21
DB-1003 : Foreground ANALYZE TABLE uses locally scoped buffer for thd…

george-lorch:
…_proc_info
- Moved buffer for process info status into class scope instead of local.
- Renamed/clarified intention of method that sets thd_proc_info.


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1247/

PR #689, opened 2016-07-22
Fix bug 1604774 (Memory leaks in unit tests)

laurynas-biveinis:
Trivailly fix memory leaks in the unit tests that have been exposed by
LeakSanitiser testing.

```
http://jenkins.percona.com/job/percona-server-5.6-param/1246/
```


percona-ysorokin:
LGTM

PR #693, opened 2016-07-22
Bug 1602986: Audit log does not transcode queries

gl-sergei:
Patch addressing following issues:
1. Convert query from client charset to system charset
2. Convert database name extracted from `use' query to system charset
3. Fix `ecsape_string' reported 0 estimated length for output string
   when output buffer was 0 length


gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1248/

plugin/audit_log/audit_log.c
```diff
@@ -246,6 +247,36 @@ void csv_escape(const char *in, size_t *inlen, char *out, size_t *outlen)
...
+  Calculate the size of the otput bufer needed to escape the string.
+
+  @param[in]  in           Input string
+  @param[in]  len          Length of the input string
+
+  @return
+    size of the otput bufer including trailing zero
+*/
+static
+size_t escape_string_dry_run(const char *in, size_t len)
```
laurynas-biveinis: A name like "calculate_escape_string_buf_len" would be more descriptive what the function does (and matches the header comment even) whereas "escape_string_dry_run" is more how it's implemented

plugin/audit_log/audit_log.c
```diff
@@ -468,8 +488,31 @@ char *audit_log_general_record(char *buf, size_t buflen,
...
-                       endptr, endbuf - endptr, &endptr, &full_outlen);
+  query_length= system_charset_info->mbmaxlen * event->general_query_length;
+
+  if (query_length < (size_t) (endbuf - endptr))
+  {
+    uint errors;
+    query_length= my_convert(endptr, query_length, system_charset_info,
+                             event->general_query,
+                             event->general_query_length,
+                             event->general_charset, &errors);
```
laurynas-biveinis, line 502: UTF-8 can represent all the other character sets, right? Should then be asserted that errors == 0?

plugin/audit_log/audit_log.c
```diff
@@ -899,15 +944,16 @@ void audit_log_update_thd_local(audit_log_thd_local *local,
...
         if (*word == '`')
         {
           word++;
           len-= 2;
         }
-        len = len < sizeof(local->db) ? len : sizeof(local->db) - 1;
-
-        memcpy(local->db, word, len);
+        len= my_convert(local->db, sizeof(local->db) - 1, system_charset_info,
+                        word, len, event_general->general_charset, &errors);
```
laurynas-biveinis, line 960: Likewise errors == 0 ?

plugin/audit_log/audit_log.c
gl-sergei, line 502: The problem here is that MySQL in reality is using 3-byte subset of UTF-8 internally and `utf8` is the alias of the `utf8mb3`. Here is an example which will give `errors==3`:

```
SET NAMES utf8mb4;
INSERT INTO t VALUES ('𦉘𦟌𦧲');
```

We'll see warning:

```
Warning        1366    Incorrect string value: '\xF0\xA6\x89\x98\xF0\xA6...' for column 'txt' at row 1
```

and ??? in the audit log:

```
"Query","<ID>","<DATETIME>","insert","<CONN_ID>",0,"INSERT INTO t VALUES ('???')","root[root] @ localhost []","localhost","","","ฐานข้อมูล"
"Query","<ID>","<DATETIME>","show_warnings","<CONN_ID>",0,"SHOW WARNINGS","root[root] @ localhost []","localhost","","","ฐานข้อมูล"
```

laurynas-biveinis, line 502: OK, then for output in standard formats we shouldn't be transcoding to a MySQL-subset of UTF-8 utf8mb3, but to actual UTF-8 (utf8mb4)?

gl-sergei:
- renamed the function
- changed logged query encoding from `utf8mb3` to `utf8mb4`
- added assert `errors == 0` for conversion client encoding -> `utf8mb4` for query
- added test case showing showing why database name has to be converted to `utf8mb3` when we capture it from `use` query

gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1258/

laurynas-biveinis:
Can you please elaborate on "added test case showing showing why database name has to be converted to utf8mb3 when we capture it from use query" ?

gl-sergei:
It is simple.

```
SET NAMES utf8mb4;
CREATE DATABASE 𦉘𦟌𦧲;
use 𦉘𦟌𦧲;
use test;
DROP DATABASE `???`;
```

`CREATE DATABASE` here creating database named `???`. And `use` changes database to `???`. So, the lines

```
       len= my_convert(local->db, sizeof(local->db) - 1, system_charset_info,
                        word, len, event_general->general_charset, &errors);
```

do the right thing and current database is showed as `???`

```
"Query","<ID>","<DATETIME>","change_db","<CONN_ID>",0,"use 𦉘𦟌𦧲","root[root] @ localhost []","localhost","","","???"
```

laurynas-biveinis:
I don't think we should be recording in the testcase that CREATE DATABASE <symbols> silently converts them to ???. It's all very suspicious:
- the first symbol is U+26258, thus above U+10000 as documented at http://dev.mysql.com/doc/refman/5.7/en/identifiers.html, thus illegal in an identifier (regardless of the quoting; but what happens with quoting is mildly interesting)
- if it's illegal, why doesn't server reject it?
- if it's converting it to ???, what would happen to a second db name that would also convert to same ???
- db names map to FS, what happens if character_set_filesystem is set to one of the unicode encodings?

If utf8mb4 minus utf8mb3 is illegal in identifiers and yet we still receive it in plugin, then maybe not converting it to ??? would have better audit value as that does not lose information?

gl-sergei:
Transcoding utf8 to utf8mb3 is lossy. 𦉘𦟌𦧲 becomes ???, any db name that would also convert to ??? will become ??? and ??? itself will become ???.

Another place in the docs saying that "Any character is legal in database or table identifiers except ASCII NUL (X'00'). MySQL encodes any characters that are problematic in the corresponding file system objects when it creates database directories or table files" and rules described here:

http://dev.mysql.com/doc/refman/5.7/en/identifier-mapping.html

Well, if we will encode db name in utf8mb4 in this only case we will face several issues:
-  in other cases db name comes to us in utf8m3, it makes plugins output inconsistent
- we'll need to allocate larger buffer for db name NAME_LEN defined with expectation that name is encoded in utf8mb3
- it'll make filtering by DB name buggy

The issue is MySQL-specific behaviour, it looks buggy, but I think plugin must do the same.

To record this in the test case or not -- my point is since behaviour is not obvious, it better be recorded.

laurynas-biveinis:
OK then, LGTM

Please file an upstream bug re CREATE DATABASE 'illegal characters'

gl-sergei:
reported http://bugs.mysql.com/bug.php?id=82419

PR #697, opened 2016-07-26
Fix bug 1604676 (Do not free SSL_COMP_get_compression_methods() direc…

laurynas-biveinis:
…tly)

The fix for bug 1341067 added
sk_SSL_COMP_free(SSL_COMP_get_compression_methods()) call to free some
of the heap memory allocated by OpenSSL. This is not safe for repeated
calls if OpenSSL is linked twice through different libraries and each
is trying to free the same. The proper fix would be to call
SSL_COMP_free_compression_methods(), but this API has been introduced
in newer OpenSSL versions than what we are supporting. Thus, fix by
not attempting to free compression methods at all, and add a
corresponding Valgrind suppression.

At the same time call ERR_remove_state(0) to free what's legal to
free, and fix-up/backport/add the rest of Valgrind suppressions in
order to get clean runs on SSL-related testcases.

http://jenkins.percona.com/job/percona-server-5.5-param/1262/

### DIFF
--- a/mysql-test/valgrind.supp
+++ b/mysql-test/valgrind.supp
@@ -84,17 +84,7 @@
    Memcheck:Leak
    fun:calloc
    fun:allocate_dtv
-   fun:_dl_allocate_tls_storage
-   fun:__GI__dl_allocate_tls
-   fun:pthread_create
-}
-
-{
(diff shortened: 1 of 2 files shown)


percona-ysorokin:
LGTM

PR #703, opened 2016-07-26
Ps 5.6 db 1005

george-lorch:


george-lorch:
Updated\* http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1260/

storage/tokudb/ha_tokudb_admin.cc
```diff
@@ -114,10 +113,19 @@ recount_rows_t::recount_rows_t(
...
+    snprintf(_parameters,
+             sizeof(_parameters),
+             "TOKUDB_ANALYZE_THROTTLE=%llu;",
+             _throttle);
+    _status[0] = '\0';
 }
 recount_rows_t::~recount_rows_t() {
 }
 void recount_rows_t::on_run() {
+    const char* orig_proc_info = NULL;
```
laurynas-biveinis, line 126: nullptr ?

storage/tokudb/ha_tokudb_admin.cc
```diff
@@ -351,13 +371,27 @@ standard_t::standard_t(
...
+
+    _status[0] = '\0';
 }
 standard_t::~standard_t() {
 }
 void standard_t::on_run() {
     DB_BTREE_STAT64 stat64;
     uint64_t rec_per_key_part[_share->_max_key_parts];
     uint64_t total_key_parts = 0;
+    const char* orig_proc_info = NULL;
```
laurynas-biveinis, line 391: nullptr ?

storage/tokudb/ha_tokudb_admin.cc
```diff
@@ -546,19 +543,69 @@ int standard_t::analyze_key_progress(void) {
         if ((_thd && thd_killed(_thd)) || cancelled()) {
             // client killed
             return ER_ABORTING_CONNECTION;
-        } else if(_time_limit > 0 &&
-                  (uint64_t)_key_elapsed_time > _time_limit) {
+        } else if (_time_limit > 0 &&
+                   (uint64_t)_key_elapsed_time > _time_limit) {
```
laurynas-biveinis: static_cast ?

storage/tokudb/ha_tokudb_admin.cc
```diff
@@ -546,19 +543,69 @@ int standard_t::analyze_key_progress(void) {
...
+                scan_direction = scan_direction_str[3];
+                break;
         }
 
+        float progress_rows = 0.0;
+        if (_share->row_count() > 0)
+            progress_rows = (float)_rows / (float)_share->row_count();
+        float progress_time = 0.0;
+        if (_time_limit > 0)
+            progress_time = (float)_key_elapsed_time / (float)_time_limit;
```
laurynas-biveinis: static_cast ?

storage/tokudb/ha_tokudb_admin.cc
```diff
@@ -43,13 +43,11 @@ class recount_rows_t : public tokudb::background::job_manager_t::job_t {
...
-        char* database,
-        char* table,
-        char* type,
-        char* params,
-        char* status);
+    virtual const char* database();
+    virtual const char* table();
+    virtual const char* type();
+    virtual const char* parameters();
+    virtual const char* status();
```
laurynas-biveinis, line 50: const methods?

storage/tokudb/tokudb_background.h
```diff
@@ -58,13 +58,20 @@ class job_manager_t {
...
 
-        // method to get info for information schema, 255 chars per buffer
-        virtual void status(
-            char* database,
-            char* table,
-            char* type,
-            char* params,
-            char* status) = 0;
+        // method to obtain the database name the job is scheduled on
+        virtual const char* database() = 0;
```
laurynas-biveinis, line 62: const methods?

storage/tokudb/ha_tokudb_admin.cc
george-lorch, line 50: No, specifically not const as they may need to alter derived class state in order to provide their responses. That _could_ be overridden in the derived class but then violates the stated intention of the API if defined as const.

storage/tokudb/ha_tokudb_admin.cc
george-lorch: Hmm, yeah, good catch, the code existed before (where I should have used static_cast then as well) and was just relocated for this issue, but relocation/touching is grounds for correcting anything to current standards.

storage/tokudb/ha_tokudb_admin.cc
george-lorch, line 126: Yes, it is correct, but it is also not correct :) This should be using the tokudb wrappers of tokudb_thd_get/set_proc_info that return and accept const char\* pointers instead of the direch thd_\* calls.

PR #705, opened 2016-07-27
Fix bug 1606572 (Spurious "[Warning] Slave I/O: SET @master_heartbeat…

laurynas-biveinis:
…_period to master failed (...)" caused by STOP SLAVE while START SLAVE in progress)

Call check_io_slave_killed and return early if needed before issuing
diagnostics if SET @master_heartbeat_period query to master failed.

http://jenkins.percona.com/job/percona-server-5.5-param/1270/?auto_refresh=true


percona-ysorokin:
LGTM

PR #708, opened 2016-07-27
Fix bug 1606782 (rpl_percona_bug1070255 may crash due to concurrent d…

laurynas-biveinis:
…bug access)

The --source include/rpl_stop.inc is preceded by set global DEBUG='';
executed with the slaves running. Thus, this has the same underlying
issue as upstream bug 58754 / 11765758 and is fixed the same way, by
executing SET GLOBAL DEBUG=... with a stopped slave.


percona-ysorokin:
LGTM

PR #711, opened 2016-07-28
Fix bug 1606782 (Shutdown hangs if redo log tracker thread stopped)

laurynas-biveinis:
If the log tracking stops due to margin violation or any other reason
(user request in debug build; I/O error), then shutdown will wait for
a long time for the log tracker thread to quit, which it never does.

This is caused by the tracking margin code setting
srv_track_changed_pages = FALSE as tracking cannot continue, and then
never raising the tracking event so that the tracker thread can wake
up and quit.

The root cause is srv_track_changed_pages attempting to handle two
different purposes: 1) whether the tracking is currently active; 2)
whether the tracker thread has been started.

Fix by splitting off srv_redo_log_thread_started flag out of
srv_track_changed_pages, and using it where appropriate.

```
http://jenkins.percona.com/job/percona-server-5.5-param/1275/
```

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -12539,7 +12539,6 @@ innodb_track_changed_pages_validate(
 						for update function */
 	struct st_mysql_value*		value)	/*!< in: incoming bool */
 {
-	static bool     enabled_on_startup = false;
 	long long	intbuf = 0;
 
 	if (value->val_int(value, &intbuf)) {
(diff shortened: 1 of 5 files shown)


percona-ysorokin:
What about `init_log_online(void)` in `srv0start.c`?
Do we need to change `srv_track_changed_pages` to `srv_redo_log_thread_started` there as well?

laurynas-biveinis:
No: srv_redo_log_thread_started is set in the tracker thread. init_log_online is what spawns that thread, based on the startup innodb_track_changed_pages value in srv_track_changed_pages.

percona-ysorokin:
LGTM

PR #720, opened 2016-07-28
Fix bug 1607359 (rpl_4threads_deadlock may crash due to concurrent db…

laurynas-biveinis:
…ug access)

A client disconnect and SET DEBUG=... running in parallel race in dbug
data structure access. Add synchronisation so that disconnects are
completed before SET DEBUG stats.

```
http://jenkins.percona.com/job/percona-server-5.5-param/1279/
```


percona-ysorokin:
LGTM

PR #723, opened 2016-07-28
Fix bug 1097870 (test rpl_cant_read_event_incident can fail sporadica…

laurynas-biveinis:
…lly)

Disable warnings around "drop table if exists" so that its output is
identical for both existing and non-existing table case.

http://jenkins.percona.com/job/percona-server-5.5-param/1280/


percona-ysorokin:
LGTM

PR #726, opened 2016-07-29
Work towards diagnosing bug 1607606 (test percona_extended_innodb_sta…

laurynas-biveinis:
…tus test is unstable)

This test sometimes fails with apparently different record lock count
than expected. Make the test dump the whole SHOW ENGINE INNODB STATUS
output if that happens.

```
http://jenkins.percona.com/job/percona-server-5.5-param/1284/
```


percona-ysorokin:
LGTM

PR #729, opened 2016-07-29
Work towards diagnosing bug 1607607 (test percona_show_temp_tables_st…

laurynas-biveinis:
…ress is unstable)

If a mysqltest MySQL query failed with ER_NO_SUCH_THREAD, and the
error is not expected, dump the PROCESSLIST to show the available
thread ids and other info at the time.

```
http://jenkins.percona.com/job/percona-server-5.5-param/1283/
```


percona-ysorokin:
LGTM

PR #732, opened 2016-07-29
Fix bug 1607671 (Test rpl.rpl_concurrency_error is unstable)

laurynas-biveinis:
The triggers in the test run on the master and slave independently,
and so may happen to get different timestamps for the same event,
which then causes differences in table n contents between master and
the slave.

Fix by replacing the timestamp column with an auto_increment one.

http://jenkins.percona.com/job/percona-server-5.5-param/1285/


percona-ysorokin:
LGTM

PR #736, opened 2016-07-30
Allow to log only queries for specific DBs in audit log plugin

gl-sergei:
Blueprint:

https://blueprints.launchpad.net/percona-server/+spec/audit-filer-db

Add two global variables:
-   `audit_log_include_databases` comma separated list of databases to
  include in audit logging
-   `audit_log_exclude_databases` comma separated list of databases to
  exclude from audit logging

Allow escaped backticks in the `next_word` function. This change
needs to be backported to the 5.6.

### DIFF
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -66,13 +66,17 @@ ulong audit_log_syslog_facility= 0;
 ulong audit_log_syslog_priority= 0;
 static char *audit_log_exclude_accounts= NULL;
 static char *audit_log_include_accounts= NULL;
+static char *audit_log_exclude_databases= NULL;
+static char *audit_log_include_databases= NULL;
 
 PSI_memory_key key_memory_audit_log_logger_handle;
 PSI_memory_key key_memory_audit_log_handler;
 PSI_memory_key key_memory_audit_log_buffer;
 PSI_memory_key key_memory_audit_log_record_buffer;
+PSI_memory_key key_memory_audit_log_query_stack;
 PSI_memory_key key_memory_audit_log_thd_local;
 PSI_memory_key key_memory_audit_log_accounts;
+PSI_memory_key key_memory_audit_log_databases;
 
 static PSI_memory_info all_audit_log_memory[]=
 {
@@ -80,8 +84,10 @@ static PSI_memory_info all_audit_log_memory[]=
   {&key_memory_audit_log_handler, "audit_log_handler", 0},
   {&key_memory_audit_log_buffer, "audit_log_buffer", 0},
   {&key_memory_audit_log_record_buffer, "audit_log_record_buffer", 0},
+  {&key_memory_audit_log_query_stack, "audit_log_query_stack", 0},
   {&key_memory_audit_log_thd_local, "audit_log_thd_local", 0},
   {&key_memory_audit_log_accounts, "audit_log_accounts", 0},
+  {&key_memory_audit_log_databases, "audit_log_databases", 0},
 };
 
 static int audit_log_syslog_facility_codes[]=
@@ -771,6 +777,25 @@ int reopen_log_file()
   return(0);
 }
 
+typedef struct
+{
+  /* number of included databases */
+  int databases_included;
+  /* number of excluded databases */
+  int databases_excluded;
+  /* number of accessed databases */
+  int databases_accessed;
+  /* query */
+  const char *query;
+} query_stack_frame;
+
+typedef struct
+{
+  size_t size;
+  size_t top;
+  query_stack_frame *frames;
+} query_stack;
+
 /*
  Struct to store various THD specific data
  */
@@ -778,12 +803,16 @@ typedef struct
 {
   /* size of allocated large buffer to for record formatting */
   size_t record_buffer_size;
-  /* skip logging session */
-  my_bool skip_logging;
+  /* skip session logging */
+  my_bool skip_session;
+  /* skip logging for the next query */
+  my_bool skip_query;
   /* default database */
   char db[NAME_LEN + 1];
   /* default database candidate */
   char init_db_query[NAME_LEN + 1];
+  /* call stack */
+  query_stack stack;
 } audit_log_thd_local;
 
 /*
@@ -797,6 +826,12 @@ audit_log_thd_local *get_thd_local(MYSQL_THD thd);
  */
 static
 char *get_record_buffer(MYSQL_THD thd, size_t size);
+
+/*
+ Allocate and return given number of stack frames.
+ */
+static
+query_stack_frame *realloc_stack_frames(MYSQL_THD thd, size_t size);
 
 
 static
@@ -825,7 +860,7 @@ int audit_log_plugin_init(MYSQL_PLUGIN plugin_info)
 
 
 static
-int audit_log_plugin_deinit(void *arg __attribute__((unused)))
+int audit_log_plugin_deinit(void *arg MY_ATTRIBUTE((unused)))
 {
   char buf[1024];
   size_t len;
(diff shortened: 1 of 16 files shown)


gl-sergei:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/275/

plugin/audit_log/audit_log.c
```diff
@@ -732,10 +736,18 @@ typedef struct
   size_t record_buffer_size;
   /* skip logging session */
   my_bool skip_logging;
```
laurynas-biveinis: Maybe rename to skip_session to avoid ambiguity with skip_query?

plugin/audit_log/audit_log.c
```diff
@@ -732,10 +736,18 @@ typedef struct
...
+  /* skip logging next query */
+  my_bool skip_query;
   /* default database */
   char db[NAME_LEN + 1];
   /* default database candidate */
   char init_db_query[NAME_LEN + 1];
+  /* number of included databases */
+  my_bool databases_included;
```
laurynas-biveinis: These three definitely should not be "my_bool"

plugin/audit_log/audit_log.c
```diff
@@ -906,6 +926,29 @@ my_bool audit_log_update_thd_local(MYSQL_THD thd,
     const struct mysql_event_general *event_general=
       (const struct mysql_event_general *) event;
 
+    if (event_general->event_subclass == MYSQL_AUDIT_GENERAL_STATUS)
+    {
+      local->skip_query= FALSE;
```
laurynas-biveinis: I like direct assignments of bool-like vars, i.e. here

local->skip_query= (first if condition)
local->skip_query|= (second if condition)

but it's matter of taste

plugin/audit_log/audit_log.c
```diff
@@ -1323,6 +1380,115 @@ static MYSQL_SYSVAR_STR(include_accounts, audit_log_include_accounts,
        audit_log_include_accounts_validate,
        audit_log_include_accounts_update, NULL);
 
+static
+int
+audit_log_exclude_databases_validate(
+          MYSQL_THD thd __attribute__((unused)),
```
laurynas-biveinis: MY_ATTRIBUTE ?

plugin/audit_log/filter.c
```diff
@@ -167,13 +263,26 @@ void audit_log_filter_init()
                20, 0, 0,
                (my_hash_get_key) account_get_key,
                my_free, HASH_UNIQUE, key_memory_audit_log_accounts);
+
+  my_hash_init(&include_databases, &my_charset_utf8_general_ci,
```
laurynas-biveinis: I think here (and for account filter hashes) there is character set collation overhead on each and every lookup? Would it be possible to create hashes with one of the dumb character sets, and convert to it on insert to hash?

plugin/audit_log/tests/mtr/audit_log_filter_db_events.inc
```diff
@@ -0,0 +1,30 @@
+
+
+CREATE TABLE db1.t (a INT);
```
laurynas-biveinis, line 4: I'd add some more tests (strictly speaking to test MYSQL_AUDIT_TABLE_ACCESS_CLASS notifications and not this plugin, but if we ship this we have to have confidence if the whole thing works):
- multi-table delete with same table listed twice
- trigger on a table that writes to a different table
- table access through event scheduler (actually this one is interesting to test outside the db filtering context too)

gl-sergei:
Adding a trigger complicates things a bit. For example statements

```
CREATE TRIGGER ins_tring BEFORE INSERT ON db1.trig FOR EACH ROW INSERT INTO db2.t VALUES (new.a + 100);
INSERT INTO trig VALUES (1), (2), (3);
```

will generate following sequence if events:

```
table access event to `db1`, query = "INSERT INTO trig VALUES (1), (2), (3)"
table access event to `db2`, query = "INSERT INTO trig VALUES (1), (2), (3)"
table access event to `db2`, query = "INSERT INTO db2.t VALUES (new.a + 100)"
general status event, query = "INSERT INTO db2.t VALUES (new.a + 100)"
table access event to `db2`, query = "INSERT INTO db2.t VALUES (new.a + 100)"
general status event, query = "INSERT INTO db2.t VALUES (new.a + 100)"
table access event to `db2`, query = "INSERT INTO db2.t VALUES (new.a + 100)"
general status event, query = "INSERT INTO db2.t VALUES (new.a + 100)"
general status event, query = "INSERT INTO trig VALUES (1), (2), (3)"
```

If plugin doesn't take query into account when updating access information,
it will produce wrong results.

We can use stack of hash table to make sure we update access information
for the correct query.

gl-sergei:
Event scheduler seems to work OK, except for it doesn't set default database. One more weird thing is that the statement logged with `sql_command` equals `create_procedure`

gl-sergei:
- added test with stored procedure
- added test with triggers
- added test with event scheduler
- added test with multi-table delete
- use stack to remember which query we are working with at the moment
- use `charset_bin' instead of`charset_utf8_general_ci', which makes
  everything case-sensitive, but has faster hash computation
- replaced my_bool with int for DB access counters (it was a mistake :)

gl-sergei:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/290/

laurynas-biveinis:
One more case which would need checking is querying / updating a VIEW

gl-sergei:
done

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/293/

gl-sergei:
Changed behaviour of user and database filters. User filter becomes case sensitive for user name and case insensitive for host name. Database filter becomes case insensitive.

plugin/audit_log/audit_log.c
```diff
@@ -954,6 +997,36 @@ my_bool audit_log_update_thd_local(MYSQL_THD thd,
...
+
+      local->stack.frames[local->stack.top].databases_included= 0;
+      local->stack.frames[local->stack.top].databases_accessed= 0;
+      local->stack.frames[local->stack.top].databases_excluded= 0;
+      local->stack.frames[local->stack.top].query= NULL;
+
+      if (local->stack.top > 0)
+        --local->stack.top;
+
+      if ((event_general->general_command.length == 4 &&
```
laurynas-biveinis: if (!local->skip_query && (...) to bypass evaluation if we are already skipping the query?

plugin/audit_log/filter.c
```diff
@@ -88,6 +101,44 @@ uchar *account_get_key(const account *acc, size_t *length,
...
+  database_init(db, name, length);
+
+  return db;
+}
+
+/*
+  Get database key
+*/
+uchar *database_get_key(const database *db, size_t *length,
+                        my_bool not_used __attribute__((unused)))
```
laurynas-biveinis: MY_ATTRIBUTE

plugin/audit_log/tests/mtr/audit_log_filter_users.test
```diff
@@ -71,6 +73,13 @@ SET GLOBAL audit_log_exclude_accounts= 'user1@localhost,, user22@127.0.0.1,admin
 
 SET GLOBAL audit_log_exclude_accounts= NULL;
 
+# testing case sensitivity
```
laurynas-biveinis, line 74: Case sensitivity also needs testing for the username part too: Foo vs foo, filter just one of them

gl-sergei:
added test sensitivity test for username
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/308/

gl-sergei:
added missing

```
  my_free(audit_log_include_databases);
  my_free(audit_log_exclude_databases);
```

in `audit_log_plugin_deinit`

laurynas-biveinis:
LGTM

PR #743, opened 2016-08-01
Fix bug 1608385 (innodb.percona_changed_page_bmp does not cleanup $MY…

laurynas-biveinis:
…SQLTEST_VARDIR/tmpdatadir)

Remove the created temp directory at the end of testcase. At the same
time replace some of the manual server shutdown/startup sequences with
include files.

http://jenkins.percona.com/job/percona-server-5.5-param/1290/


percona-ysorokin:
LGTM

PR #746, opened 2016-08-01
Fix bug 1608437 (MTR does not register unit test failures with Subunit)

laurynas-biveinis:
Register test results with Subunit for CTest and Valgrind too.

http://jenkins.percona.com/job/percona-server-5.5-param/1291/


percona-ysorokin:
LGTM

PR #748, opened 2016-08-01
Bug1608437 5.7

laurynas-biveinis:


laurynas-biveinis:
Merged through https://github.com/percona/percona-server/pull/747, which has wrong target branch

PR #750, opened 2016-08-01
Fix bug 1608515 (test group_commit is unstable)

laurynas-biveinis:
The test has a race condition where con5 catches the previous commit
group and attaches to it instead of becoming a leader of a new group.

Fix by applying a patch from MariaDB:

```
revno: 2732.26.19
committer: knielsen@knielsen-hq.org
branch nick: work-5.2-rpl
timestamp: Thu 2011-04-14 13:23:02 +0200
message:
  Fix race in testcase innodb_plugin.group_commit

  The problem was that connection con5, which is supposed to be the leader for
  the third group, could if timing was bad end up as participant in the second
  group, causing the test to fail.

  Fixed by ensuring that group 2 is running before starting the transaction
  for group 3.
```

http://jenkins.percona.com/job/percona-server-5.5-param/1293/


percona-ysorokin:
LGTM

PR #759, opened 2016-08-02
Fix bug 1608845 (Test percona_show_slave_status_nolock is unstable)

laurynas-biveinis:
Rewrite the testcase to avoid SET GLOBAL DEBUG=... in any context
where anything asynchronuous (such as slave threads) might be running,
and to simplify the testcase.

http://jenkins.percona.com/job/percona-server-5.5-param/1297/


percona-ysorokin:
LGTM

PR #762, opened 2016-08-02
Work towards fixing bug 756387 (Intermittent innodb.innodb_bug60049 f…

laurynas-biveinis:
…ailures)

Previous work added diagnostics if MTR server shutdown times out and
kill -9 is issued. Here the testcases are cleaned up to ensure clean
shutdowns according to those diagnostics, mostly by backporting 5.7
changes:
- mysql-test/include/restart_mysqld.inc: use the default shutdown
  timeout value (60 seconds) instead of 10;
- mysql-test/include/rpl_stop_server.inc: use 60 second timeout
  instead of 10, further multiply it by 6 if running under Valgrind;
- mysql-test/suite/sys_vars/t/innodb_read_ahead_tartup.test: rewrite
  to use restart_mysqld.inc.

http://jenkins.percona.com/job/percona-server-5.5-param/1296/

### DIFF
--- a/mysql-test/include/restart_mysqld.inc
+++ b/mysql-test/include/restart_mysqld.inc
@@ -19,8 +19,9 @@ if (!$restart_parameters)
 --exec echo "wait" > $_expect_file_name
 
 # Send shutdown to the connected server and give
-# it 10 seconds to die before zapping it
-shutdown_server 10;
+shutdown_server;
+
+--source include/wait_until_disconnected.inc
 
 # Write file to make mysql-test-run.pl start up the server again
 --exec echo "$restart_parameters" > $_expect_file_name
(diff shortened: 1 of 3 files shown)


mysql-test/include/restart_mysqld.inc
```diff
@@ -19,8 +19,9 @@ if (!$restart_parameters)
 --exec echo "wait" > $_expect_file_name
 
 # Send shutdown to the connected server and give
```
percona-ysorokin, line 21: truncated comment

percona-ysorokin:
LGTM

mysql-test/include/restart_mysqld.inc
laurynas-biveinis, line 21: I'd have to do redo 5.6 and 5.7 to fix this one, I'm tempted to leave it as-is

PR #766, opened 2016-08-03
Fix bug 1607606 (percona_extended_innodb_status test is unstable)

laurynas-biveinis:
If the MTR worker happened to run a test that resulted in a foreign
key errors, SHOW ENGINE INNODB STATUS will show this error and its
associated record locks, which will be counted too in this testcase,
failing it.

Fix by counting record locks only starting with "TRANSCATION" section
in the InnoDB status.

http://jenkins.percona.com/job/percona-server-5.5-param/1298/


percona-ysorokin:
LGTM

PR #773, opened 2016-08-04
Fix bug 1609523 (Needless sleep between purge thread iterations durin…

laurynas-biveinis:
…g shutdown)

After bug 1609364 fix, the 5.5 separate purge thread tries to purge up
to TRX_SYS_N_RSEGS times when there is nothing to purge before
quitting, but it sleeps needlessly for up to 1 second between these
iterations, delaying the shutdown. Fix by skipping the sleep if we are
in shutdown.

http://jenkins.percona.com/job/percona-server-5.5-param/1301/


percona-ysorokin:
LGTM

PR #776, opened 2016-08-04
Fix bug 1609422 (Unit test lf-t unstable)

laurynas-biveinis:
Adjust lf-t test threads so that they call my_thread_end before
signalling the main thread that test is done, in order to avoid race
on with_my_thread_init flag, which would cause
my_thread_init/my_thread_end mismatch. Remove obsolete "sleep(2)" from
the test framework.

http://jenkins.percona.com/job/percona-server-5.5-param/1303/

### DIFF
--- a/unittest/mysys/lf-t.c
+++ b/unittest/mysys/lf-t.c
@@ -48,12 +48,14 @@ pthread_handler_t test_lf_pinbox(void *arg)
     pins= lf_pinbox_get_pins(&lf_allocator.pinbox);
   }
   lf_pinbox_put_pins(pins);
+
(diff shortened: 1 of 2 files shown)


percona-ysorokin:
the comment is awesome
`workaround until we know why it crashes randomly on some machine`
It looks like now we do
LGTM

laurynas-biveinis:
The comment is referring to https://bugs.mysql.com/bug.php?id=22320, which boils down to that it's impossible to wrap CMPXCHG8B in GCC inline asm :), not a problem if GCC atomic builtins are available

PR #779, opened 2016-08-04
Fix bug 1607607 (test percona_show_temp_tables_stress is unstable)

laurynas-biveinis:
In the testcase, one thread queries
INFORMATION_SCHEMA.GLOBAL_TEMPORARY_TABLES in an infinite loop, called
from a stored procedure. The results of this query are never
collected, causing a network write failure from the server if the
testcase is running long enough.

Rewrite this testcase to use the event scheduler so that network I/O
is avoided.

http://jenkins.percona.com/job/percona-server-5.5-param/1304/
The test results expose a crash that was fixed in a later 5.5, tested locally with 5.5 trunk too.

### DIFF
--- a/mysql-test/t/percona_show_temp_tables_stress.test
+++ b/mysql-test/t/percona_show_temp_tables_stress.test
@@ -1,0 +1,50 @@
+--source include/have_innodb.inc
+
+CREATE USER event_runner@localhost;
+
+SET @saved_event_scheduler = @@GLOBAL.event_scheduler;
+SET GLOBAL event_scheduler = ON;
+
+delimiter |;
+CREATE DEFINER=event_runner@localhost EVENT query_temp_tables ON SCHEDULE AT CURRENT_TIMESTAMP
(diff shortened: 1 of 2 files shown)


mysql-test/t/percona_show_temp_tables_stress.test
```diff
@@ -0,0 +1,50 @@
...
+--enable_query_log
+
+--let $ev_thread_id= `SELECT ID FROM INFORMATION_SCHEMA.PROCESSLIST WHERE USER='event_runner'`
+--echo KILL QUERY \$ev_thread_id
+--disable_query_log
+eval KILL QUERY $ev_thread_id;
+--enable_query_log
+
+SET GLOBAL event_scheduler = @saved_event_scheduler;
+
```
percona-ysorokin, line 49: missing `DROP EVENT query_temp_tables`

percona-ysorokin:
LGTM

PR #782, opened 2016-08-04
DB-1006 : Cardinality=0 for all rows in a TokuDB table

george-lorch:
FT-732 : logical row counts inaccurate
- Advanced PerconaFT submodule commit pointer to point to recent master commit for FT-732.
- Minor format correction.
- Minor snprintf message correction.
- Fixed potential cardinality 0 issue if ANALYZE TABLE finds only deleted rows and no actual logical rows before it times out.


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1282/

PR #785, opened 2016-08-06
Bug 1610242: Reading past the end of heap buffer on saving the defaul…

gl-sergei:
…t DB in audit plugin

memcpy in the call

memcpy(local->db, event_general->general_query, sizeof(local->db));

always copies `NAME_LEN` bytes, but the real length of the
`general_query` can be less than `NAME_LEN`

Fix is to copy `general_query_length` bytes.


gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-asan-param/16/

laurynas-biveinis:
LGTM

PR #787, opened 2016-08-08
Bug#22173419 VALGRIND FAILURE IN MAIN.GROUP_MIN_MAX

laurynas-biveinis:
Query below produces valgrind warning:

select a1,a2,b,min(c) from t2 where
((a1 > 'a') or (a1 < '9')) and (a2 >= 'b') and (b = 'a') and (c < 'h112')
group by a1,a2,b;

Problematic expession is (c < 'h112').
SEL_ARG::min_range is is_null_string which has length=2, memcmp
is executed with the length which is bigger than is_null_string.

The fix: Do not perform comparison if one of the argument is NULL value.

http://jenkins.percona.com/job/percona-server-5.5-param/1307/


percona-ysorokin:
LGTM

PR #790, opened 2016-08-08
Fix bug 1610858 (Some MTR tests cause server kill on shutdown with Va…

laurynas-biveinis:
…lgrind)

Replace custom shutdown/startup MTR sequences with include file
use. This also removes hardcoded 10 second shutdown timeout in those
places, and using the default 60 second one, which is more appropriate
for running under Valgrind.

http://jenkins.percona.com/job/percona-server-5.5-param/1308/


percona-ysorokin:
LGTM

PR #793, opened 2016-08-09
DB-1007 : Extend percona_show_temp_tables_stress to handle TokuDB

george-lorch:
- Created new include file that will indicate if TokuDB _should_ be present.
- Altered test to use TokuDB along with other engines if TokuDB should be presentwith consideration to run successfully if there is no TokuDB.

### DIFF
--- a/mysql-test/include/has_tokudb.inc
+++ b/mysql-test/include/has_tokudb.inc
@@ -1,0 +1,18 @@
+#
+# Script tests if TokuDB _should_ be present.
+# Sets $HAS_TOKUDB to either 0 or 1
+#
+--let $HAS_TOKUDB=0
+#
+# Check if server has support for loading plugins
+#
+if (`SELECT @@have_dynamic_loading = 'YES'`) {
+  if ($TOKUDB) {
+    #
+    # Check if --plugin-dir was setup for TokuDB
+    #
+    if (`SELECT CONCAT('--plugin-dir=', REPLACE(@@plugin_dir, '\\\\', '/')) = '$TOKUDB_OPT/'`) {
+      --let $HAS_TOKUDB=1
+    }
+  }
+}
--- a/mysql-test/t/percona_show_temp_tables_stress-master.opt
+++ b/mysql-test/t/percona_show_temp_tables_stress-master.opt
@@ -1,0 +1,1 @@
+$TOKUDB_OPT $TOKUDB_LOAD_ADD --loose-tokudb-check-jemalloc=0 --loose-tokudb-cache-size=512M --loose-tokudb-block-size=1M
--- a/mysql-test/t/percona_show_temp_tables_stress.test
+++ b/mysql-test/t/percona_show_temp_tables_stress.test
@@ -1,5 +1,7 @@
 --source include/have_innodb.inc
 
+--source include/has_tokudb.inc
+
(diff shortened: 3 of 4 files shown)


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1287/

mysql-test/t/percona_show_temp_tables_stress.test
```diff
@@ -18,18 +20,35 @@ send CALL query_temp_tables();
...
 }
 --enable_query_log
-
-SELECT COUNT(*) FROM INFORMATION_SCHEMA.GLOBAL_TEMPORARY_TABLES;
+--let $table_count=`SELECT COUNT(*) FROM INFORMATION_SCHEMA.GLOBAL_TEMPORARY_TABLES`
+if ($HAS_TOKUDB == 0)
+{
+  if ($table_count != 1200)
+  {
+    --die ERROR IN TEST: Expected count of I_S.GLOBAL_TEMPORARY_TABLES to be 1200.
```
laurynas-biveinis, line 42: Looks like a job for include/assert.inc ?

mysql-test/include/has_tokudb.inc
```diff
@@ -0,0 +1,18 @@
+#
```
laurynas-biveinis, line 1: Now we have have_tokudb.inc and has_tokudb.inc and it's really hard to tell the difference from the filename which one is supposed to do what. have_tokudb_optional.inc ?

mysql-test/t/percona_show_temp_tables_stress-master.opt
```diff
@@ -0,0 +1 @@
+$TOKUDB_OPT $TOKUDB_LOAD_ADD --loose-tokudb-check-jemalloc=0 --loose-tokudb-cache-size=512M --loose-tokudb-block-size=1M
```
laurynas-biveinis, line 1: hm, does "$TOKUDB_OPT $TOKUDB_LOAD_ADD" work w/o TokuDB here?

mysql-test/include/has_tokudb.inc
george-lorch, line 1: I agree and had a bit of a hard time myself coming up with a name for it. Has seemed logical because of the precedent set by 'have' meaning "must have or fail". One of the two most difficult issues in programming :)

mysql-test/t/percona_show_temp_tables_stress-master.opt
george-lorch, line 1: Yes, according to the comments in include/plugin.defs:

> # If a listed plugin is not found, the corresponding variables will be
> 
> # set to empty, they will not be unset.

So if set to empty, they don't exist as options here and the positive results can be seen in the jenkins run. I actually tested the test locally to ensure that the TokuDB portion of the test is executed and/or not executed correctly if the plugin is not present by adding some extra 'echo's and changing the expected values of the table_counts to force the test to fail for both the TokuDB and non TokuDB cases.

mysql-test/t/percona_show_temp_tables_stress.test
george-lorch, line 42: Yup, that seems reasonable, I wasn't aware of it, I'll change that up.

PR #795, opened 2016-08-09
Fix MTR --repeat=2 failures (bug 1610102 and bug 1610110)

laurynas-biveinis:
For upstream failures:
- main.auth_rpl: CHANGE MASTER TO in cleanup resets the replication
  user name, but not the password, leaving the slave unable to connect;
- binlog.bug36391: clear binlog at the start of the test as it's
  replayed in the whole;
- parts.partition_special_myisam, main.func_str, main.loaddata,
  rpl.rpl_binlog_errors: cleanup the temp files.
- rpl.rpl_rotate_logs: cleanup master.index through RESET SLAVE.
- main.temp_pool: force server restart to get known temp file
  number.
- rpl.rpl_row_find_row_debug: do not grep the main error file
  as its previous contents are unpredictable. Restart the server for the
  test with the test-specific error file and grep that instead. Add
  arg-hiding option to include/rpl_restart_server.inc for this.
- innodb.innodb_bug12400341: the test fills up whole undo space and
  does not purge it for the next test. Force server restart with a slow
  shutdown to clean up.
- main.group_commit_binlog_pos (MariaDB upstream): RESET MASTER for
  repeatable binlog position

For Percona Server-specific failures:
- main.percona_bug643149: cleanup the testcase in general, remove the
  slow log file at the end
- main.percona_log_connection_error,
  main.percona_log_warnings_suppress: similar fixes as for
  rpl.rpl_row_find_row_debug.
- main.percona_ssl_connections_count: flush userstat before test
  start.
- main.percona_bug1016991: enable slow_query_log_timestamp_always, so
  that the quick test execution would not disab

http://jenkins.percona.com/job/percona-server-5.5-param/1309/


percona-ysorokin:
LGTM

PR #800, opened 2016-08-10
Fix bug 1611229 (Test mysql_plugin intermittently fails to shutdown t…

laurynas-biveinis:
…he server in 10 seconds)

Clean up MTR testcases not to specify non-default shutdown_server
timeout, if greater than zero. For upstream testcases, this is mostly
[1], with some follow-up cleanup not handled by that commit. For our
testcases, use include files as appropriate.

[1]:

commit 5505f669979a0504988c36fa1e13b42daeda1b6e
Author: Anitha anitha.gopi@oracle.com
Date:   Thu Dec 18 09:30:01 2014 +0100

```
Bug#20225524 TESTS SHOULD USE AT LEAST 60 SEC TIMEOUT FOR SHUTDOWN_SERVER
COMMAND

Removed timeouts less than the default from all shutdown_server commands in the
.tets and .inc files
```

http://jenkins.percona.com/job/percona-server-5.5-param/1311/

### DIFF
--- a/mysql-test/suite/innodb/t/percona_changed_page_bmp_1364315.test
+++ b/mysql-test/suite/innodb/t/percona_changed_page_bmp_1364315.test
@@ -6,9 +6,7 @@
 
 --let $TMPDATADIR= $MYSQLTEST_VARDIR/tmpdatadir_1364315
 
---exec echo "wait" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
---shutdown_server 10
---source include/wait_until_disconnected.inc
+--source include/shutdown_mysqld.inc
 --mkdir $TMPDATADIR
 --enable_reconnect
 --echo 1st restart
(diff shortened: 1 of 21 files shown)


mysql-test/suite/innodb/t/percona_changed_page_bmp_1364315.test
```diff
@@ -6,9 +6,7 @@
 
 --let $TMPDATADIR= $MYSQLTEST_VARDIR/tmpdatadir_1364315
 
---exec echo "wait" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
---shutdown_server 10
---source include/wait_until_disconnected.inc
+--source include/shutdown_mysqld.inc
 --mkdir $TMPDATADIR
 --enable_reconnect
 --echo 1st restart
```
percona-ysorokin, line 12: `--source include/start_mysqld.inc` with parameters instead of `--source include/wait_until_connected_again.inc` here?

percona-ysorokin:
LGTM with minor comment

mysql-test/suite/innodb/t/percona_changed_page_bmp_1364315.test
laurynas-biveinis, line 12: It would result in $TMPDATADIR absolute path recorded in the test output. 5.7 can do --replace_result $FOO FOO for start_mysqld.inc, but this one cannot

PR #804, opened 2016-08-10
Fix for bug 1544443 (5.6)

gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1289/


laurynas-biveinis:
LGTM

PR #809, opened 2016-08-11
Work towards fixing bug 1612076 (Test rpl.rpl_cant_read_event_inciden…

laurynas-biveinis:
…t is unstable)

The testcase fails with STOP SLAVE indicating that both slave threads
have already been stopped, when the SQL thread should have been
running. Record the running SQL thread SHOW SLAVE STATUS, so that the
next testcase failure would show the SQL thread stopping reason.

http://jenkins.percona.com/job/percona-server-5.5-param/1323/


percona-ysorokin:
LGTM

PR #813, opened 2016-08-11
Fix bug 1612084 (MTR include/wait_until_connected_again.inc may timeo…

laurynas-biveinis:
…ut too soon for Valgrind)

Backport the 5.7 wait_until_connected_again.inc timeout, which is
significantly bumped for the general case (and fix the attempt to bump
it further if running under Valgrind).

http://jenkins.percona.com/job/percona-server-5.5-param/1320/


percona-ysorokin:
LGTM

PR #816, opened 2016-08-11
Bug 1608902: Constant buffer size set on max_nss_name_len is not enou…

gl-sergei:
…gh to get results from getgrgid_r if there are thousands of members in a group

PAM plugin makes a list of the names of the groups which the user belongs
to as following:
- receive the list of group ids with `getgrouplist`
- for each group id call `getgrgid_r` to get its name

There are two issues:
- Fixed length array passed to `getgrouplist`. If its size isn't enough,
  plugin aborts authentication. Fixed by reallocating the array using
  the number of groups reported by the unsuccessful call and retrying
  the call.
- `getpwnam_r` and `getgrgid_r` return `ERANGE` in case of insufficient
  buffer.
  
  Fix:
  1. Use values of `sysconf(_SC_GETGR_R_SIZE_MAX)` and
     `sysconf(_SC_GETPW_R_SIZE_MAX)` as initial buffer size.
  2. In case of `ERANGE` allocate the buffer with the double size and
     try again until the result is different from `ERANGE`.

### DIFF
--- a/plugin/percona-pam-for-mysql/src/groups.c
+++ b/plugin/percona-pam-for-mysql/src/groups.c
@@ -1,5 +1,5 @@
 /*
-(C) 2013 Percona LLC and/or its affiliates
+(C) 2013, 2016 Percona LLC and/or its affiliates
 
 This program is free software; you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
@@ -17,19 +17,21 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
 
 #include <pwd.h>
 #include <grp.h>
+#include <errno.h>
+#include <my_sys.h>
 #include <string.h>
 #include <stdlib.h>
 #include <unistd.h>
 
-enum { max_nss_name_len = 10240 };
-enum { max_number_of_groups = 1024 };
+static int gr_buf_size= 0;
 
 /** Groups iterator. It's not exposed outsude */
 struct groups_iter {
-  char buf[max_nss_name_len];
-  gid_t groups[max_number_of_groups];
+  char *buf;
+  gid_t *groups;
   int current_group;
   int ngroups;
+  int buf_size;
 };
 
 /** Create iterator through user groups.
@@ -41,23 +43,58 @@ struct groups_iter *groups_iter_new(const char *user_name)
   int error;
   struct groups_iter *it;
 
-  it= calloc(1, sizeof(struct groups_iter));
-  if (it == NULL)
-    return NULL;
+  if (gr_buf_size <= 0)
+  {
+    long gr_size_max, pw_size_max;
+    gr_size_max= sysconf(_SC_GETGR_R_SIZE_MAX);
+    pw_size_max= sysconf(_SC_GETPW_R_SIZE_MAX);
+    gr_buf_size= gr_size_max > pw_size_max ? gr_size_max : pw_size_max;
+  }
 
-  error= getpwnam_r(user_name, &pwd, it->buf, max_nss_name_len, &pwd_result);
+  it= (struct groups_iter *) my_malloc(sizeof(struct groups_iter),
+                                       MYF(MY_FAE | MY_ZEROFILL));
+
+  it->buf_size= gr_buf_size;
+  if (it->buf_size <= 0)
+    it->buf_size= 1024;
+
+  it->buf= (char *) my_malloc(it->buf_size, MYF(MY_FAE));
+
+  while ((error= getpwnam_r(user_name, &pwd, it->buf, it->buf_size,
+                            &pwd_result)) == ERANGE)
+  {
+    it->buf_size= it->buf_size * 2;
+    it->buf= (char *) my_realloc(it->buf, it->buf_size, MYF(MY_FAE));
+  }
   if (error != 0 || pwd_result == NULL)
   {
-    free(it);
+    fprintf(stderr, "auth_pam: Unable to obtain the passwd entry for the user "
+                    "'%s'.", user_name);
+    my_free(it->buf);
+    my_free(it);
     return NULL;
   }
 
-  it->ngroups= max_number_of_groups;
+  gr_buf_size= it->buf_size;
+
+  it->ngroups= 1024;
+  it->groups= (gid_t *) my_malloc(it->ngroups * sizeof(gid_t), MYF(MY_FAE));
   error= getgrouplist(user_name, pwd_result->pw_gid, it->groups, &it->ngroups);
   if (error == -1)
   {
-    free(it);
-    return NULL;
+    it->groups= (gid_t *) my_realloc(it->groups, it->ngroups * sizeof(gid_t),
+                                     MYF(MY_FAE));
+    error= getgrouplist(user_name, pwd_result->pw_gid, it->groups,
+                        &it->ngroups);
+    if (error == -1)
+    {
+      fprintf(stderr, "auth_pam: Unable to obtain the group access list for "
+                      "the user '%s'.", user_name);
+      my_free(it->buf);
+      my_free(it->groups);
+      my_free(it);
+      return NULL;
+    }
   }
 
   return it;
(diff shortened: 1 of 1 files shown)


gl-sergei:
http://jenkins.percona.com/view/PS%205.5/job/percona-server-5.5-param/1329/

plugin/percona-pam-for-mysql/src/groups.c
```diff
@@ -41,23 +43,58 @@ struct groups_iter *groups_iter_new(const char *user_name)
   int error;
   struct groups_iter *it;
 
-  it= calloc(1, sizeof(struct groups_iter));
-  if (it == NULL)
-    return NULL;
+  if (gr_buf_size <= 0)
+  {
+    long gr_size_max, pw_size_max;
+    gr_size_max= sysconf(_SC_GETGR_R_SIZE_MAX);
```
laurynas-biveinis, line 49: sysconf(_SC_GETGR_R_SIZE_MAX) is allowed to return -1 as a valid return value indicating no max size, this case needs to be handled

plugin/percona-pam-for-mysql/src/groups.c
```diff
@@ -41,23 +43,58 @@ struct groups_iter *groups_iter_new(const char *user_name)
...
+    pw_size_max= sysconf(_SC_GETPW_R_SIZE_MAX);
```
laurynas-biveinis, line 50: Likewise

plugin/percona-pam-for-mysql/src/groups.c
```diff
@@ -17,19 +17,21 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
...
 #include <grp.h>
+#include <errno.h>
+#include <my_sys.h>
 #include <string.h>
 #include <stdlib.h>
 #include <unistd.h>
 
-enum { max_nss_name_len = 10240 };
-enum { max_number_of_groups = 1024 };
+static int32_t gr_buf_size= 0;
```
laurynas-biveinis: size_t

plugin/percona-pam-for-mysql/src/groups.c
```diff
@@ -17,19 +17,21 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
...
 
 /** Groups iterator. It's not exposed outsude */
 struct groups_iter {
-  char buf[max_nss_name_len];
-  gid_t groups[max_number_of_groups];
+  char *buf;
+  gid_t *groups;
   int current_group;
   int ngroups;
+  int buf_size;
```
laurynas-biveinis, line 34: size_t

plugin/percona-pam-for-mysql/src/groups.c
gl-sergei, line 49: it is handled

```
if (it->buf_size <= 0)
  it->buf_size= 1024;
```

plugin/percona-pam-for-mysql/src/groups.c
```diff
@@ -41,23 +43,58 @@ struct groups_iter *groups_iter_new(const char *user_name)
...
+    gr_size_max= sysconf(_SC_GETGR_R_SIZE_MAX);
+    pw_size_max= sysconf(_SC_GETPW_R_SIZE_MAX);
+    gr_buf_size= gr_size_max > pw_size_max ? gr_size_max : pw_size_max;
+  }
+
+  it= (struct groups_iter *) my_malloc(sizeof(struct groups_iter),
+                                       MYF(MY_FAE | MY_ZEROFILL));
+
+  it->buf_size= gr_buf_size;
+  if (it->buf_size <= 0)
```
laurynas-biveinis, line 58: Hmm it might be handled here

plugin/percona-pam-for-mysql/src/groups.c
```diff
@@ -41,23 +43,58 @@ struct groups_iter *groups_iter_new(const char *user_name)
...
-    return NULL;
+    it->groups= (gid_t *) my_realloc(it->groups, it->ngroups * sizeof(gid_t),
+                                     MYF(MY_FAE));
+    error= getgrouplist(user_name, pwd_result->pw_gid, it->groups,
+                        &it->ngroups);
+    if (error == -1)
+    {
+      fprintf(stderr, "auth_pam: Unable to obtain the group access list for "
+                      "the user '%s'.", user_name);
+      my_free(it->buf);
```
laurynas-biveinis, line 93: I'd add DBUG_ASSERT(0) after the fprintf, as this should be an impossible situation?

plugin/percona-pam-for-mysql/src/groups.c
gl-sergei: "size_t" is unsigned, sysconf returns "long" which is signed. And I wanted to avoid mutexes and still feel safe to load/store gr_buf_size from multiple threads. Since on x86 32-bit load/store instructions are atomic, I felt like int32 would be OK.

laurynas-biveinis: OK, then at lest gr_buf_size and groups_iter::buf_size types should agree? And it can be plain int, it will be atomic the same

plugin/percona-pam-for-mysql/src/groups.c
gl-sergei, line 93: Probably, yes. It has to be impossible. But&#x2026; when I tested
`getgrouplist` on the laptop with OS X, I discovered that it behaves
very strange and not at all as described in the manual.

I have a user which is a member of 15 groups and when I pass
`ngroups=2`, `getgrouplist` returns -1 as expected, but it doesn't
return actual `ngroups`.

When I pass `ngroups` = 4, it returns 4 groups out of 15 and
indicates success!

When I pass non-existent user name and id, it returns -1.

There is always room for the race condition. What if we authenticated
user with PAM, then started to look for his groups and someone else
dropped the user at the same time. Should we crash the server?

laurynas-biveinis, line 93: Good point re. race condition - should it loop until it's not -1? With logic that each iteration should bump ngroups, in case getgrouplist didn't do that itself as on Mac

gl-sergei, line 93: Loop doesn't work on mac unfortunately. As I said earlier, `getgrouplist` doesn't return -1 when I pass 4, but the user is the member of 15 groups... Here is an example:

```
$ ./aaa sergei 2
getgrouplist() returned -1; ngroups = 2

$ ./aaa sergei 3
ngroups = 3
20 (staff)
401 (com.apple.sharepoint.group.1)
12 (everyone)

$ ./aaa sergei 4
ngroups = 4
20 (staff)
401 (com.apple.sharepoint.group.1)
12 (everyone)
61 (localaccounts)

$ ./aaa sergei 5
ngroups = 5
20 (staff)
401 (com.apple.sharepoint.group.1)
12 (everyone)
61 (localaccounts)
79 (_appserverusr)

$ ./aaa sergei 10
ngroups = 10
20 (staff)
401 (com.apple.sharepoint.group.1)
12 (everyone)
61 (localaccounts)
79 (_appserverusr)
80 (admin)
81 (_appserveradm)
98 (_lpadmin)
33 (_appstore)
100 (_lpoperator)

$ ./aaa sergei 20
ngroups = 14
20 (staff)
401 (com.apple.sharepoint.group.1)
12 (everyone)
61 (localaccounts)
79 (_appserverusr)
80 (admin)
81 (_appserveradm)
98 (_lpadmin)
33 (_appstore)
100 (_lpoperator)
204 (_developer)
395 (com.apple.access_ftp)
398 (com.apple.access_screensharing)
399 (com.apple.access_ssh)
```

Moreover, `sysconf(_SC_NGROUPS_MAX)` returns 16 which in fact can be exceeded... Looks like we have no options on OS X if we are using `getgrouplist`.

gl-sergei, line 93: I took the source code of the example here: http://man7.org/linux/man-pages/man3/getgrouplist.3.html

laurynas-biveinis, line 93: Is mac still broken if getgrouplist is passed at least sysconf(_SC_NGROUPS_MAX) on the first iteration?

gl-sergei, line 93: Yes, it is. I have just increased group membership to 19 and verified once again that `sysconf(_SC_NGROUPS_MAX)` is 16. When I pass 16 `getgrouplist` returns 0, fills out 16 groups and doesn't change `ngroups`. When I pass 20, it fills out all 19.

laurynas-biveinis, line 93: OK - then there is little we else can do. (You can try filing a Mac bug if you feel adventurous :))

gl-sergei:
I have replaced int32_t with int and updated all branches

laurynas-biveinis:
LGTM

PR #819, opened 2016-08-12
Fix bug 1612551 (Test innodb.percona_changed_page_bmp_crash is unstable)

laurynas-biveinis:
Ensure that the second bitmap file is never zero-sized by issuing some
write workload while it's the current bitmap output file. Then a clean
shutdown happens, ensuring that the bitmap is written. Cleanup the
testcase.

http://jenkins.percona.com/job/percona-server-5.5-param/1330/


percona-ysorokin:
LGTM

PR #822, opened 2016-08-15
5.5 - Fix for ld_preload in mysqld_safe.sh

tplavcic:
More info here:
https://jira.percona.com/browse/BLD-476

### DIFF
--- a/scripts/mysqld_safe.sh
+++ b/scripts/mysqld_safe.sh
@@ -258,6 +258,16 @@ add_mysqld_ld_preload() {
   lib_to_add="$1"
   log_notice "Adding '$lib_to_add' to LD_PRELOAD for mysqld"
 
+  # Check if the library is in the reduced number of standard system directories
+  case "$lib_to_add" in
+    /usr/lib64/* | /usr/lib/* | ${MY_BASEDIR_VERSION}/lib/*)
+      ;;
+    *)
+      log_error "ld_preload libraries can only be loaded from system directories (/usr/lib64, /usr/lib, ${MY_BASEDIR_VERSION}/lib)"
+      exit 1
+      ;;
+  esac
+
   case "$lib_to_add" in
     *' '*)
       # Must strip path from lib, and add it to LD_LIBRARY_PATH


scripts/mysqld_safe.sh
```diff
@@ -258,6 +258,16 @@ add_mysqld_ld_preload() {
   lib_to_add="$1"
   log_notice "Adding '$lib_to_add' to LD_PRELOAD for mysqld"
 
+  # Check if the library is in the reduced number of standard system directories
+  case "$lib_to_add" in
+    /usr/lib64* | /usr/lib/x86_64-linux-gnu* | /usr/lib* | ${MY_BASEDIR_VERSION}/lib*)
```
EvgeniyPatlan: may be we can simplified this `in`
/usr/lib\* | ${MY_BASEDIR_VERSION}/lib\* should be enough
as \* would include all other paths

tplavcic: You're correct, but it even seems more correct for me to add "/" on the paths to really make it possible to only use /usr/lib/.. and not something like /usr/library/... although /usr is writeable by root only so it should be fine but anyway.

EvgeniyPatlan:
ok for me

Oneiroi:
Can we confirm $MY_BASEDIR_VERSION can not be modified by configuration file / runtime args ?

Also is there ever any library that we're aware of that needs to be loaded from $MY_BASEDIR_VERSION/lib ?

Also affects #823 #824

tplavcic:
@Oneiroi 
$MY_BASEDIR_VERSION can be modified as descibed here:
https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_basedir

It was used in the ld_preloading for binary tarball installations so that they can find jemalloc and TokuBackup libraries that we ship in our binary tarballs - so they are not installed in /usr/lib*.
Please comment if I should remove this path as well but then the users will need to specify those preloads manually when using binary tarball distributed packages.

Oneiroi:
@tplavcic no requirement to remove at this time; this already reduces the attack surface a fair amount and the preservation of basedir is required for a valid reason as you have mentioned (loading to Toku shared libraries) this does mean that if the user can set the BASEDIR and has write access to said BASIDIR then the attack could proceed however this is a lot of assumptions that the user has the required access to perform these modifications in the first place.

PR #825, opened 2016-08-15
Add filtering by `sql_command`

gl-sergei:
Blueprint:

https://blueprints.launchpad.net/percona-server/+spec/audit-filer-command
- add option `audit_log_include_command` to specify commands to log
- add option `audit_log_exclude_command` to specify commands to exclude
  from logging
- backport changes for `next_word` fixing the capture of default
  database with quoted backticks.
- backport case-sensitivity fix for account filters


gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1299/
there are two failures, I don't think they are related to this change, I'll take a close look tomorrow morning

laurynas-biveinis:
It looks like command filtering is case-sensitive but it shouldn't be? I.e. if filtering options are set using not fully lowercase command names

gl-sergei:
I don't think case sensitivity is a big deal here... `sql_command` is always lowercase so far. And its unlikely that someday there will be two different commands "insert" and "Insert"... Ok. Lets make it case insensitive.

gl-sergei:
Branch re-pushed in case insensitive variant. `__attribute__` replaced with `MY_ATTRIBUTE`. Added test case with filter only partially covering the executed queries.

gl-sergei:
regarding `audit_log_filter_users` failure. The following sequence:

```
SET GLOBAL audit_log_flush=ON;
connect (test,,...);
connection test;
SQL STATEMENT;
disconnect test;
--source include/wait_until_disconnected.inc
```

has been reordered in the audit log as follows:

```
SQL STATEMENT;
SET GLOBAL audit_log_flush=ON;
```

I think, this can be fixed by executing some command, "ping" or alike before `connect` to make sure `set global` has been logged. This command must be grepped away in the echo script.

I'll report a bug.

laurynas-biveinis:
For case sensitivity, I don't mean we should handle non-lower case commands in audit notifications, it's just that the user should be able to specify filters in any case

gl-sergei:
regarding `main.audit_log_old`. It looks odd to me. Parser reporting

```
not well-formed (invalid token) at line 639, column 12, byte 20241 at /usr/lib/x86_64-linux-gnu/perl5/5.20/XML/Parser.pm line 187.
```

but in my testing environment the log file is only 18919 bytes and 577 lines.

I will add the code to print the entire file in case of parser error. Will track this work separately.

gl-sergei:
case sensitivity: yeah, I have made that change. User is able to specify command in any case and plugin makes an assumption that MySQL reports commands in lower case.

laurynas-biveinis:
LGTM

gl-sergei:
Here is updated jenkins run: http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1300/

PR #827, opened 2016-08-16
Fix bug 1613641 (Test rpl.rpl_filter_tables_not_exist produces an uns…

laurynas-biveinis:
…uppressed warning in the error log)

The unsuppressed warning should be suppressed, as it already is in
5.6+.

http://jenkins.percona.com/job/percona-server-5.5-param/1331/


percona-ysorokin:
LGTM

PR #830, opened 2016-08-17
Fix bug 1613663 (Test main.mysqltest is unstable)

laurynas-biveinis:
Prevent negative time in SHOW PROCESSLIST by replacing it with 0. A
proper fix would be to use only monotonic clocks for time duration
measurements, but that's too invasive for us to fix.

http://jenkins.percona.com/job/percona-server-5.5-param/1333/


percona-ysorokin:
LGTM

PR #833, opened 2016-08-17
Fix bug 1613728 (Test rpl.rpl_bug58546 may crash server due to concur…

laurynas-biveinis:
…rent DBUG access)

The --source include/rpl_stop.inc is preceded by set global DEBUG='';
executed with the slaves running. Thus, this has the same underlying
issue as upstream bug 58754 / 11765758 and is fixed the same way, by
executing SET GLOBAL DEBUG=... with a stopped slave.

http://jenkins.percona.com/job/percona-server-5.5-param/1334/


percona-ysorokin:
LGTM

PR #839, opened 2016-08-17
Fix bug 1613986 (Test percona_changed_page_bmp_crash produces warning…

laurynas-biveinis:
…s if doublewrite was used for recovery)

The intentionally crashed server might need doublewrite buffer to
restore a data page. It produces a warning in the error log in this
case, which should be suppressed.

http://jenkins.percona.com/job/percona-server-5.5-param/1337/


percona-ysorokin:
LGTM

PR #842, opened 2016-08-18
Work towards fixing bug 904714 (Intermittent rpl.percona_bug860910 fa…

laurynas-biveinis:
…ilures)

The testcase dumps RPL debug info on failure, which is then truncated
by MTR, since it only dumps the last 20 lines of queries before the
failure. Bump that limit to 500.

http://jenkins.percona.com/job/percona-server-5.5-param/1338/

### DIFF
--- a/mysql-test/mysql-test-run.pl
+++ b/mysql-test/mysql-test-run.pl
@@ -4268,7 +4268,7 @@ sub run_testcase ($) {
       {
         $tinfo->{comment}.=
 	   "== $log_file_name == \n".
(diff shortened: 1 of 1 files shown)


laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.5-param/1339/

percona-ysorokin:
LGTM
Do we care about `mysql-test/lib/v1/mysql-test-run.pl`?

laurynas-biveinis:
We don't. IIRC it's for Random Query Generator only?

percona-ysorokin:
OK. Just wanted to confirm.

PR #846, opened 2016-08-18
Bug 1614444: Assertion `local->stack.frames[local->stack.top].query =…

gl-sergei:
…= event_general->general_query.str' failed

It is a bug which can affects filtering by DB for triggers.

There is a trigger

```
CREATE TRIGGER tr1 BEFORE INSERT ON t1 FOR EACH ROW SET @aux=1;
```

and query

```
INSERT INTO t1(c5,c6)VALUES (1,0);
```

Lets see which events are sent to the audit_log:
1. TABLE ACCESS event for "INSERT INTO t1(c5,c6)VALUES (1,0)"
2. STATUS event for "SET @aux=1"
3. STATUS event for "INSERT INTO t1(c5,c6)VALUES (1,0)"

When (1) comes, plugin pushes "INSERT INTO t1(c5,c6)VALUES (1,0)" to
the top of the stack, so it can match it later with STATUS event.

When (2) comes, plugin is trying to match "SET @aux=1" against the stack
top. But "SET @aux=1" isn't on the stack since it doesn't access any
database. In debug mode we see the assertion failure. In the release
mode, "SET @aux=1" will be attributed as accessing table "t1", which is
wrong.

The fix is to replace assert with if statement. Now the query the within
trigger which doesn't access any database will always be logged just
like any similar query executed outside of the trigger.

### DIFF
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -1013,18 +1013,28 @@ my_bool audit_log_update_thd_local(MYSQL_THD thd,
 
     if (event_general->event_subclass == MYSQL_AUDIT_GENERAL_STATUS)
     {
-      DBUG_ASSERT(local->stack.frames[local->stack.top].query == NULL
-                  || local->stack.frames[local->stack.top].query
-                      == event_general->general_query.str);
+      local->skip_query= FALSE;
 
-      local->skip_query= audit_log_include_databases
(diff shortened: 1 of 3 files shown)


laurynas-biveinis:
LGTM

gl-sergei:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/321/
I really should have started it with --suite=audit_log for speedup.

PR #848, opened 2016-08-18
5.7 - Add WITH_ASAN=ON option in build-binary.sh for debug build

tplavcic:
As per: https://jira.percona.com/browse/BLD-477
we want -DWITH_ASAN=ON option for Debug builds done with build-binary.sh but not if the build is also done with valgrind instrumentation.
Since ASAN is supported with GCC >= 4.8 we need to check if we can safely enable it.

**Test:**
**Centos7:**
Debug adds this option.

```
+ GT_VERSION=4.8.5
+ '[' 4.8.5 = 4.8.5 ']'
+ DEBUG_EXTRA='-DDEBUG_EXTNAME=OFF -DWITH_ASAN=ON'
```

Debug+Valgrind doesn't:

```
+ [[ Debug == \D\e\b\u\g ]]
+ [[  -DWITH_VALGRIND=ON != *WITH_VALGRIND=ON* ]]
```

and adding option is skipped.

**Ubuntu precise:**
Debug only:
The ASAN option is not added because gcc < 4.8.

```
+ GT_VERSION=4.8.0
+ '[' 4.8.0 = 4.6 ']'
```

Debug with valgrind - not added because it's a valgrind build also.

```
+ [[ Debug == \D\e\b\u\g ]]
+ [[  -DWITH_VALGRIND=ON != *WITH_VALGRIND=ON* ]]
```

### DIFF
--- a/build-ps/build-binary.sh
+++ b/build-ps/build-binary.sh
@@ -19,6 +19,7 @@ QUIET='VERBOSE=1'
 CMAKE_BUILD_TYPE='RelWithDebInfo'
 DEBUG_COMMENT=''
 WITH_JEMALLOC=''
+DEBUG_EXTRA=''
 WITH_SSL='/usr'
 WITH_SSL_TYPE='system'
 OPENSSL_INCLUDE=''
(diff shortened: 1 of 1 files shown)


EvgeniyPatlan:
LGTM

PR #852, opened 2016-08-19
Fix bug 1614885 (Spurious Valgrind error on main.udf)

laurynas-biveinis:
Add a Valgrind suppression for internal libc allocations stemming from
dlerror_run.

http://jenkins.percona.com/job/percona-server-5.5-valgrind-param/10/


percona-ysorokin:
LGTM

PR #861, opened 2016-08-20
Fix bug 1613728 (Test rpl.rpl_bug58546 may crash server due to concur…

laurynas-biveinis:
…rent DBUG access), 2nd attempt

The previous attempt to ensure that DEBUG variable is not set while
binlog dump threads are active failed to ensure that by DROP TABLE
being issued before SET DEBUG. Since dump threads do not necessarily
quit when slave disconnects, they are running and see the extra binlog
written. Fix by setting DEBUG when master binlog dump threads are in
the state of having sent all binlogs to the slave.

http://jenkins.percona.com/job/percona-server-5.5-param/1344/


percona-ysorokin:
LGTM

PR #864, opened 2016-08-22
Fix bug 1615468 (mysql-test/mysql-test-run.pl is not executable)

laurynas-biveinis:
Make mysql-test/mysql-test-run.pl executable again.


percona-ysorokin:
LGTM

PR #870, opened 2016-08-23
Fix bug 1614849 (INFORMATION_SCHEMA.TABLES (or other schema info tabl…

laurynas-biveinis:
…e) and GLOBAL_TEMPORARY_TABLES queries running in parallel may crash or hang)

LOCK_temporary_tables mutex was declared next to the temporary_tables
field, which happened to be in Open_tables_state class, one of the THD
ancestors. This class is used for taking backups and restoring the
open table state, by default member-wise assignment operator. Thus,
mutex itself was being copied and restored, while potentially also
being locked and unlocked in other threads, resulting in broken mutex
state.

Fix by moving LOCK_temporary_tables to class THD itself. Add a debug
sync testcase to percona_show_temp_tables_debug, and add a
INFORMATION_SCHEMA.TABLES query workload to
percona_show_temp_tables_stress.

At the same time cleanup percona_show_temp_tables_stress:
- grant all global privileges to the event runner accounts so that they
  actually get some data;
- use KILL CONNECTION instead of KILL QUERIES as it's more correct and
  avoids a bug where KILL QUERIES might be ignored by compound statements
  (http://bugs.mysql.com/bug.php?id=82679)
- use include/no_running_events.inc to wait for the event runner threads
  to quit.

http://jenkins.percona.com/job/percona-server-5.5-param/1348/

### DIFF
--- a/sql/sql_show.cc
+++ b/sql/sql_show.cc
@@ -3962,6 +3962,17 @@ static int fill_global_temporary_tables(THD *thd, TABLE_LIST *tables, COND *cond
  
   while ((thd_item=it++)) {
     mysql_mutex_lock(&thd_item->LOCK_temporary_tables);
+
+#ifndef DBUG_OFF
+    const char* tmp_proc_info= thd_item->proc_info;
+    if (tmp_proc_info &&
+        !strncmp(tmp_proc_info,
+                 STRING_WITH_LEN("debug sync point: before_open_in_get_all_tables"))) {
+      DEBUG_SYNC(thd,
+                 "fill_global_temporary_tables_thd_item_at_tables_debug_sync");
+    }
+#endif
+
     for (tmp=thd_item->temporary_tables; tmp; tmp=tmp->next) {
 
 #ifndef NO_EMBEDDED_ACCESS_CHECKS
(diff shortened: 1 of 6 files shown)


sql/sql_show.cc
```diff
@@ -3962,6 +3962,18 @@ static int fill_global_temporary_tables(THD *thd, TABLE_LIST *tables, COND *cond
  
   while ((thd_item=it++)) {
     mysql_mutex_lock(&thd_item->LOCK_temporary_tables);
+
+#ifndef DBUG_OFF
+    const char* tmp_proc_info= thd_item->proc_info;
+    if (tmp_proc_info &&
+        !strncmp(tmp_proc_info,
+                 "debug sync point: before_open_in_get_all_tables",
```
percona-ysorokin:
``` c++
STRING_WITH_LEN("debug sync point: before_open_in_get_all_tables")
```

percona-ysorokin:
LGTM with minor comment

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.5-param/1351/

percona-ysorokin:
LGTM

PR #873, opened 2016-08-23
Fix bug 1615959 (Unsuppressed Valgrind warning on plugin_auth_qa_2)

laurynas-biveinis:
Make libc _dlerror_run Valgrind suppression general enough to catch
the new unsupressed case.

http://jenkins.percona.com/job/percona-server-5.5-valgrind-param/11/


percona-ysorokin:
LGTM

PR #876, opened 2016-08-23
Work towards diagnosing bug 1615945 (Test percona_slow_query_log_time…

laurynas-biveinis:
…stamp_always is unstable)

Adjust include/log_grep.inc to optionally accept the expected match
count and dump the grepped file if it does not match. Make
percona_slow_query_log_timestamp_always slow query log filename
unique.

http://jenkins.percona.com/job/percona-server-5.5-param/1349/


percona-ysorokin:
LGTM

PR #880, opened 2016-08-23
Fix bug 1553938 (Valgrind error on hp_test2)

laurynas-biveinis:
Trivailly fix Valgrind errors on the hp_test2 testcase.

http://jenkins.percona.com/job/percona-server-5.5-param/1350/


percona-ysorokin:
LGTM

PR #883, opened 2016-08-23
Fix bug 1616091 (Test rpl.rpl_semi_sync is unstable)

laurynas-biveinis:
mysql-test/include/stop_dump_threads.inc tries to disable result and
query log, but wait_condition.inc re-enables the query log, making the
testcase produce output if the loop does more than one iteration. Fix
by making wait_condition.inc not change the query log status.

http://jenkins.percona.com/job/percona-server-5.5-param/1353/


percona-ysorokin:
LGTM

PR #886, opened 2016-08-24
Fix bug 1616333 (Test rpl.rpl_backup_locks_mts is unstable)

laurynas-biveinis:
On opening the relay log, slave may inject an
Format_description_log_event event. It might be processed by the slave
SQL thread after LOCK BINLOG FOR BACKUP is issued, and this event
requires async processing for multi-threaded slave, resulting in the
slave SQL thread being blocked instead of slave worker threads, as
expected by the testcase.

Fix by issuing some workload on the master and syncing the slave
before LOCK BINLOG FOR BACKUP, so that the format description event is
processed before binlog is locked.

http://jenkins.percona.com/job/percona-server-5.6-param/1322/


percona-ysorokin:
LGTM

PR #888, opened 2016-08-24
Fix bug 1616404 (Test perfschema.setup_instruments_defaults is unstable)

laurynas-biveinis:
Replace the custom shutdown sequence (that uses send_shutdown,
creating a race condition) with an include file use.

http://jenkins.percona.com/job/percona-server-5.6-param/1324/


percona-ysorokin:
LGTM

PR #890, opened 2016-08-24
Fix bug 1615970 ("InnoDB: Failed to set NUMA memory policy" on 32-bit…

laurynas-biveinis:
… hosts)

Suppress the warnings stemming from the NUMA calls returning ENOSYS.

http://jenkins.percona.com/job/percona-server-5.6-param/1329/


percona-ysorokin:
LGTM

PR #892, opened 2016-08-24
Work towards diagnosing bug 1616392 (Test parts.partition_debug_innod…

laurynas-biveinis:
…b is unstable)

The testcase fails to restart the server after a crash, with a very
long timeout. Adjust MTR to dump the error log for all the servers if
mysqltest died.

http://jenkins.percona.com/job/percona-server-5.6-param/1323/


percona-ysorokin:
LGTM

PR #894, opened 2016-08-24
Try to fix bug 1616320 (Test events_time_zone is unstable)

laurynas-biveinis:
Bump the test sensitivity parameter and hope it will be enough.

http://jenkins.percona.com/job/percona-server-5.6-param/1321/


percona-ysorokin:
LGTM

PR #896, opened 2016-08-25
Bug#20458574 FIX ALL THE ARRAY BOUNDS WARNINGS

laurynas-biveinis:
Remove compiler warnings when building in optimized mode:

In strings/decimal.c fix numerous warnings of the type:
array subscript is below array bounds [-Werror=array-bounds]
This part of the patch is based on a patch
from https://github.com/webscalesql/webscalesql-5.6
That patch added some bounds-checks for loops, and worked for
gcc4.8.3 but not for gcc 4.9.1 so it was re-written to unroll the loops instead.

In other source files, fix numerous warnings of the type:
variable XX set but not used [-Werror=unused-but-set-variable]
variable XX may be used uninitialized in this function [-Werror=maybe-uninitialized]
These were variables used only for debugging purposes,
or variables not proven to be set by all execution paths.

Compiled with gcc 4.7.2 4.8.3 4.9.1 in optimized mode, with -Wall -Werror

### DIFF
--- a/strings/decimal.c
+++ b/strings/decimal.c
@@ -1,4 +1,4 @@
-/* Copyright (c) 2004, 2014, Oracle and/or its affiliates. All rights reserved.
+/* Copyright (c) 2004, 2015, Oracle and/or its affiliates. All rights reserved.
 
    This program is free software; you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


strings/decimal.c
```diff
@@ -211,6 +211,74 @@ static const dec1 frac_max[DIG_PER_DEC1-1]={
...
+
+  @param   i    start index
+  @param   val  value to compare against list of powers of 10
+
+  @retval  Number of leading zeroes that can be removed from fraction.
+
+  @note Why unroll? To get rid of lots of compiler warnings [-Warray-bounds]
+        Nice bonus: unrolled code is significantly faster.
+ */
+static inline int count_leading_zeroes(int i, dec1 val)
```
percona-ysorokin, line 227: If called with `0` for the second parameter `count_leading_zeroes(XXX, 0)` will always assert

strings/decimal.c
```diff
@@ -211,6 +211,74 @@ static const dec1 frac_max[DIG_PER_DEC1-1]={
...
+
+  @param   i    start index
+  @param   val  value to compare against list of powers of 10
+
+  @retval  Number of trailing zeroes that can be removed from fraction.
+
+  @note Why unroll? To get rid of lots of compiler warnings [-Warray-bounds]
+        Nice bonus: unrolled code is significantly faster.
+ */
+static inline int count_trailing_zeroes(int i, dec1 val)
```
percona-ysorokin, line 260: The same: `count_trailing_zeroes(XXX, 0)` will always assert.

strings/decimal.c
laurynas-biveinis: All callers seem to handle it

strings/decimal.c
laurynas-biveinis: Likewise all callers seem to handle this

strings/decimal.c
```diff
@@ -493,7 +561,7 @@ static void digits_bounds(decimal_t *from, int *start_result, int *end_result)
     stop= (int) ((buf_end - from->buf + 1) * DIG_PER_DEC1);
     i= 1;
   }
-  for (; *buf_end % powers10[i++] == 0; stop--) ;
+  stop-= count_trailing_zeroes(i, *buf_end);
```
percona-ysorokin, line 564: This call seems to be suspicious - `*buf_end` can potentially be `0` here.
if the previous `while`

``` cpp
while (buf_end > buf_beg  && *buf_end == 0)
     buf_end--;
```

terminated with `buf_end == buf_beg`

All the other calls to `count_trailing_zeroes` / `count_leading_zeroes` seems to guarantee that the second arg is non-zero.

laurynas-biveinis: If the whole decimal number is zeros, then buf_end and buf_beg already meet earlier:

  /\* find non-zero digit from number begining */
  while (buf_beg < end && *buf_beg == 0)
    buf_beg++;

  if (buf_beg >= end)
  {
    /\* it is zero */
    *start_result= *end_result= 0;
    return;
  }

percona-ysorokin:
LGTM

PR #902, opened 2016-08-25
Fix bug 1616785 (Shutdown waiting for purge to complete is undiagnose…

laurynas-biveinis:
…d for the 1st minute)

Tweak shutdown loop that waits for the buffer pool to be fully flushed
so that the diagnostic message is printed at the start, not end, of
a 60-second interval.

http://jenkins.percona.com/job/percona-server-5.6-param/1331/


percona-ysorokin:
LGTM

PR #907, opened 2016-08-25
my.cnf should be provided on all distros.

EvgeniyPatlan:
As It was decided my.cnf should be provided for all distros.
The easiest way is to make symlink and include dir wit *.cnf file.
http://jenkins.percona.com/job/percona-server-5.7-RELEASE/51/
The same behavior is used for debian/ubuntu installations by oracle for original mysql.
As it is easier to maintain the same behavior for all systems centos would work in the same way.
So we shouldn't face any issues with mysql-libs.


build-ps/percona-server.spec
```diff
@@ -497,19 +501,30 @@ datadir=$(/usr/bin/my_print_defaults server mysqld | grep '^--datadir=' | sed -n
...
 # and the default system one doesn't have info needed
-%if 0%{?rhel} < 7
-if [ $1 -eq 1 ]; then
+#%if 0%{?rhel} < 7
+#if [ $1 -eq 1 ]; then
   cnflog=$(/usr/bin/my_print_defaults mysqld|grep -c log-error)
   if [ $cnflog = 0 -a -f /etc/my.cnf ]; then
     sed -i "/^\[mysqld\]$/a log-error=/var/log/mysqld.log" /etc/my.cnf
   fi
+  sed -i "/^\[mysqld\]$/a log-error=/var/log/mysqld.log" /etc/percona-server.conf.d/my.cnf
```
tplavcic: do we want this executed every time?
is the /etc/percona-server.conf.d/my.cnf correct or it should be /etc/mysql/percona-server.conf.d/my.cnf?
do we need this if we provide our own cnf file with log-error included?

tplavcic:
Did you do some tests for this?
The biggest problem here is what will happen on upgrades and if users already have their own custom cnf files...

build-ps/percona-server.spec
EvgeniyPatlan: In case if my.cnf exists we do not replace it so we should add these lines

EvgeniyPatlan:
## Sure. Everything works correctly.

[vagrant@localhost etc]$ ls -la my.cnf
lrwxrwxrwx. 1 root root 24 Aug 25 14:05 my.cnf -> /etc/alternatives/my.cnf
[vagrant@localhost etc]$ ls -la /etc/alternatives/my.cnf
lrwxrwxrwx. 1 root root 23 Aug 25 14:05 /etc/alternatives/my.cnf -> /etc/percona-server.cnf
[vagrant@localhost etc]$ sudo service mysql status

## mysqld (pid  5578) is running...

vagrant@precise64:/etc/mysql$ ls -la my.cnf
lrwxrwxrwx 1 root root 24 Aug 25 09:10 my.cnf -> /etc/alternatives/my.cnf
vagrant@precise64:/etc/mysql$ ls -la /etc/alternatives/my.cnf
lrwxrwxrwx 1 root root 29 Aug 25 09:10 /etc/alternatives/my.cnf -> /etc/mysql/percona-server.cnf
vagrant@precise64:/etc/mysql$ sudo service mysql status
- Percona Server 5.7.14-7 is running

build-ps/debian/extra/percona-server.conf.d/mysqld.cnf
```diff
@@ -0,0 +1,32 @@
...
+explicit_defaults_for_timestamp
+
+# Instead of skip-networking the default is now to listen only on
+# localhost which is more compatible and is not less secure.
+bind-address = 127.0.0.1
+
+log-error    = /var/log/mysql/error.log
+
+# Recommended in standard MySQL setup
+sql_mode=NO_ENGINE_SUBSTITUTION,STRICT_TRANS_TABLES
```
laurynas-biveinis: I am not sure why I was co-assigned for this PR, but since I am... s/STRICT_TRANS_TABLES/STRICT_ALL_TABLES ?

tplavcic:
Can you please try using "ps_tokudb_admin --enable --enable-backup" on centos. I have a feeling it will not work with this.

build-ps/debian/extra/percona-server.conf.d/mysqld.cnf
EvgeniyPatlan: exactly because of such useful comments ;) thanks.

EvgeniyPatlan:
PR has been updated. It works now

build-ps/percona-server.spec
EvgeniyPatlan: I also have changed this behavior. Now the lines should be added in case if we do not replace existent my.cnf

build-ps/percona-server.spec
```diff
@@ -495,10 +500,15 @@ datadir=$(/usr/bin/my_print_defaults server mysqld | grep '^--datadir=' | sed -n
...
-# We need this because we don't provide my.cnf on centos 6
-# and the default system one doesn't have info needed
-%if 0%{?rhel} < 7
-if [ $1 -eq 1 ]; then
+if [ ! -f /etc/my.cnf ]
+then
+  update-alternatives --install /etc/my.cnf my.cnf "/etc/percona-server.cnf" 200
+else
+  echo " -------------"
+  echo "   *  The suggested my.cnf options and settings are in percona-server.conf.d/my.cnf"
```
tplavcic: should this be: "\*  The suggested my.cnf options and settings are in /etc/percona-server.conf.d/mysqld.cnf"?

build-ps/ubuntu/percona-server-common-5.7.postinst
```diff
@@ -7,13 +7,23 @@ case "$1" in
...
+    if [ ! -f /etc/mysql/my.cnf ]
     then
-      /usr/share/mysql-common/configure-symlinks install percona-server "/etc/mysql/percona-server.cnf"
+      update-alternatives --force --install /etc/mysql/my.cnf my.cnf "/etc/mysql/percona-server.cnf" 200
     else
-      echo "Notice: configure-symlinks trigger could not be called."
-      echo "Please manually make /etc/mysql/my.cnf a symlink that points to percona-server.cnf."
-    fi  
+      echo " -------------"
+      echo "   *  The suggested my.cnf options and settings are in percona-server.conf.d/mysqld.cnf"
```
tplavcic: should this be: "\*  The suggested my.cnf options and settings are in /etc/mysql/percona-server.conf.d/mysqld.cnf"?

EvgeniyPatlan: updated

build-ps/percona-server.spec
```diff
@@ -495,10 +500,15 @@ datadir=$(/usr/bin/my_print_defaults server mysqld | grep '^--datadir=' | sed -n
...
+  echo "   *  If you want to use my.cnf as default configuration file please use the following command:"
```
tplavcic: Should this be: "\*  If you want to use mysqld.cnf as default configuration file please use the following command:"?
It got me confused like the message was saying I wasn't using my.cnf but I had custom my.cnf already.

build-ps/percona-server.spec
```diff
@@ -495,10 +500,15 @@ datadir=$(/usr/bin/my_print_defaults server mysqld | grep '^--datadir=' | sed -n
...
+  echo " update-alternatives --install /etc/my.cnf my.cnf \"/etc/percona-server.cnf\" 200"
```
tplavcic, line 512: on ubuntu we use with "--force", why? what's the difference (and it was added in this PR)?

build-ps/debian/percona-server-common-5.7.postinst
```diff
@@ -0,0 +1,38 @@
...
+#    else
+#      echo "Notice: configure-symlinks trigger could not be called."
+#      echo "Please manually make /etc/mysql/my.cnf a symlink that points to percona-server.cnf."
+#    fi
+    if [ ! -f /etc/mysql/my.cnf ]
+    then
+      update-alternatives --force --install /etc/mysql/my.cnf my.cnf "/etc/mysql/percona-server.cnf" 200
+    else
+      echo " -------------"
+      echo "   *  The suggested my.cnf options and settings are in percona-server.conf.d/mysqld.cnf"
```
tplavcic: should this be: "\* The suggested my.cnf options and settings are in /etc/mysql/percona-server.conf.d/mysqld.cnf"?
I would suggest to change the start to something like "\* The suggested mysql server options and ..." because it lowers the confusion.

build-ps/percona-server.spec
EvgeniyPatlan: done

build-ps/ubuntu/percona-server-common-5.7.postinst
EvgeniyPatlan: done

build-ps/percona-server.spec
```diff
@@ -495,10 +500,15 @@ datadir=$(/usr/bin/my_print_defaults server mysqld | grep '^--datadir=' | sed -n
...
-# We need this because we don't provide my.cnf on centos 6
-# and the default system one doesn't have info needed
-%if 0%{?rhel} < 7
-if [ $1 -eq 1 ]; then
+if [ ! -f /etc/my.cnf ]
+then
+  update-alternatives --install /etc/my.cnf my.cnf "/etc/percona-server.cnf" 200
+else
+  echo " -------------"
+  echo "   *  The suggested my.cnf options and settings are in /etc/percona-server.conf.d/my.cnf"
```
tplavcic: /etc/percona-server.conf.d/mysqld.cnf ?

build-ps/percona-server.spec
```diff
@@ -495,10 +500,15 @@ datadir=$(/usr/bin/my_print_defaults server mysqld | grep '^--datadir=' | sed -n
...
+  echo "   *  If you want to use my.cnf as default configuration file please use the following command:"
```
tplavcic: mysqld.cnf ?

build-ps/percona-server.spec
EvgeniyPatlan: done

build-ps/percona-server.spec
EvgeniyPatlan: on centos there is no option --force for update-alternatives.
force was taken from original mysql.

build-ps/percona-server.spec
EvgeniyPatlan: done

build-ps/debian/percona-server-common-5.7.postinst
EvgeniyPatlan: done

build-ps/percona-server.spec
EvgeniyPatlan: done

build-ps/percona-server.spec
```diff
@@ -495,10 +500,15 @@ datadir=$(/usr/bin/my_print_defaults server mysqld | grep '^--datadir=' | sed -n
...
-# and the default system one doesn't have info needed
-%if 0%{?rhel} < 7
-if [ $1 -eq 1 ]; then
+if [ ! -f /etc/my.cnf ]
+then
+  update-alternatives --install /etc/my.cnf my.cnf "/etc/percona-server.cnf" 200
+else
+  echo " -------------"
+  echo "   *  The suggested mysql options and settings are in /etc/percona-server.conf.d/mysqld.cnf"
+  echo "   *  If you want to use mysqld.cnf as default configuration file please use the following command:"
```
tplavcic: The problem with this is that if I already have a my.cnf and issue that command nothing happens - I guess that's why the --force option on debian/ubuntu. The users might get confused by this so maybe we should add something like "If you want to use mysqld.cnf as default configuration file please remove the existing my.cnf and use the following command:"

build-ps/percona-server.spec
```diff
@@ -495,10 +500,15 @@ datadir=$(/usr/bin/my_print_defaults server mysqld | grep '^--datadir=' | sed -n
 %else
 /sbin/chkconfig --add mysql
 %endif
-# We need this because we don't provide my.cnf on centos 6
-# and the default system one doesn't have info needed
-%if 0%{?rhel} < 7
-if [ $1 -eq 1 ]; then
+if [ ! -f /etc/my.cnf ]
```
tplavcic, line 503: Do we want to run this whole block on every minor upgrade?
Previously it was in "if [ $1 -eq 1 ]; then" which means the sed command from below and other stuff was only running during the package install and not on upgrade.

build-ps/percona-server.spec
EvgeniyPatlan: Message has been modified

build-ps/percona-server.spec
EvgeniyPatlan: sed wouldn't be executed as ﻿⁠⁠⁠⁠/usr/bin/my_print_defaults mysqld|grep -c log-error﻿⁠⁠⁠⁠would return 1

tplavcic:
On debian jessie if you make a fresh installation your my.cnf will not be used from what I see, but instead the system one is used and you get that message that you can change this with:

```
update-alternatives --force --install /etc/mysql/my.cnf my.cnf "/etc/mysql/percona-server.cnf" 200
```

not sure if this is ok or not so I can't make a decision about that.

EvgeniyPatlan:
To my mind it is correct behavior as my.cnf exists.
We don't know if there were any custom changes or not. So we provide user with the prompt what should be done in case if he wants to use our my.cnf.
But such situation appears due to dependency on mysql-common package as it is also installed

EvgeniyPatlan:
I have removed the dependency to mysql-common. In the result there is no conflict if someone would like to install mysql-common later.
Configuration file '/etc/mysql/my.cnf'
 ==> File on system created by you or by a script.
 ==> File also in package provided by package maintainer.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
**\* my.cnf (Y/I/N/O/D/Z) [default=N] ?

EvgeniyPatlan:
merge approved by AlexeyB

PR #911, opened 2016-08-26
Further work to help diagnose bug 1615945 (Test percona_slow_query_lo…

laurynas-biveinis:
…g_timestamp_always is unstable)

Make log_grep.inc dump slow query log-related configuration if the
expected match count was not seen.

http://jenkins.percona.com/job/percona-server-5.5-param/1358/


percona-ysorokin:
LGTM

PR #913, opened 2016-08-26
Disable main.events_time_zone test (bug 1616320)

laurynas-biveinis:
The test is fundamentally unstable, a bump of sensitivity parameter
was not enough for our CI, it has a low probability of regressing in
Percona Server compared to MySQL, thus let's just disable it.

http://jenkins.percona.com/job/percona-server-5.6-param/1334/


percona-ysorokin:
LGTM

PR #915, opened 2016-08-26
Work towards fixing bug 1612063 (Test show_processlist is unstable)

laurynas-biveinis:
The failure of main.show_processlist shows an extra connection at the
start of the testcase that is gone by the next processlist
query. Thus, it could be a connection by a previous testcase that
hadn't cleaned up fully yet. To catch this, add processlist query to
the MTR testcase checker, so that any open connections at the end of
testcases are caught.

http://jenkins.percona.com/job/percona-server-5.5-param/1318/

### DIFF
--- a/mysql-test/include/check-testcase.test
+++ b/mysql-test/include/check-testcase.test
@@ -8,6 +8,12 @@
 
 --disable_query_log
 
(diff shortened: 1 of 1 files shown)


percona-ysorokin:
LGTM
Really good idea, BTW.

PR #918, opened 2016-08-26
Fix bug 1617216 (Test percona_processlist_row_stats is unstable)

laurynas-biveinis:
Only clear "threads_dumped" signal after both threads have processed
it. Reset DEBUG_SYNC at the end of testcase. Clean up the testcase in
general.

http://jenkins.percona.com/job/percona-server-5.6-param/1335/


percona-ysorokin:
LGTM

PR #920, opened 2016-08-26
Fix bug 1617246 (recv_recovery_from_checkpoint_start_func -Wmaybe-uni…

laurynas-biveinis:
…nitialized warning for group_scanned_lsn)

Initialise group_scanned_lsn, even though it does get initialised
later, but the control flow is apparently too complex for compiler.

http://jenkins.percona.com/job/percona-server-5.5-param/1359/


percona-ysorokin:
LGTM

PR #923, opened 2016-08-26
Bug 1613650: Sporadic audit_log_old failures

gl-sergei:
The root cause for the failure is that the `audit_handler_file_flush'
writes footer and header into the file. At the same time buffer could
flush its contents and may create partially written record.

The fix is to pause log flush before writing the footer and resume it
after the new header is written.


gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1337/ no audit_log_old failures here

laurynas-biveinis:
LGTM

PR #924, opened 2016-08-26
Fix for bug 1613650 (5.7)

gl-sergei:


gl-sergei:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/366/ -- branch contains fixes for other test cases, but the one fixing the audit_log_old is the same as in this PR

laurynas-biveinis:
LGTM

PR #929, opened 2016-08-28
Bug 1613647: Sporadic audit_log_filter_users failures

gl-sergei:
Insert silent statements (which aren't are grepped away) before
connects to make sure the previous statement we which want to see in
the audit log actually gets there before the connect.

http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1338/


laurynas-biveinis:
LGTM

PR #934, opened 2016-08-29
Try to diagnose bug 1617889 (Test percona_log_slow_query_plan is unst…

laurynas-biveinis:
…able)

Adjust the testcase to dump slow query log and relevant server
variables on a failure.

http://jenkins.percona.com/job/percona-server-5.5-param/1362/


percona-ysorokin:
LGTM

PR #937, opened 2016-08-29
Fix bug 1617949 (HandlerSocket leaks thread objects)

laurynas-biveinis:
This is caused by hstcpsvr::threads, which is
auto_ptrcontainer<std::vector<worker_thread_type *> > (effectively
std::vector<foo *>), having clear() method, which clears the container
without deleting the member pointers, which are owned by the
container.

Fix by making clear delete the pointers, and by making destructor to
call clear instead of deleting pointers by itself.

http://jenkins.percona.com/job/percona-server-5.6-param/1341/
http://jenkins.percona.com/job/percona-server-5.6-valgrind/173/

### DIFF
--- a/plugin/HandlerSocket-Plugin-for-MySQL/libhsclient/auto_ptrcontainer.hpp
+++ b/plugin/HandlerSocket-Plugin-for-MySQL/libhsclient/auto_ptrcontainer.hpp
@@ -12,7 +12,7 @@
 namespace dena {
 
 template <typename Tcnt>
-struct auto_ptrcontainer {
+struct auto_ptrcontainer : private noncopyable {
   typedef Tcnt container_type;
   typedef typename container_type::value_type value_type;
   typedef typename container_type::pointer pointer;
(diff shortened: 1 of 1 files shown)


plugin/HandlerSocket-Plugin-for-MySQL/libhsclient/auto_ptrcontainer.hpp
```diff
@@ -56,7 +54,12 @@ struct auto_ptrcontainer {
   }
   reference operator [](size_type n) { return cnt[n]; }
   const_reference operator [](size_type n) const { return cnt[n]; }
-  void clear() { cnt.clear(); }
+  void clear() {
+    for (iterator i = begin(); i != end(); i++) {
+      delete *i;
+    }
+    cnt.clear();
+  }
```
percona-ysorokin, line 62: For containers like this hiding copy constructor and assignment operator is also a good idea.

percona-ysorokin:
LGTM

laurynas-biveinis:
Updated to inherit from noncopyable, http://jenkins.percona.com/job/percona-server-5.6-param/1345/

percona-ysorokin:
LGTM

PR #939, opened 2016-08-29
Fix bug 1617998 (HandlerSocket may access freed memory on startup)

laurynas-biveinis:
There is a race condition between HS worker threads initializing
default SE and default temp SE fields and the server main thread
initializing the same globally w/o locking.

As HS workers do not care about the default (temp) SE, initialise
their THD handles without plugin support.

http://jenkins.percona.com/job/percona-server-5.6-param/1343/


percona-ysorokin:
LGTM

PR #941, opened 2016-08-29
Bug 1617833: audit_log_default_db failures

gl-sergei:
1. Insert silent statement after UNINSTALL PLUGIN to make sure that SHOW
   WARNINGS is logged before the statement from the next connection.
2. Insert silent statement before establishing the connection which does
   'INSTALL PLUGIN', make sure 'SELECT \* FROM t' gets to the log.
3. Remove problematic 'change_user' because there is no way to
   synchronize it without changing mysqltest.

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/381/

### DIFF
--- a/client/mysqltest.cc
+++ b/client/mysqltest.cc
@@ -4308,12 +4308,16 @@ void do_send_quit(struct st_command *command)
 void do_change_user(struct st_command *command)
 {
   MYSQL *mysql = &cur_con->mysql;
-  static DYNAMIC_STRING ds_user, ds_passwd, ds_db;
+  static DYNAMIC_STRING ds_user, ds_passwd, ds_db, ds_reconnect;
+  bool reconnect = true;
   const struct command_arg change_user_args[] = {
     { "user", ARG_STRING, FALSE, &ds_user, "User to connect as" },
     { "password", ARG_STRING, FALSE, &ds_passwd, "Password used when connecting" },
     { "database", ARG_STRING, FALSE, &ds_db, "Database to select after connect" },
+    { "reconnect", ARG_STRING, FALSE, &ds_reconnect, "Reconnect on fail" },
   };
+  const char *reconnect_on_fail = "reconnect_on_fail";
+  const char *do_not_reconnect_on_fail = "do_not_reconnect_on_fail";
 
   DBUG_ENTER("do_change_user");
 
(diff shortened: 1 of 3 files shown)


laurynas-biveinis:
Would change_user need a big mysqltest change? Testing this command is a good idea, so we should see what's the tradeoff

gl-sergei:
mysqltest change is really small. I have made it. chanhe_user now takes an option telling to reconnect or not. http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/383/

gl-sergei:
@laurynas-biveinis did you have a chance to take a look?

client/mysqltest.cc
```diff
@@ -4346,8 +4347,11 @@ void do_change_user(struct st_command *command)
   {
     handle_error(curr_command, mysql_errno(mysql), mysql_error(mysql),
                  mysql_sqlstate(mysql), &ds_res);
-    mysql->reconnect= 1;
-    mysql_reconnect(&cur_con->mysql);
+    if (ds_reconnect.length == 0 || strcmp(ds_reconnect.str, "true") == 0)
```
laurynas-biveinis: Perha

plugin/audit_log/tests/mtr/audit_log_default_db.test
```diff
@@ -47,7 +49,7 @@ connection test;
 SELECT * FROM t;
 use 	`db1`;
 --error ER_DBACCESS_DENIED_ERROR
-change_user user2,111,db1;
+change_user user2,111,db1,false;
```
laurynas-biveinis: Maybe "reconnect_on_fail, do_not_reconnect_on_fail" or similar so that it's obvious what does "false" mean here?

laurynas-biveinis:
Looking, sorry for the delay

plugin/audit_log/tests/mtr/audit_log_default_db.test
gl-sergei: yeah... it was my first intention to name them this way. Now I did

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/399/

laurynas-biveinis: You posted a new Jenkins link but no branch change here?

gl-sergei:
ah, sorry... I have pushed to my local repo, but not to the github...

here is the new link http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/401/

laurynas-biveinis:
LGTM

PR #942, opened 2016-08-29
Fix bug 1617323 (main.percona_show_temp_tables_debug spuriously skipped)

laurynas-biveinis:
mysql-test/include/have_debug_sync.inc expects session DEBUG_SYNC
variable to have certain values, which is violated by some previous
testcase not cleaning it up. Add a check to
mysql-test/include/mtr_check.sql that the testcase does not change
DEBUG_SYNC value, cleanup several testcases that fail that check.

http://jenkins.percona.com/job/percona-server-5.5-param/1364/


percona-ysorokin:
LGTM

PR #947, opened 2016-08-30
BUG#1618634 [sql] [binlog] fix undefined behavior

bryonglodencissp:
Greetings, on line number 7072 of ‘[sql/binlog.cc](https://github.com/percona/percona-server/blob/5.7/sql/binlog.cc#L7072)’ there is a double-delete causing undefined behavior.

So, after calling ‘`delete pending`’ the first time (on line number 7067), we should do `pending= NULL`, so if (by any chance), ‘`delete pending`’ is called again, nothing will happen.

REF: https://bugs.launchpad.net/percona-server/+bug/1618634

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -7065,6 +7065,7 @@ MYSQL_BIN_LOG::flush_and_set_pending_rows_event(THD *thd,
           stmt_cannot_safely_rollback(thd))
         cache_data->set_incident();
       delete pending;
+      pending= NULL;
       cache_data->set_pending(NULL);
       DBUG_RETURN(1);
     }


laurynas-biveinis:
The code in question reads

```
  if (Rows_log_event* pending= cache_data->pending())
  {
    /*
      Write pending event to the cache.
    */
    if (cache_data->write_event(thd, pending))
    {
      set_write_error(thd, is_transactional);
      if (check_write_error(thd) && cache_data &&
          stmt_cannot_safely_rollback(thd))
        cache_data->set_incident();
      delete pending;
      cache_data->set_pending(NULL);
      DBUG_RETURN(1);
    }

    delete pending;
  }

```

If the first `delete pending is executed`, the method returns (`DBUG_RETURN(1)`). Thus there is no way to reach the 2nd `delete pending` from the 1st one.

laurynas-biveinis:
Please open new bugs/PRs should you find new issues. Thanks!

PR #948, opened 2016-08-30
DB-964 : tokudb.i_s_tokudb_lock_waits_released is unstable

george-lorch:
- Changed comments to echos to better assist in correlating test to result
- Added missing conditional wait that in a race could cause the test to fail.


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1348/

PR #950, opened 2016-08-31
Fix bug 1618478 (innodb.percona_changed_page_bmp might not shutdown s…

laurynas-biveinis:
…erver under 60 seconds under ASan)

Bump the default shutdown_server timeout to 90. This is in line with
e.g. rpl_shutdown_server.inc multiplying shutdown timeout by 6 if
running under Valgrind. Since we cannot determine whether we are
running under ASan or not, just bump the timeout unconditionally.

http://jenkins.percona.com/job/percona-server-5.5-param/1366/


percona-ysorokin:
LGTM

PR #955, opened 2016-08-31
Fix bug 1618811 (MTR Subunit support should report testcase check fai…

laurynas-biveinis:
…lures)

Adjust mtr_report_test_subunit to report an additional testcase
"testcase-check-<original-name>" as failing if the testcase check has
failed.

http://jenkins.percona.com/job/percona-server-5.5-param/1372/


percona-ysorokin:
LGTM

PR #958, opened 2016-08-31
Fix bug 904714 (Intermittent rpl.percona_bug860910 failures)

laurynas-biveinis:
The failing testcase sets up circular replication between two servers,
issues a workload, syncs the servers, and checks that master binlog
position on one server is equal to Exec_Master_Log_Pos. The testcase
intermittently fails that positions are not equal, and when it does,
the 1st server points to the position of 2nd server that comes before
the 1st->2nd replicated events, i.e. while the data is in sync, the
positions aren't, for that the 1st server should receive back its own
events from the 2nd server, and advance log positions without
executing them.

This is caused by rpl_sync.inc include file rpl_generate_sync_chain.inc
generating sync chains that propagate all the data but not necessarily
the log positions as described above in case of loops. Fix by
performing one more synchronisation step in the loop.

http://jenkins.percona.com/job/percona-server-5.5-param/1374/

### DIFF
--- a/mysql-test/include/rpl_generate_sync_chain.inc
+++ b/mysql-test/include/rpl_generate_sync_chain.inc
@@ -45,12 +45,13 @@
 #         a tail, such as 1->2->3->1->4->5. We should first sync the
 #         loop, and then the tail. To ensure all servers in the loop
 #         are synced, we must sync the loop two turns minus two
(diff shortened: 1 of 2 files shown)


percona-ysorokin:
LGTM

PR #968, opened 2016-09-02
Fix bug 1619572 (Test rpl.rpl_init_slave leaves max_connections = 500…

laurynas-biveinis:
… at the end intermittently)

This is caused by the slave SQL thread racing to execute init-slave
with the testcase, and no synchronisation between the two. Add some
workload on the master server that will be replicated by the SQL
thread at the next sync, forcing init-slave execution too.

http://jenkins.percona.com/job/percona-server-5.6-param/1356/


percona-ysorokin:
LGTM

PR #970, opened 2016-09-02
Fix bug 1619665 (Test innodb.innodb_bug30423 is unstable)

laurynas-biveinis:
Backport the 5.6 that masks rows value in EXPLAIN SELECT output for
this testcase, as it's not fully deterministic.

http://jenkins.percona.com/job/percona-server-5.5-param/1380/


percona-ysorokin:
LGTM

PR #973, opened 2016-09-02
Fix bug 1619547 (rpl_end.inc/check-warnings.test/check-testcase.test …

laurynas-biveinis:
…etc do not wait for client disconnects to complete)

In MTR, it is impossible to wait until the last opened server connection
finishes closing (as such check requires another connection). This
concerns e.g. include/rpl_end.inc for slave servers, and testcase
warning/side effects checkers.

Thus replace before/after PROCESSLIST comparison with an assumption that
before and after the testcase there should be only one non-binlog dump
connection opened, optionally an event scheduler thread running, and all
the other connections are in the process of being closed. Assert this
assumption by waiting until that single open connection remains. If it
fails, dump processlist.

http://jenkins.percona.com/job/percona-server-5.5-param/1379/

### DIFF
--- a/mysql-test/include/check-testcase.test
+++ b/mysql-test/include/check-testcase.test
@@ -8,11 +8,23 @@
 
 --disable_query_log
 
-# Check that the testcase has closed (and the server has finished cleaning up)
-# all the connection it opened. Exclude binlog dump threads as these may
+# Wait for any connections opened by a previous testcase (or the testcase
+# warning checker) to finish closing. Exclude binlog dump threads as these may
 # linger around indefinitely.
---replace_column 1 <Id> 6 <Time> 7 <State> 9 X 10 X 11 X
-SELECT * FROM INFORMATION_SCHEMA.PROCESSLIST WHERE COMMAND != 'Binlog Dump';
+--disable_result_log
+SET @expected_thread_count= 1;
+--let $event_scheduler_running= `SELECT @@event_scheduler = 'ON'`
+if ($event_scheduler_running == 1)
+{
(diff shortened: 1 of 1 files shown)


mysql-test/include/check-testcase.test
```diff
@@ -8,11 +8,23 @@
...
-# Check that the testcase has closed (and the server has finished cleaning up)
-# all the connection it opened. Exclude binlog dump threads as these may
+# Wait for any connections opened by a previous testcase (or the testcase
+# warning checker) to finish closing. Exclude binlog dump threads as these may
 # linger around indefinitely.
---replace_column 1 <Id> 6 <Time> 7 <State> 9 X 10 X 11 X
-SELECT * FROM INFORMATION_SCHEMA.PROCESSLIST WHERE COMMAND != 'Binlog Dump';
+--disable_result_log
+SET @expected_thread_count= 1;
+--let $event_scheduler_running= `SELECT @@event_scheduler = 'ON'`
```
percona-ysorokin, line 16: May be we should enforce that `@@event_scheduler` must always be `OFF` at the end of the test and therefore do not add this `@expected_thread_count + 1` here?

laurynas-biveinis: It is legal to have it ON at the end, if -master.opt has it ON, or even if `./mtr --suite=main --mysqld=event_scheduler=ON`

percona-ysorokin: Good point

percona-ysorokin:
LGTM

PR #974, opened 2016-09-02
Fix bug 1619547 (rpl_end.inc/check-warnings.test/check-testcase.test …

laurynas-biveinis:
…etc do not wait for client disconnects to complete)

In MTR, it is impossible to wait until the last opened server connection
finishes closing (as such check requires another connection). This
concerns e.g. include/rpl_end.inc for slave servers, and testcase
warning/side effects checkers.

Thus replace before/after PROCESSLIST comparison with an assumption that
before and after the testcase there should be only one non-binlog dump
connection opened, optionally an event scheduler thread running, and all
the other connections are in the process of being closed. Assert this
assumption by waiting until that single open connection remains. If it
fails, dump processlist.

http://jenkins.percona.com/job/percona-server-5.6-param/1358/

### DIFF
--- a/mysql-test/include/check-testcase.test
+++ b/mysql-test/include/check-testcase.test
@@ -8,12 +8,23 @@
 
 --disable_query_log
 
-# Check that the testcase has closed (and the server has finished cleaning up)
-# all the connection it opened. Exclude binlog dump threads as these may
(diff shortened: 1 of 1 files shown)


percona-ysorokin:
LGTM

PR #980, opened 2016-09-06
Fix bug 1617198 (Intermittently stalled or crashing shutdown if it's …

laurynas-biveinis:
…issued immediately after HandlerSocket startup)

If server is shutdown immediately after HandlerSocket plugin has
initialized, it may hit a race condition between worker thread
initialisation, which happens asynchronously, and at some point adds
threads to the server global thread list, and server shutdown, which
waits for the thread count to go down to zero. This race condition may
result in some worker threads not receiving kill notifications, and
thus hanging the server shutdown.

Fix by making the worker threads bump the worker thread count and
raise a signal after the last one has been started. Wait for that
signal in the spawning thread. Move the server startup check in the
worker threads after the signal.

In the core server, add assert that server is not being shutdown to
add_global_thread function.

http://jenkins.percona.com/job/percona-server-5.6-param/1359/

### DIFF
--- a/plugin/HandlerSocket-Plugin-for-MySQL/handlersocket/hstcpsvr.hpp
+++ b/plugin/HandlerSocket-Plugin-for-MySQL/handlersocket/hstcpsvr.hpp
@@ -38,9 +38,19 @@ struct hstcpsvr_shared_c {
     thread_num_conns(0) { }
 };
 
-struct hstcpsvr_shared_v : public mutex {
+struct hstcpsvr_shared_v : private noncopyable {
   int shutdown;
-  hstcpsvr_shared_v() : shutdown(0) { }
+  long threads_started;
+  pthread_cond_t threads_started_cond;
+  mutex v_mutex;
+  hstcpsvr_shared_v() : shutdown(0), threads_started(0)
+  {
+    pthread_cond_init(&threads_started_cond, NULL);
+  }
+  ~hstcpsvr_shared_v()
+  {
+    pthread_cond_destroy(&threads_started_cond);
+  }
 };
 
 struct hstcpsvr_i;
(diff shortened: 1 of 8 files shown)


plugin/HandlerSocket-Plugin-for-MySQL/handlersocket/hstcpsvr.hpp
```diff
@@ -38,9 +38,19 @@ struct hstcpsvr_shared_c {
     thread_num_conns(0) { }
 };
 
-struct hstcpsvr_shared_v : public mutex {
+struct hstcpsvr_shared_v {
```
percona-ysorokin: noncopyable?

plugin/HandlerSocket-Plugin-for-MySQL/handlersocket/hstcpsvr.hpp
```diff
@@ -38,9 +38,19 @@ struct hstcpsvr_shared_c {
...
   int shutdown;
-  hstcpsvr_shared_v() : shutdown(0) { }
+  long threads_started;
```
percona-ysorokin, line 43: why signed type?

laurynas-biveinis: To match hstcpsvr_shared_c::num_threads. Fixing

laurynas-biveinis: Changing it back, causing too much of cascading cleanups

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.6-param/1363/

percona-ysorokin:
LGTM

PR #984, opened 2016-09-06
Fix bug 1620583 (Test change_user_notembedded is unstable)

laurynas-biveinis:
Cleanup the testcase, including that the each authorization failure
can return one of of the three failures, depending on the timing.

http://jenkins.percona.com/job/percona-server-5.6-param/1361/


percona-ysorokin:
LGTM

PR #988, opened 2016-09-08
Fix bug 1621012 / 82886 (Server may crash due to a glibc bug in handl…

laurynas-biveinis:
…ing short-lived detached threads)

Glibc is affected by a bug where a short-lived detached thread, that
manages to complete before the caller thread finishes executing
pthread_create, may crash server. In MySQL, at least the InnoDB full
text parallel merge threads might be short-lived enough to hit this.

Implement a workaround to join these threads instead of detaching
them.

http://jenkins.percona.com/job/percona-server-5.6-param/1364/


percona-ysorokin:
LGTM

PR #990, opened 2016-09-08
Work towards diagnosing bug 1621060 (Test innodb.innodb_stats_auto_re…

laurynas-biveinis:
…cals fails intermittently)

Replace "die" with "echo" in mysql-test/include/wait_condition.inc in
case of failure to allow the testcase to proceed and show the result
diff. This also makes this include file consistent with both 5.5 and
5.7.

http://jenkins.percona.com/job/percona-server-5.6-param/1367/


percona-ysorokin:
LGTM

PR #992, opened 2016-09-08
Fix bug 1619559 (Thread pool and lock tables cause deadlock)

laurynas-biveinis:
The bug is caused by LOCK TABLES thread going to low-priority thread
queue with the lock taken, then SELECT connections piling up until
thread pool oversubscribe limit is reached. At this point, the LOCK
TABLES thread will never be dequeued from the low priority queue,
resulting in a deadlock.

Fix by implementing bug 1374930 (put threads with GLOBAL READ lock
(executing ftwrl) into high priority queue) suggestion: for
thread_pool_high_prio_mode = transactions mode, put into the high
priority queue connections not only with open transactions, but also
connections that hold global read, binlog, or tables backup lock, has
explicitly locked some tables, or has taken a user-level lock.

http://jenkins.percona.com/job/percona-server-5.6-param/1366/


sql/threadpool_unix.cc
```diff
@@ -437,16 +437,21 @@ inline bool too_many_busy_threads(thread_group_t *thread_group)
...
-    (mode == TP_HIGH_PRIO_MODE_TRANSACTIONS &&
-     c->tickets > 0 && thd_is_transaction_active(c->thd));
+    (mode == TP_HIGH_PRIO_MODE_TRANSACTIONS && c->tickets > 0 &&
+     (thd_is_transaction_active(c->thd) ||
+      c->thd->variables.option_bits & OPTION_TABLE_LOCK ||
+      c->thd->global_read_lock.is_acquired() ||
+      c->thd->backup_tables_lock.is_acquired() ||
+      c->thd->backup_binlog_lock.is_acquired() ||
+      c->thd->ull_hash.records > 0));
 }
```
gl-sergei, line 457: What about `thd->locked_tables_mode`? `thd->mdl_context.has_locks()`?

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.6-param/1383/

gl-sergei (approved):
LGTM

PR #994, opened 2016-09-12
Fix bug 1622449 (Test innodb.innodb_replace is unstable)

laurynas-biveinis:
Prevent INSERT on con1 overtaking SELECT LOCK IN SHARE MODE on the
default connection by reaping its query first.

http://jenkins.percona.com/job/percona-server-5.6-param/1369/


percona-ysorokin:
LGTM

PR #998, opened 2016-09-13
servers system table should explicitly set engine=MyISAM like the oth…

mig5:
I believe 868dbf4ba27ac4974b98f11c835ae367bdeb0022 may have caused a regression that breaks MySQL 'full' replication (e.g not specific db replication but all binary log, where the mysql database itself is also replicated)

Specifically the script scripts/mysql_system_tables.sql tries to run 'CREATE TABLE IF NOT EXISTS' queries for the system tables. Just one of them 'servers' does not explicitly set engine=MyISAM   (all the rest do).

Possibly due to the commit mentioned above, this now causes a strict error on a slave that is replicating queries sent to the mysql database, since Percona release 5.5.49-rel37.9-1, 'InnoDB does not support system tables'

This commit fixes the CREATE TABLE statement (but still remain concerned by the change as this could occur again in future if no explicit engine is set on any new system tables, however unlikely that may be)

See https://bugs.launchpad.net/percona-server/+bug/1600056 and https://groups.google.com/forum/#!msg/percona-discussion/2eZtWaqTMPM/KGdtgG6CBgAJ


laurynas-biveinis:
I have applied your fix in https://github.com/percona/percona-server/pull/1014. Thanks!

PR #1000, opened 2016-09-13
Fix bug 1622977 (Test connect_debug is unstable)

laurynas-biveinis:
Avoid race condition between the server and the testcase by replacing
the processlist-based disconnect check with a Threads_connected-based
one.

http://jenkins.percona.com/job/percona-server-5.6-param/1373/


percona-ysorokin:
LGTM

PR #1004, opened 2016-09-13
Fix bug 1622985 (Successful doublewrite recovery should not be a warn…

laurynas-biveinis:
…ing)

Downgrade diagnostics severity from warning to regular note if the
doublewrite user has been used for crash recovery. Remove the
corresponding suppressions from the testcases.

http://jenkins.percona.com/job/percona-server-5.6-param/1374/


percona-ysorokin:
LGTM

PR #1006, opened 2016-09-14
#1065841 use --skip-write-binlog on mysql upgrade, to avoid breaking …

mig5:
…replication on slaves


laurynas-biveinis:
We have fixed this in https://github.com/percona/percona-server/pull/1017, which changes mysql_upgrade option default itself.

PR #1007, opened 2016-09-14
#1065841 use --skip-write-binlog on mysql upgrade, to avoid breaking …

mig5:
…replication on slaves


laurynas-biveinis:
Closing, as mysql_upgrade already has --skip-write-binlog default in 5.6.

PR #1008, opened 2016-09-14
Add asserts to catch bug 1622868 (LSN rolls back by 10 on shutdown in…

laurynas-biveinis:
…termittently)

Specifically, assert that
- next checkpoint LSN only advances forward, both in log_sys and in
  log file;
- no checkpoints happen at the very last phase of InnoDB shutdown;
- last_checkpoint_lsn does not move during the same phase.

http://jenkins.percona.com/job/percona-server-5.6-param/1371/


percona-ysorokin:
LGTM

PR #1010, opened 2016-09-14
Add assert to catch bug 1622952 (Test innodb.innodb-log-file-size-1 i…

laurynas-biveinis:
…s unstable)

InnoDB shutdown hangs with error messages indicating I/O attempts to
non-existing tablespace, where the tablespace is the system one, that
must exist. Add an assert to catch this condition.

http://jenkins.percona.com/job/percona-server-5.6-param/1372/


percona-ysorokin:
LGTM

PR #1012, opened 2016-09-14
Fix bug 82980 / 1619622 (Multi-threaded slave leaks worker threads in…

laurynas-biveinis:
… case of thread create failure)

Move early exits in the slave_stop_workers to the middle of the function
so that the running slave worker threads are terminated.

http://jenkins.percona.com/job/percona-server-5.6-param/1379/


percona-ysorokin (approved):
LGTM

PR #1014, opened 2016-09-15
Fix bug 1600056 (Slave replication fails on master upgrade 'InnoDB do…

laurynas-biveinis:
…es not support system tables')

After bug 1540338, system table engine validation check is no longer
skipped for tables that don't have an explicit ENGINE=foo clause in
CREATE TABLE statement. If MySQL upgrade statements are replicated,
and slave does not have MyISAM set as a default SE, then the CREATE
TABLE mysql.server statement will attempt to create an InnoDB table
and fail.

Fix by applying the contributed fix by Miguel Jacq at
https://github.com/percona/percona-server/pull/998, which adds
explicit ENGINE= clause to CREATE TABLE mysql.server.


percona-ysorokin (approved):
LGTM

PR #1017, opened 2016-09-15
BUG#14221043: MYSQL_UPGRADE FAILS WHEN SERVER IS RUNNING WITH GTID-MO…

laurynas-biveinis:
…DE=ON

mysql_upgrade fails when server is running with gtid_mode=ON and
--disable-gtid-unsafe-statements because system tables are stored
using MyISAM.

Changed default logging behaviour of mysql_upgrade, disabling it
by default. Logging can still be enable using the option
--write-binlog.
Actions taken by mysql_upgrade depend on the server version, them
should not be replicated to slaves.

http://jenkins.percona.com/job/percona-server-5.5-param/1385/


percona-ysorokin (approved):
LGTM

PR #1023, opened 2016-09-16
Bug 1620152: The audit log plugin uses too much memory

gl-sergei:
1. Backport fix for bug #19917521 from MySQL 8.
2. Fix memory leaks in audit plugin using THDVAR_SET.
3. Fix memory leaks when non-unique keys inserted into the hash table.

### DIFF
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -116,6 +116,8 @@ ulonglong next_record_id()
 #define MAX_RECORD_ID_SIZE  50
 #define MAX_TIMESTAMP_SIZE  25
 
+void plugin_thdvar_safe_update(MYSQL_THD thd, struct st_mysql_sys_var *var,
+                               char **dest, const char *value);
 
 static
 void fprintf_timestamp(FILE *file)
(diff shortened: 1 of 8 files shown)


gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1387/

plugin/audit_log/audit_log.c
```diff
@@ -1446,6 +1448,14 @@ static struct st_mysql_sys_var* audit_log_system_variables[] =
   NULL
 };
 
+char thd_local_init_buf[sizeof(audit_log_thd_local)];
```
laurynas-biveinis, line 1451: Is this still required?

laurynas-biveinis (changes requested):
Please add a link to the public upstream bug https://bugs.mysql.com/bug.php?id=71759 to LP and an upstream tag - on 1620152 or create new one, as you see most appropriate

plugin/audit_log/audit_log.c
gl-sergei: THDVAR_SET makes a copy of provided string and stores it in thread variables. So, there are two options:
1. Create static buffer, fill it with sizeof(audit_log_local) bytes, pass it to THDVAR_SET each time
2. Allocate dynamic buffer and fill it for every new THD, pass it to THDVAR_SET and then release it.
I went with first option.

laurynas-biveinis (approved):

PR #1024, opened 2016-09-16
Fix for bug 1620152 (5.7)

gl-sergei:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-valgrind-param/42/


laurynas-biveinis (changes requested):
Same as 5.6

laurynas-biveinis (approved):

PR #1028, opened 2016-09-16
Fix bug 1622034 / 82935 (Cipher ECDHE-RSA-AES128-GCM-SHA256 listed in…

laurynas-biveinis:
… man/Ssl_cipher_list, not supported)

Invoke OpenSSL eliptic curve setup functions so that EC-DHE ciphers
are supported. Do this conditionally if the SSL library supports EC.  To
properly skip testcases when it does not, add a
have_elliptic_curve_crypto server read-only variable (the name is chosen
specifically not to contain "ssl" substring in order to avoid updating
all the "SHOW STATUS LIKE "%ssl%" testcases). Re-record testcases which
now negotiate EC ciphers by default.


laurynas-biveinis:
@Oneiroi , FYI. TL;DR: MySQL/Percona Server OpenSSL setup did not the necessary setup to enable elliptic curve-based cipher, now it does. It has an effect that server now negotiates ECDHE-RSA-AES128-GCM-SHA256 instead of DHE-RSA-AES128-GCM-SHA256 by default.

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.6-param/1386/

mysql-test/t/ssl_ecdh.test
```diff
@@ -0,0 +1,20 @@
+--source include/have_ssl_communication.inc
+--source include/have_elliptic_curve_crypto.inc
+
+--source include/count_sessions.inc
+
+--echo #
+--echo # Bug 82935: Cipher ECDHE-RSA-AES128-GCM-SHA256 listed in man/Ssl_cipher_list, not supported
+--echo #
+
+eval GRANT SELECT ON test.* TO ecdh@localhost REQUIRE CIPHER "ECDHE-RSA-AES128-GCM-SHA256";
```
percona-ysorokin: unnecessary 'eval'

percona-ysorokin (approved):
LGTM

laurynas-biveinis:
Fixed, http://jenkins.percona.com/job/percona-server-5.6-param/1401/

PR #1036, opened 2016-09-20
BUG#16906827-CAN'T SET QUERY_CACHE_TYPE TO 0 WHEN IT IS ALREADY 0

laurynas-biveinis:
BACKGROUND:
The crux of the problem reported in the bug report is User 
does not want to see the error message when the server is 
started with query cache disabled and User is disabling the
query cache by setting query_cache_type to 0 during runtime 
(in client session).

ANALYSIS:
The check in check_query_cache_type() function was throwing
error for all three values 0, 1 or 2 when query cache is disabled.
According to the bug or user requirement, User wants to bypass 
this error check when we are setting query_cache_type to 0 (or OFF)
during runtime when it is already disabled 0 (or OFF) during startup.

FIX:
As a fix for this bug, the check in check_query_cache_type() 
function is modified to take into account to _NOT_ throw the 
error when we are setting it 0 (or OFF) when query cache is 
already disabled.

http://jenkins.percona.com/job/percona-server-5.6-param/1402/


percona-ysorokin (commented):
LGTM

PR #1040, opened 2016-09-21
Work towards diagnosing bug 1625466 (Test innodb.percona_changed_page…

laurynas-biveinis:
…_bmp_debug is unstable (failure to find system record))

Add bitmap file size and read offset to the "incomplete run at the
end" diagnostics. Tighten the conditions for the
bitmap_page_2_write_error debug injection.

Commit on trunk instead of GCA to include other recent percona_changed_page_bmp_debug fixes.
http://jenkins.percona.com/job/percona-server-5.6-param/1406/


percona-ysorokin (approved):
LGTM

PR #1042, opened 2016-09-21
Fix bug 83073 / 1626002 (GCC 5 and 6 miscompile mach_parse_compressed)

laurynas-biveinis:
For buggy compiler versions, add a compiler memory barrier to prevent
moving the 4-byte load to the beginning of the function, where it's
not safe to perform it.

http://jenkins.percona.com/job/percona-server-5.5-param/1388/


percona-ysorokin (approved):
LGTM

PR #1054, opened 2016-09-22
Fix bug 1626069 (LRU manager thread might run too long on server shut…

laurynas-biveinis:
…down)

Synchronize server shutdown thread with LRU manager thread quitting:
- during SRV_SHUTDOWN_FLUSH_PHASE, wait for it to quit, in addition to
  checking whether the page cleaner thread has quit;
- adjust sleep heuristics for both cleaner and LRU manager threads to
  skip any sleeps during SRV_SHUTDOWN_FLUSH_PHASE.

http://jenkins.percona.com/job/percona-server-5.6-param/1409/


percona-ysorokin (approved):
LGTM

PR #1056, opened 2016-09-22
Bug#13923795: Stop slave where we test and restart master only

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.5-param/1389/


percona-ysorokin (approved):
LGTM

PR #1061, opened 2016-09-22
Fix bug 1626500 (Test rpl.rpl_percona_bug1070255 is unstable)

laurynas-biveinis:
Rewrite the testcase to
- avoid updating global DEBUG variable while a slave is connected to
  master;
- save and restore DEBUG around modifications;
- pop DEBUG states before pushined new ones.

http://jenkins.percona.com/job/mysql-5.7-param/441/


percona-ysorokin (approved):
LGTM

PR #1062, opened 2016-09-23
Fix bug 71761 / 1626441 (ANALYZE TABLE should remove its table from b…

laurynas-biveinis:
…ackground stat processing queue)

Make InnoDB ANALYZE TABLE remove its table from the background stat
processing queue, if it's there.

Remove the workaround for this bug from percona_log_slow_query_plan
testcase. Remove obsolete testcase result file
mysql-test/r/percona_bug1308016.result.

http://jenkins.percona.com/job/percona-server-5.6-param/1412/


percona-ysorokin (approved):
LGTM

PR #1064, opened 2016-09-23
Work towards diagnosing bug 1626545 (Test perfschema.selects is unsta…

laurynas-biveinis:
…ble)

Read Performance Schema status after the failed query to see whether
there have been any dropped events.

http://jenkins.percona.com/job/percona-server-5.6-param/1414/


percona-ysorokin (approved):
LGTM

PR #1030, opened 2016-09-18
DB-1012: Add read-free-replication support for TokuDB partition table

fungo:

PR #1070, opened 2016-09-26
DB-1012: Add read-free-replication support for TokuDB partition table

george-lorch:
Copyright (c) <2016>, <Fungo Wang> All rights reserved.

Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions
are met:
1. Redistributions of source code must retain the above copyright
   notice, this list of conditions and the following disclaimer.
2. Redistributions in binary form must reproduce the above copyright
   notice, this list of conditions and the following disclaimer in the
   documentation and/or other materials provided with the distribution.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
OF THE POSSIBILITY OF SUCH DAMAGE.

### DIFF
--- a/mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
+++ b/mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
@@ -1,0 +1,74 @@
+# test tokudb read free replication feature with partition table
+
+--source include/have_debug.inc
+--source include/have_tokudb.inc
+--source include/have_binlog_format_row.inc
+--source include/master-slave.inc
+
+call mtr.add_suppression(".*read free replication is disabled for TokuDB table.*continue with rows lookup");
+
+connection master;
+
+# partition table with explicit PK
+CREATE TABLE t1 (id int(11) NOT NULL, pid int(11), PRIMARY KEY (id)) ENGINE=TokuDB
+PARTITION BY RANGE (id)
+(PARTITION p_1 VALUES LESS THAN (10) ENGINE = TokuDB,
+ PARTITION p_2 VALUES LESS THAN (20) ENGINE = TokuDB,
+ PARTITION p_all VALUES LESS THAN MAXVALUE ENGINE = TokuDB);
+
+insert into t1 values (1, 1), (2, 2), (3, 3), (11, 11), (12, 12), (13, 13);
+
+# partition table without explicit PK
+CREATE TABLE t2 (id int(11) NOT NULL, pid int(11), key idx_1(id)) ENGINE=TokuDB
+PARTITION BY RANGE (id)
+(PARTITION p_1 VALUES LESS THAN (10) ENGINE = TokuDB,
+ PARTITION p_2 VALUES LESS THAN (20) ENGINE = TokuDB,
+ PARTITION p_all VALUES LESS THAN MAXVALUE ENGINE = TokuDB);
+
+insert into t2 values (1, 1), (2, 2), (3, 3), (11, 11), (12, 12), (13, 13);
+
+--sync_slave_with_master
+
+# set tokudb rfr crash/assert conditions if we enter lookup code
+# to make sure no unique checks or row lookups is invoked
+connection slave;
+--source include/stop_slave.inc
+let $saved_debug = `select @@debug`;
+set global debug= "+d,tokudb_crash_if_rpl_looks_up_row,tokudb_crash_if_rpl_does_uniqueness_check";
+--source include/start_slave.inc
+
+connection master;
+insert into t1 values(21, 21);
+delete from t1 where id = 11;
+update t1 set pid = 2 where id = 1;
+
+sync_slave_with_master;
+
+connection master;
+
+--let $diff_tables= master:test.t1, slave:test.t1
+--source include/diff_tables.inc
+
+# print rfr disabled warning in errlog
+connection master;
+insert into t2 values(21, 21);
+delete from t2 where id = 11;
+update t2 set pid = 2 where id = 1;
+
+sync_slave_with_master;
+
+--let $diff_tables= master:test.t2, slave:test.t2
+--source include/diff_tables.inc
+
+connection master;
+drop table t1;
+drop table t2;
+sync_slave_with_master;
+
+connection slave;
+--source include/stop_slave.inc
+set global debug= "-d,tokudb_crash_if_rpl_looks_up_row,tokudb_crash_if_rpl_does_uniqueness_check";
+set global debug= @saved_debug;
+--source include/start_slave.inc
+
+--source include/rpl_end.inc
--- a/sql/ha_partition.cc
+++ b/sql/ha_partition.cc
@@ -4089,10 +4089,14 @@ int ha_partition::update_row(const uchar *old_data, uchar *new_data)
     error instead of correcting m_last_part, to make the user aware of the
     problem!
 
+    For TokuDB Read-Free-Replication optimization, there is no need to do
+    a read before update(row lookup is omitted), so m_last_part is not
+    necessarily same with old_part_id.
+
     Notice that HA_READ_BEFORE_WRITE_REMOVAL does not require this protocol,
     so this is not supported for this engine.
   */
-  if (old_part_id != m_last_part)
+  if (old_part_id != m_last_part && rpl_lookup_rows())
   {
     m_err_rec= old_data;
     DBUG_RETURN(HA_ERR_ROW_IN_WRONG_PARTITION);
@@ -4229,13 +4233,17 @@ int ha_partition::delete_row(const uchar *buf)
     error instead of forwarding the delete to the correct (m_last_part)
     partition!
 
+    For TokuDB Read-Free-Replication optimization, there is no need to do
+    a read before delete(row lookup is omitted), so m_last_part is not
+    necessarily same with part_id.
+
     Notice that HA_READ_BEFORE_WRITE_REMOVAL does not require this protocol,
     so this is not supported for this engine.
 
     TODO: change the assert in InnoDB into an error instead and make this one
     an assert instead and remove the get_part_for_delete()!
   */
-  if (part_id != m_last_part)
+  if (part_id != m_last_part && rpl_lookup_rows())
   {
     m_err_rec= buf;
     DBUG_RETURN(HA_ERR_ROW_IN_WRONG_PARTITION);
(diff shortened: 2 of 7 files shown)


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1419/

This supersedes https://github.com/percona/percona-server/pull/1030

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
```diff
@@ -0,0 +1,75 @@
...
+insert into t1 values(21, 21);
+delete from t1 where id = 11;
+update t1 set pid = 2 where id = 1;
+
+sync_slave_with_master;
+
+# make sure no unique checks or row lookups is invoked
+connection master;
+select unix_timestamp() into @tend;
+select @tend-@tstart < 10;
```
laurynas-biveinis: Testcases may run arbitrarily slowly, please remove this

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
```diff
@@ -0,0 +1,75 @@
...
+
+--let $diff_tables= master:test.t1, slave:test.t1
+--source include/diff_tables.inc
+
+# print rfr disabled warning in errlog
+connection master;
+select unix_timestamp() into @tstart;
```
laurynas-biveinis: Likewise (this seems to be unused anyway)

sql/ha_partition.cc
```diff
@@ -9036,6 +9044,92 @@ int ha_partition::check_for_upgrade(HA_CHECK_OPT *check_opt)
...
+{
+  uint i;
+  for (i= 0; i < m_tot_parts; i++)
+  {
+    m_file[i]->rpl_before_write_rows();
+  }
+}
+
+/*
+  Clear flag of all paritions after executing Write_rows_log_event.
```
laurynas-biveinis: "partitions"

sql/ha_partition.cc
```diff
@@ -9036,6 +9044,92 @@ int ha_partition::check_for_upgrade(HA_CHECK_OPT *check_opt)
...
+
+  uint i;
+  for (i= 0; i < m_tot_parts; i++)
+  {
+    m_file[i]->rpl_before_delete_rows();
+  }
+}
+
+/*
+  Clear flag of all paritions after executing Delete_rows_log_event.
```
laurynas-biveinis: "partitions"

sql/ha_partition.cc
```diff
@@ -9036,6 +9044,92 @@ int ha_partition::check_for_upgrade(HA_CHECK_OPT *check_opt)
...
+{
+  uint i;
+  for (i= 0; i < m_tot_parts; i++)
+  {
+    m_file[i]->rpl_before_update_rows();
+  }
+}
+
+/*
+  Clear flag of all paritions after executing Update_rows_log_event.
```
laurynas-biveinis: "partitions"

laurynas-biveinis (changes requested):

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
george-lorch: This is a negative test, the delays should only get hit if RFR is not functioning. Without this, there is no way to know that the actual read free operation took place. DEBUG_SYNC would be ideal here but since we need the sync point activated on the slave SQL thread, there is no way to make use of it.

laurynas-biveinis: This snippet may run arbitrarily slowly, resulting in tend - tstart > 10. MTR cannot do timing-based testing. You can add a DBUG_EXECUTE_IF point instead of DBUG_SYNC

george-lorch: That is an excellent suggestion and exactly what I was looking for instead of DEBUG_SYNC. Don't know why I didn't think of it in the first place.

george-lorch:
Changed mtr test and code to use DBUG_EXECUTE_IF instead of timeouts. New jenkins http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1423/

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
```diff
@@ -0,0 +1,70 @@
...
+ PARTITION p_2 VALUES LESS THAN (20) ENGINE = TokuDB,
+ PARTITION p_all VALUES LESS THAN MAXVALUE ENGINE = TokuDB);
+
+insert into t2 values (1, 1), (2, 2), (3, 3), (11, 11), (12, 12), (13, 13);
+
+--sync_slave_with_master
+
+# set tokudb rfr delay
+connection slave;
+set global debug= "+d,tokudb_in_rpl_lookup_rows,tokudb_rpl_do_unique_checks";
```
laurynas-biveinis: This facility is not multi-thread-safe for concurrent read/write, here you are updating its setting while an async background thread - slave SQL thread - might be reading it, which results in crashes like https://bugs.mysql.com/bug.php?id=58754. I don't know if fully-synced SQL thread is safe here or not. Since you are restarting the slave SQL thread here anyway, just move the stop slave sql_thread above the debug setting (and replace both stop/start slave with include/stop|start_slave_sql.inc for synchronisation)

laurynas-biveinis: Also I'd name the debug point something like "crash_if_rpl_looks_up_a_row", "crash_if_rpl_does_uniqueness_check", to make it clear what is the test objective here

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
```diff
@@ -0,0 +1,70 @@
...
+start slave sql_thread;
+
+connection master;
+insert into t1 values(21, 21);
+delete from t1 where id = 11;
+update t1 set pid = 2 where id = 1;
+
+sync_slave_with_master;
+
+# make sure no unique checks or row lookups is invoked
```
laurynas-biveinis: This comment out of place?

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
```diff
@@ -0,0 +1,70 @@
...
+--let $diff_tables= master:test.t2, slave:test.t2
+--source include/diff_tables.inc
+
+connection master;
+drop table t1;
+drop table t2;
+sync_slave_with_master;
+
+connection slave;
+set global debug= default;
```
laurynas-biveinis: Same comment re. stopped slave thread. Also I think 100% correct reset would be 1) set global debug="-d,everything_what_was_+d'ed_above'; 2) set global debug=@saved_start_debug_value. This would both restore the debug state and make MTR --debug=print-trace-of-this-or-that work

laurynas-biveinis (changes requested):

storage/tokudb/ha_tokudb.cc
```diff
@@ -3695,6 +3695,7 @@ int ha_tokudb::do_uniqueness_checks(uchar* record, DB_TXN* txn, THD* thd) {
     // first do uniqueness checks
     //
     if (share->has_unique_keys && do_unique_checks(thd, in_rpl_write_rows)) {
+        DBUG_EXECUTE_IF("tokudb_rpl_do_unique_checks", DBUG_SUICIDE(););
```
laurynas-biveinis: Maybe DBUG_ASSERT_UNREACHABLE or DBUG_ASSERT(0) to make it clear that this is not a crash injection site but a bug catcher (and also so that failure results in a stacktrace)

storage/tokudb/ha_tokudb.cc
```diff
@@ -5915,6 +5916,7 @@ int ha_tokudb::rnd_pos(uchar * buf, uchar * pos) {
     // test rpl slave by inducing a delay before the point query
     THD *thd = ha_thd();
     if (thd->slave_thread && (in_rpl_delete_rows || in_rpl_update_rows)) {
+        DBUG_EXECUTE_IF("tokudb_in_rpl_lookup_rows", DBUG_SUICIDE(););
```
laurynas-biveinis: Likewise

george-lorch:
All makes sense. I didn't know about the race bug with the debug variable. Changed things up as suggested and another jenkins bit http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1427/

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
```diff
@@ -0,0 +1,71 @@
...
+ PARTITION p_all VALUES LESS THAN MAXVALUE ENGINE = TokuDB);
+
+insert into t2 values (1, 1), (2, 2), (3, 3), (11, 11), (12, 12), (13, 13);
+
+--sync_slave_with_master
+
+# set tokudb rfr crash/assert conditions if we enter lookup code
+# to make sure no unique checks or row lookups is invoked
+connection slave;
+--source include/stop_slave_sql.inc
```
laurynas-biveinis: Stop both slave threads?

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
```diff
@@ -0,0 +1,71 @@
...
+--let $diff_tables= master:test.t2, slave:test.t2
+--source include/diff_tables.inc
+
+connection master;
+drop table t1;
+drop table t2;
+sync_slave_with_master;
+
+connection slave;
+set global debug= @saved_debug;
```
laurynas-biveinis, line 71: --source include/stop_slave.inc
set global debug= "-d,foo,bar";
set global debug= @saved_debug;
Check whether it's OK to do rpl_end.inc with the slave stopped or whether start_slave.inc is required

laurynas-biveinis (changes requested):

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
george-lorch: Sure, why not, the original test was just stopping the SQL thread, but I suppose due the nature of the 'set debug' race it makes sense to stop everything and can't hurt anything except maybe take a few more CPU cycles.

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
george-lorch: Hmm, not sure, suppose will have to look into stop_slave.inc. I would have expected it to throw errors or something if it didn't like it.

george-lorch: OK, it is interesting, it is not dangerous but still probably not great. rpl_end.inc does a rpl_sync.inc which only actually syncs if both the rpl threads are running, so after rpl_end.inc one can assume that the master and slave(s) are in sync, but, if either of the rpl threads were stopped before rpl_end.inc, then they may not be in sync, breaking an assumption about rpl_end.inc. So, summary is yeah, it is probably good form to ensure rpl is running in the same state as it was after master-slave.inc.

george-lorch:
Changes pushed and new jenkins http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1429/

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
```diff
@@ -0,0 +1,73 @@
...
+--source include/diff_tables.inc
+
+connection master;
+drop table t1;
+drop table t2;
+sync_slave_with_master;
+
+connection slave;
+--source include/stop_slave.inc
+set global debug= @saved_debug;
```
laurynas-biveinis, line 71: set global debug= "-d,foo,bar" - please see the earlier comments here

laurynas-biveinis (changes requested):

mysql-test/suite/tokudb.rpl/t/rpl_tokudb_rfr_partition_table.test
george-lorch: Ack, yeah, sorry. I totally missed that part of the original comment.

george-lorch:
Made the change and a new jenkins http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1430/

laurynas-biveinis (approved):

PR #1072, opened 2016-09-27
Fix bug 1552428 (QUERY_RESPONSE_TIME_READ & QUERY_RESPONSE_TIME_WRITE…

laurynas-biveinis:
… broken if accessed through non-uppercase name)

The table selector in query_response_time_fill performed exact-case
comparison against uppercase table name constants, resulting in
read/write tables returning the common table data, if accessed
through a name that is not full uppercase.

Fix by splitting the query_response_time_fill function into three, one
per table, and move the table choice decision to
query_respone_time_info_init, with a case-insensitive comparison.

Expand the query_response_time-rw testcase to catch this and also
assert that the common table is a sum of read and write tables.

http://jenkins.percona.com/job/percona-server-5.6-param/1420/


percona-ysorokin (approved):
LGTM

PR #1077, opened 2016-09-28
Fix bug 1617150 (Test rpl.rpl_semi_sync_uninstall_plugin is unstable)

laurynas-biveinis:
The instability is caused by a race condition between master server
disconnecting the async slave asynchronously, and the testcase
checking whether the disconnect has already completed. Replace the
assert with a wait for the disconnect to complete.

http://jenkins.percona.com/job/percona-server-5.5-param/1395/


percona-ysorokin (approved):
LGTM

PR #1082, opened 2016-09-28
Release 5.6.32 78.0

EvgeniyPatlan:


scripts/mysqld_safe.sh
```diff
@@ -805,6 +806,7 @@ then
   if [ ! -h "$pid_file" ]; then
       rm -f "$pid_file"
   fi
+
```
laurynas-biveinis: Whitespace-only change

laurynas-biveinis (changes requested):
Same as for 5.5: LGTM with whitespace-only change removed

PR #1084, opened 2016-09-28
Bug#17922198 REMOVE OBSOLETE IFDEF HAVE_PURIFY CODE.

laurynas-biveinis:
Patch #5
Remove special purify implementations of the uint3korr macro.

http://jenkins.percona.com/view/PS%205.5/job/percona-server-5.5-asan-param/64/


percona-ysorokin (approved):
LGTM

PR #1086, opened 2016-09-28
Work towards diagnosing bug 1628431 (Test main.percona_log_slow_admin…

laurynas-biveinis:
…_statements is unstable)

Add expected match counts to the testcase, so that more info is dumped
on a failure.

http://jenkins.percona.com/job/percona-server-5.6-param/1426/


percona-ysorokin (approved):
LGTM

PR #1090, opened 2016-10-03
Fix bug 1617586 (SELECT DISTINCT x...ORDER BY y LIMIT N,N crashes ser…

laurynas-biveinis:
…ver)

Make the Percona Server / TokuDB-spefific test_if_skip_sort_order patch
honor the no_changes flag value if it desires to change the QEP.

http://jenkins.percona.com/job/mysql-5.7-param/456/

The testcase is 15K rows of customer data. If we succeed in reproducing the testcase with synthetic data, it will be added in a separate commit


percona-ysorokin (approved):
LGTM

PR #1091, opened 2016-10-03
DB-1016 : table_statistics and index_statistics are not updated corre…

george-lorch:
…ctly
- Added userstat counter increments in TokuDB codebase
- Split percona_userstat.test into engine agnostic test and engine specific tests, specifically, split out tests for table_statistics and index_statistics into .inc file.
- Created two new engine specific tests, one for innodb and one for tokudb that source the new percona_table_index_statistics test/


george-lorch:
jenkins here http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1431/

@laurynas-biveinis, please review percona_userstat test changes to see if we are thinking alike or if you prefer a different approach. I'm holding off the merge to 5.7 until we have decided on the best test case layout.

mysql-test/include/percona_table_index_statistics.inc
```diff
@@ -0,0 +1,63 @@
+# include file to test index and table statstics for specific storage engine
+# requires includer set the default strorage engine for the session
+
+# Bug 602047 (wrong rows_read value)
+
+--disable_warnings
+DROP TABLE IF EXISTS t1;
```
laurynas-biveinis: This idiom is redundant

storage/tokudb/ha_tokudb.cc
```diff
@@ -4109,10 +4109,13 @@ int ha_tokudb::write_row(uchar * record) {
     }
 
     trx = (tokudb_trx_data *) thd_get_ha_data(thd, tokudb_hton);
-    if (!error) {
+    if (TOKUDB_LIKELY(!error)) {
         added_rows++;
         trx->stmt_progress.inserted++;
         track_progress(thd);
+#ifdef EXTENDED_FOR_USERSTAT
```
laurynas-biveinis: Please don't add and remove EXTENDED_FOR_USERSTAT macro when you touch it

storage/tokudb/ha_tokudb.cc
```diff
@@ -4109,10 +4109,13 @@ int ha_tokudb::write_row(uchar * record) {
     }
 
     trx = (tokudb_trx_data *) thd_get_ha_data(thd, tokudb_hton);
-    if (!error) {
+    if (TOKUDB_LIKELY(!error)) {
         added_rows++;
```
laurynas-biveinis, line 4113: Consider pulling this up to handler level

laurynas-biveinis (changes requested):

george-lorch:
new jenkins after pulling all userstat into base handler http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1446/
*edit : changed jenkins job#, previous job failed with github issues.
Second job failing with same github issues, assuming some issue with runners.

sql/handler.cc
```diff
@@ -3001,6 +3013,15 @@ int handler::ha_index_read_map(uchar *buf, const uchar *key,
 
   MYSQL_TABLE_IO_WAIT(m_psi, PSI_TABLE_FETCH_ROW, active_index, 0,
     { result= index_read_map(buf, key, keypart_map, find_flag); })
+
+  if (!result)
```
laurynas-biveinis: Extract a helper method?

sql/handler.cc
```diff
@@ -2936,6 +2936,12 @@ int handler::ha_rnd_next(uchar *buf)
 
   MYSQL_TABLE_IO_WAIT(m_psi, PSI_TABLE_FETCH_ROW, MAX_KEY, 0,
     { result= rnd_next(buf); })
+
+  if (!result)
```
laurynas-biveinis: IIRC, at the InnoDB level UNIV_UNLIKELY has been added to the corresponding snippet after branch misprediction profiling, i.e. it's useful. Maybe here we should do `if (likely(!result))` as well

laurynas-biveinis (changes requested):
I've been also thinking about a Percona "engines" suite for XtraDB/Toku/Rocks, as you had suggested before. Userstat and kill idle transaction feature I'm working on are good candidates for this. But I'm not sure how to approach this, as the current engines suites need big --mysqld= incantation to work, and only test one engine. Maybe we need an internal-to-MTR loop over engines for selected tests, similar to how binlog tests loop over binlog formats.

sql/handler.cc
george-lorch: Good call.

sql/handler.cc
george-lorch: I considered one but wasn't sure if it was worth more pollution of the handler class. The Maria implementation uses one. Since you suggest it I assume that pollution isn't too big of an issue, I'll refactor and push.

george-lorch:
New branch pushed with branch prediction macros and factored out inline function to tickle counters.
jenkins at http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1449/

sql/handler.cc
```diff
@@ -7743,6 +7814,9 @@ int handler::ha_write_row(uchar *buf)
   if (unlikely(error= binlog_log_row(table, 0, buf, log_func)))
     DBUG_RETURN(error); /* purecov: inspected */
 
+  if (!is_fake_change_enabled(ha_thd()))
```
laurynas-biveinis: likely

sql/handler.cc
```diff
@@ -7773,6 +7847,10 @@ int handler::ha_update_row(const uchar *old_data, uchar *new_data)
     return error;
   if (unlikely(error= binlog_log_row(table, old_data, new_data, log_func)))
     return error;
+
+  if (!is_fake_change_enabled(ha_thd()))
```
laurynas-biveinis: likely

sql/handler.cc
```diff
@@ -7801,6 +7879,10 @@ int handler::ha_delete_row(const uchar *buf)
     return error;
   if (unlikely(error= binlog_log_row(table, buf, 0, log_func)))
     return error;
+
+  if (!is_fake_change_enabled(ha_thd()))
```
laurynas-biveinis: likely

laurynas-biveinis (changes requested):
Approved with the three "likely" added

sql/handler.cc
george-lorch: Thanks, I considered that but then if someone is using fake changes (unlikely, heh), they will have 100% branch prediction miss so didn't think it was a good case for it.

laurynas-biveinis: The use case is I/O bound, and we have to consider the needs of 99.95% users who don't use it :)

george-lorch: Fair enough

laurynas-biveinis (approved):

PR #1096, opened 2016-10-05
Fix bug 1629879 (Server won't start with innodb_force_recovery=6 beca…

laurynas-biveinis:
…use parallel doublewrite file exists)

With --innodb_force_recovery=6, the doublewrite buffer recovery is
skipped. But the instance may still write to the doublewrite buffer,
potentially overwriting the uncovered pages. With the parallel
doublewrite, the server already refuses to startup with
--innodb_force_recovery=6 if the doublewrite buffer file is present,
preventing the overwite of the uncovered pages. However the error
message is not intuitive, thus improve it.


percona-ysorokin (approved):
LGTM

PR #1097, opened 2016-10-10
Bug 1626519: Valgrind: Invalid read of size 8 at `get_record_buffer' …

gl-sergei:
…by `audit_log_notify'

There are two issues:
- SET STATEMENT ... FOR ... does not copy MEMALLOC'ed variables
  correctly
- when MEMALLOC'ed variables are copied by SET STATEMENT ... FOR ... ,
  audit log still needs to access original session values

Fix for SET STATEMENT:
- keep track of memalloc'ed variables (not just values as
  `dynamic_variables_allocs' does). Relevant code backported from the
  upstream commit 4d275c8995. 5.7 has this code in place.
- in `copy_system_variables' replace references to memmalloc'ed
  variables with newly references to strdup'ed ones.

Fix for audit plugin:
- make `audit_log_local' and`audit_log_record_buffer' immune to SET
  STATEMENT ..., by saving pointers to the originally allocated buffers
  and always accessing them.


gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1434/

gl-sergei:
with valgrind http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-valgrind/198/

plugin/audit_log/audit_log.c
```diff
@@ -772,6 +772,8 @@ typedef struct
 {
   /* size of allocated large buffer to for record formatting */
   size_t record_buffer_size;
+  /* large buffer to for record formatting */
```
laurynas-biveinis: "to for"

plugin/audit_log/audit_log.c
```diff
@@ -1471,9 +1473,13 @@ static MYSQL_SYSVAR_STR(include_commands, audit_log_include_commands,
...
 
 static MYSQL_THDVAR_STR(local,
-                        PLUGIN_VAR_READONLY | PLUGIN_VAR_MEMALLOC | \
-                        PLUGIN_VAR_NOSYSVAR | PLUGIN_VAR_NOCMDOPT,
-                        "Local store.", NULL, NULL, "");
+       PLUGIN_VAR_READONLY | PLUGIN_VAR_MEMALLOC | \
+       PLUGIN_VAR_NOSYSVAR | PLUGIN_VAR_NOCMDOPT,
+       "Local store.", NULL, NULL, "");
+
+static MYSQL_THDVAR_ULONG(local_ptr,
```
laurynas-biveinis, line 1480: static_assert(sizeof(local_ptr) >= sizeof(void *)) ?

mysql-test/include/audit_log_events.inc
```diff
@@ -37,6 +37,9 @@ alter table t1 rename renamed_t1;
 select * from t_doesnt_exist;
```
laurynas-biveinis, line 37: This is audit log format-independent test, maybe put pull this up from an include to some testcase, i.e. audit_log.test

laurynas-biveinis (changes requested):

plugin/audit_log/audit_log.c
gl-sergei: since static_assert is C++11, while audit_log.c is C99, I added DBUG_ASSERT. We would probably want to convert this plugin to C++, but it is out of scope of this fix.

plugin/audit_log/audit_log.c
gl-sergei: fixed

mysql-test/include/audit_log_events.inc
gl-sergei: True. The problem is, there is no "audit_log.test", but nothing stopped me from starting one.

gl-sergei:
new build is here: http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1436/

plugin/audit_log/audit_log.c
laurynas-biveinis: MySQL has its own static_assert, I'm not sure what's the exact name, maybe compile_time_assert

laurynas-biveinis (changes requested):

sql/sql_plugin.cc
```diff
@@ -4028,45 +4047,69 @@ int unlock_plugin_data()
...
 
-  DBUG_ASSERT(src);
+  DBUG_ASSERT(thd);
 
   dst= (struct system_variables *)
     my_malloc(sizeof(struct system_variables), MYF(MY_WME | MY_FAE));
-  *dst = *src;
+  *dst = thd->variables;
 
   if (dst->dynamic_variables_ptr)
```
gl-sergei, line 4061: I am not sure if this block should be protected by `LOCK_system_variables_hash`, in `intern_sys_var_ptr` I see

```
    thd->variables.dynamic_variables_ptr= (char*)
      my_realloc(thd->variables.dynamic_variables_ptr,
                 global_variables_dynamic_size,
                 MYF(MY_WME | MY_FAE | MY_ALLOW_ZERO_PTR));
```

is done under `LOCK_system_variables_hash` and it looks logical to me. But current set-statement implementation seems doesn't bother to acquire this lock at all.

plugin/audit_log/audit_log.c
gl-sergei: indeed, it has. updated.

sql/sql_plugin.cc
laurynas-biveinis: This is a copy between current thread's THD and a local variable - system_variables_hash does not seem to be involved?

laurynas-biveinis: Please check: LOCK_system_variables_hash has to be locked, in the same critical section as the below loop, to avoid race with INSTALL PLUGIN adding new global variables

gl-sergei: I see no issues here (at least in 5.6). THD-local variables are initialized lazily when they are accessed. I don't see any way to access them from another thread.

laurynas-biveinis (approved):

gl-sergei:
updated param build looks ok: http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1437/

PR #597, opened 2016-06-05
Bugfix lp:719368 log_slow_sp_statements interferes with log_slow_admin_statements

ihanick:
Bug fix for https://bugs.launchpad.net/percona-server/+bug/719368

Backup thd->enable_slow_log value before executing statements inside stored procedures.

http://jenkins.percona.com/job/percona-server-5.5-param/1241/

PR #1099, opened 2016-10-13
Bugfix for 5.6 lp:719368 log_slow_sp_statements interferes with log_slow_admin_statements

ihanick:


sql/rpl_master.cc
```diff
@@ -743,7 +743,8 @@ bool com_binlog_dump(THD *thd, char *packet, uint packet_length)
   uint packet_bytes_todo= packet_length;
 
   status_var_increment(thd->status_var.com_other);
-  thd->enable_slow_log= opt_log_slow_admin_statements;
+  thd->enable_slow_log= opt_log_slow_admin_statements
```
laurynas-biveinis: Why is this required here?

ihanick:
#597 fix in 5.5 allows to localize changes in sql/sp_head.cc
In 5.6 and 5.7 code I can save-restore thd->enable_slow_log, but there is no simple method to fix
log_slow_sp_statements=OFF & log_slow_admin_statements=ON case, because queries are logged outside sp_head.cc.

I have considered to use:

```
thd->enable_slow_log= opt_log_slow_admin_statements
         && (thd->sp_runtime_ctx ? opt_log_slow_sp_statements : 1);
```

in all places in code modifying enable_slow log like:
`thd->enable_slow_log= opt_log_slow_admin_statements;`

Thus I'm using thd->sp_runtime_ctx expression to detect if admin statement executed inside stored procedure.

To my mind the code will be clear if both enable_slow_log change and restore will happen at the same place, but several function executing admin statements have several return points and it's still easier to save and restore thd->enable_slow_log in sp_head.cc.

sql/rpl_master.cc
```diff
@@ -793,7 +794,8 @@ bool com_binlog_dump_gtid(THD *thd, char *packet, uint packet_length)
   Gtid_set slave_gtid_executed(&sid_map);
 
   status_var_increment(thd->status_var.com_other);
-  thd->enable_slow_log= opt_log_slow_admin_statements;
+  thd->enable_slow_log= opt_log_slow_admin_statements
```
laurynas-biveinis: Why is this required here?

ihanick:
http://jenkins.percona.com/job/percona-server-5.6-param/1441/

sql/sql_admin.cc
```diff
@@ -1065,7 +1065,8 @@ bool Sql_cmd_analyze_table::execute(THD *thd)
   if (check_table_access(thd, SELECT_ACL | INSERT_ACL, first_table,
                          FALSE, UINT_MAX, FALSE))
     goto error;
-  thd->enable_slow_log= opt_log_slow_admin_statements;
+  thd->enable_slow_log= opt_log_slow_admin_statements
+	  && (thd->sp_runtime_ctx ? opt_log_slow_sp_statements : 1 );
```
laurynas-biveinis: s/1/true/g

laurynas-biveinis: Can this and identical assignments below be pulled up to next to lex->m_sql_cmd->execute(thd) call in "case SQLCOM_ANALYZE etc" in mysql_execute_command?

laurynas-biveinis (changes requested):

ihanick:
- removed SP context check from binlog dump and CHECK TABLE commands (check table is not allowed in SP).
- changed one to true 

> Can this and identical assignments below be pulled up to next to lex->m_sql_cmd->execute(thd) call
> Should I use this code?
> 
> ```
>   case SQLCOM_ANALYZE:
>   case SQLCOM_CHECK:
>   case SQLCOM_OPTIMIZE:
>   case SQLCOM_REPAIR:
>   case SQLCOM_TRUNCATE:
>   case SQLCOM_ALTER_TABLE:
> +    thd->enable_slow_log= opt_log_slow_admin_statements
> +           && (thd->sp_runtime_ctx ? opt_log_slow_sp_statements : true);
>   case SQLCOM_HA_OPEN:
> ....
> res= lex->m_sql_cmd->execute(thd);
> ```

Add tests for each statement (including execution outside SP)
Remove thd->enable_slow_log=... statements from handlers

or current thd->enable_slow_log= opt_log_slow_admin_statements; statements patching strategy will be enough?

laurynas-biveinis:
The case fallthrough code is OK, just please add comments saying why it's done there and that it does fall through.

ihanick:
http://jenkins.percona.com/job/percona-server-5.6-param/1444/
- replaced thd->enable_slow_log= opt_log_slow_admin_statements with wrapper call:
  thd->set_slow_log_for_admin_command();

laurynas-biveinis (approved):

PR #1105, opened 2016-10-13
Merge webscalesql 5.6.27.75 to ps 5.6 myrocks import

george-lorch:
This is the initial import of the MyRocks storage engine into Percona 5.6. Most of the commits are part of the MyRocks history, only the last commits in the list from me are actually our work.

Six of those commits are from the actual import process:
5bb178e Merge branch 'mysql-test/suite/rocksdb' of ../upstream-working
7513906 Merge branch 'mysql-test/suite/rocksdb.rpl' of ../upstream-working
657e00d Merge branch 'mysql-test/suite/rocksdb.stress' of ../upstream-working
5140fe3 Merge branch 'mysql-test/suite/rocksdb.sys_vars' of ../upstream-working
3242e89 Merge branch 'master' of ../staging into merge_webscalesql-5.6.27.75_to_ps-5.6-myrocks-import
6a33ad9 Update of storage/myrocks/rocksdb submodule commit pointer to 6a17b07ca856e573fabd6345d70787d4e481f57b

The remaining two commits for MYR-2 and MYR-24 are for the cmake changes and starting to set up the mtr trees for oncoming work.

Here is a jenkins run _without_ the -DWITH_MYROCKS=1 specification to prove that the tree builds cleanly and quietly without myrocks http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1443/

I also have a 5.7 GCA branch in the exact same state ready for a pull request but will hold off until we agree on the state of this branch.


storage/rocksdb/CMakeLists.txt
```diff
@@ -1,27 +1,46 @@
...
-  ${CMAKE_SOURCE_DIR}/rocksdb/include
-  ${CMAKE_SOURCE_DIR}/rocksdb/third-party/gtest-1.7.0/fused-src
+  ${CMAKE_CURRENT_SOURCE_DIR}/rocksdb
+  ${CMAKE_CURRENT_SOURCE_DIR}/rocksdb/include
+  ${CMAKE_CURRENT_SOURCE_DIR}/rocksdb/third-party/gtest-1.7.0/fused-src
 )
 
 ADD_DEFINITIONS(-DROCKSDB_PLATFORM_POSIX -DROCKSDB_LIB_IO_POSIX -DOS_LINUX
-  -DZLIB)
+  -DZLIB -DPERCONA_SERVER)
```
laurynas-biveinis, line 43: Why?

storage/rocksdb/CMakeLists.txt
```diff
@@ -54,23 +73,23 @@ IF (NOT "$ENV{WITH_SNAPPY}" STREQUAL "")
   $ENV{WITH_SNAPPY}/lib/libsnappy${PIC_EXT}.a)
   ADD_DEFINITIONS(-DSNAPPY)
 ELSE()
-  SET(rocksdb_static_libs ${rocksdb_static_libs} snappy)
+#  SET(rocksdb_static_libs ${rocksdb_static_libs} snappy)
```
laurynas-biveinis: Please remove, do not comment out.
Why this needs to be removed in the first place?

storage/rocksdb/CMakeLists.txt
```diff
@@ -82,14 +101,10 @@ ENDIF()
 
 SET(rocksdb_static_libs ${rocksdb_static_libs} ${ZLIB_LIBRARY} "-lrt")
 
-MYSQL_ADD_PLUGIN(rocksdb_se ${ROCKSDB_SOURCES} STORAGE_ENGINE DEFAULT STATIC_ONLY
+MYSQL_ADD_PLUGIN(rocksdb ${ROCKSDB_SOURCES} STORAGE_ENGINE DEFAULT MODULE_ONLY
   LINK_LIBRARIES ${rocksdb_static_libs}
 )
 
-IF(WITH_EMBEDDED_SERVER)
```
laurynas-biveinis: Why?

mysql-test/include/have_rocksdb_as_default.inc
```diff
@@ -0,0 +1,18 @@
+--source include/have_rocksdb.inc
+
+if (`SELECT COUNT(*) = 0 FROM INFORMATION_SCHEMA.ENGINES WHERE engine = 'rocksdb' AND support IN ('DEFAULT')`)
+{
+  --skip Test requires default engine RocksDB
+}
+
+# MYR-14
```
laurynas-biveinis: I'd avoid markers which would not be understood by an external reader. It has to be explained whatever "MYR-14" is explaining.

mysql-test/include/have_rocksdb_as_default.inc
```diff
@@ -0,0 +1,18 @@
...
+#if (`select count(*) = 0 from information_schema.tables where engine='rocksdb' and table_name='slave_gtid_info'`)
```
laurynas-biveinis: Is this check required? We'd support a mixed InnoDB + RocksDB setup, so probably we'd use the InnoDB crash-safe slave?

mysql-test/lib/mtr_cases.pm
```diff
@@ -639,7 +639,7 @@ sub optimize_cases {
     # Check that engine selected by
     # --default-storage-engine=<engine> is supported
     # =======================================================
-    my %builtin_engines = ('myisam' => 1, 'memory' => 1, 'csv' => 1);
+    my %builtin_engines = ('myisam' => 1, 'memory' => 1, 'csv' => 1, 'rocksdb' => 1);
```
laurynas-biveinis: So the proper fix would be to have all plugins loaded at mysql-test-run.pl:collect_mysqld_features. I have tried to implement exactly this for TokuDB with a fix for http://bugs.mysql.com/bug.php?id=72615 (https://bugs.launchpad.net/percona-server/+bug/1318537, and then https://bugs.launchpad.net/percona-server/+bug/1387227).

What does currently happen with RocksDB inside of collect_mysqld_features?

laurynas-biveinis (changes requested):

storage/rocksdb/CMakeLists.txt
george-lorch: So I can #ifdef out sections/features that cause compilation problems until all issues are addressed. It allows incremental testing of completed fixes within the SE and also gives me clear markers within the codebase of where we differ from upstream implementation.

george-lorch: See something like https://github.com/georgelorchpercona/percona-server/commit/caedfc2d63b38efb0656a1210d7a8ad47fb24a51#diff-04bd0348da849eb9f32addccdffa7aa5R105 as an example of how I am using it.

storage/rocksdb/CMakeLists.txt
george-lorch: I  couldn't immediately remember but now I do. This is trying to use snappy as a system lib and would require that all builders have it, which I do not believe is practical right now and will need to be revisited in order for us to support snappy. I believe we will need to bundle it as we did for TokuDB unless we want to make it a true dependency and get jenkins builders updated. This was left over from my initial attempt several months ago and probably could/should be made clearer in the code and a JIRA issue attached.

storage/rocksdb/CMakeLists.txt
george-lorch: Stolen from TokuDB to be somewhat consistent since we don't build nor support Percona Server as embedded and will not likely ever support the same with RocksDB

mysql-test/include/have_rocksdb_as_default.inc
george-lorch: I specifically want these so I can refer back to the JIRA issue where I am tracking all commentary and can do an easy search through the base when I come back to an issue to address it and eventually remove them entirely. I suppose I can add the issue description to the tags to make it more clear. These are not intended to be permanent, in fact, the presence of any MYR tag/ifdef is indicative of something that must be addressed before it is released.

mysql-test/include/have_rocksdb_as_default.inc
george-lorch: This will/may be required in the future as we figure out how to address MYR-14 and will cause all mtr tests to be skipped at this point until it the issue is figured.

mysql-test/lib/mtr_cases.pm
george-lorch: I do not think those changes have any impact on this issue. All rocks db mtr tests are written with the basic assumption that rocksdb is the default storage engine. In facebooksql, it is compiled in and configured as such, not so in Percona Server. So this proposed change, running mtr with `--mysqld=--default-storage-engine=rocksdb --mysqld=--default-tmp-storage-engine=rocksdb` (or any other non-builtin engine) would result in all tests skipping with `[ skipped ]  'rocksdb' not supported`. Similarly, doing the same thing with TokuDB results in the same issue. I would certainly like a better solution than this but other than re-writing all of the tests to be more explicit in specifying the default engine or using `... ENGINE=ROCKSDB` I can't see any other way to fix it.

mysql-test/include/have_rocksdb_as_default.inc
george-lorch: *\* Clarifying - facebook/webscale sql introduces a new system table slave_gtid_info that is part of their cash safe slave implementation. Since this table does not exist in Percona Server and we haven't made it to the point to decide what to do about MyRocks and crash safe slave, leaving this check enabled will cause all myrocks tests to be skipped. Commenting and tagging it with MYR-14 means that this code will be re-evaluated and corrected when we address MYR-14, crash safe slave and myocks.

mysql-test/include/have_rocksdb_as_default.inc
laurynas-biveinis: OK. Would be good to track somewhere the smaller items that need to be resolved on the experimental -> RC transition for MYR markers and the like

mysql-test/include/have_rocksdb_as_default.inc
laurynas-biveinis: OK

mysql-test/lib/mtr_cases.pm
laurynas-biveinis: I think this is indeed the collect_mysqld_features issue, not necessarily the test issue itself. MTR runs collect_mysqld_features, does not see MyRocks there, does not set rocksdb as available, and as a result `--mysqld=--default-storage-engine=rocksdb` considers rocksdb as unavailable. With the fixes above `--mysqld=--default-storage-engine=tokudb --mysqld=--plugin-load=ha_tokudb.so=blahblahblah` works some of the time for some configurations, but it's not good enough.

storage/rocksdb/CMakeLists.txt
laurynas-biveinis: Is the current plan that -DPERCONA_SERVER will stay post-experimental stage?

storage/rocksdb/CMakeLists.txt
laurynas-biveinis: OK, let's track it.

storage/rocksdb/CMakeLists.txt
laurynas-biveinis: Is this proactive disable or is MyRocks failing already on embedded? Medium-term, we'd like to support PS (with some subset of Rocks, Toku) as embedded server, if only to match WebScaleSQL testing.

storage/rocksdb/CMakeLists.txt
george-lorch: No, and along with the MYR issues, it is my 'goal line' to define when we have successfully addressed all import related issues. Some might be to remove the offending feature/code, some will be to pull some portion of FacebookSQL, and some will be to replace the FacebookSQL implementation with the Percona Server implementation. Once this builds and passes MTR cleanly without the -DPERCONA_SERVER and without a single `#ifdef PERCONA_SERVER //MYR???` we will be ready to move into the next stage of release availability. Ideally, I want to have most of these fixed in 5.6 providing the 5.7 counterpart implementation is similar enough to not be fixing them twice. If any issue is significantly different between 5.6 and 5.7, then it will remain unfixed in 5.6 and get addressed immediately after the 5.7 port. Without this macro and the #ifdefs, there is no way to incrementally fix build and test each item.

storage/rocksdb/CMakeLists.txt
george-lorch: Proactive, I can't recall ever even trying to build an embedded Percona Server as it was always my understanding that it just didn't work. I can pull it if you feel strongly about it but embedded is something that is just not at all on my radar here.

mysql-test/include/have_rocksdb_as_default.inc
george-lorch: As mentioned above, each of these specifically refers to a JIRA issue that needs completing before we can consider the import complete. They exist at this point solely to identify where/what needs to be done and to allow a successful compilation with reduced feature set in order to be able to begin incremental testing as things progress.

mysql-test/include/have_rocksdb_as_default.inc
george-lorch: Another thing I was curious about on this particular validation is why they check for this table here as a condition to meet for 'have rocksdb as default engine'. This table is specific to a set of features surrounding the FacebookSQL crash safe replication. This table is not necessary for all tests and is only an indicator that the FacebookSQL system schema is in place. IMHO the check for this table should be its own test or include prerequisite for something like a `have_facebooksql_crash_safe_rpl.inc`. This is really a discussion to have on MYR-14 though and not here since we will very likely not be using their implementation of crash safe rpl and probably just stick with what we have in PS.

storage/rocksdb/CMakeLists.txt
laurynas-biveinis: OK

storage/rocksdb/CMakeLists.txt
laurynas-biveinis: Trying to build an embedded PS actually works in 5.6 and 5.7 - do you always build with -DWITH_EMBEDDED_SERVER=OFF? If yes, then just remove that option.

mysql-test/include/have_rocksdb_as_default.inc
laurynas-biveinis: `have_facebooksql_crash_safe_rpl.inc` is something we should suggest to upstream eventually, it makes sense.

storage/rocksdb/CMakeLists.txt
george-lorch: Yes, I do always build -DWITH_EMBEDDED_SERVER=OFF. Old habit that I never changed since IIRC we didn't ever support embedded server and at one time it would not even compile (maybe some 5.1 days).

mysql-test/lib/mtr_cases.pm
george-lorch: Yup, I should listen to you more often ;) It does work correctly without the nasty change above in an out of source build when I use the correct mtr incantation of `--mysqld=--default-storage-engine=rocksdb --mysqld=--default-tmp-storage-engine=rocksdb --mysqld=--plugin-load="rocksdb=ha_rocksdb.so;rocksdb_cfstats=ha_rocksdb.so;rocksdb_dbstats=ha_rocksdb.so;rocksdb_perf_context=ha_rocksdb.so;rocksdb_perf_context_global=ha_rocksdb.so;rocksdb_cf_options=ha_rocksdb.so;rocksdb_global_info=ha_rocksdb.so;rocksdb_ddl=ha_rocksdb.so;rocksdb_index_file_map=ha_rocksdb.so"`

george-lorch:
OK, pushed a new branch with some of the touchups mentioned and added in MYR-28 for compression libs attempting to build with system installed versions by default.

mysql-test/include/have_rocksdb_as_default.inc
```diff
@@ -0,0 +1,12 @@
+--source include/have_rocksdb.inc
```
laurynas-biveinis, line 1: Maybe push --default-storage-engine=rocksdb to suite.opt too? and then have_rocksdb_as_default.inc becomes redundant? But it needs to be checked what --d-s-e does for an unavailable SE, so that we don't silently test InnoDB, and I'm not sure what to do about `set global rocksdb_force_flush_memtable_now = true;`

laurynas-biveinis (changes requested):

mysql-test/include/have_rocksdb_as_default.inc
george-lorch: Hmm, yes. That's actually how I started this but since I was missing the --plugin-load on the mtr call it was skipped/ignored during the collect as we just got through figuring. So assuming it works correctly, it would eliminate the need for the extra mtr options --d-s-e and --d-t-s-e.

From my earlier testing, the proposed have_rocksdb.inc would 'pass' and run the test with a different default engine. This is specifically why I introtuced the have_rocksdb_as_default.inc to ensure that the tests are actually being run as intended. I think I would like to keep that as is since I imagine we will have a transition period where some tests will be written correctly to explicitly use a storage engine, only needing have_rocksdb.inc, and these current tests that depend on rocksdb being default going into the test.

IMHO the net change here should be just to add --d-s-e and --d-t-s-e to the suite.opt and keep the use of have_rocksdb.inc and have_rocksdb_as_default.inc as is. With that I believe we will be guaranteed that if rocksdb failed to load for any reasons, any new explicit test will fail on trying to do something like CREATE TABLE ... ENGINE=ROCKSDB and the current assuming tests will fail/skip on have_rocksdb_as_default.inc.

Maybe, if we wanted to assert it a bit more we could change have_rocksdb_as_default.inc --error rather than --skip on `SELECT COUNT(*) = 0 FROM INFORMATION_SCHEMA.ENGINES WHERE engine = 'rocksdb' AND support IN ('DEFAULT')`.

As far as `set global rocksdb_force_flush_memtable_now = true`, IMHO it is another bad move by the test writers and should be placed within specific tests that rely on deterministic behavior of table statistics, not in a global catch-all. I will create a new JIRA task for this and see if I can figure out how many tests would need changing and possibly suggest it upstream. For now, I will just add JIRA tags to this with the new issue id until I can figure out how to sort it through the individual test cases.

george-lorch: So testing this out and it doesn't work with the suite.opt as expected. suite.opt is `$ROCKSDB_OPT $ROCKSDB_LOAD_ADD --default-storage-engine=rocksdb --default-tmp-storage-engine=rocksdb`

and mtr incantation is `./mtr  --force --retry=0 --retry-failure=0 --max-test-fail=0 --no-warnings --testcase-timeout=600 --mysqld=--plugin-load="rocksdb=ha_rocksdb.so;rocksdb_cfstats=ha_rocksdb.so;rocksdb_dbstats=ha_rocksdb.so;rocksdb_perf_context=ha_rocksdb.so;rocksdb_perf_context_global=ha_rocksdb.so;rocksdb_cf_options=ha_rocksdb.so;rocksdb_global_info=ha_rocksdb.so;rocksdb_ddl=ha_rocksdb.so;rocksdb_index_file_map=ha_rocksdb.so" --suite=rocksdb`

Adding the `--mysqld=--default-storage-engine=rocksdb --mysqld=--default-tmp-storage-engine=rocksdb` back in works. The logs indicate it is the have_rocksdb_as_default.inc that is causing the skip. So somehow the --d-s-e and --d-t-s-e suite options are not getting passed into the server instance.

From a little poking and debugging, it looks like in  $default_storage_engine is set if specified on the command line, but it is not set if it is in suite.opt, this causes the logic in collect_one_test_case to take a different path and I believe the end result is that MyISAM becomes the default engine. So it seems that options in the suite.opts and test.opts are not run through the collect_options function which sets the global option variables. That makes sense because we wouldn't want a single test or suite option to change the global for other suites. So to fix this I think there would need to be a test by test check in the options for things like --d-s-e and perhaps others.

Honestly, at this point I'm not sure the depth of changes needed to fix this are worth the effort and risk of breaking mtr. Can you confirm my research here and advise if we should just leave as is and require --d-s-e and --d-t-s-e along with the --mysqld=--plugin-load="..." incantation in order to run rocksdb tests for now and perhaps revisit this at some later time if it becomes more critical? Perhaps we can convince upstream to fix their test cases to not pre-require rocksdb as the d-s-e which would solve it all for us anyway.

george-lorch:
Pushed a new branch that removes the `set global rocksdb_force_flush_memtable_now = true` and have a JIRA issue to deal with fixing up any fallout in mtr instability from it https://jira.percona.com/browse/MYR-29

mysql-test/include/have_rocksdb_as_default.inc
laurynas-biveinis: Yes, everything you suggest sounds good. 

BTW, if you are patching the testcases for s/have_rocksdb/have_rocksdb_as_default/g, then patching for SET default_storage_engine=... is similar amount of patching, isn't it :) I'm not saying to make this change, it's just something that could be considered in the future

laurynas-biveinis: Ah, have_rocksdb_as_default.inc can do SET default_storage_engine = ... itself, can't it?

laurynas-biveinis: No it cannot, this will "fail" instead of "skip" on -DWITHOUT_ROCKSDB=ON - which is also a reason we cannot "assert a bit more" by failing instead of skipping when RocksDB is not default, but its suites will go to the default suite list

laurynas-biveinis (approved):
Since this puts new things into trunk, please file a very short blueprint "Experimental MyRocks 5.6 merge"

mysql-test/include/have_rocksdb_as_default.inc
george-lorch: Correct, and in order for it to function properly for multi-session tests, it would have to set global, which then also means we would need a 'clean up' step at the end of each test to reset it back to previous state.

george-lorch: Also, changing `--source include/have_rocksdb.inc` to `--source include/have_rocksdb_as_default.inc` was just a simple sed. Adding the whole `get current d-e-s; set d-s-e=rocksdb; <test body>; set d-s-e=old-d-s-e` is quite a bit more interesting to ensure it is correct on all tests.

george-lorch:
> Since this puts new things into trunk, please file a very short blueprint "Experimental MyRocks 5.6 merge"

We should talk about this. When you say trunk, do you mean actually the trunk or are you simple referring to the fact that it changes server side code? We do not do this when we add new things to TokuDB.

If you do mean trunk in general, are you expecting now a new blueprint for every MP on this? If so, why, again we do not do this for TokuDB, we simply rely on the JIRA project and Hrvoje and release team know to look there for anything to report, we just would need to make them aware of this.

If we do want a blueprint for this, then I suggest maybe we use the one already created and just add commentary to it with back link to each Pull Request and possibly the MYR issues addressed wihin each Pull Request.

We should get this figured out right now because once this particular PR gets merged, there is a flood of changes coming right behind it that depend on it.

george-lorch:
https://jira.percona.com/browse/MYR-2
https://jira.percona.com/browse/MYR-24
https://jira.percona.com/browse/MYR-28
https://blueprints.launchpad.net/percona-server/+spec/experimental-myrocks-5.6
https://blueprints.launchpad.net/percona-server/+spec/integrate-myrocks-mtr-5.6
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1451/

PR #1106, opened 2016-10-14
Fix bug 1612076 (Test rpl.rpl_cant_read_event_incident is unstable)

laurynas-biveinis:
The testcase shuts down the master, replaces its binlog file contents,
restarts the master, and resets/the connects the slave. This has a
race condition, that the slave may reconnect to the master after its
restart, and replay the CREATE TABLE statement before the slave
reset/restart. This will then cause SQL thread error as the same
statement will be replayed the 2nd time.

Fix by postponing slave connect to the master until the master has the
test binlog.

http://jenkins.percona.com/job/percona-server-5.5-param/1402/

Committed on the tip as the GCA does not have the latest version of the testcase.


percona-ysorokin (approved):
LGTM

PR #1108, opened 2016-10-19
Deprecate scalability metrics plugin

laurynas-biveinis:
This implements
https://blueprints.launchpad.net/percona-server/+spec/deprecate-scalability-metrics
blueprint.

Disable scalability metrics plugin at CMake level unless
-DWITH_SCALABILITY_METRICS=ON is given.

Fix diagnostics at
mysql-test/include/have_scalability_metrics_plugin.inc so that the
skipped test is not mistakenly diagnosed as belonging to Example
plugin.

http://jenkins.percona.com/job/percona-server-5.5-param/1410/


percona-ysorokin (approved):
LGTM

PR #1113, opened 2016-10-20
Fix bug 1633430 (Test main.variables-notembedded is unstable)

laurynas-biveinis:
If server send ER_NET_PACKET_TOO_LARGE, it closes the connection
immediatelly afterwards, without waiting for the client. This may
cause client to get one of the "server lost" errors instead of the
"packet too large" one. Fix by expecting the "server lost" errors in the
testcase too.

http://jenkins.percona.com/job/percona-server-5.5-param/1416/


percona-ysorokin (approved):
LGTM

PR #1120, opened 2016-10-21
MYR-5 : myrocks uses interesting webscale abort_with_stack_traces

george-lorch:
- Imported upstream my_pstack and macros

### DIFF
--- a/storage/rocksdb/rdb_utils.h
+++ b/storage/rocksdb/rdb_utils.h
@@ -71,6 +71,9 @@ namespace myrocks {
 
   Use the power of SHIP_ASSERT() wisely.
 */
+#ifndef abort_with_stack_traces
+#define abort_with_stack_traces abort
+#endif
 
 #ifndef SHIP_ASSERT
 #define SHIP_ASSERT(expr)                                               \


george-lorch:
https://blueprints.launchpad.net/percona-server/+spec/myrocks-my-pstack-5.6
https://jira.percona.com/browse/MYR-5
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1453/

laurynas-biveinis:
The blueprint is a duplicate of https://blueprints.launchpad.net/percona-server/+spec/print-all-thread-stacktraces, please merge

include/my_stacktrace.h
```diff
@@ -59,6 +59,24 @@ void my_set_exception_pointers(EXCEPTION_POINTERS *ep);
 void my_write_core(int sig);
 #endif
 
+/* print stack traces for all threads */
+void my_pstack();
+
+/* print stack traces for all threads before assert failures */
+#define assert_with_stack_traces(expr) \
```
laurynas-biveinis: It looks we'd end up with two sets of asserts/aborts - with stacktraces and without. Is there any reason not to use the stacktrace version uniformly? One thing I can think of is that my_pstack requires some stack space, and we might be aborting because stack is full. This can be checked though.

george-lorch: True. MyRocks uses the abort_with_stack_traces for a SHIP_ASSERT or an assertion that is compiled into release builds. Presumably there not a concern or an issue with stack space for this use case. i.e. I don't believe there is any use of this assertion if we are at or nearing end of stack. Other than that one specific issue, I do not see why we could not use this for all assertions.

george-lorch: So I wonder then what is next. Do we want to merge this as-is and continue to work towards digging out all assert wrappers from different imported code bases (core my_*, InnoDB, TokuDB, ???) or do we want to hold and make this all one larger task?

laurynas-biveinis: I believe this should be merged in the final form, i.e. default assert implementation changed and verified that MyRocks / not-MyRocks work with it. This is probably small amount of code changing, but more effort in testing. If that's one larger task better done later then I guess MyRocks can be patched to default assert for now.

george-lorch: That's where I was heading, I don't want to mix some large, server-wide change with something that is a necessity to get MyRocks to compile if I can help it. I don't think changing the default global behavior is quite as easy as it seems as every system/engine has its own wrapper for dealing with debug vs release/ship assertions. Then we have the submodule kv libraries that are still behaving independently (PerconaFT, RocksDB) and don't use anything from the upper layers of the server.

george-lorch:
OK, backed out import of my_pstack for this particular issue and redefined abort_with_stack_traces as abort in rdb_utils.h since it is used both as part of SHIP_ASSERT macro and directly from several places within the code. Updated blueprints.

laurynas-biveinis (approved):

PR #1122, opened 2016-10-22
MYR-8 : myrocks implements empty handler::delete_all_rows

george-lorch:
- myrocks implements delete_all_rows that does exactly what the default implementation does.
  In FacebookSQL though, there is a slightly different function signature which causes a compiler warning: virtual int handler::delete_all_rows() was hidden
- Removed unnecessary and empty implementation of delete_all_rows


george-lorch:
https://jira.percona.com/browse/MYR-8

laurynas-biveinis (approved):
It looks like this one should be sent upstream

PR #1124, opened 2016-10-22
MYR-11 : rdb_threads.h needs to include <atomic>

george-lorch:
- Added #include <atomic>


george-lorch:
https://jira.percona.com/browse/MYR-11

laurynas-biveinis (approved):

PR #1126, opened 2016-10-22
MYR-7 : myrocks needs extra initialization step during .frm opening

george-lorch:
- Added virtual handler call init_with_fields, call from open_binary_frm and pass through partition handler as required by MyRocks and implemented in webscale.

### DIFF
--- a/sql/ha_partition.h
+++ b/sql/ha_partition.h
@@ -293,6 +293,9 @@ public:
                  ha_partition *clone_arg,
                  MEM_ROOT *clone_mem_root_arg);
    ~ha_partition();
+
+   bool init_with_fields();
+
   /*
     A partition handler has no characteristics in itself. It only inherits
     those from the underlying handlers. Here we set-up those constants to
--- a/sql/handler.h
+++ b/sql/handler.h
@@ -2021,6 +2021,15 @@ public:
   {
     cached_table_flags= table_flags();
   }
+  /**
+    For MyRocks, secondary initialization that happens after frm is parsed into
+    field information from within open_binary_frm. MyRocks uses this secondary
+    init phase to analyze the key and field definitions to determine if it can
+    expose the HA_PRIMARY_KEY_IN_READ_INDEX flag on the table as it only
+    supports that behavior for certain types of key combinations.
+    Return values: false success, true failure.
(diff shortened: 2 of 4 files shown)


george-lorch:
https://blueprints.launchpad.net/percona-server/+spec/myrocks-extra-frm-init-5.6
https://jira.percona.com/browse/MYR-7
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1454/

laurynas-biveinis:
It is not entirely clear why this needs to be done - MYR and blueprints say that it "needs", and that's all. A bit of suggestion is in the new field name and in the final diff comment re. HA_PRIMARY_KEY_IN_READ_INDEX - can you please elaborate?

george-lorch:
The patch originated from Sergei Petrunia here https://github.com/facebook/mysql-5.6/commit/c1f1f0b64eaecd132fc6ec21663faf0510afacaf : Index-only scans do not work for partitioned tables and extended keys

I have added a bit more description in the MYR and blueprint taken from Sergeis original patch.

sql/ha_partition.h
```diff
@@ -293,6 +293,14 @@ class ha_partition :public handler
                  ha_partition *clone_arg,
                  MEM_ROOT *clone_mem_root_arg);
    ~ha_partition();
+
+   /*
```
laurynas-biveinis: This comment is redundant...

sql/handler.h
```diff
@@ -2021,6 +2021,11 @@ class handler :public Sql_alloc
   {
     cached_table_flags= table_flags();
   }
+  /**
+    For MyRocks, secondary initialization that happens after frm is parsed into
```
laurynas-biveinis, line 2025: ... this one is not, and again it should mention as an example that MyRocks uses it to adjust HA_PRIMARY_KEY_IN_READ_INDEX

laurynas-biveinis: Return values: false success, true failure

laurynas-biveinis (changes requested):

george-lorch:
Done...

laurynas-biveinis (approved):

PR #1131, opened 2016-10-22
MYR-18 : myroocks uses facebook/webscale implementation of flashcache

george-lorch:
- Removed MyRocks flashcache implementation.


george-lorch:
https://jira.percona.com/browse/MYR-18

laurynas-biveinis (approved):

PR #1135, opened 2016-10-23
Fix bug 728082 (USER_STATISTICS shows incorrect value for number of concurrent connections)

ihanick:
https://bugs.launchpad.net/percona-server/+bug/728082

I have implemented counting for current connections "CONCURRENT_CONNECTIONS – The number of concurrent connections for this user." and only for userstats table.

I'm not sure if this parameter will be useful in production in opposite to max(CONCURRENT_CONNECTIONS) since last flush, because current concurrent connections parameter could be obtained from I_S.processlist.
Also not sure about diff_disconnect variable. Originally I have considered to have total number of disconnects, but bug could be resolved just with simple bool flag (maybe it should be diff_concurrent_connections).

### DIFF
--- a/sql/sql_connect.cc
+++ b/sql/sql_connect.cc
@@ -421,6 +421,7 @@ static int increment_count_by_name(const char *name, const char *role_name,
       return 1; // Out of memory
     }
   }
+  user_stats->concurrent_connections++;
   user_stats->total_connections++;
   if (thd->net.vio &&  thd->net.vio->type == VIO_TYPE_SSL)
     user_stats->total_ssl_connections++;
@@ -542,6 +543,11 @@ static void update_global_user_stats_with_user(THD* thd,
   user_stats->lost_connections+=     thd->diff_lost_connections;
   user_stats->access_denied_errors+= thd->diff_access_denied_errors;
   user_stats->empty_queries+=        thd->diff_empty_queries;
+
+  if (thd->diff_disconnects && thd->diff_denied_connections == 0) {
+    DBUG_ASSERT(user_stats->concurrent_connections > 0);
+    user_stats->concurrent_connections-=  thd->diff_disconnects;
+  }
 }
 
 static void update_global_thread_stats_with_thread(THD* thd,
@@ -1210,6 +1216,12 @@ void end_connection(THD *thd)
     of someone else.
   */
   release_user_connection(thd);
+
+  if (unlikely(opt_userstat)) {
+    thd->update_stats(false);
+    thd->diff_disconnects= 1;
+    update_global_user_stats(thd, false, time(NULL));
+  }
 
   if (thd->killed || (net->error && net->vio != 0))
   {
--- a/mysql-test/include/percona_userstat_conn_handling.inc
+++ b/mysql-test/include/percona_userstat_conn_handling.inc
@@ -1,0 +1,78 @@
+SET @userstat_old= @@userstat;
+SET GLOBAL userstat=ON;
+
+CREATE USER mysqltest_1@localhost;
+CREATE USER mysqltest_2@localhost;
+
+# Make a single connection for mysqltest_1 user
+connect (conn1,localhost,mysqltest_1,,);
+SELECT 1;
+connection default;
+
+--let $assert_text= I_S.USER_STATISTICS CONCURRENT_CONNECTIONS: single connection
+--let $query_result= query_get_value(SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1', CONCURRENT_CONNECTIONS, 1)
+--let $assert_cond= "$query_result" = 1
+--source include/assert.inc
+
+# Make two connections for mysqltest_1 user
+connect (conn2,localhost,mysqltest_1,,);
+SELECT 1;
+connection default;
+
+--let $assert_text= I_S.USER_STATISTICS CONCURRENT_CONNECTIONS: two connections
+--let $query_result= query_get_value(SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1', CONCURRENT_CONNECTIONS, 1)
+--let $assert_cond= "$query_result" = 2
+--source include/assert.inc
+
+# Check if number of concurrent connections decreased
+disconnect conn2;
+let $count_sessions= 2;
+--source include/wait_until_count_sessions.inc
+
+--let $assert_text= I_S.USER_STATISTICS CONCURRENT_CONNECTIONS: handle disconnects
+--let $query_result= query_get_value(SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1', CONCURRENT_CONNECTIONS, 1)
+--let $assert_cond= "$query_result" = 1
+--source include/assert.inc
+
+# Number of concurrent connections shouldn't be affected by different users
+connect (conn3,localhost,mysqltest_2,,);
+SELECT 1;
+connection default;
+
+--let $assert_text= I_S.USER_STATISTICS CONCURRENT_CONNECTIONS: connections from other users
+--let $query_result= query_get_value(SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1', CONCURRENT_CONNECTIONS, 1)
+--let $assert_cond= "$query_result" = 1
+--source include/assert.inc
+
+
+# if there is no connections, concurrent connections counter should be zero
+disconnect conn1;
+let $count_sessions= 2;
+--source include/wait_until_count_sessions.inc
+
+--let $assert_text= I_S.USER_STATISTICS CONCURRENT_CONNECTIONS: no connections - zero concurrency
+--let $query_result= query_get_value(SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1', CONCURRENT_CONNECTIONS, 1)
+--let $assert_cond= "$query_result" = 0
+--source include/assert.inc
+
+disconnect conn3;
+let $count_sessions= 1;
+--source include/wait_until_count_sessions.inc
+
+# Unknown users are counted in USER_STATISTICS, concurrent connections should be positive
+--disable_query_log
+--error ER_ACCESS_DENIED_ERROR
+connect (conn4,localhost,mysqltest_unknownuser,,);
+--enable_query_log
+connection default;
+
+--let $assert_text= I_S.USER_STATISTICS CONCURRENT_CONNECTIONS: unknown user connection
+--let $query_result= query_get_value(SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_unknownuser', CONCURRENT_CONNECTIONS, 1)
+--let $assert_cond= "$query_result" = 0
+--source include/assert.inc
+
+
+DROP USER mysqltest_1@localhost;
+DROP USER mysqltest_2@localhost;
+
+SET GLOBAL userstat= @userstat_old;
--- a/mysql-test/r/percona_userstat_conn_handling.result
+++ b/mysql-test/r/percona_userstat_conn_handling.result
@@ -1,0 +1,23 @@
+SET @userstat_old= @@userstat;
+SET GLOBAL userstat=ON;
+CREATE USER mysqltest_1@localhost;
+CREATE USER mysqltest_2@localhost;
+SELECT 1;
+1
+1
+include/assert.inc [I_S.USER_STATISTICS CONCURRENT_CONNECTIONS: single connection]
+SELECT 1;
+1
+1
+include/assert.inc [I_S.USER_STATISTICS CONCURRENT_CONNECTIONS: two connections]
+include/assert.inc [I_S.USER_STATISTICS CONCURRENT_CONNECTIONS: handle disconnects]
+SELECT 1;
+1
+1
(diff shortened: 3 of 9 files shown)


sql/sql_connect.cc
```diff
@@ -422,6 +422,7 @@ static int increment_count_by_name(const char *name, const char *role_name,
     }
   }
   user_stats->total_connections++;
+  user_stats->concurrent_connections++;
```
laurynas-biveinis, line 424: There is an existing USER_CONN::connections - please check e.g. how https://github.com/facebook/mysql-5.6/commit/a0109fb6cce7ff7435f33b4b27c8552b191860b4 (grep for fill_one_user_stats) does it, and please confirm whether USER_CONN::connections is suitable for this purpose. This would also get rid of diff_disconnect

mysql-test/t/percona_userstat_bug728082.test
```diff
@@ -0,0 +1,43 @@
+--source include/have_innodb.inc
```
laurynas-biveinis: Please do not introduce new test file, but extend the existing one - this is a basic test for this particular column

mysql-test/t/percona_userstat_bug728082.test
```diff
@@ -0,0 +1,43 @@
...
+
+SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1';
+
+connect (conn2,localhost,mysqltest_1,,);
+connection default;
+
+--echo check inc2
+SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1';
+
+disconnect conn2;
```
laurynas-biveinis: Disconnect will complete asynchronously on the server after the network connection is closed, this is a race condition with the statement below. You can sync by Threads_connected status var, and there are several ways to do it, easiest is probably to replace the global --source include/count_sessions.inc / wait_until_count_sessions.inc (which is BTW missed here) with a few local ones as needed

mysql-test/t/percona_userstat_bug728082.test
```diff
@@ -0,0 +1,43 @@
...
+-- echo check dec1
+SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1';
+
+connect (conn3,localhost,mysqltest_2,,);
+connection default;
+
+-- echo check not incremented
+SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1';
+
+disconnect conn1;
```
laurynas-biveinis: Likewise

mysql-test/t/percona_userstat_bug728082.test
```diff
@@ -0,0 +1,43 @@
...
+
+connect (conn1,localhost,mysqltest_1,,);
+connection default;
+
+SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1';
+
+connect (conn2,localhost,mysqltest_1,,);
+connection default;
+
+--echo check inc2
```
laurynas-biveinis: Here and below please use a bit less cryptic comments

laurynas-biveinis (changes requested):

ihanick:
- Calculating concurrent connections by full hash_user_connections scan during show command
  for some reason USER_STATISTICS stores data by user, not by user@host, like USER_CONN
  this prevents both one record hash lookup in SHOW and concurrent_connections variable update for each connect/disconnect.
- Should I put whole test case at the end of percona_userstat.test? I need additional users to test connect & disconnect changes, but additional user is only used to test flush statistics.

sql/sql_connect.cc
```diff
@@ -84,6 +84,20 @@ static int increment_connection_count(THD* thd, bool use_lock);
 
 static HASH hash_user_connections;
 
+int get_user_con_by_name(const char *user) {
```
laurynas-biveinis: Please re-check the hash function for hash_user_connections whether it's really based on user@host, and not user alone: it is initialised as

```
    my_hash_init(&hash_user_connections,system_charset_info,max_connections,
                 0,0, (my_hash_get_key) get_key_conn,
                 (my_hash_free_key) free_user, 0);
```

and get_key_conn is

```
extern "C" uchar *get_key_conn(user_conn *buff, size_t *length,
                  my_bool not_used __attribute__((unused)))
{
  *length= buff->len;
  return (uchar*) buff->user;
}
```

which suggests that is hashes on the username along.

If not, then I'm afraid we cannot have O(n) here, this will hurt with high concurrency / high userstat query ratio.

mysql-test/t/percona_userstat_bug728082.test
```diff
@@ -0,0 +1,43 @@
...
+
+SELECT CONCURRENT_CONNECTIONS = 1 AS correct FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1';
+
+connect (conn2,localhost,mysqltest_1,,);
+connection default;
+
+SELECT CONCURRENT_CONNECTIONS = 2 AS correct FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1';
+
+disconnect conn2;
+let $count_sessions= 2;
```
laurynas-biveinis: Please add a comment that connections opened by this test have overlapping lifetimes, hence the direct count_sessions variable manipulation

mysql-test/t/percona_userstat_bug728082.test
```diff
@@ -0,0 +1,43 @@
...
+
+SELECT CONCURRENT_CONNECTIONS = 1 AS correct FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1';
+
+connect (conn3,localhost,mysqltest_2,,);
+connection default;
+
+SELECT CONCURRENT_CONNECTIONS = 1 AS correct FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1';
+
+disconnect conn1;
+let $count_sessions= 2;
```
laurynas-biveinis: Likewise

mysql-test/t/percona_userstat_bug728082.test
```diff
@@ -0,0 +1,43 @@
...
+
+CREATE USER mysqltest_1@localhost;
+CREATE USER mysqltest_2@localhost;
+
+SET GLOBAL userstat=ON;
+
+connect (conn1,localhost,mysqltest_1,,);
+connection default;
+
+SELECT CONCURRENT_CONNECTIONS = 1 AS correct FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1';
```
laurynas-biveinis: Looks like a job for include/assert.inc ?

laurynas-biveinis (changes requested):

laurynas-biveinis:
Yes, please add the whole test at the end of percona_userstat.test.

sql/sql_connect.cc
ihanick:
```
  /*
     Pointer to user+host key (pair separated by '\0') defining the entity
     for which resources are counted (By default it is user account thus
     priv_user/priv_host pair is used. If --old-style-user-limits option
     is enabled, resources are counted for each user+host separately).
  */

  user_len= strlen(user);
  temp_len= (strmov(strmov(temp_user, user)+1, host) - temp_user)+1;
  mysql_mutex_lock(&LOCK_user_conn);
  if (!(uc = (struct  user_conn *) my_hash_search(&hash_user_connections,
                                               (uchar*) temp_user, temp_len)))
```

userstats using:

```
typedef struct st_user_stats {
  char user[max(USERNAME_LENGTH, LIST_PROCESS_HOST_LEN) + 1];

but:
  // Update by user name
  if ((user_stats = (USER_STATS *) my_hash_search(&global_user_stats,
                                                  (uchar *) user_string,
                                                  strlen(user_string))))

Where name:
  char* user_string=         get_valid_user_string(thd->main_security_ctx.user);

static char* get_valid_user_string(char* user) {
  return user ? user : mysql_system_user;
}
```

thd->main_security_ctx.user looks to be just a username without host information and this is confusing with USER_STATS->user:   char user[max(USERNAME_LENGTH, LIST_PROCESS_HOST_LEN) + 1]; definition.

Should I return back to independent concurrent connections calculation?
or fill USER_STATS->user in the same way as user_conn hash key?

laurynas-biveinis: You are right. We should account same user from different hosts as same user, thus please go back to the independent calculation. BTW, this also looks like a bug in Facebook patch, feel free to report to them too

ihanick:
http://jenkins.percona.com/job/percona-server-5.5-param/1442/
- moved test case to percona_userstat.test
- added assertions to track possible negative values of concurrent_connections.
- added test condition for unauthorized user (causing assertion without update_global_thread_stats_with_thread: create_user variable check).
- using include/assert.inc 
- added comments for each test part

ihanick:
http://jenkins.percona.com/job/percona-server-5.5-param/1444/

Use SELECT 1; after connect

mysql-test/t/percona_userstat.test
```diff
@@ -177,4 +177,82 @@ FROM
...
 ;
 
+CREATE USER mysqltest_1@localhost;
+CREATE USER mysqltest_2@localhost;
+
+# Make a single connection for mysqltest_1 user
+connect (conn1,localhost,mysqltest_1,,);
+SELECT 1;
+connection default;
+let $count_sessions= 2;
```
laurynas-biveinis: This and next line redundant now?

mysql-test/t/percona_userstat.test
```diff
@@ -177,4 +177,82 @@ FROM
...
+--let $assert_text= I_S.USER_STATISTICS CONCURRENT_CONNECTIONS: handle disconnects
+--let $query_result= query_get_value(SELECT CONCURRENT_CONNECTIONS FROM INFORMATION_SCHEMA.USER_STATISTICS WHERE USER LIKE 'mysqltest_1', CONCURRENT_CONNECTIONS, 1)
+--let $assert_cond= "$query_result" = 1
+--source include/assert.inc
+
+# Number of concurrent connections shouldn't be affected by different users
+connect (conn3,localhost,mysqltest_2,,);
+SELECT 1;
+connection default;
+let $count_sessions= 3;
```
laurynas-biveinis: Likewise

mysql-test/t/percona_userstat.test
```diff
@@ -177,4 +177,82 @@ FROM
...
+--source include/wait_until_count_sessions.inc
+
+# Unknown users are counted in USER_STATISTICS, concurrent connections should be positive
+--disable_query_log
+--error ER_ACCESS_DENIED_ERROR
+connect (conn4,localhost,mysqltest_unknownuser,,);
+--enable_query_log
+connection default;
+
+--let $assert_text= I_S.USER_STATISTICS CONCURRENT_CONNECTIONS: unknown user conneciton
```
laurynas-biveinis: s/conneciton/connection

sql/sql_connect.cc
```diff
@@ -1426,6 +1432,10 @@ void do_handle_one_connection(THD *thd_arg)
     if (unlikely(opt_userstat))
     {
       thd->update_stats(false);
+      if (create_user) {
+        DBUG_ASSERT(thd->diff_disconnects == 0);
+        thd->diff_disconnects = 1;
```
laurynas-biveinis: No space before assignment operator

laurynas-biveinis (changes requested):
Almost OK, only minor comments

ihanick:
http://jenkins.percona.com/job/percona-server-5.5-param/1446/
- fixed typo in assert message
- fixed space with =
- keep wait_until_count_sessions.inc only after disconnect

laurynas-biveinis (approved):

ihanick:
http://jenkins.percona.com/job/percona-server-5.5-param/1447/
- threadpool: move disconnect flag set to end_connection. Allows to avoid 5.6 merge conflict and work with thread pool
- added a test case for threadpool and user_statistics.
- using thd->diff_denied_connections instead of create_user flag to detect unauthorized users (required, because user_stats->concurrent_connections & user_stats->total_connections incremented only for authorized users).

mysql-test/r/percona_userstat_threadpool.result
```diff
@@ -0,0 +1,167 @@
+DROP TABLE IF EXISTS t1;
```
laurynas-biveinis: It looks like the tests require different organisation now that some things have to be tested with the threadpool:
- split percona_userstat.test into two parts: one that does not concern with connection handling (probably it will be the old percona_userstat.test content as before this patch), and another for userstat for connection handling. Leave the former alone
- Move the second part into an include file (e.g. mysql-test/include/percona_userstat_conn_handling.inc), create two testcases percona_userstat_conn_handling.test and percona_userstat_conn_handling_threadpool.test that include it

mysql-test/t/percona_userstat_threadpool.cnf
```diff
@@ -0,0 +1,15 @@
+!include include/default_my.cnf
```
laurynas-biveinis: Replace with a -master.opt file that simply has --connection-handling=pool-of-threads

laurynas-biveinis (changes requested):

ihanick:
http://jenkins.percona.com/job/percona-server-5.5-param/1448/
- using -master.opt file with --thread-handling=pool-of-threads

laurynas-biveinis:
The other comment does not look addressed from the last review round?

ihanick:
http://jenkins.percona.com/job/percona-server-5.5-param/1453/
- Separate include for concurrent_connections field testing
- one row test case & one row + have_thread_pool include

laurynas-biveinis (approved):

laurynas-biveinis:
This has caused https://bugs.launchpad.net/percona-server/+bug/1661488

PR #1137, opened 2016-10-24
Ps 5.7 myr 10

george-lorch:


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -44,7 +44,6 @@
 #include "rocksdb/slice_transform.h"
 #include "rocksdb/utilities/checkpoint.h"
 #include "rocksdb/utilities/convenience.h"
-#include "rocksdb/utilities/flashcache.h"
```
laurynas-biveinis: This and everything below should have been merged to 5.7 in their respective PRs, no?

george-lorch: Hmm, yeah, that is odd. That all should have been merged already and shouldn't appear here as part of the rebased starting point.

george-lorch: Ahh, I know what happened, my local 5.7 base branch has diverged from the real 5.7 branch...meaning I merged locally as part of an earlier test rather than pulling the post merge changes that were merged via github. Anyway, correcting and pushing a fixed branch for this.

laurynas-biveinis (approved):

PR #1138, opened 2016-10-25
 MYR-9 : myrocks requires use of new class in facebook mysql Regex_li…

george-lorch:
…st_handler

```
- Took facebook/webscale implementation of Regex_list_handler from their sql/handler.h and .cc and put it into rdb_utils.h and .cc
```

### DIFF
--- a/storage/rocksdb/rdb_utils.cc
+++ b/storage/rocksdb/rdb_utils.cc
@@ -308,4 +308,73 @@ bool rdb_database_exists(const std::string& db_name)
   return true;
 }
 
+/*
+  Set the patterns string.  If there are invalid regex patterns they will
+  be stored in m_bad_patterns and the result will be false, otherwise the
+  result will be true.
+*/
+bool Regex_list_handler::set_patterns(const std::string& pattern_str)
+{
+  bool pattern_valid= true;
+
+  // Create a normalized version of the pattern string with all delimiters
+  // replaced by the '|' character
+  std::string norm_pattern= pattern_str;
+  std::replace(norm_pattern.begin(), norm_pattern.end(), m_delimiter, '|');
+
+  // Make sure no one else is accessing the list while we are changing it.
+  mysql_rwlock_wrlock(&m_rwlock);
+
+  // Clear out any old error information
+  m_bad_pattern_str.clear();
+
+  try
+  {
+    // Replace all delimiters with the '|' operator and create the regex
+    // Note that this means the delimiter can not be part of a regular
+    // expression.  This is currently not a problem as we are using the comma
+    // character as a delimiter and commas are not valid in table names.
+    m_pattern.reset(new std::regex(norm_pattern));
+  }
+  catch (const std::regex_error& e)
+  {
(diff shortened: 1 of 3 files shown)


storage/rocksdb/rdb_utils.cc
```diff
@@ -308,4 +308,77 @@ bool rdb_database_exists(const std::string& db_name)
...
+  // Clear out any old error information
+  m_bad_pattern_str.clear();
+
+  try
+  {
+    // Replace all delimiters with the '|' operator and create the regex
+    // Note that this means the delimiter can not be part of a regular
+    // expression.  This is currently not a problem as we are using the comma
+    // character as a delimiter and commas are not valid in table names.
+    const std::regex* pattern= new std::regex(norm_pattern);
```
laurynas-biveinis: Consider making m_pattern a std::unique_ptr, and the try block collapses to
`m_pattern.reset(new std::regex(norm_pattern));`
and...

storage/rocksdb/rdb_utils.h
```diff
@@ -203,4 +204,62 @@ std::string rdb_hexdump(const char *data, std::size_t data_len,
...
+#if defined(HAVE_PSI_INTERFACE)
+  Regex_list_handler(const PSI_rwlock_key& key,
+                     char delimiter= ',') :
+    m_key(key),
+#else
+  Regex_list_handler(char delimiter= ',') :
+#endif
+    m_delimiter(delimiter),
+    m_bad_pattern_str(""),
+    m_pattern(nullptr)
```
laurynas-biveinis: ...and this disappears...

storage/rocksdb/rdb_utils.h
```diff
@@ -203,4 +204,62 @@ std::string rdb_hexdump(const char *data, std::size_t data_len,
...
+  {
+    mysql_rwlock_init(key, &m_rwlock);
+  }
+
+  ~Regex_list_handler()
+  {
+    mysql_rwlock_destroy(&m_rwlock);
+    delete m_pattern;
```
laurynas-biveinis: ..and this disappears too.

laurynas-biveinis (changes requested):

storage/rocksdb/rdb_utils.cc
george-lorch: True, this may be a decent improvement, but I wasn't really looking into improving the implementation, just dragging it down from the upstream handler.h into the MyRocks codebase as-is. I'll push a tested change later.

george-lorch:
Done...

laurynas-biveinis (approved):
Once it's tested, please send to upstream too

george-lorch:
It passed my local MTR run without introducing any regressions so I think it is good. Sending an actual patch upstream is a little more interesting because their implementation is in sql/handler.h but I think sending out diff with an explanation in a bug report will be enough for them to get the picture.

george-lorch:
https://github.com/facebook/mysql-5.6/issues/363

PR #1141, opened 2016-10-25
MYR-22 : myrocks depends on minor const correctness bits in the commo…

george-lorch:
…n server code
- Took const correctness changes into server

### DIFF
--- a/sql/field.cc
+++ b/sql/field.cc
@@ -10498,7 +10498,7 @@ Create_field::Create_field(Field *old_field,Field *orig_field) :
     length
 */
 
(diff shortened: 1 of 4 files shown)


george-lorch:
https://jira.percona.com/browse/MYR-22
https://blueprints.launchpad.net/percona-server/+spec/myrocks-const-correct-5.6
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1460/

laurynas-biveinis (approved):
Since this a bugfix, an LP bug instead of a blueprint would be enough, but no need to change now

PR #1144, opened 2016-10-26
Fix bug 1636500 (Threadpool "thread limit reached" / "failed to creat…

laurynas-biveinis:
…e thread" should be printed on the 1st occurrence too)

Currently, threadpool prints "max thread limit reached" and "failed to
create a thread" messages only if 1) the above situation has already
occurred, 2) more than 30 seconds have passed, and 3) the above
situation has occured again.

Adjust this so that diagnostics is printed on the first time the
situation occurrs, and then reprinted as needed, but not more often
than every 30 seconds.

This also fixes bug 1635184 (Spurious warnings on main.pool_of_threads)

http://jenkins.percona.com/job/percona-server-5.5-param/1423/


percona-ysorokin (approved):
LGTM

PR #1160, opened 2016-11-15
MYR-13 : webscale/myrocks has "Extra Stats : Per Table" patch that Pe…

george-lorch:
…rcona Server does not

- upstream patch https://github.com/facebook/mysql-5.6/commit/19f50b46bb51f335be769d1f2550c90d8b41b968
- Stripped use of new table status vars
- Fixed up mtr tests and ensure teste that require userstats have userstat=on
- Renamed userstat test case to match PS naming convention of same test for other engines.


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1468/

laurynas-biveinis (approved):

PR #1162, opened 2016-11-15
MYR-15 : Optionally block/log queries relying on Gap Locks

george-lorch:
- Implemented partial patch to raise error if SE is transactional, does not support gap locks and a gap lock is being attempted.
- Implemented new handler virtual calls for has_gap_locks() in different SEs that need it.
- Imported and adjusted test cases for gap lock errors and added partition test.
- Disabled numerous MyRocks tests as a result of using SQL that invokes 'missing' gap lock functionality and generates new gap lock error.


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1469/

mysql-test/include/gap_lock_error_cleanup.inc
```diff
@@ -0,0 +1,6 @@
+
+drop table if exists gap1, gap2, gap3, gap4;
+
+--exec grep "Got error" $MYSQLTEST_VARDIR/log/mysqld.1.err | grep "gap" | wc -l
```
laurynas-biveinis: This is going to be unstable down the road: by default the error log includes all the content from all the preceding tests, thus it can be anything. If you really need to grep the error log, add a temp error log option to mysqld for a testcase (and use include/search_pattern_in_file.inc)

mysql-test/include/gap_lock_error_cleanup.inc
```diff
@@ -0,0 +1,6 @@
...
+
+flush general logs;
```
laurynas-biveinis: Why do you need this?

mysql-test/include/gap_lock_error_cleanup.inc
```diff
@@ -0,0 +1,6 @@
+
+drop table if exists gap1, gap2, gap3, gap4;
```
laurynas-biveinis: No need for "if exists"? Better to catch testcase bugs early :)

mysql-test/include/gap_lock_error_init.inc
```diff
@@ -0,0 +1,28 @@
+--disable_warnings
+drop table if exists gap1,gap2,gap3;
+--enable_warnings
```
laurynas-biveinis: Obsolete idiom, just remove

mysql-test/t/gap_lock_error_memory_myisam.test
```diff
@@ -0,0 +1,17 @@
+# Non transactional engines should ignore gap lock errors
+
+let $engine=memory;
```
laurynas-biveinis: File name says myisam, engine says memory

sql/ha_partition.cc
```diff
@@ -5663,6 +5663,19 @@ int ha_partition::read_range_next()
...
+{
+  handler **file;
+  DBUG_ENTER("ha_partition::has_gap_locks");
+
+  file= m_file;
+  do
+  {
+    if (!(*file)->has_gap_locks())
+      DBUG_RETURN(FALSE);
+  } while (*(++file));
```
laurynas-biveinis, line 5676: Is it not enough to check only the first partition?

sql/ha_partition.cc
```diff
@@ -5663,6 +5663,19 @@ int ha_partition::read_range_next()
   DBUG_RETURN(handle_unordered_next(table->record[0], eq_range));
 }
 
+bool ha_partition::has_gap_locks()
```
laurynas-biveinis: Let's make a note somewhere (maybe in the 8.0 blueprint), that when legacy partitioning is dead, this method should move from handler (per-table property) to handlerton (per-SE property), which it fundamentally is, if not ha_partition

mysql-test/include/gap_lock_error_select.inc
```diff
@@ -0,0 +1,87 @@
...
+--error ER_UNKNOWN_ERROR
+eval select * from gap1 order by id1 desc limit 1 $select_lock;
+--error ER_UNKNOWN_ERROR
+eval select * from gap1 order by id1 desc, id2 desc, id3 desc
+  limit 1 $select_lock;
+# secondary index lookup
+--error ER_UNKNOWN_ERROR
+eval select * from gap1 force index(i) where c1=1 $select_lock;
+# unique index lookup
+eval select * from gap3 force index(ui) where value=1 $select_lock;
```
laurynas-biveinis, line 44: Add a comment for non-failing cases why they don't fail?

mysql-test/t/gap_lock_error_memory_myisam.test
```diff
@@ -0,0 +1,17 @@
...
+let $expect_gap_lock_errors=0;
+--source include/gap_lock_error_init.inc
+select sum(value) from gap1;
+update gap1 set gap1.value= (select count(*) from gap2);
+select sum(value) from gap1;
+--source include/gap_lock_error_cleanup.inc
+
+let $engine=myisam;
```
laurynas-biveinis: Duplicated?

sql/ha_partition.h
```diff
@@ -616,6 +616,8 @@ class ha_partition :public handler
 			       bool eq_range, bool sorted);
   virtual int read_range_next();
 
+  virtual bool has_gap_locks();
```
laurynas-biveinis: const

sql/handler.h
```diff
@@ -3624,7 +3635,8 @@ const char *ha_legacy_type_name(legacy_db_type legacy_type);
 const char *get_canonical_filename(handler *file, const char *path,
                                    char *tmp_path);
 bool mysql_xa_recover(THD *thd);
-
+bool can_hold_read_locks_on_select(THD *thd, thr_lock_type lock_type);
```
laurynas-biveinis: static (looks like a helper method which has little to do with handler, and not that much even with THD)

sql/handler.h
```diff
@@ -2365,7 +2365,18 @@ class handler :public Sql_alloc
     return prepare_index_key_scan(key, key_len);
   }
 
+  /**
+    Query storage engine to see if it supports gap locks on this table.
+  */
+  virtual bool has_gap_locks() { return false; }
+
 protected:
+  bool is_using_full_key(key_part_map keypart_map, uint actual_key_parts);
```
laurynas-biveinis: If possible, move somewhere more applicable, otherwise static

sql/handler.h
```diff
@@ -2365,7 +2365,18 @@ class handler :public Sql_alloc
...
+  bool is_using_full_unique_key(uint active_index,
```
laurynas-biveinis, line 2375: const

sql/handler.h
```diff
@@ -3624,7 +3635,8 @@ const char *ha_legacy_type_name(legacy_db_type legacy_type);
 const char *get_canonical_filename(handler *file, const char *path,
                                    char *tmp_path);
 bool mysql_xa_recover(THD *thd);
-
+bool can_hold_read_locks_on_select(THD *thd, thr_lock_type lock_type);
+bool can_hold_locks_on_trans(THD *thd, thr_lock_type lock_type);
```
laurynas-biveinis: Likewise

sql/handler.h
```diff
@@ -2365,7 +2365,18 @@ class handler :public Sql_alloc
...
+    Query storage engine to see if it supports gap locks on this table.
+  */
+  virtual bool has_gap_locks() { return false; }
+
 protected:
+  bool is_using_full_key(key_part_map keypart_map, uint actual_key_parts);
+  bool is_using_full_unique_key(uint active_index,
+                                key_part_map keypart_map,
+                                enum ha_rkey_function find_flag);
+  bool is_using_prohibited_gap_locks(TABLE* table,
```
laurynas-biveinis, line 2378: const

sql/handler.h
```diff
@@ -2365,7 +2365,18 @@ class handler :public Sql_alloc
     return prepare_index_key_scan(key, key_len);
   }
 
+  /**
+    Query storage engine to see if it supports gap locks on this table.
+  */
+  virtual bool has_gap_locks() { return false; }
```
laurynas-biveinis: const

storage/tokudb/ha_tokudb.h
```diff
@@ -816,6 +816,8 @@ class ha_tokudb : public handler {
     int index_first(uchar * buf);
     int index_last(uchar * buf);
 
+    bool has_gap_locks() { return true; }
```
laurynas-biveinis: const

storage/innobase/handler/ha_innodb.h
```diff
@@ -144,6 +144,8 @@ class ha_innobase: public handler
 	int index_first(uchar * buf);
 	int index_last(uchar * buf);
 
+	bool has_gap_locks() { return true; }
```
laurynas-biveinis: const

storage/blackhole/ha_blackhole.h
```diff
@@ -94,6 +94,7 @@ class ha_blackhole: public handler
   THR_LOCK_DATA **store_lock(THD *thd,
                              THR_LOCK_DATA **to,
                              enum thr_lock_type lock_type);
+  bool has_gap_locks() { return true; }
```
laurynas-biveinis: const

laurynas-biveinis (changes requested):

mysql-test/include/gap_lock_error_cleanup.inc
george-lorch: Makes sense, this was copied/inherited from the upstream test and I didn't think much of it. I might report it to them for the same reasons.

mysql-test/include/gap_lock_error_cleanup.inc
george-lorch: Not sure, again, inherited from upstream test and didn't give it a second thought. I wonder if it has something to do with clearing state of the gap lock log that would have been part of that patch. For our purposes I can't see how or where it would be needed at all.

mysql-test/t/gap_lock_error_memory_myisam.test
george-lorch: Again, inherited from upstream patch. Filename says both myisam and memory, it tests both engines. Maybe should be split up into two different tests.

sql/ha_partition.cc
george-lorch: Yeah, well I would also hope that by then we have gap lock support and can drop this entirely, but then, this still might make sense to keep around.

sql/ha_partition.cc
george-lorch: Maybe, but I believed there were conditions where a single partition could be altered to use a different engine than was created with, this is being a bit over conservative based on other behaviors in ha_partition to ensure nothing slips through the cracks.

mysql-test/include/gap_lock_error_cleanup.inc
george-lorch: Actually, this is a really good catch. I'm not sure why this is looking into the err log as there is nowhere that anything is ever written to the server error log on a gap lock violation, all errors go to the client and in the upstream patch they are also optionally written to the gap lock log. I believe this serves no useful purpose at all.

sql/ha_partition.cc
george-lorch: Originally I was going to add it just as a handlerton flag, then implement the virtual similar to has_transactions that just returns flagset & flag, but the handlerton flags seems to be getting full and is not possible/practical for the partition handler to accurately reflect what the underlying handlerton supports via the HTON flags. Also as mentioned elsewhere, it is my understanding (perhaps incorrect) that there are some ways to alter the engine of an individual partition within a partitioned table, thus requiring something like this to scan all of the partitions to ensure 100% support.

sql/handler.h
george-lorch: These two actually look like unnecessary refactors and could/should probably just have their logic expanded inline. They are used from one place in handler.cc, the new method is_using_prohibited_gap_locks, and was a factor out of ha_innodb.cc.

ghost:
Some changes made as discussed and new jenkins http://jenkins.percona.com/job/percona-server-5.6-param/1474/

laurynas-biveinis (approved):

PR #1164, opened 2016-11-16
Re-implement kill idle transaction feature on socket level

laurynas-biveinis:
This implements blueprint
https://blueprints.launchpad.net/percona-server/+spec/socket-kill-idle-trx.

Remove all the InnoDB code for the feature and replace it by setting a
custom timeout value on sockets whose connections have transactions open.

Keep the old innodb_kill_idle_trx variable as an alias for the new
kill_idle_transaction one, which works with all the storage engines.

Removing InnoDB implementation also fixes its associated bugs:
- bug 1166744 (Some issues with kill_idle_transaction implementation at
InnoDB level);
- bug 1179136 (Kill idle transaction debug build assert: safe_mutex:
Trying to destroy a mutex that was locked at sql_class.cc, line 880 at
sql_class.cc, line 1773);
- bug 1369373 (minor changes to optimize innodb_kill_idle_transaction).

At the same time expand MTR testing coverage for the feature, add TokuDB
tests, and storage engine-independent tests, and create sys_vars tests
for both the old and new variable, fixing bug 907719.

http://jenkins.percona.com/job/percona-server-5.6-param/1471/


percona-ysorokin (approved):
LGTM

PR #1166, opened 2016-11-17
Fix bug 1642230 (handle_fatal_signal (sig=11) in THD::set_new_thread_id)

laurynas-biveinis:
With threadpool enabled, if an out-of-memory error happens while
creating a new THD object for an incoming object, the object will be
NULL, and threadpool code will dereference it.

Fix by checking for this condition and properly closing the
communication channel instead.

Start a new debug build MTR testcase for threadpool,
main.threadpool_debug.

http://jenkins.percona.com/job/mysql-5.7-param/500/ and http://jenkins.percona.com/job/mysql-5.7-param/502/

### DIFF
--- a/sql/threadpool_unix.cc
+++ b/sql/threadpool_unix.cc
@@ -1301,6 +1301,7 @@ static void wait_end(thread_group_t *thread_group)
 static connection_t *alloc_connection(THD *thd)
 {
   DBUG_ENTER("alloc_connection");
+  DBUG_EXECUTE_IF("simulate_tp_alloc_connection_oom", DBUG_RETURN(NULL););
 
   connection_t* connection = (connection_t *)
       my_malloc(key_memory_thread_pool_connection,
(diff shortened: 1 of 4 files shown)


sql/threadpool_unix.cc
```diff
@@ -1328,6 +1328,13 @@ bool Thread_pool_connection_handler::add_connection(Channel_info *channel_info)
   DBUG_ENTER("Thread_pool_connection_handler::add_connection");
 
   THD* thd= channel_info->create_thd();
+
+  if (unlikely(!thd))
+  {
+    channel_info->send_error_and_close_channel(ER_OUT_OF_RESOURCES, 0, false);
+    DBUG_RETURN(true);
```
percona-ysorokin, line 1336: Do we need to delete channel_info here as well?

laurynas-biveinis: No - if true is returned caller deletes it. But this shows that the next exit from the function, if alloc_connection calls returns NULL, is wrong. Fixing.

laurynas-biveinis:
http://jenkins.percona.com/job/mysql-5.7-param/505/

percona-ysorokin (approved):
LGTM

PR #1170, opened 2016-11-18
Fix bug 1641612 (New compilation warnings with GCC 6 (-Wshift-negativ…

laurynas-biveinis:
…e-value, -Wdeprecated-declarations, -Wnonnull-compare))

"Left shift of negative value warnings are caused by (~0 << x) code
construct, where constant zero and its complement are of type signed
int. Fix by replacing 0 with 0U.

"‘readdir_r’ is deprecated" warnings are fixed by observing that glibc
readdir is thread-safe for non-shared directory streams, thus
replacing readdir_r with readdir is easy and simplifies code.

"warning: nonnull argument 'foo' compared to NULL" is fixed by
backporting [1], which removes nonnull attribute as unsafe. It also
removes pure attribute due to a compiler bug. In this backport, where
applicable, warn_unused_result attribute has been added.

A special case of the above is comparison of this to NULL in a
partition_info::get_clone member function. Fixed by backporting the
relevant bit of [2].

[1]

commit 7b899ac00c5d6a18b5b8eda661885f6f1afc502a
Author: Marko Makela <marko.makela@oracle.com>
Date: Wed Oct 8 10:31:13 2014 +0300

    Bug#19632776 Code gets incorrectly optimized away by gcc

    As reported in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=764220
    it is possible that gcc incorrectly optimizes away some code
    when a function that takes a pointer or reference as a parameter
    is declared as attribute((const)) or attribute((pure)).

    The function affected was page_zip_rec_needs_ext().

    As a preventive measure, we will remove the potentially problematic
    attributes from all functions that take pointers or references that
    they are dereferencing. Functions that perform pointer arithmetics
    without dereferencing the pointers should be safe: page_offset() is
    an example.

    While we are at it, remove also some attribute((nonnull)).
    This attribute is dangerous, because it does not always generate
    a warning when NULL may be passed, but it may optimize away code
    for handling the NULL case. We wanted this attribute for the sake
    of the warnings, not for the optimizations.

    rb#6940 approved by Vasil Dimov and Jimmy Yang

[2]

commit 33c2404b397e1077daaf0ef0ff9edba445430f5f
Author: Vasil Dimov <vasil.dimov@oracle.com>
Date: Wed Jan 21 10:33:58 2015 +0200

    Fix Bug#20380977 COMPILER WARNINGS IN MYSQL-5.6

    Remove unsued symbols and resolve other compiler warnings accordingly

http://jenkins.percona.com/job/percona-server-5.5-param/1435/


percona-ysorokin (approved):
LGTM

gl-sergei:
Upstream fix for "‘readdir_r’ is deprecated" warnings is in commit c6a1902

laurynas-biveinis:
Thanks - I missed it. The patches look similar, upstream cleans up whitespace, makes directory_file_name static, and removed dirent.h CMake checks. The difference does not look critical to backport

PR #1175, opened 2016-11-21
Bug 1641910: Trying to set audit_log_exclude_accounts crashes server

gl-sergei:
The root cause of this bug is that logic around filtering variables rely
heavily on the fact that CHECK and UPDATE functions are always called.
This is not true when variables passed via defaults file or set as
command line options.

The fix is add special handling for filtering variables on startup.

http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1477/


mysql-test/t/audit_log_acc.test
```diff
@@ -0,0 +1,71 @@
+# Bug 1641910: Trying to set audit_log_exclude_accounts crashes server.
+
+call mtr.add_suppression("Plugin 'audit_log' init function returned error.");
+call mtr.add_suppression("Plugin 'audit_log' registration as a AUDIT failed.");
+
+--exec echo "wait" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
```
laurynas-biveinis: Here and everywhere please use include/restart_mysqld.inc, it supports restart parameters.

mysql-test/t/audit_log_acc.test
```diff
@@ -0,0 +1,71 @@
+# Bug 1641910: Trying to set audit_log_exclude_accounts crashes server.
```
laurynas-biveinis: Use one of the existing test cases?

plugin/audit_log/audit_log.c
```diff
@@ -811,7 +811,53 @@ int audit_log_plugin_init(void *arg MY_ATTRIBUTE((unused)))
 
   audit_log_filter_init();
 
-  return(0);
+  if (audit_log_exclude_accounts != NULL && audit_log_include_accounts != NULL)
```
laurynas-biveinis, line 814: IMHO this added code belongs to audit_log_filter_init, up to you to consider

laurynas-biveinis (changes requested):

plugin/audit_log/audit_log.c
gl-sergei: audit_log_filer_init located in separate module and doesn't know about plugin variables. This code is aimed to validate plugin variables.

mysql-test/t/audit_log_acc.test
gl-sergei: This test case does a lot of server restarts to test how different variables combinations are tested on startup. We don't have similar test cases currently.

gl-sergei: the name is bad... maybe name it audit_log_bug1641910?

mysql-test/t/audit_log_acc.test
gl-sergei: done

mysql-test/t/audit_log_acc.test
laurynas-biveinis: audit_log_startup ?

laurynas-biveinis (changes requested):

gl-sergei:
Renamed test case to `audit_log_startup`.
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1489/

laurynas-biveinis (approved):

PR #1183, opened 2016-11-23
Fix bug 1644174 (Test rpl.rpl_init_slave_errors is unstable)

laurynas-biveinis:
Add missing slave I/O thread stop sync to the testcase.

http://jenkins.percona.com/job/percona-server-5.5-param/1450/

Not a GCA because of https://bugs.launchpad.net/percona-server/+bug/1638897 fix dependency


percona-ysorokin (approved):
LGTM

PR #1184, opened 2016-11-23
Fix bug 1539504 / 83003 (CREATE TEMPORARY TABLE creates a transaction…

laurynas-biveinis:
… in binary log on read only server)

Previously, a session disconnect causes DROP TEMPORARY TABLE IF EXISTS
to be binlogged for all the opened temp tables in that session. Even
though temporary table operation are not otherwise binlogged in row or
mixed mode, this was done regardless of binary log format in use, as
it was not tracked, whether a particular temp table was not create in
STATEMENT mode - in which case it does need the DROP.

For ROW/MIXED users, this behavior was causing spurious binlog writes
and GTIDs generated on otherwise read only servers.

Fix by tracking the binlog format at temporary table create time
(open_table_uncached), and using it to decide whether a DROP should be
logged or not in close_temporary_tables. At the same time fix some
issues with  THD:LOCK_temporary_tables critical sessions in the latter
function.

http://jenkins.percona.com/job/percona-server-5.6-param/1483/

5.7 version will be prepared after this one is reviewed


percona-ysorokin (approved):
LGTM

PR #1187, opened 2016-11-23
ps-5.6-MYR-3.2 : webscale specific performance timers and table stats

george-lorch:
- Copied in atomic_stat.h from upstream FacebookSQL ./include/atomic/
- Copied in upstream struct declarations from ./include/mysqld.h and some helper functions into
  ./storage/rocksdb/rdb_perf_context.h for my_io_perf and my_io_atomic_perf used by Rdb_perf_context.
- Removed portion FacebookSQL patch in MyRocks that implements handlerton function
  update_table_stats which is used to explose statistics in information_schema.table_statistics
- Removed updating of FacebookSQL specific additional data in ha_statistics struct.

### DIFF
--- a/mysql-test/include/plugin.defs
+++ b/mysql-test/include/plugin.defs
@@ -55,3 +55,4 @@ mysql_no_login     plugin/mysql_no_login      MYSQL_NO_LOGIN    mysql_no_login
 test_udf_services  plugin/udf_services TESTUDFSERVICES
 ha_tokudb          storage/tokudb             TOKUDB            tokudb,tokudb_trx,tokudb_locks,tokudb_lock_waits,tokudb_fractal_tree_info,tokudb_background_job_status
 tokudb_backup      plugin/tokudb_backup       TOKUDB_BACKUP     tokudb_backup
+ha_rocksdb         storage/rocksdb            ROCKSDB           rocksdb,rocksdb_cfstats,rocksdb_dbstats,rocksdb_perf_context,rocksdb_perf_context_global,rocksdb_cf_options,rocksdb_global_info,rocksdb_ddl,rocksdb_index_file_map
--- a/storage/rocksdb/rdb_perf_context.h
+++ b/storage/rocksdb/rdb_perf_context.h
@@ -17,6 +17,7 @@
 #pragma once
 
 /* C++ standard header files */
+#include <algorithm>
 #include <atomic>
 #include <cstdint>
 #include <string>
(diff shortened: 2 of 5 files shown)


george-lorch:
http://jenkins.percona.com/job/percona-server-5.6-param/1488/

mysql-test/include/plugin.defs
```diff
@@ -55,3 +55,4 @@ mysql_no_login     plugin/mysql_no_login      MYSQL_NO_LOGIN    mysql_no_login
 test_udf_services  plugin/udf_services TESTUDFSERVICES
 ha_tokudb          storage/tokudb             TOKUDB            tokudb,tokudb_trx,tokudb_locks,tokudb_lock_waits,tokudb_fractal_tree_info,tokudb_background_job_status
 tokudb_backup      plugin/tokudb_backup       TOKUDB_BACKUP     tokudb_backup
+ha_rocksdb         storage/rocksdb            ROCKSDB           rocksdb,rocksdb_cfstats,rocksdb_dbstats,rocksdb_perf_context,rocksdb_perf_context_global,rocksdb_cf_options,rocksdb_global_info,rocksdb_ddl,rocksdb_index_file_map
```
laurynas-biveinis, line 58: Spurious change or some GCA wart?

storage/rocksdb/rdb_perf_context.h
```diff
@@ -17,13 +17,109 @@
...
+
+/* Struct used for IO performance counters within a single thread */
+struct my_io_perf_struct {
+  ulonglong bytes;
+  ulonglong requests;
+  ulonglong svc_time; /*!< time to do read or write operation */
+  ulonglong svc_time_max;
+  ulonglong wait_time; /*!< total time in the request array */
+  ulonglong wait_time_max;
+ulonglong slow_ios; /*!< requests that take too long */
```
laurynas-biveinis: Indentation

storage/rocksdb/rdb_perf_context.h
```diff
@@ -17,13 +17,109 @@
...
+  atomic_stat<ulonglong> svc_time; /*!< time to do read or write operation */
+  atomic_stat<ulonglong> svc_time_max;
+  atomic_stat<ulonglong> wait_time; /*!< total time in the request array */
+  atomic_stat<ulonglong> wait_time_max;
+  atomic_stat<ulonglong> slow_ios; /*!< requests that take too long */
+};
+typedef struct my_io_perf_atomic_struct my_io_perf_atomic_t;
+
+/* Initialize an my_io_perf_t struct. */
+static inline void my_io_perf_init(my_io_perf_t* perf) {
```
laurynas-biveinis: inline functions in an anonymous namespace rather than static inline?

laurynas-biveinis: The functions are clearly asking to become class members of the above structs (while keeping them POD). Up to you if you want to make such change or not, as it will introduce merge issues unless we push it to upstream

laurynas-biveinis (changes requested):

mysql-test/include/plugin.defs
george-lorch: A little of both. Actually is needed as the merged version from MYR-24 is missing the I_S plugin rocksdb_perf_context which is added here.

storage/rocksdb/rdb_perf_context.h
george-lorch: Yeah, I actually considered doing a fair bit of factoring and C++izing of this whole copy/paste thing and merging this all into an appropriately renamed pasted atomic_stat.h since I believe this is not really going to be a temporary change, but am on the fence about it. Moving it all into a single, properly named and namespaced, stand alone header file would reduce potential merge issues and be a lot cleaner for sure.

laurynas-biveinis: OK, I suggest checking with the upstream if they'd accept the patch and going for it

george-lorch: I am unsure what you mean by them accepting a patch. This stuff was selectively copied from upstream [FacebookSQL mysqld.h](https://github.com/facebook/mysql-5.6/blob/webscalesql-5.6.27.75/sql/mysqld.h#L510) into rdb_perf_context.h. I also only copied what was necessary to satisfy the needs of MyRocks and not the complete upstream implementation.

So are you suggesting that I create a patch against the upstream mysqld.h and mysqld.cc that refactors the my_io_perf* as discussed, possibly into its own .h and .cc files, then re-copying that code back into MyRocks for PS? I am not opposed to such a task but it seems going pretty far outside of our scope here.

laurynas-biveinis: Yes - that is my suggestion - as it looks like it will both result in cleaner code for us (and them) and easier upstream merges. Up to you to make the final decision, not doing this is of course fine here

george-lorch: OK, as I said, I am not totally opposed to the idea, just wanted to make sure I understood the suggestion. It does not look to be a huge amount of work other than maybe ensuring it tests well and is fully covered by mtr somehow. I suppose I could take a stab and if I stumble into some rabbit hole just revert and go with some alternative.

george-lorch:
Superseded by https://github.com/percona/percona-server/pull/1283

PR #1188, opened 2016-11-24
Bug 1633988: Assertion `thd == _current_thd()' failed

gl-sergei:
When 'too many connections' error audit log is notified from by main
server thread, but target THD is set to the thread which failed to
connect.

When we are trying to update THD variable of target thread we checking
that current thread is the target thread to make sure operation is
thread-safe. We however are good in case of too many connections error,
i.e. when we are working from main thread and `currnt_thd' is NULL.

http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1491/

### DIFF
--- a/mysql-test/t/audit_log_many_connections.test
+++ b/mysql-test/t/audit_log_many_connections.test
@@ -1,0 +1,31 @@
+#
+# Bug #1633988: Assertion `thd == _current_thd()' failed
+#
+# Test audit log handling of too many connections error
+#
+
+--disable_query_log
+call mtr.add_suppression("Too many connections");
+--enable_query_log
+
+--source include/count_sessions.inc
+
+SET @saved_max_connections = @@global.max_connections;
+SET GLOBAL max_connections = 2;
+
+--connect (con1, localhost, root)
+--connect (con2, localhost, root)
+--replace_result $MASTER_MYPORT MYSQL_PORT $MASTER_MYSOCK MYSQL_SOCK
+--error ER_CON_COUNT_ERROR
+--connect (con3, localhost, root)
+--replace_result $MASTER_MYPORT MYSQL_PORT $MASTER_MYSOCK MYSQL_SOCK
+--error ER_CON_COUNT_ERROR
+--connect (con4, localhost, root)
+
+connection default;
+SET GLOBAL max_connections= @saved_max_connections;
+
+--disconnect con2
+--disconnect con1
+
+--source include/wait_until_count_sessions.inc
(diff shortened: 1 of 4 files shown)


mysql-test/t/audit_log_many_connections.test
```diff
@@ -0,0 +1,31 @@
+#
```
laurynas-biveinis, line 1: Doesn't the testcase need e.g. audit_null installed? Server has other "too many connections" testcases, but these don't currently crash at that assert?

sql/sql_plugin.cc
```diff
@@ -3092,7 +3092,7 @@ static bool plugin_var_memalloc_session_update(THD *thd,
 
 void plugin_thdvar_safe_update(THD *thd, st_mysql_sys_var *var, char **dest, const char *value)
 {
-  DBUG_ASSERT(thd == current_thd);
+  DBUG_ASSERT(current_thd == NULL || thd == current_thd);
```
laurynas-biveinis, line 3095: This is upstream 8.0 bug, correct? Once the patch is final, can you open a bug report with them, mark LP as its downstream copy, and send the patch?

laurynas-biveinis:
Please update the bug title to something more specific

mysql-test/t/audit_log_many_connections.test
gl-sergei: see audit_log_many_connections-master.opt, it installing audit_log

sql/sql_plugin.cc
gl-sergei: Yes. It looks like it must be reproducible with MySQL 8

mysql-test/t/audit_log_many_connections.test
laurynas-biveinis: sorry - missed it. For upstream submission can you convert the testcase to audit_null? Please also add --source include/have_audit_null... at the top

gl-sergei:
Have just figured out that MySQL 8 handles "too many connections" error in `acl_authenticate` from within correct context and `thd` is set properly. It looks like a new feature - now MySQL allows users having `SUPER` privilege to connect even if maximum connections limit has been reached.

mysql-test/t/audit_log_many_connections.test
laurynas-biveinis: Since this is not an upstream bug anymore, the conversion to audit_null becomes only "nice to have", up to you to decide. Otherwise LGTM

gl-sergei: We need specific behaviour from plugin to reproduce this issue. It needs to update its own MEMALLOC'ed variables. audit_null doesn't not do so. We can add this for the test case, but I think it OK to keep audit_log as requirement because it simply already triggers this error.

gl-sergei: The test case for THDVAR_SET in upstream added its invocation to example storage engine plugin

laurynas-biveinis (approved):

PR #1193, opened 2016-11-24
Work towards diagnosing bug 1644198 (Test rpl.rpl_binlog_errors is un…

laurynas-biveinis:
…stable)

Make mysqltest print the PID of a process it is sending
SIGABRT/SIGKILL to.

http://jenkins.percona.com/job/percona-server-5.5-param/1452/


percona-ysorokin (approved):
LGTM

PR #1201, opened 2016-11-24
Added waiting conditions to rpl_semi_sync test when semi sync master

laurynas-biveinis:
status depend on round trip to slave to avoid test failures (result
content mismatch) on slow platforms.

http://jenkins.percona.com/job/percona-server-5.5-param/1456/


percona-ysorokin (approved):
LGTM

PR #1206, opened 2016-11-24
Fix bug 1644583 (Test main.percona_expand_fast_index_creation is unst…

laurynas-biveinis:
…able)

Mask row estimate from EXPLAIN output, as it's unstable with InnoDB
random-dive statistics.

http://jenkins.percona.com/job/percona-server-5.6-param/1494/


percona-ysorokin (approved):
LGTM

PR #1213, opened 2016-11-28
MYR-23 : warning: variable length array is used [-Wvla]

george-lorch:
- Changed use of vla to my_malloc/my_free

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -2799,10 +2799,10 @@ static std::string format_string(
     res = std::string("");
   }
   else {
-    char buff[len];
-    (void) vsnprintf(buff, len, format, args_copy);
-
-    res = std::string(buff);
+    std::unique_ptr<char[], decltype(&my_free)> buff(nullptr, &my_free);
+    buff.reset(static_cast<char*>(my_malloc(len, MYF(0))));
+    (void) vsnprintf(buff.get(), len, format, args_copy);
+    res = buff.get();
   }
 
   va_end(args_copy);


george-lorch:
No jenkins as the code change is entirely within the MyRocks SE.

laurynas-biveinis (approved):
LGTM, but AFAIK upstream cares about warnings - do you have any guess why they haven't seen this one then?

george-lorch:
I do not know why they have not seen/fixed it but I suspect it is either some gcc version difference or some inherited flags difference. This one I am planning to send upstream, just ran out of time today to do it.

george-lorch:
Actually, this should probably use std::unique_ptr instead of raw C type. I'll fix and push.

laurynas-biveinis:
std::vector then, unique_ptr is suboptimal with arrays

laurynas-biveinis:
Ignore my last comment if you are keeping my_*

george-lorch:
Yeah, I already pushed a change that uses my_ and unique_ptr then later saw your comment on vector. I think for this particular use case, where resizing is not needed, and it is desirable to keep my_*, and only scoped lifetime management of allocated blob is needed, that the unique_ptr is probably the way to go.

laurynas-biveinis (approved):

george-lorch:
Sigh, upstream chose a different and IMHO more strained approach for something that is minimal impact https://github.com/facebook/mysql-5.6/pull/458/files...I think ours is simpler and less intrusive for some effectively no impact in results.

george-lorch:
Closing as we will inherit a fix in the next upstream refresh.

PR #1215, opened 2016-11-30
Add the ability to setup several plugin directories.

vlad-lesin:
plugin_dir option can contain several directories splitted with ':' delimiter.

The order of some plugin lookup is the same as the order of direcrories in
plugin_dir string. If there is error during loading some plugin the lookup is
stopped for this plugin.

Testing:
http://jenkins.percona.com/job/percona-server-5.6-param/1495/


laurynas-biveinis:
The feature needs a spec blueprint. I am not sure about the foo:bar syntax, it paints us into the corner should we start support Windows years down the road, and there is no precedent for such separator in MySQL (the closest is innodb_data_file_path, but it does not work on directories/paths, only file names (?)). Maybe --plugin-dir-add option following the --plugin-load-add convention?

sql/sql_plugin.cc
```diff
@@ -311,6 +311,14 @@ static void report_error(int where_to, uint error, ...)
   }
 }
 
+static void error_log_print_wrapper(const char *fmt, ...)
```
laurynas-biveinis, line 314: Use sql_print_error

sql/sql_plugin.cc
```diff
@@ -450,18 +458,30 @@ static inline void free_plugin_mem(struct st_plugin_dl *p)
...
+                                           int report,
+                                           bool /* OUT */ *lib_open_error,
+                                           int /* OUT */ *err,
+                                           char /* OUT */ *err_msg,
+                                           size_t err_msg_size)
 {
 #ifdef HAVE_DLOPEN
   char dlpath[FN_REFLEN];
-  uint plugin_dir_len, dummy_errors, dlpathlen, i;
+    uint plugin_dir_len, dummy_errors, dlpathlen, i;
```
laurynas-biveinis, line 471: Spurious

sql/sql_plugin.cc
```diff
@@ -472,7 +492,10 @@ static st_plugin_dl *plugin_dl_add(const LEX_STRING *dl, int report)
                                system_charset_info, 1) ||
       plugin_dir_len + dl->length + 1 >= FN_REFLEN)
   {
-    report_error(report, ER_UDF_NO_PATHS);
+    *err= ER_UDF_NO_PATHS;
```
laurynas-biveinis, line 495: Here and below, why?

sql/sql_plugin.cc
```diff
@@ -694,6 +845,7 @@ static struct st_plugin_int *plugin_find_internal(const LEX_STRING *name, int ty
 
   mysql_mutex_assert_owner(&LOCK_plugin);
 
+
```
laurynas-biveinis, line 848: Spurious

laurynas-biveinis (changes requested):
Partial review, as the specification has to be discussed first

sql/sql_plugin.cc
vlad-lesin: Good point. The reason is initially I wanted to add the ability to lookup plugin until it can be loaded, so there could be several plugin libs with load error and this error should be handled by the up-level function. But now the lookup stops on first plugin file load error, so there is no need to pass error string to caller.

PR #1216, opened 2016-11-30
Multiple plugin dir 5.7

vlad-lesin:
plugin_dir option can contain several directories splitted with ':' delimiter.

The order of some plugin lookup is the same as the order of direcrories in
plugin_dir string. If there is error during loading some plugin the lookup is
stopped for this plugin.

Testing:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/528

perfschema.show_plugin* tests fail looks suspicious, but they fail without plugin_dir changes.


laurynas-biveinis (changes requested):

PR #1217, opened 2016-12-01
Fixed LP Bug #1616477 by removing the 5.6-specific paragraph

hrvojem:
Document now passes doc8 run


laurynas-biveinis (commented):
It's hard to find non-formatting changes in this diff - is it only the "priority refill" paragraph removed?

PR #1219, opened 2016-12-01
Fix bug 1631309: rollback 1433432 and 1ms lru manager sleep increment

ihanick:
Fix for https://bugs.launchpad.net/bugs/1631309 and https://bugs.launchpad.net/percona-server/+bug/1433432
http://jenkins.percona.com/job/percona-server-5.6-param/1504/

My previous fix for  1433432 was invalid: spinning happening only with small buffer pool (lru manager jumps in burst mode and never exits to normal mode even after removing load).

During benchmarks it was found that 50ms increments for 5-20% filling of free list reducing performance for small buffer pools and slightly reducing performance for fully filled big buffer pools.

### DIFF
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -2206,13 +2206,14 @@ Clears up tail of the LRU lists:
 * Flush dirty pages at the tail of LRU to the disk
 The depth to which we scan each buffer pool is controlled by dynamic
 config parameter innodb_LRU_scan_depth.
-@return number of pages flushed */
+@return number of flushed and evicted pages */
 UNIV_INTERN
 ulint
 buf_flush_LRU_tail(void)
 /*====================*/
 {
 	ulint	total_flushed = 0;
+	ulint	total_evicted = 0;
 	ulint	start_time = ut_time_ms();
 	ulint	scan_depth[MAX_BUFFER_POOLS];
 	ulint	requested_pages[MAX_BUFFER_POOLS];
@@ -2278,6 +2279,7 @@ buf_flush_LRU_tail(void)
 				limited_scan[i]
 					= (previous_evicted[i] > n.evicted);
 				previous_evicted[i] = n.evicted;
+				total_evicted += n.evicted;
 
 				requested_pages[i] += lru_chunk_size;
 
@@ -2310,7 +2312,7 @@ buf_flush_LRU_tail(void)
 			MONITOR_LRU_BATCH_PAGES,
 			total_flushed);
 	}
-	return(total_flushed);
+	return(total_flushed + total_evicted);
 }
 
 /*********************************************************************//**
@@ -2604,6 +2606,23 @@ buf_get_total_free_list_length(void)
 	return result;
 }
 
+/** Returns the aggregate LRU list length over all buffer pool instances.
+@return total LRU list length. */
+MY_ATTRIBUTE((warn_unused_result))
+static
+ulint
+buf_get_total_LRU_list_length(void)
+{
+        ulint result = 0;
+
+        for (ulint i = 0; i < srv_buf_pool_instances; i++) {
+
+                result += UT_LIST_GET_LEN(buf_pool_from_array(i)->LRU);
+        }
+
+        return result;
+}
+
 /*********************************************************************//**
 Adjust the desired page cleaner thread sleep time for LRU flushes.  */
 MY_ATTRIBUTE((nonnull))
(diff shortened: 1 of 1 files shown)


storage/innobase/buf/buf0flu.cc
```diff
@@ -2611,25 +2611,24 @@ static
...
 /*==============================*/
-	ulint*	lru_sleep_time,	/*!< in/out: desired page cleaner thread sleep
+	ulint*	lru_sleep_time)	/*!< in/out: desired page cleaner thread sleep
 				time for LRU flushes  */
-	ulint	lru_n_flushed) /*!< in: number of flushed in previous batch */
 
 {
 	ulint free_len = buf_get_total_free_list_length();
-	ulint max_free_len = srv_LRU_scan_depth * srv_buf_pool_instances;
+	ulint lru_len = UT_LIST_GET_LEN(buf_pool_from_array(0)->LRU);
```
laurynas-biveinis: In 5.6 this is constant through server lifetime, right? Thus can be computed once?
In 5.7 this can be computed at startup and after each buffer pool resize

laurynas-biveinis (changes requested):

storage/innobase/buf/buf0flu.cc
ihanick: Almost the same expression used inside buf_flush_LRU_tail(void)  
ut_min(srv_LRU_scan_depth, UT_LIST_GET_LEN(buf_pool->LRU)); 
UT_LIST_GET_LEN(buf_pool->LRU) is not a constant and could be changed in runtime for 5.6.
Should I make a function similar to buf_get_total_free_list_length, e.g. buf_get_total_LRU_list_length ? Length parameter is cached in list and this check should be fast.

laurynas-biveinis: Right, I forgot it is cached. No changes needed, let's wait for performace confirmation with large buffer pools

ihanick:
http://jenkins.percona.com/job/percona-server-5.6-param/1561/

- Current 5.6 implementation uses just pages flushed from lru instead of flushed + evicted as done in 5.7.
After adding evicted performance became similar to rollback case.

1ms LRU manager sleep increment also improves this case.

- LRU manager adapt sleep algorithm operates with free list filling 0-1% (polling), 1-5% (decrement sleep time until 50ms), 5-20% (no change for sleep time), 20%+ (increment by 50ms, in patch increment by 1ms). Desired free list size determined as lru_scan_depth, but for small buffer pool instance size scan depth could be bigger than buffer pool size, in this situation length of LRU list (total sum for all BP instances) is considered as maximum free length.

Benchmarks for 1GB, 10GB, 100GB with several buffer pool instances shows same or better performance for both current version and for "rollback to polling + 1ms increment"(previous pull request version).

storage/innobase/buf/buf0flu.cc
```diff
@@ -2206,13 +2206,14 @@ Clears up tail of the LRU lists:
 * Flush dirty pages at the tail of LRU to the disk
 The depth to which we scan each buffer pool is controlled by dynamic
 config parameter innodb_LRU_scan_depth.
-@return number of pages flushed */
+@return number of pages flushed+evicted */
```
laurynas-biveinis: "flushed and evicted"

storage/innobase/buf/buf0flu.cc
```diff
@@ -2310,7 +2312,7 @@ buf_flush_LRU_tail(void)
 			MONITOR_LRU_BATCH_PAGES,
 			total_flushed);
 	}
-	return(total_flushed);
+	return(total_flushed+total_evicted);
```
laurynas-biveinis: Spaces around plus sign

storage/innobase/buf/buf0flu.cc
```diff
@@ -2605,6 +2607,26 @@ buf_get_total_free_list_length(void)
 }
 
 /*********************************************************************//**
+Returns the aggregate LRU list length over all buffer pool instances.
```
laurynas-biveinis: New functions should use the new InnoDB header comment style:
/** Returns ...
... */
and no /*=======*/ underneath

laurynas-biveinis (changes requested):
Minor style comments. The algorithm change looks good

ihanick:
http://jenkins.percona.com/job/percona-server-5.6-param/1565/

- fixed comments and white space

laurynas-biveinis (approved):

PR #1221, opened 2016-12-01
Fix bug 1646384 (ANALYZE TABLE crashes debug build on --innodb-read-o…

laurynas-biveinis:
…nly instance)

Do not attempt to delete the table being analyzed from the background
stats queue on a read-only instance, as that will raise an assertion
and it's impossible for the table to be in the queue in the first
place.

http://jenkins.percona.com/job/percona-server-5.6-param/1501/


percona-ysorokin (approved):
LGTM

PR #1225, opened 2016-12-07
Fix bug 1647723 (mysqltest should diagnose popen call failure)

laurynas-biveinis:
Print diagnostics for popen call returning NULL in mysqltest.

http://jenkins.percona.com/job/percona-server-5.6-param/1512/


percona-ysorokin (approved):
LGTM

PR #1227, opened 2016-12-07
Bug#20874411 INNODB SHUTDOWN HANGS IF INNODB_FORCE_RECOVERY>=3

laurynas-biveinis:
SKIPPED ANY ROLLBACK

trx_sys_any_active_transactions(): If any prepared ACTIVE transactions
exist, and their rollback was prevented by innodb_force_recovery,
convert these transactions to XA PREPARE state in the main-memory
data structures, so that shutdown will proceed normally. These transactions
will again recover as ACTIVE on the next restart, and they will be rolled
back unless innodb_force_recovery>=3 again.

trx_undo_fake_prepared(): An auxiliary function to set an undo log to
PREPARED state.

RB: 8610
Reviewed-by: Jimmy Yang <jimmy.yang@oracle.com>

http://jenkins.percona.com/job/percona-server-5.6-param/1514/

### DIFF
--- a/storage/innobase/trx/trx0sys.cc
+++ b/storage/innobase/trx/trx0sys.cc
@@ -1245,6 +1245,33 @@ trx_sys_close(void)
 	trx_sys = NULL;
 }
 
+/** @brief Convert an undo log to TRX_UNDO_PREPARED state on shutdown.
+
+If any prepared ACTIVE transactions exist, and their rollback was
+prevented by innodb_force_recovery, we convert these transactions to
+XA PREPARE state in the main-memory data structures, so that shutdown
+will proceed normally. These transactions will again recover as ACTIVE
+on the next restart, and they will be rolled back unless
+innodb_force_recovery prevents it again.
+
+@param[in]	trx	transaction
+@param[in,out]	undo	undo log to convert to TRX_UNDO_PREPARED */
+static
+void
+trx_undo_fake_prepared(
+	const trx_t*	trx,
+	trx_undo_t*	undo)
+{
+	ut_ad(srv_force_recovery >= SRV_FORCE_NO_TRX_UNDO);
+	ut_ad(trx_state_eq(trx, TRX_STATE_ACTIVE));
+	ut_ad(trx->is_recovered);
+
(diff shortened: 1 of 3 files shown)


storage/innobase/trx/trx0sys.cc
```diff
@@ -1253,15 +1280,42 @@ ulint
...
 {
-	ulint	total_trx = 0;
-
 	mutex_enter(&trx_sys->mutex);
 
-	total_trx = UT_LIST_GET_LEN(trx_sys->rw_trx_list)
-		  + UT_LIST_GET_LEN(trx_sys->mysql_trx_list);
+	ulint	total_trx = UT_LIST_GET_LEN(trx_sys->mysql_trx_list);
+
+	if (total_trx == 0) {
```
percona-ysorokin, line 1287: The previous version of this function used to return
`UT_LIST_GET_LEN(trx_sys->rw_trx_list) + UT_LIST_GET_LEN(trx_sys->mysql_trx_list) - trx_sys->n_prepared_trx`
The new one, when `UT_LIST_GET_LEN(trx_sys->mysql_trx_list)  != 0` returns just `UT_LIST_GET_LEN(trx_sys->mysql_trx_list)`, ignoring `UT_LIST_GET_LEN(trx_sys->rw_trx_list)` and `trx_sys->n_prepared_trx`
Is this OK?

laurynas-biveinis: It looks correct: while mysql_trx_list is non-empty, this function returns its length, resulting in a shutdown loop with "Waiting for x active transactions to finish" diagnostics. Once it's empty, it checks rw_trx_list, it's either same length as n_prepared_trx, in which case nothing is done and zero returned (correct), either the active transactions are coverted to prepared ones, which should result in zero return value the same.

percona-ysorokin: Yes, I agree. It looks like the calling side really cares only about the return value of this function being zero/non-zero. The value itself is used for diagnostics only.

percona-ysorokin (approved):
LGTM

PR #1229, opened 2016-12-07
Fixed LP Bug #1647728 by documenting the proper way to remove the

hrvojem:
variable before proceeding with the upgrade


laurynas-biveinis (commented):
LGTM if no changes needed for the upgrading page (see my last comment on the bug report)

laurynas-biveinis (approved):

PR #1233, opened 2016-12-09
Fix bug 1648389 (Spurious unsuppressed warning on rpl.rpl_parallel_sw…

laurynas-biveinis:
…itch_sequential)

Whether the "slave coordinator and worker threads are stopped,
possibly leaving data in inconsistent state." warning will be printed
depends solely on timing, thus suppress it for slow hosts.

http://jenkins.percona.com/job/percona-server-5.6-param/1518/


percona-ysorokin (approved):
LGTM

laurynas-biveinis:
Merged manually to correct target branch

PR #1235, opened 2016-12-09
Add the ability to use path in --plugin-load and --plugin-load-add server  command-line options.

vlad-lesin:
New environment variables PLUGVAR_LOAD_PATH and PLUGVAR_LOAD_ADD_PATH are
involved in MTR to use the feature in tests. This is usefull if test uses
several plugins located in different directories like PLUGIN_AUTH and
EXAMPLE_PLUGIN (see mysql-test/include/plugin.defs").

http://jenkins.percona.com/job/percona-server-5.6-param/1519/

### DIFF
--- a/mysql-test/t/plugin-load-add-with-path.test
+++ b/mysql-test/t/plugin-load-add-with-path.test
@@ -1,0 +1,107 @@
+# Test for ability to use path in --plugin-load and --plugin-load-add options
+
+--source include/have_dynamic_loading.inc
+
+# have_example_plugin.inc, have_plugin_auth.inc, have_udf.inc can't be used
+# here because they check if plugin_var variable contains the correspondent
+# plugin path what is wrong as these plugins are in different subdirs in
+# build dir
+
+#
+# Check if the variable EXAMPLE_PLUGIN is set
+#
+if (!$EXAMPLE_PLUGIN) {
+  --skip Example plugin requires the environment variable \$EXAMPLE_PLUGIN to be set (normally done by mtr)
+}
+
+#
+# Check if the variable PLUGIN_AUTH is set
+#
+if (!$PLUGIN_AUTH) {
+  --skip Example plugin requires the environment variable \$PLUGIN_AUTH to be set (normally done by mtr)
+}
+
+#
+# Check if the variable UDF_EXAMPLE_LIB is set
+#
+if (!$UDF_EXAMPLE_LIB) {
+  --skip UDF requires the environment variable \$UDF_EXAMPLE_LIB to be set (normally done by mtr)
+}
+
+--let MYSQL_BASEDIR= `select @@basedir`
+
+--echo ###
+--echo # Test for loading two plugins both with full path
+--echo ###
+--let $restart_parameters= restart:$EXAMPLE_PLUGIN_LOAD_PATH $PLUGIN_AUTH_LOAD_ADD_PATH
+--source include/restart_mysqld.inc
+SELECT COUNT(*) FROM INFORMATION_SCHEMA.PLUGINS WHERE PLUGIN_NAME = 'EXAMPLE';
+SELECT COUNT(*) FROM INFORMATION_SCHEMA.PLUGINS WHERE PLUGIN_NAME = 'test_plugin_server';
+SELECT COUNT(*) FROM INFORMATION_SCHEMA.PLUGINS WHERE PLUGIN_NAME = 'cleartext_plugin_server';
+
+--echo ###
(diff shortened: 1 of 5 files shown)


mysql-test/t/plugin-load-add-with-path.test
```diff
@@ -0,0 +1,107 @@
+# Test for ability to use path in --plugin-load and --plugin-load-add options
+
+#
+# Check if server has support for loading plugins
+#
+if (`SELECT @@have_dynamic_loading != 'YES'`) {
```
laurynas-biveinis: include/have_dynamic_loading.inc

mysql-test/t/plugin-load-add-with-path.test
```diff
@@ -0,0 +1,107 @@
...
+  --skip This test requires dynamic loading
+}
+
+#
+# Check if the variable EXAMPLE_PLUGIN is set
```
laurynas-biveinis, line 11: have_example_plugin.inc

mysql-test/t/plugin-load-add-with-path.test
```diff
@@ -0,0 +1,107 @@
...
+#
+if (!$EXAMPLE_PLUGIN) {
+  --skip Example plugin requires the environment variable \$EXAMPLE_PLUGIN to be set (normally done by mtr)
+}
+
+#
+# Check if the variable PLUGIN_AUTH is set
```
laurynas-biveinis, line 18: have_plugin_auth.inc (the test is different but checks the same plugin)

mysql-test/t/plugin-load-add-with-path.test
```diff
@@ -0,0 +1,107 @@
...
+#
+if (!$PLUGIN_AUTH) {
+  --skip Example plugin requires the environment variable \$PLUGIN_AUTH to be set (normally done by mtr)
+}
+
+#
+# Check if the variable UDF_EXAMPLE_LIB is set
```
laurynas-biveinis, line 25: have_udf.inc

sql/sql_plugin.cc
```diff
@@ -451,42 +451,68 @@ static inline void free_plugin_mem(struct st_plugin_dl *p)
...
-  if (check_valid_path(dl->str, dl->length) ||
-      check_string_char_length((LEX_STRING *) dl, "", NAME_CHAR_LEN,
-                               system_charset_info, 1) ||
-      plugin_dir_len + dl->length + 1 >= FN_REFLEN)
+
+  if (allow_dl_path)
+    dl_fname_pos = dirname_length(dl->str);
+
+  if (!dl_fname_pos) {
+    plugin_dir_len= strlen(opt_plugin_dir);
```
laurynas-biveinis: A bit easier to reason about this patch if plugin_dir_len becomes limited in scope (move here, make const size_t)

laurynas-biveinis (changes requested):
Needs a blueprint.
What happens if relative path is passed?

laurynas-biveinis:
@Oneiroi, FYI, we are about to allow absolute paths in --plugin-load and --plugin-load-add server options. This looks like a safe change to me, do you agree?

mysql-test/t/plugin-load-add-with-path.test
vlad-lesin: fixed

mysql-test/t/plugin-load-add-with-path.test
vlad-lesin: It does not suite because have_example_plugin.inc checks if plugin_var variable contains the plugin dir, what is wrong for our case. The correspondent comment is added in the *.test file.

mysql-test/t/plugin-load-add-with-path.test
vlad-lesin: See above.

mysql-test/t/plugin-load-add-with-path.test
vlad-lesin: see above

sql/sql_plugin.cc
vlad-lesin: fixed

vlad-lesin:
> Needs a blueprint.

https://blueprints.launchpad.net/percona-server/+spec/pathname+in+--plugin-load+and+--plugin-load-add

> What happens if relative path is passed?

dlopen() can accept relative path so it will work as well as absolute path.

laurynas-biveinis (approved):

laurynas-biveinis:
Is a new Jenkins runs needed?

PR #1236, opened 2016-12-09
Add the ability to use path in --plugin-load and --plugin-load-add server command-line options. 5.7

vlad-lesin:
See 5.6 description here https://github.com/percona/percona-server/pull/1235 .

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/540/

The following tests:

MySQL.perfschema.show_plugin
MySQL.perfschema.show_plugin_56

look suspicious but they fail without this fix.

### DIFF
--- a/mysql-test/include/plugin.defs
+++ b/mysql-test/include/plugin.defs
@@ -10,6 +10,10 @@
 # PLUGVAR_OPT: mysqld option --plugin_dir=....
 # PLUGVAR_LOAD: option --plugin_load=.... if the 4th element is present
 # PLUGVAR_LOAD_ADD: option --plugin_load_add=.... if the 4th element is present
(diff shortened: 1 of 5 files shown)


laurynas-biveinis (changes requested):

laurynas-biveinis:
Is a new Jenkins run needed? The blueprint needs a 5.7 version

vlad-lesin:
> Is a new Jenkins run needed?

I think there is no need in new Jenkins run because the changes should not influence code execution, it's just some new comments in test and changing scope of local variable in the code.

>The blueprint needs a 5.7 version

https://blueprints.launchpad.net/percona-server/+spec/pathname+in+--plugin-load+and+--plugin-load-add-5.7

PR #1237, opened 2016-12-09
Fix bug 1648718 (Testcase rpl.rpl_get_master_version_and_clock almost…

laurynas-biveinis:
… completely broken)

Fix the testcase:
- fix the debug sync point names to actually match what's in the
  source code (mostly s/debug_lock/dbug);
- remove extra quotes from dbug_sync_point variable sets, as they
  are redundant and make all the matches fail;
- for starting a slave with a debug sync point set, replace the
  --source include/start_slave.inc with a plain START SLAVE and a
  debug_sync wait. Make the testcase debug sync points signal when
  they are parked for this;
- remove redundant MTR warning suppressions.

http://jenkins.percona.com/job/percona-server-5.6-param/1518/


percona-ysorokin (approved):
LGTM

PR #1239, opened 2016-12-09
Try to fix bug 1648737 (Testcase opt_trace.temp_table is unstable)

laurynas-biveinis:
Table t1 is InnoDB table queried right after filling it. This creates
a race condition with background stats thread, which may result in
query plan/cost instability. Fix this by adding an ANALYZE TABLE
statement.

http://jenkins.percona.com/job/percona-server-5.6-param/1522/


percona-ysorokin (approved):
LGTM

PR #1241, opened 2016-12-12
Fix lp:1640810 "Adding COMPRESSED attributes to InnoDB special tables fields can lead to server crashes".

percona-ysorokin:
'dict_table_schema_check()' extended with additional logic which guarantees that there is no difference between 'COMPRESSED' attributes in column definitions.

'innodb.xtradb_compressed_columns' MTR test case extended with checks for adding 'COMPRESSED' attribute to 'stat_description' field in 'mysql.innodb_index_stats'.


percona-ysorokin:
Jenkins build link
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1527/

laurynas-biveinis (approved):

PR #1243, opened 2016-12-15
Log all statements in transaction with same id, lp:897715

ihanick:
https://bugs.launchpad.net/percona-server/+bug/897715
http://jenkins.percona.com/job/percona-server-5.6-param/

I have considered to use thread_id + statement start time as a uniq identifier.
Only explicit transactions are tracked.
I'm not sure if additional flag is required for log_slow_verbosity or existing one (e.g. microtime) could be used.
It could be also implemented for 5.5 in the same way.

This kind of functionality will be useful with:
- find a transactions caused  metadata lock
- find whole transaction (including non-innodb statements at transaction start) by thread id (from processlist) or innodb transaction id (from innodb status).
- find incorrect transactions usage (set auto_commit = 0, connection pool on application side, some application threads using transactions and some expecting auto_commit=1).

Mysql 5.7 has thd->get_transaction()->sequence_number but it always empty.

### DIFF
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -2083,7 +2083,7 @@ bool MYSQL_QUERY_LOG::write(THD *thd, ulonglong current_utime,
     thd->profiling.print_current(&log_file);
 #endif
     if ((thd->variables.log_slow_verbosity & (1ULL << SLOG_V_INNODB))
-        && thd->innodb_was_used)
+        && thd->innodb_trx_id)
     {
       char buf[20];
       snprintf(buf, 20, "%llX", thd->innodb_trx_id);
--- a/mysql-test/r/percona_log_slow_verbosity.result
+++ b/mysql-test/r/percona_log_slow_verbosity.result
@@ -5,30 +5,33 @@ SET SESSION long_query_time=0;
 SET SESSION log_slow_verbosity='microtime,innodb,query_plan';
 [log_start.inc] percona.slow_extended.log_slow_verbosity_0
 INSERT INTO t1 VALUE(0);
+BEGIN;
+INSERT INTO t1 VALUE(1);
+ROLLBACK;
 [log_stop.inc] percona.slow_extended.log_slow_verbosity_0
 log_slow_verbosity='microtime,innodb,query_plan':
 [log_grep.inc] file: percona.slow_extended.log_slow_verbosity_0 pattern: ^# Schema: .+  Last_errno: \d+  Killed: \d+$
-[log_grep.inc] lines:   2
+[log_grep.inc] lines:   5
 [log_grep.inc] file: percona.slow_extended.log_slow_verbosity_0 pattern: ^#.*Rows_affected: \d+$
-[log_grep.inc] lines:   2
+[log_grep.inc] lines:   5
 [log_grep.inc] file: percona.slow_extended.log_slow_verbosity_0 pattern: ^# Bytes_sent: \d+.*$
-[log_grep.inc] lines:   2
+[log_grep.inc] lines:   5
 [log_grep.inc] file: percona.slow_extended.log_slow_verbosity_0 pattern: ^# InnoDB_trx_id: \w+$
-[log_grep.inc] lines:   1
+[log_grep.inc] lines:   3
 [log_grep.inc] file: percona.slow_extended.log_slow_verbosity_0 pattern: ^# Bytes_sent: \d+  Tmp_tables: \d+  Tmp_disk_tables: \d+  Tmp_table_sizes: \d+$
-[log_grep.inc] lines:   2
+[log_grep.inc] lines:   5
 [log_grep.inc] file: percona.slow_extended.log_slow_verbosity_0 pattern: ^# QC_Hit: (Yes|No)  Full_scan: (Yes|No)  Full_join: (Yes|No)  Tmp_table: (Yes|No)  Tmp_table_on_disk: (Yes|No)$
-[log_grep.inc] lines:   2
+[log_grep.inc] lines:   5
 [log_grep.inc] file: percona.slow_extended.log_slow_verbosity_0 pattern: ^# Filesort: (Yes|No)  Filesort_on_disk: (Yes|No)  Merge_passes: \d+$
-[log_grep.inc] lines:   2
+[log_grep.inc] lines:   5
 [log_grep.inc] file: percona.slow_extended.log_slow_verbosity_0 pattern: ^#   InnoDB_IO_r_ops: \d+  InnoDB_IO_r_bytes: \d+  InnoDB_IO_r_wait: \d*\.\d*$
-[log_grep.inc] lines:   1
+[log_grep.inc] lines:   2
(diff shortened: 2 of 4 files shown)


sql/log.cc
```diff
@@ -2082,6 +2082,18 @@ bool MYSQL_QUERY_LOG::write(THD *thd, ulonglong current_utime,
 #if defined(ENABLED_PROFILING)
     thd->profiling.print_current(&log_file);
 #endif
+    if ((thd->variables.log_slow_verbosity & (1ULL << SLOG_V_MICROTIME))
```
laurynas-biveinis: Why is this attached to microtime verbosity?

sql/log.cc
```diff
@@ -2082,6 +2082,18 @@ bool MYSQL_QUERY_LOG::write(THD *thd, ulonglong current_utime,
...
+        && thd->trx_start_utime)
+    {
+        char buf[30];
+        snprintf(buf, 30, "%lX%llX", thd->thread_id, thd->trx_start_utime);
+        if (!(thd->server_status & SERVER_STATUS_IN_TRANS))
+            thd->trx_start_utime= 0;
+        if (my_b_printf(&log_file,
+                    "# Trx_id: %s\n", buf) == (uint) -1)
```
laurynas-biveinis: We end up with two trx ids, innodb one, and thread id. Probably it's better to implement the InnoDB trx id for "COMMIT", and nothing for "BEGIN" (maybe in some cases BEGIN gets InnoDB trx id - what about "WITH CONSISTENT SNAPSHOT"?)

laurynas-biveinis (changes requested):

ihanick:
http://jenkins.percona.com/job/percona-server-5.6-param/1535/
- changed transaction id to innodb_trx_id
- now transaction tracked after any innodb statement inside transaction until commit or rollback.

both timestamp based and Innodb_trx_id producing different numbers for base session and for consistent snapshot.

sql/sql_class.h
```diff
@@ -2431,6 +2431,7 @@ class THD :public MDL_context_owner,
     default values only if (innodb_was_used==TRUE)
   */
   ulonglong  innodb_trx_id;
+  ulonglong  saved_innodb_trx_id;
```
laurynas-biveinis: Why is this needed?

laurynas-biveinis: Assuming it is needed, then it probably needs some sort of initialisation/reset in THD or it looks like some of the reads will be uninitialized

laurynas-biveinis (changes requested):
Can you provide an example of a slow query log with changed output?

ihanick:
http://jenkins.percona.com/job/percona-server-5.6-param/1541/

- clear_slow_extended clears innodb_trx_id only if not running inside transaction
- during init innodb_trx_id always cleared, because server_status= SERVER_STATUS_AUTOCOMMIT; (no SERVER_STATUS_IN_TRANS).
- remove save_innodb_trx_id variable

Test queries:
```sql
create table t(c int) engine=innodb;
select 1;
begin;
  select 2;
  select * from t;
  select 3;
commit;
select 4;
```
Slow log example
```
# User@Host: root[root] @ localhost []  Id:     2
# Schema: test  Last_errno: 0  Killed: 0
# Query_time: 0.000131  Lock_time: 0.000000  Rows_sent: 1  Rows_examined: 0  Rows_affected: 0
# Bytes_sent: 56
# No InnoDB statistics available for this query
SET timestamp=1481896052;
select 1;
# User@Host: root[root] @ localhost []  Id:     2
# Schema: test  Last_errno: 0  Killed: 0
# Query_time: 0.000047  Lock_time: 0.000000  Rows_sent: 0  Rows_examined: 0  Rows_affected: 0
# Bytes_sent: 11
# No InnoDB statistics available for this query
SET timestamp=1481896052;
begin;
# User@Host: root[root] @ localhost []  Id:     2
# Schema: test  Last_errno: 0  Killed: 0
# Query_time: 0.000071  Lock_time: 0.000000  Rows_sent: 1  Rows_examined: 0  Rows_affected: 0
# Bytes_sent: 56
# No InnoDB statistics available for this query
SET timestamp=1481896052;
select 2;
# User@Host: root[root] @ localhost []  Id:     2
# Schema: test  Last_errno: 0  Killed: 0
# Query_time: 0.001138  Lock_time: 0.000909  Rows_sent: 0  Rows_examined: 0  Rows_affected: 0
# Bytes_sent: 57
# InnoDB_trx_id: 509
#   InnoDB_IO_r_ops: 8  InnoDB_IO_r_bytes: 131072  InnoDB_IO_r_wait: 0.000142
#   InnoDB_rec_lock_wait: 0.000000  InnoDB_queue_wait: 0.000000
#   InnoDB_pages_distinct: 22
SET timestamp=1481896052;
select * from t;
# User@Host: root[root] @ localhost []  Id:     2
# Schema: test  Last_errno: 0  Killed: 0
# Query_time: 0.000158  Lock_time: 0.000000  Rows_sent: 1  Rows_examined: 0  Rows_affected: 0
# Bytes_sent: 56
# InnoDB_trx_id: 509
# No InnoDB statistics available for this query
SET timestamp=1481896052;
select 3;
# User@Host: root[root] @ localhost []  Id:     2
# Schema: test  Last_errno: 0  Killed: 0
# Query_time: 0.000104  Lock_time: 0.000000  Rows_sent: 0  Rows_examined: 0  Rows_affected: 0
# Bytes_sent: 11
# InnoDB_trx_id: 509
# No InnoDB statistics available for this query
SET timestamp=1481896052;
commit;
# User@Host: root[root] @ localhost []  Id:     2
# Schema: test  Last_errno: 0  Killed: 0
# Query_time: 0.000684  Lock_time: 0.000000  Rows_sent: 1  Rows_examined: 0  Rows_affected: 0
# Bytes_sent: 56
# No InnoDB statistics available for this query
SET timestamp=1481896052;
select 4;
# User@Host: root[root] @ localhost []  Id:     2
# Schema: test  Last_errno: 0  Killed: 0
# Query_time: 0.000013  Lock_time: 0.000000  Rows_sent: 0  Rows_examined: 0  Rows_affected: 0
# Bytes_sent: 0
# No InnoDB statistics available for this query
SET timestamp=1481896052;
```

sql/log.cc
```diff
@@ -2083,7 +2083,7 @@ bool MYSQL_QUERY_LOG::write(THD *thd, ulonglong current_utime,
     thd->profiling.print_current(&log_file);
 #endif
     if ((thd->variables.log_slow_verbosity & (1ULL << SLOG_V_INNODB))
-        && thd->innodb_was_used)
+        && (thd->innodb_was_used || thd->innodb_trx_id))
```
laurynas-biveinis: Can we remove thd->innodb_was_used check?

laurynas-biveinis (changes requested):

ihanick:
http://jenkins.percona.com/job/percona-server-5.6-param/1579/

- removed redundant thd->innodb_was_use check:
thd->innodb_trx_id is zero if not in transaction in clear_slow_extended, called from THD::init and before running command at dispatch_command.

laurynas-biveinis (approved):

PR #1260, opened 2016-12-22
Fix bug 1625466 (INFORMATION_SCHEMA.INNODB_CHANGED_PAGES queries need…

laurynas-biveinis:
…lessly read potentially incomplete bitmap data past the query range)

The bitmap read iterator reads one more bitmap page past the end of
the range to compare its start LSN to the range end LSN to stop the
iteration. If the extra bitmap page happens to be corrupted, the
INFORMATION_SCHEMA.INNODB_CHANGED_PAGES query is needlessly failed
with ER_CANT_FIND_SYSTEM_REC. Change the iteration stop condition to
check the end LSN and the last-page-in-run flag at the end of the page
and stop iteration there.

At the same time fix bug 1651121 (Test
innodb.percona_changed_page_bmp_debug unstable under load)

An assert added for bug 1625466 - that if a 2nd bitmap page write
failure is enabled, then the 2nd page must always exist - is too
strict, as it is possible for the log tracker to do a single-page
write before reaching the two-page data. Replace the assert with an
if.

At the same time cut the testcase runtime by replacing the log-filling
workload by debug variable settings forcing the same effect of
guaranteed log tracking.

http://jenkins.percona.com/job/percona-server-5.6-param/1547/
http://jenkins.percona.com/job/percona-server-5.6-valgrind/232/


storage/innobase/log/log0online.cc
```diff
@@ -1710,6 +1714,9 @@ log_online_bitmap_iterator_next(
 		return TRUE;
 	}
 
+	if (i->end_lsn >= i->max_lsn && i->last_page_in_run)
```
percona-ysorokin, line 1717: UNIV_UNLIKELY?

percona-ysorokin (approved):
LGTM
whether to ignore the comment is on your choice

storage/innobase/log/log0online.cc
laurynas-biveinis: Not sure - at this point we are past iteration over individual bits stage, we either are finished, either need to read the next page.

PR #1262, opened 2016-12-25
Fix #1651657, failback from backup for legacy for small BP instance size

ihanick:
https://bugs.launchpad.net/percona-server/+bug/1651657
http://jenkins.percona.com/job/percona-server-5.6-param/1560/

Small buffer pool instance size and backoff algorithm causing deadlock, failback to legacy algorithm in such cases.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -701,6 +701,32 @@ innobase_is_fake_change(
 	THD*		thd);	/*!< in: MySQL thread handle of the user for
 				  whom the transaction is being committed */
 
+/** Empty free list algorithm.
+Checks if buffer pool is big enough to enable backoff algorithm.
+InnoDB empty free list algorithm backoff requires free pages
+from LRU for the best performance.
+buf_LRU_buf_pool_running_out cancels query if 1/4 of
+buffer pool belongs to LRU or freelist.
+At the same time buf_flush_LRU_list_batch
+keeps up to BUF_LRU_MIN_LEN in LRU.
+In order to avoid deadlock baclkoff requires buffer pool
+to be at least 4*BUF_LRU_MIN_LEN,
+but flush peformance is bad because of trashing
+and additional BUF_LRU_MIN_LEN pages are requested.
+@param[in]	algorithm	desired algorithm from srv_empty_free_list_t
+@return	true if it's possible to enable backoff. */
+static inline
+bool
+innodb_empty_free_list_algorithm_allowed(
+	srv_empty_free_list_t	algorithm)
+{
+	long long buf_pool_pages = srv_buf_pool_size / srv_page_size
+				/ srv_buf_pool_instances;
+
+	return(buf_pool_pages >= BUF_LRU_MIN_LEN * (4 + 1)
+			|| algorithm != SRV_EMPTY_FREE_LIST_BACKOFF);
+}
+
 /** Get the list of foreign keys referencing a specified table
 table.
 @param thd		The thread handle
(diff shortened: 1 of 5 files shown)


mysql-test/t/percona_bug1651657.test
```diff
@@ -0,0 +1,11 @@
+--source include/have_innodb.inc
+--let $assert_text= Small buffer pool instance should use legacy
```
laurynas-biveinis: s/instance/instances

mysql-test/t/percona_bug1651657.test
```diff
@@ -0,0 +1,11 @@
...
+--let $assert_cond= @@innodb_empty_free_list_algorithm = "legacy"
+--source include/assert.inc
+
+call mtr.add_suppression("InnoDB: innodb_empty_free_list_algorithm = 'backoff' requires at least 20MB buffer pool instances.");
+--error 1231
```
laurynas-biveinis: Please use ER_ mnemonics

mysql-test/t/percona_bug1651657.test
```diff
@@ -0,0 +1,11 @@
+--source include/have_innodb.inc
```
laurynas-biveinis: Please move the testcase to the innodb suite

storage/innobase/handler/ha_innodb.cc
```diff
@@ -3880,7 +3880,8 @@ innobase_init(
 	if (!innodb_empty_free_list_algorithm_backoff_allowed(
```
laurynas-biveinis: There are some code smells in innodb_empty_free_list_algorithm_backoff_allowed: 1) the name hardcodes that it is  "backoff", but the 1st arg passes the actual algorithm to verify; 2) the calculation innobase_buffer_pool_size / srv_page_size / innobase_buffer_pool_instance is the only correct way to pass the 2nd arg.

Thus:
- Please rename to innodb_empty_free_list_algorithm_allowed
- Please inline and remove the 2nd arg

laurynas-biveinis (changes requested):
Please also check the Jenkins run for sys_vars.innodb_empty_free_list_algorithm_basic failure

ihanick:
http://jenkins.percona.com/job/percona-server-5.6-param/1567/
- renamed to innodb_empty_free_list_algorithm_allowed
- moved number of pages in buffer pool calculation to this function
- using new InnoDB header comment style
- moved function declaration slightly above to avoid merge conflict
- using --error ER_WRONG_VALUE_FOR_VAR in test
- test now inside innodb suite
- there is no limitation for minimal buffer pool size to have several instances (looks like a bug, because instance could be smaller than minimum single buffer pool size),  forced sys_vars.innodb_empty_free_list_algorithm_basic to use single buffer pool instance.
- changed innodb buffer pool size and number of instances to the same value as in 5.7-ps-bug1651657

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1346,6 +1346,24 @@ innodb_enable_monitor_at_startup(
 /*=============================*/
 	char*	str);	/*!< in: monitor counter enable list */
 
+/** Checks if buffer pool is big enough to enable backoff algorithm.
```
laurynas-biveinis: "Check if enabling given empty free list algorithm is allowed"

The header comment slightly differs between this declaration and the actual definition. I'd just move definition higher in the file so that there is no need for the declaration

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17178,28 +17173,29 @@ innodb_status_output_update(
...
 At the same time buf_flush_LRU_list_batch
 keeps up to BUF_LRU_MIN_LEN in LRU.
 In order to avoid deadlock baclkoff requires buffer pool
 to be at least 4*BUF_LRU_MIN_LEN,
 but flush peformance is bad because of trashing
 and additional BUF_LRU_MIN_LEN pages are requested.
+@param[in]	algorithm	desired algorithm from srv_empty_free_list_t
 @return	true if it's possible to enable backoff. */
-static
+static inline
```
laurynas-biveinis: Just static

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17178,28 +17173,29 @@ innodb_status_output_update(
...
-innodb_empty_free_list_algorithm_backoff_allowed(
-	srv_empty_free_list_t	algorithm,	/*!< in: desired algorithm
-						from srv_empty_free_list_t */
-	long long		buf_pool_pages)	/*!< in: total number
-						of pages inside buffer pool */
+innodb_empty_free_list_algorithm_allowed(
+	srv_empty_free_list_t	algorithm)
 {
+	long long buf_pool_pages = srv_buf_pool_size / srv_page_size
+				/ (srv_buf_pool_instances > 1 ?
```
laurynas-biveinis: Why the ternary operator here?

laurynas-biveinis (changes requested):

ihanick:
http://jenkins.percona.com/job/percona-server-5.6-param/1706/
- removed declaration for innodb_empty_free_list_algorithm_backoff_allowed
- removed  ternary operator from buf_pool_pages calculation
- removed "inline"

laurynas-biveinis (approved):

PR #1264, opened 2016-12-28
Make tokudb hotbackup tests work from build dir.

vlad-lesin:
Use path in plugin-load(-add) option to make tokudb hotbackup tests work from
build dir.

http://jenkins.percona.com/job/percona-server-5.6-thb-param/10/

have_tokudb.inc and have_hotbackup.inc are fixed too because if plugin_dir does not contain some plugin directory it does not mean that this plugin can't be loaded with path in --plugin-load(-add) options.

The conflict is in "plugin.defs" file because it contains entry for rocksdb while this branch does not contain this entry. It can be fixed manually.


laurynas-biveinis (approved):

laurynas-biveinis:
This might have caused https://bugs.launchpad.net/percona-server/+bug/1654501. @vlad-lesin, can you please check?

PR #1268, opened 2016-12-30
Fix bug 1651656 (Server crash if a changed page bitmap error occurs c…

laurynas-biveinis:
…oncurrently with executing FLUSH CHANGED_PAGE_BITMAPS)

If a bitmap write I/O errors happens in the background log tracking
thread while a FLUSH CHANGED_PAGE_BITMAPS is executing concurrently,
the two threads enter a race condition where the background thread
frees the mutex while the FLUSH locks/unlocks it, resulting in a
crash. Fix by pulling up the log_bmp_sys_mutex out of the dynamic
log_bmp_sys structure, and making it available throught server
lifetime, regardless of the log tracking startup or current
setting. Add new functions log_online_init and log_online_shutdown to
create and destroy the mutex. Make log_online_read_shutdown take that
mutex around log_bmp_sys deinitialisation and also protect
srv_track_changed_pages and srv_redo_log_thread_started true to false
transitions with it. Make log_online_follow_redo_log check
srv_track_changed_pages, take the mutex, and re-check the variable
before continuing operation. Likewise make
log_online_purge_changed_page_bitmaps check
srv_redo_log_thread_started, take the mutex, and re-check the variable.

At the same time replace mutex_own asserts with a helper function
in_log_bmp_sys_with_mutex that also asserts enabled log tracking
status var. Cleanup functions a bit by moving variable declarations to
initialisation statements.

At the same time remove another testcase instability by making the
bitmap_page_write_error error injection point only trigger for bitmap
data pages with tablespace ID > 0, that is, user tablespaces, as tested
by the testcases.

http://jenkins.percona.com/job/percona-server-5.6-param/1574/

### DIFF
--- a/mysql-test/suite/innodb/r/percona_changed_page_bmp_debug.result
+++ b/mysql-test/suite/innodb/r/percona_changed_page_bmp_debug.result
@@ -46,6 +46,7 @@ SET DEBUG_SYNC="RESET";
 DROP TABLE t2;
 RESET CHANGED_PAGE_BITMAPS;
 4th restart
+5th restart
 SET DEBUG_SYNC="i_s_innodb_changed_pages_range_ready SIGNAL ready WAIT_FOR finish";
 SELECT COUNT(*) FROM INFORMATION_SCHEMA.INNODB_CHANGED_PAGES;
 SET DEBUG_SYNC="now WAIT_FOR ready";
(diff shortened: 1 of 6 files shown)


percona-ysorokin (approved):
LGTM

PR #1273, opened 2017-01-05
Fix SET STATEMENT ... FOR bug 1635927 and bug 1392375.

laurynas-biveinis:
Bug 1392375 is that repeated execution of SET STATEMENT ... FOR
<SELECT FROM view> crashes and bug 1635927 is a memory leak when using
per-query variables with subquery temp tables.

Both bugs share the root cause which is that SET STATEMENT query
variables were set with a sql_set_variables call before main query
execution. This function frees subselect joins, if any, at its end,
which is premature as they are needed for main query execution. As
the subselect lex object is marked as already freed, any subsequent
free attempts will be no-ops, resulting in either a memory leak or a
crash that the lex already has a join associated. Fix by not freeing
subselect joins for SET STATEMENT variable setting part.

At the same time add invariant asserts to st_select_lex_unit methods
that the object is in cleaned state iff it has no associated joins.

http://jenkins.percona.com/job/percona-server-5.6-param/1577/

### DIFF
--- a/mysql-test/r/percona_statement_set.result
+++ b/mysql-test/r/percona_statement_set.result
@@ -879,6 +879,34 @@ HANDLER t3 OPEN;
 SET STATEMENT max_join_size=1000 FOR SELECT * FROM t3;
 ERROR HY000: Can't reopen table: 't3'
 DROP TABLE t3;
+#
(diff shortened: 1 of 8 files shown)


percona-ysorokin (approved):
LGTM

PR #1275, opened 2017-01-05
Fix the bug of renaming to non-existen db.

vlad-lesin:
It is supposed db exists if mysql data dir contains subdirectory with the same
name as the db. Before calling rename operation from storage engine it is
checked if corresponding subdirectory exists in mysql data directory.

But this check is not enough in the case if tokudb data directory is not
the same as mysql data directory. It is also supposed PerconaFT library does
not handle directory operations because the library should not know about
databases, tables etc, it's just library for work with fractal trees. And
caller should manage directories, in our case caller is mysqld. But mysqld can
work with directories only in it's data directory and it does not create or
remove db directory in tokudb data directory if it's not the same as mysql
data directory. That is why PerconaFT is also fixed to create necessary
directories for fractal tree file path like InnoDB. But There is no
transactional directory operations in PerconaFT.

See also: https://github.com/percona/percona-server/pull/1276, https://github.com/percona/PerconaFT/pull/365.

http://jenkins.percona.com/job/percona-server-5.6-param/1581/

### DIFF
--- a/include/my_base.h
+++ b/include/my_base.h
@@ -488,7 +488,8 @@ is the global server default. */
 #define HA_ERR_TEMP_FILE_WRITE_FAILURE	189	/* Temporary file write failure */
 #define HA_ERR_INNODB_FORCED_RECOVERY 190	/* Innodb is in force recovery mode */
 #define HA_ERR_FTS_TOO_MANY_WORDS_IN_PHRASE	191 /* Too many words in a phrase */
-#define HA_ERR_LAST               191    /* Copy of last error nr */
+#define HA_ERR_DEST_SCHEMA_NOT_EXIST   192  /* Destination schema does not exist */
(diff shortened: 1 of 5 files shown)


george-lorch:
Haven't fully reviewed yet but just noting that https://github.com/percona/PerconaFT/pull/365 MUST be merged first AND the PerconaFT git submodule commit pointer MUST be moved prior to merging this change.

george-lorch:
Merged here https://github.com/percona/percona-server/commit/78afddb0cceb3f2907181017c210f7eab95f2b20 where I advanced the git submodule commit pointer for PerconaFT to https://github.com/percona/PerconaFT/commit/f56b7e86b78d529563e57290bf0b54e121c2684c to capture the PerconaFT portion of the issue.

PR #1286, opened 2017-01-11
Add diagnostics to catch bug 1628079 (Intemittent hangs on shutdown, …

laurynas-biveinis:
…ASan build)

Diagnose kill syscall failing in mysqltest with anything else than
ESRCH.

http://jenkins.percona.com/job/percona-server-5.6-param/1586/


percona-ysorokin (approved):
LGTM

PR #1288, opened 2017-01-12
Fix bug 1655587 / 84381 (main.events_2 test fails because of hardcode…

laurynas-biveinis:
…d 2017-01-01 date)

Replace the hardcoded date with one at one year in the future from
NOW().

http://jenkins.percona.com/job/percona-server-5.5-param/1507/


percona-ysorokin (approved):
LGTM

PR #1291, opened 2017-01-17
Fix bug 1657026 (Lack of free pages in the buffer pool is not diagnos…

laurynas-biveinis:
…ed with innodb_empty_free_list_algorithm=backoff)

The introduction of backoff to buf_LRU_get_free_block made it skip some
of the code paths of the legacy case, including diagnostics in the error
log for the lack of free pages, enabling of the InnoDB monitor, and
Innodb_buffer_pool_wait_free status variable bump. Fix by factoring out a
helper function to print diagnostics and enable the InnoDB monitor. Bump
the status variable directly in buf_LRU_get_free_block. Add a simulated
testcase. For diagnostics, also throttle their printing to one around
every two seconds.

Also, as a related change, add diagnostics to help catch bug 1655657
(Test innodb_fts.innodb_fts_misc_debug is unstable)

If crash recovery fails to find free buffer pool pages for read, it
loops forever in buf_LRU_get_free_block waiting for a block to appear
and attempting to print the InnoDB status to the server error
log. This never happens, since the server monitor thread has not been
started yet at the crash recovery time. Thus move the thread start
earlier, and patch the status printer to skip over the parts that are
not initialized yet. Add a simulated testcase.

http://jenkins.percona.com/job/percona-server-5.6-param/1593/

### DIFF
--- a/mysql-test/suite/innodb/t/percona_crash_recovery_innodb_status.test
+++ b/mysql-test/suite/innodb/t/percona_crash_recovery_innodb_status.test
@@ -1,0 +1,29 @@
+--source include/have_debug.inc
+--source include/have_innodb.inc
+--source include/not_embedded.inc
+
+--echo #
+--echo # Test printing of InnoDB status during crash recovery
+--echo #
+
+--let SEARCH_FILE= $MYSQLTEST_VARDIR/log/my_restart.err
+
+CREATE TABLE t1 (a INT PRIMARY KEY) ENGINE=InnoDB;
+
+BEGIN;
+INSERT INTO t1 VALUES (1);
+
+--source include/kill_mysqld.inc
+--error 137,3
+--exec $MYSQLD_CMD --loose-console --debug=+d,simulate_recovery_lack_of_pages > $SEARCH_FILE 2>&1
+
+--let SEARCH_PATTERN= difficult to find free blocks
+--source include/search_pattern_in_file.inc
+
+--let SEARCH_PATTERN=END OF INNODB MONITOR OUTPUT
+--source include/search_pattern_in_file.inc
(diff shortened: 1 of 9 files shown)


mysql-test/suite/innodb/t/percona_lru_flusher_debug.test
```diff
@@ -0,0 +1,36 @@
+--source include/have_debug.inc
+--source include/have_innodb.inc
+
```
percona-ysorokin, line 4: --source include/not_embedded.inc

mysql-test/suite/innodb/t/percona_crash_recovery_innodb_status.test
```diff
@@ -0,0 +1,29 @@
...
+
+--let SEARCH_FILE= $MYSQLTEST_VARDIR/log/my_restart.err
+
+CREATE TABLE t1 (a INT PRIMARY KEY) ENGINE=InnoDB;
+
+BEGIN;
+INSERT INTO t1 VALUES (1);
+
+--source include/kill_mysqld.inc
+--error 137
```
percona-ysorokin: From `log_file_size.test`
```# On Windows, DBUG_SUICIDE() does exit(3) instead of SIGKILL (128+9).```
So (just in case one day we support Windows)
`--error 137,3`

percona-ysorokin (approved):
LGTM
just minor MTR test case comments.

laurynas-biveinis:
Fixed, http://jenkins.percona.com/job/percona-server-5.6-param/1596/

percona-ysorokin (approved):
LGTM

PR #1293, opened 2017-01-18
Bug#995624 RBR events are not reflected in userstat's Rows_updated

robgolebiowski:
The Rows_updated value in the userstat feature is calculated at the statement level, i.e. in mysql_insert()/mysql_delete()/mysql_update(). However, row-based replic$
tion event operate on the lower handler level, so that value is not updated with RBR as a result.


laurynas-biveinis (changes requested):
IMHO this bug is not important enough to warrant a 5.5 fix, it should be fixed in 5.6+ only

laurynas-biveinis:
This is the PR should have been closed instead

PR #1294, opened 2017-01-18
5.6 ps bug995624

robgolebiowski:

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_percona_bug995624.test
+++ b/mysql-test/suite/rpl/t/rpl_percona_bug995624.test
@@ -1,0 +1,46 @@
+source include/master-slave.inc;
+source include/not_gtid_enabled.inc;
+
+connection slave;
+set global userstat=ON;
+--echo Rows_updated on slave - starting situation
+select ROWS_UPDATED from information_schema.client_statistics;
+
+connection master;
+set global userstat=ON;
+--echo Rows_updated on master - starting situation
+select ROWS_UPDATED from information_schema.client_statistics;
+
+create table t1 (m int);
+--echo 3 "updates" to rows
+insert into t1 values(15),(16),(17); 
+--echo 1 "update" to rows (4 in total)
+update t1 set m=20 where m=16;
+--echo 1 "update" to rows (5 in total)
+delete from t1 where m=17;
+create table t2 (n int);
+--echo 2 "updates" to rows (7 in total)
+insert into t2 values(30),(30);
+--echo 2 "updates" to rows (9 in total)
+update t2 set n=10 where n=30;
+--echo 2 "updates" to rows (11 in total)
+delete from t2 where n=10;
+
+--echo 2 "updates" to rows (13 in total)
+insert into t2 (n)
+select t1.m
+from t1;
+--echo Rows_updated on master - ending situation
+select ROWS_UPDATED from information_schema.client_statistics;
+set global userstat=OFF;
+
+--source include/sync_slave_sql_with_master.inc
+--echo Rows_updated on slave - ending situation
+select ROWS_UPDATED from information_schema.client_statistics;
+set global userstat=OFF;
+select * from t1 ORDER BY m;
+
+connection master;
+drop table t1, t2;
+--source include/sync_slave_sql_with_master.inc
+--source include/rpl_end.inc
--- a/sql/log_event.cc
+++ b/sql/log_event.cc
@@ -59,6 +59,9 @@ slave_ignored_err_throttle(window_size,
 #include "rpl_utility.h"
 
 #include "sql_digest.h"
+#ifndef EMBEDDED_LIBRARY
+#include "sql_connect.h" //update_global_user_stats
+#endif
 
 using std::min;
 using std::max;
@@ -11554,6 +11557,9 @@ int Rows_log_event::do_apply_event(Relay_log_info const *rli)
 
       error= (this->*do_apply_row_ptr)(rli);
 
+      if (!error)
+        thd->updated_row_count++;
+
       if (handle_idempotent_and_ignored_errors(rli, &error))
         break;
 
@@ -11561,6 +11567,14 @@ int Rows_log_event::do_apply_event(Relay_log_info const *rli)
       do_post_row_operations(rli, error);
 
     } while (!error && (m_curr_row != m_rows_end));
+
+    if (unlikely(opt_userstat))
+    {
+      thd->update_stats(false);
+#ifndef EMBEDDED_LIBRARY
+      update_global_user_stats(thd, true, time(NULL));
+#endif
+    }
 
 AFTER_MAIN_EXEC_ROW_LOOP:
 
--- a/mysql-test/suite/rpl/r/rpl_percona_bug995624.result
+++ b/mysql-test/suite/rpl/r/rpl_percona_bug995624.result
@@ -1,0 +1,52 @@
+include/master-slave.inc
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
+Note	####	Storing MySQL user name or password information in the master info repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START SLAVE; see the 'START SLAVE Syntax' in the MySQL Manual for more information.
+[connection master]
+set global userstat=ON;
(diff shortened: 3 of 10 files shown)


laurynas-biveinis:
Please update https://bugs.launchpad.net/percona-server/+bug/995624 accordingly

laurynas-biveinis:
The commit message (in 5.5 but it will become the 5.6 one, replacing the current merge commit) only repeats the bug description. It should say what was the fix

sql/log_event.cc
```diff
@@ -93,6 +93,11 @@ TYPELIB binlog_checksum_typelib=
 
 #define FLAGSTR(V,F) ((V)&(F)?#F" ":"")
 
+#ifndef EMBEDDED_LIBRARY
+// Uses the THD to update the global stats by user name and client IP
+void update_global_user_stats(THD* thd, bool create_user, time_t now);
```
laurynas-biveinis: The existing patch uses this "declarations everywhere" practice instead of a header file, this has to end - please use sql_connect.h (? not sure if actually the best place) and replace this and other existing declarations with an include

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
```diff
@@ -0,0 +1,46 @@
+source include/master-slave.inc;
```
laurynas-biveinis, line 1: Please name rpl_percona_userstat, and add a
--echo #
--echo # Bug 995624 blah blah
--echo #
section. So that further replication-related userstat testcases could be added to the same file

laurynas-biveinis, line 1: For "source" MTR command we usually use --source foo/bar style

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
```diff
@@ -0,0 +1,46 @@
...
+source include/not_gtid_enabled.inc;
```
laurynas-biveinis, line 2: Why is this check required?

laurynas-biveinis, line 2: If this check is required, then the cheap testcase-skipping checks should go before the expensive testcase setup actions (master-slave.inc), or MTR will go through the trouble of starting a slave server, configuring replication, only to see that the test is going to be skipped in some cases

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
```diff
@@ -0,0 +1,46 @@
...
+
+connection slave;
+set global userstat=ON;
+--echo Rows_updated on slave - starting situation
+select ROWS_UPDATED from information_schema.client_statistics;
```
laurynas-biveinis, line 7: Please don't record this to output, we don't care about this number before the testcase start, we care about its increase. So store it instead, and check the delta in the end

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
```diff
@@ -0,0 +1,46 @@
...
+
+connection master;
+set global userstat=ON;
```
laurynas-biveinis, line 10: The bug is not about the master, so strictly speaking this is redundant. But, it could be a good idea to check that the userstat rows_updated on master and slave increase by the same amount

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
```diff
@@ -0,0 +1,46 @@
+source include/master-slave.inc;
+source include/not_gtid_enabled.inc;
+
+connection slave;
+set global userstat=ON;
```
laurynas-biveinis, line 5: Please save the existing userstat value and restore it in the end (so that MTR --mysqld=--userstat and similar work)

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
```diff
@@ -0,0 +1,46 @@
...
+select ROWS_UPDATED from information_schema.client_statistics;
+
+create table t1 (m int);
+--echo 3 "updates" to rows
+insert into t1 values(15),(16),(17); 
+--echo 1 "update" to rows (4 in total)
+update t1 set m=20 where m=16;
+--echo 1 "update" to rows (5 in total)
+delete from t1 where m=17;
+create table t2 (n int);
```
laurynas-biveinis, line 21: t2 test looks redundant - please test instead, that a table, that is not replicated by the slave-side replication filtering rules, does not cause userstat bump

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
```diff
@@ -0,0 +1,46 @@
...
+
+--source include/sync_slave_sql_with_master.inc
+--echo Rows_updated on slave - ending situation
+select ROWS_UPDATED from information_schema.client_statistics;
+set global userstat=OFF;
+select * from t1 ORDER BY m;
+
+connection master;
+drop table t1, t2;
+--source include/sync_slave_sql_with_master.inc
```
laurynas-biveinis, line 45: IIRC rpl_end will sync too?

laurynas-biveinis (changes requested):

sql/log_event.cc
robgolebiowski: Done.

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
robgolebiowski: Done.

robgolebiowski: Done.

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
robgolebiowski: This check was not required. Removed.

robgolebiowski: Not needed.

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
robgolebiowski: Done.

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
robgolebiowski: This is what I am doing, but in a slightly different way. I record number of Rows_updated on master and slave, and of course compare it to the value stored in result file - i.e. 0. Then after the inserts/updates/deletes I am checking Rows_updated on master and slave once again. The recorded value in result file is 13. It should be the same on master and slave (at least for the tables that are replicated). In other words I am checking how much Rows_updated on master and slave increased by means of comparing it with the value stored in result file, i.e. 13.

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
robgolebiowski: Please see the comment above.

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
robgolebiowski: I just wanted to check that userstat is updated on slave when I do updates not only to one table, but also to two. However, I have added a t3 table that is not replicated.

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
robgolebiowski: Right, corrected.

laurynas-biveinis:
No need to close the existing PR: git push --force to the same branch will update it

mysql-test/suite/rpl/t/rpl_percona_bug995624.test
laurynas-biveinis: I see, but then it looks like it's better suited include/assert.inc: as the testcase is maintained going forward (merged across major versions etc), relying on .result file review after any changes is less robust, possible to miss something, whereas assert.inc would fail at once.

robgolebiowski: OK, changed to assert.

PR #1295, opened 2017-01-18
5.7 ps bug995624

robgolebiowski:


laurynas-biveinis (changes requested):
Same comments as for 5.6 + if you add slave filtering test, make it use 5.7 SQL to set it up instead of 5.6 command-line options

laurynas-biveinis (approved):

percona-ysorokin (approved):
LGTM

PR #1299, opened 2017-01-19
Bug#1592694 Query plan temp flag set is missed in create_duplicate_we…

robgolebiowski:


laurynas-biveinis:
Please always update the bug status once you are working on it (assignment, 5.5/5.6/5.7 series, In Progress, Fix Committed, Fix Released, Invalid, Won't Fix etc.). For this bug 5.5 and 5.6 were New, that is, the bug was found on 5.7, and the earlier versions have not been checked yet - that needs to be checked and the earliest fix version determined

mysql-test/include/subquery_sj.inc
```diff
@@ -7142,3 +7142,41 @@ eval EXPLAIN extended $query;
 eval $query;
 
 DROP TABLE t;
+
```
laurynas-biveinis: This should go to one of percona slow query log testcases

sql/sql_class.cc
```diff
@@ -4505,6 +4505,7 @@ void THD::inc_status_created_tmp_disk_tables()
 void THD::inc_status_created_tmp_tables()
 {
   status_var.created_tmp_tables++;
+  query_plan_flags|= QPLAN_TMP_TABLE;
```
laurynas-biveinis, line 4511: After the fix, this same statement in create_tmp_table becomes redundant and should be removed

laurynas-biveinis:
The commit message should describe the fix

laurynas-biveinis:
Needs a Jenkins run

laurynas-biveinis (changes requested):

sql/sql_class.cc
robgolebiowski: Thanks a lot. I was going to removed it, but in the end as it turned out I had not.

robgolebiowski:
Jenkin's run: http://jenkins.percona.com/job/mysql-5.7-param/615/

robgolebiowski:
Still needs to check if this problem occurs on 5.5 and 5.6

laurynas-biveinis:
Looks good - please check lower versions and re-submit this patch as a merge

robgolebiowski:
This is merged from https://github.com/percona/percona-server/pull/1332. Self-approving.

laurynas-biveinis:
@robgolebiowski, please update the Launchpad bug

PR #1300, opened 2017-01-19
 DB-1033 : TokuDB ::info uses bad/lengthy function to provide table s…

george-lorch:
…tatus

- Changed ::info call to simply return the 'deleted' size s the diff between
  data size and file size. This isn't entirely accurate but is far better than
  having PerconaFT scan the complete block table every time ::info is called.


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1598/

laurynas-biveinis (approved):
Not sure whom you wanted to review this, but, hey, LGTM

george-lorch:
LOL! No specific target, just giving our crowd a short period to object before I merge.

PR #1302, opened 2017-01-20
5.7 ps bug897715

ihanick:
Fix for https://bugs.launchpad.net/percona-server/+bug/897715
5.6 PR: #1243 

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/580/

The fix is the same as for 5.6 (clear innodb_trx_id after slow log actions only if transaction is not active).
For 5.7 trx->id is also cleared after each transaction by trx_init (called by trx_commit_in_memory and during init).
Thus for 5.7, bare statement trx->id is always 0 in slow log. For 5.6 all innodb statements having valid transaction id.

### DIFF
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -927,7 +927,7 @@ bool File_query_log::write_slow(THD *thd, ulonglong current_utime,
 #endif
 
   if ((thd->variables.log_slow_verbosity & (1ULL << SLOG_V_INNODB))
-      && thd->innodb_was_used)
+      && thd->innodb_trx_id)
   {
     char buf[20];
     snprintf(buf, 20, "%llX", thd->innodb_trx_id);
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -2083,7 +2083,7 @@ bool MYSQL_QUERY_LOG::write(THD *thd, ulonglong current_utime,
     thd->profiling.print_current(&log_file);
 #endif
     if ((thd->variables.log_slow_verbosity & (1ULL << SLOG_V_INNODB))
-        && thd->innodb_was_used)
+        && thd->innodb_trx_id)
     {
       char buf[20];
       snprintf(buf, 20, "%llX", thd->innodb_trx_id);
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -16475,7 +16475,7 @@ ha_innobase::external_lock(
 
 		if (UNIV_UNLIKELY(trx->take_stats)) {
 			increment_thd_innodb_stats(thd,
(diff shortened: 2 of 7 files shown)


sql/log.cc
```diff
@@ -927,7 +927,7 @@ bool File_query_log::write_slow(THD *thd, ulonglong current_utime,
 #endif
 
   if ((thd->variables.log_slow_verbosity & (1ULL << SLOG_V_INNODB))
-      && thd->innodb_was_used)
+      && thd->innodb_trx_id )
```
laurynas-biveinis: No space before closing parenthesis

storage/innobase/handler/ha_innodb.cc
```diff
@@ -16452,14 +16452,15 @@ ha_innobase::external_lock(
 
 		if (UNIV_UNLIKELY(trx->take_stats)) {
 			increment_thd_innodb_stats(thd,
-						   (unsigned long long) trx->id,
+						   (unsigned long long) trx->id? trx->id: trx->id_saved,
```
laurynas-biveinis: Please put the ternary expression into parens (or the type cast applies to trx->id only)
No spaces before ? and :
Since you are touching this line, please use C++ style cast

storage/innobase/include/trx0trx.h
```diff
@@ -928,6 +928,7 @@ struct trx_t {
 					it can */
 
 	trx_id_t	id;		/*!< transaction id */
+	trx_id_t	id_saved;	/*!< transaction id */
```
laurynas-biveinis: Needs a describing comment different from plain id

storage/innobase/trx/trx0trx.cc
```diff
@@ -124,6 +124,7 @@ trx_init(
 	trx_t::state here to NOT_STARTED. The FORCED_ROLLBACK
 	status is required for asynchronous handling. */
 
+	trx->id_saved = trx->id;
```
laurynas-biveinis, line 127: I am not sure if that's the right place to save the ID - could this result in old committed RW trx IDs reused by new transactions in the slow log?

laurynas-biveinis (changes requested):

ihanick:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/675/
- rebase to fresh 5.7
- fixed whitespaces
- c++ static_cast
- trx_init and clear trx->id happens before slow logging, thus I'm keeping transaction id in trx->id_saved until slow log details sent to thread (increment_thd_innodb_stats). For read-only queries and transactions trx->id is always zero and not reused from previous transaction.

laurynas-biveinis (approved):

PR #1304, opened 2017-01-20
Fix bug 1646100 (SHOW STATUS blocked by binlog commit)

laurynas-biveinis:
In the existing binary log status variables implementation,
MYSQL_BIN_LOG::set_status_variables is called with locked LOCK_status
and then it attempts to lock LOCK_log. This means that e.g. a
long-running commit blocks any SHOW STATUS, and the latter then blocks
e.g. client connects/disconnects.

While MySQL has not sorted the core server locks by the required
locking order, there is no precedent in the core server of LOCK_status
owners taking any other mutexes, and most LOCK_status critical
sections there are short. This suggests that LOCK_status should come
near the end of locking order, and so fix the issue by locking
LOCK_log before LOCK_status.

For that, create two new static variables binlog_global_snapshot_file
and binlog_global_snapshot_position to binlog.cc that hold last
"published" global binary log position. Add a new method
MYSQL_BIN_LOG::publish_global_status_variables, that, under LOCK_log
taken already, locks LOCK_status, and sets the global position
variables. Call it in the group commit after flush phase, and
optionally after binlog rotate. Change show_binlog_vars to read the
global position variables if required (no snapshot; binlog open), and
avoid taking LOCK_log there.

http://jenkins.percona.com/job/percona-server-5.6-param/1595/

### DIFF
--- a/mysql-test/suite/binlog/t/binlog_consistent_debug.test
+++ b/mysql-test/suite/binlog/t/binlog_consistent_debug.test
@@ -1,0 +1,55 @@
+--source include/have_debug_sync.inc
+--source include/have_innodb.inc
+--source include/have_log_bin.inc
+
+--echo #
+--echo # Debug build tests for consistent binlog snapshot
+--echo #
+
+RESET MASTER;
+
+--echo #
+--echo # Bug 1646100: Server becomes unresponsive during flushing after loading
+--echo # big files through LOAD DATA INFILE
+--echo #
+
+SET @saved_sync_binlog = @@sync_binlog;
+SET GLOBAL sync_binlog = 1;
+
+CREATE TABLE t1 (a INT);
+
+--source include/count_sessions.inc
+
+--connect(con1,localhost,root,,)
+--echo # connection con1
+
+BEGIN;
+INSERT INTO t1 VALUES (0);
+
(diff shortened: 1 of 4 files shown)


laurynas-biveinis:
@robgolebiowski, please review as well (couldn't add you directly yet)

mysql-test/suite/binlog/t/binlog_consistent_debug.test
```diff
@@ -0,0 +1,55 @@
...
+SET DEBUG_SYNC="before_sync_binlog_file SIGNAL commit_ready WAIT_FOR finish_commit";
+
+send COMMIT;
+
+--connection default
+--echo # connection default
+
+SET DEBUG_SYNC="now WAIT_FOR commit_ready";
+
+--replace_result 405 POSITION 426 POSITION
```
percona-ysorokin, line 38: Why 2 different values are expected?

laurynas-biveinis: Different binlog formats

percona-ysorokin (approved):
LGTM

robgolebiowski:
LGTM

laurynas-biveinis:
Updated method name

robgolebiowski:
LGTM

PR #1305, opened 2017-01-20
Bug1646100 5.7

laurynas-biveinis:
http://jenkins.percona.com/job/mysql-5.7-param/584/

@robgolebiowski, please review as well

### DIFF
--- a/sql/binlog.h
+++ b/sql/binlog.h
@@ -689,7 +689,8 @@ public:
   inline void unlock_index() { mysql_mutex_unlock(&LOCK_index);}
   inline IO_CACHE *get_index_file() { return &index_file;}
   inline uint32 get_open_count() { return open_count; }
-  void set_status_variables(THD *thd);
+private:
+  void publish_coordinates_for_global_status(void) const;
 };
 
 typedef struct st_load_file_info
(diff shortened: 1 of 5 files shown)


percona-ysorokin (approved):
LGTM

sql/binlog.h
```diff
@@ -949,6 +948,8 @@ class MYSQL_BIN_LOG: public TC_LOG
       @retval !=0    Error
   */
   int get_gtid_executed(Sid_map *sid_map, Gtid_set *gtid_set);
+private:
+  void publish_global_status_variables(void) const;
```
robgolebiowski: maybe I would just changed the name of this function to publish_global_snapshot_status_variable.

robgolebiowski:
I added only one comment. LGTM

laurynas-biveinis:
@robgolebiowski, please take a look at https://github.com/percona/percona-server/pull/1304 too for 5.6

sql/binlog.h
laurynas-biveinis: Hmm indeed not the best name. publish_position_for_global_status ?

robgolebiowski: I think this is better than the previous one. Just note that apart from position you are also publishing binlog_global_snapshot_file. Maybe publish_position_for_global_snapshot_status, but publish_position_for_global_status is good too.

laurynas-biveinis: Ah, "coordinates" then (a pair of name, position), making  it publish_coordinates_for_global_status ?

robgolebiowski: Agreed :+1:

laurynas-biveinis:
Updated method name

robgolebiowski:
LGTM

PR #1308, opened 2017-01-20
FT-732 : logical row counts inaccurate

george-lorch:
- Advanced PerconaFT git submodule pointer to new commit for fix


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1600/

PR #1313, opened 2017-01-24
Fix bug 1658055 (Log tracking initialisation does not find the last g…

laurynas-biveinis:
…ood bitmap data correctly)

On server startup, log tracking tries to find the last good bitmap
data, discard any later junk, and re-track its log. The intention of
the algorithm is to open the last bitmap file, read the last page,
check its checksum and last-page flag, and if either check fails,
discard it, and seek back in the file.

However, the read loop condition currently is "while (!checksum_ok &&
read_offset > 0 && !is_last_page)", that is, only read back if both
the checksum is wrong and the last page flag is set. Fix by making it
"(!checksum_ok || !is_last_page) && read_offset > 0".

Also, if the last bitmap file ends up being zero-sized (if all of its
data was junk), then a bitmap file rotation was needlessly being
performed. Fix by continuing writing to that zero-sized file instead.

Replace some of the bitmap file name listings with an
INFORMATION_SCHEMA.INNODB_CHANGED_PAGES query that reads all the
bitmaps instead as a stricter check.

For the testcase, backport the change from the higher versions where
bitmap_page_write_error fault injection triggers only for non-system
tablespaces. Do the same for the crash-injection site
crash_before_bitmap_write, upmerge of which will fix bug 1658021 (Test
innodb.percona_changed_page_bmp_crash is unstable). Fix debug
injection site bitmap_page_2_write_error to trigger only if the second
bitmap data page is indeed available.

http://jenkins.percona.com/job/percona-server-5.5-param/1515/

@robgolebiowski

### DIFF
--- a/mysql-test/suite/innodb/r/percona_changed_page_bmp.result
+++ b/mysql-test/suite/innodb/r/percona_changed_page_bmp.result
@@ -5,15 +5,18 @@ INSERT INTO t1 VALUES (1),(2),(3),(4),(5);
 ib_modified_log_1
 1st restart
 INSERT INTO t1 VALUES (1),(2),(3),(4),(5);
+include/assert.inc [There should not be a hole in the tracked LSN range]
 ib_modified_log_1
 ib_modified_log_2
 2nd restart
 INSERT INTO t1 VALUES (1),(2),(3),(4),(5);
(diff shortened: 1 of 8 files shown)


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1316, opened 2017-01-24
MYR-27 : facebook/webscale mysqldump changes missing

george-lorch:
- Port from Facebook MySQL mysqldump option "--order-by-primary-desc=true|false
  (default false) : take backup ORDER BY primary key DESC"
- Detect when MyRocks is installed and available by seeing if there is a session
  variable named rocksdb_skip_fill_cache and "SET SESSION
  rocksdb_skip_fill_cache=1" if it exists.
- Add logic into dump file that automatically enables session rocksdb_bulk_load
  if it is supported by target server.
- Touched up mtr cases to encompass new behaviors.


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1607/

client/mysqldump.c
```diff
@@ -742,6 +746,14 @@ static void write_header(FILE *sql_file, char *db_name)
...
             path?"":"NO_AUTO_VALUE_ON_ZERO",compatible_mode_normal_str[0]==0?"":",",
             compatible_mode_normal_str);
+    fprintf(sql_file,
+            "/*!50601 SELECT count(*) INTO @is_rocksdb_supported FROM"
+            " information_schema.SESSION_VARIABLES WHERE"
+            " variable_name='rocksdb_bulk_load' */;\n"
+            "/*!50601 SET @enable_bulk_load = IF (@is_rocksdb_supported,"
+            " 'SET SESSION rocksdb_bulk_load=1', 'SET @dummy = 0') */;\n"
+            "/*!50601 PREPARE s FROM @enable_bulk_load */;\n"
+            "/*!50601 EXECUTE s */;\n");
```
laurynas-biveinis: DEALLOCATE PREPARE ?

client/mysqldump.c
```diff
@@ -756,6 +768,11 @@ static void write_footer(FILE *sql_file)
   }
   else if (!opt_compact)
   {
+    fprintf(sql_file,
+            "/*!50601 SET @disable_bulk_load = IF (@is_rocksdb_supported, "
+            "'SET SESSION rocksdb_bulk_load=0', 'SET @dummy = 0') */;\n"
+            "/*!50601 PREPARE s FROM @disable_bulk_load */;\n"
+            "/*!50601 EXECUTE s */;\n");
```
laurynas-biveinis: DEALLOCATE PREPARE?

client/mysqldump.c
```diff
@@ -5628,6 +5646,29 @@ static int purge_bin_logs_to(MYSQL *mysql_con, char* log_name)
...
+
+  my_snprintf(query, sizeof(query), "SELECT COUNT(*) FROM"
+              " INFORMATION_SCHEMA.SESSION_VARIABLES WHERE VARIABLE_NAME LIKE"
+              " %s", quote_for_like(var_name, buf));
+  if (mysql_query_with_error_report(mysql_con, &res, query))
+    return FALSE;
+
+  row = mysql_fetch_row(res);
+  val = row ? (char*)row[0] : NULL;
+  if (val && strcmp(val, "0") != 0)
```
laurynas-biveinis: has_var= (val && strcmp...);

client/mysqldump.c
```diff
@@ -5652,7 +5693,10 @@ static int start_transaction(MYSQL *mysql_con)
     if (!ignore_errors)
       exit(EX_MYSQLERR);
   }
-
+  if (has_session_variables_like(mysql_con, "rocksdb_skip_fill_cache"))
+    mysql_query_with_error_report(mysql_con, 0,
+                                  "/*!50601 SET SESSION"
```
laurynas-biveinis: No need for !50601: the variable name is checked for, and SET SESSION itself has been here forever

client/mysqldump.c
```diff
@@ -5874,6 +5918,10 @@ static char *primary_key_fields(const char *table_name)
     {
       quoted_field= quote_name(row[4], buff, 0);
       result_length+= strlen(quoted_field) + 1; /* + 1 for ',' or \0 */
+      if (desc)
+      {
+        result_length+= 5; /* + 1 for space, and + 4 for "DESC" */
```
laurynas-biveinis: I'd do static const char *desc_index= " DESC"; here strlen, below use the var directly, but this is strictly up to you

client/mysqldump.c
```diff
@@ -5895,7 +5943,11 @@ static char *primary_key_fields(const char *table_name)
     while ((row= mysql_fetch_row(res)) && atoi(row[3]) > 1)
     {
       quoted_field= quote_name(row[4], buff, 0);
-      end= strxmov(end, ",", quoted_field, NullS);
+      end= strxmov(end, desc? " DESC," : ",", quoted_field, NullS);
```
laurynas-biveinis: space before ?

client/mysqldump.c
```diff
@@ -5652,7 +5693,10 @@ static int start_transaction(MYSQL *mysql_con)
     if (!ignore_errors)
       exit(EX_MYSQLERR);
   }
-
+  if (has_session_variables_like(mysql_con, "rocksdb_skip_fill_cache"))
```
laurynas-biveinis: Better to query this before any locking, i.e. move from start_transation to its own function, call it if (opt_single_transaction) from main

client/mysqldump.c
```diff
@@ -756,6 +768,11 @@ static void write_footer(FILE *sql_file)
   }
   else if (!opt_compact)
   {
+    fprintf(sql_file,
+            "/*!50601 SET @disable_bulk_load = IF (@is_rocksdb_supported, "
+            "'SET SESSION rocksdb_bulk_load=0', 'SET @dummy = 0') */;\n"
```
laurynas-biveinis: Should save the original value and restore to it instead of disabling

client/mysqldump.c
```diff
@@ -742,6 +746,14 @@ static void write_header(FILE *sql_file, char *db_name)
             "/*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;\n",
             path?"":"NO_AUTO_VALUE_ON_ZERO",compatible_mode_normal_str[0]==0?"":",",
             compatible_mode_normal_str);
+    fprintf(sql_file,
+            "/*!50601 SELECT count(*) INTO @is_rocksdb_supported FROM"
+            " information_schema.SESSION_VARIABLES WHERE"
+            " variable_name='rocksdb_bulk_load' */;\n"
+            "/*!50601 SET @enable_bulk_load = IF (@is_rocksdb_supported,"
+            " 'SET SESSION rocksdb_bulk_load=1', 'SET @dummy = 0') */;\n"
```
laurynas-biveinis: Maybe some other SQL for ELSE part here and below? This one clobbers a user variable

client/mysqldump.c
```diff
@@ -742,6 +746,14 @@ static void write_header(FILE *sql_file, char *db_name)
             "/*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;\n",
             path?"":"NO_AUTO_VALUE_ON_ZERO",compatible_mode_normal_str[0]==0?"":",",
             compatible_mode_normal_str);
+    fprintf(sql_file,
+            "/*!50601 SELECT count(*) INTO @is_rocksdb_supported FROM"
```
laurynas-biveinis: !50601 is too late for a i_s.S_V query, the table exists starting from 5.1

mysql-test/suite/rocksdb/r/mysqldump.result
```diff
@@ -76,7 +76,7 @@ UNLOCK TABLES;
 
 rollback;
 connection con1;
-1
+0
```
laurynas-biveinis: Shouldn't it stay at one?

mysql-test/suite/rocksdb/r/mysqldump.result
```diff
@@ -123,7 +129,7 @@ UNLOCK TABLES;
 /*!40101 SET COLLATION_CONNECTION=@OLD_COLLATION_CONNECTION */;
 /*!40111 SET SQL_NOTES=@OLD_SQL_NOTES */;
 
-2
+0
```
laurynas-biveinis: Likewise

laurynas-biveinis (changes requested):

laurynas-biveinis:
openssl_1 test failures are caused by this patch, need re-record

client/mysqldump.c
george-lorch: Alrighty, part of the original patch and didn't really consider that too much.

client/mysqldump.c
george-lorch: Hmm, perhaps, maybe a specifically named dummy or a SELECT 1; or something. Again, part of original patch that I really didn't consider much.

client/mysqldump.c
george-lorch: I had wondered about this myself. Original patch has none. Documentation indicates tat is session scoped so no real loss once session ends, but I suppose session pollution and just general good practice calls for deallocation.

client/mysqldump.c
george-lorch: Yeah, good point, and again, original patch... :-\

client/mysqldump.c
george-lorch: Another copy/past from original patch

client/mysqldump.c
laurynas-biveinis: I see, this and other similar bits could be something to send upstream then

mysql-test/suite/rocksdb/r/mysqldump.result
george-lorch: Unsure, that'ts a good catch. That looks to be the result of the grep for STWCS which should remain the same due to the --single-transaction flag. I tested that specifically when applying part of one of the patches but must have munged it somehow when applying or fixing another of the patches.

client/mysqldump.c
george-lorch: Yeah, I'm thinking the same thing.

george-lorch: Another perspective here, as I understand, the purpose of these is in the case that the result of this dump it attempted to be imported/replayed against an older server version. For the purposes of MyRocks, we _know_ that there will never be a server version older then 5.6 with the plugin installed, not from us, not from Facebook, not from MariaDB. Even if we did see one, not setting this bulk load option is not a catastrophic event, it is simply an optional optimization. So I'm thinking this is OK as is, otherwise, (since I am not terribly familiar with mysqldump history and semantics) I am not sure how these should be treated. Simply change the version requirement to some 5.1? Remove them entirely?

client/mysqldump.c
george-lorch: True, and really this has nothing to do with locking, it should probably be done just ahead of start_transaction to tell rocksdb, if it is in use, not to fill the cache on this session. This is where the original patch intersects with our minor change to STWCS and the patch flag --rocksdb which we we are not carrying over.

client/mysqldump.c
laurynas-biveinis: I only don't see a point in pushing a known-wrong version marker, but even that is benign as you say. Either patching up it to 5.1 or removing is fine

george-lorch: I can't argue with that. I'll see if I can figure out what version that i_s table was introduced and use those version markers.

george-lorch: Alright, I finally found when these tables were introduced https://docs.oracle.com/cd/E17952_01/mysql-5.1-en/variables-table.html seems 5.1.12

george-lorch:
OK, I think I got everything mentioned in the comments. The grep failure in rocksdb.mysqldump was due to a difference between the Facebook MySQL mysqldump and ours where they were expecting "START TRANSACTION WITH CONSISTENT ROCKSDB SNAPSHOT" and we were expecting "START TRANSACTION /*!40100 WITH CONSISTENT SNAPSHOT */". My original fix took care of the "ROCKSDB" portion but missed the version marker difference.

New jenkins run http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1625/

mysql-test/suite/rocksdb/r/mysqldump.result
```diff
@@ -54,16 +59,16 @@ CREATE TABLE `r1` (
   PRIMARY KEY (`id1`,`id2`,`id3`,`id4`)
 ) ENGINE=ROCKSDB DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
-/* ORDERING KEY (DESC) : PRIMARY */;
```
laurynas-biveinis, line 57: Why this change? Is it related to --print-ordering-key removal?

mysql-test/suite/rocksdb/r/mysqldump.result
```diff
@@ -106,13 +119,16 @@ CREATE TABLE `r1` (
   PRIMARY KEY (`id1`,`id2`,`id3`,`id4`)
 ) ENGINE=ROCKSDB DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
-/* ORDERING KEY : (null) */;
```
laurynas-biveinis, line 109: Likewise

mysql-test/suite/rocksdb/t/mysqldump.test
```diff
@@ -34,30 +34,31 @@ insert into r1 values (2,2,2,1,15,15,15,15);
 insert into r1 values (2,2,2,2,16,16,16,16);
 
 connection con2;
+SET SESSION TRANSACTION ISOLATION LEVEL READ COMMITTED;
 BEGIN;
 insert into r1 values (5,5,5,5,5,5,5,5);
 update r1 set value1=value1+100 where id1=1 and id2=1 and id3='1';
 
---exec $MYSQL_DUMP --skip-comments --single-transaction --master-data=2 --print-ordering-key --rocksdb --order-by-primary-desc --rocksdb_bulk_load test
```
laurynas-biveinis, line 41: What's the deal re. --print-ordering-key?

laurynas-biveinis (commented):
Need couple of clarifications, otherwise LGTM

mysql-test/suite/rocksdb/t/mysqldump.test
george-lorch: --print-ordering-key was a Facebook MySQL addition to mysqldump that we have not/did not take  in and is not related to MyRocks. It came in this commit to Facebook MySQL 2+ years ago https://github.com/facebook/mysql-5.6/commit/0f2578a086ce7822ce16fb9ac5c47e515ea2b1c5

It is what generated the "/* ORDERING KEY (DESC) : PRIMARY */;" output above and came in the MyRocks test results during the initial import. Now that we are re-recording these MyRocks tests to actually pass with our implementation of consistent snapshot and other bits of this pull request, these are disappearing.

laurynas-biveinis (approved):

PR #1317, opened 2017-01-24
Ps 5.7 myr 27

george-lorch:


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/595/

laurynas-biveinis (changes requested):
Same as in 5.6

george-lorch:
New jenkins http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/625/

sql/handler.cc
```diff
@@ -8566,6 +8566,7 @@ bool handler::is_using_prohibited_gap_locks(TABLE* table,
   if (!using_full_primary_key
       && has_transactions()
       && !has_gap_locks()
+      && thd_tx_isolation(thd) >= ISO_REPEATABLE_READ
```
laurynas-biveinis: Something is not right with this merge - this change and the listed conflicts

laurynas-biveinis (changes requested):

sql/handler.cc
george-lorch: Yeah, something is wrong here, unsure of what, likely a GCA and working branch mixup when I forward merged this to 5.7.

george-lorch:
This is strange, I can explain/fix the sql/handler.cc issue, that was simply an accidental git add from stashed changes off a working branch.

I can't explain all of the conflicts. these are legit as there are slight differences between 5.6 and 5.7 that need manual reconciliation:
client/client_priv.h
client/mysqldump.c

But the rest make no sense at all, there are no changes to these files whatsoever so how can there be a conflict?

I'm thinking this might be a broken GCA merge attempt now that our 5.6 and 5.7 branches have maybe moved ahead. A cherry-pick of this same merge works fine, except it is a cherry-pick and not a GCA merge.

laurynas-biveinis:
Could it be that your 5.7 base commit is very old? Otherwise, once you fix handler.cc, check through command line interface if the diff looks OK and that's fine

george-lorch:
Yeah, it's likely just a very old base 5.7 for these, going back to when I did the first GCA import. This is what the status looks like when merging to 5.7 master before committing:
```
git status
On branch 5.7
Your branch is up-to-date with 'origin/5.7'.

All conflicts fixed but you are still merging.
  (use "git commit" to conclude merge)

Changes to be committed:

        modified:   client/client_priv.h
        modified:   client/mysqldump.c
        modified:   mysql-test/r/backup_locks_mysqldump.result
        modified:   mysql-test/r/mysqldump-max.result
        modified:   mysql-test/r/mysqldump.result
        modified:   mysql-test/r/openssl_1.result
        modified:   mysql-test/r/percona_mysqldump_innodb_optimize_keys.result
        modified:   mysql-test/suite/rocksdb/r/mysqldump.result
        modified:   mysql-test/suite/rocksdb/t/mysqldump.test
        modified:   mysql-test/suite/rocksdb/t/mysqldump2.test
        modified:   sql/handler.cc
```

So this is exactly the commit content that we want, it is safe to merge.

george-lorch:
And local mtr built on this proposed new 5.7 trunk passes as desired. So I think this is just some weird old-base-branch issue. Moving forward into cherry-pick instead of GCA should eliminate this kind of oddity from happening.

laurynas-biveinis:
GCA process is what makes the 5.5/5.6 branches old - but it does not prevent from using the latest 5.7.

george-lorch:
Yes, I know. I have a whole series of branches and scripts at work here in order to maintain GCA start points for individual 5.6 commits and a working (compiling) 5.6 branch for MyRocks and have been mostly blindly doing forward merges to 5.7 without having done a local update to it. There is no reason not to update my 5.7 base other than I haven't done it due to just having too many things running in parallel. Jumping ship to a cherry-pick based workflow make no excuse for not having all commits based on both 5.6 and 5.7 branch heads.

PR #1320, opened 2017-01-26
Fix nondeterminism in a mtr test

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.6-param/1608/

### DIFF
--- a/mysql-test/suite/sys_vars/t/innodb_buffer_pool_load_now_basic.test
+++ b/mysql-test/suite/sys_vars/t/innodb_buffer_pool_load_now_basic.test
@@ -5,6 +5,19 @@
 -- source include/have_innodb.inc
 --source include/not_embedded.inc
 
+# Restart mysqld to ensure that the status variables below do not contain
+# stale value from previous executions of this or other tests. The following
+# race could happen without this restart:
+# 0. innodb_buffer_pool_load_status is 'Buffer pool(s) load completed ...'
+# 1. SET innodb_buffer_pool_load_now = ON
+# (1. does not start executing yet)
+# 2. Wait for innodb_buffer_pool_load_status to become '... completed ...',
+#    the condition is satisfied immediately
+# (1. starts executing now)
+# 3. Query innodb_buffer_pool_load_status, expecting 'completed', but it
+#    contains something like 'Loading page 100/150'
+-- source include/restart_mysqld.inc
+
 # Check the default value
 SET @orig = @@global.innodb_buffer_pool_load_now;
 SELECT @orig;


mysql-test/suite/sys_vars/t/innodb_buffer_pool_load_now_basic.test
```diff
@@ -5,6 +5,19 @@
...
 --source include/not_embedded.inc
 
+# Restart mysqld to ensure that the status variables below do not contain
+# stale value from previous executions of this or other tests. The following
+# race could happen without this restart:
+# 0. innodb_buffer_pool_load_status is 'Buffer pool(s) load completed ...'
+# 1. SET innodb_buffer_pool_load_now = ON
+# (1. does not start executing yet)
+# 2. Wait for innodb_buffer_pool_load_status to become '... completed ...',
+#    the condition is satisfied immediately
```
robgolebiowski, line 15: I am not sure, but maybe setting SET GLOBAL innodb_buffer_pool_load_abort=ON; would also clear innodb_buffer_pool_load_status, instead of restarting the server ?

laurynas-biveinis: Probably - but here I'm only backporting an upstream fix from a higher version, then the diff disappears in those higher versions. We need to have a good reason to change such backported fixes, as then the maintenance becomes different (diff does not disappear - send it back to MySQL)

robgolebiowski: Ah, OK - then LGTM. Is there a way of knowing which bug fixes are backported, so maybe the code review process could be quicker ?

percona-ysorokin (approved):
LGTM

mysql-test/suite/sys_vars/t/innodb_buffer_pool_load_now_basic.test
laurynas-biveinis: Right, we have to use "-x" option for git cherry-pick going forward:

"       -x
           When recording the commit, append a line that says "(cherry picked from commit ...)" to the original commit message in order to indicate which commit
           this change was cherry-picked from. This is done only for cherry picks without conflicts. Do not use this option if you are cherry-picking from your
           private branch because the information is useless to the recipient. If on the other hand you are cherry-picking between two publicly visible branches
           (e.g. backporting a fix to a maintenance branch for an older release from a development branch), adding this information can be useful."

robgolebiowski (commented):
LGTM

PR #1322, opened 2017-01-26
Add diagnostics to catch bug 1659224 (Test main.status is unstable)

laurynas-biveinis:
Dump the general log in case executing "SELECT f1(); CALL p1();"
results in a different delta of Queries status variable than
expected.

http://jenkins.percona.com/job/percona-server-5.6-param/1609/

### DIFF
--- a/mysql-test/t/status.test
+++ b/mysql-test/t/status.test
@@ -319,6 +319,11 @@ let $wait_condition =
 #
 # Bug#41131 "Questions" fails to increment - ignores statements instead stored procs
 #
+SET @old2_log_output = @@GLOBAL.log_output;
+SET GLOBAL log_output = 'TABLE';
+SET @old_general_log = @@GLOBAL.general_log;
+SET GLOBAL general_log = 'ON';
+TRUNCATE TABLE mysql.general_log;
 connect (con1,localhost,root,,);
 connection con1;
 --disable_warnings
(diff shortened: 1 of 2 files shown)


robgolebiowski:
LGTM

mysql-test/t/status.test
```diff
@@ -319,6 +319,10 @@ let $wait_condition =
 #
 # Bug#41131 "Questions" fails to increment - ignores statements instead stored procs
 #
+SET GLOBAL log_output = 'TABLE';
```
percona-ysorokin, line 323: May be store the value of `log_output` here as well
```sql
SET @old_log_output = @@GLOBAL.log_output;
```

mysql-test/t/status.test
```diff
@@ -347,10 +351,16 @@ let $new_queries= `SHOW STATUS LIKE 'Queries'`;
...
 eval SELECT $diff;
+if ($diff != 9)
+{
+  SELECT * FROM mysql.general_log;
+}
 disconnect con1;
 connection default;
 DROP PROCEDURE p1;
 DROP FUNCTION f1;
+SET GLOBAL log_output = 'FILE';
```
percona-ysorokin: restore it here

percona-ysorokin (approved):
LGTM
with minor comments

mysql-test/t/status.test
laurynas-biveinis: That is already done by the surrounding testcase, so I only restore to what the testcase sets it to previously. But what you suggest is more robust if the surrounding testcase changes

percona-ysorokin: Anyway, the PR is already approved

laurynas-biveinis:
Addressed review comments, http://jenkins.percona.com/job/percona-server-5.6-param/1620/

robgolebiowski (approved):
LGTM

PR #1270, opened 2016-12-30
Disable tokudb backup on --innodb-use-native-aio=on

vlad-lesin:
See https://tokutek.atlassian.net/browse/BACKUP-121.
http://jenkins.percona.com/job/percona-server-5.6-param/1573/

PR #1324, opened 2017-01-26
Store binlog position and gtid in separate file for hotbackup.

vlad-lesin:
The algorithm for getting master-info is the following:

1) LOCK BINLOG FOR BACKUP just before stop capturing.
This statement will block any queries which would change binlog position or
executed gtids set and flushes master info in innodb;
2) Stop capturing;
3) Get master info and store it in separate file in backup directory;
4) UNLOCK BINLOG

Currently HotBackup DOES NOT give any guaranties of non-transactional tables
consistency because it does not use any flush-and-lock commands for backup.

See also:
https://github.com/percona/percona-server/pull/1325
https://github.com/percona/Percona-TokuBackup/pull/80

http://jenkins.percona.com/job/percona-server-5.6-param/1611/

### DIFF
--- a/mysql-test/include/filter_file.inc
+++ b/mysql-test/include/filter_file.inc
@@ -15,6 +15,7 @@
 # [--let $select_columns= <LIST OF NUMBERS>]
 # [--let $pre_script= <PERL_SCRIPT>]
 # [--let $rpl_debug= 1]
+# [--let $skip_column_names= 0]
 # --source include/filter_file.inc
 #
 # Parameters:
@@ -45,6 +46,9 @@
 #     numbers, it will print only the numbered columns, in the given
 #     order.
 #
+#   $skip_column_names
+#     If true then skip reading columns line. false by default.
+#
 #   $pre_script
 #     This script will be evaluated before starting to iterate over
 #     the lines of $input_file.  It can be useful if you need some
@@ -70,6 +74,10 @@ if ($rpl_debug)
 --let _FF_INPUT_FILE= $input_file
 --let _FF_OUTPUT_FILE= $output_file
 --let _FF_SELECT_COLUMNS= $select_columns
+if (!$skip_column_names) {
+  --let $skip_column_names= 0;
+}
+--let _FF_SKIP_COLUMN_NAMES= $skip_column_names
 --let _FF_DEBUG= $rpl_debug
 if (!$output_file)
 {
@@ -83,6 +91,7 @@ perl;
   my $input_file = $ENV{'_FF_INPUT_FILE'};
   my $output_file = $ENV{'_FF_OUTPUT_FILE'};
   my $select_columns = $ENV{'_FF_SELECT_COLUMNS'};
+  my $skip_column_names = $ENV{'_FF_SKIP_COLUMN_NAMES'};
   my $debug = $ENV{'_FF_DEBUG'};
   if ($select_columns)
   {
@@ -114,7 +123,7 @@ perl;
   {
     chomp;
     s/\015?\012?$//;
-    if (!%column_names)
+    if (!%column_names && $skip_column_names == 0)
     {
       my $n = 1;
       %column_names = map { $_ => $n++ } split(/\t/, $_);
--- a/plugin/tokudb-backup-plugin/tokudb_backup.cc
+++ b/plugin/tokudb-backup-plugin/tokudb_backup.cc
@@ -3,6 +3,7 @@
 #ident "Copyright (c) 2014 Tokutek Inc.  All rights reserved."
 
 #define MYSQL_SERVER
+#define HAVE_REPLICATION
 #include <my_config.h>
 #include <mysql_version.h>
 #include <mysql/plugin.h>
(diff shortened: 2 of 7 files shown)


plugin/tokudb-backup-plugin/tokudb_backup.cc
```diff
@@ -3,6 +3,7 @@
 #ident "Copyright (c) 2014 Tokutek Inc.  All rights reserved."
 
 #define MYSQL_SERVER
+#define HAVE_REPLICATION
```
george-lorch, line 6: I believe this may not be necessary anymore as per @laurynas-biveinis from earlier reviews

sql/sql_parse.cc
```diff
@@ -2572,7 +2572,7 @@ static bool lock_tables_for_backup(THD *thd)
   @return FALSE in case of success, TRUE in case of error.
 */
 
-static bool lock_binlog_for_backup(THD *thd)
+bool lock_binlog_for_backup(THD *thd)
```
george-lorch, line 2575: These changes, being up in the server code, will need a small blueprints to attach to for 5.6 and 5.7

sql/sql_parse.h
```diff
@@ -214,4 +214,6 @@ inline bool is_supported_parser_charset(const CHARSET_INFO *cs)
 
 extern "C" bool sqlcom_can_generate_row_events(enum enum_sql_command command);
 
+bool lock_binlog_for_backup(THD *thd);
```
george-lorch, line 217: Same, need blueprint to explain changes

mysql-test/include/filter_file.inc
```diff
@@ -114,7 +123,7 @@ perl;
   {
     chomp;
     s/\015?\012?$//;
-    if (!%column_names)
+    if (!%column_names && $skip_column_names == 0)
```
george-lorch, line 126: This all looks reasonable to me but I'm not terribly familiar with this functionality in mtr internals. @laurynas-biveinis do you have a compelling opinion?

plugin/tokudb-backup-plugin/tokudb_backup.cc
george-lorch: Actually, I may be mistaking this with PARTITIONING...

mysql-test/include/filter_file.inc
george-lorch: Also likely need blueprint for these common server mtr changes.

plugin/tokudb-backup-plugin/tokudb_backup.cc
vlad-lesin: Some necessary types and functions are not accessible without this define. I could try to pass this define from server cmake configuration. But in this case the most of the changes will be wrapped with ifdef.

sql/sql_parse.cc
vlad-lesin: This changes are quite obvious. I did it to make this function accessible from tokudb-hotbackup plugin. I don't think we need blueprints for such small changes. At least it was not required for this pull request https://github.com/percona/percona-server/pull/1270 , where I did the same thing for global variable in innodb. @laurynas-biveinis, what do you think?

mysql-test/include/filter_file.inc
vlad-lesin: Guys, with such rules our work will stop. There is no need to be mtr guru to understand this code, it's just perl. This changes allow to optionally skip parsing the first line of a file. And filter_file.inc contains quite detailed comments, the examples of it's usage can be found with grep. And Jenkins testing results show this changes do not affect existing tests.

vlad-lesin: @laurynas-biveinis, we are waiting for your decision about if blueprint is required.

vlad-lesin:
What is necessary is to rename the file to which master info is stored, because apart from this functionality we also need to store binlog position for master. Will do renaming. See my chat with Shakko here https://tokutek.atlassian.net/browse/BACKUP-118?focusedCommentId=26706 .

mysql-test/include/filter_file.inc
laurynas-biveinis: This one does not need a blueprint for sure

sql/sql_parse.cc
laurynas-biveinis: That's right, I missed it in #1270. Yes, we would need a blueprint "expose LOCK BINLOG FOR BACKUP function to plugins", but it's borderline silly in this instance. I'm willing to let this one slide

plugin/tokudb-backup-plugin/tokudb_backup.cc
vlad-lesin: Well, actually, it's not required for 5.6, but it's required for 5.7. But I would prefer to have as more similar code as possible for 5.6 and 5.7 to avoid merge conflicts. Otherwise I would not use vector to gather slave information in 5.6 because multisource replication is not implemented for 5.6.

sql/sql_parse.cc
vlad-lesin: I think the rules must be common and if we require blueprints for such changes, this changes must not be exception. I created blueprints, can be found here: https://blueprints.launchpad.net/percona-server/+spec/lock-binlog-for-backup-expose-5.6, https://blueprints.launchpad.net/percona-server/+spec/lock-binlog-for-backup-expose-5.7.

george-lorch: The rules AFAIK are common, but this is an area that not a lot of people tread into. Basically any change to the core server or core engines needs a blueprint or bug. This often means that you end up with both JIRA issues and launchpad blueprint/bug for the same issue. This comes into play with TokuDB, BakupPlugin, MyRocks, etc...

george-lorch: As Laurynas said, sometimes if the change is trivial, check with him to decide if a bp is really necessary.

vlad-lesin:
Added storing master state and some refactoring, see here https://github.com/percona/percona-server/pull/1372 .

PR #1330, opened 2017-01-27
Fix bug 1659548 (innodb_buffer_pool_evict_uncompressed might attempt …

laurynas-biveinis:
…to process non-unzip_LRU page, crashing debug build)

Fix iteration logic in innodb_buffer_pool_evict_uncompressed: if
buf_LRU_free_page returned true, relock the LRU list mutex, restart
the LRU list iteration, and, since it's restarted, reset the all pages
evicted flag back to true. If that function returned false, the LRU
list mutex has never been released, and we can safely use prev_block
pointer to iterate over the list.

http://jenkins.percona.com/job/percona-server-5.6-param/1612/

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -16319,15 +16319,17 @@ innodb_buffer_pool_evict_uncompressed(void)
 			ut_ad(block->page.in_LRU_list);
 
 			mutex_enter(&block->mutex);
-			if (!buf_LRU_free_page(&block->page, false)) {
-				mutex_exit(&block->mutex);
-				all_evicted = false;
-			} else {
-				mutex_exit(&block->mutex);
-				mutex_enter(&buf_pool->LRU_list_mutex);
-			}
+			all_evicted = buf_LRU_free_page(&block->page, false);
+			mutex_exit(&block->mutex);
 
-			block = prev_block;
+			if (all_evicted) {
+
+				mutex_enter(&buf_pool->LRU_list_mutex);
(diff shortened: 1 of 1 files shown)


robgolebiowski:
LGTM

storage/innobase/handler/ha_innodb.cc
```diff
@@ -16319,15 +16319,17 @@ innodb_buffer_pool_evict_uncompressed(void)
 			ut_ad(block->page.in_LRU_list);
 
 			mutex_enter(&block->mutex);
-			if (!buf_LRU_free_page(&block->page, false)) {
-				mutex_exit(&block->mutex);
-				all_evicted = false;
-			} else {
-				mutex_exit(&block->mutex);
+			all_evicted = buf_LRU_free_page(&block->page, false);
```
percona-ysorokin, line 16322: No test case because the problem was originally reveled by existing `xtradb_compressed_columns_consistency_barracuda_compressed`, right?

laurynas-biveinis: No testcase because I'm not sure how to write one: we'd have have at least two unzip_LRU pages with know IDs, DEBUG_SYNC stop here after the first one, do something remove the 2nd one from unzip_LRU - looks extremely laborious. Maybe I just don't know how to write a simpler one. It certainly helps that xtradb_compressed_columns_* stress-test this.

percona-ysorokin (approved):
LGTM

robgolebiowski (commented):
LGTM

PR #1333, opened 2017-01-27
Fix bug 1626545 (Test perfschema.selects is unstable)

laurynas-biveinis:
The unstable test bit joins performance_schema.events_waits_current
with performance_schema.events_waits_history. Such join does not have
meaningful semantics, as (thread_id; event_id) pairs in
events_waits_current are not duplicated in events_waits_history but
rather moved. So, in order for the testcase to pass, it relies that
only event_id is joined on, thus it's likely that different threads
have different IDs, and so a lower event_id by one thread in
events_waits_current will be already logged by another one that had
more events in events_waits_history. But it appears that it is not
guaranteed, with the join returning empty set.

Fix by removing events_waits_current from the join. Also remove SHOW
STATUS LIKE 'perf%' which was added earlier for diagnostics of this
bug.

http://jenkins.percona.com/job/percona-server-5.6-param/1618/

### DIFF
--- a/mysql-test/suite/perfschema/r/selects.result
+++ b/mysql-test/suite/perfschema/r/selects.result
@@ -23,38 +23,12 @@ LIMIT 1;
 EVENT_ID
 [EVENT_ID]
 SELECT DISTINCT EVENT_ID
-FROM performance_schema.events_waits_current
-JOIN performance_schema.events_waits_history USING (EVENT_ID)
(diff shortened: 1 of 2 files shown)


percona-ysorokin (approved):
LGTM

robgolebiowski:
LGTM

robgolebiowski (approved):
LGTM

PR #1337, opened 2017-01-30
Fix bug 1660239 (Test sys_vars.thread_cache_size_func is unstable)

laurynas-biveinis:
The testcase disconnects some threads, and waits for them to disappear
from INFORMATION_SCHEMA.PROCESSLIST before checking the Threads_cached
status variable. This introduces a race condition, as a thread will be
removed from PROCESSLIST before caching it.

Fix by waiting for the Threads_cached status variable to become equal
to its intended value instead.

http://jenkins.percona.com/job/percona-server-5.6-param/1627/

### DIFF
--- a/mysql-test/suite/sys_vars/t/thread_cache_size_func.test
+++ b/mysql-test/suite/sys_vars/t/thread_cache_size_func.test
@@ -75,12 +75,15 @@ DISCONNECT conn4;
 # Checking the status
 #
 
-# Wait until all disconnects ready
-let $wait_condition= SELECT COUNT(*)= 1 FROM INFORMATION_SCHEMA.PROCESSLIST;
---source include/wait_condition.inc
+--echo # Threads_cached must become 3
+--let $status_var= Threads_cached
+--let $status_var_value= 3
+--source include/wait_for_status_var.inc
 
-SHOW STATUS LIKE 'Threads_cached';
---echo 3 Expected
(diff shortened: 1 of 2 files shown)


mysql-test/suite/sys_vars/t/thread_cache_size_func.test
```diff
@@ -75,12 +75,10 @@ DISCONNECT conn4;
...
-# Wait until all disconnects ready
-let $wait_condition= SELECT COUNT(*)= 1 FROM INFORMATION_SCHEMA.PROCESSLIST;
---source include/wait_condition.inc
-
-SHOW STATUS LIKE 'Threads_cached';
---echo 3 Expected
+--echo # Threads_cached must become 3
+--let $status_var= Threads_cached
+--let $status_var_value= 3
+--source include/wait_for_status_var.inc
```
robgolebiowski, line 82: After the Threads_cached variable becomes 3 - it could be useful to check if SELECT COUNT(*)= 1 FROM INFORMATION_SCHEMA.PROCESSLIST;

robgolebiowski (changes requested):

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.6-param/1661/

robgolebiowski (approved):
LGTM

PR #1344, opened 2017-01-30
MYR-4 : myrocks #includes across engines to pull ut0counter.h from in…

george-lorch:
…nodb

Facebook MySQL copied(not moved) ./storage/innobase/include/ut0counter.h to
  ./include and altered/removed the 'InnoDB-isms' from it. They also altered
  some behaviors in this implementation. InnoDB still maintains its own copy
  of this file. MyRocks makes use of this counter class and templates.
- Rather than make the same copy into the same location, I think it best to
  copy their (Facebook) ut0counter.h from their ./include into our
  ./storage/rocksdb, rename it slightly to ib_ut0counter.h, and add an
  upstream merge 'watch' on this file so we are sure to properly merge down
  any subsequent changes.
- Altered CMakeLists.txt to detect presence of sched_getcpu.
- Altered uses of ib_counter to use new indexer provided by MyRocks version
  of the ib_counter.

### DIFF
--- a/storage/rocksdb/CMakeLists.txt
+++ b/storage/rocksdb/CMakeLists.txt
@@ -11,6 +11,11 @@ IF (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/rocksdb/Makefile")
   MESSAGE(SEND_ERROR "Missing Makefile in rocksdb directory. Try \"git submodule update\".")
 ENDIF()
 
+CHECK_FUNCTION_EXISTS(sched_getcpu  HAVE_SCHED_GETCPU)
+IF(HAVE_SCHED_GETCPU)
+  ADD_DEFINITIONS(-DHAVE_SCHED_GETCPU=1)
+ENDIF()
+
 ## adds a compiler flag if the compiler supports it
 INCLUDE(CheckCCompilerFlag)
 INCLUDE(CheckCXXCompilerFlag)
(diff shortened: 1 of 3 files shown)


laurynas-biveinis (approved):

laurynas-biveinis:
Have you confirmed that making this change results in the right file being compiled for Rocks?
Maybe there are some internal InnoDB include paths that can be removed from RocksDB CMake then?

laurynas-biveinis (commented):
See questions in the first tab

laurynas-biveinis (commented):
Same questions as for 5.6

george-lorch:
Well, yes. If including the InnoDB version of the file, MyRocks fails to link due to missing symbols from InnoDB (os_thread_get_curr_id). The Facebook MySQL version of the file changes that to use the MySQL provided get_curr_thread_id.

In answering this though I realized that this PR is incomplete. One of the reasons they changed this file is to add a new indexer that is based on sched_getcpu and to use that indexer when available in ha_rocksdb.cc. So I need to go back and re-add those changes that got caught up between the November 2016 merge and this patch.

george-lorch:
Also, if you are concerned over file name conflicts, I'll rename the MyRocks copy slightly just to ensure the correct file is in use.

george-lorch:
OK, renamed and pulled in upstream changes to use more cache friendly indexer and renamed copy of ut0counter.h

laurynas-biveinis:
Rename sounds like a good very good idea

PR #1345, opened 2017-01-30
MYR-4 : myrocks #includes across engines to pull ut0counter.h from in…

george-lorch:
…nodb

- cherry-pick of 79a197d from 5.6
- Facebook MySQL copied(not moved) ./storage/innobase/include/ut0counter.h to
  ./include and altered/removed the 'InnoDB-isms' from it. They also altered
  some behaviors in this implementation. InnoDB still maintains its own copy
  of this file. MyRocks makes use of this counter class and templates.
- Rather than make the same copy into the same location, I think it best to
  copy their (Facebook) ut0counter.h from their ./include into our
  ./storage/rocksdb, rename it slightly to ib_ut0counter.h, and add an
  upstream merge 'watch' on this file so we are sure to properly merge down
  any subsequent changes.
- Altered CMakeLists.txt to detect presence of sched_getcpu.
- Altered uses of ib_counter to use new indexer provided by MyRocks version
  of the ib_counter.

### DIFF
--- a/storage/rocksdb/CMakeLists.txt
+++ b/storage/rocksdb/CMakeLists.txt
@@ -11,6 +11,11 @@ IF (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/rocksdb/Makefile")
   MESSAGE(SEND_ERROR "Missing Makefile in rocksdb directory. Try \"git submodule update\".")
 ENDIF()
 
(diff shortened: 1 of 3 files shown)


george-lorch:
cherry-pick 5.6 branch https://github.com/percona/percona-server/pull/1344

PR #1348, opened 2017-01-31
Fix bug 1660348 (Intermittent testcase check warning on

laurynas-biveinis:
innodb_stress.innodb_hugestress_blob)

Save and restore innodb_status_output variable around the
innodb_stress testcases in case InnoDB enables it internally due to an
e.g. long lock wait.

http://jenkins.percona.com/job/percona-server-5.6-param/1631/


robgolebiowski (approved):
LGTM

PR #1356, opened 2017-02-02
MYR-26 : rocksdb.slow_query_log fails using unsupported --log-slow-extra

george-lorch:
- Removed test option for unsupported option
- Ensured test does not require MyRocks as default storage engine
- Touched up awk file to not look for new slow log fields
- Re-recorded result

### DIFF
--- a/mysql-test/suite/rocksdb/r/slow_query_log.result
+++ b/mysql-test/suite/rocksdb/r/slow_query_log.result
@@ -1,6 +1,5 @@
 SET @cur_long_query_time = @@long_query_time;
 SET @@long_query_time = 600;
-DROP TABLE IF EXISTS t1;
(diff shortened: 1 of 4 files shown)


laurynas-biveinis (approved):
LGTM, but curious, which part of the change does "Ensured test does not require MyRocks as default storage engine"?

george-lorch:
A visual/manual review to ensure that the test either properly specifies the required engine on CREATE TABLE and other statement or otherwise specifies default storage engine in the test or .opt file.

I am starting to try to 'clear' tests and get them using have_rocksdb.inc instead of have_rocksdb_as_default.inc so that we can eventually get mtr to run without needing --mysqld=--plugin-load=<...> --mysqld=--default-storage-engine=rocksdb incantation for running MyRocks suites.

PR #1359, opened 2017-02-02
Fix LP1661123

EvgeniyPatlan:


scripts/mysqld_safe.sh
```diff
@@ -694,8 +694,13 @@ safe_mysql_unix_port=${mysql_unix_port:-${MYSQL_UNIX_PORT:-@MYSQL_UNIX_ADDR@}}
 mysql_unix_port_dir=`dirname $safe_mysql_unix_port`
 if [ ! -d $mysql_unix_port_dir ]
 then
-  log_error "Directory '$mysql_unix_port_dir' for UNIX socket file don't exists."
-  exit 1
+  if [ ! -h $mysql_unix_port_dir ];
+  then
+    install -d -m 0755 -o $user $mysql_unix_port_dir
+  else 
+    log_error "Directory '$mysql_unix_port_dir' for UNIX socket file don't exists."
```
laurynas-biveinis, line 701: "does not exist", as pointed out in the bug report too

scripts/mysqld_safe.sh
```diff
@@ -694,8 +694,13 @@ safe_mysql_unix_port=${mysql_unix_port:-${MYSQL_UNIX_PORT:-@MYSQL_UNIX_ADDR@}}
 mysql_unix_port_dir=`dirname $safe_mysql_unix_port`
 if [ ! -d $mysql_unix_port_dir ]
 then
-  log_error "Directory '$mysql_unix_port_dir' for UNIX socket file don't exists."
-  exit 1
+  if [ ! -h $mysql_unix_port_dir ];
```
laurynas-biveinis, line 697: Is this symlink check safe?

EvgeniyPatlan: install will replace any existent link and create new directory

PR #1360, opened 2017-02-03
Fix bug 1660591 (5.7 range optimizer crash)

laurynas-biveinis:
After upstream commit [1], which fixed
https://bugs.mysql.com/bug.php?id=70236, SEL_ARG::increment_use_count
became more correct in propagating reference counts. This included
reference decreases too, and exposed a missing reference count bump in
and_all_keys. Because of this, there was a SEL_ARG object whose count
dropped to zero prematurely, and was pointed to from a different key
part. A new interval later was inserted through this new pointer,
coloring the original RB-tree root node red. Then the tree was
inserted to through the old pointer, resulting in a crash due to
RB-tree invariant violation above.

Fix by bumping reference count by one more in and_all_keys.

http://jenkins.percona.com/job/mysql-5.7-param/642/

### DIFF
--- a/sql/opt_range.cc
+++ b/sql/opt_range.cc
@@ -742,7 +742,7 @@ public:
   SEL_ARG *rb_insert(SEL_ARG *leaf);
   friend SEL_ARG *rb_delete_fixup(SEL_ARG *root,SEL_ARG *key, SEL_ARG *par);
 #ifndef DBUG_OFF
-  friend int test_rb_tree(SEL_ARG *element,SEL_ARG *parent);
+  friend int test_rb_tree(const SEL_ARG *element, const SEL_ARG *parent);
 #endif
   bool test_use_count(SEL_ARG *root);
   SEL_ARG *first();
@@ -781,6 +781,8 @@ public:
     {
       if (cur_selarg->next_key_part)
       {
+        DBUG_ASSERT(count >= 0
+                    || (long)cur_selarg->next_key_part->use_count >= count);
         cur_selarg->next_key_part->use_count+= count;
         cur_selarg->next_key_part->increment_use_count(count);
       }
@@ -8024,14 +8026,15 @@ static SEL_ARG *
 and_all_keys(RANGE_OPT_PARAM *param, SEL_ARG *key1, SEL_ARG *key2, 
              uint clone_flag)
 {
+  DBUG_ASSERT(key1->part < key2->part);
+
   SEL_ARG *next;
   ulong use_count=key1->use_count;
 
-  if (key1->elements != 1)
-  {
-    key2->use_count+=key1->elements-1; //psergey: why we don't count that key1 has n-k-p?
-    key2->increment_use_count((int) key1->elements-1);
-  }
+  DBUG_ASSERT(key1->elements > 0);
+  key2->use_count+= key1->elements;
+  key2->increment_use_count(key1->elements);
+
   if (key1->type == SEL_ARG::MAYBE_KEY)
   {
     // See todo for left/right pointers
@@ -9270,8 +9273,8 @@ SEL_ARG *rb_delete_fixup(SEL_ARG *root,SEL_ARG *key,SEL_ARG *par)
 
 #ifndef DBUG_OFF
 	/* Test that the properties for a red-black tree hold */
-
-int test_rb_tree(SEL_ARG *element,SEL_ARG *parent)
+static
+int test_rb_subtree(const SEL_ARG *element, const SEL_ARG *parent)
 {
   int count_l,count_r;
 
@@ -9294,8 +9297,8 @@ int test_rb_tree(SEL_ARG *element,SEL_ARG *parent)
     sql_print_error("Wrong tree: Found right == left");
     return -1;
   }
-  count_l=test_rb_tree(element->left,element);
-  count_r=test_rb_tree(element->right,element);
+  count_l=test_rb_subtree(element->left,element);
+  count_r=test_rb_subtree(element->right,element);
   if (count_l >= 0 && count_r >= 0)
   {
     if (count_l == count_r)
@@ -9304,6 +9307,16 @@ int test_rb_tree(SEL_ARG *element,SEL_ARG *parent)
 	    count_l,count_r);
   }
   return -1;					// Error, no more warnings
+}
+
+int test_rb_tree(const SEL_ARG *element, const SEL_ARG *parent)
+{
+  if (element->color == SEL_ARG::RED)
+  {
+    sql_print_error("Wrong tree: root node is red");
+    return -1;
+  }
+  return test_rb_subtree(element, parent);
 }
 #endif
 
--- a/mysql-test/r/bug84736.result
+++ b/mysql-test/r/bug84736.result
@@ -1,0 +1,24 @@
+#
+# Bug 84736: 5.7 range optimizer crash
+#
+CREATE TABLE t1 (
+col1 INT(11) NOT NULL,
+col2 INT(11) NOT NULL,
+col3 TEXT,
+PRIMARY KEY (col1, col2)
+);
+INSERT INTO t1 VALUES (2, 1, "a");
+INSERT INTO t1 VALUES (3, 10, "b");
+INSERT INTO t1 VALUES (1, 2, "c");
+INSERT INTO t1 VALUES (100, 100, "x");
+SELECT * FROM t1
+WHERE
+(col1 = 2 AND col2 = 1) OR
+(col1 IN (3, 100) AND col2 = 10) OR
+(col1 = 1 AND col2 = 2) OR
+(col1 IN (1, 2, 3) AND col2 = 2);
+col1	col2	col3
+1	2	c
+2	1	a
+3	10	b
+DROP TABLE t1;
--- a/mysql-test/t/bug84736.test
+++ b/mysql-test/t/bug84736.test
@@ -1,0 +1,24 @@
+--echo #
+--echo # Bug 84736: 5.7 range optimizer crash
+--echo #
+
+CREATE TABLE t1 (
+       col1 INT(11) NOT NULL,
+       col2 INT(11) NOT NULL,
+       col3 TEXT,
+       PRIMARY KEY (col1, col2)
+);
+
+INSERT INTO t1 VALUES (2, 1, "a");
+INSERT INTO t1 VALUES (3, 10, "b");
+INSERT INTO t1 VALUES (1, 2, "c");
+INSERT INTO t1 VALUES (100, 100, "x");
+
+SELECT * FROM t1
+       WHERE
+       (col1 = 2 AND col2 = 1) OR
+       (col1 IN (3, 100) AND col2 = 10) OR
+       (col1 = 1 AND col2 = 2) OR
+       (col1 IN (1, 2, 3) AND col2 = 2);
+
+DROP TABLE t1;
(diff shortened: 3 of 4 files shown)


sql/opt_range.cc
```diff
@@ -8029,8 +8031,8 @@ and_all_keys(RANGE_OPT_PARAM *param, SEL_ARG *key1, SEL_ARG *key2,
 
   if (key1->elements != 1)
```
gl-sergei: What is this condition for? Previously it looked like we wanted to avoid invocation of `key2->increment_use_count(0)` but now we skipping `key2->increment_use_count(1)`?

george-lorch:
As discussed in standup meeting, I reviewed some, but this is the first time I have ever laid eyes on this code, so my review perhaps may not be worth very much. It seems based on the comment by Sergei P that you are in the correct general area, why would key2 not take the full count of key1? As mentioned, the worst case would be retaining extra keys through the duration of the query, holding more memory for perhaps keys that would not be full parts of the result set. IMHO that is far superior to returning incorrect query results.

vlad-lesin:
Your changes in and_all_keys() look reasonable only for the case if key1->part < key2->part. Otherwise we should increase use_count by element-1 to take into account that after the operation SEL_TREE::keys[] will point to key1 but not to key2, and we should decrement reference counter for key2. To be absolutely sure I would add assert(key1->part < key2->part) before changing use_count in  and_all_keys().

I have not yet finished the review. I would like to revert your code changes and repeat the issue with your test and look how it goes in debugger to fully understand the case, and play a little bit with WHERE clause in the test, I would add the third smallest keypart and push condition in the end of WHERE clause to be sure and_all_keys() is called only for the case if key1->part < key2->part.

laurynas-biveinis:
Implemented @gl-sergei and @vlad-lesin suggestions

http://jenkins.percona.com/job/mysql-5.7-param/664/

vlad-lesin:
Laurynas, I removed your code changes and started your test. I set breakpoint on and_all_keys and printed the content of key1 and key2. This is what I got:

(gdb) p *key1
$1 = {<Sql_alloc> = {<No data fields>}, min_flag = 0 '\000', max_flag = 0 '\000', maybe_flag = 0 '\000', part = 0 '\000', maybe_null = 0 '\000', 
  rkey_func_flag = HA_READ_INVALID, elements = 1, use_count = 0, field = 0x7fffa400d2c0, min_value = 0x7fffa4002a48 "\002", max_value = 0x7fffa4002a48 "\002", 
  left = 0x2ec9040 <null_element>, right = 0x2ec9040 <null_element>, next = 0x0, prev = 0x0, parent = 0x0, next_key_part = 0x0, color = SEL_ARG::BLACK, 
  type = SEL_ARG::KEY_RANGE}
(gdb) p *key2
$2 = {<Sql_alloc> = {<No data fields>}, min_flag = 0 '\000', max_flag = 0 '\000', maybe_flag = 0 '\000', part = 1 '\001', maybe_null = 0 '\000', 
  rkey_func_flag = HA_READ_INVALID, elements = 1, use_count = 1, field = 0x7fffa400d378, min_value = 0x7fffa4002b28 "\001", max_value = 0x7fffa4002b28 "\001", 
  left = 0x2ec9040 <null_element>, right = 0x2ec9040 <null_element>, next = 0x0, prev = 0x0, parent = 0x0, next_key_part = 0x0, color = SEL_ARG::BLACK, 
  type = SEL_ARG::KEY_RANGE}

(gdb) p *key1
$3 = {<Sql_alloc> = {<No data fields>}, min_flag = 0 '\000', max_flag = 0 '\000', maybe_flag = 0 '\000', part = 0 '\000', maybe_null = 0 '\000', 
  rkey_func_flag = HA_READ_INVALID, elements = 2, use_count = 0, field = 0x7fffa400d2c0, min_value = 0x7fffa4002c08 "\003", max_value = 0x7fffa4002c08 "\003", 
  left = 0x2ec9040 <null_element>, right = 0x7fffa4002cf0, next = 0x7fffa4002cf0, prev = 0x0, parent = 0x0, next_key_part = 0x0, color = SEL_ARG::BLACK, 
  type = SEL_ARG::KEY_RANGE}
(gdb) p *key2
$4 = {<Sql_alloc> = {<No data fields>}, min_flag = 0 '\000', max_flag = 0 '\000', maybe_flag = 0 '\000', part = 1 '\001', maybe_null = 0 '\000', 
  rkey_func_flag = HA_READ_INVALID, elements = 1, use_count = 1, field = 0x7fffa400d378, min_value = 0x7fffa4002dc8 "\n", max_value = 0x7fffa4002dc8 "\n", 
  left = 0x2ec9040 <null_element>, right = 0x2ec9040 <null_element>, next = 0x0, prev = 0x0, parent = 0x3e72017, next_key_part = 0x0, color = SEL_ARG::BLACK, 
  type = SEL_ARG::KEY_RANGE}

(gdb) p *key1
$5 = {<Sql_alloc> = {<No data fields>}, min_flag = 0 '\000', max_flag = 0 '\000', maybe_flag = 0 '\000', part = 0 '\000', maybe_null = 0 '\000', 
  rkey_func_flag = HA_READ_INVALID, elements = 1, use_count = 0, field = 0x7fffa400d2c0, min_value = 0x7fffa4002ea8 "\001", max_value = 0x7fffa4002ea8 "\001", 
  left = 0x2ec9040 <null_element>, right = 0x2ec9040 <null_element>, next = 0x0, prev = 0x0, parent = 0x0, next_key_part = 0x0, color = SEL_ARG::BLACK, 
  type = SEL_ARG::KEY_RANGE}
(gdb) p *key2
$6 = {<Sql_alloc> = {<No data fields>}, min_flag = 0 '\000', max_flag = 0 '\000', maybe_flag = 0 '\000', part = 1 '\001', maybe_null = 0 '\000', 
  rkey_func_flag = HA_READ_INVALID, elements = 1, use_count = 1, field = 0x7fffa400d378, min_value = 0x7fffa4002f88 "\002", max_value = 0x7fffa4002f88 "\002", 
  left = 0x2ec9040 <null_element>, right = 0x2ec9040 <null_element>, next = 0x0, prev = 0x0, parent = 0x0, next_key_part = 0x0, color = SEL_ARG::BLACK, 
  type = SEL_ARG::KEY_RANGE}

(gdb) p *key1
$7 = {<Sql_alloc> = {<No data fields>}, min_flag = 0 '\000', max_flag = 0 '\000', maybe_flag = 0 '\000', part = 0 '\000', maybe_null = 0 '\000', 
  rkey_func_flag = HA_READ_INVALID, elements = 3, use_count = 0, field = 0x7fffa400d2c0, min_value = 0x7fffa4003148 "\002", max_value = 0x7fffa4003148 "\002", 
  left = 0x7fffa4003070, right = 0x7fffa4003230, next = 0x7fffa4003230, prev = 0x7fffa4003070, parent = 0x0, next_key_part = 0x0, color = SEL_ARG::BLACK, 
  type = SEL_ARG::KEY_RANGE}
(gdb) p *key2
$8 = {<Sql_alloc> = {<No data fields>}, min_flag = 0 '\000', max_flag = 0 '\000', maybe_flag = 0 '\000', part = 1 '\001', maybe_null = 0 '\000', 
  rkey_func_flag = HA_READ_INVALID, elements = 1, use_count = 1, field = 0x7fffa400d378, min_value = 0x7fffa4003308 "\002", max_value = 0x7fffa4003308 "\002", 
  left = 0x2ec9040 <null_element>, right = 0x2ec9040 <null_element>, next = 0x0, prev = 0x0, parent = 0x0, next_key_part = 0x0, color = SEL_ARG::BLACK, 
  type = SEL_ARG::KEY_RANGE}

As you can see in all cases: 
1) key1->elements > 0
2) and what is more important key2->use_count = 1

Let's follow the algorithm for increasing use_count described in commit 951d0e82aa562f05c4b7859a2906aee7bc2e8639 (see https://bugs.launchpad.net/percona-server/+bug/1660591/comments/3).

I suppose key2->use_count = 1 because SEL_TREE::keys[] points to key2. And in this case the old code looks reasonable because:
1) we must decrement key2->use_count as SEL_TREE::keys[] does not point to key2, it points to key1 after the operation,
2) we must increase key2->use_count by the key1->elements as key->elements points to key2 now,

So if we take into account (1) and (2) then we must increase key2->use_count by key1->elements-1 if key1->elements != 1 and don't touch it if key1->elements = 1 because 1 - 1 = 0. 

I have not yet understood the case when key1->elements = 0, but it does not matter at the moment.

So, let's consider the simple example:
1) key1->elements = 1
2) key2->use_count = 1 (what is true as SEL_TREE::keys[] points to key2)

in this case after and_all_keys() execution we have:
1) key1->use_count = 1 (SEL_TREE::keys[] points to key1)
2) key2->use_count = 1 (only one element in key1 points to key2)

Again, this is true if we follow the algorithm described here https://bugs.launchpad.net/percona-server/+bug/1660591/comments/3, and code without your changes.

If we follow the code with your changes then the result for the above example will be:
1) key1->use_count = 1
2) key2->use_count = 2 <<<===  it's not the same value as should be if we follow this https://bugs.launchpad.net/percona-server/+bug/1660591/comments/3 algorithm.

So it looks like your changes are wrong. Or this https://bugs.launchpad.net/percona-server/+bug/1660591/comments/3 algorithm is wrong. For me the rules of increasing and decreasing in this algorithm look reasonable. But I have some doubts about the part about propagation, I would explore it more deeply.

The general question for me is the following. If we decrement key2->use_count(actually increase it by key1->elements-1 instead of key1->elements) because some pointer(SEL_TREE::keys[]) does not point to key2 so far because it points to key1, i.e. we don't do propagation for SEL_TREE::keys[] pointer, then why we do propagation for other pointers from key1 tree. And vice versa, if we do propagation for all pointers of the graph, then why we don't do propagation for SEL_TREE::keys[] pointers? It does not look logical.

laurynas-biveinis:
Vlad - I might be misunderstanding something, but look at the tree_and routine, which suggests that key2 is still pointed to the same after the routine, thus its use_count decrement would be incorrect?

vlad-lesin:
> Vlad - I might be misunderstanding something, but look at the tree_and routine, which suggests that key2 is still pointed to the same after the routine, thus its use_count decrement would be incorrect?

This is true, but this code at the end of  tree_and() looks like the code which merger tree2 into tree1 and after this tree2->keys[] is not used.

imerge_list_and_list(&tree1->merges, &tree2->merges);
DBUG_RETURN(tree1);

But, anyway, it looks like the question on my answer about references counter propagation of SEL_TREE::keys[] can be found in the following ctor:
```c++
SEL_TREE::SEL_TREE(SEL_TREE *arg, RANGE_OPT_PARAM *param):
  Sql_alloc(), keys(param->mem_root, param->keys), n_ror_scans(0)
{
...
 for (uint idx= 0; idx < param->keys; idx++)
  {
    if (arg->keys[idx])
    {
      keys[idx]= arg->keys[idx]->clone_tree(param);
      if (!keys[idx])
        break;
      keys[idx]->use_count++;
      keys[idx]->increment_use_count(1);
    }
    else
      keys[idx]= NULL;
  }
...
}
```

If this counter should be propagated then the code:
```c++
static SEL_ARG *
and_all_keys(RANGE_OPT_PARAM *param, SEL_ARG *key1, SEL_ARG *key2, 
             uint clone_flag)
{
 ...
  if (key1->elements != 0)
  {
    key2->use_count+= key1->elements;
    key2->increment_use_count(key1->elements);
  }
...
}
```
looks good to me.

So if we have the following trees:

tree1.keys[n]->key1
tree2.keys[m]->key2

then for key2 in and_all_keys() we should:
1) increase use_count by key1->elements
2) decrement use_count as key2.keys[m] do not refer to it so far (as tree2 will be merged to tree1 in tree_and() )
3) increment use_count as tree1.keys[n] refers to key1 which in turn refers to key2.

(2) and (3) result to 0, so there is no need to decrease use_count as it was before your changes.

So now it looks good to me.

PR #1363, opened 2017-02-03
Fix bug 1660255 (Test innodb.innodb_mysql is unstable)

laurynas-biveinis:
The testcase creates a table with a secondary table, deletes all rows
from it, and checks the query plan. The problem is that it expects a
range access method, which it gets most of the time, because purge has
not completed yet. However, sometimes purge completes before the
query, and then the access method becomes ref, failing the testcase.

If the testcase were stabilised by fully completing purge before the
query, the access plan would be a stable ref method. But the goal of
the testcase is to test specifically range optimiser, and specifically
with a just-cleared table. To avoid making this a deterministic but a
debug-only testcase, fix by removing EXPLAIN SELECT altogether, but
leaving its SELECT, so that the absence of the crash is confirmed
regardless of the query plan.

http://jenkins.percona.com/job/percona-server-5.6-param/1633/

### DIFF
--- a/mysql-test/suite/innodb/r/innodb_mysql.result
+++ b/mysql-test/suite/innodb/r/innodb_mysql.result
@@ -386,10 +386,6 @@ name	dept
 rs5	cs10
 rs5	cs9
 DELETE FROM t1;
-# Masking (#) number in "rows" column of the following EXPLAIN output, as it may vary (bug#47746).
-EXPLAIN SELECT DISTINCT t1.name, t1.dept FROM t1 WHERE t1.name='rs5';
-id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
-1	SIMPLE	t1	range	name	name	44	NULL	#	Using where; Using index for group-by
 SELECT DISTINCT t1.name, t1.dept FROM t1 WHERE t1.name='rs5';
 name	dept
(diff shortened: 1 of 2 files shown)


robgolebiowski (approved):
LGTM

mysql-test/suite/innodb/r/innodb_mysql.result
```diff
@@ -386,10 +386,7 @@ name	dept
 rs5	cs10
 rs5	cs9
 DELETE FROM t1;
-# Masking (#) number in "rows" column of the following EXPLAIN output, as it may vary (bug#47746).
 EXPLAIN SELECT DISTINCT t1.name, t1.dept FROM t1 WHERE t1.name='rs5';
```
percona-ysorokin: Why this line is still present in the `.result` file although `EXPLAIN SELECT` has been removed from the `.inc`?

laurynas-biveinis: Good catch - this is a --big-test, and the final patch version was not tested properly

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.6-param/1676/

percona-ysorokin (approved):
LGTM

PR #1368, opened 2017-02-03
Ps 5.7 myr merge november2016

george-lorch:
Merge upstream November 2016 commits from 35ad59416fee5d840ab2e9144cfc30e70306f64e through 0151b45ae15087e4339ee8214cd52ed1b545d5ac


laurynas-biveinis (changes requested):
Same comments as for 5.6

PR #1370, opened 2017-02-04

george-lorch:
- Added __STDC_FORMAT_MACROS to CMakeLists


george-lorch:
Still doesn't fix the issue...

PR #1378, opened 2017-02-07
Ps 5.7 myr merge december2016

george-lorch:


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/659/

laurynas-biveinis (changes requested):

PR #1380, opened 2017-02-07
Bug 1650321: Valgrind error on main.audit_log_filter_commands

gl-sergei:
Valgrind complained about `cmd' being used uninitialized. However, I
don't see a path when it could happen, since the first command was to
initialize it unconditionally.

Still this error message pointed me to the fact that we don't need `cmd'
at all, we can simply use `name' and `length' provided as arguments.

http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1674/

### DIFF
--- a/plugin/audit_log/filter.c
+++ b/plugin/audit_log/filter.c
@@ -360,17 +360,12 @@ void audit_log_set_exclude_commands(const char *val)
 */
 my_bool audit_log_check_command_included(const char *name, size_t length)
 {
-  command cmd;
   my_bool res;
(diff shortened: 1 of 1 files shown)


laurynas-biveinis (approved):
LGTM, I assume you checked that the Valgrind error goes away?

gl-sergei:
I did

laurynas-biveinis:
This probably caused https://bugs.launchpad.net/percona-server/+bug/1663251

PR #1387, opened 2017-02-08
Bug 1650322: Valgrind error on main.audit_log_threadpool

gl-sergei:
When connection created in thread-per-connection handle there is a
thd_prepare_connection which starts as following:

    bool rc;
    lex_start(thd);
    rc= login_connection(thd);
    MYSQL_AUDIT_NOTIFY_CONNECTION_CONNECT(thd);

In case of pool-of-threads handler, connection goes to
threadpool_add_connection which is missing the lex_start call.

    bool rc= login_connection(thd);
    MYSQL_AUDIT_NOTIFY_CONNECTION_CONNECT(thd);

The fix is to simply add missing lex_start invocation.

### DIFF
--- a/sql/threadpool_common.cc
+++ b/sql/threadpool_common.cc
@@ -214,6 +214,7 @@ int threadpool_add_connection(THD *thd)
 
   if (!setup_connection_thread_globals(thd))
   {
+    lex_start(thd);
     bool rc= login_connection(thd);
     MYSQL_AUDIT_NOTIFY_CONNECTION_CONNECT(thd);
     if (!rc)


gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1677/

sql/threadpool_common.cc
```diff
@@ -214,7 +214,9 @@ int threadpool_add_connection(THD *thd)
 
   if (!setup_connection_thread_globals(thd))
   {
-    bool rc= login_connection(thd);
+    bool rc;
```
laurynas-biveinis: lext_start(thd);

bool rc= login_connection(thd);

laurynas-biveinis (changes requested):

sql/threadpool_common.cc
gl-sergei: Does it conform MySQL coding style?

laurynas-biveinis: I have not found anything in https://dev.mysql.com/doc/internals/en/general-development-guidelines.html. Given the typical function length in the core server code declaring variables at first use helps to keep sanity

gl-sergei:
Maybe it is not there, but upstream code follows C practice for variable declarations. It is really a matter of personal taste. For me, declaring all variables all at once makes my program easier to understand. However, I'm not in mood to argue.

gl-sergei:
@laurynas-biveinis anything else needed for this PR?

laurynas-biveinis (approved):
Sorry, didn't know it was ready for review.

PR #1392, opened 2017-02-10
Add further diagnostics for bug 1661534 (Test innodb.percona_changed_…

laurynas-biveinis:
…pages is unstable)

Test innodb.percona_changes_pages breaks intermittently due to LSN
advacing during read-only workload. Previous diagnostics showed that
this happens due to InnoDB transaction system header rewrite in the
system tablespace. Now add diagnostics to confirm whether that rewrite
happens due to maximum InnoDB transaction ID advancing.

http://jenkins.percona.com/job/percona-server-5.6-param/1686/


robgolebiowski (approved):
LGTM

percona-ysorokin (approved):
LGTM

PR #1393, opened 2017-02-10
Add further diagnostics for bug 1661534 (Test innodb.percona_changed_…

laurynas-biveinis:
…pages is unstable)

Test innodb.percona_changes_pages breaks intermittently due to LSN
advacing during read-only workload. Previous diagnostics showed that
this happens due to InnoDB transaction system header rewrite in the
system tablespace. Now add diagnostics to confirm whether that rewrite
happens due to maximum InnoDB transaction ID advancing.

(cherry picked from commit 8b6eacae8a7e77c585d9b2372a2dbfc3a356a56b)

http://jenkins.percona.com/job/mysql-5.7-param/668/


percona-ysorokin (approved):
LGTM

PR #1394, opened 2017-02-10
Fix lp:1652912 "Test main.bug80134 fails under AddressSanitizer default options" (5.6)

percona-ysorokin:
Under address sanitizer (both 'gcc' and 'clang') unsuccessful attempts to
allocate very large memory blocks result in printing the following warning
to stderr:
'==<PID>== WARNING: AddressSanitizer failed to allocate <NNN> bytes'.
As the result this message is added to Percona Server error log, which in
turn makes MTR test cases fail.

Unfortunately, there is no way to suppress specific ASan warnings and the
only way to resolve this issue is to add a suppression on the MTR level.
Instead of calling 'call mtr.add_suppression("...")' in individual tests,
this commit adds a new record to 'mtr.global_suppressions' table as testing
for OOM is pretty common case among existing MTR tests.

As some MTR test cases check for OOM, it is necessary to instruct address
sanitizer to not terminate the process when an allocation of a very
large memory block is requested and return NULL as expected.
This is done by defining '__asan_default_options()' ASan special function
which sets 'allocator_may_return_null' option to 1.

(cherry picked from commit 74a6e63)

### DIFF
--- a/mysql-test/include/mtr_warnings.sql
+++ b/mysql-test/include/mtr_warnings.sql
@@ -251,6 +251,9 @@ INSERT INTO global_suppressions VALUES
  */
  ("Insecure configuration for --secure-file-priv:*"),
 
+ /* ASan memory allocation warnings */
+ ("==[0-9]*== WARNING: AddressSanitizer failed to allocate 0x[0-9a-f]+ bytes"),
+
  ("THE_LAST_SUPPRESSION")||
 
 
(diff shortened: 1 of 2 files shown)


percona-ysorokin:
Cherry picked as 5.5 GCA does not have `main.bug80134` test case yet.

percona-ysorokin:
Jenkins build links:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-asan-param/109/ (ASan)
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1688/ (5.6-param with MTR_ARGS=main.bug80134)

laurynas-biveinis (approved):

PR #1399, opened 2017-02-10
MYR-19 : webscale/facebook binlog group commit interfaces and interac…

george-lorch:
…ts with SE

         differently
- MyRocks 5.6 has equivalent of PS 5.7 interface with small variance, but is
  not compatible with PS 5.6 at all.
- For PS 5.6, stripped out uncommon functionality. Cherry picking to 5.7 should
  properly re-introduce the group commit logic.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -2432,8 +2432,6 @@ static std::string rdb_xid_to_string(const XID &src) {
   transactions.
 */
 static bool rocksdb_flush_wal(handlerton *const hton
-                              __attribute__((__unused__)),
-                              ulonglong target_lsn
                               __attribute__((__unused__))) {
   DBUG_ASSERT(rdb != nullptr);
   rocksdb_wal_group_syncs++;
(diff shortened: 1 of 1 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2474,13 +2472,6 @@ static int rocksdb_prepare(handlerton *const hton, THD *const thd,
       if (!tx->prepare(rdb_xid_to_string(xid))) {
         return HA_EXIT_FAILURE;
       }
-      if (thd->durability_property == HA_IGNORE_DURABILITY &&
```
laurynas-biveinis, line 2477: Can you please explain this snippet?

george-lorch: No I can not, this was part of the original upstream behavior/patch that integrated with Facebook MySQL group commit, which is similar to what is implemented in PS 5.7 but missing from PS 5.6, see https://github.com/facebook/mysql-5.6/commit/14a0d4a97c09b52fa7450e6a3d56ebe7ed193ab6

It may not be required at all for the purpose of PS 5.6

gl-sergei: `thd_store_lsn` is a part of facebook/mysql-5.6@a869c56d361bb44f46c0efeb11a8f03561676247

laurynas-biveinis (approved):

PR #1400, opened 2017-02-10
MYR-19 : webscale/facebook binlog group commit interfaces and interac…

george-lorch:
…ts with SE

         differently
- MyRocks 5.6 has equivalent of PS 5.7 interface with small variance, but is
  not compatible with PS 5.6 at all.
- For PS 5.6, stripped out uncommon functionality. Cherry picking to 5.7 should
  properly re-introduce the group commit logic.
- Cherry picked to 5.7 reverted the change so MyRocks interacts with
  thd_store_lsn during prepare.
- Cherry pick merging commit e05afd2bd340f70eeb5ce29e44f57573f352f650 from ps-5.6-MYR-19


laurynas-biveinis (approved):
LGTM but please see the question on https://github.com/percona/percona-server/pull/1399

PR #1401, opened 2017-02-10

george-lorch:
- Added __STDC_FORMAT_MACROS to CMakeLists
- Added include <cinttypes> to top of ha_rocksdb.h to ensure CMakeLists macro
  and PRI get picked up before other includes.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.h
+++ b/storage/rocksdb/ha_rocksdb.h
@@ -20,6 +20,7 @@
 #endif
 
 /* C++ standard header files */
+#include <cinttypes>
 #include <set>
 #include <string>
 #include <unordered_set>


laurynas-biveinis (approved):
Interesting, the internets say this should not be required: http://stackoverflow.com/questions/12497894/d-stdc-format-macros-gnu-compiler-option

george-lorch:
That may be correct and is leftover from when I was trying with inttypes.h and assuming it was still necessary with cinttypes. I will try w/o the -D... and just the cinttypes and see if it compiles properly. If so I'll drop the -D... and merge the change.

george-lorch:
OK, you are correct, the -D is not needed if cinttypes is used. I'll pop that change out, re-cherry-pick to 5.7 push and merge.

PR #1402, opened 2017-02-10

george-lorch:
- Added include <cinttypes> to top of ha_rocksdb.h to ensure CMakeLists macro
  and PRI get picked up before other includes.
- Cherry pick merging commit aa1ecdcf149289a17b193beb2bf822a7b2a3a03c from ps-5.6-MYR-30


laurynas-biveinis (approved):
http://stackoverflow.com/questions/12497894/d-stdc-format-macros-gnu-compiler-option

PR #1407, opened 2017-02-10
MYR-51 : rocksdb.rocksdb_cf_options missing upstream include file inc…

george-lorch:
…/restart_mysqld_with_options.inc

- Copied in upstream Facebook MySQL include file.

### DIFF
--- a/mysql-test/suite/rocksdb/t/rocksdb_cf_options.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_cf_options.test
@@ -1,4 +1,4 @@
---source include/have_rocksdb_as_default.inc
+--source include/have_rocksdb.inc
 
 let MYSQLD_DATADIR= `SELECT @@datadir`;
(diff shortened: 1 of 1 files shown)


george-lorch:
This one might need a launchpad blueprint...

laurynas-biveinis (changes requested):
This is already handled by our restart_mysqld.inc backport that introduced restart_parameters argument.

george-lorch:
OK, changed test case over to using PS restart_myslqd.inc with restart_parameters. Tests out fine. Also audited the test and verified it is good for have_rocksdb without as_default.

laurynas-biveinis (approved):

laurynas-biveinis:
This is going to need a re-record in 5.7

george-lorch:
It will definitely need re-recorded as part of "5.7 mtr - re-record tests that use restart_mysqld.inc due to new comment in output" and possibly others.

PR #1408, opened 2017-02-10
MYR-51 : rocksdb.rocksdb_cf_options missing upstream include file inc…

george-lorch:
…/restart_mysqld_with_options.inc

- Copied in upstream Facebook MySQL include file.
- Cherry pick merging commit 272fdd805ddeaaee5afc6946f7a9e47bbd25aa1a from ps-5.6-MYR-51


laurynas-biveinis (changes requested):
restart_mysqld.inc with restart_parameters should be used instead

PR #1409, opened 2017-02-11
MYR-54 : rocksdb.sys_vars fails due to missing include files

george-lorch:
- Copied in upstream files to new directory mysql-test/rocksdb.sys_vars/include
- Fixed up all rocksdb.sys_vars tests to use new locations

### DIFF
--- a/mysql-test/suite/rocksdb.sys_vars/include/correctboolvalue.inc
+++ b/mysql-test/suite/rocksdb.sys_vars/include/correctboolvalue.inc
@@ -1,0 +1,68 @@
+##
+# Maps true, True, TRUE, on, On, ON to 1
+# and false, False, FALSE, off, Off, OFF to 0
+#
+# $input the value of a boolean type
+# $output the value of int type
+##
+--let $int_value=$value
+if ($value==on)
+{
+  --let $int_value=1
+}
+
+if ($value==On)
+{
+  --let $int_value=1
+}
+
+if ($value==ON)
+{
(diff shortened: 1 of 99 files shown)


george-lorch:
Upstream patch sent as https://github.com/facebook/mysql-5.6/pull/529

mysql-test/suite/rocksdb.sys_vars/include/correctboolvalue.inc
```diff
@@ -0,0 +1,25 @@
+##
```
laurynas-biveinis, line 1: Could use a better header comment ("maps TRUE FALSE 1 0 to 1 0")

laurynas-biveinis: Does not handle ON OFF?

laurynas-biveinis (approved):
Very minor comments

mysql-test/suite/rocksdb.sys_vars/include/correctboolvalue.inc
george-lorch: Good ideas, probably I should re-do this to handle all common variants of on, ON, On, off, OFF, Off, true, TRUE, True, false, FALSE, and False and send it upstream.

george-lorch:
OK, touched up correctboolvalue.inc, please take a peek

laurynas-biveinis (approved):

PR #1410, opened 2017-02-11
MYR-54 : rocksdb.sys_vars fails due to missing include files

george-lorch:
- Copied in upstream files to new directory mysql-test/rocksdb.sys_vars/include
- Fixed up all rocksdb.sys_vars tests to use new locations
- Cherry pick merging commit 86acdd7442907aa69422802ca40568746e8ed504 from ps-5.6-MYR-54


laurynas-biveinis (approved):
Same as for 5.6

PR #1415, opened 2017-02-15
Fix bug 1664280 (Race condition between buffer pool page optimistic a…

laurynas-biveinis:
…ccess and eviction | InnoDB: Failing assertion: buf_page_in_file(bpage) in buf_page_make_young)

buf_page_optimistic_get is passed a buffer pool block pointer with
nothing locked. Thus it is possible, for example, for the pointed
block to get evicted from the buffer pool and a different data page to
be read in. This function does not detect this condition correctly:
first, it checked whether block state is BUF_BLOCK_FILE_PAGE,
buffer-pinned the block, latched it, and then compared the
modify_clock value. The modify_clock value mismatch catches the
condition of block going to the free list and back as a data page, but
after this caught, the block is unfixed before returning. However,
when buf_page_init_for_read was called for this same block, it reset
buf_fix_count to zero in buf_page_init_low, overwriting the effect of
the previous buf_block_buf_fix_inc, and making the return unfix result
in UINT32_MAX as the fix count.

Fix by catching the block pointer pointing to evicted and/or re-read
data page earlier in the function, before the buffer pool fix
attempt, together with BUF_BLOCK_FILE_PAGE check. This is correct
because buf_LRU_block_remove_hashed bumps the modify clock while under
buffer page mutex protection, in the same critical section as
BUF_BLOCK_FILE_PAGE -> BUF_BLOCK_REMOVE_HASH transition.

Also replace the buf_fix_count reset in buf_page_init_low with a release
build assert for zero. Fix buf_block_unfix assert to actually catch
the unsigned underflow, as the previous expression type was promoted
to a wider type and thus never became zero.

At the same time move all the side effects of buf_page_optimistic_get
after the final modify_clock check, to avoid needlessly marking wrong
pages as young or accessed. Also clarify in the comments and asserts
that buf_block_modify_clock_inc may be called with the buffer page
locked instead of the  LRU list mutex.

http://jenkins.percona.com/job/mysql-5.7-param/671/

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -4802,7 +4802,9 @@ buf_page_optimistic_get(
 
 	buf_page_mutex_enter(block);
 
-	if (UNIV_UNLIKELY(buf_block_get_state(block) != BUF_BLOCK_FILE_PAGE)) {
+	if (UNIV_UNLIKELY(block->modify_clock != modify_clock
+			  || (buf_block_get_state(block)
+			      != BUF_BLOCK_FILE_PAGE))) {
 
 		buf_page_mutex_exit(block);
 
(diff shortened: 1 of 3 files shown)


laurynas-biveinis (commented):
The fix is incomplete, please don't review yet

laurynas-biveinis:
Ready for review

http://jenkins.percona.com/job/mysql-5.7-param/672/

percona-ysorokin:
Did you make 2 separate commits on purpose?

laurynas-biveinis:
Ugh, forgot to squash. But the composite diff is correct for review

percona-ysorokin (approved):
LGTM if squashed

robgolebiowski (approved):
LGTM

PR #1418, opened 2017-02-17
MYR-68 : rocksdb.show_engine fails

george-lorch:
- Removed portion of the test and result irrelevant to Percona Server
  functionality.
- Audited test and changed to have_rocksdb, dropping as_default


george-lorch:
Passes local mtr.

laurynas-biveinis (approved):

PR #1422, opened 2017-02-18
MYR-70 : rocksdb.misc fails

george-lorch:
- Re-recorded test to add/remove columns different from Facebook MySQL


george-lorch:
Passed local mtr.

laurynas-biveinis (approved):

PR #1424, opened 2017-02-18
MYR-71 : rocksdb.select_for_update_skip_locked_nowait fails

george-lorch:
- Removed test as 'skip locked' does not exist in Percona Server and that is
  all this test is intentded to validate with MyRocks.


laurynas-biveinis (approved):
Because https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-1.html
"InnoDB: InnoDB now supports NOWAIT and SKIP LOCKED options with SELECT ... FOR SHARE and SELECT ... FOR UPDATE locking read statements.",
please add a note to
https://blueprints.launchpad.net/percona-server/+spec/initial-5.8-port
to restore the test when the time comes

PR #1425, opened 2017-02-18
MYR-71 : rocksdb.select_for_update_skip_locked_nowait fails

george-lorch:
- Removed test as 'skip locked' does not exist in Percona Server and that is
  all this test is intentded to validate with MyRocks.
- Cherry pick merging commit 71b49fa6597bdd675c33e7340d79079420144973 from ps-5.6-MYR-71


laurynas-biveinis (approved):
Shared comment with 5.6 re. 8.0 blueprint

PR #1426, opened 2017-02-18
MYR-60 : rocksdb.rocksdb_icp fails

george-lorch:
- This was broken during the re-record for MYR-13 here
  https://github.com/percona/percona-server/commit/757260fda1d16f05b6823d060c6dad19628ce35e
- Unsure how or why this made it into the result. This value comes from deep
  inside RocksDB (not MyRocks). I suspect there was some issue of either an
  unmatched submodule or perhaps something else. Either way, this fixes it to
  be the same as upstream Facebook MySQL 5.6.35 and on the current PS run and
  rocksdb_icp passes.


george-lorch:
Passes local mtr.

laurynas-biveinis (approved):

PR #1428, opened 2017-02-18
MYR-9 : myrocks requires use of new class in facebook mysql Regex_lis…

george-lorch:
…t_handler

MYR-52: rocksdb.collation missing upstream include file have_regex.inc
lp : Implement class Regex based on my_regex
- https://blueprints.launchpad.net/percona-server/+spec/regex-list-handler-5.6
- Implemented new class Regex that is a simple wrapper around my_regex
  which manages lifecycle, concurent access, compilation, matching, and cleanup
  of a regex. located in sql/sql_regex.h
- Altered MyRocks to remove Regex_list_handler entirely and use new Regex class
- Fixed rocksdb.collation test to remove requirement for have_full_regex
- Fixed rocksdb.collation test to validate supported regex and some bad regex,
  removing Regex_list_handler semantica that replaced some delimiter in the
  option stream(',') with '|' before passing to std::regex
- Fixed rocksdb.collation test to use mtr semantics to restart the server with
  an alternate log_error file and use mtr pattern match rather than exec grep.

### DIFF
--- a/mysql-test/suite/rocksdb/r/collation.result
+++ b/mysql-test/suite/rocksdb/r/collation.result
@@ -1,5 +1,3 @@
-SET @start_global_value = @@global.ROCKSDB_STRICT_COLLATION_EXCEPTIONS;
-DROP TABLE IF EXISTS t1;
 CREATE TABLE t1 (id INT primary key, value varchar(50), value2 varbinary(50), value3 text) engine=rocksdb charset utf8;
 DROP TABLE t1;
 CREATE TABLE t1 (id INT primary key, value varchar(50), value2 varbinary(50), value3 text, index(value)) engine=rocksdb charset utf8;
@@ -33,63 +31,7 @@ CREATE TABLE xt123 (id INT primary key, value varchar(50), index(value)) engine=
 DROP TABLE xt123;
 CREATE TABLE t123 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
 ERROR HY000: Unsupported collation on string indexed column test.t123.value Use binary collation (binary, latin1_bin, utf8_bin).
-SET GLOBAL rocksdb_strict_collation_exceptions="s.*,t.*";
-CREATE TABLE s1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE s1;
-CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE t1;
-CREATE TABLE u1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-ERROR HY000: Unsupported collation on string indexed column test.u1.value Use binary collation (binary, latin1_bin, utf8_bin).
 SET GLOBAL rocksdb_strict_collation_exceptions="s.*|t.*";
-CREATE TABLE s1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE s1;
-CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE t1;
-CREATE TABLE u1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-ERROR HY000: Unsupported collation on string indexed column test.u1.value Use binary collation (binary, latin1_bin, utf8_bin).
-SET GLOBAL rocksdb_strict_collation_exceptions=",s.*,t.*";
-CREATE TABLE s1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE s1;
-CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE t1;
-CREATE TABLE u1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-ERROR HY000: Unsupported collation on string indexed column test.u1.value Use binary collation (binary, latin1_bin, utf8_bin).
-SET GLOBAL rocksdb_strict_collation_exceptions="|s.*|t.*";
-CREATE TABLE s1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE s1;
-CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE t1;
-CREATE TABLE u1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-ERROR HY000: Unsupported collation on string indexed column test.u1.value Use binary collation (binary, latin1_bin, utf8_bin).
-SET GLOBAL rocksdb_strict_collation_exceptions="s.*,,t.*";
-CREATE TABLE s1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE s1;
-CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE t1;
-CREATE TABLE u1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-ERROR HY000: Unsupported collation on string indexed column test.u1.value Use binary collation (binary, latin1_bin, utf8_bin).
-SET GLOBAL rocksdb_strict_collation_exceptions="s.*||t.*";
-CREATE TABLE s1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE s1;
-CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE t1;
-CREATE TABLE u1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-ERROR HY000: Unsupported collation on string indexed column test.u1.value Use binary collation (binary, latin1_bin, utf8_bin).
-SET GLOBAL rocksdb_strict_collation_exceptions="s.*,t.*,";
-CREATE TABLE s1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE s1;
-CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE t1;
-CREATE TABLE u1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-ERROR HY000: Unsupported collation on string indexed column test.u1.value Use binary collation (binary, latin1_bin, utf8_bin).
-SET GLOBAL rocksdb_strict_collation_exceptions="s.*|t.*|";
-CREATE TABLE s1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE s1;
-CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE t1;
-CREATE TABLE u1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
-ERROR HY000: Unsupported collation on string indexed column test.u1.value Use binary collation (binary, latin1_bin, utf8_bin).
-SET GLOBAL rocksdb_strict_collation_exceptions="||||,,,,s.*,,|,,||,t.*,,|||,,,";
 CREATE TABLE s1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
 DROP TABLE s1;
 CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
@@ -107,7 +49,6 @@ ALTER TABLE t2 ADD INDEX(value);
 ERROR HY000: Unsupported collation on string indexed column test.t2.value Use binary collation (binary, latin1_bin, utf8_bin).
 DROP TABLE t2;
 SET GLOBAL rocksdb_strict_collation_exceptions="[a-b";
- Invalid pattern in strict_collation_exceptions: [a-b
 CREATE TABLE a (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
 ERROR HY000: Unsupported collation on string indexed column test.a.value Use binary collation (binary, latin1_bin, utf8_bin).
 SET GLOBAL rocksdb_strict_collation_exceptions="[a-b]";
@@ -117,12 +58,3 @@ CREATE TABLE c (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rock
 ERROR HY000: Unsupported collation on string indexed column test.c.value Use binary collation (binary, latin1_bin, utf8_bin).
 DROP TABLE a, b;
 SET GLOBAL rocksdb_strict_collation_exceptions="abc\\";
- Invalid pattern in strict_collation_exceptions: abc\
-CREATE TABLE abc (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
-ERROR HY000: Unsupported collation on string indexed column test.abc.value Use binary collation (binary, latin1_bin, utf8_bin).
-SET GLOBAL rocksdb_strict_collation_exceptions="abc";
-CREATE TABLE abc (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
-CREATE TABLE abcd (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
-ERROR HY000: Unsupported collation on string indexed column test.abcd.value Use binary collation (binary, latin1_bin, utf8_bin).
-DROP TABLE abc;
-SET GLOBAL rocksdb_strict_collation_exceptions=@start_global_value;
--- a/mysql-test/suite/rocksdb/t/collation.test
+++ b/mysql-test/suite/rocksdb/t/collation.test
@@ -1,11 +1,7 @@
---source include/have_rocksdb_as_default.inc
---source include/have_fullregex.inc
-
-SET @start_global_value = @@global.ROCKSDB_STRICT_COLLATION_EXCEPTIONS;
-
---disable_warnings
-DROP TABLE IF EXISTS t1;
---enable_warnings
+# tests the basic functionality of rocksdb_strict_collation,
+# rocksdb_strict_collation_exceptions, and my_regex functionality behind
+# rocksdb_strict_collation_exceptions
+--source include/have_rocksdb.inc
 
 # ci non-indexed column is allowed
 CREATE TABLE t1 (id INT primary key, value varchar(50), value2 varbinary(50), value3 text) engine=rocksdb charset utf8;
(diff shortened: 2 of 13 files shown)


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1709/

mysql-test/suite/rocksdb/r/collation.result
```diff
@@ -117,12 +60,11 @@ CREATE TABLE c (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rock
...
 DROP TABLE a, b;
 SET GLOBAL rocksdb_strict_collation_exceptions="abc\\";
- Invalid pattern in strict_collation_exceptions: abc\
 CREATE TABLE abc (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
 ERROR HY000: Unsupported collation on string indexed column test.abc.value Use binary collation (binary, latin1_bin, utf8_bin).
 SET GLOBAL rocksdb_strict_collation_exceptions="abc";
 CREATE TABLE abc (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
-CREATE TABLE abcd (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
-ERROR HY000: Unsupported collation on string indexed column test.abcd.value Use binary collation (binary, latin1_bin, utf8_bin).
+CREATE TABLE bcd (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
```
laurynas-biveinis: Why the table rename?

sql/sql_regex.h
```diff
@@ -0,0 +1,132 @@
...
+class Regex
+{
+private:
+  std::string m_pattern;
+  my_regex_t m_expr;
+  bool m_compiled;
+
+  mutable mysql_rwlock_t m_rwlock;
+
+  Regex(const Regex& other) {}
```
laurynas-biveinis: = delete

sql/sql_regex.h
```diff
@@ -0,0 +1,132 @@
...
+  Regex& operator=(const Regex& other)
```
laurynas-biveinis: = delete

sql/sql_regex.h
```diff
@@ -0,0 +1,132 @@
...
+  Regex() :
+    m_pattern(""),
+    m_compiled(false)
+  {
+    mysql_rwlock_init(0, &m_rwlock);
+  }
+
+  ~Regex()
+  {
+    mysql_rwlock_wrlock(&m_rwlock);
```
laurynas-biveinis: Destructor probably does not need to latch m_rwlock, as there would be a race condition between subsequent mysql_rwlock_destroy and any latching in a parallel thread anyway - by the time we get to the destructor the instance must be non shared already. I'd replace with a debug assert that the latch is free.

sql/sql_regex.h
```diff
@@ -0,0 +1,132 @@
...
+                         &my_charset_utf8_general_ci);
+      if (!error)
+      {
+        m_compiled = true;
+      }
+    }
+
+    mysql_rwlock_unlock(&m_rwlock);
+
+    return error == 0 ? true : false;
```
laurynas-biveinis: return error == 0

sql/sql_regex.h
```diff
@@ -0,0 +1,132 @@
...
+    // Make sure no one else changes the list while we are accessing it.
+    mysql_rwlock_rdlock(&m_rwlock);
+
+    // See if the table name matches the regex we have created
+    int found = my_regexec(&m_expr, str.c_str(), 0, (my_regmatch_t*)0, 0);
+
+    // Release the lock
+    mysql_rwlock_unlock(&m_rwlock);
+
+    return found ? false : true;
```
laurynas-biveinis: return !found

sql/sql_regex.h
```diff
@@ -0,0 +1,132 @@
...
+    m_pattern(""),
+    m_compiled(false)
+  {
+    mysql_rwlock_init(0, &m_rwlock);
+  }
+
+  ~Regex()
+  {
+    mysql_rwlock_wrlock(&m_rwlock);
+    if (m_compiled)
```
laurynas-biveinis: I'd probably make a small private helper method out ouf this if statement

sql/sql_regex.h
```diff
@@ -0,0 +1,132 @@
...
+    if (pattern)
+    {
+      m_pattern.assign(pattern);
+    }
+
+    if (!m_pattern.empty())
+    {
+      error = my_regcomp(&m_expr,
+                         m_pattern.c_str(),
+                         MY_REG_EXTENDED | MY_REG_NOSUB | MY_REG_ICASE,
```
laurynas-biveinis: MY_REG_ICASE suggests a subtle bug here: MySQL table name case sensitivity depends on the underlying filesystem and lower_case_table_names value, see https://dev.mysql.com/doc/refman/5.7/en/identifier-case-sensitivity.html.

To resolve, I'd add bool case_sensitive arg to constructor or here (whichever makes more sense), and work out the correct value at the caller. I'd add some MTR too (extend the existing one with lower_case_table_names = 1, and since this depends on FS, add a new testcase with include/have_case_sensitive_file_system.inc for l_c_t_n = 0. For absolute completeness yet another new testcase with include/have_case_insensitive_file_system and l_c_t_n = 2 would also be required, but we don't have a supported platform aka MacOS for that).

Since std::regex is case-sensitive by default, this also suggests that the upstream has the opposite bug here which should be reported to them (option broken if table names case-insensitive)

sql/sql_regex.h
```diff
@@ -0,0 +1,132 @@
...
+                         &my_charset_utf8_general_ci);
```
laurynas-biveinis: This sets off an alarm although I cannot show a particular bug: the rest of the server uses table_alias_charset for e.g. table name hash table collations. I'd use it too to be safe (by passing to the constructor)

sql/sql_regex.h
```diff
@@ -0,0 +1,132 @@
...
+    if (!m_compiled)
+    {
+      return false;
+    }
+
+    // Make sure no one else changes the list while we are accessing it.
+    mysql_rwlock_rdlock(&m_rwlock);
+
+    // See if the table name matches the regex we have created
+    int found = my_regexec(&m_expr, str.c_str(), 0, (my_regmatch_t*)0, 0);
```
laurynas-biveinis: 4th arg plain "nullptr"

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -473,8 +473,8 @@ static MYSQL_THDVAR_BOOL(
 
 static MYSQL_THDVAR_STR(
     read_free_rpl_tables, PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_MEMALLOC,
-    "List of tables that will use read-free replication on the slave "
-    "(i.e. not lookup a row during replication)",
+    "List of tables (using regex) that will use read-free replication on the "
```
laurynas-biveinis: "List" suggests commas :)
I'd word this "A regexp matching the tables ... " or something similar

sql/sql_regex.h
```diff
@@ -0,0 +1,132 @@
...
+  {
+    return *this;
+  }
+
+public:
+  Regex() :
+    m_pattern(""),
+    m_compiled(false)
+  {
+    mysql_rwlock_init(0, &m_rwlock);
```
laurynas-biveinis: Looks like you have removed PFS key support?

storage/rocksdb/rdb_utils.h
```diff
@@ -247,60 +248,6 @@ std::string rdb_hexdump(const char *data, const std::size_t data_len,
...
-  // See the list of bad patterns
-  const std::string& bad_pattern() const
-  {
-    return m_bad_pattern_str;
-  }
-};
-
-void warn_about_bad_patterns(const Regex_list_handler* regex_list_handler,
-                             const char *name);
+void warn_about_bad_patterns(const Regex *regex, const char *name);
```
laurynas-biveinis: const reference 1st arg?

laurynas-biveinis (changes requested):

laurynas-biveinis:
Does the option accept "database.table" syntax too?

mysql-test/suite/rocksdb/r/collation.result
george-lorch: Difference in regex behavior from 'traditional' to 'modern'. 'traditional' matches "abc" in the "abcd", modern does not.

sql/sql_regex.h
george-lorch: Doesn't compile under C++11.

sql/sql_regex.h
george-lorch: true enough, assertion is probably the better route here.

sql/sql_regex.h
laurynas-biveinis: Huh? I thought it was C++11

sql/sql_regex.h
george-lorch: Good idea, a 'reset' kind of thing.

sql/sql_regex.h
george-lorch: MyRocks is, but server is not, which is where this new source file resides. It is likely we will need this in server to introduce similar new option for RFR such as rfr_tables according to discussion in https://jira.percona.com/browse/MYR-6 where we are talking about moving bulk of RFR state tracking up into handler layer.

sql/sql_regex.h
george-lorch: Acually, full regex flag set really should be passed on compile and not just assumed as different users will have different requirements. I think maybe similar applies to charset. Upstream possibly has inverse issue but has consistent usage from specific areas so is likely not caring nor consistent about alternative users.

sql/sql_regex.h
laurynas-biveinis: Argh, of course. I believe in C++03 the canonical way is to declare private (already done) and then also omit { } so that a linker error on use would result

george-lorch: Ahh yes, I forgot about the linker trick and just relied on the compiler error of unaccessable member, both tools to ensure no implicit call.

sql/sql_regex.h
george-lorch: Yeah, initial attempt had PFS types not accessible, but should be now with proposed location and implementation.

sql/sql_regex.h
george-lorch: Question though, can we assume that the option/regex charset being compiled is the same as the table name used in the match which will then likely be table_alias_charset?

george-lorch:
Does not appear that database.table is supported, only table as option uses sql/sql_table filename_to_tablename which appears to only deal with table names not schema names.

sql/sql_regex.h
laurynas-biveinis: Good question, I'd look into how e.g. replicate-do-table is implemented

george-lorch: Looks like these rpl_do and rpl_ignore options are not treated to any charset, they are just placed into storage and later on when lookup/comparison time comes they use table_alias_charset just as you mentioned.

sql/sql_regex.h
george-lorch: mysql_rwlock has no functions to test/assert for lock state.

mysql-test/suite/rocksdb/r/collation.result
george-lorch: Actually, this is slightly incorrect. The original test is incorrect. A proper regex of "abc" should match "abcd" as "abc" is within "abcd".

Within MySQL REGEX:

> mysql> SELECT 'abcd' REGEXP 'abc';
> +---------------------+
> | 'abcd' REGEXP 'abc' |
> +---------------------+
> |                   1 |
> +---------------------+
> 1 row in set (0.00 sec)

std::regex should make this same match based on the documented syntax http://en.cppreference.com/w/cpp/regex/ecmascript but it does not.

george-lorch:
Please review again. I did not feel comfortable adding any test for lower_case_table_name=2 due to inability to test, record, or validate results. I would rather go with no test than a broken one.

mysql-test/suite/rocksdb/r/collation.result
laurynas-biveinis: A bug in std::regex might have serious consequences for the upstream, please let them know

sql/sql_regex.h
```diff
@@ -0,0 +1,144 @@
...
+  {
+    if (!m_compiled)
+    {
+      return false;
+    }
+
+    // Make sure no one else changes the list while we are accessing it.
+    mysql_rwlock_rdlock(&m_rwlock);
+
+    // See if the table name matches the regex we have created
```
laurynas-biveinis: not necessarily a "table name"

sql/sql_regex.h
```diff
@@ -0,0 +1,144 @@
...
+    int found = my_regexec(&m_expr, str.c_str(), 0, nullptr, 0);
+
+    // Release the lock
```
laurynas-biveinis: Comment too obvious?

mysql-test/suite/rocksdb/t/collation.test
```diff
@@ -166,16 +96,76 @@ CREATE TABLE c (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rock
...
 CREATE TABLE abcd (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
+CREATE TABLE ABCD (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
+DROP TABLE ABCD;
+
+--error ER_UNKNOWN_ERROR
+CREATE TABLE ZABC (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
+
+
+# test with server comparing in lower case
+--let $restart_parameters="restart: --lower_case_table_names=1"
```
laurynas-biveinis: Not having a lctn=2 test is OK - but there is no lctn=0 test in a separate testcase with include/have_case_sensitive_file_system.inc?

sql/sql_regex.h
```diff
@@ -0,0 +1,144 @@
...
+  Regex& operator=(const Regex& other);
+
+  void reset()
+  {
+    // assumes object is write locked where necessary
+    if (m_compiled)
+    {
+      m_pattern.clear();
+      my_regfree(&m_expr);
+      m_compiled = false;
```
laurynas-biveinis: No space before operator

sql/sql_regex.h
```diff
@@ -0,0 +1,144 @@
...
+  const std::string& pattern() const
+  {
+    return m_pattern;
+  }
+
+  // Compile the pattern into an expression
+  // see regex/my_regex.h for definition of flags
+  bool compile(const char *pattern, int flags, const CHARSET_INFO *charset)
+  {
+    int error = 0;
```
laurynas-biveinis: Likewise

sql/sql_regex.h
```diff
@@ -0,0 +1,144 @@
...
+      m_pattern.assign(pattern);
+    }
+
+    if (!m_pattern.empty())
+    {
+      error = my_regcomp(&m_expr,
+                         m_pattern.c_str(),
+                         flags,
+                         charset);
+      if (!error)
```
laurynas-biveinis: I believe this should be "m_compiled= !error" - if you have a previous pattern assigned, and pass a new invalid pattern, m_compiled cannot stay true.

sql/sql_regex.h
```diff
@@ -0,0 +1,144 @@
...
+    mysql_rwlock_wrlock(&m_rwlock);
+
+    reset();
+
+    if (pattern)
+    {
+      m_pattern.assign(pattern);
+    }
+
+    if (!m_pattern.empty())
```
laurynas-biveinis, line 106: What should happen to m_compiled value if m_pattern.empty() ?

sql/sql_regex.h
```diff
@@ -0,0 +1,144 @@
...
+    if (!m_compiled)
+    {
+      return false;
+    }
+
+    // Make sure no one else changes the list while we are accessing it.
+    mysql_rwlock_rdlock(&m_rwlock);
+
+    // See if the table name matches the regex we have created
+    int found = my_regexec(&m_expr, str.c_str(), 0, nullptr, 0);
```
laurynas-biveinis: No space before assignment operator

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3071,10 +3071,9 @@ static int rocksdb_init_func(void *const p) {
                    MY_MUTEX_INIT_FAST);
 
 #if defined(HAVE_PSI_INTERFACE)
-  rdb_collation_exceptions =
-      new Regex_list_handler(key_rwlock_collation_exception_list);
+  rdb_collation_exceptions = new Regex(key_rwlock_collation_exception_list);
```
laurynas-biveinis, line 3074: By now I am thoroughly confused whether MyRocks code is using MySQL code style or not, IIRC I have seen upstream pull requests moving both directions

laurynas-biveinis (changes requested):

sql/sql_regex.h
george-lorch: Crap, right, server code style...

sql/sql_regex.h
george-lorch: Contract with reset() is that upon completion m_compiled is false and m_pattern is empty.

sql/sql_regex.h
george-lorch: Again, same, contract with reset() is that upon completion m_compiled is false and m_pattern is empty.

sql/sql_regex.h
laurynas-biveinis: Ah, reset makes my comments invalid

sql/sql_regex.h
george-lorch: Correct, left over from upstream Regex_list_handler

sql/sql_regex.h
laurynas-biveinis: I'd probably still do this even if reset is done before for a straighter flow

sql/sql_regex.h
george-lorch: Left over from upstream Regex_list_handler

mysql-test/suite/rocksdb/t/collation.test
george-lorch: Well, the existing test is based on the default value of lctn which is 0. Honestly I do not like the idea of deep testing the regex functionality in this specific test case for a MyRocks option (there will be more), it is way out of scope. Unfortunately, there isn't much of another option outside of exposing some other debug only option that exists solely to test regex options vs table names.

laurynas-biveinis: Dependency on lctn = 0 is platform-specific, thus any tests for this should be moved to a specific lctn = 0 testcase, that gets skipped on some platforms, and only lctn-agnostic tests should remain in the main testcase. 

Having said that, this only needs to happen if we are to upstream at least some of the regex changes.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Yes, their code style is roughly "whatever I feel like today" but I believe it is close to google style guidelines. They did a major clang reformat recently but unfortunately did not place a .clang-format into their tree so we have some idea of what they style is and can 'enforce' it in our own changes. This is actually a good thing to ask them...

mysql-test/suite/rocksdb/t/collation.test
george-lorch: Again though, I bent a little to cover some of this, but I still think this is not appropriate to place the testing of what is basically a new sysvar type within a test for a this specific option. That implies then that we need to replicate all the test work we do here for every variable that uses this functionality. I am not at all opposed to testing but it must be done in the correct way and in the correct place otherwise we end up with more test fragmentation. By your suggestion, I now need a collation, collation_lctn_0, collation_lctn_1, and ideally a collation_lctn_2 test, all of which exist solely to test regex and char case which has nothing to do with the functionality of the strict collation check. We will also now need the same for the existing rocksdb read_free_rpl_tables since it uses the same facility (and eventually to be moved up into SQL layer as discussed in MYR-6), and we will need to continue the redundancy to every and all options that use this Regex facility. I do not believe that is at all reasonable. That being said, I can not offer an alternative as there is no other mtr accessible UI to test this specific regex functionality.

laurynas-biveinis: George, I don't think this is nearly as bad as you'd seem to see. This is simply about testing the interaction of two server variables, which does have different behavior depending on the settings. Were this random two other variables, we'd simply add some lines to a testcase and forget about it. But so it happens that lctn requires server restart and compatible FS, forcing the testcase split, and note, that the lctn_0, lctn_2 testcases would be very short each, as there is no need to re-test the whole core functionality. Yes, probably read_free_rpl_tables would need the same so that we know that this filter does actually DTRT on different FS and lctn values, but this is not different from testing any other option combination?

Also the above is what I see the ideal for this particular testing, which is a prerequisite for keeping the same cross-platform functionality as the upstream server. We don't really need to implement this due to our limited supported platform set, but as we touch things, I believe it's good to have this in mind and work towards that direction to some extent.

george-lorch:
OK, pushed with code format changes and such and two new tests for lctn that I think covers the gist of what you were looking for.

mysql-test/suite/rocksdb/t/collation.test
```diff
@@ -166,16 +94,12 @@ CREATE TABLE c (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rock
...
-CREATE TABLE abc (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
---error ER_UNKNOWN_ERROR
-CREATE TABLE abcd (id INT PRIMARY KEY, value varchar(50), index(value)) engine=rocksdb charset utf8;
-DROP TABLE abc;
+--let SEARCH_PATTERN=Invalid pattern in strict_collation_exceptions: abc\\\\
+--source include/search_pattern_in_file.inc
 
 # cleanup
-SET GLOBAL rocksdb_strict_collation_exceptions=@start_global_value;
+--let SEARCH_PATTERN=
```
laurynas-biveinis, line 102: No need to reset MTR variables, including environment MTR ones, at the end (no need to address this comment)

laurynas-biveinis (changes requested):
You mention new tests but I don't see them - forgotten git add?

george-lorch:
Grrr, son of a....

george-lorch:
And there are the new tests, fortunately I still had them sitting in a test/install directory.

laurynas-biveinis (approved):

PR #1429, opened 2017-02-18
MYR-9 : myrocks requires use of new class in facebook mysql Regex_lis…

george-lorch:
…t_handler

MYR-52: rocksdb.collation missing upstream include file have_regex.inc
lp : Implement class Regex based on my_regex
- https://blueprints.launchpad.net/percona-server/+spec/regex-list-handler-5.6
- Implemented new class Regex that is a simple wrapper around my_regex
  which manages lifecycle, concurent access, compilation, matching, and cleanup
  of a regex. located in sql/sql_regex.h
- Altered MyRocks to remove Regex_list_handler entirely and use new Regex class
- Fixed rocksdb.collation test to remove requirement for have_full_regex
- Fixed rocksdb.collation test to validate supported regex and some bad regex,
  removing Regex_list_handler semantica that replaced some delimiter in the
  option stream(',') with '|' before passing to std::regex
- Fixed rocksdb.collation test to use mtr semantics to restart the server with
  an alternate log_error file and use mtr pattern match rather than exec grep.

### DIFF
--- a/mysql-test/suite/rocksdb/r/collation.result
+++ b/mysql-test/suite/rocksdb/r/collation.result
@@ -1,5 +1,3 @@
-SET @start_global_value = @@global.ROCKSDB_STRICT_COLLATION_EXCEPTIONS;
-DROP TABLE IF EXISTS t1;
 CREATE TABLE t1 (id INT primary key, value varchar(50), value2 varbinary(50), value3 text) engine=rocksdb charset utf8;
 DROP TABLE t1;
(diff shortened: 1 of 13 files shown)


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/680/

laurynas-biveinis (changes requested):
Same as for 5.6

PR #1430, opened 2017-02-18
MYR-76 : rocksdb.i_s_ddl fails

george-lorch:
- Test failed due to non-deterministic ordering of a query output.
  Added --sorted_result and re-recorded.


mysql-test/suite/rocksdb/t/i_s_ddl.test
```diff
@@ -17,6 +17,7 @@ CREATE TABLE is_ddl_t2 (x INT, y INT, z INT,
                         PRIMARY KEY (z, y) COMMENT 'zy_cf',
                         KEY (x)) ENGINE = ROCKSDB;
 
+--sorted_result
```
laurynas-biveinis: Why not SQL ORDER BY?

george-lorch: IIRC I used ORDER BY a while back on some TokuDB issue only to get feedback that --sorted-result might be the more correct mtr semantic, can't remember where, even though I have continued to use ORDER BY to correct non-deterministic ordering for I_S type tables. For some reason, right or wrong, when fixing this, the thought of "should use --sorted-result" came to mind. IMHO it needs to be a rule that some ordering be specified for all non-keyed tables such as I_S and PFS tables as these have caused numerous mtr instabilities over time where results were the same but orders different for whatever 'under the hood' reasons. If you think natural ORDER BY is the way to go, I'm there with you and can easily change this around.

laurynas-biveinis: A potential ORDER BY issue would be lack of unique key and too many potentially equal columns, resulting in very long ORDER BY, and I don't know the TokuDB issue, which could have been something legitimate too. Otherwise I do prefer SQL over MTR sort

george-lorch: Agreed, I can not recall the specific issue other than I got to learn about --sorted-result and something about it being preferred mtr semantics. I think ORDER BY is much more specific and clear to the intention. Will change it up and re-push to the PR.

PR #1432, opened 2017-02-18

george-lorch:
… new comment in output

- Added all # restart echos via re-recording and selective editing.


george-lorch:
Passes local mtr on 5.7 on top of MYR-55 except for truncate_table3 that needs MYR-61 (i_s.global_status) result changes as well.

laurynas-biveinis (approved):

PR #1437, opened 2017-02-20
Ps 5.7 myr 55

george-lorch:


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -200,7 +193,7 @@ struct Rdb_open_tables_map {
   void init_hash(void) {
     (void)my_hash_init(&m_hash, my_core::system_charset_info, TABLE_HASH_SIZE,
                        0, 0, (my_hash_get_key)Rdb_open_tables_map::get_hash_key,
-                       0, 0);
+                       0, 0, 0); // TODO: instrument for PFS
```
laurynas-biveinis: Is there a MYR issue? Also strictly speaking nullptr

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1511,24 +1504,25 @@ class Rdb_transaction {
...
         char user_host_buff[MAX_USER_HOST_SIZE + 1];
-        make_user_name(thd, user_host_buff);
+        make_user_name(thd->security_context(), user_host_buff);
         // NO_LINT_DEBUG
         sql_print_warning("Got snapshot conflict errors: User: %s "
-                          "Query: %s",
-                          user_host_buff, thd->query());
+                          "Query: %.*s",
+                          user_host_buff,
+                          static_cast<int>(thd->query().length),
```
laurynas-biveinis: Why? Is the string not NUL-terminated?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2758,7 +2765,7 @@ class Rdb_trx_info_aggregator : public Rdb_tx_list_walker {
     DBUG_ASSERT(tx != nullptr);
 
     THD *const thd = tx->get_thd();
-    ulong thread_id = thd->thread_id;
+    ulong thread_id = thd->thread_id();
```
laurynas-biveinis: The correct type is my_thread_id, all versions

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2378,25 +2372,28 @@ static int rocksdb_close_connection(handlerton *const hton, THD *const thd) {
  * be used as a rocksdb transaction name
  */
 static std::string rdb_xid_to_string(const XID &src) {
-  DBUG_ASSERT(src.gtrid_length >= 0 && src.gtrid_length <= MAXGTRIDSIZE);
-  DBUG_ASSERT(src.bqual_length >= 0 && src.bqual_length <= MAXBQUALSIZE);
+  DBUG_ASSERT(src.get_gtrid_length() >= 0 &&
```
laurynas-biveinis, line 2373: I'd look into extracting a helper function out of these two asserts (all versions)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2785,9 +2792,8 @@ class Rdb_trx_info_aggregator : public Rdb_tx_list_walker {
       }
 
       std::string query_str;
-      LEX_STRING *const lex_str = thd_query_string(thd);
-      if (lex_str != nullptr && lex_str->str != nullptr) {
-        query_str = std::string(lex_str->str);
+      if (thd->query().str != nullptr) {
```
laurynas-biveinis, line 2793: This dropped the check "lex_str != nullptr", aka "thd->query() != nullptr" in 5.7. Maybe it can never be nullptr?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2378,25 +2372,28 @@ static int rocksdb_close_connection(handlerton *const hton, THD *const thd) {
...
   /*
    * expand formatID to fill 8 bytes if it doesn't already
    * then reinterpret bit pattern as unsigned and store in network order
    */
   uchar fidbuf[RDB_FORMATID_SZ];
-  int64 signed_fid8 = src.formatID;
+  int64 signed_fid8 = src.get_format_id();
   const uint64 raw_fid8 = *reinterpret_cast<uint64 *>(&signed_fid8);
   rdb_netbuf_store_uint64(fidbuf, raw_fid8);
   buf.append(reinterpret_cast<const char *>(fidbuf), RDB_FORMATID_SZ);
```
laurynas-biveinis, line 2389: Is const_cast not enough here? (yes this is unrelated to the PR, general code review)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3479,9 +3473,10 @@ Rdb_open_tables_map::get_table_handler(const char *const table_name) {
             &m_hash, reinterpret_cast<const uchar *>(table_name), length)))) {
     // Since we did not find it in the hash map, attempt to create and add it
     // to the hash map.
+    // TODO: instrument for PFS
     if (!(table_handler = reinterpret_cast<Rdb_table_handler *>(my_multi_malloc(
```
laurynas-biveinis, line 3476: Here and in all other my_ memory allocators - why reinterpret_cast instead of static_cast?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5608,11 +5608,9 @@ int ha_rocksdb::read_range_first(const key_range *const start_key,
       DBUG_RETURN(HA_ERR_LOCK_DEADLOCK);
     }
 
-    MYSQL_TABLE_IO_WAIT(m_psi, PSI_TABLE_FETCH_ROW, active_index, 0, {
-      result =
-          index_read_map_impl(table->record[0], start_key->key,
-                              start_key->keypart_map, start_key->flag, end_key);
-    })
+    result =
```
laurynas-biveinis, line 5610: Is MYSQL_TABLE_IO_WAIT dropped on purpose?

storage/rocksdb/ha_rocksdb.h
```diff
@@ -1045,10 +1047,10 @@ class ha_rocksdb : public my_core::handler {
...
       __attribute__((__warn_unused_result__));
 
-  my_bool register_query_cache_table(THD *const thd, char *const table_key,
-                                     uint key_length,
-                                     qc_engine_callback *const engine_callback,
-                                     ulonglong *const engine_data) override {
+  my_bool register_query_cache_table(THD *thd, char *table_key,
+                                     size_t key_length,
+                                     qc_engine_callback *engine_callback,
+                                     ulonglong *engine_data) override {
```
laurynas-biveinis: If "override" is not complaining, I'd keep the consts

storage/rocksdb/rdb_index_merge.cc
```diff
@@ -14,6 +14,10 @@
    along with this program; if not, write to the Free Software
    Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA */
 
+/* MySQL header files */
```
laurynas-biveinis: Probably should go below #include rdb_index_merge.h ?

storage/rocksdb/rdb_mutex_wrapper.cc
```diff
@@ -41,7 +41,7 @@ static const int64_t ONE_SECOND_IN_MICROSECS = 1000 * 1000;
 static const int64_t ONE_YEAR_IN_MICROSECS =
     ONE_SECOND_IN_MICROSECS * 60 * 60 * 24 * 365;
 
-Rdb_cond_var::Rdb_cond_var() { mysql_cond_init(0, &m_cond, nullptr); }
+Rdb_cond_var::Rdb_cond_var() { mysql_cond_init(0, &m_cond); }
```
laurynas-biveinis: // TODO PFS

storage/rocksdb/rdb_threads.cc
```diff
@@ -19,6 +19,9 @@
 #pragma implementation // gcc: Class implementation
 #endif
 
+/* MySQL header files */
```
laurynas-biveinis: Should be moved below include rdb_threads.h?

storage/rocksdb/rdb_threads.cc
```diff
@@ -61,17 +64,6 @@ int Rdb_thread::create_thread(const std::string &thread_name
   int err = mysql_thread_create(background_psi_thread_key, &m_handle, nullptr,
                                 thread_func, this);
 
-  if (!err) {
```
laurynas-biveinis, line 64: Why dropped?

laurynas-biveinis (changes requested):

storage/rocksdb/ha_rocksdb.cc
george-lorch: There is now, and probably more as other things need keys created for pfs instrumentation
https://jira.percona.com/browse/MYR-89

storage/rocksdb/ha_rocksdb.cc
george-lorch: Unsure if one could depend on string to be properly null terminated, so just being overly cautious.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Possibly, I have seen a few other instances of questionable casting used, but trying to keep changes minimal and properly scoped to MYR issues. Maybe we create a MYR issue to review all casting and ensure the minimal needed casts are being used.

storage/rocksdb/ha_rocksdb.cc
george-lorch: lex_str I think can never be NULL as thd->query().str is an instance, not a pointer, but is it possible that the contents of .str can be {NULL, 0}?

storage/rocksdb/ha_rocksdb.cc
george-lorch: Yes, first, I believe it is inappropriate here and second, it is not 'visible' outside of sql/handler.cc as noted in commit message.

storage/rocksdb/rdb_threads.cc
george-lorch: my_thread* interface does not have facility to set thread name and IMHO this is bad form to mix use of pthread primitives with my_thread*. It is of little real use IMHO but I did make suggestion in commit msg that maybe we properly extend my_thread to expose thread naming functionality and then we can use that from here.

george-lorch:
Addressed a few of the real suggestions, left cleanup stuff not really directly related to the commits and task at hand. These should get recommended to upstream. Please take another look and le me know what you think.

storage/rocksdb/rdb_threads.cc
laurynas-biveinis: Please log a task to restore if you haven't done so already - this feature is useful

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: Let's fix things on touched lines at least

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: What do you think?

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: It must be a bug to have it in the 5.6 version, right?

laurynas-biveinis (approved):

storage/rocksdb/ha_rocksdb.cc
george-lorch: Possibly, but actually seems like it should be some server functionality tied to XID like DBUG_ASSERT_XID_IS_VALID or something. Doesn't seem like something that should be done down here.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Don't disagree, but that's not actually a touched line, just diff context.

laurynas-biveinis: Sure, was referring in general

storage/rocksdb/ha_rocksdb.cc
george-lorch: I think it is in the sense that it might be over counting certain index operations, but, I am far from an expert here. The upstream commit that introduces this was https://github.com/facebook/mysql-5.6/commit/aa89833f8fcef6ce7346ae94db9a1463e069b7ec and comes from Sergei P but I think it was wrong originally and is definitely wrong in 5.7 considering that the macros are hidden in handler.cc

storage/rocksdb/rdb_threads.cc
george-lorch: Well, if you want it, I'll just go ahead and add a BP for PS to expose the call, implement it and change this up to use it. I don't find it particularly useful, I've lived without it or years so far and managed to debug things :-\

george-lorch: Logged here https://blueprints.launchpad.net/percona-server/+spec/5.7-myrocks-add-setname and here https://jira.percona.com/browse/MYR-93.

storage/rocksdb/ha_rocksdb.cc
george-lorch: @laurynas-biveinis Looking at xid_t in sql/xa.h, this seems like it would be reasonable to add something like an is_valid() method to the struct that validates something like
`bool is_Valid() const {
  return get_format_id() != -1 &&
            get_gtrid_length() >= 0 && get_gtrid_length() <= MAXGTRIDSIZE &&
           get_bqual_length() >= 0 && get_bqual_length() <= MAXBQUALSIZE;
 `

The problem with this I think is that MAXGTRIDSIZE and MAXBQUALSIZE are not visible within this scope (they are in handler.h). I could see doing the same but maybe passing in values for max_gtrid_length and max_bqual_length to be compared to.

Getting something like that in place then makes it easy to just DBUG_ASSERT(src.is_valid()) from MyRocks.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Actually fixing all of these on my_malloc in a coming PSI branch that instruments memory and some other missing keys.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: I see, I am not sure if sorting out MAX...SIZE visibility wouldn't be too much of a yak to shave, in which case it's fine to leave it alone. Re. passing the max values - would anyone ever pass different values than these constants?

laurynas-biveinis (approved):

PR #1439, opened 2017-02-21
Bug #84415 slave don't report Seconds_Behind_Master when

robgolebiowski:
running slave_parallel_workers > 0

The fix: When a task from GAQ finishes executing, last_master_timestamp
gets assigned timestamp from next task in GAQ. In case GAQ is empty
last_master_timestamp gets reset to 0. Because last_master_timestamp
is only re-assigned when task leaves GAQ it must be also assigned
when first event (from parallel execution) arrives.

### DIFF
--- a/mysql-test/include/wait_for_mts_checkpoint.inc
+++ b/mysql-test/include/wait_for_mts_checkpoint.inc
@@ -1,0 +1,11 @@
+# ==== Purpose ====
+#
+# Waits until mts_checkpoint_routine has been fully
+# executed - from the beginning till the end
+#
+SET @@global.debug= "+d,mts_checkpoint";
+#If end was hit before start - clear it
+SET DEBUG_SYNC="RESET";
+SET DEBUG_SYNC="now WAIT_FOR mts_checkpoint_start";
+SET DEBUG_SYNC="now WAIT_FOR mts_checkpoint_end";
+SET @@global.debug= "-d,mts_checkpoint";
--- a/mysql-test/suite/rpl/t/bug84415.test
+++ b/mysql-test/suite/rpl/t/bug84415.test
@@ -1,0 +1,291 @@
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+--let $rpl_skip_start_slave= 1
+--source include/master-slave.inc
+
+--echo #
+--echo # Bug #84415: slave don't report Seconds_Behind_Master when
+--echo # running slave_parallel_workers > 0
+--echo #
+
+# restart slave in MTS mode
+connection slave;
+SET @@global.debug= "+d,dbug.mts.force_clock_diff_eq_0";
+SET @save.slave_parallel_workers= @@global.slave_parallel_workers;
+SET @save_slave_transaction_retries= @@global.slave_transaction_retries;
+SET @@global.slave_transaction_retries= 0;
+SET @@global.slave_parallel_workers= 4;
+--source include/start_slave.inc
+
+# 1st scenario. slave executes insert to t1 and is blocked due to lock on t1, after 3s another worker
+# on slave starts to execute insert to t2, this worker is also blocked on insert to t2. After 2 more
+# seconds we check Seconds_Behind_Master. Since worker is still  executing insert to t1 this means
+# that slave is at least 5s behind master (might be more due to slow network connection).
+
+connect (slave2,127.0.0.1,root,,test,$SLAVE_MYPORT,);
+connection master;
+
+CREATE TABLE t1 (f1 INT);
+CREATE DATABASE test2;
+USE test2;
+CREATE TABLE t2 (f1 INT);
+--source include/sync_slave_sql_with_master.inc
+
+connection slave1;
+LOCK TABLE test.t1 WRITE;
+
+connection slave2;
+LOCK TABLE test2.t2 WRITE;
+
+connection master;
+
+USE test;
+let $start= `SELECT UNIX_TIMESTAMP()`;
+INSERT INTO t1 VALUES (1);
+
+--real_sleep 3
+
+USE test2;
+let $start2= `SELECT UNIX_TIMESTAMP()`;
+INSERT INTO t2 VALUES (1);
+--real_sleep 2
+
+#
+# all events are in relay-log
+#
+source include/sync_slave_io_with_master.inc;
+
+# Check that two workers are waiting
+let $wait_condition= SELECT count(*) = 2 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Waiting for table metadata lock';
+--source include/wait_condition.inc
+let $wait_condition= SELECT count(*) = 0 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Executing event'
+                                                                             OR State = 'update';
+--source include/wait_condition.inc
+
+# $upper_bound = "timestamp on slave after reading Seconds_Behind_Master" -
+# "timestamp on master before the insert statement".
+
+source include/wait_for_mts_checkpoint.inc;
+
+let $sbm= query_get_value("SHOW SLAVE STATUS", Seconds_Behind_Master, 1);
+let $stop= `SELECT UNIX_TIMESTAMP()`;
+let $upper_bound= `SELECT $stop - $start`;
+let $assert_text= Seconds_Behind_Master must be between 5 and upper_bound;
+let $assert_cond= 5 <= $sbm AND $sbm <= $upper_bound;
+source include/assert.inc;
+
+connection slave1;
+UNLOCK TABLES;
+
+# Now the first worker which was 5s behind the master has completed its task. However the second
+# worker is still running its task. This means that Seconds_Behind_Master should now be 2s
+# or slightly more.
+
+# Check that one worker is still running (waiting)
+let $wait_condition= SELECT count(*) = 1 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Waiting for table metadata lock';
+--source include/wait_condition.inc
+let $wait_condition= SELECT count(*) = 0 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Executing event'
+                                                                             OR State = 'update';
+--source include/wait_condition.inc
+
+source include/wait_for_mts_checkpoint.inc;
+
+let $sbm= query_get_value("SHOW SLAVE STATUS", Seconds_Behind_Master, 1);
+let $stop= `SELECT UNIX_TIMESTAMP()`;
+let $upper_bound= `SELECT $stop - $start2`;
+let $assert_text= Seconds_Behind_Master must be between 2 and upper_bound;
+let $assert_cond= 2 <= $sbm AND $sbm <= $upper_bound;
+source include/assert.inc;
+
+connection slave2;
+UNLOCK TABLES;
+
+--connection master
+--source include/sync_slave_sql_with_master.inc
+
+# Now both workers have completed their tasks and no new tasks have arrived, thus
+# Seconds_Behind_Master should be equal to 0.
+
+let $wait_condition= SELECT count(*) = 0 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Waiting for table metadata lock';
+source include/wait_for_mts_checkpoint.inc;
+let $wait_condition= SELECT count(*) = 0 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Executing event'
+                                                                             OR State = 'update';
+--source include/wait_condition.inc
+
+let $sbm= query_get_value("SHOW SLAVE STATUS", Seconds_Behind_Master, 1);
+let $assert_text= Seconds_Behind_Master must be 0;
+let $assert_cond= $sbm = 0;
+source include/assert.inc;
+
+# 2nd scenario. In this scenario we check if Seconds_Behind_Master is updated correctly when first
+# worker compeletes its tasks first.
+
+connection slave2;
+LOCK TABLE test2.t2 WRITE;
+
+connection master;
+
+USE test;
+let $start= `SELECT UNIX_TIMESTAMP()`;
+INSERT INTO t1 VALUES (1);
+
+--real_sleep 3
+
+USE test2;
+let $start2= `SELECT UNIX_TIMESTAMP()`;
+INSERT INTO t2 VALUES (1);
+--real_sleep 2
+
+#
+# all events are in relay-log
+#
+source include/sync_slave_io_with_master.inc;
+
+# Check that one worker is still running
+let $wait_condition= SELECT count(*) = 1 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Waiting for table metadata lock';
+--source include/wait_condition.inc
+let $wait_condition= SELECT count(*) = 0 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Executing event'
+                                                                             OR State = 'update';
+--source include/wait_condition.inc
+
+# First worker has completed its task, second one is still running.
+# We should be 2 seconds behind the master.
+
+source include/wait_for_mts_checkpoint.inc;
+
+let $sbm= query_get_value("SHOW SLAVE STATUS", Seconds_Behind_Master, 1);
+let $stop= `SELECT UNIX_TIMESTAMP()`;
+let $upper_bound= `SELECT $stop - $start2`;
+let $assert_text= Seconds_Behind_Master must be between 2 and upper_bound;
+let $assert_cond= 2 <= $sbm AND $sbm <= $upper_bound;
+source include/assert.inc;
+
+connection slave2;
+UNLOCK TABLES;
+
+let $wait_condition= SELECT count(*) = 0 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Waiting for table metadata lock';
+--source include/wait_condition.inc
+let $wait_condition= SELECT count(*) = 0 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Executing event'
+                                                                             OR State = 'update';
+--source include/wait_condition.inc
+
+# Now both workers have completed their tasks and no new tasks have arrived, thus
+# Seconds_Behind_Master should be equal to 0.
+
+source include/wait_for_mts_checkpoint.inc;
+
+let $sbm= query_get_value("SHOW SLAVE STATUS", Seconds_Behind_Master, 1);
+let $assert_text= Seconds_Behind_Master must be 0;
+let $assert_cond= $sbm = 0;
+source include/assert.inc;
+
+# 3rd scenario. Three workers are running. In this scenario we check if correct Seconds_Behind_Master
+# is calculated when second worker completes first.
+
+--connection master
+
+CREATE DATABASE test3;
+USE test3;
+CREATE TABLE t3 (f1 INT);
+--source include/sync_slave_sql_with_master.inc
+
+connect (slave3,127.0.0.1,root,,test,$SLAVE_MYPORT,);
+
+connection slave1;
+LOCK TABLE test.t1 WRITE;
+
+connection slave3;
+LOCK TABLE test3.t3 WRITE;
+
+connection master;
+
+let $start= `SELECT UNIX_TIMESTAMP()`;
+
+USE test;
+INSERT INTO t1 VALUES (1);
+--real_sleep 3
+
+use test2;
+INSERT INTO t2 VALUES (1);
+--real_sleep 2
+
+USE test3;
+let $start3= `SELECT UNIX_TIMESTAMP()`;
+INSERT INTO t3 VALUES (1);
+--real_sleep 2
+
+source include/sync_slave_io_with_master.inc;
+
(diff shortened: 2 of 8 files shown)


robgolebiowski:
jenkins run => http://jenkins.percona.com/job/percona-server-5.6-param/1711/

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+
+connection slave;
+--source include/start_slave_sql.inc
+
+# Check that two workers are running
+let $wait_condition= SELECT count(*) = 2 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Waiting for table metadata lock';
+--source include/wait_condition.inc
+
+# $upper_bound = "timestamp on slave after reading Seconds_Behind_Master" -
+# "timestamp on master defore the insert statement" (+1 due to the fact that
```
percona-ysorokin: s/defore/before/

sql/rpl_rli.h
```diff
@@ -650,7 +650,7 @@ class Relay_log_info : public Rpl_info
      Coordinator notifies Workers about this event. Coordinator and Workers
      maintain a bitmap of executed group that is reset with a new checkpoint. 
   */
-  void reset_notified_checkpoint(ulong, time_t, bool);
+  void reset_notified_checkpoint(ulong, time_t*, bool);
```
percona-ysorokin: I would make it `const time_t*` as it is an in-only parameter

sql/rpl_rli.cc
```diff
@@ -222,10 +222,10 @@ void Relay_log_info::reset_notified_relay_log_change()
    @param shift          number of bits to shift by Worker due to the
                          current checkpoint change.
    @param new_ts         new seconds_behind_master timestamp value
-                         unless zero. Zero could be due to FD event.
+                         unless NULL. NULL could be due to FD event.
    @param need_data_lock False if caller has locked @c data_lock
 */
-void Relay_log_info::reset_notified_checkpoint(ulong shift, time_t new_ts,
+void Relay_log_info::reset_notified_checkpoint(ulong shift, time_t *new_ts,
```
percona-ysorokin: I would make it `const time_t *new_ts` as it is an in-only parameter.

sql/rpl_slave.cc
```diff
@@ -4401,11 +4401,18 @@ static int exec_relay_log_event(THD* thd, Relay_log_info* rli)
...
+      In case of parallel execution last_master_timestamp is only updated when
+      an event is taken out of GAQ. Thus when last_master_timestamp is 0 we need
+      to initialize it with a timestamp from the first event to be executed in
+      parallel.
     */
-    if (!(rli->is_parallel_exec() ||
-          ev->is_artificial_event() || ev->is_relay_log_event() ||
-          ev->when.tv_sec == 0 || ev->get_type_code() == FORMAT_DESCRIPTION_EVENT ||
-          ev->server_id == 0))
+    if (((rli->is_parallel_exec() && rli->last_master_timestamp == 0) ||
```
percona-ysorokin: a && b || !a === !a || b
So
```c++
(!rli->is_parallel_exec() || rli->last_master_timestamp == 0) && !(...)
```

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
+-- source include/master-slave.inc
```
percona-ysorokin: General concerns about stability of this test: it relies a lot on delays, which may show unexpected results in highly load environments.

Try running locally
`./mtt --debug-server --parallel=16 --repeat=128 bug84415 bug84415 bug84415 bug84415 bug84415 bug84415 bug84415 bug84415 bug84415 bug84415 bug84415 bug84415 bug84415 bug84415 bug84415 bug84415`

laurynas-biveinis: --source include/have_innodb.inc, above master-slave.inc. Or if the test is SE-agnostic, remove ENGINE=... clauses

laurynas-biveinis: Sleeps in test are always a red flag, but here they are almost fine: it's OK to use sleep to ensure that at least this much time has passed, i.e. to create some slave delay. But it's not OK to 1) have upper bound for time passed, it's arbitrarily large on arbitrarily slow machine, i.e. $upper_bound must go away; 2) use sleep for synchronisation: "--real_sleep 1 # give mts checkpoint chance to update slave status" must be replaced by some condition wait or some other sync

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+
+--echo #
+--echo # Bug #84415: slave don't report Seconds_Behind_Master when
+--echo # running slave_parallel_workers > 0
+--echo #
+
+# restart slave in MTS mode
+connection slave;
+source include/stop_slave.inc;
```
laurynas-biveinis: This can be optimized away by $rpl_skip_start_slave= 1 for master-slave.inc

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+SET @save.slave_parallel_workers= @@global.slave_parallel_workers;
+SET @@global.slave_parallel_workers= 4;
+--source include/start_slave.inc
+
+# 1st scenario. slave executes insert to t1 and is blocked due to lock on t1, after 3s another worker
+# on slave starts to execute insert to t2, this worker is also blocked on insert to t2. After 2 more
+# seconds we check Seconds_Behind_Master. Since worker is still  executing insert to t1 thise means
+# that slave is at least 5s behind master (might be more due to slow network connection).
+
+CREATE USER test;
```
laurynas-biveinis: Why not just use existing root?

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+connection slave1;
+LOCK TABLE test.t1 WRITE;
+
+connection slave2;
+LOCK TABLE test2.t2 WRITE;
+
+connection master;
+
+USE test;
+let $start= `SELECT UNIX_TIMESTAMP()`;
```
laurynas-biveinis, line 43: Remove because $upper_bound needs removing

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+INSERT INTO t1 VALUES (1);
+
+--real_sleep 3
+
+USE test2;
+let $start2= `SELECT UNIX_TIMESTAMP()`;
```
laurynas-biveinis, line 49: Remove because $upper_bound needs removing

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+INSERT INTO t2 VALUES (1);
+--real_sleep 2
+
+#
+# all events are in relay-log
+#
+source include/sync_slave_io_with_master.inc;
+
+connection slave;
+--source include/start_slave_sql.inc
```
laurynas-biveinis: Was it stopped before?

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+
+# Check that two workers are running
+let $wait_condition= SELECT count(*) = 2 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE State = 'Waiting for table metadata lock';
+--source include/wait_condition.inc
+
+# $upper_bound = "timestamp on slave after reading Seconds_Behind_Master" -
```
laurynas-biveinis: Remove, see above

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+# "timestamp on master defore the insert statement" (+1 due to the fact that
+# actual check is done after calculating upper_bound).
+
+--real_sleep 1 # give mts checkpoint chance to update slave status
```
laurynas-biveinis: Replace with sync, see above

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+
+let $sbm= query_get_value("SHOW SLAVE STATUS", Seconds_Behind_Master, 1);
+let $stop= `SELECT UNIX_TIMESTAMP()`;
```
laurynas-biveinis: Remove

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+let $upper_bound= `SELECT $stop - $start + 1`;
```
laurynas-biveinis: Remove

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+let $assert_text= Seconds_Behind_Master must be between 6 and upper_bound;
+let $assert_cond= 6 <= $sbm AND $sbm <= $upper_bound;
```
laurynas-biveinis: Only 6 <= $sbm

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,265 @@
...
+let $assert_cond= 4 <= $sbm AND $sbm <= $upper_bound;
+source include/assert.inc;
+
+connection slave2;
+UNLOCK TABLES;
+
+--connection master
+--source include/sync_slave_sql_with_master.inc
+
+# Now both workers have compeleted their tasks and no new tasks have arrived, thus
```
laurynas-biveinis: "completed" here and elsewhere

sql/log_event.cc
```diff
@@ -6984,10 +6984,12 @@ int Rotate_log_event::do_update_pos(Relay_log_info *rli)
     thd->backup_binlog_lock.release_protection(thd);
 
     if (rli->is_parallel_exec())
+    {
+      time_t ts= when.tv_sec + static_cast<time_t>(exec_time);
```
laurynas-biveinis, line 6988: Do not initialize ts if !server_id?
I am not sure how rli->is_parallel_exec() and !server_id can be true at the same time, but this is outside the scope

sql/rpl_rli_pdb.cc
```diff
@@ -1435,18 +1435,14 @@ ulong circular_buffer_queue::en_queue(void *item)
   return ret;
 }
 
-void* circular_buffer_queue::head_queue()
+uchar* circular_buffer_queue::head_queue()
```
laurynas-biveinis: The change is correct but it's a cleanup instead of a part of a fix - I see that the result value is cast anyway? Or is it required? If it's a cleanup only, please remove, it adds to our maintenance diff

sql/rpl_slave.cc
```diff
@@ -5725,6 +5732,7 @@ bool mts_checkpoint_routine(Relay_log_info *rli, ulonglong period,
   bool error= FALSE;
   struct timespec curr_clock;
   bool binlog_prot_acquired= false;
+  time_t ts=0;
```
laurynas-biveinis, line 5736: Move declaration to the use? can even mark it const there

sql/rpl_slave.cc
```diff
@@ -5861,8 +5869,10 @@ bool mts_checkpoint_routine(Relay_log_info *rli, ulonglong period,
     cnt is zero. This value means that the checkpoint information
     will be completely reset.
   */
-  rli->reset_notified_checkpoint(cnt, rli->gaq->lwm.ts, need_data_lock);
-
+  ts= rli->gaq->empty()
+    ? 0
```
laurynas-biveinis, line 5881: s/0/NULL

laurynas-biveinis (changes requested):
I haven't fully understood the overall design of the fix: at which point now SBM gets reset to zero when a parallel slave fully catches up?

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: @laurynas-biveinis This is how other Seconds_Behind_Master tests are constructed. See rpl_seconds_behind_master.test. Upper bound is not arbitrary high - it is time between just before applying changes on master and the moment we are checking SBM on slave - so it is bounded. As to point 2) real_sleep 1 - is also deterministic. MTS checkpoint which refreshes SBM is to be executed every 300ms. Thus if after 1 sec SBM is not updated it means that somehting is broken and the test should fail.

laurynas-biveinis: The host executing test might take arbitrarily long time to execute any piece of code due to load. There is no guarantee that something that has to executed every 300ms will be executed at least once in one second, it is only known for sure that it will not be executed more often than that, and that's all.

sql/rpl_rli.h
robgolebiowski: I took it one step further and changed it to "const time_t *const new_ts"

sql/rpl_rli.cc
robgolebiowski: I took it one step further and changed it to "const time_t *const new_ts"

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: Removed ENGINE= ... clauses

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: Done.

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: Done, changed to root

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: Done.

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: Removed.

sql/rpl_slave.cc
robgolebiowski: Daaamn...right :)

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: @laurynas-biveinis as discussed over slack : upper_bound stays, but wait for mts checkpoint and + 1 to upper bound go away.

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: Done.

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: upper_bound stays as discussed on slack.

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: upper_bound stays as discussed on slack.

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: upper_bound stays as discussed on slack.

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: Done.

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: upper_bound stays as discussed on slack.

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: upper_bound stays as discussed on slack.

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: upper_bound stays as discussed on slack.

mysql-test/suite/rpl/t/bug84415.test
robgolebiowski: @percona-ysorokin Now, together with the recommendations the test is "rock solid". I have run it in the configuration you wrote couple of times, all seems to work fine. The test itself is based on other SBM tests and they work just fine. Now it is far more secured than the other tests.

sql/rpl_rli_pdb.cc
robgolebiowski: reverted it back to void*

sql/rpl_slave.cc
robgolebiowski: Can't move it to use as there is a goto statement before it.

sql/rpl_slave.cc
robgolebiowski: It should be 0. This is the place when last_master_timestamp is reset to 0 - when all events were proceed and slave fully catches up with master.

robgolebiowski:
Jenkins run => http://jenkins.percona.com/job/percona-server-5.6-param/1766/ (one build is failing, but it is not related to my changes)

mysql-test/include/wait_for_mts_checkpoint.inc
```diff
@@ -0,0 +1,11 @@
+# ==== Purpose ====
+#
+# Waits until mts_checkpoint_routine have been fully
```
laurynas-biveinis: s/have/has

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,289 @@
+--let $rpl_skip_start_slave= 1
+--source include/master-slave.inc
+--source include/have_debug_sync.inc
```
laurynas-biveinis: Add --source include/have_debug.inc

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,289 @@
+--let $rpl_skip_start_slave= 1
+--source include/master-slave.inc
```
laurynas-biveinis, line 4: Move below have_ includes

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,289 @@
...
+# restart slave in MTS mode
+connection slave;
+SET @@global.debug= "+d,dbug.mts.force_clock_diff_eq_0";
+SET @save.slave_parallel_workers= @@global.slave_parallel_workers;
+SET @@global.slave_parallel_workers= 4;
+--source include/start_slave.inc
+
+# 1st scenario. slave executes insert to t1 and is blocked due to lock on t1, after 3s another worker
+# on slave starts to execute insert to t2, this worker is also blocked on insert to t2. After 2 more
+# seconds we check Seconds_Behind_Master. Since worker is still  executing insert to t1 thise means
```
laurynas-biveinis: s/thise/this

mysql-test/suite/rpl/t/bug84415.test
```diff
@@ -0,0 +1,289 @@
...
+let $start2= `SELECT UNIX_TIMESTAMP()`;
+INSERT INTO t2 VALUES (1);
+--real_sleep 2
+
+#
+# all events are in relay-log
+#
+source include/sync_slave_io_with_master.inc;
+
+connection slave;
```
laurynas-biveinis: Redundant due to include/sync above

sql/rpl_rli_pdb.cc
```diff
@@ -1437,16 +1437,12 @@ ulong circular_buffer_queue::en_queue(void *item)
 
 void* circular_buffer_queue::head_queue()
 {
-  uchar *ret= NULL;
   if (entry == size)
   {
     DBUG_ASSERT(len == 0);
+    return NULL;
```
laurynas-biveinis, line 1443: Is the diff for this file still required?

laurynas-biveinis (changes requested):
Minor comments

sql/rpl_slave.cc
laurynas-biveinis: Right, misread the type of ts

sql/rpl_rli_pdb.cc
robgolebiowski: Yes, the implementation of head_queue was not working - nobody has checked that as this function was never used. get_dynamic copies element of an array into argument void *element passed to it. However, it requires that this element has allocated memory. In previous implementation we were always passing NULL to it - thus segfault. We do not need to copy the element though, we only need access to it - so I have changed it to dynamic_array_ptr - which just returns the pointer. If a caller of head_queue needs a copy of this memory he can copy it on his own.

percona-ysorokin (commented):
LGTM provided that comments from Laurynas are addressed.

percona-ysorokin (approved):

robgolebiowski:
Jenkins run => http://jenkins.percona.com/job/percona-server-5.6-param/1772/ (one build is failing, but it is not related to my changes)

laurynas-biveinis (approved):

laurynas-biveinis:
This might have caused https://bugs.launchpad.net/percona-server/+bug/1674299

robgolebiowski:
@laurynas-biveinis I will look into it asap.

PR #1441, opened 2017-02-21
Fix bug 1666213 (Debug build shutdown may crash with void add_global_…

laurynas-biveinis:
…thread(THD*): Assertion `!shutdown_in_progress' failed.)

Since there is no synchronisation between setting shutdown_in_progress
flag and wake up of cached connection handling threads, remove the
assert.

http://jenkins.percona.com/job/percona-server-5.6-param/1710/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1443, opened 2017-02-21
Bug 1663251: ASan errors on main.audit_log_filter_commands

gl-sergei:
`my_hash_search` takes two arguments - key to search for and it's
length. When length is 0, `my_hash_search` considers that key has the
same length as hash length.

Since this behaviour may be relied on somewhere, the fix is to avoid
calling `my_hash_search` for empty strings.

### DIFF
--- a/plugin/audit_log/filter.c
+++ b/plugin/audit_log/filter.c
@@ -306,6 +306,9 @@ my_bool audit_log_check_account_included(const char *user, size_t user_length,
 
   account_init(&acc, user, user_length, host, host_length);
 
+  if (acc.length == 0)
+    return FALSE;
+
   mysql_rwlock_rdlock(&LOCK_account_list);
 
   res= my_hash_search(&include_accounts,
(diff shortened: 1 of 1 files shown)


gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1715/

laurynas-biveinis (approved):

laurynas-biveinis:
This needs one more review from either @robgolebiowski or @percona-ysorokin

laurynas-biveinis:
Is 

p event_general->general_sql_command
(MYSQL_LEX_STRING) $0 = (str = "error", length = 0)

an upstream bug that needs to be reported?

gl-sergei:
This is probably by design.  There is no `sql_command` for "Ping".

```
p thd->lex->sql_command
(enum_sql_command) $16 = SQLCOM_END
```

and in `mysqld.cc`:

```
sql_statement_names[(uint) SQLCOM_END].str= "error";
```

But the length is left 0. I don't know if it is intentional.

robgolebiowski (approved):
LGTM

percona-ysorokin (approved):
LGTM

PR #1445, opened 2017-02-21
Bug 1537554 (SIGTERM kills server instead of shutdown w/ threadpool)

laurynas-biveinis:
5.5 is not affected by the bug, thus only add the testcase. Take the
existing percona_signal_handling one and extend it with the various
ignored signals and SIGTERM, whose handling is buggy in 5.7 with
threadpool. Test all those signals with both threadpool and
one-thread-per-connection handlers.

http://jenkins.percona.com/job/percona-server-5.5-param/1527/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1450, opened 2017-02-22
Bug 1666496: Server crashes on startup if plugin is unable to create …

gl-sergei:
…file pointed by --audit_log_file

In order to expose this bug we need to set both
--audit-log-exclude-commands and --audit-log-file at start up.

Second one has to be incorrect, so that plugin would refuse to start
with error:

    mysqld: File './data-dir/mysql-audit.json' not found (Errcode: 2 - No such file or directory)
    2017-02-21T11:58:42.392280Z 0 [ERROR] Plugin audit_log reported: 'Cannot open file ./data-dir/mysql-audit.json.'
    2017-02-21T11:58:42.392350Z 0 [ERROR] Plugin audit_log reported: 'Error: No such file or directory'
    2017-02-21T11:58:42.392355Z 0 [ERROR] Plugin 'audit_log' init function returned error.
    2017-02-21T11:58:42.392360Z 0 [ERROR] Plugin 'audit_log' registration as a AUDIT failed.

The cause for the crash is common with bug 1641910. MEMALLOC'ed
variables are not properly initialized by server (UPDATE and CHECK
functions aren't invoked).

The fix is to make sure that initialization of MEMALLOC'ed variables is
the very first thing we do.

### DIFF
--- a/mysql-test/r/audit_log_startup.result
+++ b/mysql-test/r/audit_log_startup.result
@@ -8,3 +8,5 @@ SHOW VARIABLES LIKE 'audit_log%';
 Variable_name	Value
 SHOW VARIABLES LIKE 'audit_log%';
 Variable_name	Value
+SHOW VARIABLES LIKE 'audit_log%';
+Variable_name	Value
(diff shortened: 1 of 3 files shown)


gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1719/

laurynas-biveinis (approved):

laurynas-biveinis:
Needs one more review from either @robgolebiowski or @percona-ysorokin

robgolebiowski (approved):
LGTM

percona-ysorokin (approved):
LGTM

PR #1452, opened 2017-02-22
Fix lp:1653104 "Compression dictionaries are not being used in partitioned tables".

percona-ysorokin:
Although, upon creating a partitioned table, references to compression
dictionaries are successfully added to 'SYS_ZIP_DICT_COLS', dictionary data
was never used for compression/decompression.

So, specifying 'WITH COMPRESSION_DICTIONARY ...' in field definitions of
partitioned tables did not have any effect.

Fixed by adding an overload of 'update_field_defs_with_zip_dict_info()' to
'ha_partition' class which simply redirects to
'update_field_defs_with_zip_dict_info()' of the first of its handlers.

Moreover, 'ha_innobase::update_field_defs_with_zip_dict_info()' now also
works properly when is called from 'ha_partition' and table name includes
partitioning part ('#p#').

'xtradb_compressed_columns_ibd_sizes' MTR test case generalized to work with
both partitioned and regular tables.

Added 'xtradb_compressed_columns_ibd_sizes_partitioning' MTR test case which
checks that using compression dictionaries increases data compression ratio
for tables with partitions / subpartitions as well.

Added 'xtradb_compressed_columns_partitioning' MTR test case which includes
'SHOW CREATE TABLE' test for a partitioned table with a reference to a
compression dictionary from the original bug description.


percona-ysorokin:
Jenkins build link
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1720/

laurynas-biveinis (approved):

robgolebiowski (approved):
LGTM

PR #1455, opened 2017-02-28
Fix lp:1653104 "Compression dictionaries are not being used in partitioned tables" (5.7).

percona-ysorokin:


laurynas-biveinis:
Jenkins?

laurynas-biveinis:
5.6 to 5.7 partitioning upgrade could be actually tested, please review innodb.partition-upgrade testcase

sql/partition_info.h
```diff
@@ -557,6 +557,23 @@ bool fill_partition_tablespace_names(
...
+
+  @param[in]  part_info       - Partition info.
+  @param[in]  normalized_path - Normalized path name of table and database
+  @param[out] first_name      - The name of the first partition.
+                                Must be at least FN_REFLEN bytes long.
+
+  @return true - On failure.
+  @return false - On success.
+*/
+bool fill_first_partition_name(
```
laurynas-biveinis: MY_ATTRIBUTE((warn_unused_result)) ?

sql/partition_info.cc
```diff
@@ -3301,6 +3301,68 @@ bool fill_partition_tablespace_names(
 }
 
 /**
+  Fill the Tablespace_hash_set with the tablespace names
```
laurynas-biveinis: "Fill the Tablespace_hash_set"?

sql/partition_info.cc
```diff
@@ -3301,6 +3301,68 @@ bool fill_partition_tablespace_names(
...
+  @param[in]  part_info       - Partition info.
+  @param[in]  normalized_path - Normalized path name of table and database
+  @param[out] first_name      - The name of the first partition.
+                                Must be at least FN_REFLEN bytes long.
+
+  @return true - On failure.
+  @return false - On success.
+*/
+bool fill_first_partition_name(
+       partition_info *part_info,
```
laurynas-biveinis: const? or everything breaks down

sql/partition_info.h
```diff
@@ -557,6 +557,23 @@ bool fill_partition_tablespace_names(
        partition_info *part_info,
        Tablespace_hash_set *tablespace_set);
 
+/**
+  Fill the Tablespace_hash_set with the tablespace names
```
laurynas-biveinis: "Fill the Tablespace_hash_set"?

sql/sql_partition.h
```diff
@@ -122,6 +122,30 @@ bool get_partition_tablespace_names(
...
+  @param[in]  normalized_path    - Normalized path name of table and database
+  @param[in]  partition_info_str - The partition expression.
+  @param[in]  partition_info_len - The partition expression length.
+  @param[out] first_name         - The name of the first partition.
+                                   Must be at least FN_REFLEN bytes long.
+
+  @retval true  - On failure.
+  @retval false - On success.
+*/
+bool get_first_partition_name(
```
laurynas-biveinis: MY_ATTRIBUTE((warn_unused_result)) ?

sql/sql_partition.cc
```diff
@@ -253,20 +253,13 @@ static bool is_name_in_list(const char *name, List<String> list_names)
     FALSE                         Success
 */
 
-bool partition_default_handling(TABLE *table, partition_info *part_info,
+bool partition_default_handling(Partition_handler *part_handler,
```
laurynas-biveinis: Since you are touching the declaration anyway, might as well make it static and MY_ATTRIBUTE((warn_unused_result))

laurynas-biveinis: Possible to mark as const?

sql/sql_partition.cc
```diff
@@ -1674,7 +1667,16 @@ bool fix_partition_func(THD *thd, TABLE *table,
   if (!is_create_table_ind ||
        thd->lex->sql_command != SQLCOM_CREATE_TABLE)
   {
-    if (partition_default_handling(table, part_info,
+    Partition_handler *part_handler= table->file->get_partition_handler();
```
laurynas-biveinis, line 1671: const if possible?

laurynas-biveinis (changes requested):

sql/partition_info.h
percona-ysorokin: fixed

sql/partition_info.h
percona-ysorokin: added

sql/partition_info.cc
percona-ysorokin: fixed

sql/partition_info.cc
percona-ysorokin: added const here, although  had to apply const_cast for list iterations

sql/sql_partition.h
percona-ysorokin: added

sql/sql_partition.cc
percona-ysorokin: added `static` and `MY_ATTRIBUTE((warn_unused_result))`
changing to `const Partition_handler *` is not valid as this function calls a non-const method  `get_num_parts()`

sql/sql_partition.cc
percona-ysorokin: likewise

percona-ysorokin:
Jenkins build link (including 5.6->5.7 compressed column upgrade tests)
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/702/

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade_partitioning.test
```diff
@@ -0,0 +1,93 @@
...
+
+# creating a 5.7 partitioned table with the same structure as in pre-included
+# 'std_data/t5635_par*'
+eval CREATE TABLE $table_name(
+  id INT,
+  a BLOB COLUMN_FORMAT COMPRESSED,
+  b BLOB COLUMN_FORMAT COMPRESSED WITH COMPRESSION_DICTIONARY numbers
+) ENGINE=InnoDB
+PARTITION BY RANGE (id) (
+  PARTITION p0 VALUES LESS THAN (128) ENGINE = InnoDB,
```
laurynas-biveinis: Should subpartitions be tested too?

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade_partitioning.test
```diff
@@ -0,0 +1,93 @@
...
+--remove_file $MYSQLD_DATADIR/test/$table_name.frm
+--remove_file $MYSQLD_DATADIR/test/$table_name#P#p0.ibd
+--remove_file $MYSQLD_DATADIR/test/$table_name#P#p1.ibd
+--copy_file std_data/parts/t5635_par.frm $MYSQLD_DATADIR/test/$table_name.frm
+--copy_file std_data/parts/t5635_par.par $MYSQLD_DATADIR/test/$table_name.par
+--copy_file std_data/parts/t5635_par#P#p0.ibd $MYSQLD_DATADIR/test/$table_name#P#p0.ibd
+--copy_file std_data/parts/t5635_par#P#p1.ibd $MYSQLD_DATADIR/test/$table_name#P#p1.ibd
+
+# showing table structure from the 5.6 data files
+eval SHOW CREATE TABLE $table_name;
```
laurynas-biveinis: Insert a record into a legacy-partitioned table and check?

laurynas-biveinis (commented):
Please see if you agree with the comments

percona-ysorokin:
New Jenkins build link
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/708/
Now includes `xtradb_compressed_columns_upgrade` and `xtradb_compressed_columns_upgrade_partitioning` tests.

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade-master.opt
```diff
@@ -0,0 +1 @@
+--loose-compressed-columns-unknown1
```
laurynas-biveinis: ?

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade.test
```diff
@@ -0,0 +1,125 @@
...
+
+CREATE TABLE dummy_t (id INT) ENGINE=InnoDB;
+CREATE INDEX dummy_idx ON dummy_t(id);
+--let $initial_max_tablespace_id = `SELECT MAX(space) FROM information_schema.innodb_sys_tablespaces`
+--let $initial_max_index_id = `SELECT MAX(index_id) FROM information_schema.innodb_sys_indexes`
+
+if($initial_max_tablespace_id >= $desired_tablespace_id)
+{
+  DROP TABLE dummy_t;
+  --skip Unable to create table with desired tablespace id
```
laurynas-biveinis: Don't skip but fail? So that the test is fixed for any environment?

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade_partitioning-master.opt
```diff
@@ -0,0 +1 @@
+--loose-compressed-columns-unknown2
```
laurynas-biveinis: ?

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade_partitioning.test
```diff
@@ -0,0 +1,252 @@
...
+
+CREATE TABLE dummy_t (id INT) ENGINE=InnoDB;
+CREATE INDEX dummy_idx ON dummy_t(id);
+--let $initial_max_tablespace_id = `SELECT MAX(space) FROM information_schema.innodb_sys_tablespaces`
+--let $initial_max_index_id = `SELECT MAX(index_id) FROM information_schema.innodb_sys_indexes`
+
+if($initial_max_tablespace_id >= $desired_tablespace_id)
+{
+  DROP TABLE dummy_t;
+  --skip Unable to create table with desired tablespace id
```
laurynas-biveinis: Likewise

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade_partitioning.test
```diff
@@ -0,0 +1,252 @@
...
+}
+if(`SELECT $initial_max_index_id + $desired_tablespace_id - $initial_max_tablespace_id - 1 >= $desired_index_id`)
+{
+  DROP TABLE dummy_t;
+  --skip Unable to create table with desired index id
```
laurynas-biveinis: Likewise

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade-master.opt
percona-ysorokin: The idea behind this was to create a unique set of options for this test case (which in turn do nothing) so that it would always start from a fresh copy of system tables. This is needed to make sure that MAX tablespace ID and max index ID are still lower than those in pre-included 5.6 .ibd files (100 and 150 in our case). Without this hack whether this test is successful or not will depend on how many tests MTR has run before executing this one.

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade.test
percona-ysorokin: The problem is that I did not find find a way to make this test successful when running with `--repeat=2`, because MTR does not "refresh" system tables between repetitions and after the first successful execution MAX tablespace id will be  already greater than 100.
So, the workaround is to make only first attempt a real test and skip  all the subsequent ones.

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade-master.opt
laurynas-biveinis: Would --force-restart here achieve the same effect here (and would also help with the next comment)?

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade.test
laurynas-biveinis: I see, but what I'm worried about is that after some perturbation we can arrive in the situation where all the test executions are skipped, and it would be very hard to notice. Maybe --force-restart would help here?

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade-master.opt
percona-ysorokin: It looks like I re-invented the wheel. `--force-restart` does exactly what is needed here. I just didn't know about its existence.

`--loose-compressed-columns-unknown1` and `--loose-compressed-columns-unknown2` in the option files changed to `--force-restart`

mysql-test/suite/innodb/t/xtradb_compressed_columns_upgrade.test
percona-ysorokin: `--force-restart` also fixes the issue with `--repeat=2` as with this option the server is also restarted between the repeated runs. So, no there is no need for `--skip` any more.

All `--skip` commands changed to `--die`

percona-ysorokin:
Jenkins build link (after switching to `--force-restart`)
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/718/
(`./mtr --do-test=xtradb_compressed_columns*`)

laurynas-biveinis (approved):

PR #1458, opened 2017-03-02
Add tools to edit TokuDB directory file.

vlad-lesin:
The new system variables are involved:

===tokudb_dir_cmd===

This variable is used to send commands to edit directory file. The format of
the command line is the following:

command arg1 arg2 .. argn

I.e, if we want to execute some command the following statement can be used:

SET tokudb_dir_cmd = "command arg1 ... argn"

Currently the following commands are availabe:

attach dname iname - attach iname to dname, if dname exists - override
                     the previous value, add new record otherwise
detach dname       - remove record with corresponding dname,
                     the corresponding iname file stays untouched
move old_dname new_dname - rename (only) dname from old_dname to new_dname

If table and database names can be parsed from dname then MDL X lock is
acquired and the table is removed from mysql table dictionary cache. When some
table is removed from TDC it's TABLE_SHARE instance is released. TABLE_SHARE
contains refereces counter, when it equal to zero, the table is closed.
At the other hand tokudb cachetable contains a list of active cachefiles and a
list of stalled cachefiles. Each cachefile contains ft_handle which in turn
has refference counter too. When TABLE_SHARE is closed it decrements
the refference counter of corresponding ft_handles (there are usually several
ft_handles associated with one TABLE_SHARE instance as one table consists of
the several ft-files like main, status and keys). When the refference counter
of some ft_handle becomes zero it's cachefile is moved from active to stalled
list in cachetable. Cachefiles from stalled list are the first candidates for
eviction.

So it's enought just to remove table from TDC and it's cachefiles will be moved
to stalled list sooner or later. Sooner or later because there can be some
other objects which use the table's TABLE_SHARE instance, for example, tokudb
storage engine gathers some statistics in background and it can hold
TABLE_SHARE instances for this purpose.

As it was said when cachefiles are placed into stalled list they will be
evicted sooner or latter unless they are reopened before eviction. When new
ft-file is opened it's file_id is obtained. file_id consists of device id and
inode number. The cachefile is searched in stalled list by this file_id. If
it's not found then new cachefiles is created.

So if some iname was changed/deleted for some dname with the commands described
above the corresponding cachefile will be moved to stalled list and evicted
eventually unless the same iname is not used for opening ft-file with some
other dname before the eviction.

===tokudb_dir_cmd_last_error===

Contains error number of the last executed command

===tokudb_dir_cmd_last_error_string===

Contains error string of the last executed command

Testing: http://jenkins.percona.com/job/percona-server-5.6-param/1730/
tokudb.backup.tokudb_backup_exclude fails because the test uses tokubackup master which has updated api while the PS GCA branch still uses the old API.

See also:
https://github.com/percona/PerconaFT/pull/367
https://github.com/percona/percona-server/pull/1459

### DIFF
--- a/storage/tokudb/tokudb_sysvars.cc
+++ b/storage/tokudb/tokudb_sysvars.cc
@@ -25,6 +25,9 @@ Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved.
 #ident "Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved."
 
 #include "hatoku_hton.h"
+#include "sql_acl.h"
+#include "tokudb_dir_cmd.h"
+#include "sql_parse.h"
 
 namespace tokudb {
 namespace sysvars {
@@ -39,6 +42,8 @@ namespace sysvars {
 #else
 #define TOKUDB_VERSION_STR NULL
 #endif
+
+const size_t error_buffer_max_size = 1024;
 
 ulonglong   cache_size = 0;
 uint        cachetable_pool_threads = 0;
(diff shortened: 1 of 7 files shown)


george-lorch:
I haven't reviewed fully yet, now that the FT 367 has been approved, please ensure that you move the PS submodule commit pointers once 367 is merged.

Since this is a new feature, it will need QA and documentation, so please create proper JIRA issues for RQA and DOC.

storage/tokudb/tokudb_sysvars.cc
```diff
@@ -918,7 +923,91 @@ static MYSQL_THDVAR_BOOL(
...
+    THDVAR(thd, dir_cmd_last_error) = error;
+    char *old_error_string = THDVAR(thd, dir_cmd_last_error_string);
+    if (error_fmt) {
+        va_list varargs;
+        va_start(varargs, error_fmt);
+        char   buff[error_buffer_max_size];
+
+        vsnprintf(buff, sizeof(buff), error_fmt, varargs);
+
+        THDVAR(thd, dir_cmd_last_error_string) =
```
george-lorch: We have had some serious memory issues with plugin setting string variables to use as error indicators. Please test with valgrind on both 5.6 and 5.7 where a regular MySQL cli sets these values, then the plugin sets them, then the client sets them again.

Also try SET STATEMENT on the value, guaranteed to cause issue.

george-lorch (changes requested):

storage/tokudb/tokudb_sysvars.cc
vlad-lesin: Sure, I will recheck this moment. But it looks like the issue is the same as described here https://bugs.launchpad.net/percona-server/+bug/1671082/comments/1 for TokuDB Hotbackup.

george-lorch: Yes, it is the same, and I was warning against introducing another like it. Unfortunately, this use is a very nice way to report session status/error back without having to change SQL layer error/warning messages.

george-lorch (approved):

george-lorch:
Please be sure to create JIRA RQA and DOC issues for this and cross link them to the TokuDB issue when merging.

vlad-lesin:
George, you are right, I have changed the work with session variable to set error string by analogy with https://github.com/percona/percona-server/pull/1531. The new testing looks clean: http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1785/.

vlad-lesin:
https://jira.percona.com/browse/RQA-81,
https://jira.percona.com/browse/DOC-413

laurynas-biveinis:
This likely caused https://bugs.launchpad.net/percona-server/+bug/1675763

PR #1460, opened 2017-03-02
Fix bug 1659746 (Test main.index_merge_innodb is unstable)

laurynas-biveinis:
The previous testcase fix for bug 1625151 missed one more instability
case in the testcase, where
"union(intersect(key1,key2),intersect(key3,key4))" could be
"union(intersect(key1,key2),intersect(key4,key3))".

Previous cases were resolved by restarting the server with a slow
shutdown so that purge completes, but here, since key3 and key4 are
interchangeable in this query plan, replace them both in the output to
something common, avoiding a restart.

http://jenkins.percona.com/job/percona-server-5.6-param/1732/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1464, opened 2017-03-06
Fix bug 1658006 (Test sys_vars.rpl_init_slave_func is unstable)

laurynas-biveinis:
A slave SQL thread sets its Running state to Yes very early in its
initialisation, before the majority of initialisation actions,
including executing the init_slave command, are done. Thus the
testcase has a race condition where the initial replication setup
might finish executing later than the testcase SET GLOBAL init_slave,
making the testcase see its effect where it checks for its absence.

Add synchronisation by running some workload on the master, which is
then synced to slave, ensuring that its initialisation is fully
done. Replace the apparently needless RESET MASTER / RESET SLAVE
etc. with just slave SQL restarts, as the former breaks in the
presence of added master workload.

http://jenkins.percona.com/job/percona-server-5.6-param/1747/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1468, opened 2017-03-06
Ps 5.7 myr 89 90 91 92

george-lorch:


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2921,10 +2852,10 @@ static inline void rocksdb_register_tx(handlerton *const hton, THD *const thd,
                                        Rdb_transaction *const tx) {
   DBUG_ASSERT(tx != nullptr);
 
-  trans_register_ha(thd, FALSE, rocksdb_hton, NULL); // TODO: PFS
+  trans_register_ha(thd, FALSE, rocksdb_hton, nullptr);
```
laurynas-biveinis: s/FALSE/false, optionally?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2921,10 +2852,10 @@ static inline void rocksdb_register_tx(handlerton *const hton, THD *const thd,
...
   if (my_core::thd_test_options(thd, OPTION_NOT_AUTOCOMMIT | OPTION_BEGIN)) {
     tx->start_stmt();
-    trans_register_ha(thd, TRUE, rocksdb_hton, NULL); // TODO: PFS
+    trans_register_ha(thd, TRUE, rocksdb_hton, nullptr);
```

storage/rocksdb/rdb_mutex_wrapper.h
```diff
@@ -75,7 +75,11 @@ class Rdb_cond_var : public rocksdb::TransactionDBCondVar {
   Rdb_cond_var &operator=(const Rdb_cond_var &) = delete;
 
 public:
+#ifdef HAVE_PSI_INTERFACE
+  Rdb_cond_var(PSI_memory_key psi_key);
```
laurynas-biveinis: explicit? (as typecasting PSI_memory_key to Rdb_cond_var should be wrong)

laurynas-biveinis (changes requested):
Minor optional comments, please decide whether to address here

storage/rocksdb/ha_rocksdb.cc
george-lorch: Good call, fixing the NULL is enough to warrant fixing the bool at the same time.

storage/rocksdb/rdb_mutex_wrapper.h
george-lorch: Another nice catch, yes, this should definitely be explicit. I'll make these changes then merge.

PR #1470, opened 2017-03-06
MYR-94 : 5.7 mtr - incorrect thd::query string access causes assertion

george-lorch:
percona-server/include/thr_mutex.h:153: void safe_mutex_assert_not_owner(
  const my_mutex_t*): Assertion `!mp->count || !my_thread_equal(
  my_thread_self(), mp->thread)' failed.
- Fixed issue in Rdb_trx_info_aggregator to use thd_query_safe since this
  will be reading from thd->query across thread boundaries.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -2690,6 +2690,7 @@ public:
         {rocksdb::Transaction::AWAITING_ROLLBACK, "AWAITING_ROLLBACK"},
         {rocksdb::Transaction::ROLLEDBACK, "ROLLEDBACK"},
     };
+    static const size_t trx_query_max_len = 1024; // length stolen from InnoDB
 
     DBUG_ASSERT(tx != nullptr);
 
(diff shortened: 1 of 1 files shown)


george-lorch:
This passes local mtr and clears the issue from tests noted in the MYR issue. Depends on MYR-66 landing in PR 1469 to fully pass tests.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2790,9 +2791,10 @@ class Rdb_trx_info_aggregator : public Rdb_tx_list_walker {
       }
 
       std::string query_str;
-      if (thd->query().str != nullptr) {
-        query_str = std::string(thd->query().str);
-      }
+      query_str.reserve(trx_query_max_len + 1);
+      size_t query_len = thd_query_safe(
+          thd, const_cast<char *>(query_str.c_str()), trx_query_max_len);
```
laurynas-biveinis: const_casting const away and writing to the result of std::string::c_str is undefined behaviour - replace it with &query_str[0]? Shorter, no need for const_cast, no UB

laurynas-biveinis (changes requested):

george-lorch:
Changes made, quite minor, passed local testing with no issue. Also rebased to recent 5.7 changes.

laurynas-biveinis (approved):

PR #1473, opened 2017-03-07
MYR-97 : 5.7 missing DBUG_RETURNs on gap lock error

george-lorch:
- Added missing DBUG_RETURNs to handler::ha_index_first and
  handler::ha_index_last
- Corrected gap_lock_error.result expected warnings revealed by fixing this
  issue.


george-lorch:
Self approving as this is trivial and blocking further trunk testing of MyRocks

PR #1480, opened 2017-03-07
MYR-98 : MYR-7 is incomplete patch, MyRocks fails to allow index only…

george-lorch:
… scans on

  partitioned tables
- Imported full upstream Facebook MySQL patch for MyRocks in sql/table.cc
- This patch is to enable MyRocks to properly use index only scans for
  partitoned tables and extended keys.


george-lorch:
https://bugs.launchpad.net/percona-server/+bug/1670834
https://jira.percona.com/browse/MYR-98
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1755/

I'm not proposing the 5.7 yet as there will probably be some discussion on this.

laurynas-biveinis (approved):

PR #1482, opened 2017-03-08
Fix bug 1671076 (Test main.mysqldump is unstable)

laurynas-biveinis:
The testcase invokes a command, redirects its stderr to stdout, and
expects a certain interleaving between the two. But it is not
deterministic due to stdout buffering. Fix by redirecting the two
streams separately.

http://jenkins.percona.com/job/percona-server-5.6-param/1758/


george-lorch (approved):

percona-ysorokin (approved):
LGTM

PR #1484, opened 2017-03-09
lp1671152 : tokudb does not use index even if cardinality is good

george-lorch:
- ha_tokudb::records_in_range returns the logical row count for each
  individual index when the range is -infinity to + infinity.
- This is problematic as different indices on the same table can report
  wildly different logical row count estimations for many reasons too deep to
  discuss here.
- In comparison, InnoDB returns only a single table row count for all indices.
- The optimizer then uses the index cardinality along with the indexes
  individual row estimate to compute the cost of using that index. When the
  reported row counts are wildly different, the cost calculation is badly and
  inaccurately skewed, making a good index possibly look like a bad choice.
- The fix is to return only the row logical count from the PK regardless of the
  index being counted in ha_tokudb::records_in_range.
- Also touched up ha_tokudb::info call to eliminate redundant call to PerconaFT
  statt::64 on PK dictionary through ha_tokudb::estimate_num_rows.

### DIFF
--- a/storage/tokudb/ha_tokudb.cc
+++ b/storage/tokudb/ha_tokudb.cc
@@ -6121,7 +6121,6 @@ int ha_tokudb::info(uint flag) {
         stats.records = share->row_count() + share->rows_from_locked_table;
         stats.deleted = 0;
         if (!(flag & HA_STATUS_NO_LOCK)) {
-            uint64_t num_rows = 0;
             TOKU_DB_FRAGMENTATION_S frag_info;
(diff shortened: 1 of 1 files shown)


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1761/

laurynas-biveinis (approved):

percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1490, opened 2017-03-09
Fix bug 1671013 (CREATE TEMPORARY TABLE ... SELECT ... gets a DROP bi…

laurynas-biveinis:
…nlogged on disconnect in mixed)

In case of CREATE ... SELECT, a temporary table is created before the
logging format is decided for the current statement, resulting in
premature binlog-on-drop flag value. Fix by setting the flag again,
after the final decide_logging_format call.

Add testcases for this and other cases of temporary table binlogging
on session disconnect.

http://jenkins.percona.com/job/percona-server-5.6-param/1762/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1492, opened 2017-03-13
Improve ACL cache concurrency

efirs:
The ACL cache is fixed size and is protected with exclusive mutex. If the lookup entry is not in the cache or the entry has a wildcard, then MySQL resorts to linear search through all the ACL entries. This change replaces exclusive mutex with RWLock and adds a configuration option to control the ACL cache size:

| **Command-Line Format** | --acl-cache-size=# | |
|---|---|---|
| **System Variable** | **Name** | acl_cache_size |
| | **Variable Scope** | Global |
| | **Dynamic Variable** | Yes |
| **Permitted Values**	| **Type** | integer |
| | **Default** | 256 |
| | **Min Value** | 1 |
| | **Max Value** | 1048576 |

### DIFF
--- a/mysql-test/r/mysqld--help-notwin.result
+++ b/mysql-test/r/mysqld--help-notwin.result
@@ -11,6 +11,7 @@ The following options may be given as the first argument:
  --abort-slave-event-count=# 
  Option used by mysql-test for debugging and testing of
  replication.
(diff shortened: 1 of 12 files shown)


dutow:
Hello!

PR #1495, opened 2017-03-13
MYR-53 : Missing tool sst_dump causes rocksdb.compact_deletes massive…

george-lorch:
… failures

- Initial work on CMakeLists.txt did not correctly build extra tools. Fixed
  to build Myocks/RocksDB tools.
- Added mtr environment variable MYSQL_SST_DUMP so test(s) can access binary.
- Touched compact deletes test to audit and change to have_rocksdb.inc.
- Touched compact_deletes test to add check that MYSQL_SST_DUMP exists prioo
  to executing test.
- Removed obsolete "DROP TABLE IF EXISTS"
- Cherry pick merging commit 28ea2285c59d10cb91894164f605cd8e05585df0 from ps-5.6-MYR-53

### DIFF
--- a/mysql-test/mysql-test-run.pl
+++ b/mysql-test/mysql-test-run.pl
@@ -2778,6 +2778,16 @@ sub environment_setup {
   $ENV{'INNOCHECKSUM'}= native_path($exe_innochecksum);
 
   # ----------------------------------------------------
+  # sst_dump
+  # ----------------------------------------------------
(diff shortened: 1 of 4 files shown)


laurynas-biveinis (changes requested):
Same extra two spaces as in 5.6 (OK to push as is to both branches if fixing/remerging effort is too big)

PR #1497, opened 2017-03-14
Adaptive network buffers

efirs:
# Net_buffer_length status variable

New net_buffer_length status variable shows the size of the buffer of the current connection. If SHOW GLOBAL is used then cumulative size of buffers of all connections is shown
This variable will help to estimate how big the network buffers overhead in every particular case.

# Periodic network buffers shrink

When long lived connections are used, network buffers tend to grow towards max_allowed_packet size and never shrink automatically unless the connection is terminated. For example, occasional big query will eventually increase buffers of all connections(if connections picked at random from the pool), while at any moment of time only one such buffer is sufficient. People usually set max_allowed_packet_size to some relaxed value, like 64-128Mb, which, along with significant number of connections(256-1024), may lead to enormous memory overhead.

This feature allows periodically shrink network buffers according to the workload during the specified interval.
Thread's network buffer size is set to the maximum packet size seen in the last interval.
Thus it allows to "forget" spikes in the packet size which happened before the last interval.

Similar affect can also be achieved by disconnecting and reconnecting TCP connections, but that much heavier process(and it reconnects connections with small buffers as well), while this feature is basically a no-op during the stable workload.

Shrink check timeout can be configured through global `net_buffer_shrink_interval` MySQL variable.

| **Command-Line Format** | --net-buffer-shrink-interval=# | |
|---|---|---|
| **System Variable** | **Name** | net_buffer_shrink_interval |
| | **Variable Scope** | Global |
| | **Dynamic Variable** | Yes |
| **Permitted Values**	| **Type** | integer (seconds) |
| | **Default** | 0 (disabled) |
| | **Min Value** | 0 |
| | **Max Value** | 31536000 |


laurynas-biveinis:
@EvgeniyPatlan @abychko can you please review and comment on the library minor version bump and packaging changes?

mysql-test/suite/sys_vars/t/net_buffer_shrink_interval_basic.test
```diff
@@ -0,0 +1,122 @@
...
+--echo see con2 buffer after grow
+show status like 'net_buffer_length';
+--echo See total usage after buffer has been grown
+show global status like 'net_buffer_length';
+
+--echo Wait for first shrink check interval expire
+SELECT SLEEP(3);
```
laurynas-biveinis: This is going to be unstable and needs to replaced by some sync

mysql-test/suite/sys_vars/t/net_buffer_shrink_interval_basic.test
```diff
@@ -0,0 +1,122 @@
...
+
+--echo Shrink check triggered by queries, so trigger it in both threads
+show status like 'net_buffer_length';
+connection con_int1;
+
+show status like 'net_buffer_length';
+connection con_int2;
+
+--echo Wait for second shrink check interval expire
+SELECT SLEEP(3);
```
laurynas-biveinis: Likewise

mysql-test/suite/sys_vars/t/net_buffer_shrink_interval_basic.test
```diff
@@ -0,0 +1,122 @@
...
+
+connection default;
+
+--echo Check default connections size
+show status like 'net_buffer_length';
+
+disconnect con_int1;
+disconnect con_int2;
+
+--echo "This also serves as a delay to give time for above disconnects to finish"
```
laurynas-biveinis: The disconnect finish can be ensured by include/count_sessions.inc include/wait_until_count_sessions.inc pair

mysql-test/suite/sys_vars/t/net_buffer_shrink_interval_basic.test
```diff
@@ -0,0 +1,122 @@
+--source include/restart_mysqld.inc
```
laurynas-biveinis: I'd split this test in two: in sys_vars I'd leave a skeleton test for variable manipulation, and have functionality test in the main suite.

sql/net_serv.cc
```diff
@@ -848,6 +850,11 @@ static ulong net_read_packet(NET *net, size_t *complen)
   if ((pkt_data_len >= net->max_packet) && net_realloc(net, pkt_data_len))
     goto error;
 
+  pkt_data_len = (pkt_data_len+IO_SIZE-1) & ~(IO_SIZE-1);
```
laurynas-biveinis: MySQL coding style: no space before assignment operator, spaces around + -

sql/net_serv.cc
```diff
@@ -848,6 +850,11 @@ static ulong net_read_packet(NET *net, size_t *complen)
   if ((pkt_data_len >= net->max_packet) && net_realloc(net, pkt_data_len))
     goto error;
 
+  pkt_data_len = (pkt_data_len+IO_SIZE-1) & ~(IO_SIZE-1); 
+
+  if(net->max_interval_packet < pkt_data_len)
```
laurynas-biveinis: Space after if

sql/net_serv.cc
```diff
@@ -861,6 +868,36 @@ static ulong net_read_packet(NET *net, size_t *complen)
   return packet_error;
 }
 
+my_bool my_net_shrink_buffer(NET *net, ulong min_buf_size)
+{
+    /* Buffer is already of smallest possible size */
+    if(net->max_packet <= min_buf_size)
+        return false;
```
laurynas-biveinis: s/false/FALSE, likewise for true, because of my_bool ?

laurynas-biveinis (commented):
Thank you for your contribution. I have some relatively minor comments which could be addressed by either you or ourselves. Note that we'd merge to 5.7 by default.

mysql-test/suite/sys_vars/t/net_buffer_shrink_interval_basic.test
efirs: This works because the only guarantee we need is "not less then 2 sec".

efirs:
Checked in the fixes as a separate commit, will squash later when you are ready to accept.

mysql-test/suite/sys_vars/t/net_buffer_shrink_interval_basic.test
laurynas-biveinis: OK

laurynas-biveinis:
@EvgeniyPatlan @abychko can you please review and comment on the library minor version bump and packaging changes?

EvgeniyPatlan (commented):
looks good to me

build-ps/debian/libperconaserverclient18.2.docs
```diff
@@ -0,0 +1,2 @@
+CMakeLists.txt
```
abychko, line 1: Evgeniy @efirs: do we need to pack CMakeLists.txt? for whom?

cmake/mysql_version.cmake
```diff
@@ -18,7 +18,7 @@
 #
 
 SET(SHARED_LIB_MAJOR_VERSION "18")
-SET(SHARED_LIB_MINOR_VERSION "1")
+SET(SHARED_LIB_MINOR_VERSION "2")
```
abychko, line 21: I'm not sure about 18.2, usually version bump depends on ABI incompatibility. @laurynas-biveinis it's OK if it's compatible with previous version except new option, could you please confirm?

build-ps/debian/libperconaserverclient18.2.docs
EvgeniyPatlan: @abychko as I understand it is just file renaming from libperconaserverclient18.1.docs to libperconaserverclient18.2.docs. In our current file we have CMakeLists.txt included.

abychko:
I'm OK with merge if @laurynas-biveinis confirms the versioning

cmake/mysql_version.cmake
laurynas-biveinis: NET is not a part of public API unless I'm missing something. We might be OK without a version bump, will figure out in due course

EvgeniyPatlan:
Hi @efirs. Thanks for your contribution. Our development team has reworked it and merged it https://github.com/percona/percona-server/pull/4178

PR #1498, opened 2017-03-14
lp1671152 : tokudb does not use index even if cardinality is good

george-lorch:
- Advance PerconaFT git submodule commit pointer to keyrange fix.


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1770/

PR #1501, opened 2017-03-15

george-lorch:
… new

         comment in output
- Fixed missed instance in rocksdb.collation.result masked by other failures.
- Fixed rocksdb.collation.test to properly replace result so absolute path is
  not contained within result file.


george-lorch:
Minor change, passes mtr locally now, self approving.

PR #1502, opened 2017-03-15
MYR-63 : rocksdb.bloomfilter3 fails

george-lorch:
lp1672871 : MyRocks - missing server optimizer patch for ICP
- Percona Server is missing a small optimizer patch that was implemented in
  upstream Facebook MySQL here:
  https://github.com/facebook/mysql-5.6/commit/723bfc97c7f7309c531c410133be3acd073f8129
- This commit takes this small server side patch into Percona Server.
- The change also benefits TokuDB in that TokuDB can now properly terminate on
  out of range condition during descending range scan. Re-recorded test with
  better results, partially removed fix for DB-233, and fixed up test to remove
  conditions for case that we should never get to.


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1771/

laurynas-biveinis (approved):

PR #1505, opened 2017-03-15
Implement index-only query optimisation for prefix indexes.

laurynas-biveinis:
This implements
https://blueprints.launchpad.net/percona-server/+spec/index-only-prefix-index-query
by porting the feature from Facebook patch, revisions
5f9e36706fee7295b3ba736899717cf57bfb9859 and
8f784ea5ed30f796e7f16fb3fbae03e9e5ee1f74.

"Currently InnoDB will always fetch the clustered index (primary key
index) for all prefix columns in an index, even when the value of a
particular record is smaller than the prefix length. This change
optimizes that case to use the record from the secondary index and avoid
the extra lookup.

Also adds two status vars that track how effective this is:

    innodb_secondary_index_triggered_cluster_reads:
	Times secondary index lookup triggered cluster lookup.

    innodb_secondary_index_triggered_cluster_reads_avoided:
	Times prefix optimization avoided triggering cluster lookup."

Changes from Facebook version:

- Fixed multibyte character handling in the prefix index.
- The system variable to control the optimisation has been removed,
  leaving the optimisation turned on permanently.
- ibool has been replaced with bool in the source.
- Extra assert in dict_index_get_nth_col_or_prefix_pos to enforce
  caller's contract prefix_col_pos == NULL if inc_prefix == true.
- Minor cleanup in build_template_field.
- Relaxed assert in row_sel_field_store_in_mysql_format_func to accept
  prefix indexes, and restored the assert in row_sel_store_mysql_func
  with added exception for them.
- The testcase has been rewritten to use an include file and asserts,
  moved to innodb suite.

http://jenkins.percona.com/job/mysql-5.7-param/724/

### DIFF
--- a/storage/innobase/include/row0mysql.h
+++ b/storage/innobase/include/row0mysql.h
@@ -681,6 +681,12 @@ struct mysql_row_templ_t {
 					Innobase record in the current index;
 					not defined if template_type is
 					ROW_MYSQL_WHOLE_ROW */
+	bool	rec_field_is_prefix;	/* is this field in a prefix index? */
+	ulint	rec_prefix_field_no;	/* record field, even if just a
+					prefix; same as rec_field_no when not a
+					prefix, otherwise rec_field_no is
+					ULINT_UNDEFINED but this is the true
+					field number*/
 	ulint	clust_rec_field_no;	/*!< field number of the column in an
 					Innobase record in the clustered index;
 					not defined if template_type is
@@ -784,7 +790,9 @@ struct row_prebuilt_t {
 					columns through a secondary index
 					and at least one column is not in
 					the secondary index, then this is
-					set to TRUE */
+					set to TRUE; note that sometimes this
+					is set but we later optimize out the
+					clustered index lookup */
 	unsigned	templ_contains_blob:1;/*!< TRUE if the template contains
 					a column with DATA_BLOB ==
 					get_innobase_type_from_mysql_type();
--- a/mysql-test/suite/innodb/include/prefix_index_only_query_check.inc
+++ b/mysql-test/suite/innodb/include/prefix_index_only_query_check.inc
@@ -1,0 +1,32 @@
+#
+# A helper include file for prefix index index-only query tests
+#
+# Parameters:
+# $prefix_index_check_title - title of the test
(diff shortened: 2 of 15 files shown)


storage/innobase/include/row0mysql.h
```diff
@@ -681,6 +681,12 @@ struct mysql_row_templ_t {
 					Innobase record in the current index;
 					not defined if template_type is
 					ROW_MYSQL_WHOLE_ROW */
+	bool	rec_field_is_prefix;	/* is this field in a prefix index? */
```
percona-ysorokin, line 684: is `rec_field_is_prefix` member needed at all?
can `(rec_prefix_field_no != ULINT_UNDEFINED)` be used as the same indicator?

laurynas-biveinis: Right now, rec_prefix_field_no == rec_field_no != ULINT_UNDEFINED for full fields as returned by dict_index_get_nth_col_pos. Checking whether it's possible to make it return rec_prefix_field_no == ULINT_UNDEFINED for full fields.

robgolebiowski (approved):
LGTM

storage/innobase/include/row0mysql.h
laurynas-biveinis: OK, having prototyped it I think it's better to keep this flag. Making the change in dict_index_get_nth_col_or_prefix_pos to return prefix_col_pos == ULINT_UNDEFINED whenever pos != ULINT_UNDEFINED makes sure that build_template_field starts out with rec_field_is_prefix == (rec_field_no == ULINT_UNDEFINED && rec_prefix_field_no != ULINT_UNDEFINED) holding. However this stops holding at the end of ha_innobase::build_template where templ->rec_field_no is reset to templ->clust_rec_field_no, optionally, and it complicates the control flow everywhere.

The diff of the prototype is at https://gist.github.com/laurynas-biveinis/d57b618a909001d4eb575c9cea63e20f

percona-ysorokin: OK, does make sense.

percona-ysorokin (approved):
LGTM

PR #1507, opened 2017-03-15
MYR-64 : 5.7 mtr - rocksdb.rocksdb_checksums fails

george-lorch:
MYR-59 : 5.7 mtr - re-record tests for different EXPLAIN format
- Test fails due to slightly different timestamp formatting between 5.6 and 5.7
  and use of grep with explicit pattern.
- Test also needed re-recording of EXPLAIN differences due to being masked by
  failure.
- Test used grep explicitly against mysqld.err file, converted to using
  explicit log for this test and search_pattern.inc
- Removed deprecated idioms.

### DIFF
--- a/mysql-test/suite/rocksdb/t/rocksdb_checksums.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_checksums.test
@@ -5,33 +5,57 @@
 #
 --source include/have_debug.inc
 
+--let SEARCH_FILE=$MYSQLTEST_VARDIR/tmp/rocksdb_checksums.err
+--let $restart_parameters=restart:--log-error=$SEARCH_FILE
+--replace_result $SEARCH_FILE LOG_FILE
+--source include/restart_mysqld.inc
+
+create database rocksdb_checksums_test;
+use rocksdb_checksums_test;
+
 set @save_rocksdb_store_row_debug_checksums=@@global.rocksdb_store_row_debug_checksums;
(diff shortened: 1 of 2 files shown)


mysql-test/suite/rocksdb/t/rocksdb_checksums.test
```diff
@@ -121,8 +153,8 @@ select a from t3 force index(a) where a<4;
 select a from t4 force index(a) where a<1000000;
 set session debug= "-d,myrocks_simulate_bad_key_checksum1";
 
-set @@global.rocksdb_store_row_debug_checksums=@save_rocksdb_store_row_debug_checksums;
-set @@global.rocksdb_verify_row_debug_checksums=@save_rocksdb_verify_row_debug_checksums;
-set @@global.rocksdb_checksums_pct=@save_rocksdb_checksums_pct;
-
+#cleanup
```
laurynas-biveinis: Is $SEARCH_FILE removed at the end?

laurynas-biveinis (approved):
LGTM if testcase cleans up properly, another good candidate for --repeat=3

mysql-test/suite/rocksdb/t/rocksdb_checksums.test
george-lorch: Nope, it is not...good catch...--repeat missed it because the test clears it before each search...still sloppy though and needs cleanup.

PR #1509, opened 2017-03-15
MYR-72 : rocksdb.rocksdb_datadir fails

george-lorch:
- Test fails die to ha_rocksdb.so not being loaded as part of the
  MYSQL_BOOTSTRAP_CMD. Added --plugin-load=ha_rocksdb.so to satisfy and
  modernized the test a little.
- Fixed test to run without rocksdb as default engine.

### DIFF
--- a/mysql-test/suite/rocksdb/t/rocksdb_datadir.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_datadir.test
@@ -1,14 +1,14 @@
---source include/have_rocksdb_as_default.inc
+--source include/have_rocksdb.inc
 
-let $ddir = $MYSQL_TMP_DIR/.rocksdb_datadir.test.install.db;
-let $rdb_ddir = $MYSQL_TMP_DIR/.rocksdb_datadir.test;
-let $sql_file = $MYSQL_TMP_DIR/rocksdb_datadir.sql;
+let $ddir = $MYSQLTEST_VARDIR/.rocksdb_datadir.test.install.db;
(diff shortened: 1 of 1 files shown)


laurynas-biveinis (approved):
LGTM, but also a good candidate for a single local MTR --repeat=3 run

george-lorch:
Yeah, I'm ahead of you there, already done a repeat=4 on it...even then it still seems to me like it would be leaving something behind.

mysql-test/suite/rocksdb/t/rocksdb_datadir.test
```diff
@@ -19,12 +19,14 @@ EOF
...
 
 # Launch mysqld with non-standard rocksdb_datadir
-exec $MYSQLD_BOOTSTRAP_CMD --datadir=$ddir --rocksdb_datadir=$rdb_ddir --default-storage-engine=rocksdb --skip-innodb --default-tmp-storage-engine=MyISAM --rocksdb < $sql_file;
+exec $MYSQLD_BOOTSTRAP_CMD --plugin-load=rocksdb=ha_rocksdb.so --datadir=$ddir --rocksdb_datadir=$rdb_ddir < $sql_file;
 
 --echo Check for the number of MANIFEST files
 exec ls $rdb_ddir/MANIFEST-0000* | wc -l;
 
 # Clean up
+remove_files_wildcard $ddir *;
```
laurynas-biveinis, line 28: Doesn't the next line make it redundant?

PR #1511, opened 2017-03-16
Fixed lp:1631954 "It is possible to create COMPRESSED columns in MyISAM tables via specifying partition engines" and lp:1671492 "Unable to add compression columns to a partitioned table via 'ALTER TABLE'"

percona-ysorokin:
When adding a new compressed column to a partitioned InnoDB table,
'ALTER TABLE' used to report
"Table storage engine '<ENGINE>' does not support the create option 'COMPRESSED COLUMNS'".
This happened because 'check_engine()' function was relying on whether table
handletron object had non-null 'create_zip_dict' and 'drop_zip_dict' members.
And this was not the case for existing partitioned tables for which
associated handletron object was 'partition_hton'.

Fixed by extending 'handlerton' flags with one additional value
'HTON_SUPPORTS_COMPRESSED_COLUMNS' which can be used to determine whether
compressed columns are supported by the engine.

'innobase_hton' and 'partition_hton' extended to report that this flag is
supported. Although whether partition SE supports compressed columns or not
depends on what engine was specified for individual partitions, we make
'partition_hton' report 'HTON_SUPPORTS_COMPRESSED_COLUMNS' meaning that this
engine 'potentially' supports this feature.

'Alter_info' class extended with one additional method
'has_compressed_columns()' which is used to determine whether field
definitions have at least one column with compressed attribute among them.

It was possible to create a 'MyISAM' table with compressed columns via
CREATE TABLE t(...) PARTITION BY RANGE(...)
(
  PARTITION p0 VALUES LESS THAN (...) ENGINE=MyISAM,
  PARTITION p1 VALUES LESS THAN MAXVALUE ENGINE=MyISAM
);
when 'default_storage_engine' was set to 'InnoDB'. This used to happen
because at the stage of 'check_engine()' call, 'partition_info' was not
analyzed yet and storage engine for table itself was deduced from
the 'default_storage_engine' system variable rather than from individual
partitions. Fixed by adding extra check after analyzing partition info.


percona-ysorokin:
Jenkins build link
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1774/

sql/sql_alter.cc
```diff
@@ -84,6 +84,24 @@ bool Alter_info::set_requested_lock(const LEX_STRING *str)
...
+  attribute among field definitions in create_list.
+
+  @retval false there are no compressed columns
+  @retval true there is at least one compressed column
+*/
+bool Alter_info::has_compressed_columns() const
+{
+  List_iterator<Create_field> it(
+    const_cast<List<Create_field>&>(create_list));
+  Create_field* sql_field = it++;
```
laurynas-biveinis: Can the list be empty so that iterator already points to one-past-list-end here, before we post-increment? I.e. why not idiomatic iteration loop?

laurynas-biveinis: const Create_field *sql_field?

laurynas-biveinis: No space before assignment

sql/sql_alter.cc
```diff
@@ -84,6 +84,24 @@ bool Alter_info::set_requested_lock(const LEX_STRING *str)
...
+  while (sql_field != 0 &&
+    sql_field->column_format() != COLUMN_FORMAT_TYPE_COMPRESSED)
+    sql_field = it++;
```
laurynas-biveinis: No space before assignment

laurynas-biveinis (changes requested):

sql/sql_alter.cc
percona-ysorokin: 1. In this particular case the list cannot be empty (empty field list is not allowed).
2.  `it++` for a iterator created from an empty list is a valid construction and will return NULL (base list has a global termination node `list_node end_of_list` for such cases)

robgolebiowski (approved):
LGTM

percona-ysorokin:
Updated Jenkins build link
http://jenkins.percona.com/job/percona-server-5.6-param/1777/

laurynas-biveinis (approved):

PR #1515, opened 2017-03-16
Added missing have_rocksdb.inc

george-lorch:
Upstream commit ID: 813cb8923aae97773803833d89dbc04c9de21669

### DIFF
--- a/mysql-test/suite/rocksdb.sys_vars/t/all_vars.test
+++ b/mysql-test/suite/rocksdb.sys_vars/t/all_vars.test
@@ -1,3 +1,4 @@
+--source include/have_rocksdb_as_default.inc
 --source include/not_embedded.inc
 --source include/not_threadpool.inc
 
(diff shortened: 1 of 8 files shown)


laurynas-biveinis (commented):
Why not convert to plain have_rocksdb.inc at the same time?

mysql-test/suite/rocksdb.sys_vars/t/all_vars.test
```diff
@@ -1,3 +1,4 @@
+--source include/have_rocksdb_as_default.inc
```
laurynas-biveinis, line 1: This testcase only needs MyISAM, not RocksDB :)

george-lorch: I was kind of thinking the same thing, it _does_ need RocksDB for the vars to be present in the first place to 'see' the vars, but no, it does not invoke RocksDB for the purposes of executing anything against it.

george-lorch:
Not converting at the same time just to keep scope of the commit down as some of the tests may actually need fixup and re-recording. Also, this was effectively a merge in of a (proposed) upstream commit. Converting tests will come soon enough, I want to get the ones that are converted either passing or disabled until their features are properly ported/implemented.

PR #1517, opened 2017-03-16

george-lorch:
- Fixed bad include directives.


laurynas-biveinis (changes requested):
Don't the testcases still fail? grep for "rocksdb_rpl/" in mysql-test returns --exec perl suite/rocksdb_rpl/t/rpl_check_for_binlog_info.pl $slave_pid_file $MYSQLTEST_VARDIR/log/mysqld.2.err

in the same testcases. (also the testcases probably need custom error log restart)

george-lorch:
Yeah, missed that because the tests still fail before getting to that point due to the missing FB MySQL rpl system tables. This is one of the tests that needs a lot of further work before it passes and still requires myrocks specific mtr incantation as it has not yet been converted. I probably should just fix this include when I address the other issues with these tests if they are even keepable at all due to the missing sys tables.

PR #1518, opened 2017-03-16

george-lorch:
- Fixed bad include directives.
- Cherry pick merging commit 181b1bfdca2604aa600624fbfc3dd0023cb72074 from ps-5.6/ps-5.6-MYR-102


laurynas-biveinis (changes requested):
Same as for 5.6

PR #1521, opened 2017-03-16
MYR-61 : 5.7 mtr - fix and re-record tests using deprecated i_s.globa…

george-lorch:
…l_status

- Changed out all uses of information_schema.global_status to
  performance_schema.global_status and re-recorded results

### DIFF
--- a/mysql-test/suite/rocksdb.sys_vars/r/all_vars.result
+++ b/mysql-test/suite/rocksdb.sys_vars/r/all_vars.result
@@ -1,15 +1,15 @@
 create table t1 (test_name text) engine=MyISAM;
 create table t2 (variable_name text) engine=MyISAM;
 load data infile "MYSQLTEST_VARDIR/tmp/rocksdb_sys_vars.all_vars.txt" into table t1;
-insert into t2 select variable_name from information_schema.global_variables where variable_name like "rocksdb_%";
-insert into t2 select variable_name from information_schema.session_variables where variable_name like "rocksdb_%";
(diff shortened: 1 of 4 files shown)


mysql-test/suite/rocksdb.sys_vars/r/all_vars.result
```diff
@@ -1,15 +1,15 @@
...
-insert into t2 select variable_name from information_schema.global_variables where variable_name like "rocksdb_%";
-insert into t2 select variable_name from information_schema.session_variables where variable_name like "rocksdb_%";
+insert into t2 select variable_name from performance_schema.global_variables where variable_name like "rocksdb_%";
+insert into t2 select variable_name from performance_schema.session_variables where variable_name like "rocksdb_%";
 select variable_name as `There should be *no* long test name listed below:` from t2
 where length(variable_name) > 50;
 There should be *no* long test name listed below:
 select variable_name as `There should be *no* variables listed below:` from t2
 left join t1 on variable_name=test_name where test_name is null ORDER BY variable_name;
 There should be *no* variables listed below:
```
laurynas-biveinis, line 11: Is there a MYR/upstream bug for "There should be *no* variables listed below"?

laurynas-biveinis (approved):

mysql-test/suite/rocksdb.sys_vars/r/all_vars.result
george-lorch: LOL, I noticed that too and I don't think so...

I just sent one for them https://github.com/facebook/mysql-5.6/issues/575

PR #1523, opened 2017-03-17
Fix bug 1668602 (DROP TEMPORARY TABLE creates a transaction in binary…

laurynas-biveinis:
… log on read only server)

The issue is that DROP TEMPORARY TABLE statements are logged pretty
much unconditionally, regardless of binlog format and actual temporary
table presence. Fix by logging only successful DROP TEMPORARY TABLE
statements whose corresponding CREATE TEMPORARY TABLE statement has
been logged too.

Implement the fix by the following.
- Only count those temporary tables whose DROPs are actually going to
  be binlogged in mysql_rm_table for
  ER_GTID_UNSAFE_BINLOG_SPLITTABLE_STATEMENT_AND_GTID_GROUP
  diagnostic.
- In mysql_rm_table_no_locks:
  - split out the bool flags of trans/non_trans tmp tables deletes to
    flags of tables deleted and whether deletes have to be binloggeed
  - check whether the qualified table name in DROP has to be used in
    all cases instead of only if the current binlog format is row and
    IF EXISTS is given.
  - only build the DROP ... strings for the tables whose DROPs
    actually have to be binlogged because of binlog format at table
    create time.

Update the testsuite for the above. Among other changes,
rpl.rpl_row_drop test case becomes handled by existing
rpl_ddl-including testcases, thus delete it.

(cherry picked from commit 2c35dd86804637ae19d683fada79baa76b1efe81)

http://jenkins.percona.com/job/mysql-5.7-param/728/
+ http://jenkins.percona.com/job/mysql-5.7-param/731/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1531, opened 2017-03-21
Bug fix for lp:1674476.

vlad-lesin:
There is no need in additional memory managment for session variable if
THDVAR_SET() macros is used and the variable flags contain PLUGIN_VAR_MEMALLOC.

See also https://github.com/percona/percona-server/pull/1534 for 5.7.

Testing: http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-valgrind/273/
Look at the end of console output. There is no stack trace in valgrind reports described here: https://bugs.launchpad.net/percona-server/+bug/1674476 .

### DIFF
--- a/plugin/tokudb-backup-plugin/tokudb_backup.cc
+++ b/plugin/tokudb-backup-plugin/tokudb_backup.cc
@@ -42,19 +42,15 @@ static MYSQL_SYSVAR_STR(version, tokudb_backup_version,
     "version of the tokutek backup library",
     NULL, NULL, NULL);
 
-static MYSQL_THDVAR_ULONG(last_error,
+static MYSQL_THDVAR_INT(last_error,
     PLUGIN_VAR_THDLOCAL,
     "error from the last backup. 0 is success",
-    NULL, NULL, 0, 0, ~0ULL, 1);
-
-static void tokudb_backup_update_last_error_str(THD* thd,
-                                                struct st_mysql_sys_var* var,
-                                                void* var_ptr, const void* save);
+    NULL, NULL, 0, 0, 0, 1);
 
 static MYSQL_THDVAR_STR(last_error_string,
-    PLUGIN_VAR_THDLOCAL,
+    PLUGIN_VAR_THDLOCAL + PLUGIN_VAR_MEMALLOC,
     "error string from the last backup",
-    NULL, tokudb_backup_update_last_error_str, NULL);
+    NULL, NULL, NULL);
 
 static MYSQL_THDVAR_STR(exclude,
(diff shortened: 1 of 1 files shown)


george-lorch (approved):

plugin/tokudb-backup-plugin/tokudb_backup.cc
```diff
@@ -42,19 +42,15 @@ static MYSQL_SYSVAR_STR(version, tokudb_backup_version,
...
-static MYSQL_THDVAR_ULONG(last_error,
+static MYSQL_THDVAR_INT(last_error,
     PLUGIN_VAR_THDLOCAL,
     "error from the last backup. 0 is success",
-    NULL, NULL, 0, 0, ~0ULL, 1);
-
-static void tokudb_backup_update_last_error_str(THD* thd,
-                                                struct st_mysql_sys_var* var,
-                                                void* var_ptr, const void* save);
+    NULL, NULL, 0, 0, 0, 1);
```
laurynas-biveinis, line 48: Why is allowed min value == allowed max value == 0 ?

plugin/tokudb-backup-plugin/tokudb_backup.cc
```diff
@@ -42,19 +42,15 @@ static MYSQL_SYSVAR_STR(version, tokudb_backup_version,
...
 
 static MYSQL_THDVAR_STR(last_error_string,
-    PLUGIN_VAR_THDLOCAL,
+    PLUGIN_VAR_THDLOCAL + PLUGIN_VAR_MEMALLOC,
```
laurynas-biveinis, line 51: Shouldn't this use | operator?

plugin/tokudb-backup-plugin/tokudb_backup.cc
george-lorch: This was a Tokutek thing. This 'variable' is never, ever meant to be set by a user, only by the plugin, which can set it to any value regardless of the SYSVAR limits. It has no meaning whatsoever if a user sets the value. These should have been done as session status, not system variables.

plugin/tokudb-backup-plugin/tokudb_backup.cc
vlad-lesin: In the case if the flags are just constants with the only bit set "+" is the same as "|", but the other code in this file contains "+" instead of "|". If the result is the same why not write the code with uniform style?

laurynas-biveinis: I believe it's wrong throughout the file, but not wrong enough to fix it only here, or to fix it globally, thus it's OK

PR #1544, opened 2017-03-22
MYR-65 : rocksdb.type_text_indexes fails

george-lorch:
- Imported Facebook MySQL upstream server patches:

- https://github.com/facebook/mysql-5.6/commit/3fda3eb9bc7f7c5fec6ff2edd4057b7101102d69

  Fix sort order for empty strings in 'text' columns

  Summary:
    This is a fix for an upstream bug: http://bugs.mysql.com/bug.php?id=81810

    The problem is that Field_blob::make_sort_key will memzero the sort key for
    the empty string. However, for most collations, the empty string should sort
    as spaces, not as NUL characters.

    The fix is to memzero only if the collation pad character happens to be 0
    as well.

- https://github.com/facebook/mysql-5.6/commit/fecef54c24b5b64037af1ecf2f8e6ced5a90a3ac

  Fix length in Field_blob::make_sort_key

  Summary:
    For the binary collation in Field_blob::make_sort_key, we write out the
    length of the entire blob, even though the key itself may be shorter. We can
    miss out on duplicate key errors because of this:

  Suppose key length is 1, and we want to encode "ab", "a". Then we get the
  following sort keys:

  'a' 0 2
  'a' 0 1

  We want the two of them to produce the same key instead. To fix, define key
  length to be the min of blob length and requested key length.

- Cherry pick merging commit 0db909d87ca190bded0daa45949122a98b17d0d7 from ps-5.6/ps-5.6-MYR-65

Conflicts:
	strings/ctype-utf8.c

### DIFF
--- a/mysql-test/r/myisam.result
+++ b/mysql-test/r/myisam.result
@@ -1120,8 +1120,8 @@ length(c1)	c1
 0	
 SELECT DISTINCT length(c1), c1 FROM t1 ORDER BY c1;
 length(c1)	c1
-0	
 2		A
+0	
 2	 B
 DROP TABLE t1;
 End of 4.1 tests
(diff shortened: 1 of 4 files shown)


george-lorch:
https://bugs.launchpad.net/mysql-server/+bug/1674867
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/741/

laurynas-biveinis (approved):

PR #1546, opened 2017-03-23
lp:1669005 is fixed as well as some other valgrind issues.

vlad-lesin:
As a result https://tokutek.atlassian.net/browse/BACKUP-135 is fixed too.

innodb_use_native_aio_enabled.test is fixed to remove backup directory
at the end of the test to avoid test failing with --repeat mtr option.

Testing:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1780/
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-valgrind/277/BUILD_TYPE=valgrind,Host=ubuntu-xenial-64bit/consoleText

If you look at the valgrind output you will not see the stacktrace described here: https://bugs.launchpad.net/percona-server/+bug/1669005, as well as other valgrind backtraces which are concerned with uninitialized THD fields for mysql function calls.

This is not GCA branch because the similar changes  must be done for 5.7, but current 5,6 GCA branch does not contain the code to fix. So the corresponding commit was cherry-picked to 5.7. See https://github.com/percona/percona-server/pull/1547 for 5.7 changes.


vlad-lesin:
See https://github.com/percona/percona-server/pull/1608 to continue...

PR #1548, opened 2017-03-23
Fix bug 1662908 (Testcases restart in InnoDB read only mode incorrectly)

laurynas-biveinis:
A restart in InnoDB read-only mode requires a preceding slow
shutdown. innodb.percona_log_archive_ro and
innodb.xtradb_compressed_columns_read_only do not perform it, and the
former fails with error log errors intermittently. Cleanup these
testcases to use include/restart_readonly_mysqld.inc instead. At the
same time, since its option file disappears, merge the
percona_log_archive_ro with the main log archiving testcase
innodb.percona_log_arch_func. At the same time cleanup Percona
testcases that restart the server in InnoDB read-only mode correctly
to use restart_readonly_mysqld.inc too. At the same time add a
--force-restart MTR option file for innodb.percona_bug_1433197 to ensure
deterministic undo segment state, as it became needed due to different
MTR testcase scheduling due to the above changes.

http://jenkins.percona.com/job/percona-server-5.6-param/1787/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1564, opened 2017-03-27
MYR-14 : MYR-19 : MYR-62 : MYR-99 : MYR-102

george-lorch:
- Combined commit to address 5.7 binlog group commit interaction. Intersects
  with tests and removal of Facebook MySQL MyRocks crash safe slave
  implementation.

  - MYR-14 : myrocks stores full binlog and gtid info for crash safe slave
    and
    - Original MYR-14 commit was incomplete and failed to remove rocksdb.rpl
      tests that were explicitly testing crash safe slave with MyRocks as the SE
      for system tables and slave info. This is why MYR-102 was discovered as
      these teste are specifically for MyRocks CSS.
    - Removed tests and utility files that are aimed specifically at testing
      different angles of MySQL CSS.
    - Attempted fixing of multiclient_2pc, renamed mis-named option file,
      re-worked test to modern semantics and reported upstream issue
      https://github.com/facebook/mysql-5.6/issues/581. Based on my testing
      and upstream comments, test is extremely unstable due to dependence on
      a single instance of any test running on a fresh, single instance of a
      mysqld process and dataset. As a result, disabled test until upstream
      provides some fixes. Is more of a stress/random test as functionality is
      already explicitly tested within 2pc_group_commit.test

  - MYR-19 : webscale/facebook binlog group commit interfaces and interacts
             with SE differently
    and
    MYR-62 : rocksdb.2pc_group_commit fails
    - Fixed/removed deprected in 5.7 rocksdb_enable_2pc functionality that is
      similar to deprecated innodb_support_xa and tokudb_support_xa features
      that allow full 2pc/xa support to be optional when it is required for
      binlog group commit that these never be disabled.
    - Removed tests and test sections that deal with this variable and
      re-recorded where necessary.

  - MYR-99 : Convert tests to not require have_rocksdb_as_default
    - Touched tests audited and converted to have_rocksdb


mysql-test/suite/rocksdb.rpl/t/multiclient_2pc.test
```diff
@@ -23,7 +19,7 @@ connect (con2, localhost, root,,);
 # has been recovered and commited
 connection con1;
 --echo 'con1'
---exec echo "restart" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+--exec echo "$restart_parameters" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
 SET SESSION debug="d,crash_commit_after_log";
 SET DEBUG_SYNC='rocksdb.prepared SIGNAL parked WAIT_FOR go';
 --error 0,2013
```
laurynas-biveinis, line 25: Why does this and all the other crash tests accept "0" for a possible error? The crash is in query thread and as such should be 2013 every single time (which should be replaced by an ER_... code)

laurynas-biveinis (commented):
"dependence on
a single instance of any test running on a fresh, single instance of a
mysqld process and dataset." - maybe all you need is --force-restart in the option file?

george-lorch:
Ahh, I was unaware of the --force-restart. Does this also prevent any upcoming/scheduled tests from running against this instance as that will also cause a failure.

mysql-test/suite/rocksdb.rpl/t/multiclient_2pc.test
george-lorch: No idea, original Facebook code, easy enough to change to modern semantics and drop the 0.

laurynas-biveinis:
I don't recall from the top of my head what, if any, effect does --force-restart have at the *end* of the testcase

george-lorch:
End doesn't matter, this test just needs to ensure that the _only_ traffic that hits the server instance is specifically from this test, no others, none before, none during. I believe the way mtr manages server instances this is a guarantee, but I do not know that much about mtr internals.

laurynas-biveinis:
Then you should be good, a server instance is always private to a test

george-lorch:
Cool, that is what I thought, private to a test but can/will be reused by another test.

laurynas-biveinis:
That's why I wrote that I don't know what does --force-restart do at the end of the testcase. Normally the next testcase takes the instance as is

george-lorch:
--force-restart is an mtr input option (like --suite, --parallel, etc...) that forces server restarts between tests oneach runner. It is not something usable within the test option files or anywhere within the test suites that I could find documented. The test can always just start of with a --source include/restart_mysqld.inc but honestly this test I think has more issues than just that. I am trying a few different things to see if I can make it more deterministic across --repeat=3 and similar conditions anyway.

laurynas-biveinis:
No, --force-restart is a special option file option too, which is unfortunately underdocumented yet widely used. While grepping for it I also found include/force_restart.inc, which forces a restart *after* the testcase

george-lorch:
OK, grepping I see it but it is absolutely not documented, I spent a good 20 minutes rooting through and searching the mtr docs on it. This test cares not about what happens after, what it needs is a clean memory/cache/log state at the begin, so a reused server from a prior test causes problems. Force restarting before the test is what is needed. That is only one of the issues this test has, which again, is why upstream skips it in their own environment. I have fixed a few things but there are likely more to show, we shall see.

mysql-test/suite/rocksdb.rpl/t/multiclient_2pc.test
george-lorch: I tried CR_SERVER_LOST and it doesn't work since it is a client error code. It seems 5.7 does not yet support symbolic name matching for client errors, only numerical. According to the docs [here](https://dev.mysql.com/doc/mysqltest/2.0/en/writing-tests-expecting-errors.html), 8.0 will support symbolic names for client error codes. So I'm just removing the 0

george-lorch:
Made some changes to reduce the test size and increase the stability that seem to work, but I get the feeling that something is still not right and will show as an instability, even though testing with --repeat and --parallel locally it passes cleanly now.

mysql-test/suite/rocksdb.rpl/t/multiclient_2pc.test
laurynas-biveinis: Ah, client error, OK then

laurynas-biveinis (approved):

PR #1570, opened 2017-03-28
Fix bug 1676705 (mysql-test/extra/binlog_tests/tmp_table.test does no…

laurynas-biveinis:
…t disconnect temp table connections, intermittently failing following tests)

The testcase opens a connection, creates some temporary tables there,
and does not close it. This causes a disconnect at the end of the
testcase, which is not waited for to complete, which in turn may
caused DROP TEMPORARY TABLE binlogged for the next testcase.

Fix by disconnecting the sessions as soon as they are not needed and
waiting for the disconnects to complete.

http://jenkins.percona.com/job/percona-server-5.6-param/1796/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1574, opened 2017-03-29
lp1676401 : mysqldump using deprecated variable

george-lorch:
INFORMATION_SCHEMA.SESSION_VARIABLES
- Altered two cases in mysqldump to check for and use performance schema
  instead of information schema to check for existence of different
  MyRocks variables.
- In the case where SQL is being injected into the resulting dump in order to
  enable rocksdb_bulk_load, use version specifiers to just completely ignore
  anything older than current 5.7 releases as MyRocks will never show up there
  anyway, and if it somehow does, we don't care as it is not a supported case
  for us.
- Fixed up/re-recorded all tests where these changes may appear.


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/758/

george-lorch:
A few jenkins instances are seeing an issue on one of the modified tests. I can't imagien how these changes would be the cause and can not say if this is a regression or just an unstable test

23:01:16 main.backup_locks_mysqldump              w3 [ fail ]  Found warnings/errors in server log file!
23:01:16         Test ended at 2017-03-29 02:01:16
23:01:16 line
23:01:16 2017-03-28T23:01:15.901261Z 28 [ERROR] Failed to write to mysql.general_log: Deadlock found when trying to get lock; try restarting transaction
23:01:16 ^ Found warnings in /mnt/workspace/mysql-5.7-param/BUILD_TYPE/release/Host/ubuntu-precise-64bit/build/mysql-test/var/3/log/mysqld.1.err
23:01:16 ok
23:01:16 
23:01:16  - the logfile can be found in '/mnt/workspace/mysql-5.7-param/BUILD_TYPE/release/Host/ubuntu-precise-64bit/build/mysql-test/var/log/main.backup_locks_mysqldump/backup_locks_mysqldump.log'
23:01:16 
23:01:16 Retrying test main.backup_locks_mysqldump, attempt(2/3)...

This seems to be occurring on previous builds without these fixes so I believe that it is unrelated unless perhaps it is something introduced with the original changes that introduced this regression in the first place. Ideas?

client/mysqldump.c
```diff
@@ -787,22 +787,34 @@ static void write_header(FILE *sql_file, char *db_name)
...
-            "/*!50112 SET @enable_bulk_load = IF (@is_rocksdb_supported,"
-            " 'SET SESSION rocksdb_bulk_load = 1',"
-            " 'SET @dummy_rocksdb_bulk_load = 0') */;\n"
-            "/*!50112 PREPARE s FROM @enable_bulk_load */;\n"
-            "/*!50112 EXECUTE s */;\n"
-            "/*!50112 DEALLOCATE PREPARE s */;\n");
+            "/*!50717 SET @rocksdb_bulk_load_var_name='rocksdb_bulk_load' */;\n"
+            "/*!50717 SELECT COUNT(*) INTO @rocksdb_has_p_s_session_variables"
+            " FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA LIKE"
+            " 'PERFORMANCE_SCHEMA' AND TABLE_NAME LIKE 'SESSION_VARIABLES'"
```
laurynas-biveinis: Lowercase 'performance_schema', 'session_variables' - it should be more robust to match the schema name case too

laurynas-biveinis (changes requested):
Have you figured out why mysqldump execution failure did not fail the testcases?

client/mysqldump.c
george-lorch: perf schema is lowercase, using uppercase fails to match unless LIKE is used.

george-lorch: LOL, and that is pretty much what you just said...sheesh

george-lorch:
No idea how they passed before unless mtr was in 5.6 compatibility mode, which I don't think is/was the case as even in 5.6 compatibility mode, a warning would be produced, causing a different result. The only thing I can figure is that mysqldump was returning with 0 instead of 1, which the code paths do not align with. The failure as reported by the bug should be returning >0, which should have caused an immediate mtr failure from the --exec call. It makes no sense and I see no smoking gun as to how it passed before, still trying to figure that out.

laurynas-biveinis:
main.backup_locks_mysqldump is a long standing issue

george-lorch:
Pushed changes to correct case on result embedded SQL and re-record results. Tested locally with a new jenkins running here http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/759/ which should be finished in a few hours.

george-lorch:
OK, I think I figured out why nothing caught this.
There are two changes as part of the original patch:

First change:
1 - mysqldump itself sends the SQL "SELECT COUNT(*) FROM INFORMATION_SCHEMA.SESSION_VARIABLES WHERE VARIABLE_NAME LIKE 'rocksdb\_skip\_fill\_cache';"
2 - This is to check if MyRocks is present and supports the skip_fill_cache session variable which tells RocksDB not to populate its internal cache with any data queried on this session. Basically is a way of scan resistence.
3a - With show_compatibility_56 off, this query will return an error, and mysqldump will fail with an error (2).
3b - With show_compatibility_56 on, the server returns a SQL warning, but the query completes correctly. This warning is not seen or detected by mysqldump via mysql_query_with_error_report

Second change:
1 - mysqldump injects SQL into the resulting dump so that when that dump is reloaded in a session, it detects if MyRocks is present and if it supports bulk loading. If so, it enables the MyRocks bulk loader.
2 - No tests actually replay this dump against the server, most simply echo the output of the dump to the mtr result file.
3 - With show_compatibility_56 off, any replay of these dumps would generate an error from the mysql client.
4 - With show_compatibility_56 on, any replay of these dumps would succeed as the SQL warnings returned to the mysql client basically go nowhere, they are not echoed to any output.

So, I have verified that for all of these tests, mtr starts the server with show_compatibility_56 on (very bad idea all across the board) and even created a simple test that executes the 'bad' SQL with both show_compatibility_56 on and off. In the case of off, the test detects the mysqldump failure as it executes the first fix and the test fails. Thus, as a result of the poor choice within mtr to run with show_compatibility_56 on, all tests succeed exactly as expected and do not fail because they are never run against a server with the 'default' settings.

laurynas-biveinis (changes requested):
The patch looks good, but don't we need a most basic mysqldump sanity testcase with show_compatibility_56 = off ?

george-lorch:
Yes, I am working on something to that effect, but in the bigger picture, we must eliminate mtr overriding the default of show_compatibility_56. IMHO wherever this idea came from is extremely bad and needs fixed. mtr tests _must_ run complete and pass with server defaults unless explicitly overridden by .opt and other methods where the test writer understands and expects the intended result. I can't say in public how stupid this idea of changing server defaults 'mid product release stream' and backing off the mtr suite in place of actually rewriting and reevaluating tests is as I might get banned from the MySQL community.

george-lorch:
OK, I push with one new test, no other changes. jenkins here http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/771/

I tried to take the mysqldump.test and make it a .inc, then create a new test that includes it with show_compatibility_56 both on and off. That won't work as apparently show_compatibility_56 does a lot more than just change the I_S to P_S, it seems to change the sql_mode back to 5.6 as well </facepalm>. It was not worth trying to fit that all up, so I created a simple new test, an ensured that it failed against the 'bad' mysqldump and passed against the new 'good' mysqldump.

laurynas-biveinis (changes requested):
Have you "git add"'ed that new test?

george-lorch:
Grrr...I did it again. I keep forgetting git commit --amend --all doesn't pick up new files unless they have been explicitly added first.

george-lorch:
OK, here is the new jenkins with the new test http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/776/

mishomvp:
When can we expect fix?
Thanks ;)

laurynas-biveinis:
@mishomvp , ASAP

laurynas-biveinis (approved):
* For a new test like this, I'd prefer feature, not bug number-based naming, e.g. percona_mysqldump_wo_56_compat

* The test should replay its dump. I know most existing tests don't.

Both comments are for the future though, LGTM

laurynas-biveinis:
Merging myself due to an urgent request for PXC by @kbauskar

PR #1590, opened 2017-03-30
Fix bug 1674286 / 85678 (field-t deletes Fake_TABLE objects through b…

laurynas-biveinis:
…ase TABLE pointer w/o virtual dtor)

Avoid deleting child class objects through a base class pointer, since
the base class does not have a virtual destructor and we do not want
to add one. Instead, typecast the pointers to the child class when
passing to delete.

Tested together with another unit test fix at http://jenkins.percona.com/job/percona-server-5.6-asan-param/149/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

robgolebiowski (commented):
Should not the commit message stand bug 1677130 instead of 1674286 ?

laurynas-biveinis:
Yes, it should, fixing before pushing

PR #1592, opened 2017-03-31
Fix bug 1677943 (Bug 1668602 fix did not re-record rpl_semi_sync test)

laurynas-biveinis:
Flush status after CREATE / DROP TEMPORARY TABLE so that the testcase
has the same count of Rpl_semi_sync_master_yes_tx in all three binary
log formats.

http://jenkins.percona.com/job/mysql-5.7-param/772/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1593, opened 2017-03-31
Fix bug 1674284 (Test funcs_1.processlist_priv_no_prot is unstable)

laurynas-biveinis:
Instead of checking at the start of the test that there is only one
connection to the server wait until the connection count becomes one
instead.

http://jenkins.percona.com/job/percona-server-5.6-param/1821/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1597, opened 2017-03-31
Fix bug 1674281 (MTR testcase timeouts should include server error logs)

laurynas-biveinis:
Add server logs to the testcase output in the case of timeout.

http://jenkins.percona.com/job/percona-server-5.5-param/1538/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1602, opened 2017-04-03
lp1676401 : mysqldump using deprecated variable

EvgeniyPatlan:
INFORMATION_SCHEMA.SESSION_VARIABLES
- Altered two cases in mysqldump to check for and use performance schema
  instead of information schema to check for existence of different
  MyRocks variables.
- In the case where SQL is being injected into the resulting dump in order to
  enable rocksdb_bulk_load, use version specifiers to just completely ignore
  anything older than current 5.7 releases as MyRocks will never show up there
  anyway, and if it somehow does, we don't care as it is not a supported case
  for us.
- Fixed up/re-recorded all tests where these changes may appear.
- Created new test to validate thet mysqldump can run against a server operating
  with show_compatibility_56 both on nd off.

(cherry picked from commit d2d17dd676a2d68cb579b882c15d9ff817e9fccf)


abychko (approved):
looks good to me

PR #1603, opened 2017-04-03
Fix bug 1678792 (mysql-test/include/restart_readonly_mysqld.inc waits…

laurynas-biveinis:
… only for 10 seconds for shutdown)

Make restart_readonly_mysqld.inc use the same timeout as the rest of
MTR include files by making it use restart_mysqld.inc.

http://jenkins.percona.com/job/percona-server-5.6-param/1827/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1605, opened 2017-04-03
Add diagnostics for bug 1673800 (Test main.type_temporal_fractional i…

laurynas-biveinis:
…s unstable)

A possible way for the test to fail is for CURRENT_TIMESTAMP() to
happen to have four trailing zeros during execution, and then
comparing equal to an inserted TIME(2) value. To confirm, replace the
unstable query with an assert, which has a debug statement in case of
failure to dump the table.

http://jenkins.percona.com/job/percona-server-5.6-param/1830/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1610, opened 2017-04-04
Fix bug 1679155 / 81814 (InnoDB adaptive hash index uses a bad partit…

laurynas-biveinis:
…ioning algorithm for the real world)

by applying a patch by Alexey Kopytov at https://github.com/mysql/mysql-server/pull/122:

Bug #81814: InnoDB adaptive hash index uses a bad partitioning algorithm for the real world

This patch replaces hash-based AHI partitioning with index ID based
one.

Index ID based partitioning is better than the current bad hash
function. It is also not any worse than any 'good' hash function for
real world workloads.

Adding space IDs would only make sense if index IDs were not guaranteed
to be unique. They are unique in 5.7 and the current public release of
8.0.

(cherry picked from commit d0fb95d1cbab6670b508145954e57a1fe3383885)

http://jenkins.percona.com/job/mysql-5.7-param/779/

### DIFF
--- a/storage/innobase/btr/btr0sea.cc
+++ b/storage/innobase/btr/btr0sea.cc
@@ -1262,9 +1262,7 @@ retry:
 	const index_id_t	index_id
 		= btr_page_get_index_id(block->frame);
 	const ulint		ahi_slot
-		= ut_fold_ulint_pair(static_cast<ulint>(index_id),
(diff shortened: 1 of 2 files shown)


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1611, opened 2017-04-04
Fix bug 1672389 (Make MTR dump server's error log in case warnings fo…

laurynas-biveinis:
…und)

In case the testcase fails its warning check, include the "comment" in
the output too, which in turn includes servers' error logs.

http://jenkins.percona.com/job/percona-server-5.5-param/1539/


percona-ysorokin (approved):
LGTM

robgolebiowski (commented):
LGTM

PR #1618, opened 2017-04-04
ps-5.6-MYR-74 : rocksdb.rpl_row_not_found fails

george-lorch:
- Facebook MySQL added a new slave_exec_mode called SEMI_STRICT. It is
  introduced here  https://github.com/facebook/mysql-5.6/commit/212e57c96129718b76d0ceb5b883f76ee9063f24
  This has no changes to any storage engines, just server code. This new mode
  makes the slave SQL thread ignore KEY_NOT_FOUND errors instead of stopping.
- A subsequent change to MyRocks was made to return HA_ERR_END_OF_FILE instead
  of HA_ERR_KEY_NOT_FOUND from index operations in order to work with this new
  mode here https://github.com/facebook/mysql-5.6/commit/a69a630c6bc721a0906b9a61b62ff74334517a6d
- These tests fail because this mode does not exist in Percona Server.
- Rewrote test to validate the opposite, that the slave SQL thread stops with
  expected error.
- Also corrected MYR-15 issue with gap lock detector fouling test.


mysql-test/suite/rocksdb/t/rpl_row_not_found.inc
```diff
@@ -1,12 +1,14 @@
+# Facebook MySQL has slave_exec_mode=SEMI_STRICT which allows a slave to
+# 'absorb' or ignore key not found errors. Percona Server does not.
+# This test has been modified to only test that the slave SQL thread stops on
+# error as it should when running with the default slave_exec_mode=STRICT
 --source include/have_rocksdb.inc
 --source include/master-slave.inc
```
laurynas-biveinis: Heavy include files (such as master-slave.inc) should be included last in the header, after the light have_* ones

mysql-test/suite/rocksdb/t/rpl_statement_not_found.test
```diff
@@ -1,2 +1,4 @@
---source rpl_row_not_found.inc
+--source include/have_binlog_format_statement.inc
```
laurynas-biveinis: Slightly misleading testcase name: should be something like rpl_stm_row_not_found.test

laurynas-biveinis (changes requested):

george-lorch:
Made minor changes suggested, tested and repushed. Self approving and merging as changes were minor.

PR #1619, opened 2017-04-04
ps-5.6-MYR-74 : rocksdb.rpl_row_not_found fails

george-lorch:
- Facebook MySQL added a new slave_exec_mode called SEMI_STRICT. It is
  introduced here  https://github.com/facebook/mysql-5.6/commit/212e57c96129718b76d0ceb5b883f76ee9063f24
  This has no changes to any storage engines, just server code. This new mode
  makes the slave SQL thread ignore KEY_NOT_FOUND errors instead of stopping.
- A subsequent change to MyRocks was made to return HA_ERR_END_OF_FILE instead
  of HA_ERR_KEY_NOT_FOUND from index operations in order to work with this new
  mode here https://github.com/facebook/mysql-5.6/commit/a69a630c6bc721a0906b9a61b62ff74334517a6d
- These tests fail because this mode does not exist in Percona Server.
- Rewrote test to validate the opposite, that the slave SQL thread stops with
  expected error.
- Also corrected MYR-15 issue with gap lock detector fouling test.
- Cherry pick merging commit 7958cb83459a24369ef308c4b04a326cd6b5f322 from ps-5.6/ps-5.6-MYR-74
- 5.7 removes the test of rpl_statement_not_found as the error does not show
  in stmt for any engine in this same case and the test is covered elsewhere
  in rpl_statement.test.


mysql-test/suite/rocksdb/t/rpl_row_not_found.inc
```diff
@@ -1,12 +1,14 @@
+# Facebook MySQL has slave_exec_mode=SEMI_STRICT which allows a slave to
+# 'absorb' or ignore key not found errors. Percona Server does not.
+# This test has been modified to only test that the slave SQL thread stops on
+# error as it should when running with the default slave_exec_mode=STRICT
 --source include/have_rocksdb.inc
```
laurynas-biveinis: Add --source include/have_myisam.inc

laurynas-biveinis (changes requested):
5.6 comments + one 5.7 comment

george-lorch:
Same comment as 5.6 Made minor changes suggested, tested and repushed. Self approving and merging as changes were minor. Also added have_myisam

PR #1621, opened 2017-04-05
ps-5.7-MYR-99 : Convert tests to not require have_rocksdb_as_default

george-lorch:
- Audited rocksdb suite tests beginning with 0-a making minor updates and
  re-recording.
- rocksdb.add_index_inplace needed to pick these up once passing from other issues
  - MYR-58 : 5.7 mtr - re-record tests for Warning 1287 - deprecated partition handler
  - MYR-79 : 5.7 mtr - re-record tests that use restart_mysqld.inc due to new comment in output
- rocksdb.add_index_inplace also needed some TLC to make the .cnf a proper .opt
  file. Also had an -error expectation where it looks like it shouldn't have
  been needed and caused a test failure.


george-lorch:
Starting to audit cases and fix to not require MyRocks as default engine. Breaking into small-ish easier to handle chunks rather than one massive commit. This one captures suite.rocksdb tests beginning with '0-a'.

jenkins http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/785/

laurynas-biveinis (approved):

PR #1627, opened 2017-04-05
MYR-78: rocksdb.information_schema fails

gl-sergei:
Fixed upstream issue #589 and re-recorded test case.

Differences with upstream are:

- we don't create rocksdb tables during bootstrap
- we don't provide binlog info via ROCKSDB_GLOBAL_INFO

### DIFF
--- a/mysql-test/suite/rocksdb/r/information_schema.result
+++ b/mysql-test/suite/rocksdb/r/information_schema.result
@@ -1,6 +1,13 @@
-DROP TABLE IF EXISTS t1;
-DROP TABLE IF EXISTS t2;
-DROP TABLE IF EXISTS t3;
+select * from INFORMATION_SCHEMA.ROCKSDB_GLOBAL_INFO;
+TYPE	NAME	VALUE
+CF_FLAGS	0	default [0]
+CF_FLAGS	1	__system__ [0]
+select count(*) from INFORMATION_SCHEMA.ROCKSDB_GLOBAL_INFO;
+count(*)
+2
+select VALUE into @keysIn from INFORMATION_SCHEMA.ROCKSDB_COMPACTION_STATS where CF_NAME = 'default' and LEVEL = 'Sum' and TYPE = 'KeyIn';
+CREATE TABLE t1 (i1 INT, i2 INT, PRIMARY KEY (i1)) ENGINE = ROCKSDB;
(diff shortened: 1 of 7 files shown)


george-lorch:
Sergei, I should have given you more info on some things we are doing, so I will use this review to show what I would like.

First, the commit message should mention the MYR issue IDs and titles of what is being fixed, then a full description of the discovered problem and fix. See https://github.com/percona/percona-server/pull/1621 as an example.

Second, the MyRocks test cases are very inconsistent and use old idioms that are no longer necessary. Some test cases do not check to see if MyRocks is available as a storage engine(source include/have_myrocks.inc) and some assume that MyRocks is the default storage engine. So when we touch a test case for any reason, we fix these issues and note them in the commit message with whatever issues might have been present. Typical MYR issues that get treatment are 58, 59, 61, 99, and some others. I will mark up for review here...

george-lorch:
Ohh, I see you did not touch the actual test here, only a re-record, so no need for any additional changes. The remaining problems with the test will be caught up by MYR-99 once I get to it.

gl-sergei:
PR updated:

   5.6-ps-MYR-78: rocksdb.information_schema fails

   1. Cherry-picked fix for facebook/mysql-5.6#589 at commit
      facebook/mysql-5.6@3fffa58333c8f352a8ed9ef51f550110fdde3388
   2. Added flags initialization code to `rocksdb_init_func` to create
      flags for default CF.
   3. Cleaned-up test case:
      - removed binlog and GTID setup from the test case since we
        removed binlog coordinates from I_S.ROCKSDB_GLOBAL_INFO
      - replaced have_rocksdb_as_default with have_rocksdb
   4. Re-recorded test case

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/827/

See also comments in upstream issue

laurynas-biveinis:
Shouldn't we just merge upstream more often instead of cherry-picking?

george-lorch:
In a perfect world, yes. This world, and this project are far from perfect. In order to achieve any semblance of stability, merging is currently on hold until mtr is running clean and I can easily spot regressions introduced by the merge. BTW, the upstream merge process actually is a form of cherry picking.

The upstream fix for this issue came about from our bug report. They made a slightly different change than we did originally for it, so we opted to cherry pick their fix down rather then keep ours and then have to deal with the potential conflict later on.

george-lorch:
+1 - I'll merge it now for you...

PR #1628, opened 2017-04-05
Bug#21616632 RPL.RPL_BUG33931 HAS OCCASIONAL FAILURES ON PB2

laurynas-biveinis:
Problem and Analysis:

The test case is failing because it forces the two replication threads to
error out once asked to start, but it is only waiting for the SQL thread to
stop reporting the error before trying to apply an "RESET SLAVE" statement.

When a not so fast server didn't stopped the I/O thread before the
"RESET SLAVE" statement be applied, the test case fails as reported.

Fix:

In order to fix the issue, the test case must also wait for the I/O thread to
be stopped before trying to reset the slave.

(cherry picked from commit 581c5a76baed872e1a139aa101847b2d7c407d4f)

http://jenkins.percona.com/job/percona-server-5.5-param/1540/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1640, opened 2017-04-08
MYR-111 : rocksdb mtr suite .cnf and .opt files missing --loose prefi…

george-lorch:
…x for rocksdb_* options

- Initial commit missed a few instances of .cnf splits and .opt files without
  --loose prefixes om rocksdb variables.
- Renamed and split if needed all .cnf files to -master|slave.opt files.
- Removed mysql-test/suite/rocksdb/my.cnf and moved all options into suite.opt
- Changed all instances of rocksdb_* options in .opt files to --loose


george-lorch:
Self approving and merging, changes minimal and similar previous work already approved.

PR #1641, opened 2017-04-08
MYR-110 : mtr suite rocksdb overrides default sql_mode for full suite

george-lorch:
MYR-118 : rocksdb mtr .inc files should be relocated to include dir
MYR-119 : Fix tests disabled by MYR-15
MYR-120 : rocksdb.type_varchar unstable
MYR-59 : 5.7 mtr - fix and re-record tests for deprecated EXPLAIN EXTENDED,
         new EXPLAIN result format, and format=json
- This is a rather large change set primarily addressing MYR-110. Any test
  touched in the process of addressing MYR-110 has all needed and known
  changes and re-records needed to correct all issues listed above. An
  attempt was made to ensure that all changes related to this effort were
  obvious via comments.
- The type_* tests and the col_opt_* tests and their include files are a
  part of a full test 'system' with some amount of dependencies and
  interactions across them.
  - Extra .inc variables are now required when using the type_*.inc files
    to specify if either a ddl or dml error is expected as a result of 5.7
    and/or the new default sql_mode.
- If a test behavior change was due specifically to a change in sql_mode
  defaults, the test behavior is duplicated with the impacted statements
  run once with the default sql_mode and any errors caught and logged,
  then run again with sql_mode similar to 5.6.
- 5.7 introduced the inability to create a PRIMARY KEY with non unique NULL
  columns. This created an issue with one portion of the test series where
  an alternate PRIMARY KEY needed to be introduced to allow the test tables
  to be created.
- All .inc files related to any tests changed were relocated from the ./t
  to the ./include directory and the entire suite searched to ensure that
  other tests referencing the .inc files were updated to the new location.
- Any test that was touched and disabled previously by MYR-15 was corrected
  to prevent the gap lock error detector from waking and generating an
  error on some statements. The fix was generally to temporarily change the
  session ISO level to READ_COMMITTED for the statement and back to
  default.
- Any test that was touched and used have_rocksdb_as_default.inc was
  reviewed and fixed if necessary and changed to use have_rocksdb.inc
- One test, rocksdb.loaddata, required an interesting trick in order to
  work correctly with the new 5.7 sql_mode. It uses LOAD DATA INFILE which
  needed both --replace_result and --error. This can not be done directly
  in mtr as each one is telling mtr what to expect on the next immediate
  command. It was suggested then to use the combination of
  --replace_result, send_eval, --error, then reap in order to mask paths
  for the LOAD DATA INFILE and also expect an error from the loading of the
  file. This worked out very nice.
- A few statements still did not have proper output masking of
  non-deterministic data due to the changes in EXPLAIN results (MYR-59).
  This was mostly within the disabled tests, since they were disabled, the
  incorrect results were never noticed.
- Despite the large number of .test and .inc changes, the results remained
  fairly similar to the originals.


george-lorch:
Here is a jenkins run with these changes included http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/807/

It is a matrix reload due to a few minor changes and is based on the current committed branch.

The two failing tests rocksdb.rocksdb_datadir and rocksdb.rpl_row_not_found are unrelated to this pull request and need fixing on their own (tracked in MYR-72 and MYR-74)

laurynas-biveinis (approved):

PR #1642, opened 2017-04-08
MYR-74 : rocksdb.rpl_row_not_found fails

george-lorch:
- When running with full debug build and --debug-server, extra suppressions are
  needed for log file.

### DIFF
--- a/mysql-test/suite/rocksdb/include/rpl_row_not_found.inc
+++ b/mysql-test/suite/rocksdb/include/rpl_row_not_found.inc
@@ -9,8 +9,8 @@
 --source include/master-slave.inc
 
 connection slave;
-call mtr.add_suppression("Slave SQL: Worker.*failed executing transaction");
-
+call mtr.add_suppression("Slave SQL for channel '': Worker.*failed executing transaction");
+call mtr.add_suppression("Slave SQL for channel '': ... The slave coordinator and worker threads are stopped, possibly leaving data in inconsistent state.");
 connection master;
 
 create table t0 (a int) engine=myisam;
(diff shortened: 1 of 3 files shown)


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/808/ shows the issue fixed

mysql-test/suite/rocksdb/include/rpl_row_not_found.inc
```diff
@@ -10,7 +10,8 @@
 
 connection slave;
 call mtr.add_suppression("Slave SQL: Worker.*failed executing transaction");
```
laurynas-biveinis: Is this one still needed? The one added below looks too similar, is it as 5.6 vs 5.7 difference?

george-lorch: I think that is all it is, a 5.6 to 5.7 diff with 5.7 bringing in the "for channel" but can't say offhand for sure as I have not gone back and looked at the actual 5.6 error being logged.

laurynas-biveinis (changes requested):
Replace the suppression instead of adding a new one

george-lorch:
Yup, will do, it just didn't occur to me that the current was 5.6 specific.

george-lorch:
OK, re-pushed with the deprecated suppression removed and test re-recorded. jenkins here http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/811/ and we can see the test passing.

I'm going to go ahead and merge it at this point, nothing more really to do.

PR #1647, opened 2017-04-11
MYR-72 : rocksdb.rocksdb_datadir fails

george-lorch:
- The path to the installed plugins can not be assumed by a test case to be the
  default plugin dir. When calling the bootstrap with MyRocks, it needs to be
  given the explicit path to the .so.

### DIFF
--- a/mysql-test/suite/rocksdb/t/rocksdb_datadir.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_datadir.test
@@ -1,8 +1,13 @@
 --source include/have_rocksdb.inc
 
+if (!$ROCKSDB_LOAD_PATH) {
+  --skip rocksdb.rocksdb_datadir requires that \$ROCKSDB_PLUGIN_LOAD_PATH is set.
+}
+
 let $ddir = $MYSQLTEST_VARDIR/.rocksdb_datadir.test.install.db;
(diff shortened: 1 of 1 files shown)


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/816/

mysql-test/suite/rocksdb/t/rocksdb_datadir.test
```diff
@@ -1,8 +1,13 @@
 --source include/have_rocksdb.inc
 
+if (!$ROCKSDB_LOAD_PATH) {
```
laurynas-biveinis, line 3: Is it possible have include/have_rocksdb.inc passing and $ROCKSDB_LOAD_PATH not set?

george-lorch: That I am not sure of which is why I added the extra checking. My initial assumption was that no, if we pass have_rocksdb, then all of these plugin.defs should be defined, except that if the plugin.def for rocksdb were to change in a way that allowed some of the env vars to be defined but not all. I can't imagine such a scenario off the top of my head, but it may be possible.

laurynas-biveinis: OK, I'm fine either way

laurynas-biveinis (approved):

PR #1648, opened 2017-04-11
MYR-109: Fail to compile on jenkins yakkety

gl-sergei:
Removed redundant assert:

    DBUG_ASSERT_IMP(m_maybe_unpack_info, pk_unpack_info)

because `pk_unpack_info` cannot be NULL according to the function
declaration.

This commit addresses issue facebook/mysql-5.6#605

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -3664,7 +3664,6 @@ bool ha_rocksdb::init_with_fields() {
 void ha_rocksdb::convert_record_to_storage_format(
     const rocksdb::Slice &pk_packed_slice,
     Rdb_string_writer *const pk_unpack_info, rocksdb::Slice *const packed_rec) {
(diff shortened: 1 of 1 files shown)


gl-sergei:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/817/

gl-sergei:
Actually may be fixed as side effect of https://github.com/facebook/mysql-5.6/commit/5107b14404f76617756cd4c0f04cb95e41f82e2c

george-lorch:
Yeah, so then this looks good. Let's take you fix for now to get building, then later on when the upstream patch for TTL is merged we'll be in sync.

laurynas-biveinis (changes requested):
I believe it is the assert which should be kept and attribute removed. It is useless and dangerous as it does not protect against NULL passed but it will optimize away any NULL checks in the implementation.

PR #1649, opened 2017-04-11
MYR-109: Fail to compile on jenkins yakkety

gl-sergei:
Removed redundant assert:

    DBUG_ASSERT_IMP(m_maybe_unpack_info, pk_unpack_info)

because `pk_unpack_info` cannot be NULL according to the function
declaration.

This commit addresses issue facebook/mysql-5.6#605

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -3748,7 +3748,6 @@ bool ha_rocksdb::init_with_fields() {
 void ha_rocksdb::convert_record_to_storage_format(
     const rocksdb::Slice &pk_packed_slice,
     Rdb_string_writer *const pk_unpack_info, rocksdb::Slice *const packed_rec) {
-  DBUG_ASSERT_IMP(m_maybe_unpack_info, pk_unpack_info);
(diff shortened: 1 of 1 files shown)


laurynas-biveinis (changes requested):
I believe it is the assert which should be kept and attribute removed. It is useless and dangerous as it does not protect against NULL passed but it will optimize away any NULL checks in the implementation.

george-lorch:
Gotta love useless syntactical sugar...either way, a coming merge from upstream changes the assertion so that it does not involve the passed argument. The current, and coming change is not an ASSERT != NULL, it is an assert that  `if (a==true) then b must also be true, if (!a) then b doesn't matter` where a is a class member and b happens to be the pointer argument.

The fact that it was operating on pointers is what invoked the NULL check error. In the end, both the attribute and the assertion are effectively useless, so being pedantic about which to remove is sort of a time sink.

laurynas-biveinis:
OK, if it's temporary, it does not matter.

george-lorch:
+1 - I'll merge now, upstream merge to replace/overwrite should be coming in a short time.

PR #1650, opened 2017-04-11
MYR-120 : rocksdb.type_varchar unstable

george-lorch:
- Fixed more non-deterministic result masks
- Added explicit ORDER BY to force results appear in deterministic order as it
  may change depending on the state of the SST and WAL files on test start.
- Added explicit ANALYZE TABLE to force cardinality updates otherwise they are
  cause query plan to float and breaks EXPLAIN SELECT.

### DIFF
--- a/mysql-test/suite/rocksdb/include/type_varchar_endspace.inc
+++ b/mysql-test/suite/rocksdb/include/type_varchar_endspace.inc
@@ -42,18 +42,19 @@ insert into t1 values (0, 'ab', 'a-b');
 insert into t1 values (1, 'a ', 'a-space');
 insert into t1 values (2, 'a',  'a');
 insert into t1 values (3, 'a  \t', 'a-tab');
+analyze table t1;
 
 --echo # Must show 'using index' for latin1_bin and utf8_bin:
 --replace_column 10 # 11 #
 explain
-select col1, hex(col1) from t1;
-select col1, hex(col1) from t1;
+select col1, hex(col1) from t1 order by col1;
+select col1, hex(col1) from t1 order by col1;
 
 --echo # Must show 'using index' for latin1_bin and utf8_bin:
 --replace_column 5 # 7 # 8 # 10 # 11 #
 explain
-select col1, hex(col1) from t1 where col1 < 'b';
-select col1, hex(col1) from t1 where col1 < 'b';
+select col1, hex(col1) from t1 where col1 < 'b' order by col1;
+select col1, hex(col1) from t1 where col1 < 'b' order by col1;
 
 delete from t1;
 insert into t1 values(10, '', 'empty');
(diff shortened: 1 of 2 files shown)


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/821/

mysql-test/suite/rocksdb/include/type_varchar_endspace.inc
```diff
@@ -42,18 +42,24 @@ insert into t1 values (0, 'ab', 'a-b');
 insert into t1 values (1, 'a ', 'a-space');
 insert into t1 values (2, 'a',  'a');
 insert into t1 values (3, 'a  \t', 'a-tab');
+# force cardinality updates otherwise explains below are non-deterministic
```
laurynas-biveinis: No need for such comment (when a testcase does CREATE TABLE ... INSERT ... ANALYZE ... [EXPLAIN ]SELECT,  the role of ANALYZE is always clear).

If this were the only review comment then it could have stayed

mysql-test/suite/rocksdb/include/type_varchar_endspace.inc
```diff
@@ -42,18 +42,24 @@ insert into t1 values (0, 'ab', 'a-b');
...
 
 --echo # Must show 'using index' for latin1_bin and utf8_bin:
 --replace_column 5 # 7 # 8 # 10 # 11 #
 explain
-select col1, hex(col1) from t1 where col1 < 'b';
-select col1, hex(col1) from t1 where col1 < 'b';
+select col1, hex(col1) from t1 where col1 < 'b' order by col1;
+#set optimizer_trace="enabled=on";
+select col1, hex(col1) from t1 where col1 < 'b' order by col1;
+#select * from information_schema.optimizer_trace;
```
laurynas-biveinis: Remove

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb/include/type_varchar_endspace.inc
george-lorch: Ohh crap, that was some of my debug code trying to figure out where the instability was coming from. Should have never made it into the push.

mysql-test/suite/rocksdb/include/type_varchar_endspace.inc
george-lorch: The comment was really a reminder of where our test is differing from upstream. Seems 5.6 is not as sensitive to the differences of a 4 row table as 5.7 is which causes the instability, but I do get the point, the resulting test is fairly obvious.

george-lorch:
Done, and a new jenkins http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/828/

laurynas-biveinis (approved):

PR #1655, opened 2017-04-17

george-lorch:
- Reviewed, fixed, and re-recorded rocksdb.sys_vars where needed.


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/839/

rocksdb.sys_vars.rocksdb_max_open_files_basic fails and is tracked here
https://jira.percona.com/browse/MYR-117
and
https://jira.percona.com/browse/MYR-123

I actually believe it is an upstream MySQL 5.7 regression where signed INT sysvar -1 is being displayed as unsigned 4294967295

laurynas-biveinis (approved):

PR #1656, opened 2017-04-19

george-lorch:
MYR-58 : 5.7 mtr - re-record tests for Warning 1287 - deprecated partition
         handler
- Audited remainder of rocksdb suite tests making minor updates and
  re-recording.
- Cleaned a few naming case conventions where easy enough.
- Corrected a few places to make use of restart_mysqld.inc
- Added occasional SHOW CREATE TABLE to ensure table was created as the test
  intended it to be.
- Corrected a few --error to use literal ER_* name instead of number id.
- Added some count_sessions/wait_until_count_sessions
- rocksdb.read_only_tx has different output on binlog pos due to different
  sized binlog event in 5.7, captured via re-record.
- rocksdb.rocksdb_icp and rocksdb.rocksdb_icp_rev have slightly different
  results due to different engine default state.
- rocksdb.show_table_status was using SHOW GLOBAL VARIABLES LIKE "log_bin" to
  determine if binlog was enabled rather than
- rocksdb.read_only_tx re-enabled as it was missed when fixing MYR-16
- rocksdb_deadlock_stress.py moved from rocksdb/t to rocksdb/include and
  fixed up tests.
- Renamed test index_file_map to correct i_s_index_file_map
- Renamed test perf_context to correct i_s_perf_context


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/842/

Failures are known as:
https://jira.percona.com/browse/MYR-117 : rocksdb.sys_vars.rocksdb_max_open_files_basic fails
https://jira.percona.com/browse/MYR-123 : PS 5.7 rocksdb_max_open_files differs from default
and
https://jira.percona.com/browse/MYR-105 : rocksdb.parts unstable

mysql-test/suite/rocksdb/t/drop_index_inplace.test
```diff
@@ -18,7 +14,7 @@ ALTER TABLE t1 DROP INDEX ka, ALGORITHM=INPLACE;
 SHOW CREATE TABLE t1;
 
 # Key ka does not exist in table t1
---error 1176
+--error ER_KEY_DOES_NOT_EXITS
```
laurynas-biveinis, line 17: Fantastic, this typo is now set in stone

mysql-test/suite/rocksdb/t/update_with_keys.test
```diff
@@ -27,7 +23,7 @@ DROP TABLE t1;
 --echo # RocksDB: skip the test for secondary UNIQUE keys.
 --disable_parsing
 --error ER_GET_ERRMSG
-CREATE TABLE t1 (a INT, b CHAR(8), pk INT AUTO_INCREMENT PRIMARY KEY, UNIQUE INDEX(a)) ENGINE=innodb;
+CREATE TABLE t1 (a INT, b CHAR(8), pk INT AUTO_INCREMENT PRIMARY KEY, UNIQUE INDEX(a)) ENGINE=INNODB;
```
laurynas-biveinis: The testcase strongly suggests that InnoDB is not the intended engine here. OK to fix in a separate commit

laurynas-biveinis (approved):
One InnoDB misuse can be fixed here or as a follow-up, in both cases there is no need for more reviews

mysql-test/suite/rocksdb/t/drop_index_inplace.test
george-lorch: LOL, I found it quite funny myself :)

mysql-test/suite/rocksdb/t/update_with_keys.test
george-lorch: How in the world did I miss that one...There are places MyISAM is explicitly used, and places where InnoDB was explicitly used where it should have been MyRocks that I fixed. This now has me questioning some of the work. I need to re-review this all again myself now that I have 'looked away' for a bit.

george-lorch:
Fixed InnoDB issue, re-reviewed and a new jenkins at http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/852/

Merging \o/

PR #1657, opened 2017-04-19
Fix bug 1683993 (Test main.rpl_slow_query_log is unstable)

laurynas-biveinis:
Remove checks that fast queries are fast enough not to be logged to
the slow query log, as they still might be arbitrarily slow on loaded
machines.

http://jenkins.percona.com/job/percona-server-5.5-param/1547/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1667, opened 2017-04-19
Fix bug 1673800 (Test main.type_temporal_fractional is unstable)

laurynas-biveinis:
The previously added asserts confirmed that the failure happens when
CURRENT_TIMESTAMP() at INSERT time has four trailing zeroes in
microseconds, comparing equal to TIME(2). Fix by checking if that's
the case and adding one microsecond.

(cherry picked from commit 116b01a347d68724c4e0545cc37304c60edf1261)

http://jenkins.percona.com/job/mysql-5.7-param/845/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1672, opened 2017-04-20
Fix bug 1684264 (Test innodb.monitor is unstable)

laurynas-biveinis:
Replace regexs to accept any positive integer for
buffer_page_written_index_leaf value after some workload. The exact
value is not recorded as purge from a previous testcase might be
running in parallel. Remove regexs where the expected counter value is
zero.

http://jenkins.percona.com/job/percona-server-5.6-param/1848/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1674, opened 2017-04-20
Fix bug 1684601 / 86016 (Make MTR show core dump stacktraces from uni…

laurynas-biveinis:
…t tests too)

Implement MTR showing the stacktraces of crashed unit tests by the
following:
- at the end of test run, if --unit-tests was set, search for core
  dumps in the build directory subtree and invoke My::CoreDump->show
  on them;
- removing _verify_binpath call in My::CoreDump::_gdb as it does not
  seem to be required on our supported platforms and if present, it
  assigns all the unit test core dumps to mysqltest executable.
- stop expected-crash tests from producing core dumps by introducing
  MY_EXPECT_DEATH and MY_EXPECT_DEATH_IF_SUPPORTED macros, which
  disable core dumping around the expected-crash calls. The macros and
  support functions are piggybacked on existing thread_utils.h and
  thread_utils.cpp, which is not a fully correct place for them, but
  IMHO acceptable enough.

http://jenkins.percona.com/job/percona-server-5.6-param/1849/


unittest/gunit/thread_utils.cc
```diff
@@ -200,5 +203,33 @@ void Notification::notify()
   ASSERT_FALSE(failed);
 }
 
-
 }  // namespace thread
+
+static struct rlimit initial_core_rlimit;
+static struct rlimit disabled_core_rlimit;
+
+MY_ATTRIBUTE((constructor))
```
percona-ysorokin: Why compiler specific attribute, instead of `static struct rlimit initial_core_rlimit = init_initial_core_limit()`?

unittest/gunit/thread_utils.cc
```diff
@@ -200,5 +203,33 @@ void Notification::notify()
...
+  assert(ret == 0);
+
+  disabled_core_rlimit.rlim_cur= 0;
+  disabled_core_rlimit.rlim_max= initial_core_rlimit.rlim_max;
+}
+
+void disable_core_dumps(void)
+{
+  int ret= setrlimit(RLIMIT_CORE, &disabled_core_rlimit);
+  ASSERT_EQ(ret, 0);
```
percona-ysorokin: Similarly to 'restore_core_dump_rlimit()'
```c++
if (ret != 0)
{
  perror("setrlimit(RLIMIT_CORE) failed (disable)");
  ASSERT_EQ(ret, 0);
}
```

unittest/gunit/thread_utils.cc
```diff
@@ -200,5 +203,33 @@ void Notification::notify()
...
+}
+
+void restore_core_dump_rlimit(void)
+{
+  int ret= setrlimit(RLIMIT_CORE, &initial_core_rlimit);
+  if (ret == -1)
```
percona-ysorokin: `if (ret != 0)` is more reliable

unittest/gunit/thread_utils.cc
```diff
@@ -200,5 +203,33 @@ void Notification::notify()
   ASSERT_FALSE(failed);
 }
 
-
 }  // namespace thread
+
+static struct rlimit initial_core_rlimit;
+static struct rlimit disabled_core_rlimit;
```
percona-ysorokin: do we need this 'disabled_core_rlimit' at all? Looks like pre-caching a struct of 2 integers.

unittest/gunit/thread_utils.h
```diff
@@ -115,4 +115,22 @@ class Notification
 
 }  // namespace thread
 
+extern void disable_core_dumps(void);
+extern void restore_core_dump_rlimit(void);
+
+
+#define MY_EXPECT_DEATH_IF_SUPPORTED(statement, regex)  \
+  do {                                                  \
+    disable_core_dumps();                               \
```
percona-ysorokin: May be a RAII capsule which would store old and set new value in the constructor and restore in the destructor?
Something like the following with proper error checking. No static initialization at all here.
```c++
class disable_core_dumps_capsule
{
  public:
    disable_core_dumps_capsule(): captured_core_rlimit_()
    {
      ... =getrlimit(RLIMIT_CORE, &captured_core_rlimit_);
      struct rlimit disabled_core_rlimit= captured_core_rlimit_;
      disabled_core_rlimit.rlim_cur= 0;
      ...= setrlimit(RLIMIT_CORE, &disabled_core_rlimit);
    }
    ~disable_core_dumps_capsule()
    {
      ...= setrlimit(RLIMIT_CORE, &captured_core_rlimit_);
    }
  private:
    disable_core_dumps_capsule(const disable_core_dumps_capsule&);
    disable_core_dumps_capsule& operator = (const disable_core_dumps_capsule&);

    struct rlimit captured_core_rlimit_;
};
```

percona-ysorokin (changes requested):
A few suggestions on enabling/disabling core dumps.

unittest/gunit/thread_utils.cc
robgolebiowski: +1

unittest/gunit/thread_utils.h
laurynas-biveinis: Implemented. No ASSERT_EQ in set/getrlimit failure handling, as these macros then make the constructor/destructor to return a value

percona-ysorokin (approved):
LGTM

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.6-param/1924/

robgolebiowski (approved):
LGTM

PR #1676, opened 2017-04-20
MYR-105 : rocksdb.rocksdb_parts unstable

george-lorch:
- Added column mask for non-deterministic result from explain select and
  re-recorded

### DIFF
--- a/mysql-test/suite/rocksdb/r/rocksdb_parts.result
+++ b/mysql-test/suite/rocksdb/r/rocksdb_parts.result
@@ -57,7 +57,7 @@ insert into t1 values (11,20,20);
 insert into t1 values (12,20,20);
 explain select * from t1 force index(col1) where col1=10;
 id	select_type	table	partitions	type	possible_keys	key	key_len	ref	rows	filtered	Extra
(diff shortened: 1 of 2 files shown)


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/856/

mysql-test/suite/rocksdb/r/rocksdb_parts.result
```diff
@@ -57,7 +57,7 @@ insert into t1 values (11,20,20);
 insert into t1 values (12,20,20);
 explain select * from t1 force index(col1) where col1=10;
 id	select_type	table	partitions	type	possible_keys	key	key_len	ref	rows	filtered	Extra
-1	SIMPLE	t1	p0,p1	ref	col1	col1	5	const	2000	100.00	NULL
+1	SIMPLE	t1	p0,p1	ref	col1	col1	5	const	#	100.00	NULL
```
laurynas-biveinis, line 60: Is MyRocks table row count statistics approximate like in InnoDB?

george-lorch: Sometimes, like TokuDB, not even approximate but a wild guess.

PR #1681, opened 2017-04-24
Bug 1650294: Test main.audit_log_filter_users is unstable

gl-sergei:
The source of instability is the fact that event of creating new
connection and closing previous one are not synchronized. It makes
"Quit" record to appear after "Connect" one.

It cannot be easily resolved even with using debug_sync. Simplest fix is
to grep "Quit" events away.

### DIFF
--- a/mysql-test/r/audit_log_filter_users.result
+++ b/mysql-test/r/audit_log_filter_users.result
@@ -56,9 +56,6 @@ NULL	NULL
 SET GLOBAL audit_log_flush=ON;
 SET GLOBAL audit_log_flush=ON;
 SET GLOBAL audit_log_include_accounts= '';
-SELECT 'skip this';
-skip this
-skip this
 SELECT 'user1';
 user1
 user1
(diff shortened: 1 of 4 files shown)


gl-sergei:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1858/

gl-sergei:
Likely fixes 1626559 as well

laurynas-biveinis (changes requested):
I think this can be fixed by replacing include/wait_until_disconnected.inc (which waits for its own connection to be closed and then server carries on, async, on the remaining disconnect business) with include/wait_until_count_sessions.inc (which waits for a specific Threads_connected value in the server, which is updated close to the very end of disconnect)

gl-sergei:
It looks like a good idea. I changed tests to use wait_until_count_sessions.
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1864/

laurynas-biveinis (approved):

laurynas-biveinis:
If you believe this fixes 1626559 too, please set that one to Fix Released or duplicate or Incomplete

gl-sergei:
Now 1626559 doesn't look like duplicate... or maybe there are two different issues there. Some comments report that sometimes SHOW WARNINGS record appears in audit log. It needs further investigation.

PR #1684, opened 2017-05-01
Bug #1686934: Test rpl_semi_sync_ack_thread is failing on 5.7

robgolebiowski:
Test sets up a debug sync point "before_get_MASTER_UUID". This sync
point emits signal in_get_master_version_and_clock and waits for
singal get_master_uuid. The fix is to consume signal
in_get_master_version_and_clock as we care only about "wait for" part
in this test.

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_semi_sync_ack_thread.test
+++ b/mysql-test/suite/rpl/t/rpl_semi_sync_ack_thread.test
@@ -23,6 +23,8 @@
 # "Semi-sync master failed on net_flush() before waiting for slave reply"
 CALL mtr.add_suppression("Semi-sync master failed on net_flush().*");
 
+call mtr.add_suppression("Slave SQL.*Request to stop slave SQL Thread received while applying a group that has non-transactional changes; waiting for completion of the group");
+
 --echo ####################################################################
 --echo # Test Case: semisync master can be enabled and disabled sucessfully
 --echo # without any live slave connection and also test ON, OFF can be set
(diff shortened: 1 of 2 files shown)


robgolebiowski:
Jenkins run: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/872/

mysql-test/suite/rpl/t/rpl_semi_sync_ack_thread.test
```diff
@@ -240,6 +242,7 @@ SET GLOBAL debug= 'd,dbug.before_get_MASTER_UUID';
 --source include/install_semisync_slave.inc
 
 # Clear the debug point and signal IO thread to resume
+SET DEBUG_SYNC= 'now WAIT_FOR in_get_master_version_and_clock';
```
laurynas-biveinis, line 245: Can this be merged with two lines below: SET debug_sync= 'now WAIT_FOR this SIGNAL that' ?

laurynas-biveinis (changes requested):

mysql-test/suite/rpl/t/rpl_semi_sync_ack_thread.test
laurynas-biveinis: OK, as you wrote on Slack, the merged debug_sync signals first, before waiting. LGTM with two separate SETs

laurynas-biveinis (approved):

PR #1685, opened 2017-05-01
Bug#1687386: Test innodb.create_tablespace is failing on 5.7

robgolebiowski:
Test was failing on command mkdir $MYSQL_TMP_DIR/tablespace.ibd. The
reason is that, although we create this directory in tmp dir, the
directory itself can be already present in tmp dir. The fix is to
try remove the directory before creating it.


robgolebiowski:
Jenkins run: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/871/

laurynas-biveinis (approved):

PR #1686, opened 2017-05-01
Bug#1687432 rpl.rpl_read_only is failing on 5.7

robgolebiowski:
The test was failing because test run before it : rpl_percona_userstart
failed to clear REPLICATION FILTER. Thus user table was not replicated
in this test.


robgolebiowski:
Jenkins run: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/869/

laurynas-biveinis (approved):
LGTM. This is something MTR testcase check should be able to catch, filed https://bugs.launchpad.net/percona-server/+bug/1687521 for that

PR #1687, opened 2017-05-02
Bug#1687600 Test innodb.innodb_file_limit_check is failing on 5.7

robgolebiowski:
The test was failing when allowed number of innodb_open_files was
greater than limit set in this test i.e. 1000000. innodb_open_files
on Percona Jenkins machines can reach number higher than that.
On tested Jenkins machine it was 1048576. The fix was to set limit
to higher value i.e. 2000000.


robgolebiowski:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/883/

laurynas-biveinis (approved):

PR #1689, opened 2017-05-03
MYR-123 : PS 5.7 rocksdb_max_open_files differs from default

george-lorch:
- For plugin variables that are signed numbers, doing a SHOW VARIABLES will
  always show an unsigned number. Fix this by adding more types to the
  SHOW_* enums and then based on that, calling the *10_to_str functions
  appropriately.
- This commit is a port of upstream Facebook MySQL patch from
  https://github.com/facebook/mysql-5.6/commit/97acbd60eccf96962cdce9b01b2bc67f4291bcce


george-lorch:
Previous jenkins run with same code but different commit message http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/857/

include/mysql/plugin.h
```diff
@@ -143,6 +143,9 @@ enum enum_mysql_show_type
   SHOW_INT,        ///< shown as _unsigned_ int
   SHOW_LONG,       ///< shown as _unsigned_ long
   SHOW_LONGLONG,   ///< shown as _unsigned_ longlong
+  SHOW_SIGNED_INT,
+  SHOW_SIGNED_LONG,
```
laurynas-biveinis, line 147: Duplicate with the one in sql_plugin_enum.h. Probably that one has to be removed.

laurynas-biveinis (changes requested):

include/mysql/plugin.h
george-lorch: I think you are misreading the diff, SHOW_SIGNED_LONG was removed from sql_plugin_enum.h and added to the enum_mysql_show_type along with the new SHOW_SIGNED_INT.

laurynas-biveinis: Sorry, you are right

laurynas-biveinis (approved):

PR #1690, opened 2017-05-03
MYR-124 : START TRANSACTION WITH CONSISTENT SNAPSHOT behaves different

george-lorch:
- MYR-16 : myrocks/webscale uses different method to implement START
           TRANSACTION WITH CONSISTENT SNAPSHOT
  - This is where we diverged from upstream Facebook MySQL to make STWCS
    behave more like upstream Oracle MySQL in that is just returns SQL warning
    but not an error. Tests were unstable then so were never corrected.
- Corrected SQL warning to indicate the source of the warning is RocksDB.
- Enabled, fixed, and re-recorded snapshot tests to capture current, correct
  behavior in results.

### DIFF
--- a/mysql-test/suite/rocksdb/t/disabled.def
+++ b/mysql-test/suite/rocksdb/t/disabled.def
@@ -1,6 +1,4 @@
 compression_zstd : MYR-28 - Percona Server requires compilation with zstd
-cons_snapshot_read_committed : MYR-124 - Percona Server with MyRocks has different behavior of STWCS
-cons_snapshot_repeatable_read : MYR-124 - Percona Server with MyRocks has different behavior of STWCS
 cons_snapshot_serializable : MyRocks does not yet support ISO serializable
 level_read_uncommitted : MyRocks does not yet suppore ISO read uncommitted
 level_serializable : MyRocks does not yet support ISO serializable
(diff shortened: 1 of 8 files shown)


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/896/
Two instabilities appeared and have been logged as MYR-126 and MYR-127, unrelated to this change.

mysql-test/suite/rocksdb/t/disabled.def
```diff
@@ -1,6 +1,6 @@
 compression_zstd : MYR-28 - Percona Server requires compilation with zstd
-cons_snapshot_read_committed : MYR-124 - Percona Server with MyRocks has different behavior of STWCS
-cons_snapshot_repeatable_read : MYR-124 - Percona Server with MyRocks has different behavior of STWCS
+#cons_snapshot_read_committed : MYR-124 - Percona Server with MyRocks has different behavior of STWCS
```
laurynas-biveinis: Why comment out instead of removing?

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb/t/disabled.def
george-lorch: Good catch, those were supposed to be removed. The comment was temporary during testing so I could remember exactly which tests needed work. I'll fix and re-push before merging.

PR #1692, opened 2017-05-04
Bug#25604587 BACKPORT BUG#24947217 TO 5.6 AND 5.7

laurynas-biveinis:
Change is needed in MySQL 5.6 and 5.7 as well.

In the 5.5 backport, replace std::string with a DYNAMIC_STRING.

(cherry picked from commit 6f8e0f19de4e60eca8f648a472174aba9baa00b2)

http://jenkins.percona.com/job/percona-server-5.5-param/1550/


client/mysqltest.cc
```diff
@@ -4159,7 +4159,13 @@ void do_perl(struct st_command *command)
       die("Failed to create temporary file for perl command");
     my_close(fd, MYF(0));
 
-    str_to_file(temp_file_path, ds_script.str, ds_script.length);
+    /* Compatibility for Perl 5.24 and newer. */
+    static DYNAMIC_STRING script;
+    init_dynamic_string(&script, "push @INC, \".\";\n", 1024, 1024);
```
robgolebiowski, line 4164: Would it not be better to move push @INC, "." to ds_stript initialization, i.e. init_dynamic_string(&ds_script, "", 1024, 1024); and abandon script variable ?

robgolebiowski (changes requested):

client/mysqltest.cc
laurynas-biveinis: I am not sure control flow actually lets do that, there is "    command->content= ds_script;
" above

robgolebiowski: ack

robgolebiowski (approved):
LGTM

PR #1696, opened 2017-05-05
Bug#1688542: Test test_service_sql_api.test_sql_errors is failing on …

robgolebiowski:
…5.7 trunk

This test was creating 5 sessions from test plugin. These sessions were
meant to outlive the test plugin after it had been unloaded. However,
after the plugin was unloaded it was not possible to kill or close the
opened sessions, thus the check at the end of the test reported
existing threads. The fix was to remove opening of the sessions.
In test_sql_errors.cc there was a comment from upstream that suggested
that this test violets mtr post condition and was disabled from regular
test runs:

/* Disabled open without close as these 5 sessions stay open until
server shutdow, That violates the rules of a valid regression test. */

### DIFF
--- a/mysql-test/suite/test_service_sql_api/r/test_sql_errors.result
+++ b/mysql-test/suite/test_service_sql_api/r/test_sql_errors.result
@@ -101,6 +101,25 @@ ID	USER	HOST	DB	COMMAND	TIME	STATE	INFO	TIME_MS	ROWS_SENT	ROWS_EXAMINED
 ##########################################
 # Cleanup
 ##########################################
+CREATE PROCEDURE kill_sleeping_threads()
+BEGIN
+DECLARE done INT DEFAULT FALSE;
+DECLARE thread_id INT;
+DECLARE thread_id_curr CURSOR FOR SELECT ID FROM information_schema.processlist WHERE STATE = 'Sleep';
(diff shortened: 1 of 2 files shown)


robgolebiowski:
Jenkins run: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/901/

laurynas-biveinis (commented):
I have looked at this one before and attempted to fix by including a server restart in the testcase. This resulted in a crash and I haven't got around analysing it. Could you take a look?

robgolebiowski:
I have change the fix to killing threads and killing the server. Seems to work.
Jenkins run: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/910/

laurynas-biveinis:
Does plain restart_mysqld.inc w/o killing threads first result in a server crash?

robgolebiowski:
Yes it did, on joining threads in main server thread it crashed.

laurynas-biveinis:
Can you file an upstream bug for that crash? No need to fix it currently

laurynas-biveinis (approved):

robgolebiowski:
Create bug: Bug#1690171: Crash on restart, when session opened with srv_session_open, but not closed

laurynas-biveinis:
Please file an upstream bug as we wouldn't work on this bug immediately

PR #1697, opened 2017-05-05
Bug#1688319: Test binlog.binlog_gtid_state_update_deadlock is failing

robgolebiowski:
on 5.7 trunk

According to this test sync point after_log_file_name_initialized should
be initialized on changing GLOBAL.GTID_MODE. However, debug sync
after_log_file_name_initialized  was never hit on changing GTID_MODE.
The purpose of this debug sync point was to hold execution after
gtid_mode_lock was acquired, but before global_sid_lock was locked. The
fix was to add new debug sync
gtid_mode_update_gtid_mode_lock_wrlock_taken_will_take_global_sid_lock
which fulfills this requirements. Without the fix the test just timeout
on debug sync point   after_log_file_name_initialized – that is why
–include/big_test.inc was added to this test by upstream, which now have
been removed.


robgolebiowski:
Jenkins run: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/902/

laurynas-biveinis (approved):
LGTM. Since this one carries a source code patch, please submit to upstream / PS 8.0 seed branch too

robgolebiowski:
Created an upstream pull request, cannot find this test on PS 8.0 seed branch.

PR #1700, opened 2017-05-08
Fix bug 1688698 / 86209 (audit plugin + MB collation connection + PRE…

laurynas-biveinis:
…PARE stmt parse error crash)

Upstream revision [1] introduced code to check the previous token when
0 token is encountered. In the setup above (audit plugin + MB
collation connection + PREPARE statement with a parse error), the
previous token was also 0, thus not existing in the
digest_storage->m_token_array, and causing the previous token index
computed in digest_add_token and referenced in read_token to
overflow.

Fix by not checking the previous token if there isn't any.

http://jenkins.percona.com/job/percona-server-5.6-param/1870/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1713, opened 2017-05-10
Fix bug 1689552 (Missing synchronization between LRU manager and page…

laurynas-biveinis:
… cleaner threads on shutdown)

Synchronize shutdown of LRU manager and page cleaner threads by making
the page cleaner thread loop in the SRV_SHUTDOWN_CLEANUP loop even
after srv_shutdown_state moves to SRV_SHUTDOWN_FLUSH_PHASE, until the
LRU manager thread quits. Add read/write barriers to
buf_lru_manager_is_active accesses.

http://jenkins.percona.com/job/percona-server-5.6-param/1873/

### DIFF
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -2764,7 +2764,10 @@ DECLARE_THREAD(buf_flush_page_cleaner_thread)(
 	when SRV_SHUTDOWN_CLEANUP is set other threads like the master
 	and the purge threads may be working as well. We start flushing
 	the buffer pool but can't be sure that no new pages are being
-	dirtied until we enter SRV_SHUTDOWN_FLUSH_PHASE phase. */
+	dirtied until we enter SRV_SHUTDOWN_FLUSH_PHASE phase. Because
+	the LRU manager thread is also flushing at SRV_SHUTDOWN_CLEANUP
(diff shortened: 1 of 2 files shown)


percona-ysorokin (commented):
Do we also need to add memory barriers to `buf_LRU_get_free_block()` in `buf0lru.cc` where `buf_lru_manager_is_active` is also used?

laurynas-biveinis:
mutex_exit in buf_LRU_get_free_only call above it is close enough here (it's also OK to do one iteration in the wrong srv_empty_free_list_algorithm branch in case of a parallel buf_lru_manager_is_active change)

robgolebiowski:
This comment would need changing then: 	During shutdown we pass through two stages. In the first stage,
	when SRV_SHUTDOWN_CLEANUP is set other threads like the master
	and the purge threads may be working as well. We start flushing
	the buffer pool but can't be sure that no new pages are being
	dirtied until we enter SRV_SHUTDOWN_FLUSH_PHASE phase. */

It should be added that we need to wait for LRU manager to become inactive.

percona-ysorokin (approved):
LGTM

laurynas-biveinis:
Updated the comment, no new Jenkins run

robgolebiowski (approved):
LGTM

PR #1716, opened 2017-05-11
Add a testcase for bug 1689736 / 86260 (Assert on KILL'ing a stored r…

laurynas-biveinis:
…outine invocation)

5.5 is not affected by the bug. Only add the testcase and its
associated debug sync point.

http://jenkins.percona.com/job/percona-server-5.5-param/1555/


percona-ysorokin (approved):
LGTM

PR #1727, opened 2017-05-11
[BLD-719] new installation on debian/ubuntu installs the distribution my.cnf

EvgeniyPatlan:
Once installed on fresh installation:
Setting up percona-server-common-5.7 (5.7.18-14-1.trusty) ...
update-alternatives: using /etc/mysql/percona-server.cnf to provide /etc/mysql/my.cnf (my.cnf) in auto mode


tplavcic (approved):
LGTM

tplavcic:
I have approved this but there's one thing to think about - do all ubuntu/debian distros that we support have `mysql-common` package or on some it's already `mariadb-common`?
In this case if it's `mariadb-common` this might not work correctly.

PR #1729, opened 2017-05-12
Fix bug 1989989 (Test innodb.percona_changed_page_bmp_requests is uns…

laurynas-biveinis:
…table)

Close to the start of the test, server restarts twice in a row with no
write workload in between. This might result in a zero-sized bitmap
file, which would then get reused on the next restart, whereas the
testcase expects a new file to be created. Fix by adding some write
workload. At the same time simplify the testcase.

http://jenkins.percona.com/job/percona-server-5.6-param/1882/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1736, opened 2017-05-15
Bug#1690531: Test rpl.rpl_killed_ddl is failing on 5.7

robgolebiowski:
Event created by this test was to be executed on a date based on
CURRENT_TIMESTAMP. Later creation of this event was executed on slave
and a diff of statements between master and slave was done. This diff
resulted in failure for statement based replication, as
CURRENT_TIMESTAMP on slave was different than the CURRENT_TIMESTAMP on
master. Fixed by replacing CURRENT_TIMESTAMP with fixed timestamp.

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_killed_ddl.test
+++ b/mysql-test/suite/rpl/t/rpl_killed_ddl.test
@@ -37,7 +37,7 @@ source include/master-slave.inc;
 # Use the DBUG_SYNC_POINT to make sure the thread running the DDL is
 # waiting before creating the query log event
 
-let $debug_lock= "debug_lock.before_query_log_event";
+#let $debug_lock= "debug_lock.before_query_log_event"; #uncomment this line to enable debug
 
 ######## INITIALIZATION ########
 
--- a/mysql-test/include/kill_query_and_diff_master_slave.inc
+++ b/mysql-test/include/kill_query_and_diff_master_slave.inc
@@ -19,6 +19,7 @@
 # information
 
 source include/kill_query.inc;
+source include/sync_slave_sql_with_master.inc;
 
 # Release the debug lock if used, so that the statements in
 # rpl_diff.inc will not be blocked.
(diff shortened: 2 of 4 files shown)


robgolebiowski:
jenkins run: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/925/

mysql-test/suite/rpl/t/rpl_killed_ddl.test
```diff
@@ -162,12 +162,12 @@ let $rpl_diff_statement= SELECT event_name, event_body, execute_at
                          ORDER BY event_name;
 
 send CREATE EVENT e2
-  ON SCHEDULE AT CURRENT_TIMESTAMP + INTERVAL 1 DAY
+  ON SCHEDULE AT '2020-01-01 00:00:00' + INTERVAL 1 DAY
```
percona-ysorokin: 2020 is just 3 years from now - it does not look unrealistic.
Funny thing is that Oracle fixed similar problem in its latest minor releases
https://github.com/percona/percona-server/commit/e7a3ccf6b1b1b41e0fa191d26e9e775517c6deee

May be, we can still use `CURRENT_TIMESTAMP` but eliminate the problem on MTR level. Something like
```sql
--let $timestamp= `SELECT CURRENT_TIMESTAMP()`
send_eval CREATE EVENT e2
  ON SCHEDULE AT '$timestamp' + INTERVAL 1 DAY
```

percona-ysorokin (changes requested):

mysql-test/suite/rpl/t/rpl_killed_ddl.test
robgolebiowski: It does not have to look unrealistic - it can be any date. It is to test what happens when ddl gets killed - in this case creating event.

percona-ysorokin: As far as I remember, events cannot be created in the past. So, my concern is that in 3 years this test  will be failing.

robgolebiowski: Oh, good to know, will change this, thanks

PR #1738, opened 2017-05-16
The ability to kill query blocked in PerconaFT lock tree is added.

vlad-lesin:
TDB-27 : Add ability to kill query that is awaiting FT locktree lock

MTR test uses DEBUG_SYNC built in PerconaFT.

See also https://github.com/percona/percona-server/pull/1739, https://github.com/percona/PerconaFT/pull/370.

Testing:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1896/


george-lorch (approved):
Due to the dependency on https://github.com/percona/PerconaFT/pull/370 with the API change, I suggest that you add the movement of the PerconaFT submodule commit pointer to this branch and PR so that the FT change and the DB change are merged atomically, otherwise there will be a brief window of time where DB will fail to build against the new FT API.

PR #1739, opened 2017-05-16
Tokudb kill query in locktree 5.7 gca

vlad-lesin:
TDB-27 : Add ability to kill query that is awaiting FT locktree lock

See https://github.com/percona/percona-server/pull/1738, https://github.com/percona/PerconaFT/pull/370.

Testing:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/934/

### DIFF
--- a/mysql-test/suite/tokudb/r/kill_query_blocked_in_lt.result
+++ b/mysql-test/suite/tokudb/r/kill_query_blocked_in_lt.result
@@ -1,0 +1,26 @@
+### connection default
+CREATE TABLE t (a INT PRIMARY KEY, b INT) ENGINE=TokuDB;
+INSERT INTO t (a, b) VALUES (1, 1), (2, 2), (3, 3), (4, 4), (5, 5),
+(6, 6), (7, 7), (8, 8), (9, 9), (10, 10),
+(11, 11), (12, 12), (13, 13), (14, 14), (15, 15),
+(16, 16), (17, 17), (18, 18), (19, 19), (20, 20);
+### connection con1
+SET DEBUG_SYNC= 'toku_range_lock_granted_immediately SIGNAL lock_granted WAIT_FOR lock_granted_continue';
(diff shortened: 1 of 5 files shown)


georgelorchpercona:
This does not build for me with an out of source build using the following cmake incantation
`cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=../percona-server-install-5.7 -DMYSQL_MAINTAINER_MODE=OFF -DWITH_EMBEDDED_SERVER=OFF -DWITH_JEMALLOC=system -DWITH_BOOST=../../boost_1_59_0 -DZLIB_INCLUDE_DIR=/usr/include/ $@ ../percona-server`

include/my_global.h
```diff
@@ -28,7 +28,9 @@
 #include "my_config.h"
 
 #define __STDC_LIMIT_MACROS	/* Enable C99 limit macros */
+#ifndef __STDC_FORMAT_MACROS
 #define __STDC_FORMAT_MACROS	/* Enable C99 printf format macros */
+#endif /* __STDC_FORMAT_MACROS */
 #define _USE_MATH_DEFINES       /* Get access to M_PI, M_E, etc. in math.h */
```
george-lorch: Why is this changed? This is in server code and will require a launchpad bug or blueprint explaining the change.

george-lorch: I see why, PerconaFT sets it in TokuSetupCompiler.cmake and that conflicts with the my_global.h definition, but I think the problem is not the server code but PerconaFT and probably best addressed there if possible. Even if allowed in my_global.h, shouldn't all __STDC definitions here get the same treatment to prevent other/future issues?

vlad-lesin: Agree, fixed in PerconaFT: portability/toku_debug_sync.h. Waiting for testing results...

george-lorch (approved):
Same comments re: FT submodule commit pointer as 5.6

PR #1740, opened 2017-05-16
Fix for lp:1671125.

vlad-lesin:
The issue was in reset debug_sync actions before the certain signal
was processed in the certain point. The solution it to remove debug_sync reset
from the test, the reset is not necessary in the test.

See also: https://github.com/percona/percona-server/pull/1741

Testing:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1897/

PR #1741, opened 2017-05-16
Fix for lp:1671125.

vlad-lesin:
The issue was in reset debug_sync actions before the certain signal
was processed in the certain point. The solution it to remove debug_sync reset
from the test, the reset is not neccessary in the test.

(cherry picked from commit b3a941e2752da0f0e50fce546641eea5da38eff2)

See also: https://github.com/percona/percona-server/pull/1740

Testing:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/940/


vlad-lesin:
As we have new GA, I would like to rebase this commit on the newest GA and don't use cherry-pick for 5.7.

vlad-lesin:
Closing, see: https://github.com/percona/percona-server/pull/1786 .

PR #1742, opened 2017-05-17
Update source and binary paths in installation.rst

zintrepid:
This also corrects the name of the source tarball in the example tar command to extract source.


hrvojem (approved):
LGTM

PR #1746, opened 2017-05-19
Fix lp:1657941 (Server crashes when querying partitioning table MySQL_5.7.14) (5.7)

percona-ysorokin:


percona-ysorokin:
Jenkins build link
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/948/

george-lorch:
Did you/can you validate that this is also fixed for TokuDB or other engines as well? 5.7 split up the partition handler for native partitioning which InnoDB is the only one that supports it, so I am curious if this change is common to both native and legacy or if the change needs to be done in two places. Just grepping the code makes me believe this looks like common to both but hard to say for sure.

robgolebiowski (approved):
LGTM

percona-ysorokin:
@georgelorchpercona 
This is 100% SQL-level bug (not specific SE-level), so it is storage engine independent.
Just in case I double checked the same `percona_bug1657941` with both MyISAM and TokuDB storage engines on 5.7.

george-lorch (approved):

laurynas-biveinis:
Partition testcases should go to parts suite

PR #1751, opened 2017-05-22
Ps 5.7 tdb 2

george-lorch:


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/951/

george-lorch:
Just advancing submodule commit pointer to pick up new FT changes, self-approving

PR #1752, opened 2017-05-23
Preliminary fix for bug 1679025 / 85835.

percona-ysorokin:


percona-ysorokin:
This is a quick fix from @laurynas-biveinis 
Although the main problem seems to be much deeper, the fix was confirmed by the customer who requested it.

george-lorch (approved):
@percona-ysorokin I read through the internal issue history and lightly reviewed the patch. I have never been through this area before so I may be missing many details, but in general, the fix looks harmless. I think though we need to ensure that QA looks very close at this and any impacts it might have to queries that make use of LIMIT to ensure the correct result is returned.

robgolebiowski (approved):
LGTM

percona-ysorokin:
Jenkins build link
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/953/

laurynas-biveinis:
I wrote to @peterschwaller why I thought LIMIT handling needs further review

Please prepare upstream / 8.0 seed submission too

RoelVdP:
Discussed with @laurynas-biveinis and agreed that - as far as this ticket and limited patch goes 
 an additional QA run/review is not required.

PR #1753, opened 2017-05-23
[TDB-6] src/ydb.cc:2917 can_acquire_table_lock: Assertion `r == 0'

Jun-Yuan:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/969/


george-lorch (changes requested):
Same general comments as 5.6

Jun-Yuan:
Thanks. How do these comments(both 5.6 and 6.7) look like now?

george-lorch (approved):

PR #1756, opened 2017-05-24
 lp1686603 : mysqldump execute 'SELECT 0' with params - ERROR 1210 (H…

george-lorch:
…Y000)

- The issue stemmed from a change to mysqldump to facilitate setting a MyRocks
  specific session variable that help with bulk load performance during the
  loading of a generated dump if MyRocks is present on the destination/load
  server.
  In order to prevent the use of this MyRocks specific variable when MyRocks
  it is not present on the destination server, some SQL probing code was added
  to prevent the setting of the variable if it can not be detected to exist on
  the destination.
  This probing SQL checks the performance_schema.session_variables table to test
  for the MyRocks variables existance.
  This brings us to the cause of the error: there are cases/conditions where
  the performance_schema.session_variables either does not exist at all
  (compiled out), or, it is accessable, but does not appear in the
  INFORMATION_SCHEMA.TABLES table for whatever reason (is this some other
  bug?). When the p_s table could not be detected, the probe SQL code prepared a
  procedure that took no parameters, but then tried to execute it with a
  parameter, resulting in the "'SELECT 0' with params - ERROR 1210 (HY000)".
- Fixed the mysqldump code to be consistent in how it prepares procedures
  to expect parameters.
- Introduced a new test, percona_bug1686603, that fully exercises and
  reproduces the original error, then recorded it upon implementing the
  correction to mysqldump.
- Re-recorded main.mysqldump test to account for different output.
- Re-recorded main.mysqldump-max test to account for different output.
- Re-recorded main.openssl_1 test to account for different output.
- Re-recorded percona_bug1676401 test to account for different output.
- Re-recorded percona_mysqldump_innodb_oprimize_keys test to account
  for different output.
- Re-recorded rocksdb.mysqldump test to account for different output.


george-lorch:
Here is a straight jenkins run without MyRocks : http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/955/

Here is a straight jenkins run with MyRocks built and 'visible' to the servers in mtr : http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/956/

Here is a jenkins run with just the MyRocks suites : http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/957/

george-lorch:
I added a little more situational validation and cleanup to percona_bug1686603.test and launched another jenkins run on just this test --repeat=3

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/960/

mysql-test/t/percona_bug1686603.test
```diff
@@ -0,0 +1,27 @@
...
+CREATE TABLE dump.t(a int primary key auto_increment);
+INSERT INTO dump.t VALUES(NULL), (NULL), (NULL), (NULL);
+
+--connect (conn1,localhost,dumper,,dump)
+--connection conn1
+SELECT COUNT(*) FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA='performance_schema' AND TABLE_NAME='session_variables';
+--connection default
+--disconnect conn1
+
+--exec $MYSQL_DUMP dump t > $MYSQLTEST_VARDIR/tmp/t.sql
```
percona-ysorokin: I would use a more test-case-specific name here, like `percona_bug1686603_t.sql`.
`t.sql` seems to be very common, may have collisions with other tests.

mysql-test/t/percona_bug1686603.test
```diff
@@ -0,0 +1,27 @@
+--source include/count_sessions.inc
+
+CREATE DATABASE dump;
+CREATE USER 'dumper'@'localhost';
```
percona-ysorokin: Why do we need additional user/connection at all?

george-lorch:
OK, renamed temp sql file and added a comment to the test as well. Here is the diff just pushed:
```
diff --git a/mysql-test/t/percona_bug1686603.test b/mysql-test/t/percona_bug1686603.test
index a104797..ac54c19 100644
--- a/mysql-test/t/percona_bug1686603.test
+++ b/mysql-test/t/percona_bug1686603.test
@@ -9,18 +9,21 @@ INSERT INTO dump.t VALUES(NULL), (NULL), (NULL), (NULL);
 
 --connect (conn1,localhost,dumper,,dump)
 --connection conn1
+# this _must_ be 0 to correctly test the core of the issue. It means that even
+# though the p_s.session_variables table may be accessable, it still doesn't
+# show in I_S
 SELECT COUNT(*) FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA='performance_schema' AND TABLE_NAME='session_variables';
 --connection default
 --disconnect conn1
 
---exec $MYSQL_DUMP dump t > $MYSQLTEST_VARDIR/tmp/t.sql
+--exec $MYSQL_DUMP dump t > $MYSQLTEST_VARDIR/tmp/percona_bug1686603_t.sql
 DROP TABLE dump.t;
 
---exec $MYSQL -u dumper dump < $MYSQLTEST_VARDIR/tmp/t.sql
+--exec $MYSQL -u dumper dump < $MYSQLTEST_VARDIR/tmp/percona_bug1686603_t.sql
 
 SELECT * FROM dump.t;
 
---remove_file $MYSQLTEST_VARDIR/tmp/t.sql
+--remove_file $MYSQLTEST_VARDIR/tmp/percona_bug1686603_t.sql
 DROP DATABASE dump;
 DROP USER 'dumper'@'localhost';
```

And here is a new percona_bug1686603.test jenkins run with  --repeat=3 http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/964/

percona-ysorokin (approved):
LGTM

PR #1759, opened 2017-05-25
TDB-44 : tokudb.locks-select-update-3 is unstable

george-lorch:
- Increased lock wait timeout to allow for slow ystem switching time between
  concurrent mtr connections.
- Modernized test a bit introducing count_sessions.inc and removing
  'drop table .. if exists'


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1918/

Self approving as change is very limited to mtr test scope, is very minor, and passes jenkins cleanly.

PR #1760, opened 2017-05-25
Ps 5.7 tdb 44

george-lorch:


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/967/

Self approving as change is very limited to mtr test scope, is very minor, and passes jenkins cleanly.

PR #1763, opened 2017-05-25
TDB-42 : Misleading progress status for partitioned TokuDB table ALTERs

george-lorch:
- This is a display/status issue where TokuDB really doesn't know that it is
  operating on a partition of a table or how many partitions there are, or how
  much data is in the other partitions. As far as it is concerned it is working
  on a single table.
- To make the display issue a little clearer, added the ful table name to the
  status, thus making it visible which partition is currently being processed.
- Also applied this same idea to bulk the bulk loader since it was right there.
- Restructured the code a little bit to eliminate global static functions and
  provide proper access to the table name from the TOKUDB_SHARE object.

### DIFF
--- a/storage/tokudb/ha_tokudb.cc
+++ b/storage/tokudb/ha_tokudb.cc
@@ -532,51 +532,6 @@ typedef struct index_read_info {
     DBT* orig_key;
 } *INDEX_READ_INFO;
 
-static int ai_poll_fun(void *extra, float progress) {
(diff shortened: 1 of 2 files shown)


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1919/

george-lorch:
Self approving...

PR #1766, opened 2017-05-28
Add diagnostics for bug 1690298 (Test engines/funcs.db_alter_characte…

laurynas-biveinis:
…r_set is unstable)

Currently, the only charset loading error that gets diagnosed is
Index.xml XML parse error. Print diagnostics in case
my_read_charset_file finds a too-large Index.xml file, and enable
diagnostics for other failures modes by passing MY_WME to it from
init_available_charsets. Only limit this change to server as it's quite
regular for clients to run without loading charsets.

http://jenkins.percona.com/job/percona-server-5.6-param/1920/

### DIFF
--- a/mysys/charset.c
+++ b/mysys/charset.c
@@ -416,10 +416,23 @@ my_read_charset_file(MY_CHARSET_LOADER *loader,
   int  fd;
   size_t len, tmp_len;
   MY_STAT stat_info;
-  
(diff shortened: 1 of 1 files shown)


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1770, opened 2017-05-29
Fix bug 1675256/85588 (Regular expressions work incorrectly for strin…

Olernov:
…gs having 00 byte)


Olernov:
Fix for https://bugs.launchpad.net/percona-server/+bug/1675256

PR #1772, opened 2017-05-29
TDB-41 : PerconaFT does not pass cmake flags on to snappy cmake

george-lorch:
- Advanced git submodule commit pointer


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1923/

george-lorch:
Self approving, just moving submodule commit pointer

PR #1773, opened 2017-05-29
TDB-41 : PerconaFT does not pass cmake flags on to snappy cmake

george-lorch:
- Advanced git submodule commit pointer


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/982/

george-lorch:
Self approving

PR #1774, opened 2017-05-30
Fix bug 1684601 / 86016 (Make MTR show core dump stacktraces from uni…

laurynas-biveinis:
…t tests too)

Implement MTR showing the stacktraces of crashed unit tests by the
following:
- at the end of test run, if --unit-tests was set, search for core
  dumps in the build directory subtree and invoke My::CoreDump->show
  on them;
- removing _verify_binpath call in My::CoreDump::_gdb as it does not
  seem to be required on our supported platforms and if present, it
  assigns all the unit test core dumps to mysqltest executable.
- remove TAP signal handling so that the interesting signals generate
  core dumps by default

http://jenkins.percona.com/job/percona-server-5.5-param/1560/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1779, opened 2017-06-02
Fix lp:1673656 "SSL Certificate Subject ALT Names with IPs or DNS: not respected with --ssl-verify-server-cert"

percona-ysorokin:
Original implementation of 'ssl_verify_server_cert()' in
'sql-common/client.c', which corresponds to '--ssl-verify-server-cert' client
option, did not respect neither wildcard nor SAN-enabled SSL certificates. It
just checked 'commonName' ('CN') field for equality with server host name.

Fixed by calling 'X509_check_host()' instead.
Please notice that this function is only available on platforms with
OpenSSL >= 1.0.2.
On other platforms, first we iterate over DNS entries in the SAN list
('NID_subject_alt_name') and, only if this list does not exist or is empty,
we check 'commonName' ('CN').

Also fixed inconsistency with 'errptr' being set to
"SSL certificate validation failure" even when the function returns
with non-error code.

As Oracle does not include private key along with self-signed CA certificate
'std_data/cacert.pem', generated a new self-signed Percona-specific CA
certificate 'std_data/percona-cacert.pem'. Corresponding private key
'std_data/percona-cakey.pem' also added to the source control.

Generated new Percona-specific server certificate
'std_data/percona-serversan-cert.pem' signed with
'std_data/percona-cacert.pem' which has 'CN' == 'percona-server' and
'localhost' in 'SAN' (Subject Alternative Names from v3). Corresponding
private key 'std_data/percona-serversan-key.pem' also added to the source
control.

Added the bash script which was used to generate Percona CA an Percona Server
certificates ('mysql-test/lib/generate_percona_ssl_certs.sh').

Added new MTR test case 'main.ssl_san' which checks if it is possible to
connect to the server with '--ssl-verify-server-cert' option provided that
it has SAN-enabled SSL certificate. This test case must be skipped for any
Percona Server build based on a crypto library other than OpenSSL.


percona-ysorokin:
Jenkins build link
http://jenkins.percona.com/view/PS%205.5/job/percona-server-5.5-param/1567 (OpenSSL)
http://jenkins.percona.com/view/PS%205.5/job/percona-server-5.5-param/1568 (YaSSL)

george-lorch (approved):
Nice work.

mysql-test/lib/generate_percona_ssl_certs.sh
```diff
@@ -0,0 +1,25 @@
...
+set -xe
+
+# boilerplace for "openssl ca" and /etc/ssl/openssl.cnf
+rm -rf demoCA
+mkdir demoCA demoCA/newcerts
+touch demoCA/index.txt
+echo 01 > demoCA/serial
+
+# Generate a new self-signed CA certificate
+openssl req -x509 -newkey rsa:2048 -keyout percona-cakey.pem -out percona-cacert.pem -days 7300 -nodes -subj '/CN=percona-cacert/C=US/ST=NC/L=Raleigh/O=Percona' -text
```
robgolebiowski, line 12: add also -outform PEM

mysql-test/lib/generate_percona_ssl_certs.sh
```diff
@@ -0,0 +1,25 @@
...
+# The following operation, althought looking redundant, will remove 'passphrase' from the RSA key (which is currently unsupported in YaSSL)
+openssl rsa -in percona-cakey.pem -out percona-cakey.pem
+
+# Generating a new certificate signing request for the server
+openssl req -newkey rsa:2048 -keyout percona-serversan-key.pem -out demoCA/percona-serversan-req.pem -days 7300 -nodes -subj '/CN=percona-server/C=US/ST=NC/L=Raleigh/O=Percona'
```
robgolebiowski: certificate signing request have usually .csr extension

sql-common/client.c
```diff
@@ -1897,15 +1897,27 @@ mysql_get_ssl_cipher(MYSQL *mysql __attribute__((unused)))
 
 #if defined(HAVE_OPENSSL) && !defined(EMBEDDED_LIBRARY)
 
+#include <openssl/x509v3.h>
+
+#if !defined(HAVE_YASSL) && !defined(LIBRESSL_VERSION_NUMBER)
+  #if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER >= 0x10002000L
+    #define HAVE_X509_check_host 1
```
robgolebiowski: I believe this can be done from CMakeLists.txt level with:
CHECK_FUNCTION_EXISTS(HAVE_X509_check_host HAVE_X509_CHECK_HOST)

mysql-test/lib/generate_percona_ssl_certs.sh
```diff
@@ -0,0 +1,25 @@
...
+# The following operation, althought looking redundant, will remove 'passphrase' from the RSA key (which is currently unsupported in YaSSL)
+openssl rsa -in percona-serversan-key.pem -out percona-serversan-key.pem
+# Generate a new server certificate signed with new CA
+# with SubjectAltName, only for OpenSSL 1.0.2+
+cat > demoCA/sanext.conf <<EOF
+subjectAltName=DNS:localhost
```
robgolebiowski: 1. Apart from DNS:localhost subjectAltName can also take IP addresses. Please add test for IP also, for instance IP:127.0.0.1.
2. Also we should have a test certificate which would allow multiple subjectAltNames and it should be tested

sql-common/client.c
```diff
@@ -1948,50 +1960,116 @@ static int ssl_verify_server_cert(Vio *vio, const char* server_hostname, const c
...
+    int number_of_sans= sk_GENERAL_NAME_num(sans);
+    for (i= 0; ret_validation != 0 && i < number_of_sans; ++i)
+    {
+        GENERAL_NAME *san= sk_GENERAL_NAME_value(sans, i);
+        if(san == NULL)
+        {
+          *errptr = "Failed to get item from SAN list";
+          goto error;
+        }
+        if (san->type == GEN_DNS)
```
robgolebiowski: What if subjectAltName will be provided in form IP:127.0.0.1 (example) ? I do not see support for this.

sql-common/client.c
```diff
@@ -1897,15 +1897,27 @@ mysql_get_ssl_cipher(MYSQL *mysql __attribute__((unused)))
...
 
+#include <openssl/x509v3.h>
+
+#if !defined(HAVE_YASSL) && !defined(LIBRESSL_VERSION_NUMBER)
+  #if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER >= 0x10002000L
+    #define HAVE_X509_check_host 1
+  #endif
+#endif
+
 static int ssl_verify_server_cert(Vio *vio, const char* server_hostname, const char **errptr)
```
robgolebiowski: General remark on this function - it is huge. Would you consider splitting it into smaller functions - it would be much easier to read. There is quite a distinction between SAN validation and CN validation.

sql-common/client.c
```diff
@@ -1948,50 +1960,116 @@ static int ssl_verify_server_cert(Vio *vio, const char* server_hostname, const c
...
+            goto error;
+          }
+          // There should not be any NULL embedded in the DNS name
+          if ((size_t)ASN1_STRING_length(cn_asn1) != strlen(cn))
+          {
+            *errptr = "NULL embedded in the certificate SAN DNS name";
+            goto error;
+          }
+          DBUG_PRINT("info", ("SAN DNS name in cert: %s", cn));
+          if (!strcmp(cn, server_hostname))
```
robgolebiowski: I do not see support for wildcards in SAN.

sql-common/client.c
```diff
@@ -1948,50 +1960,116 @@ static int ssl_verify_server_cert(Vio *vio, const char* server_hostname, const c
...
+
+    DBUG_PRINT("info", ("Server hostname in cert: %s", cn));
+    if (!strcmp(cn, server_hostname))
+    {
+      /* Success */
+      ret_validation= 0;
+    }
+  }
+#endif
+  *errptr= (ret_validation != 0 ? "SSL certificate validation failure" : "");
```
robgolebiowski: brackets are not needed

robgolebiowski (changes requested):

mysql-test/lib/generate_percona_ssl_certs.sh
percona-ysorokin: looks like PEM is the default format even without `-outform PEM`

mysql-test/lib/generate_percona_ssl_certs.sh
percona-ysorokin: fixed

mysql-test/lib/generate_percona_ssl_certs.sh
percona-ysorokin: Added the following section
```
DNS.1 = localhost
DNS.2 = localhost.localdomain
IP.1 = 127.0.0.1
IP.2 = ::1
```

sql-common/client.c
percona-ysorokin: Fixed. both IPV4 and IPV6 entries are now  checked. MTR test case extended accordingly.

sql-common/client.c
percona-ysorokin: This patch won't include wildcards support.

sql-common/client.c
percona-ysorokin: The problem is that `sql_common/client.c` is used in 3 different  targets (`libmysql`, `libmysqld` and `mysqld`). In my opinion, it's better to have these defines in one place rather than put `CHECK_FUNCTION_EXISTS` in 3 different `CMakeLists.txt` files and have potential  "updated only in one place" problem. Leaving as is.

sql-common/client.c
percona-ysorokin: SAN-related checks put into separate `ssl_verify_server_cert_san()` function.

sql-common/client.c
percona-ysorokin: fixed

percona-ysorokin:
Updated Jenkins build links:
http://jenkins.percona.com/view/PS%205.5/job/percona-server-5.5-param/1581 (OpenSSL)
http://jenkins.percona.com/view/PS%205.5/job/percona-server-5.5-param/1577 (YaSSL)

sql-common/client.c
robgolebiowski: I think it could be located in configure.cmake

mysql-test/t/ssl_san_ipv6.test
```diff
@@ -0,0 +1,39 @@
...
+# connections for testing, put them after 'count_sessions'
+--source include/check_ipv6.inc
+--source include/have_ipv4_mapped.inc
+
+#
+# lp:1673656 "SSL Certificate Subject ALT Names with IPs or DNS: not respected with --ssl-verify-server-cert"
+# Oracle Bug #68052 "SSL Certificate Subject ALT Names with IPs not respected with --ssl-verify-serve"
+#
+
+# This is a IPV6 companion of 'main.ssl_san' MTR test case
```
robgolebiowski, line 23: Isn't this test also checking v4 ? If so - this comment and the test name is misleading.

sql-common/client.c
```diff
@@ -1878,6 +1878,128 @@ mysql_get_ssl_cipher(MYSQL *mysql __attribute__((unused)))
...
+    {
+      *errptr= "Failed to get item from SAN list";
+      goto error;
+    }
+    if(hostname_or_ip_len == 0)
+    {
+      /* server host name was provided, check only GEN_DNS entries */
+      if (san->type == GEN_DNS)
+      {
+        char* cn;
```
robgolebiowski: handling for GET_DNS and GET_IPADD could be moved into separate functions.

robgolebiowski (changes requested):

sql-common/client.c
```diff
@@ -1878,6 +1878,128 @@ mysql_get_ssl_cipher(MYSQL *mysql __attribute__((unused)))
 }
 
 
+#if defined(HAVE_OPENSSL) && !defined(EMBEDDED_LIBRARY)
+
+#include <openssl/x509v3.h>
+
+#if !defined(HAVE_YASSL) && (OPENSSL_VERSION_NUMBER >= 0x10002000L)
```
robgolebiowski: I think it could be located in configure.cmake

mysql-test/t/ssl_san_ipv6.test
percona-ysorokin: No, this test checks only IPV6 addresses: `::1` and `::ffff:127.0.0.1` (IPV6  mapped IPV4 localhost)

sql-common/client.c
percona-ysorokin: Done

sql-common/client.c
percona-ysorokin: `HAVE_X509_CHECK_HOST` and `HAVE_X509_CHECK_IP` are now detected in `ssl.cmake`

sql-common/client.c
```diff
@@ -1878,6 +1878,190 @@ mysql_get_ssl_cipher(MYSQL *mysql __attribute__((unused)))
 }
 
 
+#if defined(HAVE_OPENSSL) && !defined(EMBEDDED_LIBRARY)
+
+#include <openssl/x509v3.h>
+
+#if defined(HAVE_X509_CHECK_HOST) && defined(HAVE_X509_CHECK_IP)
```
robgolebiowski, line 1885: I think this also could be moved into ssl.cmake - if both variables exist, set a new variable and add #cmakedefine on it.

percona-ysorokin: I was also thinking about doing this but for testing purposes (to be able to easily switch between 'emulated' and 'native' implementations on platforms with newer OpenSSL) it's better to have a single define in a source file rather than move it to cmake level. I would prefer to leave it as is.

robgolebiowski (approved):
LGTM

percona-ysorokin:
Jenkins build links
http://jenkins.percona.com/view/PS%205.5/job/percona-server-5.5-param/1582 (OpenSSL)
http://jenkins.percona.com/view/PS%205.5/job/percona-server-5.5-param/1584 (YaSSL)

PR #1782, opened 2017-06-05
Bug lp:1594964 fix.

vlad-lesin:
1) In Rpl_info_table::do_init_info() and some other Rpl_info_table::* functions
the thd is gotten with the following code:
```c++
  THD *thd= access->create_thd();
```

2) If we look into Rpl_info_table_access::create_thd() we will see that thd is
created only in the case if the current_thd is NULL, otherwise the current_thd
is returned. For our case that means the replication thread is used to store
data in rpl info table.

```c++
THD *Rpl_info_table_access::create_thd()
{
  THD *thd= current_thd;

  if (!thd)
  {
    thd= System_table_access::create_thd();
    thd->system_thread= SYSTEM_THREAD_INFO_REPOSITORY;
    thd_created= true;
  }

  return(thd);
}
```

3) If wee look at the massif.out file we will see that new memory is allocated
with "new" operator with parameters, and thd->mem_root is used as the
parameter, so the memory is allocated in thd->mem_root arena.

```
 ->09.32% (578,806,592B) 0x188E91C: my_raw_malloc (in /usr/sbin/mysqld-debug)
| ->09.32% (578,806,592B) 0x188E519: my_malloc (in /usr/sbin/mysqld-debug)
| ->09.03% (561,046,608B) 0x1887A44: alloc_root (in /usr/sbin/mysqld-debug)
| | ->08.75% (543,172,776B) 0xEFC948: Sql_alloc::operator new(unsigned long, st_mem_root*) (in /usr/sbin/mysqld-debug)
| | | ->06.25% (388,424,352B) 0x153B929: LEX::new_query(st_select_lex*) (in /usr/sbin/mysqld-debug)
| | | | ->06.25% (388,424,352B) 0x153BEEE: LEX::new_top_level_query() (in /usr/sbin/mysqld-debug)
| | | | ->06.25% (388,424,352B) 0x153B5E2: lex_start(THD*) (in /usr/sbin/mysqld-debug)
| | | | ->06.25% (388,424,352B) 0x1858CB9: Rpl_info_table_access::before_open(THD*) (in /usr/sbin/mysqld-debug)
| | | | | ->06.25% (388,424,352B) 0x181FA59: System_table_access::open_table(THD*, st_mysql_lex_string, st_mysql_lex_string, unsigned int, thr_lock_type, TABLE**, Open_tables_backup*) (in /usr/sbin/mysqld-debug)
| | | | | ->06.25% (388,424,352B) 0x186E3D3: Rpl_info_table::do_flush_info(bool) (in /usr/sbin/mysqld-debug)
```

4) If we look at the end of Rpl_info_table::do_init_info() and some other
Rpl_info_tables::* functions we will see the call of access->drop_thd(thd)
which code is the following:

```c++
void Rpl_info_table_access::drop_thd(THD *thd)
{
  DBUG_ENTER("Rpl_info::drop_thd");

  if (thd_created)
  {
    System_table_access::drop_thd(thd);
    thd_created= false;
  }

  DBUG_VOID_RETURN;
}
```

For our case thd_created is false, and that is why thd is not deleted and
the memory managed with thd->mem_root is not deallocated.

The solution is to use local MEMROOT instance which would be freed after the
corresponding function finish threir work with thd.

Testing: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/993/
See also: https://bugs.launchpad.net/percona-server/+bug/1594964
The simple way to check if the fix fixes anything: https://bugs.launchpad.net/percona-server/+bug/1594964/comments/8

### DIFF
--- a/sql/rpl_info_table.cc
+++ b/sql/rpl_info_table.cc
@@ -22,6 +22,50 @@
 #include "sql_class.h"              // THD
 
 
+class Thd_local_mem_root
+{
+  THD *m_thd;
+  MEM_ROOT *m_old_mem_root;
+  MEM_ROOT m_local_mem_root;
+  bool m_restored;
+
+  Thd_local_mem_root(const Thd_local_mem_root&);
+  Thd_local_mem_root& operator=(const Thd_local_mem_root&);
+
+public:
+
+  Thd_local_mem_root(THD *thd) : m_thd(thd), m_restored(false)
+  {
+    if (m_thd == current_thd)
+    {
+      m_old_mem_root= thd->mem_root;
+      init_sql_alloc(PSI_NOT_INSTRUMENTED,
+                     &m_local_mem_root,
+                     global_system_variables.query_alloc_block_size,
+                     global_system_variables.query_prealloc_size);
+      m_thd->mem_root= &m_local_mem_root;
+    }
+  }
+
+  ~Thd_local_mem_root()
+  {
+    if (!m_restored)
+      restore();
+  }
+
+  void restore()
+  {
+    if (m_thd == current_thd)
+    {
+      DBUG_ASSERT(m_thd->mem_root == &m_local_mem_root);
+      m_thd->mem_root= m_old_mem_root;
+      free_root(&m_local_mem_root, MYF(0));
+      m_restored= true;
+    }
+  }
+
+};
+
 Rpl_info_table::Rpl_info_table(uint nparam,
                                const char* param_schema,
                                const char *param_table,
@@ -95,6 +139,7 @@ int Rpl_info_table::do_init_info(enum_find_method method, uint instance)
   DBUG_ENTER("Rlp_info_table::do_init_info");
 
   THD *thd= access->create_thd();
+  Thd_local_mem_root local_mem_root(thd);
 
   saved_mode= thd->variables.sql_mode;
   tmp_disable_binlog(thd);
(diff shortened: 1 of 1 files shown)


sql/rpl_info_table.cc
```diff
@@ -21,6 +21,46 @@
 #include "rpl_info_values.h"        // Rpl_info_values
```
laurynas-biveinis: Would it make sense to add that MTR testcase anyway?

sql/rpl_info_table.cc
```diff
@@ -21,6 +21,46 @@
...
 #include "sql_class.h"              // THD
 
+class Thd_local_mem_root
```
percona-ysorokin: Make this class non-copyable (add private undefined copy constructor and assignment operator).

sql/rpl_info_table.cc
vlad-lesin: I am not sure, Laurynas, because that test just measures the amount of allocated memory before some action and after it. And if the difference between two measures is big enough then the test failed. But the question is how to understand that the difference is big enough? Where is the guarantee that delta is abnormal?

sql/rpl_info_table.cc
vlad-lesin: Fixed.

sql/rpl_info_table.cc
laurynas-biveinis: I'm not sure neither: this testcase would not test the memory leak bug exactly as you explain - it would just simply provide more coverage for the code paths executed and would serve as a correctness regresssion testcase.

vlad-lesin: But how this test will fail in the case of regression? There are no strict conditions of failure. Ok, we will have a little bit better code coverage, but coverage itself is useless until we have criteria of failure or correctness, what is not true for this test case.

PR #1783, opened 2017-06-05
Bug lp:1667307 fix.

vlad-lesin:
Bug description:

1) When binlog_rows_query_log_events is enabled, Rows_query_log_event
is created, written to binary log on master prior Rows_log_event.

2) Slave is reading this event from relay log and applying it,
see Rows_query_log_event::do_apply_event(). When the event is applied
the event itself is stored in Relay_log_info::rows_query_ev:
```c++
int Rows_query_log_event::do_apply_event(Relay_log_info const *rli)
{
  DBUG_ENTER("Rows_query_log_event::do_apply_event");
  DBUG_ASSERT(rli->info_thd == thd);
  /* Set query for writing Rows_query log event into binlog later.*/
  thd->set_query(m_rows_query, strlen(m_rows_query));

  DBUG_ASSERT(rli->rows_query_ev == NULL);

  const_cast<Relay_log_info*>(rli)->rows_query_ev= this;
  /* Tell worker not to free the event */
  worker= NULL;
  DBUG_RETURN(0);
}
```

3) Usually Relay_log_info::rows_query_ev is removed after Rows_log_event
execution with the following stack trace:

```
#0 Rows_query_log_event::~Rows_query_log_event (this=0x7fff9401df00, __in_chrg=<optimized out>, __vtt_parm=<optimized out>)
    at ./sql/log_event.h:3795
#1 0x0000000001958d1e in Relay_log_info::cleanup_context (this=0x7fffa40413d0, thd=0x7fff94000950, error=false)
    at ./sql/rpl_rli.cc:1777
#2 0x00000000018d984b in rows_event_stmt_cleanup (rli=0x7fffa40413d0, thd=0x7fff94000950) at ./sql/log_event.cc:11297
#3 0x00000000018d9500 in Rows_log_event::do_apply_event (this=0x7fff94012fc0, rli=0x7fffa40413d0)
    at ./sql/log_event.cc:11175
#4 0x00000000018c00c7 in Log_event::apply_event (this=0x7fff94012fc0, rli=0x7fffa40413d0) at ./sql/log_event.cc:3371
#5 0x0000000001938e0f in apply_event_and_update_pos (ptr_ev=0x7ffff139e8a0, thd=0x7fff94000950, rli=0x7fffa40413d0)
    at ./sql/rpl_slave.cc:4750
#6 0x000000000193a5d7 in exec_relay_log_event (thd=0x7fff94000950, rli=0x7fffa40413d0) at ./sql/rpl_slave.cc:5265
#7 0x00000000019418eb in handle_slave_sql (arg=0x7fffa4139800) at ./sql/rpl_slave.cc:7461
#8 0x0000000001e6a679 in pfs_spawn_thread (arg=0x7fffa413d640) at ./storage/perfschema/pfs.cc:2188
#9 0x00007ffff71616ba in start_thread (arg=0x7ffff139f700) at pthread_create.c:333
#10 0x00007ffff65f682d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```

But if we look into Rows_log_event::do_apply_event() the function
rows_event_stmt_cleanup() is called at the end:

```c++
int Rows_log_event::do_apply_event(Relay_log_info const *rli)
{
...
end:
  if (get_flags(STMT_END_F))
  {
    if((error= rows_event_stmt_cleanup(rli, thd)))
    {
      ...
    }
    ...
  }
...
}
```

4) But when the statement must be skipped because it has been already
applied from another channel, Rows_log_event::do_apply_event() exits before
rows_event_stmt_cleanup() call. See the following code:

```c++
int Rows_log_event::do_apply_event(Relay_log_info const *rli)
{
...
    enum_gtid_statement_status state= gtid_pre_statement_checks(thd);
    if (state == GTID_STATEMENT_EXECUTE)
    {
      if (gtid_pre_statement_post_implicit_commit_checks(thd))
        state= GTID_STATEMENT_CANCEL;
    }

    if (state == GTID_STATEMENT_CANCEL)
    {
      uint error= thd->get_stmt_da()->mysql_errno();
      DBUG_ASSERT(error != 0);
      rli->report(ERROR_LEVEL, error,
                  "Error executing row event: '%s'",
                  thd->get_stmt_da()->message_text());
      thd->is_slave_error= 1;
      DBUG_RETURN(-1);
    }
    else if (state == GTID_STATEMENT_SKIP)
      DBUG_RETURN(0);
...
end:
  if (get_flags(STMT_END_F))
  {
    if((error= rows_event_stmt_cleanup(rli, thd)))
    {
      ...
    }
    ...
  }
...
}
```

So in this case rli->rows_query_ev will not be deleted. In debug build such
case will lead to crash due to the corresponding assert in
Rows_query_log_event::do_apply_event():

```c++
int Rows_query_log_event::do_apply_event(Relay_log_info const *rli)
{
...
  DBUG_ASSERT(rli->rows_query_ev == NULL);
...
}
```

But for release build there will be just memory leak.

How to fix:

delete Rows_query_log_event instance if the next Rows_log_event instance read
from relay log must be skipped.

Testing: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/983/
rpl_multi_source_slave_files looks very suspicious but it fails without my fix, and here is the bug report: https://bugs.mysql.com/bug.php?id=79402
See also: https://bugs.launchpad.net/percona-server/+bug/1667307

### DIFF
--- a/sql/log_event.cc
+++ b/sql/log_event.cc
@@ -10723,7 +10723,21 @@ int Rows_log_event::do_apply_event(Relay_log_info const *rli)
       DBUG_RETURN(-1);
     }
     else if (state == GTID_STATEMENT_SKIP)
+    {
+      if (rli->rows_query_ev)
+      {
+        /*
+          thd->m_query_string now points to the data from
+          rli->rows_query_ev->m_rows_query
+          (see  Rows_query_log_event::do_apply_event()), don't let it point
+          to unallocated memory, reset query string first
+        */
+        thd->reset_query();
+        delete rli->rows_query_ev;
+        const_cast<Relay_log_info*>(rli)->rows_query_ev= NULL;
+      }
       DBUG_RETURN(0);
+    }
 
     /*
       The current statement is just about to begin and 
--- a/mysql-test/suite/rpl/r/bug85371.result
+++ b/mysql-test/suite/rpl/r/bug85371.result
@@ -1,0 +1,31 @@
+include/rpl_init.inc [topology=1->3,3->2,1->2]
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
+Note	####	Storing MySQL user name or password information in the master info repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START SLAVE; see the 'START SLAVE Syntax' in the MySQL Manual for more information.
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
+Note	####	Storing MySQL user name or password information in the master info repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START SLAVE; see the 'START SLAVE Syntax' in the MySQL Manual for more information.
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
+Note	####	Storing MySQL user name or password information in the master info repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START SLAVE; see the 'START SLAVE Syntax' in the MySQL Manual for more information.
+### Create some binlog events on server_1
+CREATE TABLE t1(a INT);
+INSERT INTO t1 (a) VALUES (1), (2), (3), (4), (5);
+INSERT INTO t1 (a) VALUES (6), (7), (8), (9), (10);
+DROP TABLE t1;
+### sync the server_3 with server_1
+[connection server_1]
+include/sync_slave_sql_with_master.inc
+### sync the server_2 with server_3
+[connection server_3]
+include/sync_slave_sql_with_master.inc
+### sync the server_2 with server_1
+### here crash must take place if the bug is not fixed
+[connection server_1]
+include/sync_slave_sql_with_master.inc
+### finish replication
+[connection server_1]
+include/rpl_end.inc
+RESET SLAVE ALL FOR CHANNEL  'channel_1';
+RESET SLAVE ALL FOR CHANNEL  'channel_3';
+RESET SLAVE ALL FOR CHANNEL  'channel_1';
--- a/mysql-test/suite/rpl/t/bug85371.cnf
+++ b/mysql-test/suite/rpl/t/bug85371.cnf
@@ -1,0 +1,30 @@
+!include ../my.cnf
+
+[mysqld.1]
+enforce-gtid-consistency=ON
+gtid-mode=ON
+master-info-repository=TABLE
+relay-log-info-repository=TABLE
+binlog-rows-query-log-events= ON
+log-slave-updates
+
+[mysqld.2]
+enforce-gtid-consistency=ON
+gtid-mode=ON
+master-info-repository=TABLE
+relay-log-info-repository=TABLE
+binlog-rows-query-log-events= ON
+log-slave-updates
+
+[mysqld.3]
+enforce-gtid-consistency=ON
+gtid-mode=ON
+master-info-repository=TABLE
+relay-log-info-repository=TABLE
+binlog-rows-query-log-events= ON
+log-slave-updates
+
+[ENV]
+SERVER_MYPORT_3=		@mysqld.3.port
+SERVER_MYSOCK_3=		@mysqld.3.socket
+
--- a/mysql-test/suite/rpl/t/bug85371.test
+++ b/mysql-test/suite/rpl/t/bug85371.test
@@ -1,0 +1,51 @@
+#
+# In the case if the bug is not fixed the test will crash on assert
+# in Rows_query_log_event::do_apply_event() on debug build.
+#
+--source include/have_binlog_format_row.inc
+--source include/have_gtid.inc
+--source include/not_group_replication_plugin.inc
+
+--let $rpl_topology= 1->3,3->2,1->2
+--let $rpl_multi_source= 1
+--let $use_gtids= 1
+--source include/rpl_init.inc
+
+--echo ### Create some binlog events on server_1
+--connection server_1
+CREATE TABLE t1(a INT);
+INSERT INTO t1 (a) VALUES (1), (2), (3), (4), (5);
+INSERT INTO t1 (a) VALUES (6), (7), (8), (9), (10);
+DROP TABLE t1;
+
+--echo ### sync the server_3 with server_1
+--let $rpl_connection_name= server_1
+--source include/rpl_connection.inc
+
+--let $rpl_channel_name= channel_1
+--let $sync_slave_connection= server_3
+--source include/sync_slave_sql_with_master.inc
+
+--echo ### sync the server_2 with server_3
+--let $rpl_connection_name= server_3
+--source include/rpl_connection.inc
+
+--let $rpl_channel_name= channel_3
+--let $sync_slave_connection=server_2
+--source include/sync_slave_sql_with_master.inc
+
+--echo ### sync the server_2 with server_1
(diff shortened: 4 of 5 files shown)


mysql-test/suite/rpl/r/rpl_bug-lp-1667307.result
```diff
@@ -0,0 +1,31 @@
+include/rpl_init.inc [topology=1->3,3->2,1->2]
```
laurynas-biveinis: Since this is an upstream bug, use upstream bug number rpl.bug85371 for testcase naming (for easier upstream submission - see Confluence)

laurynas-biveinis (changes requested):

mysql-test/suite/rpl/t/rpl_bug-lp-1667307.test
```diff
@@ -0,0 +1,51 @@
+#
+# In the case if the bug is not fixed the test will crash on assert
+# in Rows_query_log_event::do_apply_event().
+#
```
percona-ysorokin: Also `--source include/not_group_replication_plugin.inc`

mysql-test/suite/rpl/t/rpl_bug-lp-1667307.test
```diff
@@ -0,0 +1,51 @@
...
+--source include/have_debug.inc
```
percona-ysorokin: Why do we need `have_debug.inc`?

mysql-test/suite/rpl/t/rpl_bug-lp-1667307.cnf
```diff
@@ -0,0 +1,30 @@
+!include ../my.cnf
+
+[mysqld.1]
+enforce-gtid-consistency=ON
+gtid-mode=ON
+master-info-repository=TABLE
```
percona-ysorokin: Strictly speaking, we don't need
`master-info-repository=TABLE`
`relay-log-info-repository=TABLE`
for the first server

mysql-test/suite/rpl/t/rpl_bug-lp-1667307.test
```diff
@@ -0,0 +1,51 @@
+#
```
percona-ysorokin: test case naming as per Laurynas' comments

sql/log_event.cc
```diff
@@ -10723,7 +10723,21 @@ int Rows_log_event::do_apply_event(Relay_log_info const *rli)
       DBUG_RETURN(-1);
     }
     else if (state == GTID_STATEMENT_SKIP)
+    {
```
percona-ysorokin, line 10726: Do we need similar cleanup logic for the previous branch
`if (state == GTID_STATEMENT_CANCEL)`?

percona-ysorokin (changes requested):

mysql-test/suite/rpl/r/rpl_bug-lp-1667307.result
vlad-lesin: Fixed.

mysql-test/suite/rpl/t/rpl_bug-lp-1667307.test
vlad-lesin: Fixed.

mysql-test/suite/rpl/t/rpl_bug-lp-1667307.test
vlad-lesin: Fixed.

mysql-test/suite/rpl/t/rpl_bug-lp-1667307.test
vlad-lesin: Yes, we need it, because the "assert", on which the test crashes, if the bug is not fixed, happens only on debug build as "DBUG_ASSERT" macros is used. See the following code:

```c++
int Rows_query_log_event::do_apply_event(Relay_log_info const *rli)
{
  ...
  DBUG_ASSERT(rli->rows_query_ev == NULL);
...
}
```

mysql-test/suite/rpl/t/rpl_bug-lp-1667307.cnf
vlad-lesin: No, we need it, see topology in the test source code:

```
--let $rpl_topology= 1->3,3->2,1->2
--let $rpl_multi_source= 1
--let $use_gtids= 1
--source include/rpl_init.inc
```

For such topology we need multi-source replication, what requires master and relay log info format to be "TABLE". I am not sure if it is possible to have some servers with "multisource" option on, and some of them with the option switched off in replication graph, but, anyway, "rpl_init.inc" does not allow this, if "$rpl_multi_source" is on, then it's on for all servers pointed in the graph.

sql/log_event.cc
vlad-lesin: I am not sure about this. This question requires additional research and there indeed must be other test to model different situations(I counted at least four such cases) when gtid_pre_statement_checks() returns GTID_STATEMENT_CANCEL and how higher-level functions process the case when Rows_log_event::do_apply_event() returns error. I think such cases must be considered separately from the original case, and there must be separate bug report.

mysql-test/suite/rpl/t/rpl_bug-lp-1667307.test
laurynas-biveinis: I see no reason not to run this on release build as well, this a regression test case after all

percona-ysorokin:
Looks good with the only comment that I also think that this test case must be run not only in debug mode (even though the initial problem was caused by `DBUG_ASSERT`, it won't hurt to enable this test case in release mode anyway).

vlad-lesin:
Yura, including this test in release testing does not have any sense because for release build it will pass anyway, regardless the fact of regression.

And the initial problem was caused by memory leak, not by DBUG_ASSERT.

mysql-test/suite/rpl/t/rpl_bug-lp-1667307.test
vlad-lesin: Laurynas, I just want to clarify about regression testing. My understanding of regression supposes that there are the criteria of regression. And this is not only my understanding. https://en.wikipedia.org/wiki/Regression_testing : "Regression testing is a type of software testing which verifies that software which was previously developed and tested still performs correctly after it was changed or interfaced with other software." What means the criteria of correctness must exist. In our case, such criteria exist only in debug build. Even if something is broken in release build we will not see this in the current version of the test. In this test, it is supposed that some code (assertion) must be executed to specify such criteria, what happens only in debug build. So what the sense of this test in a release build if there is no way to catch regression for release build?

laurynas-biveinis: For the release build the criteria would be admittedly weaker "runs, does not crash, completes correctly". According to your criteria the testcase should contain e.g. --source include/not_valgrind.inc, not_asan.inc at least, as it only trips on DBUG_ASSERT, shouldn't it? Having said that, if you considered everything above and still think this must be a debug-only testcase, then I'm OK with your decision. (I don't know about other reviewers of course)

vlad-lesin: Good point, Laurynas. valgrind should catch it. "have_debug.inc" has been removed.

vlad-lesin:
Guys, it seems there is no more dissidence for this pull request. I am waiting for your approvals to merge it.

percona-ysorokin (approved):
LGTM

laurynas-biveinis (approved):
My review is only partial

PR #1784, opened 2017-06-05
Add geoWithin feature using S2 Geometry library.

SunguckLee:
With S2 geometry, latitude and longitude value is converted and stored 1-dimensional data.
So this feature make you create compound index with general scalar field and GPS location field(This is great feature to boost categorized geometry search performance).
This feature is based on MySQL UDF and QueryRewrite Plugin.

e.g.) Search all restaurant within 2000 meters with center position (37.547196, 127.047331)
SELECT * FROM poi_table WHERE type='restaurant' AND S2WITHIN(s2cellid, 37.547196, 127.047331, 2000, 20);

This pull-request contains S2 geometry library (https://code.google.com/archive/p/s2-geometry-library/source) based on Apache License 2.0. I am not sure Apache License 2.0 is compatible with percona-server license. If not, I will remove this pull-request.


SunguckLee:
I have prepared simple document for this feature.

https://sungucklee.github.io/s2_geometry_search

PR #1792, opened 2017-06-14
lp:1698019 - build-ps: take only the last value from my_print_defaults

comel:
Take only the last value from my_print_defaults. This is the same as e.g. in mysql-systemd-start.


dutow:
Closing this pull request as the issue was already fixed elsewhere.

PR #1795, opened 2017-06-16
Fix bug 1385352 (SET STATEMENT does not work after SET GLOBAL / SHOW …

laurynas-biveinis:
…GLOBAL STATUS and affects the global value)

SET STATEMENT ... FOR ... variable assignments depended on
lex->option_type being set to OPT_SESSION for the correct operation
(setting the current session and not the global variables). However,
SET STATEMENT ... FOR ... parser never set it, resulting in the
previous statement values reused. If a previous statement was SET
GLOBAL or SHOW GLOBAL STATUS, lex->option_type was OPT_GLOBAL, making
SET STATEMENT to operate on global variables. Fix by setting
lex->option_type to OPT_SESSION in SET STATEMENT parser grammar rule.

http://jenkins.percona.com/job/percona-server-5.6-param/1936/

### DIFF
--- a/mysql-test/r/percona_statement_set.result
+++ b/mysql-test/r/percona_statement_set.result
@@ -934,3 +934,39 @@ TESTDIR
 SELECT @@innodb_tmpdir;
 @@innodb_tmpdir
 TMPDIR
+#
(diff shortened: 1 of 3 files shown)


percona-ysorokin (approved):
LGTM

gl-sergei (approved):
LGTM

PR #1798, opened 2017-06-19
TDB-23 : PerconaFT can assert when opening a dictionary with no usefu…

george-lorch:
…l information to error

TDB-24 : PerconaFT asserts for various reasons deserializing nodes with no useful error output.
- Advancing git submodule commit pointer for PerconaFT
- Cherry-pick merge to 5.7, no GCA


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1937/

PR #1801, opened 2017-06-21
MYR-133 : rocksdb.select failing

george-lorch:
- Test uses PROCEDURE ANALYZE which is deprecated as of MySQL 5.7.18, and is
  removed in MySQL 8.0. This results in a new warning:
  Warning  1681  'PROCEDURE ANALYSE' is deprecated and will be removed in a
  future release.
- Since this is still valid syntax in 5.7, just re-recorded the test to capture
  the warning and will need to remove the use of PROCEDURE ANALYZE in 8.0

### DIFF
--- a/mysql-test/suite/rocksdb/r/select.result
+++ b/mysql-test/suite/rocksdb/r/select.result
@@ -107,6 +107,8 @@ SUM(a)	MAX(a)	b
 602	200	NULL
 SELECT * FROM t2 WHERE a>0 PROCEDURE ANALYSE();
 Field_name	Min_value	Max_value	Min_length	Max_length	Empties_or_zeros	Nulls	Avg_value_or_avg_length	Std	Optimal_fieldtype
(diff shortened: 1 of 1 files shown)


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1021/

laurynas-biveinis (approved):
https://blueprints.launchpad.net/percona-server/+spec/initial-5.8-port could be the right place to make a note to remove PROCEDURE ANALYZE (although the 1st attempt to run the testcase will tell the same)

george-lorch:
Agreed, I was wondering myself if it was worth noting on the 5.8/8.0 port list for this or just wait and deal with it as it comes.

george-lorch:
I think I will also start labeling issues that need special 8.0 attention in JIRA with something like 8.0_consideration or something.

PR #1805, opened 2017-06-23
Fix bug 1121072 (Redundant DB_TABLESPACE_DELETED handling for crash r…

laurynas-biveinis:
…ecovery page reads)

XtraDB for a long time carried code in buf_read_page_low and
buf_read_recv_pages which attempts, if a page being read is not found
due to a missing tablespace, to remove all its recovery entries from
the recovery redo log hash table. This handling is redundant in the
server as the recovery code makes sure to add redo log entries to the
hash table only for the existing tablespaces, in 5.7 specifically this
is handled by WL#7142. Remove that code.

http://jenkins.percona.com/job/mysql-5.7-param/1024/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1809, opened 2017-06-29
PS-8129: bug in  threadpool_unix.cc  tp_set_threadpool_size() (5.7)

lipengchong:
if io_poll_create failed ,we shouldn't break.we should unlock the group mutex and return i. if  break here,we don't unlock the group mutex,and 
group_count= size.So it is a bug here.


dutow:
@lipengchong I rebased your work on our current 5.7 branch.

PR #1810, opened 2017-06-29
PMM-240 - fixing typo of trottling -> throttling

michaelcoburn:


laurynas-biveinis:
You did submit a PR, I only ensured that the fix starts at the lowest branch and is upmerged

PR #1811, opened 2017-07-05
MYR-137 : Review replication testing

george-lorch:
- Expanded rocksdb.rpl 'combinations' to test all binlog fomats and not just row
- Added --loose-rocksdb_unsafe_for_binlog=true to rocksdb.rpl/t/suite.opt to
  allow rocksdb tests to proceed with all binlog formats even though some
  combinations of MyRocks options can lead to unsafe/inconsistent replication.
- New tests implemented around extra/rpl_tests, recorded:
  - rpl_rocksdb_auto_increment
  - rpl_rocksdb_commit_after_flush
  - rpl_rocksdb_crash_safe (stm_mixed, row)  ** needed to set transaction_level=READ-COMMITTED in .opt files **
  - rpl_rocksdb_extra_col_master
  - rpl_rocksdb_extra_col_slave
  - rpl_rocksdb_insert_ignore ** needed to set transaction_level=READ-COMMITTED in .opt files **
  - rpl_rocksdb_log (stm, row)
  - rpl_rocksdb_lower_case_table_names (stm_mixed, row)
  - rpl_rocksdb_mixed_ddl ** needed to set rocksdb_strict_collation_check=off in .opt files **
  - rpl_rocksdb_multi_update (mixed, row) ** needed to set transaction_level=READ-COMMITTED in .opt files, stm will not work with this ISO forced **
  - rpl_rocksdb_multi_update2 (mixed, row) ** needed to set transaction_level=READ-COMMITTED in .opt files, stm will not work with this ISO forced **
  - rpl_rocksdb_multi_update3 (mixed, row) ** needed to set transaction_level=READ-COMMITTED in .opt files, stm will not work with this ISO forced **
  - rpl_rocksdb_not_null
  - rpl_rocksdb_parallel_ddl
  - rpl_rocksdb_sv_relay_space
  - rpl_rocksdb_record_compare
  - rpl_rocksdb_row_basic ** needed to set rocksdb_strict_collation_check=off and transaction_level=READ-COMMITTED in .opt files **
  - rpl_rocksdb_row_blob
  - rpl_rocksdb_row_sp003
  - rpl_rocksdb_row_sp007
  - rpl_rocksdb_row_tabledefs
  - rpl_rocksdb_row_trig004
  - rpl_rocksdb_set_null
  - rpl_rocksdb_truncate


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1029/

mysql-test/suite/rocksdb.rpl/t/rpl_rocksdb_row_basic.test
```diff
@@ -0,0 +1,11 @@
+-- source include/have_rocksdb.inc
+-- source include/have_binlog_format_row.inc
+-- source include/master-slave.inc
+
+connection slave;
+let $bit_field_special = ALL_LOSSY;
+let $type= 'RocksDB' ;
```
laurynas-biveinis: No space before semicolon

mysql-test/suite/rocksdb.rpl/t/rpl_rocksdb_stm_mixed_crash_safe.test
```diff
@@ -0,0 +1,18 @@
+--source include/big_test.inc
+--source include/master-slave.inc
```
laurynas-biveinis: master-slave.inc should go last in the header

mysql-test/suite/rocksdb.rpl/t/suite.opt
```diff
@@ -1 +1 @@
-$ROCKSDB_OPT $ROCKSDB_LOAD_ADD
+$ROCKSDB_OPT $ROCKSDB_LOAD_ADD --loose-rocksdb_unsafe_for_binlog=true
```
laurynas-biveinis: That doesn't sound safe. Perhaps some of the existing tests need stricter --source include/have_binlog_format_*.inc instead?

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb.rpl/t/suite.opt
george-lorch: This is a bit complicated. Facebook really only wants to support ROW with MyRocks but AFAIK there is no practical way to enforce this and IMHO it is just not right. MyRocks should be tested as much as possible with all formats. This option disables some of their (incomplete) checks for things that would be unsafe if mixed with other options like read free, etc... that are not validated in combination with this option check. So without this option, most of these newly added tests would only work for ROW.

mysql-test/suite/rocksdb.rpl/t/rpl_rocksdb_stm_mixed_crash_safe.test
george-lorch: All of these were copied from elsewhere in the mtr suite and switched to RocksDB, I tried to catch and fixup some of this kind of stuff as I went along and obviously missed some in my haste :-\

george-lorch:
New jenkins http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1044/

mysql-test/suite/rocksdb.rpl/t/suite.opt
laurynas-biveinis: Then perhaps this option should be added to individual .opt files for the unsafe tests? I am worried that mixed/stmt testing is going to be sub-standard here even for the subset of the testcases that don't have any inherent row format dependencies.

If you considered the above and still want a global option here, then it's OK

laurynas-biveinis (approved):
The final mysql-test/suite/rocksdb.rpl/t/suite.opt decision up to you

mysql-test/suite/rocksdb.rpl/t/suite.opt
george-lorch: Well, my thinking here is a bit contrary to our normal approach. We generally design and implement tests to be as deterministic and 100% passable as possible. In this case, I am wanting to cast as wide a net as I can, intentionally looking for instabilities that will (hopefully) point to places in the code where there are not adequate checks for combinations of unsafe option settings and table key layouts. I suppose individual .opt files for tests that we know might become unstable as a result is not an unrealistic thing to ask for then take action if/when things show up.

laurynas-biveinis: OK

george-lorch: OK, so I reworked a few of these that need the rocksdb_unsafe_for_binlog=true which was mostly splitting a few of the tests into a row and a stm_mixed test, re-recorded and a new jenkins run here http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1052/

This looks pretty good as is so I will go ahead and merge so I can continue rebasing/fixing/merging forward.

PR #1815, opened 2017-07-06
MYR-28 : Support lz4 and zstd compression by static link

george-lorch:
- This commit implements lz4, lz4hc, and zstd compression by pulling in lz4
  code from  https://github.com/lz4/lz4, zstd code from
  https://github.com/facebook/zstd vie new git submodule pointers and
  compiling the needed source directly into MyRocks.
- Added simple tests for compression_lz4 and compression_lz4hc.
- Enabled and improved exising simple test for compression_zstd.


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1034/

laurynas-biveinis (changes requested):
The simple tests should insert something, flush memtable, and query - otherwise I believe the tests are no-ops

george-lorch:
I can add a little activity to the tests, but they are not totally no-ops as they stand. They test the setting and recognition of the compression algos. If the option values were bad or compression missing, MyRocks would fail to initialize and the tests would fail.

george-lorch:
OK, added a compression.inc that jams in a few rows, flushes the memtable and forces a compaction, then selects a SUM. Re-recorded the compression.

Started a small jenkins job here http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1043/

laurynas-biveinis (approved):

george-lorch:
OK, I rebased this on the 5.7 trunk after the Feb 2017 upstream merge and also added a little more validation of the table creation and column family settings and re-recorded. Also added missing zlib test. This all looks pretty good now I think. Thanks for the review!

PR #1821, opened 2017-07-12
Fix lp:1702903 (Cannot build Percona Server on Debian Stretch)

percona-ysorokin:
Percona Server code made compatible with OpenSSL 1.1.

As 'DH' became an internal-only struct, it is not possible to update
its members directly any more. Fixed by using 'DH_set0_pqg()'
instead.

All occurrences of 'ERR_remove_state()' are now wrapped into '#if'
directive to make sure that this function is called only with
OpenSSL < 1.1.

As Percona Server built with OpenSSL 1.1 reports 'ECDHE-RSA-...' for
'ssl_Cipher' status variable, whereas for OpenSSL < 1.1 these values
are 'DHE-RSA-...', a number of '--replace-result' directives in SSL
MTR test cases extended with new values.

Reworked and re-recorded 'main.openssl_1' MTR test case to detect
whether to use OpenSSL 1.1 'ECDHE-RSA-...' cipher values or not for
'GRANT' statements.


percona-ysorokin:
Jenkins build links:
http://jenkins.percona.com/view/PS%205.5/job/percona-server-5.5-param/1572 (OpenSSL)
http://jenkins.percona.com/view/PS%205.5/job/percona-server-5.5-param/1573 (YaSSL)

laurynas-biveinis (approved):
Is this based on one of the existing patches?

percona-ysorokin:
Yes, @laurynas-biveinis 
This is based on your patch
https://github.com/laurynas-biveinis/percona-server/commit/8a4295c2417e738d93a64771880e92994d02ef52
and Sergei's patch
https://github.com/percona/percona-xtrabackup/pull/360/commits/8853f2dd913c6a966a6ff0a7248130733c953968
Plus some `#ifdefs` to make sure that deprecated OpenSSL functions are not used (required for `-DMYSQL_MAINTAINER_MODE=ON`).
Regarding `main.openssl_1` MTR test case, rearranging those code fragments was a way to determine whether elliptic curves ciphers are available on 5.5 (unfortunately, the patch with `have_elliptic_curves` you did only for 5.6 and 5.7).

laurynas-biveinis:
Ack, any substantial differences from the patch on the upstream bug / MariaDB?

sql/slave.cc
```diff
@@ -4353,8 +4353,8 @@ static int connect_to_master(THD* thd, MYSQL* mysql, Master_info* mi,
   mysql_options(mysql, MYSQL_SET_CHARSET_DIR, (char *) charsets_dir);
 
   /* Set MYSQL_PLUGIN_DIR in case master asks for an external authentication plugin */
-  if (opt_plugin_dir_ptr && *opt_plugin_dir_ptr)
-    mysql_options(mysql, MYSQL_PLUGIN_DIR, opt_plugin_dir_ptr);
+  if (opt_plugin_dir_ptr && *opt_plugin_dir_ptr)
+    mysql_options(mysql, MYSQL_PLUGIN_DIR, opt_plugin_dir_ptr);
```
george-lorch, line 4357: These look identical, curious why git picks this out as a diff///

george-lorch (approved):

sql/slave.cc
percona-ysorokin: Could be inconsistent CRLF line endings which my IDE  automatically fixed.

PR #1825, opened 2017-07-13
MYR-146 : rocksdb.sys_var tests need have_myisam.inc

george-lorch:
- Added have_myisam.inc where needed and cleaned up some extra whitespace


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1053/

laurynas-biveinis (approved):

PR #1826, opened 2017-07-13
Fix bug 1669414 (Failed to set O_DIRECT on xb_doublewrite when runnin…

laurynas-biveinis:
…g MTR test cases)

Fix handling of failure to set O_DIRECT on the parallel doublewrite
buffer by the following:
- downgrade diagnostics severity from warning to note by introducing
  an extra flag to os_file_setnocache;
- introduce new flag parallel_dblwr_t::needs_flush. Initialize it in
  buf_parallel_dblwr_file_create according to innodb_flush_method
  value and whether setting O_DIRECT succeeded. Use it instead of
  innodb_flush_method value in buf_dblwr_flush_buffered_writes.
- fix an incorrect comment at os_file_create_func.

http://jenkins.percona.com/job/mysql-5.7-param/1054/

### DIFF
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -3706,9 +3706,8 @@ os_file_create_func(
 	} else if (!srv_read_only_mode
 		   && *success
 		   && srv_unix_file_flush_method == SRV_UNIX_ALL_O_DIRECT) {
-		/* Do fsync() on log and parallel doublewrite files
-		when setting O_DIRECT fails.
-		See log_io_complete() and buf_dblwr_flush_buffered_writes() */
+		/* Do fsync() on log and files when setting O_DIRECT fails.
+		See log_io_complete() */
 		if (!os_file_set_nocache(file.m_file, name, mode_str)) {
 			srv_unix_file_flush_method = SRV_UNIX_O_DIRECT;
 		}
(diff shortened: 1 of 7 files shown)


percona-ysorokin (approved):
LGTM

storage/innobase/os/os0file.cc
```diff
@@ -6053,11 +6071,12 @@ os_file_set_nocache(
...
 short_warning:
 # endif
-			ib::warn()
-				<< "Failed to set O_DIRECT on file "
+			ib::logger *logger = ib_warn_or_info(failure_warning);
+			*logger << "Failed to set O_DIRECT on file "
 				<< file_name << "; " << operation_name
 				<< " : " << strerror(errno_save)
 				<< " continuing anyway.";
+			delete logger;
```
gl-sergei: Actual write to log happens when destructor invoked. Since ib::logger does not have virtual destructor, I suspect this message will not be logged in any case. Maybe it would be better to implement it the same way `error_or_warn` is implemented - subclass `ib::logger` and define specific destructor.

gl-sergei (changes requested):

storage/innobase/os/os0file.cc
```diff
@@ -6037,14 +6052,17 @@ os_file_set_nocache(
...
+					= ib_warn_or_info(failure_warning);
+				*logger
+					<< "Failed to set O_DIRECT on file "
 					<< file_name << ";" << operation_name
 					<< ": " << strerror(errno_save) << ", "
 					<< "continuing anyway. O_DIRECT is "
 					"known to result in 'Invalid argument' "
 					"on Linux on tmpfs, "
 					"see MySQL Bug#26662.";
+				delete logger;
```
gl-sergei: same for this message

gl-sergei (changes requested):

storage/innobase/os/os0file.cc
laurynas-biveinis: Thanks, this is a great catch

laurynas-biveinis:
http://jenkins.percona.com/job/mysql-5.7-param/1067/

laurynas-biveinis:
@percona-ysorokin or @robgolebiowski, please re-review

percona-ysorokin (approved):
LGTM

PR #1827, opened 2017-07-13
Add testcase for bug 1704056 (Enabling and disabling slow query log r…

laurynas-biveinis:
…otation spuriously adds version suffix to the next slow query log file name)

5.5 is not affected by the bug, thus only add the testcase.

http://jenkins.percona.com/job/percona-server-5.5-param/1575/


percona-ysorokin (approved):
LGTM

gl-sergei (approved):
LGTM

PR #1830, opened 2017-07-14
Fix bug 1704267 (Spurious lock_wait_timeout_thread wakeup in lock_wai…

laurynas-biveinis:
…t_suspend_thread())

InnoDB spuriously wakes up lock_timeout thread on each lock wait
start, which is unnecessary and contributes to lock_sys_wait mutex
pressure if there are many lock waits ongoing. Also, in some cases
lock_timeout thread is woken up completely unnecessarily when the
intention is to wake up the InnoDB monitor thread instead.

Fix both of these issues by applying [1] from WebScaleSQL / MySQL
Facebook patch.

[1]:

commit 6e06bbfa315ffb97d713dd6e672d6054036ddc21
Author: Inaam Rana <irana@twitter.com>
Date:   Thu Apr 10 10:46:14 2014 -0400

    Spurious wakeup in lock_wait_suspend_thread()

    Summary:
    Feature: Performance improvement in locking subsystem

    This is a fix for upstream bug#72123

    lock_timeout thread works in a tight loop waking up every second
    and checking for lock_wait_timeout. In addition, when a mysql
    thread is forced to wait on a lock, it signals the lock_timeout thread
    as well. This call is not required. In a heavily contended workload
    each thread going to wait will signal the lock_timeout thread making
    it work all the time. As lock_timeout thread scans the array of
    waiting threads under lock_sys::wait_mutex which is already very
    hot in contneded loads, these extra scans can cause significanct
    performance regression.

    Also, in various codepaths lock_timeout thread is signalled where
    actual intention was to signal the innodb monitor thread.

    Test Plan: mtr

    Reviewers: pivanof, darnaut

    Reviewed By: darnaut

    CC: jtolmer, CalvinSun, WebScaleSQL, steaphan

    Differential Revision: https://reviews.facebook.net/D17673

http://jenkins.percona.com/job/mysql-5.7-param/1059/

### DIFF
--- a/storage/innobase/buf/buf0lru.cc
+++ b/storage/innobase/buf/buf0lru.cc
@@ -1334,7 +1334,7 @@ buf_LRU_check_size_of_non_data_objects(
 
 			buf_lru_switched_on_innodb_mon = true;
 			srv_print_innodb_monitor = TRUE;
-			os_event_set(lock_sys->timeout_event);
+			os_event_set(srv_monitor_event);
 		}
 
 	} else if (buf_lru_switched_on_innodb_mon) {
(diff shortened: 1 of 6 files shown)


percona-ysorokin (approved):
LGTM

gl-sergei (approved):
LGTM

PR #1844, opened 2017-07-25
Fix bug 1706262 (innodb.bug-76142 leaves $MYSQL_TMP_DIR/test behind)

laurynas-biveinis:
The testcase creates an InnoDB table with a specified data directory
$MYSQL_TMP_DIR, resulting in $MYSQL_TMP_DIR/test directory being
created, which is not removed at the end of the testcase. Remove it.

http://jenkins.percona.com/job/percona-server-5.6-param/1960/


percona-ysorokin (approved):
LGTM

PR #1849, opened 2017-07-26
[BLD-778] TokuDB upgrade fails on ubuntu/debian

EvgeniyPatlan:


tplavcic (commented):
This doesn't seem to check if LD_PRELOAD and THP_SETTING are already in /etc/default/mysql so on multiple upgrades it will add duplicate rows to the file?

delgod (approved):

EvgeniyPatlan:
good catch @tplavcic! thanks. fixed

tplavcic (approved):
LGTM

PR #1808, opened 2017-06-27
General tablespace encryption

gl-sergei:
https://blueprints.launchpad.net/percona-server/+spec/encrypted-general-tablespaces

This patch does following:

1. Adds syntax `CREATE TABLESPACE ... ENCRYPTED='y/n'', this option then
   passed to InnoDB and added to tablespace flags. Since InnoDB
   encryption is really tablespace-based, not table-based, no further
   actions needed to encrypt the tablespace.
2. Validates encryption options for `CREATE TABLE' and `ALTER TABLE'
   statements

PR #1858, opened 2017-08-01
InnoDB temporary tablespace encryption

gl-sergei:
Based on PR #1808
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1093/


laurynas-biveinis:
Is general tablespace encryption (https://github.com/percona/percona-server/pull/1808) a prerequisite? If so, I'll get that reviewed thoroughly first so it can be merged and this PR reduced

gl-sergei:
Rebased on 5.7 trunk http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1203/

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,93 @@
...
+# test that we can turn encryption OFF and ON
+
+SET GLOBAL innodb_temp_tablespace_encrypt = OFF;
+
+CREATE TEMPORARY TABLE t07 (a INT) ENGINE=InnoDB;
+INSERT INTO t07 VALUES (1), (2), (3);
+
+SET GLOBAL innodb_temp_tablespace_encrypt = ON;
+
+# make sure encryption cannot be be turned ON/OFF for session
```
laurynas-biveinis: "be be", although I'd just leave this part for a sys_vars test and remove it from here altogether

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,93 @@
...
+INSERT INTO t10 (b) SELECT b FROM t10;
+INSERT INTO t10 (b) SELECT b FROM t10;
+
+CREATE INDEX t10_b ON t10 (b);
+DROP INDEX t10_b ON t10;
+
+CREATE INDEX t10_b ON t10 (b) ALGORITHM=COPY;
+DROP INDEX t10_b ON t10 ALGORITHM=COPY;
+
+# make sure server survives restart
```
laurynas-biveinis: What does it mean for the "server to survive restart"?

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,93 @@
...
+let $restart_hide_args=1
+let $restart_parameters = restart: --early-plugin-load="keyring_file=$KEYRING_PLUGIN" --loose-keyring_file_data=$MYSQL_TMP_DIR/ts_encrypt_keyring --loose-innodb-temp-tablespace-encrypt $KEYRING_PLUGIN_OPT;
+--source include/kill_and_restart_mysqld.inc
+
+CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB;
+INSERT INTO t01 VALUES (1), (2), (3);
+
+# and restart without keyring plugin
+let $restart_hide_args=1
+let $restart_parameters = restart: ;
```
laurynas-biveinis: consistent spacing around = 
resetting restart_parameters to empty has the same effect as setting it to "restart:"

storage/innobase/fil/fil0fil.cc
```diff
@@ -5470,18 +5470,33 @@ fil_io_set_encryption(
...
-	from the system tablespace. */
-	if (!req_type.is_log() && page_id.page_no() > 0
-	    && space->encryption_type != Encryption::NONE)
-	{
-		req_type.encryption_key(space->encryption_key,
-					space->encryption_klen,
-					space->encryption_iv);
-		req_type.encryption_algorithm(Encryption::AES);
-	} else {
+	/* Don't encrypt page 0 of all tablespaces except redo log
```
laurynas-biveinis: "Don't encrypt page 0 of any tablespace except if it is the redo log tablespace" ?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21349,6 +21375,11 @@ static MYSQL_SYSVAR_STR(temp_data_file_path, innobase_temp_data_file_path,
   "Path to files and their sizes making temp-tablespace.",
   NULL, NULL, NULL);
 
+static MYSQL_SYSVAR_BOOL(temp_tablespace_encrypt, srv_tmp_tablespace_encrypt,
+  PLUGIN_VAR_OPCMDARG,
+  "Enable or disable Encrypt of temporary tablespace.",
```
laurynas-biveinis: s/Encrypt/encryption

storage/innobase/srv/srv0srv.cc
```diff
@@ -2668,6 +2671,55 @@ srv_master_do_shutdown_tasks(
...
+srv_enable_temp_encryption_if_set()
+{
+	fil_space_t*	space;
+
+	if (srv_shutdown_state != SRV_SHUTDOWN_NONE) {
+		return;
+	}
+
+	if (is_system_tablespace(srv_tmp_space.space_id())) {
+		/* there is no separate tablespace for temorary tables */
```
laurynas-biveinis: "temporary"

laurynas-biveinis (changes requested):
Minor comments. This will also need a review by other encryption developers

storage/innobase/fil/fil0fil.cc
```diff
@@ -7353,9 +7369,9 @@ fil_set_encryption(
 	byte*			key,
 	byte*			iv)
 {
-	ut_ad(!is_system_or_undo_tablespace(space_id));
+	ut_ad(!is_shared_system_tablespace(space_id));
```
percona-ysorokin: Both assertion and subsequent `if(...) return` with the same condition

percona-ysorokin (changes requested):

storage/innobase/srv/srv0srv.cc
```diff
@@ -2749,6 +2801,9 @@ DECLARE_THREAD(srv_master_thread)(
 		}
 	}
 
+	/* Enable temorary tablespace encryption if set */
```
gl-sergei: s/temorary/temporary/

mysql-test/suite/innodb/t/temp_table_encrypt.test
gl-sergei: removed

mysql-test/suite/innodb/t/temp_table_encrypt.test
gl-sergei: changed comment to "test that server comes booting up successfully after shutdown"

mysql-test/suite/innodb/t/temp_table_encrypt.test
gl-sergei: fixed

storage/innobase/fil/fil0fil.cc
gl-sergei: changed to "Don't encrypt page 0 of all tablespaces except, don't encrypt redo log and system tablespace."

storage/innobase/handler/ha_innodb.cc
gl-sergei: fixed

storage/innobase/srv/srv0srv.cc
gl-sergei: fixed

storage/innobase/fil/fil0fil.cc
gl-sergei: Yep. And it was so in upstream actually. It means in debug build server will crash and in release build it will refuse to create table.

storage/innobase/srv/srv0srv.cc
gl-sergei: fixed

gl-sergei:
All comments fixed and replied. Param build started, will update with the link once it completes.

laurynas-biveinis:
@gl-sergei, has the last Jenkins run completed?

gl-sergei:
@laurynas-biveinis sorry, forgot to paste the link. http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1212/

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,87 @@
...
+INSERT INTO t10 (b) SELECT b FROM t10;
+INSERT INTO t10 (b) SELECT b FROM t10;
+
+CREATE INDEX t10_b ON t10 (b);
+DROP INDEX t10_b ON t10;
+
+CREATE INDEX t10_b ON t10 (b) ALGORITHM=COPY;
+DROP INDEX t10_b ON t10 ALGORITHM=COPY;
+
+# test that server comes booting up successfully after shutdown
```
laurynas-biveinis: "test restarting" would say the same?

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,87 @@
...
+let $restart_hide_args= 1
+let $restart_parameters= restart: --early-plugin-load="keyring_file=$KEYRING_PLUGIN" --loose-keyring_file_data=$MYSQL_TMP_DIR/ts_encrypt_keyring --loose-innodb-temp-tablespace-encrypt $KEYRING_PLUGIN_OPT;
+--source include/restart_mysqld.inc
+
+CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB;
+INSERT INTO t01 VALUES (1), (2), (3);
+
+# and crash
```
laurynas-biveinis: Likewise "test crashing"

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,87 @@
...
+INSERT INTO t10 (b) SELECT b FROM t10;
+
+CREATE INDEX t10_b ON t10 (b);
+DROP INDEX t10_b ON t10;
+
+CREATE INDEX t10_b ON t10 (b) ALGORITHM=COPY;
+DROP INDEX t10_b ON t10 ALGORITHM=COPY;
+
+# test that server comes booting up successfully after shutdown
+let $restart_hide_args= 1
```
laurynas-biveinis: No ";" at the end of line and no "--" at the start? Shouldn't this be a parse error?

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,87 @@
...
+let $restart_parameters= restart: --early-plugin-load="keyring_file=$KEYRING_PLUGIN" --loose-keyring_file_data=$MYSQL_TMP_DIR/ts_encrypt_keyring --loose-innodb-temp-tablespace-encrypt $KEYRING_PLUGIN_OPT;
+--source include/restart_mysqld.inc
+
+CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB;
+INSERT INTO t01 VALUES (1), (2), (3);
+
+# and crash
+let $restart_hide_args=1
```
laurynas-biveinis: No ";" at the end of line and no "--" at the start? Shouldn't this be a parse error?

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,87 @@
...
+let $restart_parameters= restart: --early-plugin-load="keyring_file=$KEYRING_PLUGIN" --loose-keyring_file_data=$MYSQL_TMP_DIR/ts_encrypt_keyring --loose-innodb-temp-tablespace-encrypt $KEYRING_PLUGIN_OPT;
+--source include/kill_and_restart_mysqld.inc
+
+CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB;
+INSERT INTO t01 VALUES (1), (2), (3);
+
+# and able to startup without keyring plugin
+let $restart_hide_args=1
```
laurynas-biveinis: Likewise

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,87 @@
...
+let $restart_parameters=;
+--source include/restart_mysqld.inc
+
+CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB;
+INSERT INTO t01 VALUES (1), (2), (3);
+
+let $restart_hide_args=1
```
laurynas-biveinis: Likewise

mysql-test/suite/sys_vars/t/innodb_temp_tablespace_encrypt_basic.test
```diff
@@ -0,0 +1,64 @@
...
+--disable_warnings
+select * from performance_schema.global_variables where variable_name='innodb_temp_tablespace_encrypt';
+select * from performance_schema.session_variables where variable_name='innodb_temp_tablespace_encrypt';
+--enable_warnings
+
+#
+# show that it's writable
+#
+set global innodb_temp_tablespace_encrypt=1;
+--sleep 2
```
laurynas-biveinis: This is unstable, same issue and fix as https://bugs.mysql.com/bug.php?id=87303

storage/innobase/fil/fil0fil.cc
```diff
@@ -5470,18 +5470,33 @@ fil_io_set_encryption(
...
-	from the system tablespace. */
-	if (!req_type.is_log() && page_id.page_no() > 0
-	    && space->encryption_type != Encryption::NONE)
-	{
-		req_type.encryption_key(space->encryption_key,
-					space->encryption_klen,
-					space->encryption_iv);
-		req_type.encryption_algorithm(Encryption::AES);
-	} else {
+	/* Don't encrypt page 0 of all tablespaces except, don't encrypt
```
laurynas-biveinis: Don't encrypt redo log and system tablespaces, for all the other types tablespaces, don't encrypt page 0
?

storage/innobase/fil/fil0fil.cc
```diff
@@ -5470,18 +5470,33 @@ fil_io_set_encryption(
...
+	redo log and system tablespace. */
+	if (space->encryption_type == Encryption::NONE
+	    || (page_id.page_no() == 0 && !req_type.is_log())) {
+		req_type.clear_encrypted();
+		return;
+	}
+
+	/* For writting redo log, if encryption for redo log is disabled,
```
laurynas-biveinis: writing

storage/innobase/fil/fil0fil.cc
```diff
@@ -5470,18 +5470,33 @@ fil_io_set_encryption(
...
+	skip set encryption. */
+	if (req_type.is_log() && req_type.is_write()) {
 		req_type.clear_encrypted();
+		return;
 	}
+
+	/* For writting temporary tablespace, if encryption for temporary
```
laurynas-biveinis: writing

storage/innobase/fil/fil0fil.cc
```diff
@@ -5470,18 +5470,33 @@ fil_io_set_encryption(
...
+	/* Don't encrypt page 0 of all tablespaces except, don't encrypt
+	redo log and system tablespace. */
+	if (space->encryption_type == Encryption::NONE
+	    || (page_id.page_no() == 0 && !req_type.is_log())) {
+		req_type.clear_encrypted();
+		return;
+	}
+
+	/* For writting redo log, if encryption for redo log is disabled,
+	skip set encryption. */
```
laurynas-biveinis: setting

storage/innobase/fil/fil0fil.cc
```diff
@@ -5470,18 +5470,33 @@ fil_io_set_encryption(
...
+	if (req_type.is_log() && req_type.is_write()) {
 		req_type.clear_encrypted();
+		return;
 	}
+
+	/* For writting temporary tablespace, if encryption for temporary
+	tablespace is disabled, skip set encryption. */
```
laurynas-biveinis: setting

storage/innobase/handler/ha_innodb.cc
```diff
@@ -10977,18 +10977,44 @@ create_table_info_t::create_table_def()
...
+							  DICT_TF2_ENCRYPTION);
+				}
+			}
+
+			if (err == DB_SUCCESS) {
 
-			/* Get a new table ID */
-			dict_table_assign_new_id(table, m_trx);
+				/* Get a new table ID */
+				dict_table_assign_new_id(table, m_trx);
```
laurynas-biveinis, line 11056: I wonder why github shows the green bars here but not below. Tabs vs spaces?

storage/innobase/include/fsp0sysspace.h
```diff
@@ -282,6 +282,17 @@ extern SysTablespace srv_sys_space;
...
 
+/** Check if the space_id is for a shared system tablespace.
+@param[in]	id	Space ID to check
+@return true if id is a system tablespace, false if not. */
+UNIV_INLINE
+bool
+is_shared_system_tablespace(
+	ulint	id)
+{
+	return(id == srv_sys_space.space_id());
```
laurynas-biveinis: In theory the implementation should go fsp0sysspace.ic file, but I don't think we care about UNIV_NONINL (?) build

storage/innobase/include/os0file.h
```diff
@@ -369,6 +369,12 @@ static const ulint ENCRYPTION_INFO_SIZE_V2 = (ENCRYPTION_MAGIC_SIZE \
 					 + ENCRYPTION_SERVER_UUID_LEN \
 					 + 2 * sizeof(ulint));
 
+/** Default master key for bootstrap */
+static const char ENCRYPTION_DEFAULT_MASTER_KEY[] = "DefaultMasterKey";
```
laurynas-biveinis: This one is never used

storage/innobase/include/os0file.h
```diff
@@ -369,6 +369,12 @@ static const ulint ENCRYPTION_INFO_SIZE_V2 = (ENCRYPTION_MAGIC_SIZE \
...
+
+/** Default master key id for bootstrap */
+static const ulint ENCRYPTION_DEFAULT_MASTER_KEY_ID = 0;
```
laurynas-biveinis: This one is used in a single source file. Move there? Easy to change later as needed.

storage/innobase/srv/srv0srv.cc
```diff
@@ -2668,6 +2671,55 @@ srv_master_do_shutdown_tasks(
 	return(n_bytes_merged || n_tables_to_drop);
 }
 
+void
+srv_enable_temp_encryption_if_set()
+{
+	fil_space_t*	space;
```
laurynas-biveinis: Move to initialization

storage/innobase/srv/srv0srv.cc
```diff
@@ -2668,6 +2671,55 @@ srv_master_do_shutdown_tasks(
...
+	if (srv_shutdown_state != SRV_SHUTDOWN_NONE) {
+		return;
+	}
+
+	if (is_system_tablespace(srv_tmp_space.space_id())) {
+		/* there is no separate tablespace for temporary tables */
+		return;
+	}
+
+	space = fil_space_get(srv_tmp_space.space_id());
```
laurynas-biveinis: Move to if

storage/innobase/srv/srv0srv.cc
```diff
@@ -2668,6 +2671,55 @@ srv_master_do_shutdown_tasks(
...
+			mtr.set_named_space(space->id);
+
+			space = mtr_x_lock_space(space->id, &mtr);
+
+			if (!fsp_header_rotate_encryption(space,
+							  encrypt_info,
+							  &mtr)) {
+				mtr_commit(&mtr);
+
+				ib::error() << "Can't set temporary"
```
laurynas-biveinis: Setting the variable and making the change are far from each other, how to communicate this to user? Who would not necessarily check the error log. Is ib::fatal() an option here?

storage/innobase/srv/srv0start.cc
```diff
@@ -1231,6 +1231,20 @@ srv_open_tmp_tablespace(
...
+				fil_space_t*		space;
+
+				space = fil_space_get(temp_space_id);
+				space->flags |= FSP_FLAGS_MASK_ENCRYPTION;
+
+				err = fil_set_encryption(space->id,
+							 Encryption::AES,
+							 NULL,
+							 NULL);
+				ut_ad(err == DB_SUCCESS);
```
laurynas-biveinis: If we are asserting here that fil_set_encryption is "impossible" (so that crashing a debug build instead of printing errors, aborting server startup is reasonable) then what about release build? Maybe ut_a here?

laurynas-biveinis (changes requested):

mysql-test/suite/innodb/t/temp_table_encrypt.test
gl-sergei: strangely, it is not.. there is one more test `keyring_udf.test` with the same issue... I'll fix it. Actually not the only one... plenty of them ....

storage/innobase/handler/ha_innodb.cc
gl-sergei: No, just one more level of indentation added

storage/innobase/include/fsp0sysspace.h
gl-sergei: upstream `is_system_tablespace` and `is_system_or_undo_tablespace` are both in .h file

storage/innobase/srv/srv0srv.cc
gl-sergei: agree, ib::fatal() makes sense to me

storage/innobase/srv/srv0start.cc
gl-sergei: agree on that too

storage/innobase/fil/fil0fil.cc
gl-sergei: I copied  this spelling from MySQL 8 I guess... Fixing

gl-sergei:
PR updated. Jenkins build: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1325/

storage/innobase/srv/srv0srv.cc
```diff
@@ -2756,6 +2807,9 @@ DECLARE_THREAD(srv_master_thread)(
 		}
 	}
 
+	/* Enable temporary tablespace encryption if set */
+	srv_enable_temp_encryption_if_set();
```
laurynas-biveinis: Why here - at the start of shutdown? Especially when the called function exits first thing in this srv_shutdown_state

storage/innobase/handler/ha_innodb.cc
```diff
@@ -10994,18 +10994,44 @@ create_table_info_t::create_table_def()
 			dict_mem_table_free(table);
 		} else if (m_create_info->encrypt_type.length > 0
 			   && !Encryption::is_none(
-				   m_create_info->encrypt_type.str)) {
+				   m_create_info->encrypt_type.str)
+			   && !srv_tmp_tablespace_encrypt) {
```
laurynas-biveinis: Two loads in a row of srv_tmp_tablespace_encrypt can race with changing the variable value.

laurynas-biveinis (changes requested):

storage/innobase/srv/srv0srv.cc
gl-sergei: hmm... I was pretty sure I was putting it inside of the loop :)

storage/innobase/handler/ha_innodb.cc
gl-sergei: I don't think that changing of `srv_tmp_tablespace_encrypt` between loads here is a big deal. But now I think that instead of `srv_tmp_tablespace_encrypt` we must check encryption flag of `srv_tmp_space`.

storage/innobase/srv/srv0srv.cc
laurynas-biveinis: Yet the current testcase succeeds - maybe need a new one where the server is started with off and switched to on dynamically?

gl-sergei:
Test case changed to start with encryption turned off (grep used to check that), then it turns encryption on and checks that by grep as well.
I put the code to enable encryption on existing tablespace into separate function in `fsp0fsp.cc`. Not sure if it would be better to have two functions - `fsp_` and `fil_` ones.
Reading through comments on my second PR, I think I should use `p_s` instead of `i_s` here as well.
Jenkins run is here: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1364/

laurynas-biveinis (changes requested):
The issue of variable being enabled dynamically become effective asynchronously still stands: query thread sets the variable, master thread makes it effective, and there is a race window between the master thread and any query thread depending on it being effective.

gl-sergei:
@laurynas-biveinis  Agree. But it is an issue with the test case. I was thinking about making the test require debug_sync for that reason. Do you think it is a good idea?

laurynas-biveinis:
No, it's a feature issue. User sets variable, how long has he wait until he's sure the setting has taken effect? Please check 8.0 redo log encryption how they solved this.

gl-sergei:
There is another possible issue which I need to investigate. When update to temp tablespace is done concurrently with CREATE TEMP TABLE. This actually may cause some damage.

gl-sergei:
Time window between variable update and it taking effect doesn't look like a big issue to me.

laurynas-biveinis:
If a sync bug can happen, it will, https://bugs.mysql.com/bug.php?id=87303. The testcases can sync by the InnoDB master thread loop count status variable

laurynas-biveinis:
On a second thought, please do review 8.0 behavior for the redo log encryption enabling. Syncing by thread loop count is fragile and if user needs confirmation that the feature is indeed enabled, it's the least intuitive thing to remember

gl-sergei:
@laurynas-biveinis  MySQL 8 doesn't care much about the time gape between variable being set and it taking effect. They seems to be OK with it to take into effect eventually or silently flipping back to OFF... Do you think changing our implementation to be synchronous is the right way to go?

laurynas-biveinis:
MySQL 8 does - the variable value changes only after it becomes effective, check that linked bug. It is not synchronous implementation (and ours does not have to be), it is the master thread which updates the variable value.

gl-sergei:
Variable changes immediately. It flips back to OFF in case of unsuccessful attempt to turn encryption on.

gl-sergei:
Looks like all comments are taken to attention. Behaviour is now exactly the same as in MySQL 8 when it turns ON encryption for redo/undo logs. Yes, there is still a time gap and it is impossible to know exact moment when change became effective. In case of error, variable becomes OFF and error message gets logged. Turning it OFF manually doesn't turn encryption off.

Added a test case creating temp table before encryption is enabled and destroying it afterwards. It also is trying to create a table when encryption is already enabled on FIL level, but haven't yet enabled at FSP level.

https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1384/

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,113 @@
...
+SET GLOBAL innodb_temp_tablespace_encrypt = ON;
+
+--error ER_ILLEGAL_HA_CREATE_OPTION
+CREATE TEMPORARY TABLE t02 (a INT) ENGINE=InnoDB ENCRYPTION='Y';
+
+CREATE TEMPORARY TABLE t03 (a TEXT) ENGINE=InnoDB ENCRYPTION='N';
+INSERT INTO t03 VALUES ('Curabitur laoreet, velit non interdum venenatis');
+
+# wait until pages flushed
+--let $wait_condition= SELECT COUNT(*) = 0 FROM performance_schema.global_status WHERE LOWER(variable_name) = 'innodb_buffer_pool_pages_dirty' AND variable_value > 0
```
gl-sergei: Maybe `set global innodb_buf_flush_list_now = 1;` is better, but it requires debug build.

laurynas-biveinis: Release/debug testing > debug testing, thus the current way is by necessity better

laurynas-biveinis: How about setting innodb_max_dirty_pages_pct to 0?

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,113 @@
...
+
+# make sure ibtmp1 is encrypted now
+
+--let $grep_pattern= Curabitur laoreet, velit non interdum venenatis
+--let $grep_file= $MYSQL_DATA_DIR/ibtmp1
+--let $grep_output= boolean
+--source include/grep_pattern.inc
+
+# this table created in separate file per table tablespace
+CREATE TEMPORARY TABLE t04 (a INT) ENGINE=InnoDB ROW_FORMAT=COMPRESSED;
```
laurynas-biveinis: What is tested here?

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,113 @@
...
+INSERT INTO t04 VALUES (1), (2), (3);
+
+--error ER_ILLEGAL_HA_CREATE_OPTION
+CREATE TEMPORARY TABLE t05 (a INT) ENGINE=InnoDB ROW_FORMAT=COMPRESSED ENCRYPTION='y';
+
+CREATE TEMPORARY TABLE t06 (a INT) ENGINE=InnoDB ROW_FORMAT=COMPRESSED ENCRYPTION='n';
```
laurynas-biveinis: And here?

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,113 @@
...
+INSERT INTO t10 (b) VALUES (FLOOR(RAND() * 10000)), (FLOOR(RAND() * 10000)), (FLOOR(RAND() * 10000));
+INSERT INTO t10 (b) VALUES (FLOOR(RAND() * 10000)), (FLOOR(RAND() * 10000)), (FLOOR(RAND() * 10000));
+INSERT INTO t10 (b) VALUES (FLOOR(RAND() * 10000)), (FLOOR(RAND() * 10000)), (FLOOR(RAND() * 10000));
+INSERT INTO t10 (b) VALUES (FLOOR(RAND() * 10000)), (FLOOR(RAND() * 10000)), (FLOOR(RAND() * 10000));
+INSERT INTO t10 (b) VALUES (FLOOR(RAND() * 10000)), (FLOOR(RAND() * 10000)), (FLOOR(RAND() * 10000));
+
+INSERT INTO t10 (b) SELECT b FROM t10;
+INSERT INTO t10 (b) SELECT b FROM t10;
+INSERT INTO t10 (b) SELECT b FROM t10;
+INSERT INTO t10 (b) SELECT b FROM t10;
```
laurynas-biveinis: What is the purpose of making t10 big?

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,113 @@
...
+# test restarting
+let $restart_hide_args = 1;
+let $restart_parameters = restart: --early-plugin-load="keyring_file=$KEYRING_PLUGIN" --loose-keyring_file_data=$MYSQL_TMP_DIR/ts_encrypt_keyring --loose-innodb-temp-tablespace-encrypt $KEYRING_PLUGIN_OPT;
+--source include/restart_mysqld.inc
+
+CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB;
+INSERT INTO t01 VALUES (1), (2), (3);
+
+# test crashing
+let $restart_hide_args= 1;
```
laurynas-biveinis: Some MTR variable assignments done with space before =, some without

mysql-test/suite/innodb/t/temp_table_encrypt_debug.test
```diff
@@ -0,0 +1,50 @@
...
+
+connect (other2,localhost,root,,test);
+connection other2;
+
+CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB;
+INSERT INTO t01 VALUES (1), (2), (3);
+
+connection default;
+
+# unpause master thread
```
laurynas-biveinis: s/unpause/resume

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -1050,6 +1045,53 @@ fsp_header_rotate_encryption(
...
 
+/** Enable encryption for already existing tablespace.
+@param[in]	space_id	tablespace id
+@return true if success */
+bool
+fsp_enable_encryption(
+	ulint space_id)
+{
+	byte		encrypt_info[ENCRYPTION_INFO_SIZE_V2];
+	mtr_t		mtr;
```
laurynas-biveinis: Declare at first use

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -1050,6 +1045,53 @@ fsp_header_rotate_encryption(
...
+	fil_space_t*	space = fil_space_get(space_id);
+
+	const page_size_t	page_size(space->flags);
```
laurynas-biveinis: Likewise

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21686,6 +21720,12 @@ static MYSQL_SYSVAR_BOOL(sync_debug, srv_sync_debug,
   PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_READONLY,
   "Enable the sync debug checks",
   NULL, NULL, FALSE);
+
+static MYSQL_SYSVAR_ULONG(master_encrypt_debug,
+  srv_master_encrypt_debug,
+  PLUGIN_VAR_OPCMDARG,
+  "Pause in the middle of the enabling of temporary tablespace encryption.",
```
laurynas-biveinis: Could you please describe 0, 1, 2 values here to make it easier to use for any future testcases?

storage/innobase/include/fsp0sysspace.h
```diff
@@ -282,6 +282,17 @@ extern SysTablespace srv_sys_space;
 /** The control info of a temporary table shared tablespace. */
 extern SysTablespace srv_tmp_space;
 
+/** Check if the space_id is for a shared system tablespace.
+@param[in]	id	Space ID to check
+@return true if id is a system tablespace, false if not. */
+UNIV_INLINE
+bool
```
laurynas-biveinis, line 289: warn_unused_result?

storage/innobase/include/os0file.h
```diff
@@ -514,6 +514,11 @@ struct Encryption {
 		ulint			dst_len)
 		MY_ATTRIBUTE((warn_unused_result));
 
+#ifndef UNIV_INNOCHECKSUM
+	/** Check if keyring plugin loaded. */
+	static bool check_keyring();
```
laurynas-biveinis: warn_unused_result

storage/innobase/os/os0file.cc
```diff
@@ -9657,6 +9657,49 @@ Encryption::decrypt(
...
 
+#ifndef UNIV_INNOCHECKSUM
+
+/** Default master key for bootstrap */
+static const char ENCRYPTION_DEFAULT_MASTER_KEY[] = "DefaultMasterKey";
+
+/** Check if keyring plugin loaded. */
+bool Encryption::check_keyring()
+{
+	char*	key_type = NULL;
```
laurynas-biveinis: Declare at first uses

storage/innobase/srv/srv0srv.cc
```diff
@@ -2675,6 +2681,73 @@ srv_master_do_shutdown_tasks(
 	return(n_bytes_merged || n_tables_to_drop);
 }
 
+void
```
laurynas-biveinis: static, header comment

storage/innobase/srv/srv0srv.cc
```diff
@@ -2675,6 +2681,73 @@ srv_master_do_shutdown_tasks(
...
+	if (srv_shutdown_state != SRV_SHUTDOWN_NONE) {
+		return;
+	}
+
+	if (is_shared_system_tablespace(srv_tmp_space.space_id())) {
+		/* there is no separate tablespace for temporary tables */
+		return;
+	}
+
+	if (srv_tmp_tablespace_encrypt) {
```
laurynas-biveinis: if (!srv_tmp_tablespace_encrypt) return (and make it 1st or 2nd if in the function, looks the cheapest), then everything below has one level less indentation

storage/innobase/srv/srv0srv.cc
```diff
@@ -2675,6 +2681,73 @@ srv_master_do_shutdown_tasks(
...
+
+		if (srv_read_only_mode) {
+			srv_tmp_tablespace_encrypt = false;
+			ib::error() << "Can't set temporary tablespace to be"
+				<< " encrypted in read-only mode.";
```
laurynas-biveinis: InnoDB read only mode, to tell apart from the core server read only mode

storage/innobase/srv/srv0srv.cc
```diff
@@ -2675,6 +2681,73 @@ srv_master_do_shutdown_tasks(
...
+		/* Make sure the keyring is loaded. */
+		if (!Encryption::check_keyring()) {
+			srv_tmp_tablespace_encrypt = false;
+			ib::error() << "Can't set temporary tablespace to be"
+				<< " encrypted because keyring plugin is"
+				<< " not available.";
+			return;
+		}
+
+		fil_space_t* space = fil_space_get(srv_tmp_space.space_id());
```
laurynas-biveinis: const ?

storage/innobase/srv/srv0srv.cc
```diff
@@ -2675,6 +2681,73 @@ srv_master_do_shutdown_tasks(
...
+
+		if (srv_read_only_mode) {
+			srv_tmp_tablespace_encrypt = false;
+			ib::error() << "Can't set temporary tablespace to be"
+				<< " encrypted in read-only mode.";
+			return;
+		}
+
+		/* Make sure the keyring is loaded. */
+		if (!Encryption::check_keyring()) {
```
laurynas-biveinis: I wonder if this is really required here - fsp_enable_encryption below would call fsp_header_fill_encryption_info and would check the same anyway? But maybe worse diagnostics would be produced?

storage/innobase/srv/srv0srv.cc
```diff
@@ -2675,6 +2681,73 @@ srv_master_do_shutdown_tasks(
...
+				<< " encrypted because keyring plugin is"
+				<< " not available.";
+			return;
+		}
+
+		fil_space_t* space = fil_space_get(srv_tmp_space.space_id());
+
+		ut_ad(fsp_is_system_temporary(space->id));
+
+		if (!FSP_FLAGS_GET_ENCRYPTION(space->flags)) {
```
laurynas-biveinis: Is it possible for this condition to be false legally? Maybe it can be replaced by an assert?

storage/innobase/srv/srv0srv.cc
```diff
@@ -2675,6 +2681,73 @@ srv_master_do_shutdown_tasks(
...
+
+			if (!fsp_enable_encryption(srv_tmp_space.space_id())) {
+				srv_tmp_tablespace_encrypt = false;
+				ib::warn() << "Can't set temporary"
```
laurynas-biveinis: One failure above is ib::error(), these two are ib::warn() - should they all be the same?

storage/innobase/srv/srv0srv.cc
```diff
@@ -2675,6 +2681,73 @@ srv_master_do_shutdown_tasks(
...
+				srv_master_encrypt_debug = 2;
+				while (srv_master_encrypt_debug != 0) {
+					os_thread_sleep(10000);
+				}
+			}
+#endif
+
+			space->flags |= FSP_FLAGS_MASK_ENCRYPTION;
+
+			if (fil_set_encryption(space->id,
```
laurynas-biveinis: Looking at what fil_set_encryption may return under what conditions, it looks like this check can be replaced by an assert?

storage/innobase/srv/srv0start.cc
```diff
@@ -1231,6 +1231,20 @@ srv_open_tmp_tablespace(
 		it stays open until shutdown. */
 		if (fil_space_open(tmp_space->name())) {
 
+			if (srv_tmp_tablespace_encrypt) {
+
+				fil_space_t*		space;
+
+				space = fil_space_get(temp_space_id);
```
laurynas-biveinis: Merge with declaration, declare as const

laurynas-biveinis (changes requested):

laurynas-biveinis:
Please also do a test MTR run with the feature and its supporting keyring enabled by default, for any crashes/surprises

mysql-test/suite/innodb/t/temp_table_encrypt.test
gl-sergei: Added innodb_max_dirty_pages_pct=0

mysql-test/suite/innodb/t/temp_table_encrypt.test
gl-sergei: This test intended to check that temp table created outside of shared temp tablespace has no issues. I added a check that it is encrypted as well.

mysql-test/suite/innodb/t/temp_table_encrypt.test
gl-sergei: Basically I wanted to check that ENCRYPTION='y' is still not allowed for temp tables.

mysql-test/suite/innodb/t/temp_table_encrypt.test
gl-sergei: It needs to be big enough so that CREATE INDEX uses temp table

mysql-test/suite/innodb/t/temp_table_encrypt.test
gl-sergei: fixed

mysql-test/suite/innodb/t/temp_table_encrypt_debug.test
gl-sergei: fixed

storage/innobase/fsp/fsp0fsp.cc
gl-sergei: fixed

storage/innobase/fsp/fsp0fsp.cc
gl-sergei: fixed

storage/innobase/handler/ha_innodb.cc
gl-sergei: done

storage/innobase/include/fsp0sysspace.h
gl-sergei: added

storage/innobase/include/os0file.h
gl-sergei: done

storage/innobase/os/os0file.cc
gl-sergei: done

storage/innobase/srv/srv0srv.cc
gl-sergei: done

storage/innobase/srv/srv0srv.cc
gl-sergei: done

storage/innobase/srv/srv0srv.cc
gl-sergei: done

storage/innobase/srv/srv0srv.cc
gl-sergei: Yes, mainly because fsp_header_fill_encryption_info does not tell what was the error.

storage/innobase/srv/srv0srv.cc
gl-sergei: const doesn't work because of `space->flags |= FSP_FLAGS_MASK_ENCRYPTION;`

storage/innobase/srv/srv0srv.cc
gl-sergei: Yes, it possible and basically it is condition to not update FSP headers each time.

storage/innobase/srv/srv0srv.cc
gl-sergei: changed all to be ib:error

storage/innobase/srv/srv0srv.cc
gl-sergei: replaced by assert

storage/innobase/srv/srv0start.cc
gl-sergei: cannot be const, but merged with declaration

gl-sergei:
Also 
- added test with keyring_vault which revealed race condition in the test case and fixed that race condition
- enforced encryption for file per table TEMP tables

Jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1529/

I ran innodb test suite with keyring_file and innotb-temp-table-encrypt=ON locally, had trouble making keyring_file to be preloaded in Jenkins

mysql-test/suite/innodb/include/temp_table_encrypt.inc
```diff
@@ -0,0 +1,158 @@
...
+
+CREATE INDEX t10_b ON t10 (b);
+DROP INDEX t10_b ON t10;
+
+CREATE INDEX t10_b ON t10 (b) ALGORITHM=COPY;
+DROP INDEX t10_b ON t10 ALGORITHM=COPY;
+
+# test restarting
+let $restart_hide_args = 1;
+let $restart_parameters = $KEYRING_RESTART_PARAM --innodb-temp-tablespace-encrypt --innodb_max_dirty_pages_pct=0;
```
laurynas-biveinis, line 129: innodb_max_dirty_pages_pct not required in this and further restarts?

storage/innobase/srv/srv0srv.cc
```diff
@@ -2675,6 +2681,77 @@ srv_master_do_shutdown_tasks(
...
+		space->flags |= FSP_FLAGS_MASK_ENCRYPTION;
+
+		dberr_t	err = fil_set_encryption(space->id,
+			Encryption::AES, NULL, NULL);
+
+		ut_a(err == DB_SUCCESS);
+
+#ifdef UNIV_DEBUG
+		if (srv_master_encrypt_debug != 0) {
+			srv_master_encrypt_debug = 2;
```
laurynas-biveinis, line 2733: How is this working without atomics? Not even volatile, which would also be wrong, but at least it would force loads and stores as written

gl-sergei: There is a `os_thread_sleep` call in the loop which can potentially use and modify any global variable, compiler would be terribly wrong if it would optimize `srv_master_encrypt_debug = 2` away.

mysql-test/suite/innodb/include/temp_table_encrypt.inc
gl-sergei: well, maybe it is not required, but it does no harm either :) I think for this test case it even useful as it makes actual encryption to happen with higher probability.

gl-sergei:
Should I rebase this PR before the next review round?

mysql-test/suite/perfschema/t/show_sanity.test
```diff
@@ -416,6 +416,7 @@ insert into test.sanity values
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "INNODB_CHANGE_BUFFERING"),
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "INNODB_CHANGE_BUFFERING_DEBUG"),
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "INNODB_CHANGE_BUFFER_MAX_SIZE"),
+  ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "INNODB_MASTER_ENCRYPT_DEBUG"),
```
percona-ysorokin: Try to maintain alphabetical order here.

percona-ysorokin: `INNODB_TEMP_TABLESPACE_ENCRYPT` should also be added to this list.

storage/innobase/os/os0file.cc
```diff
@@ -9657,6 +9657,48 @@ Encryption::decrypt(
...
+bool Encryption::check_keyring()
+{
+	bool	ret = false;
+
+	char	key_name[ENCRYPTION_MASTER_KEY_NAME_MAX_LEN];
+	memset(key_name, 0, ENCRYPTION_KEY_LEN);
+	strcpy(key_name, ENCRYPTION_DEFAULT_MASTER_KEY);
+
+	/* We call key ring API to generate master key here. */
+	int	my_ret = my_key_generate(key_name, "AES",
```
percona-ysorokin, line 9673: Checking `my_ret` result here?

percona-ysorokin (changes requested):

laurynas-biveinis:
Yes, please rebase

gl-sergei:
Already did today, and addressed Yura's comments. Here is the Jenkins build https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1623/

percona-ysorokin (approved):

mysql-test/suite/innodb/include/temp_table_encrypt.inc
```diff
@@ -0,0 +1,158 @@
...
+# we need restart to make sure keyring settings are picked up
+let $restart_hide_args = 1;
+let $restart_parameters = $KEYRING_RESTART_PARAM --innodb_max_dirty_pages_pct=0;
+--source include/restart_mysqld.inc
+
+CREATE TEMPORARY TABLE t01 (a TEXT) ENGINE=InnoDB;
+INSERT INTO t01 VALUES ('Lorem ipsum dolor sit amet, consectetur adipiscing elit');
+
+# wait until pages flushed
+--let $wait_condition= SELECT COUNT(*) = 0 FROM performance_schema.global_status WHERE LOWER(variable_name) = 'innodb_buffer_pool_pages_dirty' AND variable_value > 0
```
laurynas-biveinis: Looks like a typo: isn't wait_condition supposed to be wait *termination* condition, i.e. value = 0 ?

mysql-test/suite/innodb/include/temp_table_encrypt.inc
```diff
@@ -0,0 +1,158 @@
...
+--source include/grep_pattern.inc
+
+
+# this table created in separate file per table tablespace, make sure it not
+# encrypted
+CREATE TEMPORARY TABLE t04 (a TEXT) ENGINE=InnoDB ROW_FORMAT=COMPRESSED;
+INSERT INTO t04 VALUES ('Praesent tristique eros a tempus fringilla');
+
+# wait until pages flushed
+--let $wait_condition= SELECT COUNT(*) = 0 FROM performance_schema.global_status WHERE LOWER(variable_name) = 'innodb_buffer_pool_pages_dirty' AND variable_value > 0
```
laurynas-biveinis: Likewise

mysql-test/suite/innodb/include/temp_table_encrypt.inc
```diff
@@ -0,0 +1,158 @@
...
+--source include/wait_condition.inc
+
+--error ER_ILLEGAL_HA_CREATE_OPTION
+CREATE TEMPORARY TABLE t02 (a INT) ENGINE=InnoDB ENCRYPTION='Y';
+
+CREATE TEMPORARY TABLE t03 (a TEXT) ENGINE=InnoDB ENCRYPTION='N';
+INSERT INTO t03 VALUES ('Curabitur laoreet, velit non interdum venenatis');
+
+# wait until pages flushed
+--let $wait_condition= SELECT COUNT(*) = 0 FROM performance_schema.global_status WHERE LOWER(variable_name) = 'innodb_buffer_pool_pages_dirty' AND variable_value > 0
```
laurynas-biveinis: Likewise

mysql-test/suite/innodb/include/temp_table_encrypt.inc
```diff
@@ -0,0 +1,158 @@
...
+--let $grep_file= $MYSQL_DATA_DIR/ibtmp1
+--let $grep_output= boolean
+--source include/grep_pattern.inc
+
+# this table created in separate file per table tablespace, make sure it is
+# encrypted as well
+CREATE TEMPORARY TABLE t04 (a TEXT) ENGINE=InnoDB ROW_FORMAT=COMPRESSED;
+INSERT INTO t04 VALUES ('Praesent tristique eros a tempus fringilla');
+
+# wait until pages flushed
```
laurynas-biveinis: Likewise

mysql-test/suite/innodb/include/temp_table_encrypt.inc
```diff
@@ -0,0 +1,158 @@
...
+# test that we can turn encryption OFF and ON
+
+SET GLOBAL innodb_temp_tablespace_encrypt = OFF;
+
+CREATE TEMPORARY TABLE t07 (a INT) ENGINE=InnoDB;
+INSERT INTO t07 VALUES (1), (2), (3);
+
+SET GLOBAL innodb_temp_tablespace_encrypt = ON;
+
+# create large enought table inorted to make CREATE INDEX to use temporary table
```
laurynas-biveinis: "enought", "inorted"

storage/innobase/include/srv0srv.h
```diff
@@ -292,6 +292,9 @@ extern my_bool	srv_undo_log_truncate;
 /** UNDO logs not redo logged, these logs reside in the temp tablespace.*/
 extern const ulong	srv_tmp_undo_logs;
 
+/** Enable or disable Encrypt of temporary tablespace.*/
```
laurynas-biveinis: s/Encrypt/encryption

storage/innobase/srv/srv0srv.cc
```diff
@@ -2675,6 +2681,77 @@ srv_master_do_shutdown_tasks(
...
+		return;
+	}
+
+	fil_space_t* space = fil_space_get(srv_tmp_space.space_id());
+
+	ut_ad(fsp_is_system_temporary(space->id));
+
+	if (!FSP_FLAGS_GET_ENCRYPTION(space->flags)) {
+
+		if (srv_read_only_mode) {
```
laurynas-biveinis: Add a test?

laurynas-biveinis (changes requested):
Please log a JIRA issue if not done already and mention it in the commit message header, remove the LP blueprint URL.

mysql-test/suite/innodb/include/temp_table_encrypt.inc
gl-sergei: Actual condition here is `COUNT(*) = 0`, i.e. there is no variable with name `innodb_buffer_pool_pages_dirty` and `value > 0`

gl-sergei: I'll rewrite it to be more readable

storage/innobase/srv/srv0srv.cc
gl-sergei: I guess this branch is not reachable because there is `ut_ad(!srv_read_only_mode);` in master thread. Will replace it with assert.

gl-sergei:
- changed commit message to refer JIRA issue
- rewrote SELECT to be a bit shorter
- replaced if (srv_read_only_mode) with assert, checked that it is indeed not reachable
- grammar errors in comments

https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1659/

laurynas-biveinis (approved):

PR #1863, opened 2017-08-03
[BLD-782] Add werning message

EvgeniyPatlan:


abychko (approved):
LGFM

george-lorch:
+1, the more places we can put this in users face, the better.

PR #1867, opened 2017-08-04
Ps 5.7 myr 154

george-lorch:
https://jira.percona.com/browse/MYR-157

Cherry-picked the following : 
https://github.com/facebook/mysql-5.6/commit/001ca1fd52157a802656ca6d3b362c56aa8cfc02
https://github.com/facebook/mysql-5.6/commit/6f99962f8d2d2411d7a68c27037dc96801c4b487
https://github.com/facebook/mysql-5.6/commit/d344ec1b59508b5e4956b398eef632563529ca98
https://github.com/facebook/mysql-5.6/commit/a44d6c287111c63d35fde0222adfebeb54d601c7
https://github.com/facebook/mysql-5.6/commit/5159416ed973bd9cbbb090d98ec6307054a11493

These alter the server errmsg file sql/share/errmsg-utf8.txt to add MyRocks engine specific error codes to the 12000+ range.


george-lorch:
Jenkins http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1105/

mysql-test/suite/rocksdb/t/bulk_load_errors.test
```diff
@@ -0,0 +1,39 @@
...
+# Make sure we get an error with overlapping data
+SET rocksdb_bulk_load=1;
+INSERT INTO t1 VALUES(1);
+INSERT INTO t1 VALUES(2);
+INSERT INTO t1 VALUES(20);
+INSERT INTO t1 VALUES(21);
+
+# This last crashes the server (intentionally) because we can't return any
+# error information from a SET <variable>=<value>
+--exec echo "wait" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
```
laurynas-biveinis: --source include/expect_crash.inc

mysql-test/suite/rocksdb/t/bulk_load_errors.test
```diff
@@ -0,0 +1,39 @@
...
+--error 2013 # CR_SERVER_LOST
+SET rocksdb_bulk_load=0;
+
+--exec grep "RocksDB: Error 197 finalizing last SST file while setting bulk loading variable" $MYSQLTEST_VARDIR/log/mysqld.1.err | cut -d] -f2
```
laurynas-biveinis: This does not work if some previous test on the same MTR worker also happened to have the same bulk load error (the fix is the same as for the next comment)

mysql-test/suite/rocksdb/t/bulk_load_errors.test
```diff
@@ -0,0 +1,39 @@
...
+--exec echo "" >$MYSQLTEST_VARDIR/log/mysqld.1.err
```
laurynas-biveinis: This ruins troubleshooting, and CURRENT_TEST markers in the log which MTR uses. Please restart the server with custom error log location at the start of the testcase, restart with default at the end, and remove the custom error log.

mysql-test/suite/rocksdb/t/bulk_load_errors.test
```diff
@@ -0,0 +1,39 @@
...
+
+# restart the crashed server
```
laurynas-biveinis: --source include/start_mysqld.inc

mysql-test/suite/rocksdb/t/bulk_load_errors.test
```diff
@@ -0,0 +1,39 @@
+--source include/have_rocksdb.inc
```
laurynas-biveinis, line 1: Add --source include/not_crashrep.inc

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -247,30 +243,27 @@ static int rocksdb_create_checkpoint(THD *const thd MY_ATTRIBUTE((__unused__)),
                             checkpoint_dir.c_str());
       rocksdb::Checkpoint *checkpoint;
       auto status = rocksdb::Checkpoint::Create(rdb, &checkpoint);
+      // We can only return HA_EXIT_FAILURE/HA_EXIT_SUCCESS here which is why
+      // the return code is ignored, but by calling into rdb_error_to_mysql,
+      // it will call my_error for us, which will propogate up to the client.
+      int rc __attribute__((__unused__));
```
laurynas-biveinis: MY_ATTRIBUTE

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb/t/bulk_load_errors.test
george-lorch: Ahh crap, yeah, I missed this when importing the commits, should use the $restart_parameters=restart:--log-error=$some_file idiom

george-lorch:
OK, fixed this stuff and pushed, new jenkins http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1108/ release builds reporting failed but really didn't fail due to https://jira.percona.com/browse/JEN-676

mysql-test/suite/rocksdb/t/bulk_load_errors.test
```diff
@@ -0,0 +1,47 @@
...
+--source include/not_crashrep.inc
+
+--let SEARCH_FILE=$MYSQLTEST_VARDIR/tmp/rocksdb.bulk_load_errors.err
+--let $restart_parameters=restart:--log-error=$SEARCH_FILE
+--replace_result $SEARCH_FILE LOG_FILE
```
laurynas-biveinis, line 6: A nit which does not require fixing, but something to consider in the future: I'd prefer replacing $MYSQLTEST_VARDIR with MYSQLTEST_VARDIR only, there is no need to hide the constant file path part

mysql-test/suite/rocksdb/t/bulk_load_errors.test
```diff
@@ -0,0 +1,47 @@
...
+# Make sure the error exists in the .err log and then restart the server
+--let SEARCH_PATTERN=RocksDB: Error 207 finalizing last SST file while setting bulk loading variable
+--source include/search_pattern_in_file.inc
+
+--let SEARCH_PATTERN=
+--let SEARCH_FILE=
+--let $restart_parameters=
+--source include/start_mysqld.inc
+
+--enable_reconnect
```
laurynas-biveinis: These two lines are consumed by start_mysqld.inc

mysql-test/suite/rocksdb/t/bulk_load_errors.test
```diff
@@ -0,0 +1,47 @@
...
+# error information from a SET <variable>=<value>
+--source include/expect_crash.inc
+--error 2013 # CR_SERVER_LOST
+SET rocksdb_bulk_load=0;
+
+# Make sure the error exists in the .err log and then restart the server
+--let SEARCH_PATTERN=RocksDB: Error 207 finalizing last SST file while setting bulk loading variable
+--source include/search_pattern_in_file.inc
+
+--let SEARCH_PATTERN=
```
laurynas-biveinis: Nit: no need to reset those if you don't search_pattern_in_file again (and SEARCH_FILE is required by one of the later comments, see below)

mysql-test/suite/rocksdb/t/bulk_load_errors.test
```diff
@@ -0,0 +1,47 @@
...
+--let SEARCH_FILE=
+--let $restart_parameters=
+--source include/start_mysqld.inc
+
+--enable_reconnect
+--source include/wait_until_connected_again.inc
+
+DROP TABLE t1;
```
laurynas-biveinis: --remove_file $SEARCH_FILE or MTR --repeat is broken

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb/t/bulk_load_errors.test
george-lorch: Ahh, cool, I did not know this, good tip.

mysql-test/suite/rocksdb/t/bulk_load_errors.test
george-lorch: Doh, yup, you're right, and I should have caught this. I usually do --repeat=3 before pushing but obviously forgot this time. This specific sequence (use alternate .err log file) seems frequent enough that I wonder if there is some relatively easy way to automate this via some --source and validate proper cleanup. Maybe even going a little higher level and add some facility to mtr like 'get_tmp_filename' that gives a properly pathed tmp file name and validates that the test has cleaned up (or auto cleanup?) on exit.

laurynas-biveinis: I'm thinking of listing tmp and log dirs before and after the testcase run and adding its diff to the testcase state checker.

A funny thing is that --repeat breaks here by false negative, so it appears to run fine

george-lorch:
OK, new one http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1109/

laurynas-biveinis (approved):

PR #1871, opened 2017-08-08
[BLD-785] Service isn't started after installation on Ubuntu

EvgeniyPatlan:


abychko (approved):
OK, reverting changes

PR #1873, opened 2017-08-09
Post fix for ps-5.5.57-merge

percona-ysorokin:
Added '@PERL_PATH@' to 'mysqldumpslow.sh' missing in the main merge commit.


laurynas-biveinis (approved):
No need to ask for review, since the same change in 5.6 has been preapproved, same applies here

PR #1885, opened 2017-08-11
Ps 5.7 myr 147

george-lorch:
Jenkins just started. Prior and local runs look good, lets see what this turns out http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1121/


george-lorch:
Bad job, fat finger on mtr args, try http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1122/

mysql-test/suite/rocksdb/t/ttl_primary.test
```diff
@@ -0,0 +1,498 @@
...
+
+INSERT INTO t1 values (UNIX_TIMESTAMP());
+--sleep 1
+INSERT INTO t1 values (UNIX_TIMESTAMP());
+SELECT COUNT(*) FROM t1;
+
+set global rocksdb_force_flush_memtable_now=1;
+set global rocksdb_compact_cf='default';
+# should all still be there..
+SELECT COUNT(*) FROM t1;
```
laurynas-biveinis, line 175: Unstable test (requires < 5s wall time since INSERT, which is not guaranteed). I am not sure whether it applies to this test directly, but upstream have switched to using debug variables for determinism in this scenarios, perhaps it's in the upstream merge queue.

mysql-test/suite/rocksdb/t/ttl_primary.test
```diff
@@ -0,0 +1,498 @@
...
+
+INSERT INTO t1 values (3, 'i', 'j', UNIX_TIMESTAMP(), 'k');
+INSERT INTO t1 values (4, 'm', 'n', UNIX_TIMESTAMP(), 'o');
+
+set global rocksdb_force_flush_memtable_now=1;
+set global rocksdb_compact_cf='default';
+--sorted_result
+SELECT a FROM t1;
+
+--sleep 5
```
laurynas-biveinis, line 254: Likewise

mysql-test/suite/rocksdb/t/ttl_primary.test
```diff
@@ -0,0 +1,498 @@
...
+INSERT INTO t1 VALUES (2,2);
+
+ALTER TABLE t1 DROP PRIMARY KEY, ADD PRIMARY KEY(b);
+set global rocksdb_force_flush_memtable_now=1;
+--sleep 1
+set @@global.rocksdb_compact_cf = 'default';
+--sorted_result
+SELECT COUNT(*) FROM t1;
+
+--sleep 4
```
laurynas-biveinis, line 386: Likewise

mysql-test/suite/rocksdb/t/ttl_primary.test
```diff
@@ -0,0 +1,498 @@
...
+set global rocksdb_force_flush_memtable_now=1;
+set global rocksdb_compact_cf='default';
+
+SELECT COUNT(*) FROM t1;
+
+ALTER TABLE t1 COMMENT = 'adsf;;ttl_duration=5;asfasdfa;ttl_col=a;asdfasdf;';
+INSERT INTO t1 values (UNIX_TIMESTAMP());
+set global rocksdb_force_flush_memtable_now=1;
+
+--sleep 1
```
laurynas-biveinis, line 413: Likewise

mysql-test/suite/rocksdb/t/ttl_primary.test
```diff
@@ -0,0 +1,498 @@
...
+INSERT INTO t1 values (1);
+INSERT INTO t1 values (3);
+INSERT INTO t1 values (5);
+--sleep 5
+INSERT INTO t1 values (7);
+INSERT INTO t1 values (9);
+
+UPDATE t1 SET a=a+1;
+--sorted_result
+SELECT * FROM t1;
```
laurynas-biveinis, line 438: Likewise

mysql-test/suite/rocksdb/t/ttl_primary_with_partitions.test
```diff
@@ -0,0 +1,245 @@
+--source include/have_rocksdb.inc
```
laurynas-biveinis, line 1: --source include/have_partition.inc

mysql-test/suite/rocksdb/t/ttl_primary_with_partitions.test
```diff
@@ -0,0 +1,245 @@
...
+INSERT INTO t1 values (4,4,'aaaa');
+INSERT INTO t1 values (5,5,'aaaaa');
+INSERT INTO t1 values (6,6,'aaaaaa');
+INSERT INTO t1 values (7,7,'aaaaaaa');
+INSERT INTO t1 values (8,8,'aaaaaaaa');
+INSERT INTO t1 values (9,9,'aaaaaaaaa');
+
+--sorted_result
+SELECT * FROM t1;
+--sleep 1
```
laurynas-biveinis, line 82: Likewise

mysql-test/suite/rocksdb/t/ttl_primary_with_partitions.test
```diff
@@ -0,0 +1,245 @@
...
+INSERT INTO t1 VALUES (9, 9, "nine", null);
+set global rocksdb_force_flush_memtable_now=1;
+set @@global.rocksdb_compact_cf = 'foo';
+set @@global.rocksdb_compact_cf = 'baz';
+
+--sorted_result
+SELECT * FROM t1;
+
+# TTL should be reset after alter table
+ALTER TABLE t1 DROP PRIMARY KEY, ADD PRIMARY KEY(`c2`,`c1`) COMMENT 'custom_p0_cfname=foo;custom_p1_cfname=bar;custom_p2_cfname=baz;';
```
laurynas-biveinis, line 140: Likewise

mysql-test/suite/rocksdb/t/ttl_primary_with_partitions.test
```diff
@@ -0,0 +1,245 @@
...
+INSERT INTO t1 VALUES (2, UNIX_TIMESTAMP(), "two", null);
+INSERT INTO t1 VALUES (3, UNIX_TIMESTAMP(), "three", null);
+INSERT INTO t1 VALUES (4, UNIX_TIMESTAMP(), "four", null);
+INSERT INTO t1 VALUES (5, UNIX_TIMESTAMP(), "five", null);
+INSERT INTO t1 VALUES (6, UNIX_TIMESTAMP(), "six", null);
+INSERT INTO t1 VALUES (7, UNIX_TIMESTAMP(), "seven", null);
+INSERT INTO t1 VALUES (8, UNIX_TIMESTAMP(), "eight", null);
+INSERT INTO t1 VALUES (9, UNIX_TIMESTAMP(), "nine", null);
+set global rocksdb_force_flush_memtable_now=1;
+--sleep 1
```
laurynas-biveinis, line 189: Likewise

mysql-test/suite/rocksdb/t/ttl_primary_with_partitions.test
```diff
@@ -0,0 +1,245 @@
...
+
+INSERT INTO t1 values (1, UNIX_TIMESTAMP());
+INSERT INTO t1 values (2, UNIX_TIMESTAMP());
+INSERT INTO t1 values (3, UNIX_TIMESTAMP());
+
+--sleep 1
+set global rocksdb_force_flush_memtable_now=1;
+set global rocksdb_compact_cf='default';
+
+# everything should still be here
```
laurynas-biveinis, line 233: Likewise

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -36,12 +36,12 @@ shutdown_server 10;
 
 --exec echo "" >$MYSQLTEST_VARDIR/log/mysqld.1.err
 
---exec echo "restart:--rocksdb_use_direct_writes=1 --rocksdb_allow_mmap_writes=1" >$_expect_file_name
+--exec echo "restart:--rocksdb_use_direct_io_for_flush_and_compaction=1 --rocksdb_allow_mmap_writes=1" >$_expect_file_name
```
laurynas-biveinis: Not in the scope of this PR, but this whole testcase must be cleaned up WRT error log handling

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -417,6 +433,8 @@ static char *rocksdb_strict_collation_exceptions;
 static my_bool rocksdb_collect_sst_properties = 1;
 static my_bool rocksdb_force_flush_memtable_now_var = 0;
 static my_bool rocksdb_force_flush_memtable_and_lzero_now_var = 0;
+static my_bool rocksdb_enable_ttl = 1;
```
laurynas-biveinis: TRUE

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -417,6 +433,8 @@ static char *rocksdb_strict_collation_exceptions;
...
+static my_bool rocksdb_reset_stats = 0;
```
laurynas-biveinis: FALSE

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -465,11 +486,37 @@ static void rocksdb_set_rocksdb_info_log_level(
...
   rocksdb_info_log_level = *static_cast<const uint64_t *>(save);
-  rocksdb_db_options.info_log->SetInfoLogLevel(
+  rocksdb_db_options->info_log->SetInfoLogLevel(
       static_cast<const rocksdb::InfoLogLevel>(rocksdb_info_log_level));
   RDB_MUTEX_UNLOCK_CHECK(rdb_sysvars_mutex);
 }
 
+static void rocksdb_set_reset_stats(
+    my_core::THD *const /* unused */,
+    my_core::st_mysql_sys_var *const var MY_ATTRIBUTE((__unused__)),
```
laurynas-biveinis: Two different styles of "unused" (I'd prefer the former as shorter)

storage/rocksdb/rdb_datadic.h
```diff
@@ -113,6 +113,13 @@ const size_t RDB_UNPACK_DATA_LEN_SIZE = sizeof(uint16_t);
 const size_t RDB_UNPACK_HEADER_SIZE =
     sizeof(RDB_UNPACK_DATA_TAG) + RDB_UNPACK_DATA_LEN_SIZE;
 
+/*
+  Data dictionary index info field sizes.
+*/
+const size_t RDB_SIZEOF_INDEX_INFO_VERSION = sizeof(uint16);
```
laurynas-biveinis: constexpr (no need to allocate storage)

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb/t/ttl_primary.test
george-lorch: True. There are two commits coming from upstream, one in next batch (May) and one in the following (June) that work towards addressing flakyness:
https://github.com/facebook/mysql-5.6/commit/ad90bf77f3a353336c0e33f5784eb9b30039880a
https://github.com/facebook/mysql-5.6/commit/891a9115e3d16e1a2b2efab11eff1880eb69ce1e

I would prefer to keep these aligned with upstream for now until I get through the merges and if we still have stability issues, then we can work through fixing them and sending the changes upstream.

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
george-lorch: I had cleaned up some stuff already in this series and missed this one, I'll fix it.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Actually, these are all GLOBAL sys_vars, they should not need initialization at all as the defaults should be set via the MYSQL_SYSVAR_* macros and init process, no?

laurynas-biveinis: No! Unless MyRocks has explicit code to iterate over its sysvars, they are not set up during init process. You'd think it's a natural thing to have, but I've been burnt several times by this and have seen others trip over the same assumption

laurynas-biveinis: Actually the SE must have some startup init code, or setting the vars through command line would simply be ignored

george-lorch: Really? Interesting, then what is the point of a 'default' value for the GLOBAL SYSVAR macros. Seems to me it should be one or the other, explicit init of the static or explicit init via SYSVAR default, not some weird mix of both. Either way, I'll fix it but they have many my_bools just using 0|1 so I think rather than just fixing these in their cherry pick commits, I will leave them as merged and append a commit of my own to the whole change set that fixed them all in one shot and watch for more of the same in future merges.

george-lorch: This is going to be nice, MyRocks uses a static global singleton `std::unique_ptr<rocksdb::DBOptions> rocksdb_db_options = rdb_init_rocksdb_db_options();` to get default values for the SYSVARS macros and thus eventually the sysvar values, following on the belief/assumption I made above. So if what you say is true, then they have potential startup init and use ordering issues waiting to happen.

george-lorch: And we have a winner:
`static unsigned long rocksdb_persistent_cache_size_mb;`
...
`static MYSQL_SYSVAR_ULONG(
    persistent_cache_size_mb, rocksdb_persistent_cache_size_mb,
    PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_READONLY,
    "Size of cache in MB for BlockBasedTableOptions::persistent_cache "
    "for RocksDB", nullptr, nullptr, rocksdb_persistent_cache_size_mb,
    /* min */ 0L, /* max */ ULONG_MAX, 0);
`
<<< Using uninitialized static for the default value in the macro. This value is not explicitly set anywhere in the code. This may explain some of the instability I have with the persistent_cache test.

george-lorch: Sure enough, my persistent_cache test instability is gone by fixing up the variable inits. This is a very good and worthwhile effort.

george-lorch:
OK, addressed the agreed stuff and also fixed another missed issue in one of the commits for rocksdb.type_varchar that was missing a --loose- prefix in .opt file.
On top of that I added 3 new and very minor commits for:
https://jira.percona.com/browse/MYR-163 : global sysvars initialized incorrectly
https://jira.percona.com/browse/MYR-164 : rocksdb.persistent_cache needs cleanup and re-record to avoid fouling mtr server err log
https://jira.percona.com/browse/MYR-165 : rocksdb.collation does not clean up alternate err log from tmp

Respun jenkins on all of this here : http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1125/

mysql-test/suite/rocksdb/t/persistent_cache.test
```diff
@@ -1,41 +1,33 @@
...
---exec echo "wait" >$_expect_file_name
-
-# restart server with correct parameters
-shutdown_server 10;
---exec echo "restart:--rocksdb_persistent_cache_path=$_cache_file_name --rocksdb_persistent_cache_size_mb=100" >$_expect_file_name
---sleep 5
---enable_reconnect
---source include/wait_until_connected_again.inc
---disable_reconnect
+--let $cache_file_name=$MYSQLTEST_VARDIR/tmp/roksdb.persistent_cache
```
laurynas-biveinis: Based on the comment next to the added exec rm -rf at the bottom, this is not a "file" but "directory"?

mysql-test/suite/rocksdb/t/persistent_cache.test
```diff
@@ -1,41 +1,33 @@
...
 
+--let $restart_parameters=restart:--rocksdb_persistent_cache_path=$cache_file_name --rocksdb_persistent_cache_size_mb=100
+--replace_result $cache_file_name cache_file
```
laurynas-biveinis: I'd mildly prefer masking out only the non-constant $MYSQLTEST_VARDIR part (IIRC I have suggested this before but you haven't stated your preference so I can keep or stop suggesting this)

mysql-test/suite/rocksdb/t/persistent_cache.test
```diff
@@ -1,41 +1,33 @@
...
 # restart server to re-read cache
---exec echo "wait" >$_expect_file_name
-shutdown_server 10;
---exec echo "restart:--rocksdb_persistent_cache_path=$_cache_file_name --rocksdb_persistent_cache_size_mb=100" >$_expect_file_name
---sleep 5
---enable_reconnect
---source include/wait_until_connected_again.inc
---disable_reconnect
+--let $restart_parameters=restart:--rocksdb_persistent_cache_path=$cache_file_name --rocksdb_persistent_cache_size_mb=100
+--replace_result $cache_file_name cache_file
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/t/ttl_primary_with_partitions.test
laurynas-biveinis: Still --source include/have_partition.inc unless, as with the previous test, there is something in the next merge?

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -5,43 +5,37 @@
...
---let $_expect_file_name= $MYSQLTEST_VARDIR/tmp/mysqld.$_server_id.expect
---exec echo "wait" >$_expect_file_name
-shutdown_server 10;
+--let SEARCH_FILE=$MYSQLTEST_VARDIR/tmp/rocksdb.use_direct_reads_writes.err
 
-# Clear the log
---exec echo "" >$MYSQLTEST_VARDIR/log/mysqld.1.err
+# Expect server crash on startup with invalid combination of arguments
+--source include/expect_crash.inc
+shutdown_server 10;
```
laurynas-biveinis: I'd swap these two lines (now it looks like the crash is expected on shutdown)

laurynas-biveinis: And replace shutdown_server with shutdown_mysqld.inc. That it will write a non-crash-expecting MTR file is immaterial, it will be overwritten by expect_crash.inc, OTOH the "10" timeout is very problematic with Valgrind runs etc.

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -5,43 +5,37 @@
...
---sleep 0.1  # Wait 100ms - that is how long the sleep is in check_expected_crash_and_restart
---exec echo "restart:" >$_expect_file_name
-
-# Cleanup
---enable_reconnect
---source include/wait_until_connected_again.inc
---disable_reconnect
+--let $restart_parameters=restart:--log-error=$SEARCH_FILE --rocksdb_use_direct_reads=1 --rocksdb_allow_mmap_reads=1
+--replace_result $SEARCH_FILE LOG_FILE
+--source include/start_mysqld.inc
```
laurynas-biveinis: This works, but with an extremely long timeout (even worse with Valgrind). The canonical MTR way is --error, --exec $MYSQLD (grep main testsuite for examples)

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -5,43 +5,37 @@
...
+--source include/search_pattern_in_file.inc
+
+
 
 # Repeat with --rocksdb-use-direct-writes
---let $_server_id= `SELECT @@server_id`
---let $_expect_file_name= $MYSQLTEST_VARDIR/tmp/mysqld.$_server_id.expect
---exec echo "wait" >$_expect_file_name
+--source include/expect_crash.inc
 shutdown_server 10;
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -5,43 +5,37 @@
...
 
---exec echo "" >$MYSQLTEST_VARDIR/log/mysqld.1.err
+# Attempt to restart the server with invalid options
+--let $restart_parameters=restart:--log-error=$SEARCH_FILE --rocksdb_use_direct_io_for_flush_and_compaction=1 --rocksdb_allow_mmap_writes=1
+--replace_result $SEARCH_FILE LOG_FILE
+--source include/start_mysqld.inc
```
laurynas-biveinis: Likewise

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -390,62 +406,82 @@ static void rocksdb_set_max_background_compactions(
 //////////////////////////////////////////////////////////////////////////////
 // Options definitions
 //////////////////////////////////////////////////////////////////////////////
-static long long rocksdb_block_cache_size;
+const ulong RDB_MAX_LOCK_WAIT_SECONDS = 1024 * 1024 * 1024;
```
laurynas-biveinis: constexpr?

laurynas-biveinis (changes requested):
The variable initialisation bugfix is better upstreamed sooner than later, right?

mysql-test/suite/rocksdb/t/persistent_cache.test
george-lorch: This does not work as you suggest, tried it, seems env var is not picked up through the process. Maybe you have an exact similar example where this does work with $restart_parameters.

laurynas-biveinis: mysql-test/include/percona_signal_handling.inc:

```
--let $test_error_log= $MYSQLTEST_VARDIR/tmp/test_threadpool_sig_handler.err
--let $restart_parameters=restart:--log-error=$test_error_log
--replace_result $MYSQLTEST_VARDIR MYSQLTEST_VARDIR
--source include/restart_mysqld.inc

```

george-lorch: Ahh, I totally misunderstood your initial request/idea and get it now.

mysql-test/suite/rocksdb/t/persistent_cache.test
george-lorch: Yeah, correct. This is/was upstream naming convention and I really didn't know until I dug into the test but didn't go back and rename the vars.

mysql-test/suite/rocksdb/t/ttl_primary_with_partitions.test
george-lorch: Doesn't exist in upstream, I'll just add it to ours on the commit introducing the test and put it into the great list of upstream mtr diffs.

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
george-lorch: There is no _exact_ example that matches this case with a plugin load needed. For some reason the changes I pushed run correctly locally, but fail across 100% of our jenkins farm:
> 17:26:54 rocksdb.use_direct_reads_writes          w3 [ fail ]
> 17:26:54         Test ended at 2017-08-15 21:26:54
> 17:26:54 
> 17:26:54 CURRENT_TEST: rocksdb.use_direct_reads_writes
> 17:26:54 mysqld-debug: Can't change dir to '/var/lib/mysql/' (Errcode: 2 - No such file or directory)
> 17:26:54 2017-08-15T21:26:50.065769Z 0 [Warning] TIMESTAMP with implicit DEFAULT value is deprecated. Please use --explicit_defaults_for_timestamp server option (see documentation for more details).
> 17:26:54 2017-08-15T21:26:50.065865Z 0 [Note] --secure-file-priv is set to NULL. Operations related to importing and exporting data are disabled
> 17:26:54 2017-08-15T21:26:50.065897Z 0 [Note] /mnt/workspace/mysql-5.7-param/BUILD_TYPE/debug/Host/ubuntu-trusty-64bit/build/sql/mysqld-debug (mysqld 5.7.18-16-debug) starting as process 17851 ...
> 17:26:54 2017-08-15T21:26:50.065941Z 0 [ERROR] Can't find error-message file '/usr/share/mysql/errmsg.sys'. Check error-message file location and 'lc-messages-dir' configuration directive.
> 17:26:54 2017-08-15T21:26:50.067389Z 0 [Warning] Can't create test file /var/lib/mysql/os-trusty-x64-2030.lower-test
> 17:26:54 2017-08-15T21:26:50.067408Z 0 [Warning] Can't create test file /var/lib/mysql/os-trusty-x64-2030.lower-test
> 17:26:54 2017-08-15T21:26:50.067811Z 0 [ERROR] failed to set datadir to /var/lib/mysql/
> 17:26:54 2017-08-15T21:26:50.067823Z 0 [ERROR] Aborting

I suspect this might be some odd difference between in-source and out-of-source builds. 
I have tried other ways of invoking --exec $MYSQLD based on numerous existing tests but every combination seems to run into one issue or another. So I'm about out of ideas here how this is supposed to work and can't find anything reasonably close in any documentation or existing suites.

Please take a look at the pushed changes and see if there is something obvious that I am missing here.

george-lorch: Actually, I think I might have just stumbled on the correct incantation...stay tuned...

george-lorch: Nope, but I have confirmed that this behaves very differently between an in-source and out-of-source build. out-of-source works fine, in-source doesn't.

laurynas-biveinis: @georgelorchpercona what is the incantation you tested that works for out-of-source build for you is the one pushed right now, right?
```
--error 1
--exec $MYSQLD --log-error=$SEARCH_FILE --rocksdb_use_direct_reads=1 --rocksdb_allow_mmap_reads=1
```

laurynas-biveinis: @georgelorchpercona , try $MYSQLD_CMD instead of $MYSQLD, this is the MTR difference between the two:

```
  my $exe_mysqld= find_mysqld($basedir);
  $ENV{'MYSQLD'}= $exe_mysqld;
  my $extra_opts= join (" ", @opt_extra_mysqld_opt);
  $ENV{'MYSQLD_CMD'}= "$exe_mysqld --defaults-group-suffix=.1 ".
    "--defaults-file=$path_config_file $extra_opts";

```

george-lorch: Correct, actually I think it is not just out-of-source but using INSTALL_PREFIX as well.

george-lorch: Yup, tried that. Without overriding all of the defaults with --no-defaults, the problem is that the rocksdb plugin is not loaded. Adding the $ROCKSDB_OPT and $ROCKSDB_LOAD or $ROCKSDB_LOAD_ADD just causes the server launch to hang. I believe it is because these options are not properly quoted for use on the command line via direct --exec

laurynas-biveinis: Can you post the resulting failing command line w/o $ROCKSDB_* options and the contents of $path_config_file?

george-lorch: Yeah, this is bizzarre...it seems that trying to use '--exec $MYSQLD' and '--exec $MYSQLD_CMD' with any kind of plugin load just hangs, no output log files, nothing. There must be some issue with the way the new shell is being created. I even tried something as simple as --plugin_load=$ROCKSDB_DIR/$ROCKSDB --plugin_load_add=$ROCKSDB_DIR/$ROCKSDB and it just doesn't want to launch the process. Echoing out the -exec just shows

> /ssd/percona/myrocks/percona-server-5.7/sql/mysqld --defaults-group-suffix=.1 --defaults-file=/ssd/percona/myrocks/percona-server-5.7/mysql-test/var/my.cnf --loose-tokudb-cache-size=512M --log-syslog=0 --log-error=/ssd/percona/myrocks/percona-server-5.7/mysql-test/var/tmp/rocksdb.use_direct_reads_writes.err --plugin_load_add=/ssd/percona/myrocks/percona-server-5.7/storage/rocksdb/ha_rocksdb.so --rocksdb_use_direct_reads=1 --rocksdb_allow_mmap_reads=1

Which looks fine to me :-\



This has got to be something totally stupid and obvious that I am missing.

laurynas-biveinis: What about defaults-file=...my.cnf contents? What does machine ps show during the hung launch? Does MTR --debug or --verbose (both can be repeated several times IIRC) show what it's actually trying to invoke or what it is waiting for? We can take this offline from PR review if more convenient

george-lorch: w/o the ROCKSDB_* options, the --exec succeeds, but then the test fails because the options being tested for mutual exclusive values don't exist as the plugin is not loaded.

Yes, lets just drop to email...

laurynas-biveinis: Does this refer to ROCKSDB_ option added manually for testing or are they present in the $MYSQLD_CMD? If it's not the latter, then regular testcase restarts shouldn't work neither, suspicious

george-lorch:
As discussed in slack, root of test problem found and test is now clean and clear. Jenkins here http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1150/

george-lorch:
Worth noting, as mentioned above, the unstable ttl_* tests due to relying on sleeps for synchronization.

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -3,45 +3,29 @@
 # Issue221
 # Turning on both --rocksdb-allow-mmap-reads and --rocksdb-use-direct-reads
 # caused an assertion in RocksDB.  Now it should not be allowed and the
-# server will not start with that configuration
```
laurynas-biveinis, line 6: Great catch to fix the comment here ;)

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -3,45 +3,29 @@
...
 
---exec echo "restart:--rocksdb_use_direct_writes=1 --rocksdb_allow_mmap_writes=1" >$_expect_file_name
---sleep 0.1
---exec echo "restart:" >$_expect_file_name
-
---enable_reconnect
---source include/wait_until_connected_again.inc
---disable_reconnect
+# We should now have an error message
+--let SEARCH_PATTERN=enable both use_direct_io_for_flush_and_compaction
```
laurynas-biveinis, line 25: Is there a chance of SEARCH_PATTERN mix-up, i.e. the 2nd restart picking up something printed by the 1st one? If none at all, OK to keep as-is, otherwise the 2nd restart should use a different error log

laurynas-biveinis (approved):

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
george-lorch: No, the search text was chosen specifically to avoid that.

PR #1889, opened 2017-08-13
TDB-72 : Can not rename table in database with non alphanum character…

george-lorch:
…s in its name.

- Fixed check for db dir existence to not use MySQL call that transforms again
  the directory name that has already been transformed.
- Added new test to validate the issue is addressed.

### DIFF
--- a/mysql-test/suite/tokudb/r/dir_per_db.result
+++ b/mysql-test/suite/tokudb/r/dir_per_db.result
@@ -1,3 +1,4 @@
+SET @orig_tokudb_dir_per_db=@@global.tokudb_dir_per_db;
 ########
 #  tokudb_dir_per_db = 1
 ########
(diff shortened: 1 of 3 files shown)


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1968/

george-lorch:
Just re-pushed to remove --source include/have_innodb.inc as it was leftover/unneeded. No new jenkins for that.

mysql-test/suite/tokudb/t/dir_per_db_rename_with_non_alnum_chars.test
```diff
@@ -0,0 +1,22 @@
+--source include/have_tokudb.inc
+
+SET @orig_tokudb_dir_per_db=@@global.tokudb_dir_per_db;
+SET GLOBAL tokudb_dir_per_db=true;
+
+CREATE DATABASE `a::a@@`;
+CREATE TABLE `a::a@@`.`t1` (a INT) ENGINE=TOKUDB;
+CREATE DATABASE `!@#$%^&*()`;
+ALTER TABLE `a::a@@`.`t1` rename `!@#$%^&*()`.`t1`;
```
laurynas-biveinis: uppercase RENAME

mysql-test/suite/tokudb/t/dir_per_db_rename_with_non_alnum_chars.test
```diff
@@ -0,0 +1,22 @@
+--source include/have_tokudb.inc
```
laurynas-biveinis: I'd avoid starting a new testcase and just add an
--echo #
--echo # TDB-72: blah blah
--echo #

block to the existing dir_per_db testcase

laurynas-biveinis (changes requested):
Minor comments, up to you to decide whether to address them

mysql-test/suite/tokudb/t/dir_per_db_rename_with_non_alnum_chars.test
george-lorch: Hmm, sounds reasonable enough, particularly since the size and scope of this test is pretty small.

PR #1890, opened 2017-08-13
Ps 5.7 tdb 72

george-lorch:


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1126/

george-lorch:
Just re-pushed to remove --source include/have_innodb.inc as it was leftover/unneeded. No new jenkins for that.

laurynas-biveinis (changes requested):

PR #1896, opened 2017-08-22
[BLD-796] New authentication module 'authentication_ldap'

EvgeniyPatlan:


build-ps/percona-server.spec
```diff
@@ -190,6 +191,7 @@ Requires:       grep
 Requires:       procps
 Requires:       shadow-utils
 Requires:       net-tools
+Requires:       openldap-devel
```
abychko: requires should be handled automatically at binary level. anyway, are you sure that development library is required for runtime?

EvgeniyPatlan:
fixed

abychko (commented):
looks good now

PR #1901, opened 2017-08-23
Ps 5.7 myr 148

george-lorch:
Merge May 2017 changes


george-lorch:
jenkins http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1164/

rocksdb.ttl_* and rocksdb.parts tests still going to be unstable as they only changed from sleep to real_sleep in mtr tests. Coming merges add code and debug variables for better synchronization and test stability.

mysql-test/suite/rocksdb/t/index_merge_rocksdb-master.opt
```diff
@@ -0,0 +1 @@
+--loose-rocksdb_strict_collation_check=off --binlog_format=row --log-bin
```
laurynas-biveinis: Why the last two?

mysql-test/suite/rocksdb/t/index_merge_rocksdb2-master.opt
```diff
@@ -0,0 +1 @@
+--loose-rocksdb_strict_collation_check=off --binlog_format=row --log-bin --transaction-isolation='READ-COMMITTED'
```
laurynas-biveinis: Dynamic variables better set in the testcase directly

storage/rocksdb/rdb_sst_info.cc
```diff
@@ -98,6 +98,8 @@ Rdb_sst_file_ordered::Rdb_sst_file::put(const rocksdb::Slice &key,
   DBUG_ASSERT(m_sst_file_writer != nullptr);
 
   // Add the specified key/value to the sst file writer
+#pragma GCC diagnostic push
```
laurynas-biveinis, line 101: This is a .cc file and thus benign, but I'd prefer to have a matching pop in order for this to survive any future refactoring better

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb/t/index_merge_rocksdb-master.opt
george-lorch: Dunno, came from upstream and I didn't really notice. Good question as binlog/rpl shouldn't have anything to do with this test.

mysql-test/suite/rocksdb/t/index_merge_rocksdb2-master.opt
george-lorch: true, binlog options again are probably not needed here, but, trx iso I think is as IIRC the test includes files that use multiple connections...I'll have to look to be sure.

storage/rocksdb/rdb_sst_info.cc
george-lorch: yeah, and as I understand this is supposed to be temporary. I'll look ahead in merge history to see if there is another change coming, if not, I'll add a pop.

mysql-test/suite/rocksdb/t/index_merge_rocksdb2-master.opt
george-lorch: Actually, I see the problem, index_merge1.inc and possibly others restart mysqld, which wipes/resets any dynamic setting back to default. Thus iso needs to be set here in the .opt file to ensure it survives restarts.

george-lorch:
Forgot the new jenkins link http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1166/

george-lorch:
Going ahead and merging since everything was addressed and jenkins runs are clean.
Reported MYR-169 : rocksdb.sys_vars.rocksdb_pause_background_work_basic is unstable

PR #1905, opened 2017-08-25
Release notes for PS 5.6.37-82.2 (5.7)

azhebel:


laurynas-biveinis:
Does this actually need a review?

azhebel:
@laurynas-biveinis always good to double-check. As I get used to the process, there will be no need to review those merges to higher versions.

laurynas-biveinis (approved):
On github, I can review the diff only anyway. For process, it's very tedious to check the parents of commits and essentially replay your workflow, a log of your actual git commands executed would be needed.

PR #1906, opened 2017-08-25
Implemented Step 1 of bp:logical-read-ahead (Merge aio page read requests) (5.6)

percona-ysorokin:
Blueprint:
https://blueprints.launchpad.net/percona-server/+spec/logical-read-ahead

Cherry-picked "Merge aio page read requests" from Facebook
commit facebook/mysql-5.6@210450a

Summary:
WebScaleSQL Feature: Logical Read-Ahead

Tries to submit multiple aio page read requests together to improve read
performance.

This code adds an array to buffer aio requests in os_aio_array_t - though
only for os_aio_read_array, for now. A new parameter (should_buffer) is
added to indicate whether an aio request should be buffered or submitted.
If should_submit is true, it will submit all bufferred aio requests on
the os_aio_array.

Only buf_read_ahead_linear is modified to utilize this functionality so
far. All other call sites are setting should_submit to true. Other
os_aio_array_t arrays will also ignore this.

If one thread calling buf_read_ahead_linear is buffering io requests but
another thread issues a normal os_aio_request, that other request will
submit all the buffered requests from buf_read_ahead_linear. This is
still better than nothing I suppose.

Test Plan:
Initial perf tests were run manually.

Jenkins runs showed all tests should be clean.

This has been working well in production at Facebook for > 1 year.

Reviewers: darnaut, inaam-rana, steaphan, pivanof, weixiang.zhai

Reviewed By: inaam-rana, pivanof, steaphan, weixiang.zhai

Subscribers: minlizhu, rongrong, jtolmer, CalvinSun

Differential Revision: https://reviews.facebook.net/D23463
Differential Revision: https://reviews.facebook.net/D32379

### DIFF
--- a/storage/innobase/buf/buf0rea.cc
+++ b/storage/innobase/buf/buf0rea.cc
@@ -123,7 +123,12 @@ buf_read_page_low(
 			use to stop dangling page reads from a tablespace
 			which we have DISCARDed + IMPORTed back */
 	ulint	offset,	/*!< in: page number */
-	trx_t*	trx)
+	trx_t*	trx,
+	bool	should_buffer)	/*!< in: whether to buffer an aio request.
+			AIO read ahead uses this. If you plan to
+			use this parameter, make sure you remember
+			to call os_aio_dispatch_read_array_submit()
+			when you're ready to commit all your requests.*/
 {
 	buf_page_t*	bpage;
 	ulint		wake_later;
@@ -229,14 +234,15 @@ not_to_recover:
 		*err = _fil_io(OS_FILE_READ | wake_later
 			       | ignore_nonexistent_pages,
 			       sync, space, zip_size, offset, 0, zip_size,
-			       bpage->zip.data, bpage, trx);
+			       bpage->zip.data, bpage, trx, should_buffer);
 	} else {
 		ut_a(buf_page_get_state(bpage) == BUF_BLOCK_FILE_PAGE);
 
 		*err = _fil_io(OS_FILE_READ | wake_later
 			      | ignore_nonexistent_pages,
 			      sync, space, 0, offset, 0, UNIV_PAGE_SIZE,
-			      ((buf_block_t*) bpage)->frame, bpage, trx);
+			      ((buf_block_t*) bpage)->frame, bpage, trx,
+			      should_buffer);
 	}
 
 	if (sync) {
@@ -395,7 +401,7 @@ read_ahead:
 				&err, false,
 				ibuf_mode | OS_AIO_SIMULATED_WAKE_LATER,
 				space, zip_size, FALSE,
-				tablespace_version, i, trx);
+				tablespace_version, i, trx, false);
 			if (err == DB_TABLESPACE_DELETED) {
 				ut_print_timestamp(stderr);
 				fprintf(stderr,
@@ -459,7 +465,7 @@ buf_read_page(
 
 	count = buf_read_page_low(&err, true, BUF_READ_ANY_PAGE, space,
 				  zip_size, FALSE,
-				  tablespace_version, offset, trx);
+				  tablespace_version, offset, trx, false);
 	srv_stats.buf_pool_reads.add(count);
 	if (err == DB_TABLESPACE_DELETED) {
 		ut_print_timestamp(stderr);
@@ -507,7 +513,7 @@ buf_read_page_async(
 				  | OS_AIO_SIMULATED_WAKE_LATER
 				  | BUF_READ_IGNORE_NONEXISTENT_PAGES,
 				  space, zip_size, FALSE,
-				  tablespace_version, offset, NULL);
+				  tablespace_version, offset, NULL, false);
 	srv_stats.buf_pool_reads.add(count);
 
 	/* We do not increment number of I/O operations used for LRU policy
@@ -775,7 +781,8 @@ buf_read_ahead_linear(
 			count += buf_read_page_low(
 				&err, false,
 				ibuf_mode,
-				space, zip_size, FALSE, tablespace_version, i, trx);
+				space, zip_size, FALSE, tablespace_version,
+				i, trx, true);
 			if (err == DB_TABLESPACE_DELETED) {
 				ut_print_timestamp(stderr);
 				fprintf(stderr,
@@ -788,6 +795,7 @@ buf_read_ahead_linear(
 			}
 		}
 	}
+	os_aio_dispatch_read_array_submit();
 
 	/* In simulated aio we wake the aio handler threads only after
 	queuing all aio requests, in native aio the following call does
@@ -865,7 +873,7 @@ buf_read_ibuf_merge_pages(
 		buf_read_page_low(&err, sync && (i + 1 == n_stored),
 				  BUF_READ_ANY_PAGE, space_ids[i],
 				  zip_size, TRUE, space_versions[i],
-				  page_nos[i], NULL);
+				  page_nos[i], NULL, false);
 
 		if (UNIV_UNLIKELY(err == DB_TABLESPACE_DELETED)) {
 tablespace_deleted:
@@ -1005,12 +1013,13 @@ not_to_recover:
 		if ((i + 1 == n_stored) && sync) {
 			buf_read_page_low(&err, true, BUF_READ_ANY_PAGE, space,
 					  zip_size, TRUE, tablespace_version,
-					  page_nos[i], NULL);
+					  page_nos[i], NULL, false);
 		} else {
 			buf_read_page_low(&err, false, BUF_READ_ANY_PAGE
 					  | OS_AIO_SIMULATED_WAKE_LATER,
 					  space, zip_size, TRUE,
-					  tablespace_version, page_nos[i], NULL);
+					  tablespace_version, page_nos[i],
+					  NULL, false);
 		}
 	}
 
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -247,6 +247,16 @@ struct os_aio_array_t{
 				There is one such event for each
 				possible pending IO. The size of the
 				array is equal to n_slots. */
+	struct iocb**		pending;
+				/* Array to buffer the not-submitted aio
+				requests. The array length is n_slots.
+				It is divided into n_segments segments.
+				pending requests on each segment are buffered
+				separately.*/
+	ulint*			count;
+				/* Array of length n_segments. Each element
+				counts the number of not-submitted aio
+				request on that segment.*/
 #endif /* LINUX_NATIV_AIO */
 };
 
@@ -4084,6 +4094,13 @@ os_aio_array_create(
 	memset(io_event, 0x0, sizeof(*io_event) * n);
 	array->aio_events = io_event;
 
+	array->pending = static_cast<struct iocb**>(
+		ut_malloc(n * sizeof(struct iocb*)));
+	memset(array->pending, 0x0, sizeof(struct iocb*) * n);
+	array->count = static_cast<ulint*>(
+		ut_malloc(n_segments * sizeof(ulint)));
+	memset(array->count, 0x0, sizeof(ulint) * n_segments);
+
 skip_native_aio:
 #endif /* LINUX_NATIVE_AIO */
 	for (ulint i = 0; i < n; i++) {
@@ -4140,6 +4157,16 @@ os_aio_array_free(
 	if (srv_use_native_aio) {
 		ut_free(array->aio_events);
 		ut_free(array->aio_ctx);
+
+#ifdef UNIV_DEBUG
+		for (size_t idx = 0; idx < array->n_slots; ++idx)
+			ut_ad(array->pending[idx] == NULL);
+		for (size_t idx = 0; idx < array->n_segments; ++idx)
+			ut_ad(array->count[idx] == 0);
+#endif
+
+		ut_free(array->pending);
+		ut_free(array->count);
 	}
 #endif /* LINUX_NATIVE_AIO */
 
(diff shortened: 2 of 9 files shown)


percona-ysorokin:
Jenkins build links
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1982/
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1984/

storage/innobase/buf/buf0rea.cc
```diff
@@ -788,6 +795,10 @@ buf_read_ahead_linear(
 			}
 		}
 	}
+#if defined(LINUX_NATIVE_AIO)
+	/* Tell aio to submit all buffered requests. */
```
laurynas-biveinis: Useless comment (retells the called function name)

storage/innobase/os/os0file.cc
```diff
@@ -4771,31 +4790,100 @@ readahead requests. */
...
+					   / array->n_segments;
+			int submitted;
+			submitted = io_submit(array->aio_ctx[i], count,
+					      &(array->pending[iocb_index]));
+			if (submitted == count) {
+				total_submitted += submitted;
+			} else {
+				/* io_submit returns number of successfully
+				queued requests or -errno. */
+				fprintf(stderr, "Trying to sumbit %d aio "
```
laurynas-biveinis: Replace fprintf(stderr with InnoDB diagnostic API

storage/innobase/os/os0file.cc
```diff
@@ -4771,31 +4790,100 @@ readahead requests. */
...
+					"requests, io_submit returns %d.\n",
+					count, submitted);
+				if (submitted < 0) {
+					errno = -submitted;
```
laurynas-biveinis: 1) Is this the actual libaio contract?
2) If it is and we get the right errno value here, what consumes it?

storage/innobase/os/os0file.cc
```diff
@@ -4771,31 +4790,100 @@ readahead requests. */
...
+					rather than submit. */
 {
 	int		ret;
-	ulint		io_ctx_index;
+	ulint		io_ctx_index = 0;
 	struct iocb*	iocb;
+	ulint slots_per_segment;
 
-	ut_ad(slot != NULL);
+	ut_ad(slot);
```
laurynas-biveinis: Spurious diff

storage/innobase/os/os0file.cc
```diff
@@ -4771,31 +4790,100 @@ readahead requests. */
...
 	ut_ad(array);
-
```
laurynas-biveinis: Spurious diff

storage/innobase/os/os0file.cc
```diff
@@ -4771,31 +4790,100 @@ readahead requests. */
...
-	os_aio_slot_t*	slot)	/*!< in: an already reserved slot. */
+	os_aio_slot_t*	slot,	/*!< in: an already reserved slot. */
+	bool		should_buffer)	/*!< in: should buffer the request
+					rather than submit. */
 {
 	int		ret;
-	ulint		io_ctx_index;
+	ulint		io_ctx_index = 0;
 	struct iocb*	iocb;
+	ulint slots_per_segment;
```
laurynas-biveinis: Move declaration to initialization

storage/innobase/os/os0file.cc
```diff
@@ -4771,31 +4790,100 @@ readahead requests. */
...
 /*==================*/
 	os_aio_array_t*	array,	/*!< in: io request array. */
-	os_aio_slot_t*	slot)	/*!< in: an already reserved slot. */
+	os_aio_slot_t*	slot,	/*!< in: an already reserved slot. */
+	bool		should_buffer)	/*!< in: should buffer the request
+					rather than submit. */
 {
 	int		ret;
-	ulint		io_ctx_index;
+	ulint		io_ctx_index = 0;
```
laurynas-biveinis: Since you touch this, might as well move declaration to initialization

storage/innobase/os/os0file.cc
```diff
@@ -4771,31 +4790,100 @@ readahead requests. */
...
 	/* Find out what we are going to work with.
 	The iocb struct is directly in the slot.
 	The io_context is one per segment. */
 
+	slots_per_segment = array->n_slots / array->n_segments;
 	iocb = &slot->control;
-	io_ctx_index = (slot->pos * array->n_segments) / array->n_slots;
-
+	io_ctx_index = slot->pos / slots_per_segment;
+	if (should_buffer && array == os_aio_read_array) {
```
laurynas-biveinis: Since all the callers know exactly what I/O request they ask or not ask to be buffered, add an assert that array == os_aio_read_array if should_buffer

storage/innobase/os/os0file.cc
```diff
@@ -4771,31 +4790,100 @@ readahead requests. */
...
+		ulint n;
```
laurynas-biveinis: Declarations at first use

storage/innobase/os/os0file.cc
```diff
@@ -4771,31 +4790,100 @@ readahead requests. */
...
+		/* There are array->n_slots elements in array->pending,
+		which is divided into array->n_segments area of equal size.
+		The iocb of each segment are buffered in its corresponding area
+		in the pending array consecutively as they come.
+		array->count[i] records the number of buffered aio requests
+		in the ith segment.*/
+		n = io_ctx_index * slots_per_segment
+			  + array->count[io_ctx_index];
+		array->pending[n] = iocb;
+		array->count[io_ctx_index] ++;
```
laurynas-biveinis: Spurious space before increment

laurynas-biveinis: Debug-assert count[io_ctx_index] is not zero post increment (or not == slots_per_segment pre increment, see a comment below)

storage/innobase/os/os0file.cc
```diff
@@ -4771,31 +4790,100 @@ readahead requests. */
...
+		count = array->count[io_ctx_index];
+		os_mutex_exit(array->mutex);
+		if (count == slots_per_segment) {
```
laurynas-biveinis: I am not 100% certain how slots_per_segment (specifically, array->n_segments) is used, but this reeks of a race condition. Thread 1: submit a request, count becomes == slots_per_segment, release array mutex. Thread 2: submit a request, make count go over slots_per_segment, then ???

storage/innobase/os/os0file.cc
```diff
@@ -4140,6 +4157,8 @@ os_aio_array_free(
 	if (srv_use_native_aio) {
 		ut_free(array->aio_events);
 		ut_free(array->aio_ctx);
+		ut_free(array->pending);
+		ut_free(array->count);
```
laurynas-biveinis: Debug-build assert that arrays (all of them; not just the read one) are empty on shutdown.

storage/innobase/os/os0file.cc
```diff
@@ -4771,31 +4790,100 @@ readahead requests. */
 
 #if defined(LINUX_NATIVE_AIO)
 /*******************************************************************//**
+Submit buffered AIO requests on the given segment to the kernel. */
+UNIV_INTERN
+void
+os_aio_linux_dispatch_read_array_submit()
```
laurynas-biveinis: The submission API (should_buffer arg everywhere) is platform-independent, so the "flush" API should not be either, please remove "linux" from name, push #ifdef inside, remove #ifdefs from callers

laurynas-biveinis (changes requested):
Please replace "LRA part 1" with an actual new (or maybe existing; check) blueprint describing this change ("merging of AIO requests")

storage/innobase/buf/buf0rea.cc
percona-ysorokin: removed

storage/innobase/os/os0file.cc
percona-ysorokin: done

storage/innobase/os/os0file.cc
percona-ysorokin: done

storage/innobase/os/os0file.cc
percona-ysorokin: 1. Yes, it is true  for libaio. See 'NOTES' section in http://man7.org/linux/man-pages/man2/io_submit.2.html
2. Changed so that `errno` is printed to `ib_log`

storage/innobase/os/os0file.cc
percona-ysorokin: done

storage/innobase/os/os0file.cc
percona-ysorokin: done

storage/innobase/os/os0file.cc
percona-ysorokin: fixed

storage/innobase/os/os0file.cc
percona-ysorokin: fixed

storage/innobase/os/os0file.cc
percona-ysorokin: done

storage/innobase/os/os0file.cc
percona-ysorokin: fixed

storage/innobase/os/os0file.cc
percona-ysorokin: Good catch.

Reworked by introducing `os_aio_dispatch_read_array_submit_low(bool acquire_mutex)` and putting `if (count == slots_per_segment)` under mutex protection.
```c++
os_mutex_enter(array->mutex);
...
if (count == slots_per_segment) {
	os_aio_dispatch_read_array_submit_low(false);
}
os_mutex_exit(array->mutex);
```

storage/innobase/os/os0file.cc
percona-ysorokin: Done

storage/innobase/os/os0file.cc
laurynas-biveinis: This is not necessarily an active code in upstream (FB not WebScaleSQL) nor do we have a smoking-gun testcase, but a quick "JFYI" bug report might be useful IMHO

percona-ysorokin: Reported
https://github.com/facebook/mysql-5.6/issues/700

storage/innobase/os/os0file.cc
percona-ysorokin: Done

percona-ysorokin:
Jenkins build link
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1990/

storage/innobase/os/os0file.cc
```diff
@@ -4769,6 +4796,75 @@ readahead requests. */
...
+@param acquire_mutex specifies whether to lock array mutex
+*/
+static
+void
+os_aio_dispatch_read_array_submit_low(bool acquire_mutex MY_ATTRIBUTE((unused)))
+{
+#if defined(LINUX_NATIVE_AIO)
+	os_aio_array_t*	array = os_aio_read_array;
+	ulint		total_submitted = 0;
+	if (!srv_use_native_aio) {
```
laurynas-biveinis: Pull up above #ifdef LINUX_NATIVE_AIO ?

storage/innobase/os/os0file.cc
```diff
@@ -4769,6 +4796,75 @@ readahead requests. */
...
+			int	submitted = io_submit(
+					array->aio_ctx[i], count,
+					&(array->pending[iocb_index]));
+			if (submitted == count) {
+				total_submitted += submitted;
+			} else {
+				/* io_submit returns number of successfully
+				queued requests or -errno. */
+				if (submitted >= 0) {
+					ib_logf(IB_LOG_LEVEL_ERROR,
```
laurynas-biveinis: Is it possible for io_submit to return partial success value < count? Looking at its interface, I don't see how, either -errno, either complete success, leaving no means to restart a partial submission. If this is correct, add a DBUG_ASSERT(0)?

storage/innobase/os/os0file.cc
```diff
@@ -4769,6 +4796,75 @@ readahead requests. */
...
+						"Trying to sumbit %d aio "
+						"requests, io_submit() "
+						"returns %d",
+						count, submitted);
+				} else {
+					const char*	errmsg =
+						strerror(-submitted);
+					ib_logf(IB_LOG_LEVEL_ERROR,
```
laurynas-biveinis: IB_LOG_LEVEL_FATAL (sic?). The ut_error below can still be kept for safety

laurynas-biveinis (changes requested):

storage/innobase/os/os0file.cc
percona-ysorokin: From https://www.fsl.cs.sunysb.edu/~vass/linux-aio.txt
```
io_submit()'s return code can be one of the following values:

A) ret = (number of iocbs sumbmitted)
	Ideal case, all iocbs were accepted for processing.

B) 0 < ret <  (number of iocbs sumbmitted)
	io_submit() system call processes iocbs one by one starting from
	the first entry in the passed array. If submission of some iocb fails,
	it stops at this point and returns the index of iocb that failed.
	There is no way to know what is the exact reason of a failure.
	However, if the very first iocb submission fails, see point C.

C) ret < 0
	There are two reasons why this could happen:
	1) Some error happened even before io_submit() started to iterate
	   over iocbs in the array (e.g., AIO context was invalid). 
	2) The submission of the very first iocb (cbx[0]) failed).
```
So, potentially, when `0 < ret <  (number of iocbs sumbmitted)` we can wrap this in a loop. Terminate when we did not make any progress (`ret == 0` then  fatal error) or  `ret ==  (number of iocbs sumbmitted)`

storage/innobase/os/os0file.cc
percona-ysorokin: done

storage/innobase/os/os0file.cc
percona-ysorokin: done

storage/innobase/os/os0file.cc
laurynas-biveinis: OK, ideally this has to be loop. Please double-check ret == 0 case, from my reading, it's impossible unless submitted array was also zero-length, this could be asserted I guess.

percona-ysorokin: Yes, you are right. From the comments and the source code itself, it looks like `ret == 0` only when `nr == 0`. If `io_submit()` fails at the very first `iocb` it returns actual `errno` (propagated from `io_submit_one()`)
https://github.com/torvalds/linux/blob/master/fs/aio.c#L1700

percona-ysorokin:
Jenkins build link
http://jenkins.percona.com/job/percona-server-5.6-param/1991/

laurynas-biveinis (approved):
(as usual please have 5.7 ready before pushing 5.6)

PR #1908, opened 2017-08-29
Ps 5.7 myr 149

george-lorch:


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1174/

storage/rocksdb/rdb_io_watchdog.cc
```diff
@@ -0,0 +1,233 @@
...
+  // Verify the write access to all directories we care about.
+  for (const std::string &directory : m_dirs_to_check) {
+    ret = check_write_access(directory);
+
+    // We'll log a warning and attept to continue to see if the problem happens
+    // in other cases as well.
+    if (unlikely(ret != HA_EXIT_SUCCESS)) {
+      // NO_LINT_DEBUG
+      sql_print_warning("Unable to verify write access to %s (error code %d).",
+                        directory.c_str(), ret);
```
laurynas-biveinis: It would be more useful to report errno/errstr here (actual reason for a failure) rather than ret (-1 most of the time)

storage/rocksdb/rdb_io_watchdog.cc
```diff
@@ -0,0 +1,233 @@
...
+  int fd = open(fname.c_str(), O_WRONLY | O_DIRECT | O_CREAT | O_SYNC,
+                S_IRWXU | S_IWUSR);
+
+  if (unlikely(fd == -1)) {
+    return fd;
+  }
+
+  int ret = write(fd, m_buf, RDB_IO_WRITE_BUFFER_SIZE);
+
+  if (unlikely(ret != RDB_IO_WRITE_BUFFER_SIZE)) {
```
laurynas-biveinis: This is broken with write setting errno to EINTR, and returning zero bytes written, because HA_EXIT_SUCCESS is also defined to zero.

storage/rocksdb/rdb_io_watchdog.cc
```diff
@@ -0,0 +1,233 @@
...
+  timer_spec.it_value.tv_sec = m_write_timeout;
+  timer_spec.it_interval.tv_sec = m_write_timeout;
+
+  ret = timer_settime(m_io_check_timer, 0, &timer_spec, nullptr);
+
+  if (unlikely(ret)) {
+    // NO_LINT_DEBUG
+    sql_print_warning("Setting time for a watchdog I/O timer failed with %d.",
+                      errno);
+  }
```
laurynas-biveinis: Proceeding to return HA_EXIT_SUCCESS here seems incorrect

storage/rocksdb/rdb_io_watchdog.h
```diff
@@ -0,0 +1,113 @@
...
+        static_cast<Rdb_io_watchdog *>(timer_data.sival_ptr);
+    DBUG_ASSERT(io_watchdog != nullptr);
+
+    io_watchdog->expire_io_callback(timer_data);
+  }
+
+ public:
+  explicit Rdb_io_watchdog(const std::vector<std::string> &directories)
+      : m_io_check_timer(nullptr), m_io_check_watchdog_timer(nullptr),
+        m_io_in_progress(false), m_dirs_to_check(std::move(directories)),
```
laurynas-biveinis: Do const ref args and std::move mesh well? The suggested pattern here seems to be pass by value and move std::move one level up, to the constructor caller

laurynas-biveinis (changes requested):

storage/rocksdb/rdb_io_watchdog.h
george-lorch: This is all upstream goodness. I already made several changes to this commit to clean it up and am hesitant to go much further as it will make future auto-merging nearly impossible for the area touched. I _almost_ did not take this commit at all as I saw very little value added for the amount of code it changed and introduced, but, I did not want the same auto-merge issues that would result from the missing commit.

I did not notice the issues with the error handling and that is probably something worth fixing and notifying upstream of the issues or just rolling this commit back out of the merge and taking it in manually in order to correct the issues. I'll have to look closer, review the remaining upstream changes coming and decide whether to keep and fix, just keep as-is, or reject it outright and forward your commentary upstream.

laurynas-biveinis: All options should result in almost the same cumulative amount of work ultimately

george-lorch: Perhaps, but some options allow me to progress through the merging of useful changes and get to fixing real, GA blocking issues while others allow me to punt upstream to fix while accepting some possible short term merge pain.

storage/rocksdb/rdb_io_watchdog.cc
george-lorch: I don't think so, if EINTR occurs, write is supposed to return -1, not 0.

storage/rocksdb/rdb_io_watchdog.cc
george-lorch: Looking at the use of this function, I see no reason why it can/should even return a value as the return from it is never checked nor any action taken. It seems that the error handling for it should always be to just sql_print_warning and return;

storage/rocksdb/rdb_io_watchdog.h
george-lorch: This specific case is interesting, I am surprised that it even compiles as it seems implied that a parameter to std::move must be mutable otherwise no actual move is possible. I suspect there is an implementation detail in std::move that provides a 'const' constructor as well that does no move but a raw copy, which eliminates the entire purpose of even involving std::move here in the first place.

The one place where this is explicitly called, 'directories' is a temporary that is destroyed via scope exit right after an instance of Rdb_io_watchdog is instantiated. It seems to me sending in a list of directories to monitor is construction overkill and the interface could be better defined instead by an 'add_directory' followed by a contract change that reset_timeout must be called after any directory changes have been made.

george-lorch: Hmm, that would also produce a poor interface as once the m_directories is 'set' it must be immutable as other state within the implementation depends on the vector remaining unchanged. In fact, due to this, m_directories should be declared as const. Looking at an implementation of std::move, it appears that there are 'const' specifications that allow its use and perform a copy/addref behavior rather than an actual move. So, while its use here is not beneficial, it is also not harmful either.

george-lorch:
OK, so I backed out my changes to the original commit for Rdb_io_watchdog and incorporated them along with some of the other things discussed into a stand alone commit that is part of this merge series. The individual commit can be seen here https://github.com/percona/percona-server/pull/1908/commits/e98493a15fd82b22a0ba4ff8619b8f328405afc2

Most of the changes are fairly minor but address most of the problems. Once we agree on the changes needed for the Rdb_io_watchdog, I will do a got format-patch and send my commit upstream. Their current trunk has not changed in this area since this commit was introduced.

Here is a jenkins run with the new changes http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1181/

storage/rocksdb/rdb_io_watchdog.cc
laurynas-biveinis: You are right re. EINTR. What's broken here is write returning success having written zero bytes. I don't know if it ever does that, but I'm unable to find why this would be a write contract violation. What's also broken - and, as opposed to zero bytes written with success case, actually happens - is writing more than zero but less than requested. Then the diagnostics one level up will be "Unable to verify write access to blah, errno 0"

laurynas-biveinis: The absolutely proper fix is as usual wrapping write in an EINTR-, and returned bytes-checking loop in its own function, and every non-trivial program/lib should have one already. MySQL mysys has my_write for this, maybe MyRocks has another one too already?

storage/rocksdb/rdb_io_watchdog.h
laurynas-biveinis: The interface to take directories in the constructor is OK, now the question is how to pass them. 1) The interface could accept only rvalue ref, then std::move in it would be efficient, the only question is again whether that's a good interface. Since it's all very locally used, it could be good enough
2) The interface could not worry about the cost and take the classic const ref, drop the useless std::move then, optimize things later only if the cost is seen to be worth optimizing (probably it won't be). Looks like the current state of the branch.
3) The idomatic way - assuming that std::move on vectors is cheap which AFAIK it is - is to take arg by value and have std::move. When passed an lvalue, the arg is copy constructed, then moved. We have extra move compared to 2), that's OKish (see the "cheap move" requirement above). When passed an rvalue, the arg is move constructed, then moved with std::move. We have an extra move compared to 1), again that's OKish.

laurynas-biveinis:
LGTM if going with the constructor interface option that matches the current branch (const lvalue ref for which useless std::move is removed). If you want to have nit addressed, that bit is LGTM in advance.

storage/rocksdb/rdb_io_watchdog.cc
george-lorch: I was considering going the whole EINTR route for this similar to InnoDB and PerconaFT, but both implementations span layers and are quite messy for what amounts to a tiny, and fairly unimportant check in this unimportant class to see if there is write access to a file. MyRocks has no i/o layer wrapping because MyRocks itself, like TokuDB, does no i/o directly on its own. All i/o is done through the lower level dependent code (rocksdb, perconaft) and those do not expose their low level i/o APIs through their high level interfaces. So, anything and everything related to i/o or OS primitives that does not use the my_* has to be reinvented here at the storage engine layer. 

The only negative side effect that I can see in this thing from not fully and properly handling all cases where write does not write exactly the data expected is that we will have false sql_print_warnings done as a result.

I reiterate, I almost didn't even take this patch as I see it as a completely useless chunk of code to exist and run within the storage engine. It may be interesting as an experimental performance monitor or debugging tool, but useless in a 'soon to be production' engine. Just my $.02 worth

laurynas-biveinis: I see sloppy code, I point it out. Up to you to decide where to cut losses.

george-lorch: Ohh no doubt, and I totally appreciate it. I just wish that it didn't exist and need pointing out in the first place.

george-lorch: Hmm, looks like it causes issues w/ OSX now too https://lists.launchpad.net/maria-developers/msg10851.html

george-lorch: You know what, I really don't like what this patch is doing. My initial assertion that the only 'downside' is excessive sql_print_warnings, but that is not true, it aborts the server if there is an i/o timeout. I am going to remove the entirely. I do not think this behavior is at all appropriate and this is just not worth the effort to fix.

laurynas-biveinis: Fine with me, just please update upstream with both the high level issues and the code changes so far. Re. macOS, it's OK not to care since it's not our supported platform, and someone with a Mac can always fix it later. Depending on what is the exact issue, the MariaDB fix is probably better reworked as a CMake check anyway.

PR #1909, opened 2017-08-29
TDB-76 : CREATE TABLE ... LIKE ... does not use source row_format on …

george-lorch:
…target table

- Implemented fix in TDB to take row_format if it is not the same as DEFAULT.
- Added new SQL warning if a row_format value can not be translated to a valid
  TokuDB row_format type.
- Enhanced the tokudb.row_format test to cover these cases.

### DIFF
--- a/storage/tokudb/ha_tokudb.cc
+++ b/storage/tokudb/ha_tokudb.cc
@@ -7227,10 +7227,28 @@ int ha_tokudb::create(
     const tokudb::sysvars::format_t row_format =
         (tokudb::sysvars::row_format_t)form->s->option_struct->row_format;
 #else
-    const tokudb::sysvars::row_format_t row_format =
-        (create_info->used_fields & HA_CREATE_USED_ROW_FORMAT)
-        ? row_type_to_row_format(create_info->row_type)
-        : tokudb::sysvars::row_format(thd);
+    // TDB-76 : CREATE TABLE ... LIKE ... does not use source row_format on
+    //          target table
+    // Original code would only use create_info->row_type if
+    // create_info->used_fields & HA_CREATE_USED_ROW_FORMAT was true. This
+    // would cause us to skip transferring the row_format for a table created
+    // via CREATE TABLE tn LIKE tn. We also take on more InnoDB like behavior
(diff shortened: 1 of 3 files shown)


george-lorch:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1986/

storage/tokudb/ha_tokudb.cc
```diff
@@ -7227,10 +7227,28 @@ int ha_tokudb::create(
     const tokudb::sysvars::format_t row_format =
         (tokudb::sysvars::row_format_t)form->s->option_struct->row_format;
 #else
-    const tokudb::sysvars::row_format_t row_format =
-        (create_info->used_fields & HA_CREATE_USED_ROW_FORMAT)
-        ? row_type_to_row_format(create_info->row_type)
-        : tokudb::sysvars::row_format(thd);
+    // TDB-76 : CREATE TABLE ... LIKE ... does not use source row_format on
```
laurynas-biveinis, line 7230: I don't understand why a comment describing previous buggy version of this code is required?

george-lorch: Lack of history in case it breaks something else? Part of the problem I have with so much of the TokuDB and PerconaFT code and issue trackers is there is no history to understand why things were done the way they were. I suppose the commit message alone might be enough but to me this is exactly the kind of thing that will break something else  later on for some odd reason and there will be no reminder or historic reason documented anywhere why this change was made. The commit message/blame may be enough, assuming that we can keep it forever, but it seemed like a reasonably unclear enough thing to understand the logic and reason why the new code is the way it is to justify a few lines explaining the change.

laurynas-biveinis: Source repo history has also crossed my mind. Anyway, both versions LGTM if that's OK with you

george-lorch: In all fairness, this has been a bug for over 6 years and PZ just found it this weekend, so, how many more times do we expect to have to revisit these 6-8 lines of code :-\

PR #1910, opened 2017-08-29
Ps 5.7 tdb 76

george-lorch:


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1175/

laurynas-biveinis (commented):
Same as for 5.6

PR #1912, opened 2017-08-30
MYR-151 : json data type overwrites data in table on insert

george-lorch:
MYR-152 : the json binary value contains invalid data
- MyRocks was entirely missing JSON support, which seems to be to just treat,
  parck, and unpack MYSQL_TYPE_JSON exacely like MYSQL_TYPE_BLOB.
- Added changes to support JSON types.
- Added new test that includes some tests from mysql-test/suite/json/include
  and also does some basic JSON store/retrieve.


george-lorch:
The json test feels quite hackish but I could not find any other decent example, so I included some of the appropriate tests from mysql-test/suite/json/inc and appended the test cases that Tomislav was using. Incidentally, there is absolutely no TokuDB test for JSON in 5.7 at all.

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1182/

mysql-test/suite/rocksdb/t/type_json.test
```diff
@@ -0,0 +1,21 @@
+--source include/have_rocksdb.inc
+
+# needed by theses json .inc tests
```
laurynas-biveinis: "theses"

laurynas-biveinis (approved):
OK with typo fixed. Please see that docs mark JSON support as Beta or similar even when the rest of MyRocks is GA

george-lorch:
Actually that is a good point re: beta JSON. I did ask Tomislav to test JSON pretty thoroughly but this really is probably worth creating an RQA task to 'clear' it before we call it a GA feature.

PR #1919, opened 2017-09-05
Ps 5.7 myr 167 168 170 181

george-lorch:


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1188/

I stacked these in one PR since they all are unique but impact the same general space of functionality and that is validating CREATE TABLE options that ha_rocksdb::create doesn't support. It also allowed just one jenkins run with the changes rather then three individual runs.

mysql-test/suite/rocksdb/t/create_table.test
```diff
@@ -189,3 +190,41 @@ CREATE TABLE t1(c1 INT,c2 INT,c3 INT,c4 INT,c5 INT,c6 INT,c7 INT,c8 INT,c9 INT,
 SHOW CREATE TABLE t1;
 
 DROP TABLE t1;
+
```
laurynas-biveinis, line 193: Do we need "End of 5.6 tests", "end of 5.7 tests" etc. markers in the testcase for easier future testcase maintenance WRT 5.7 -> 8.0 etc upmerges?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5732,6 +5732,24 @@ int ha_rocksdb::create(const char *const name, TABLE *const table_arg,
     DBUG_RETURN(HA_ERR_ROCKSDB_TABLE_INDEX_DIRECTORY_NOT_SUPPORTED);
   }
 
+  if (create_info->encrypt_type.length) {
```
laurynas-biveinis: "unlikely" branch annotations?

laurynas-biveinis (approved):
LGTM unless you consider the minor comments actionable

mysql-test/suite/rocksdb/t/create_table.test
george-lorch: Hmm, maybe not a bad idea if it helps the forward merging.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Good catch, MyRocks currently doesn't make a lot of use of branch prediction so it has been sort of an "out of site, out of mind" thing for me, but that doesn't mean we shouldn't use it in our changes where it makes sense. I'll add some in for these and retest.

laurynas-biveinis: If you wish, you could also check if my_error itself is annotated as a cold function (and not invoked on some hot path with "no error")

george-lorch:
Touched up as discussed, new jenkins http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1198/

laurynas-biveinis (approved):

PR #1921, opened 2017-09-11
Bug: PS-247 - keyring_vault mtr tests should be skipped if Vault server

robgolebiowski:
is unavailable

Added check file is_vault_server_up.inc. This script checks if
Hashicorp Vault server that is to be used in a test is available. In
case it is not the test will be skipped. This test file was included in
all keyring_vault's MTR tests.

### DIFF
--- a/plugin/keyring_vault/tests/mtr/is_vault_server_up.inc
+++ b/plugin/keyring_vault/tests/mtr/is_vault_server_up.inc
@@ -1,0 +1,94 @@
+# In order to check whether Vault server is accessible we try to get list
+# of server backends. This list should be short.
+
+let KEYRING_CONF_FILE=$KEYRING_CONF_FILE_1;
+let SERVER_UUID= query_get_value(SELECT @@SERVER_UUID, @@SERVER_UUID, 1);
+if (!$CURL_TIMEOUT)
+{
+  --let CURL_TIMEOUT=4
+}
+--perl
+  use strict;
+  use IO::File;
+  my $curl_timeout= $ENV{CURL_TIMEOUT} or die "Need CURL_TIMEOUT";
+  my $keyring_conf_file= $ENV{'KEYRING_CONF_FILE'} or die("KEYRING_CONF_FILE not set\n");
+  my $server_uuid= $ENV{'SERVER_UUID'} or die("SERVER_UUID not set\n");
+  my $token;
+  my $vault_url;
+  my $secret_mount_point;
+  my $vault_ca;
+  my $CONF_FILE;
+  open(CONF_FILE, "$keyring_conf_file") or die("Could not open configuration file.\n");
+  while (my $row = <CONF_FILE>)
+  {
+    if ($row =~ m/token[ ]*=[ ]*(.*)/)
+    {
+      $token=$1;
+    }
+    elsif ($row =~ m/vault_url[ ]*=[ ]*(.*)/)
+    {
+      $vault_url=$1;
+    }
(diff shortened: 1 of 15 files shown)


percona-ysorokin (commented):
Just a suggestion: it may be better to make an attempt to establish a connection to `<vault_host>:443` rather than ping the host because our IT guys could decide to filter out ICMP traffic one day and they will do this silently.

plugin/keyring_vault/tests/mtr/is_vault_server_up.inc
```diff
@@ -0,0 +1,54 @@
...
+  $F->close();
+  $p->close();
+EOF
+
+--source $MYSQLTEST_VARDIR/tmp/ping_result.inc
+--remove_file $MYSQLTEST_VARDIR/tmp/ping_result.inc
+
+if(!$PING_SUCCESSFUL)
+{
+  --skip Seems that Hashicorp Vault is down.
```
laurynas-biveinis: ... testing server ...

laurynas-biveinis (changes requested):
+1 to Yura's comment if feasible to implement

robgolebiowski:
It is not really possible to try establishing connection with Vault Server. As it uses REST protocol I can only send https requests that Vault server will understand. I.e. I can list secrets, create or remove secret. Listing secrets is not an option as that would mean retrieving list of all keys from the server - which can be time consuming and just too much for simple check. Also creating and removing a test secret seems like an overkill. I could try using tokens backed to retrieve information about current token that I am using - but this mean upgrading token's policy to the level that is not needed. Token used in test should only be able to create/remove/list secrets from the backend they have access to. I really doubt they will change ICMP routing.

robgolebiowski:
Rebased on the latest 5.7

plugin/keyring_vault/tests/mtr/is_vault_server_up.inc
robgolebiowski: I am not sure I understand this comment. Should --skip state something like --skip While testing server turned out that Hashicorp Vault is down ?

laurynas-biveinis: "Seems that Hashicorp Vault testing server is down"

laurynas-biveinis:
If this has been updated and ready for review, then a new Jenkins run for the affected testcases is needed

robgolebiowski:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1318/

plugin/keyring_vault/tests/mtr/is_vault_server_up.inc
laurynas-biveinis: The comment not addressed but too minor to hold the merge on its own

laurynas-biveinis (approved):

percona-ysorokin (approved):
LGTM

Good as an initial version. However, still believe that just establishing a TCP connection to `$vault_url` port 443 (even without SSL handshake and sending any HTTP request) could be a good improvement.

robgolebiowski:
I have changed the implementation and addressed (I hope ;) Yura's comment. Please have another look, if necessary I will create separate PR.

laurynas-biveinis (approved):

percona-ysorokin (approved):
LGTM

robgolebiowski:
JR => https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1395/

laurynas-biveinis (approved):

PR #1924, opened 2017-09-12
Bug PS-248: keyring_udf mtr tests should be skipped if keyring_udf

robgolebiowski:
plugin is unavailable

Added a check file have_keyring_udf_plugin.inc that makes a test
skip itself in case keyring_udf plugin is not available. This check
file was added to the begging of all keyring_udf mtr tests


percona-ysorokin (approved):
LGTM

laurynas-biveinis (approved):

PR #1928, opened 2017-09-15
MYR-186 : Fix for MYR-168 introduced mtr regression

george-lorch:
- Corrected test that makes use of tablespaces where MyRocks absorbed the
  invalid CREATE TABLE specifiers but now fails due to errors generated as a
  result of the fix for MYR-168. This was missed in testing as this test,
  rocksdb.rocksdb_parts is still unstable and fails occasionally elsewhere.


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1208/

laurynas-biveinis (approved):

PR #1929, opened 2017-09-15
Fix bug 1711781 (Redundant GTID unsafe mark for CREATE/DROP TEMPORARY…

laurynas-biveinis:
… TABLE in RBR/MBR)

After fixing bug 1668602, bug 1539504, and bug 1313901, CREATE
TEMPORARY TABLE is only logged under statement binary log mode, and
DROP TEMPORARY TABLE is only logged if the corresponding CREATE
TEMPORARY TABLE has been logged. However, a corresponding
enforce_gtid_consistency check in THD::is_ddl_gtid_compatible has not
been relaxed accordingly. This resulted that CREATE/DROP TEMPORARY
TABLE statements were forbidden incorrectly in transactional contexts,
including function and trigger calls, even when they required no
binary logging at all.

Fix by keeping only the CREATE TEMPORARY TABLE check in
THD::is_ddl_gtid_compatible and lifting its restriction for row/mixed
binary log modes. For DROP TEMPORARY TABLE, since its check requires
knowing the binlog format at the corresponding CREATE TEMPORARY TABLE,
move it to mysql_rm_table.

Convert binlog.binlog_enforce_gtid_consistency test to an include file
that is shared between two new binlog format-dependent tests
binlog_stm_enforce_gtid_consistency and
binlog_row_mix_enforce_gtid_consistency.

http://jenkins.percona.com/job/percona-server-5.6-param/1988/


percona-ysorokin (approved):
LGTM

robgolebiowski (approved):
LGTM

PR #1931, opened 2017-09-15
MYR-179 : Basic Sysbench run fails with basic MyROCKS install due to …

george-lorch:
…lack of open files

- Changed default value to 1000 from infinite.
- Added detection if rocksdb_max_open_files is > open_files_limit to reduce it to open_files_limit.
- Extended rocksdb.sys_vars.rocksdb_max_open_files_basic.test to validate new behavior for minimum, maximum, and unlimited.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -798,8 +798,7 @@ static MYSQL_SYSVAR_BOOL(
 static MYSQL_SYSVAR_INT(max_open_files, rocksdb_db_options->max_open_files,
                         PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_READONLY,
                         "DBOptions::max_open_files for RocksDB", nullptr,
-                        nullptr, rocksdb_db_options->max_open_files,
-                        /* min */ -1, /* max */ INT_MAX, 0);
+                        nullptr, 1000, /* min */ -1, /* max */ INT_MAX, 0);
 
 static MYSQL_SYSVAR_ULONG(max_total_wal_size,
                           rocksdb_db_options->max_total_wal_size,
@@ -3438,6 +3437,14 @@ static int rocksdb_init_func(void *const p) {
                         HTON_SUPPORTS_EXTENDED_KEYS | HTON_CAN_RECREATE;
 
   DBUG_ASSERT(!mysqld_embedded);
+
+  if (rocksdb_db_options->max_open_files > (long) open_files_limit) {
+    sql_print_information("RocksDB: rocksdb_max_open_files should not be "
+                          "greater than the open_files_limit, effective value "
+                          "of rocksdb_max_open_files is being set to "
+                          "open_files_limit.");
+    rocksdb_db_options->max_open_files = open_files_limit;
+  }
 
   rocksdb_stats = rocksdb::CreateDBStatistics();
   rocksdb_db_options->statistics = rocksdb_stats;
--- a/mysql-test/suite/rocksdb.sys_vars/r/rocksdb_max_open_files_basic.result
+++ b/mysql-test/suite/rocksdb.sys_vars/r/rocksdb_max_open_files_basic.result
@@ -1,7 +1,7 @@
 SET @start_global_value = @@global.ROCKSDB_MAX_OPEN_FILES;
 SELECT @start_global_value;
 @start_global_value
--1
+1000
 "Trying to set variable @@global.ROCKSDB_MAX_OPEN_FILES to 444. It should fail because it is readonly."
 SET @@global.ROCKSDB_MAX_OPEN_FILES   = 444;
 ERROR HY000: Variable 'rocksdb_max_open_files' is a read only variable
--- a/mysql-test/suite/rocksdb/r/max_open_files.result
+++ b/mysql-test/suite/rocksdb/r/max_open_files.result
@@ -1,0 +1,15 @@
+CALL mtr.add_suppression("RocksDB: rocksdb_max_open_files should not be greater than the open_files_limit*");
+# restart:--log-error=MYSQLTEST_VARDIR/tmp/rocksdb.max_open_files.err --rocksdb_max_open_files=over_rocksdb_max_open_files
+# restart:--rocksdb_max_open_files=0
+SELECT @@global.rocksdb_max_open_files;
+@@global.rocksdb_max_open_files
+0
+CREATE TABLE t1(a INT) ENGINE=ROCKSDB;
+INSERT INTO t1 VALUES(0),(1),(2),(3),(4);
+SET GLOBAL rocksdb_force_flush_memtable_and_lzero_now=1;
+DROP TABLE t1;
+# restart:--rocksdb_max_open_files=-1
+SELECT @@global.rocksdb_max_open_files;
+@@global.rocksdb_max_open_files
+-1
+# restart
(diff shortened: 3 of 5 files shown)


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1207/

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3337,6 +3336,17 @@ static int rocksdb_init_func(void *const p) {
 
   DBUG_ASSERT(!mysqld_embedded);
 
+  if (rocksdb_db_options->max_open_files < 100 &&
```
laurynas-biveinis: Why?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3337,6 +3336,17 @@ static int rocksdb_init_func(void *const p) {
...
+      rocksdb_db_options->max_open_files > 0) {
+    rocksdb_db_options->max_open_files = 100;
+  }
+
+  if (rocksdb_db_options->max_open_files > (long) open_files_limit) {
+    sql_print_information("RocksDB: rocksdb_max_open_files should not be "
+                          "greater than the open_files_limit.");
+    rocksdb_db_options->max_open_files = 100;
```
laurynas-biveinis: This will not work if open_files_limit < 100

laurynas-biveinis (changes requested):
Having read through MYR-179 this PR does not quite implement what was your last plan there. If the goal of this PR is simply to tweak the default, then OK, but why not s/100/1000 then

storage/rocksdb/ha_rocksdb.cc
george-lorch: Correct, but if open_files_limit is < 100 we probably should just abort because no reasonable instance would ever function at that level. The end result will be the same problem that initiated the bug report in the first place, an assertion down low in rocksdb because an open failed.

george-lorch:
Correct, this does not implement exactly what was discussed in the issue as after some playing and testing IMHO that is not really a practical solution. It also makes testing and some test cases impossible to make deterministic. Instead, I followed the InnoDB paradigm somewhat where it tries to set a minimum rational floor and ceiling based on the users setting and the actual open_files_limit (which if not explicitly set by the user seems is set to the process ulimit). All I am trying to do is to change the default from the former value of the cache size from "unlimited", which will nearly always cause the reported assertion over time, to some reasonable low default that will allow an instance to run indefinitely without running out of descriptors (assuming ulimit is a rational value).

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: I am not sure: with open_files_limit == 99 my guesstimate is that the instance would support ~30 InnoDB tables,  maybe ~45 with innodb_file_per_table=0, each arbitrarily large as long as unpartitioned. I could imagine a workable production instance with a single table too. Another surprising thing I have learned is that numerous tiny MySQL instances as backend for application correctness testing are an actually existing thing (e.g. bump buffer pool size minimum by a trivial amount, get a bug report that the test machine OOMs). In any case the code-level bug should be fixed, and if the minimum should be indeed bumped to 100 regardless of the above, that should be done as a server-level patch.

george-lorch: "Tiny instance" and the use of MyRocks are somewhat mutually exclusive. 100 was just an arbitrary choice, InnoDB starts at 300 if innodb_open_files < 10, then, it sets to table_cache_size if fpt=1 or if it is > open_files_limit.

The reason that I did not just set the variable definition minimum to 100 (or whatever abstract amount) is that -1 is a valid value and used to specify 'unlimited'. Unlike InnoDB, I also did not use table_cache_size for anything as there is absolutely no relationship between the number of database.table.indexes and the number of files in the RocksDB set. I believe RocksDB would actually function with a low-ish limit but would be opening and closing SST files for nearly every operation. I think though it might be possible for compaction to require many concurrently open files.

laurynas-biveinis: IMHO "tiny MyRocks instance" would still make sense in that application backend testing scenario because MyRocks is functionally different from other SEs.

Re. minimum 100 as the server-level patch, I was referring to open_files_limit. But I have not fully thought it through whether it's a good idea or not.

george-lorch:
@laurynas-biveinis, finally coming back to this, so where are we at? What do you suggest we change about this to get a reasonable default and enforcable minimum? Should we change the enforcable minimum to some lower value like 10? Should the default be higher than 100? Maybe 300? Or should it be some calculated value based on open_files_limit? (This will be somewhat problematic then at keeping some mtr tests deterministic). Anyway, I would like to reach consensus and have something merged before we release again.

laurynas-biveinis:
I would like to see "This will not work if open_files_limit < 100" code-level bug fixed regardless of any other decision re. variables, and what would be your thought about setting open_files_limit minimum to 100 in the server? Not by the variable allowed value range, but, since its value is adjusted on the startup, ensuring a minimum 100 there? As for rocksdb_max_open_files, I'd make the default value the one that makes all reasonable workloads work without requiring its tweaking, but I don't know what that value would be. It looks safe to me to err on the high side, so 1000?

george-lorch:
OK, so I removed the floor limit of 100 and allow all the way down to -1. I manually tested that MyRocks still seems operational at 0, 1, 2 (just create a table, insert a bunch of records, force compaction, etc...). I also changes the rocksdb_max_open_files > open_files_limit to just set rocksdb_max_open_files = open_files_limit. I bumped the default up to 1000. Here is a new jenkins run with these changes http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1266

mysql-test/suite/rocksdb.sys_vars/t/rocksdb_max_open_files_basic.test
```diff
@@ -4,3 +4,35 @@
 --let $read_only=1
 --let $session=0
 --source ../include/rocksdb_sys_var.inc
+
+
+# MYR-179 : Basic Sysbench run fails with basic MyROCKS install due to lack of open files
```
laurynas-biveinis: OK but in this other PR I just reviewed a similar test went to a separate test in the main suite.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3439,6 +3438,12 @@ static int rocksdb_init_func(void *const p) {
 
   DBUG_ASSERT(!mysqld_embedded);
 
+  if (rocksdb_db_options->max_open_files > (long) open_files_limit) {
+    sql_print_information("RocksDB: rocksdb_max_open_files should not be "
```
laurynas-biveinis, line 3442: Does it need to say that rocksdb_max_open_files effective value is being adjusted to open_files_limit ?

mysql-test/suite/rocksdb.sys_vars/t/rocksdb_max_open_files_basic.test
george-lorch: Yeah, you're right. I actually did it in the other work as a result of discussing it elsewhere, which was probably here. I'll split this out so we stay somewhat consistent.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Yeah, that would be more clear. IIRC I copied some of this from InnoDB and probably altered it in trying to fit it into here.

george-lorch:
Split the test case out and touched it up a bit considering the changes to limits. Added a little bit that illustrated that MyRocks can still function to some minimal extent with rocksdb_max_open_files=0
New jenkins http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1272/

laurynas-biveinis (approved):

PR #1932, opened 2017-09-15
MYR-184 : More MyRocks metrics for PMM

george-lorch:
- Added missing statiistics to global status.


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1210/

laurynas-biveinis (approved):

PR #1933, opened 2017-09-19
InnoDB temporary files and temporary tables encryption

gl-sergei:
InnoDB is using temporary files for merge sort for online alter table.
It also creates temporary tables for online alter.

When destination table for online alter operation is encrypted merge
file will also be encrypted. InnoDB writing and reading merge files in
blocks. Each block encrypted with tablespace key of target tablespace.
Key version used to encrypt this block prepended to each block. Since
Percona Server does not support key rotation currently, key version is
always 0. Block encrypted with AES256 in CBC mode with IV consisting of
space_id+offset. IV is encrypted to make it non-predictable. Since CBC
can only encrypt data in multiples of block size, remainder of the
buffer is XOR'ed with encrypted IV.

When InnoDB needs to create temporary table for online alter operation
on encrypted table, temp table will be encrypted as well.

https://blueprints.launchpad.net/percona-server/+spec/innodb-temp-files

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1211/


include/my_aes.h
```diff
@@ -131,6 +131,46 @@ int my_aes_get_size(uint32 source_length, enum my_aes_opmode mode);
...
+  original one.
+
+  @param source         [in]  Pointer to data for encryption
+  @param source_length  [in]  Size of encryption data
+  @param dest           [out] Buffer to place encrypted data (same size as src)
+  @param key            [in]  Key to be used for encryption
+  @param key_length     [in]  Length of the key. Will handle keys of any length
+  @param mode           [in]  encryption mode
+  @param iv             [in]  16 bytes initialization vector.
+  @param padding        [in]  if padding needed.
```
percona-ysorokin: no such parameter

include/my_aes.h
```diff
@@ -131,6 +131,46 @@ int my_aes_get_size(uint32 source_length, enum my_aes_opmode mode);
...
+  original one.
+
+  @param source         Pointer to data for decryption
+  @param source_length  size of encrypted data
+  @param dest           buffer to place decrypted data (same size as src)
+  @param key            Key to be used for decryption
+  @param key_length     Length of the key. Will handle keys of any length
+  @param mode           encryption mode
+  @param iv             16 bytes initialization vector.
+  @param padding        if padding needed.
```
percona-ysorokin: no such parameter

mysql-test/t/innodb_onlinealter_encryption-master.opt
```diff
@@ -0,0 +1 @@
+$KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_LOAD --keyring_file_data=$MYSQL_TMP_DIR/mysecret_keyring
```
percona-ysorokin: no eol

mysys_ssl/my_aes.cc
```diff
@@ -57,3 +57,97 @@ void my_aes_create_key(const unsigned char *key, uint key_length,
...
   }
 }
+
+int my_aes_encrypt_nopad(const unsigned char *source, uint32 source_length,
+                         unsigned char *dest,
+                         const unsigned char *key, uint32 key_length,
+                         enum my_aes_opmode mode, const unsigned char *iv)
+{
+  unsigned char eiv[MY_AES_BLOCK_SIZE];
+  uint32 num_blocks= (source_length / MY_AES_BLOCK_SIZE) * MY_AES_BLOCK_SIZE;
```
percona-ysorokin: misleading name, I would suggest something like `aligned_source_len`

mysys_ssl/my_aes.cc
```diff
@@ -57,3 +57,97 @@ void my_aes_create_key(const unsigned char *key, uint key_length,
...
+                          key, key_length, mode, eiv, false);
+  if (dst_len == MY_AES_BAD_DATA)
+    return dst_len;
+
+  if (num_blocks < source_length)
+  {
+    unsigned char mask[MY_AES_BLOCK_SIZE];
+    uint32 len= source_length - num_blocks;
+
+    DBUG_ASSERT(dst_len == static_cast<int>(num_blocks));
```
percona-ysorokin: should be outside of `if` block?

mysys_ssl/my_aes.cc
```diff
@@ -57,3 +57,97 @@ void my_aes_create_key(const unsigned char *key, uint key_length,
     *ptr^= *sptr;
   }
 }
+
+int my_aes_encrypt_nopad(const unsigned char *source, uint32 source_length,
```
percona-ysorokin: Can we implement this by using one of the AES block modes which do not require input message to be a multiple of AES_BLOCK_SIZE?
CFB, OFB or CTR for instance.

mysys_ssl/my_aes.cc
```diff
@@ -57,3 +57,97 @@ void my_aes_create_key(const unsigned char *key, uint key_length,
...
+
+    DBUG_ASSERT(dst_len == static_cast<int>(num_blocks));
+
+    dst_len= my_aes_encrypt(eiv, MY_AES_BLOCK_SIZE, mask,
+                            key, key_length, ecb_mode, NULL, false);
+
+    if (dst_len == MY_AES_BAD_DATA)
+      return dst_len;
+
+    for (uint32 i= 0; i < len; i++)
```
percona-ysorokin: Is this "xoring with encrypted eiv" a well-known approach or a clever workaround? If yes, please give a reference.

mysys_ssl/my_aes.cc
```diff
@@ -57,3 +57,97 @@ void my_aes_create_key(const unsigned char *key, uint key_length,
...
+  return source_length;
+}
+
+int my_aes_decrypt_nopad(const unsigned char *source, uint32 source_length,
+                         unsigned char *dest,
+                         const unsigned char *key, uint32 key_length,
+                         enum my_aes_opmode mode, const unsigned char *iv)
+{
+  unsigned char eiv[MY_AES_BLOCK_SIZE];
+  uint32 num_blocks= (source_length / MY_AES_BLOCK_SIZE) * MY_AES_BLOCK_SIZE;
```
percona-ysorokin: as in `_encrypt_`: misleading name

mysys_ssl/my_aes.cc
gl-sergei: `my_aes.h` only provides ECB and CBC for yassl. We build PS with OpenSSL, but I think we still need to be compatible. InnoDB page encryption using CBC. I would prefer CTR of course. MariaDB provides choice between CTR and CBC (CTR only available for file_key_management plugin though) with CBC being default. Trick with last partial block is taken from MariaDB.

mysys_ssl/my_aes.cc
gl-sergei: It actually very similar to how CTR mode works. It takes counter, encrypts it and XOR with plaintext, which allows to parallelize encryption.

mysys_ssl/my_aes.cc
gl-sergei: Outside of the block it can be false actually. `DBUG_ASSERT(dst_len == static_cast<int>(source_length));` should be true outside of the `if` block :)

gl-sergei: Hmm... I got and idea.. Yes, it can be placed *before* `if` the block.

gl-sergei:
Fixed comment, added newline, renamed variable and moved DBUG_ASSERT few lines upward.

mysql-test/t/innodb_onlinealter_encryption.test
```diff
@@ -0,0 +1,74 @@
+-- source include/have_innodb.inc
+-- source include/not_valgrind.inc
```
laurynas-biveinis: Why?

mysql-test/t/innodb_onlinealter_encryption.test
```diff
@@ -0,0 +1,74 @@
...
+-- source include/not_embedded.inc
+-- source include/not_windows.inc
```
laurynas-biveinis: Why?

mysql-test/t/innodb_onlinealter_encryption.test
```diff
@@ -0,0 +1,74 @@
...
+
+CREATE TABLE t1 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB ENCRYPTION='y';
+CREATE TABLE t2 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB;
+CREATE TABLE t3 (id INT, a VARCHAR(255)) ENGINE=InnoDB ENCRYPTION='y';
+CREATE TABLE t4 (id INT, a VARCHAR(255)) engine=InnoDB;
+CREATE TABLE t5 (id INT NOT NULL PRIMARY KEY, a TEXT(500), b VARCHAR(255), FULLTEXT(b)) ENGINE=InnoDB ENCRYPTION='y';
+CREATE TABLE t6 (id INT, a TEXT(500), b VARCHAR(255), FULLTEXT(b)) ENGINE=InnoDB;
+CREATE TABLE t7 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB row_format=compressed ENCRYPTION='y';
```
laurynas-biveinis: Consistent upper case for row_format clause

mysql-test/t/innodb_onlinealter_encryption.test
```diff
@@ -0,0 +1,74 @@
+-- source include/have_innodb.inc
+-- source include/not_valgrind.inc
+-- source include/not_embedded.inc
+-- source include/not_windows.inc
+
+CREATE TABLE t1 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB ENCRYPTION='y';
+CREATE TABLE t2 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB;
+CREATE TABLE t3 (id INT, a VARCHAR(255)) ENGINE=InnoDB ENCRYPTION='y';
+CREATE TABLE t4 (id INT, a VARCHAR(255)) engine=InnoDB;
```
laurynas-biveinis: Consistent upper case engine=

mysql-test/t/innodb_onlinealter_encryption.test
```diff
@@ -0,0 +1,74 @@
...
+end//
+delimiter ;//
+commit;
+
+set autocommit=0;
+call innodb_insert_proc(15000);
+commit;
+set autocommit=1;
+
+ALTER TABLE t1 ADD COLUMN b int default 2;
```
laurynas-biveinis: Here and everywhere consistent uppercase (ideally the innodb_insert_proc definition and call above too)

mysql-test/t/innodb_onlinealter_encryption.test
```diff
@@ -0,0 +1,74 @@
+-- source include/have_innodb.inc
```
laurynas-biveinis: IIRC there is now some have_keyring_*.inc too, please check and source

mysql-test/t/innodb_onlinealter_encryption.test
```diff
@@ -0,0 +1,74 @@
...
+SHOW CREATE TABLE t1;
+SHOW CREATE TABLE t2;
+SHOW CREATE TABLE t3;
+SHOW CREATE TABLE t4;
+SHOW CREATE TABLE t5;
+SHOW CREATE TABLE t6;
+SHOW CREATE TABLE t7;
+
+--echo # Restarting server
+let $restart_parameters = restart: $KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_LOAD --keyring_file_data=$MYSQL_TMP_DIR/mysecret_keyring
```
laurynas-biveinis: The arguments seem to match the -master.opt file exactly. Are they still required to be specified here in that case?

mysys_ssl/my_aes.cc
laurynas-biveinis: It's OK to provide degraded functionality / missing features for YaSSL builds if that makes the OpenSSL implementation better.

storage/innobase/handler/handler0alter.cc
```diff
@@ -5569,6 +5570,31 @@ ha_innobase::prepare_inplace_alter_table(
 		goto err_exit_no_heap;
 	}
 
+	/* create_table_info_t::innobase_table_flags does not set encryption
+	flags. There are places where it is done afterwards, there are places
+	where it isn't done. We need to inspect all code paths and check if
+	encryption flag cen be set in one place. */
```
laurynas-biveinis: "can"

storage/innobase/row/row0ftsort.cc
```diff
@@ -195,6 +196,7 @@ row_fts_psort_info_init(
 	fts_psort_t*		merge_info = NULL;
 	ulint			block_size;
 	ibool			ret = TRUE;
+	bool			encrypted = false;
```
laurynas-biveinis: Declare at initialization, as "const"?

storage/innobase/row/row0ftsort.cc
```diff
@@ -267,6 +271,31 @@ row_fts_psort_info_init(
...
+
+			/* If tablespace is encrypted, allocate additional
+			buffer for encryption/decryption. */
+			if (encrypted) {
+
+				/* Need to align memory for O_DIRECT write */
+				psort_info[j].crypt_alloc[i] =
+					static_cast<row_merge_block_t*>(
+						ut_malloc_nokey(
+							block_size + 1024));
```
laurynas-biveinis, line 280: IIRC we might need 4K for some O_DIRECT setups, please check - but if this is a copy-paste from a different temp file allocation doing same I/O, then OK to leave as-is

storage/innobase/row/row0ftsort.cc
```diff
@@ -318,6 +347,10 @@ row_fts_psort_info_destroy(
 
 				ut_free(psort_info[j].block_alloc[i]);
 				ut_free(psort_info[j].merge_file[i]);
+
+				if (psort_info[j].crypt_alloc[i]) {
+					ut_free(psort_info[j].crypt_alloc[i]);
```
laurynas-biveinis: Does ut_free DTRT with NULL arg, allowing to remove "if" here?

storage/innobase/row/row0ftsort.cc
```diff
@@ -733,6 +766,7 @@ fts_parallel_tokenization(
 	ibool			processed = FALSE;
 	merge_file_t**		merge_file;
 	row_merge_block_t**	block;
+	row_merge_block_t**	crypt_block;
```
laurynas-biveinis: Declare at initialization?

storage/innobase/row/row0ftsort.cc
```diff
@@ -1517,6 +1561,7 @@ row_fts_merge_insert(
 	mrec_buf_t**		buf;
 	int*			fd;
 	byte**			block;
+	byte**			crypt_block;
```
laurynas-biveinis: Declare at initialization?

storage/innobase/row/row0merge.cc
```diff
@@ -259,12 +261,115 @@ class index_tuple_info_t {
...
+Encrypt a merge block. */
+static
+void
+row_merge_encrypt_buf(
+/*==================*/
+	ulint			offset,		/*!< in: offset where to
+						write */
+	ulint			space_id,	/*!< in: tablespace id */
+	const byte*		input_buf,	/*!< in: input buffer */
+	byte*			crypted_buf)	/*!< out: crypted buffer */
```
laurynas-biveinis: "encrypted"

storage/innobase/row/row0merge.cc
```diff
@@ -259,12 +261,115 @@ class index_tuple_info_t {
...
+{
+	const ulint key_version = 0;
+	int dstlen;
+	os_offset_t ofs = (os_offset_t)srv_sort_buf_size * (os_offset_t)offset;
```
laurynas-biveinis: What is the allowed maximum offset value here? Is it possible to overflow os_offset_t?

storage/innobase/row/row0merge.cc
```diff
@@ -259,12 +261,115 @@ class index_tuple_info_t {
...
+
+	/* Store key_version at the begining of the input buffer. InnoDB
+	does not support tablespace key rotation for now. */
```
laurynas-biveinis: This is not a "tablespace" is it?

storage/innobase/row/row0merge.cc
```diff
@@ -259,12 +261,115 @@ class index_tuple_info_t {
...
+		input_buf + ROW_MERGE_RESERVE_SIZE,
+		srv_sort_buf_size - ROW_MERGE_RESERVE_SIZE,
+		crypted_buf + ROW_MERGE_RESERVE_SIZE,
+		space->encryption_key,
+		space->encryption_klen,
+		my_aes_256_cbc,
+		iv);
+
+	if (dstlen == MY_AES_BAD_DATA) {
+		ib::fatal()
```
laurynas-biveinis: fatal? Maybe only the ALTER TABLE should be failed?

storage/innobase/row/row0merge.cc
```diff
@@ -259,12 +261,115 @@ class index_tuple_info_t {
...
+Decrypt a merge block. */
+static
+bool
+row_merge_decrypt_buf(
+/*==================*/
+	ulint			offset,		/*!< in: offset where to
+						write */
+	ulint			space_id,	/*!< in: tablespace id */
+	const byte*		input_buf,	/*!< in: input buffer */
+	byte*			crypted_buf)	/*!< out: crypted buffer */
```
laurynas-biveinis: "decrypted"

storage/innobase/row/row0merge.cc
```diff
@@ -259,12 +261,115 @@ class index_tuple_info_t {
...
+{
+	ulint key_version;
+	int dstlen;
+	os_offset_t ofs = (os_offset_t)srv_sort_buf_size * (os_offset_t)offset;
```
laurynas-biveinis: What is the allowed maximum offset value here? Is it possible to overflow os_offset_t?

storage/innobase/row/row0merge.cc
```diff
@@ -259,12 +261,115 @@ class index_tuple_info_t {
...
+
+	/* Read key_version from begining of the buffer */
+	key_version = mach_read_from_4((byte *)input_buf);
+
+	/* InnoDB does not support tablespace key rotation for now */
+	ut_a(key_version == 0);
```
laurynas-biveinis: Can we just fail the ALTER TABLE instead?

storage/innobase/row/row0merge.cc
```diff
@@ -259,12 +261,115 @@ class index_tuple_info_t {
...
+		input_buf + ROW_MERGE_RESERVE_SIZE,
+		srv_sort_buf_size - ROW_MERGE_RESERVE_SIZE,
+		crypted_buf + ROW_MERGE_RESERVE_SIZE,
+		space->encryption_key,
+		space->encryption_klen,
+		my_aes_256_cbc,
+		iv);
+
+	if (dstlen == MY_AES_BAD_DATA) {
+		ib::fatal()
```
laurynas-biveinis: Same comment re. fatal as above

storage/innobase/row/row0merge.cc
```diff
@@ -259,12 +261,115 @@ class index_tuple_info_t {
...
+			<< "Unable to decrypt data-block "
+			" src: " << static_cast<const void*>(input_buf)
+			<< " srclen: " << srv_sort_buf_size
+			<< " buf: " << static_cast<const void*>(crypted_buf)
+			<< " buflen: " << dstlen
+			<< ". Can't continue!";
+	}
+
+	return true;
```
laurynas-biveinis: It only returns true - either errors should be returned if possible to implement above review comments, either make it void. The caller only handles true return correctly.

storage/innobase/row/row0merge.cc
```diff
@@ -1172,6 +1307,10 @@ row_merge_read_rec(
 
 	DBUG_ENTER("row_merge_read_rec");
 
+	if (b == &block[0]) {
+		b+= ROW_MERGE_RESERVE_SIZE;
```
laurynas-biveinis: Space before +=

storage/innobase/row/row0merge.cc
```diff
@@ -1377,6 +1521,10 @@ row_merge_write_rec(
 	size = extra_size + (extra_size >= 0x80)
 		+ rec_offs_data_size(offsets);
 
+	if (b == &block[0]) {
+		b+= ROW_MERGE_RESERVE_SIZE;
```
laurynas-biveinis: Space before +=

storage/innobase/row/row0merge.cc
```diff
@@ -1434,6 +1590,10 @@ row_merge_write_eof(
 		    reinterpret_cast<const void*>(block),
 		    fd, ulong(*foffs)));
 
+	if (b == &block[0]) {
+		b+= ROW_MERGE_RESERVE_SIZE;
```
laurynas-biveinis: Space before +=

storage/innobase/row/row0merge.cc
```diff
@@ -4336,7 +4546,9 @@ row_merge_build_indexes(
 {
 	merge_file_t*		merge_files;
 	row_merge_block_t*	block;
+	row_merge_block_t*	crypt_block;
```
laurynas-biveinis: Declare at initialization

storage/innobase/row/row0merge.cc
```diff
@@ -4369,6 +4581,21 @@ row_merge_build_indexes(
...
 
+	/* Get crypt data from tablespace if present. */
+	crypt_block = NULL;
+
+	/* If tablespace is encrypted, allocate additional buffer for
+	encryption/decryption. */
+	if (dict_table_is_encrypted(new_table)) {
+		crypt_block = static_cast<row_merge_block_t*>(
+			alloc.allocate_large(
+				3 * srv_sort_buf_size, &crypt_pfx));
```
laurynas-biveinis: This might be ridiculously defensive but please consider (the upstream might rewrite the same and we wouldn't notice): extract srv_sort_buf_size sysvar max value to a define, add a static assert that 3 * max does not overflow os_offset_t

storage/innobase/row/row0merge.cc
```diff
@@ -4462,6 +4689,10 @@ row_merge_build_indexes(
 	/* Now we have files containing index entries ready for
 	sorting and inserting. */
 
+	DBUG_EXECUTE_IF(
+		"ib_merge_wait_after_read",
```
laurynas-biveinis: Please don't :) Luckily, this does not seem to have any associated testcase (something not imported from MariaDB?). Can it be removed, or if a testcase import is missed, replaced with something else?

storage/innobase/row/row0merge.cc
```diff
@@ -4545,9 +4776,13 @@ row_merge_build_indexes(
 			row_merge_dup_t	dup = {
 				sort_idx, table, col_map, 0};
 
+			DBUG_EXECUTE_IF(
+				"ib_merge_wait_after_sort",
```
laurynas-biveinis: Likewise

laurynas-biveinis (changes requested):

mysys_ssl/my_aes.cc
```diff
@@ -57,3 +57,99 @@ void my_aes_create_key(const unsigned char *key, uint key_length,
...
+  unsigned char eiv[MY_AES_BLOCK_SIZE];
+  uint32 aligned_src_len= (source_length / MY_AES_BLOCK_SIZE)
+                          * MY_AES_BLOCK_SIZE;
+  enum my_aes_opmode ecb_mode;
+  int dst_len;
+
+  if (key_length == 64)
+    ecb_mode= my_aes_256_ecb;
+  else if (key_length == 32)
+    ecb_mode= my_aes_128_ecb;
```
robgolebiowski: This seems to be wrong. In AES 128 and 256 mean key sizes. Thus key length of 64 bytes means aes 512 and key length of 32 bytes means aes 256.

mysys_ssl/my_aes.cc
```diff
@@ -57,3 +57,99 @@ void my_aes_create_key(const unsigned char *key, uint key_length,
...
+                         const unsigned char *key, uint32 key_length,
+                         enum my_aes_opmode mode, const unsigned char *iv)
+{
+  unsigned char eiv[MY_AES_BLOCK_SIZE];
+  uint32 aligned_src_len= (source_length / MY_AES_BLOCK_SIZE)
+                          * MY_AES_BLOCK_SIZE;
+  enum my_aes_opmode ecb_mode;
+  int dst_len;
+
+  if (key_length == 64)
```
robgolebiowski: This seems to be wrong. In AES 128 and 256 mean key sizes. Thus key length of 64 bytes means aes 512 and key length of 32 bytes means aes 256.

gl-sergei:
Last week MariaDB 10.2.29 has been released with support for online alter logs encryption. Merge file encryption has changed significantly as well. I will review that changes and update this PR. This PR should be put on hold for now.

gl-sergei:
I pushed some changes. Implementation slightly changed. There is no more key version stored with each block. Instead single encryption key is used. See https://jira.mariadb.org/browse/MDEV-12634. Added variable `innodb_encrypt_online_alter_logs`. Online alter logs are encrypted as well.

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1302/

gl-sergei:
For AES modes, lets discuss our options. I suggest to use CTR when available otherwise (YaSSL?) revert to CBC.

storage/innobase/row/row0ftsort.cc
gl-sergei: yes. it is copy-paste from
``` c++
			/* Need to align memory for O_DIRECT write */
			psort_info[j].block_alloc[i] =
				static_cast<row_merge_block_t*>(ut_malloc_nokey(
					block_size + 1024));
```

mysql-test/t/innodb_onlinealter_encryption.test
gl-sergei: There is a `include/have_keyring_vault_plugin.inc`, do you suggest to add `have_keyring_file.inc` as well?

mysql-test/t/innodb_onlinealter_encryption.test
gl-sergei: cleaned up

mysql-test/t/innodb_onlinealter_encryption.test
gl-sergei: cleaned up

storage/innobase/row/row0ftsort.cc
gl-sergei: I think ut_free is OK with NULL args. Changed. Not sure about `dealloc_large`, so I left that as it was

storage/innobase/row/row0merge.cc
gl-sergei: Replaced fatal with error, not sure how to cover it with a test case though.

storage/innobase/row/row0merge.cc
gl-sergei: os_offset_t should be 64bit. But it is probably OK to overflow it here since it is just a part of IV.

storage/innobase/row/row0merge.cc
gl-sergei: Currently `sort_buffer_size` max value is 64 MiB, so it is very unlike to overflow. But if you insist, I will add static assert

storage/innobase/row/row0merge.cc
gl-sergei: Test cases were cleaned up and use existing upstream DBUG_SYNC points

mysys_ssl/my_aes.cc
gl-sergei: Good catch. Fixed

mysql-test/t/innodb_onlinealter_encryption.test
laurynas-biveinis: It looks like this testcase needs have_keyring_file.inc and not have_keyring_vault_plugin.inc ?

storage/innobase/row/row0merge.cc
laurynas-biveinis: I don't insist :)

mysql-test/t/innodb_onlinealter_encryption.test
gl-sergei: I mean we do have `have_keyring_vault_plugin.inc` in our test suite, but not `have_keyring_file.inc` yet. I will create it.

mysys_ssl/my_aes.cc
robgolebiowski: Guys, Mode of operations have nothing to do with block sizes. AES will always demand input to be multiple of 128 bits. If it is not multiple the padding is needed. Mode of operations (apart from ECB) are just the ways of getting random encryption. ECB is the only mode where no randomness is introduced into the output, thus each time when I encrypt text A with key K I will get B. ECB can be only used when text A is random - for instance INNODB is using it to encrypt innodb tablespace keys - and this is fine as keys are random. About which mode we should be using, 
there is a bit of discussion on this topic, but I think we should chose CBC. The reason being that in case of CTR we need to be sure to be always encrypting each input block with unique counter, the counter cannot repeat. In past gurus like Bruce Schneier (https://books.google.pl/books?id=nnqdQsrZnJgC&pg=PT141&lpg=PT141&dq=Bruce+Schneier+cbc+or+ctr&source=bl&ots=6GXUxllCq7&sig=ca_lQu9fkiu0Kf_qcBhhYkXVF2Q&hl=en&sa=X&ved=0ahUKEwiK-9Hy5c_XAhVKCpoKHXCZDOIQ6AEIXTAJ#v=onepage&q=Bruce%20Schneier%20cbc%20or%20ctr&f=false) would recommend CTR, but now they go with CBC. It safer this way with CTR it is quite possible one will repeat the counter.

include/my_aes.h
```diff
@@ -131,6 +131,44 @@ int my_aes_get_size(uint32 source_length, enum my_aes_opmode mode);
...
+  @param source_length  [in]  Size of encryption data
+  @param dest           [out] Buffer to place encrypted data (same size as src)
+  @param key            [in]  Key to be used for encryption
+  @param key_length     [in]  Length of the key. Will handle keys of any length
+  @param mode           [in]  encryption mode
+  @param iv             [in]  16 bytes initialization vector.
+  @return              size of encrypted data, or negative in case of error
+*/
+
+int my_aes_encrypt_nopad(const unsigned char *source, uint32 source_length,
```
robgolebiowski: In binlog_encryption PR I am introducing the same functionality. The difference is only that I am introducing almost the same thing as MariaDB as I need the "context" feature they introduced. So I will be pushing this into repo and this functionalities will overlap. With what I am pushing the same will be possible by setting NO_PAD flag. However, I think the refactoring will needed to be done by one who pushes last

robgolebiowski (changes requested):

mysql-test/suite/innodb/t/innodb_onlinealter_encryption-master.opt
```diff
@@ -0,0 +1,4 @@
+$KEYRING_PLUGIN_OPT
```
robgolebiowski: Tests that require keyring plugins shoud be extracted into include files. There should be a test in innodb testsuite that loads keyring_file and sources the test as include file and there should be a test in keyring_vault  that loads keyring_vault plugin and sources the include file. For exmple see table_encypt_2 test case.

robgolebiowski (changes requested):

mysys_ssl/my_aes.cc
gl-sergei: Streaming modes such as CTR or OFB do not require input to be multiple of block size. They actually perform AES operations on key and XOR plaintext with produced stream (somewhat simplified explanation :) ). So the resulting length is always the same as input length.

mysql-test/suite/innodb/r/innodb_onlinealter_encryption.result
```diff
@@ -0,0 +1,127 @@
+CREATE TABLE t1 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB ENCRYPTION='y';
```
laurynas-biveinis: online_alter in the testcase name

mysql-test/suite/innodb/r/innodb_rowlog.result
```diff
@@ -0,0 +1,41 @@
+CREATE TABLE t1(id INT NOT NULL PRIMARY KEY AUTO_INCREMENT, credit_card VARCHAR(200), private VARCHAR(50)) ENGINE=InnoDB ENCRYPTION='y';
```
laurynas-biveinis: row_log in the testcase name

mysql-test/suite/innodb/t/innodb_rowlog.test
```diff
@@ -0,0 +1,71 @@
...
+DROP TABLE t2;
+
+connection default;
+disconnect con2;
+
+#
+# Verify that both merge blocks and row log blocks are encryted and decrypted
+#
+
+SELECT variable_value > 0 FROM information_schema.global_status WHERE variable_name = 'innodb_num_pages_encrypted';
```
laurynas-biveinis: Can you query PFS instead? (add have_perfschema.inc too then)

mysql-test/suite/innodb/t/innodb_rowlog.test
```diff
@@ -0,0 +1,71 @@
...
+#
+# Now we create secondary index for credit_card column in parallel we create new rows
+# forcing alter table to wait so that row log is used.
+#
+SET DEBUG_SYNC= 'row_merge_after_scan SIGNAL opened WAIT_FOR rotated';
+send ALTER TABLE t1 ADD INDEX secret (credit_card), ALGORITHM=INPLACE, LOCK=NONE;
+#
+# Force key rotation and create second index for same table
+#
+connect (con2,localhost,root,,);
```
laurynas-biveinis: --source include/count_sessions.inc

mysql-test/suite/innodb/t/innodb_rowlog.test
```diff
@@ -0,0 +1,71 @@
...
+connection default;
+reap;
+disconnect con1;
+
+connection con2;
+reap;
+DROP TABLE t2;
+
+connection default;
+disconnect con2;
```
laurynas-biveinis: --source include/wait_until_count_sessions.inc somewhere below this line

mysys_ssl/my_aes.cc
```diff
@@ -57,3 +57,103 @@ void my_aes_create_key(const unsigned char *key, uint key_length,
...
+  int dst_len;
+
+  if (key_length == 32)
+    ecb_mode= my_aes_256_ecb;
+  else if (key_length == 16)
+    ecb_mode= my_aes_128_ecb;
+  else
+    return MY_AES_BAD_DATA;
+
+  my_aes_encrypt(iv, MY_AES_BLOCK_SIZE, eiv,
```
laurynas-biveinis: Check return value?

mysys_ssl/my_aes.cc
```diff
@@ -57,3 +57,103 @@ void my_aes_create_key(const unsigned char *key, uint key_length,
...
+  if (key_length == 32)
+    ecb_mode= my_aes_256_ecb;
+  else if (key_length == 16)
+    ecb_mode= my_aes_128_ecb;
+  else
+    return MY_AES_BAD_DATA;
+
+  DBUG_ASSERT(aligned_src_len <= source_length);
+
+  my_aes_encrypt(iv, MY_AES_BLOCK_SIZE, eiv,
```
laurynas-biveinis: Likewise?

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,138 @@ struct row_log_t {
 				log operation */
 };
 
+struct crypt_info_t {
+
+	/** Encryption algorithm */
+	Encryption::Type	encryption_type;
+
+	/** Encrypt key */
```
laurynas-biveinis: Encryption key

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,138 @@ struct row_log_t {
...
+	byte			encryption_key[ENCRYPTION_KEY_LEN];
+
+	/** Encrypt key length*/
```
laurynas-biveinis: Encryption key length

storage/innobase/srv/srv0srv.cc
```diff
@@ -2754,6 +2767,9 @@ DECLARE_THREAD(srv_master_thread)(
 		} else {
 			srv_master_do_idle_tasks();
 		}
+
+		/* Enable row log encryption if it is set */
+		log_tmp_enable_encryption_if_set();
```
laurynas-biveinis: The corresponding plugin variable is not dynamic - shouldn't this be one-time initialisation thing?

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,138 @@ struct row_log_t {
...
+	ulint			encryption_klen;
+};
+
+my_bool	srv_encrypt_online_alter_logs = FALSE;
+
+crypt_info_t crypt_info = crypt_info_t();
+
+/** Find out is temporary log files encrypted.
```
laurynas-biveinis: Find out if temporary log files are encrypted

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,138 @@ struct row_log_t {
 				log operation */
 };
 
+struct crypt_info_t {
+
+	/** Encryption algorithm */
+	Encryption::Type	encryption_type;
```
laurynas-biveinis, line 221: Essentially duplicates each other with srv_encrypt_online_alter_logs. Not sure whether it needs fixing, and how

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,138 @@ struct row_log_t {
...
+/** Find out is temporary log files encrypted.
+@return true if temporary log file should be encrypted, false if not */
+bool
+log_tmp_is_encrypted()
+{
+	return (crypt_info.encryption_type != Encryption::NONE);
+}
+
+/** Check the row log encryption is enabled or not.
+It will try to enable the row log encryption. */
```
laurynas-biveinis: It will enable, there is no "try" :)

storage/innobase/row/row0merge.cc
```diff
@@ -4370,6 +4471,21 @@ row_merge_build_indexes(
 		DBUG_RETURN(DB_OUT_OF_MEMORY);
 	}
 
+	/* Get crypt data from tablespace if present. */
```
laurynas-biveinis: Comment not clear, there is no "tablespace" here?

storage/innobase/row/row0merge.cc
```diff
@@ -4370,6 +4471,21 @@ row_merge_build_indexes(
...
+	row_merge_block_t* crypt_block = NULL;
+
+	/* If tablespace is encrypted, allocate additional buffer for
```
laurynas-biveinis: Likewise

laurynas-biveinis (changes requested):

mysys_ssl/my_aes.cc
robgolebiowski: Yeah, my bad. I forgot that also CTR and OFB can be used as stream ciphers (however they still use block cipher inside).

storage/innobase/row/row0log.cc
gl-sergei: Right now we have implemented only AES, but maybe in future there will be more algorithms to choose from?

gl-sergei:
Rebased on 5.7 HEAD and addressed comments. https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1385/

Once binlog encryption get merged, I will update tests as suggested

laurynas-biveinis:
@gl-sergei, binlog encryption has been merged. Does this PR need updating?

gl-sergei:
rebased on latest 5.7, updated test cases to run with keyring_file and keyring_vault, switched to use of my_crypt instead of my_aes

https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1595/

mysql-test/include/innodb_row_log_encryption.inc
```diff
@@ -0,0 +1,82 @@
+--source include/have_innodb.inc
+--source include/not_embedded.inc
+--source include/have_debug_sync.inc
+--source include/have_perfschema.inc
+
+--source include/count_sessions.inc
```
laurynas-biveinis: Not before the server restart

mysql-test/include/innodb_row_log_encryption.inc
```diff
@@ -0,0 +1,82 @@
...
+# forcing alter table to wait so that row log is used.
+#
+SET DEBUG_SYNC= 'row_merge_after_scan SIGNAL opened WAIT_FOR rotated';
+send ALTER TABLE t1 ADD INDEX secret (credit_card), ALGORITHM=INPLACE, LOCK=NONE;
+#
+# Force key rotation and create second index for same table
+#
+connect (con2,localhost,root,,);
+SET DEBUG_SYNC= 'now WAIT_FOR opened';
+# TODO: once we add key rotation and scrubbing make sure to add it to this test
```
laurynas-biveinis: Track this in JIRA on the main feature tickets

mysql-test/suite/perfschema/r/show_sanity.result
```diff
@@ -672,6 +674,10 @@ where (show_mode = "JUNK: GLOBAL-ONLY")
 and source = "I_S.SESSION_STATUS")
 order by show_mode, source, variable_name;
 SHOW_MODE	SOURCE	VARIABLE_NAME
+5.7	P_S.SESSION_STATUS	Innodb_encryption_n_merge_blocks_decrypted
```
laurynas-biveinis: Was show_sanity updated correctly? I.e. no need to mention the new vars in the testcase itself anywhere?

storage/innobase/include/row0log.h
```diff
@@ -255,6 +256,44 @@ row_log_estimate_work(
 	const dict_index_t*	index);
 #endif /* HAVE_PSI_STAGE_INTERFACE */
 
+bool
```
laurynas-biveinis: Header comment

storage/innobase/include/row0log.h
```diff
@@ -255,6 +256,44 @@ row_log_estimate_work(
...
+log_tmp_is_encrypted();
+
+void
```
laurynas-biveinis: Header comment

storage/innobase/include/row0log.h
laurynas-biveinis: MY_ATTRIBUTE((warn_unused_result))

storage/innobase/include/row0log.h
```diff
@@ -255,6 +256,44 @@ row_log_estimate_work(
...
+log_tmp_enable_encryption_if_set();
+
+/** Encrypt a temporary file block.
+@param[in]	src		block to encrypt or decrypt
+@param[in]	size		size of the block
+@param[out]	dst		destination block
+@param[in]	offs		offset to block
+@param[in]	space_id	tablespace id
+@return whether the operation succeeded */
+bool
```
laurynas-biveinis: MY_ATTRIBUTE((warn_unused_result)) (and this will show one unchecked call site)

storage/innobase/include/row0log.h
```diff
@@ -255,6 +256,44 @@ row_log_estimate_work(
...
+	ulint			space_id);
+
+/** Decrypt a temporary file block.
+@param[in]	src		block to encrypt or decrypt
+@param[in]	size		size of the block
+@param[out]	dst		destination block
+@param[in]	offs		offset to block
+@param[in]	space_id	tablespace id
+@return whether the operation succeeded */
+bool
```
laurynas-biveinis: MY_ATTRIBUTE((warn_unused_result))

storage/innobase/include/row0merge.h
```diff
@@ -362,12 +365,14 @@ row_merge_file_create(
...
-@param[in]	dup	descriptor of index being created
-@param[in,out]	file	file containing index entries
-@param[in,out]	block	3 buffers
-@param[in,out]	tmpfd	temporary file handle
-@param[in,out]	stage	performance schema accounting object, used by
+@param[in]	trx		transaction
+@param[in]	dup		descriptor of index being created
+@param[in,out]	file		file containing index entries
+@param[in,out]	block		3 buffers
+@param[in,out]	crypt_block	crypted file buffer
```
laurynas-biveinis: "encrypted". I am not sure what's up with MariaDB inventing verb "to crypt"

storage/innobase/log/log0log.cc
```diff
@@ -56,6 +56,7 @@ Created 12/9/1995 Heikki Tuuri
 #include "trx0roll.h"
 #include "srv0mon.h"
 #include "sync0sync.h"
+#include "row0log.h"
```
laurynas-biveinis: Sorry if I asked this before, but why this? There are new declarations in row0log.h but none appear to be moved from somewhere else so that otherwise untouched source file starts needing them

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,158 @@ struct row_log_t {
...
+
+	res = my_aes_crypt(MY_AES_CBC,
+		ENCRYPTION_FLAG_ENCRYPT | ENCRYPTION_FLAG_NOPAD,
+		src_block, size, dst_block, &dst_len,
+		crypt_info.encryption_key, crypt_info.encryption_klen,
+		iv, MY_AES_BLOCK_SIZE);
+
+	if (res != MY_AES_OK) {
+		ib::error()
+			<< "Unable to encrypt data-block "
```
laurynas-biveinis: No hyphen

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,158 @@ struct row_log_t {
...
+
+	res = my_aes_crypt(MY_AES_CBC,
+		ENCRYPTION_FLAG_DECRYPT | ENCRYPTION_FLAG_NOPAD,
+		src_block, size, dst_block, &dst_len,
+		crypt_info.encryption_key, crypt_info.encryption_klen,
+		iv, MY_AES_BLOCK_SIZE);
+
+	if (res != MY_AES_OK) {
+		ib::error()
+			<< "Unable to decrypt data-block "
```
laurynas-biveinis: Likewise

laurynas-biveinis (changes requested):
Please also rebase again to take 5.7.21

laurynas-biveinis:
Please also file a feature request on JIRA and reference its id in the commit message header

gl-sergei:
Rebased and fixed:
- count sessions in the test case
- header comments
- grammar errors
- removed row0log.h (probably leftover from previous implementation)
- added new status variables into corresponding tables instead of re-recordling show_sanoty.result

https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1607/

mysql-test/include/innodb_row_log_encryption.inc
laurynas-biveinis: Have you logged something in JIRA?

storage/innobase/row/row0log.cc
```diff
@@ -179,6 +182,7 @@ struct row_log_t {
 	dict_table_t*	table;	/*!< table that is being rebuilt,
 				or NULL when this is a secondary
 				index that is being created online */
+	dict_index_t*	index;	/*!< index to be build */
```
laurynas-biveinis: s/build/built

storage/innobase/row/row0log.cc
```diff
@@ -3202,10 +3417,23 @@ row_log_allocate(
...
 
 	dict_index_set_online_status(index, ONLINE_INDEX_CREATION);
 	index->online_log = log;
 
+	if (log_tmp_is_encrypted()) {
+		ulint size = srv_sort_buf_size;
+		log->crypt_head = static_cast<byte *>(os_mem_alloc_large(&size));
+		log->crypt_tail = static_cast<byte *>(os_mem_alloc_large(&size));
+
+		if (!log->crypt_head || !log->crypt_tail) {
```
laurynas-biveinis, line 3431: Doesn't log->crypt_head != NULL, log->crypt_tail == NULL result in a memory leak for the former?

storage/innobase/row/row0merge.cc
```diff
@@ -1637,6 +1690,7 @@ NULL if old_table == new_table
 ULINT_UNDEFINED if none is added
 @param[in,out]	sequence	autoinc sequence
 @param[in,out]	block		file buffer
+@param[in,out]	crypt_block	crypted file buffer
```
laurynas-biveinis: encrypted

storage/innobase/row/row0merge.cc
```diff
@@ -2708,13 +2776,15 @@ row_merge_read_clustered_index(
...
-@param[in]	file	file containing index entries
-@param[in,out]	block	3 buffers
-@param[in,out]	foffs0	offset of first source list in the file
-@param[in,out]	foffs1	offset of second source list in the file
-@param[in,out]	of	output file
-@param[in,out]	stage	performance schema accounting object, used by
+@param[in]	dup		descriptor of index being created
+@param[in]	file		file containing index entries
+@param[in,out]	block		3 buffers
+@param[in,out]	crypt_block	crypted file buffer
```
laurynas-biveinis: encrypted

storage/innobase/row/row0merge.cc
```diff
@@ -2809,17 +2885,21 @@ row_merge_blocks(
...
-@param[in]	index	index being created
-@param[in]	file	input file
-@param[in,out]	block	3 buffers
-@param[in,out]	foffs0	input file offset
-@param[in,out]	of	output file
-@param[in,out]	stage	performance schema accounting object, used by
+@param[in]	index		index being created
+@param[in]	file		input file
+@param[in,out]	block		3 buffers
+@param[in,out]	crypt_block	crypted file buffer
```
laurynas-biveinis: encrypted

storage/innobase/row/row0merge.cc
```diff
@@ -2894,6 +2980,8 @@ row_merge_blocks_copy(
 @param[in]	dup		descriptor of index being created
 @param[in,out]	file		file containing index entries
 @param[in,out]	block		3 buffers
+@param[in,out]	crypt_block	crypted file buffer
```
laurynas-biveinis: encrypted

laurynas-biveinis (changes requested):
Minor comments, I'd suggest to get @percona-ysorokin review before the next Jenkins test

laurynas-biveinis:
...and/or @robgolebiowski review.

Also please file a JIRA issue for this, reference in the commit message header

storage/innobase/row/row0log.cc
gl-sergei: `row_log_free` will deallocate both as I read

gl-sergei:
repushed with grammar fixes in comments

gl-sergei:
Tracked here: https://jira.percona.com/browse/PS-3819

laurynas-biveinis (approved):
The only thing remaining from my comments is to track "TODO" in JIRA

gl-sergei:
I created both, just in the wrong project, then moved to PS ... here is the second one  https://jira.percona.com/browse/PS-3820

mysql-test/suite/perfschema/r/show_sanity.result
```diff
@@ -412,6 +412,7 @@ SHOW_MODE	SOURCE	VARIABLE_NAME
 5.6	I_S.SESSION_VARIABLES	INNODB_COMPRESSED_COLUMNS_THRESHOLD
 5.6	I_S.SESSION_VARIABLES	INNODB_COMPRESSED_COLUMNS_ZIP_LEVEL
 5.6	I_S.SESSION_VARIABLES	INNODB_DEADLOCK_DETECT
+5.6	I_S.SESSION_VARIABLES	INNODB_ENCRYPT_ONLINE_ALTER_LOGS
```
percona-ysorokin: Put `INNODB_ENCRYPT_ONLINE_ALTER_LOGS` as an exception to `show_sdanity.test`

storage/innobase/include/row0log.h
```diff
@@ -255,6 +256,51 @@ row_log_estimate_work(
...
+log_tmp_is_encrypted();
+
+/** Check the row log encryption is enabled or not.
+It will enable the row log encryption. */
+void
+log_tmp_enable_encryption_if_set();
+
+/** Encrypt a temporary file block.
+@param[in]	src		block to encrypt or decrypt
```
percona-ysorokin: `block to encrypt`

storage/innobase/include/row0log.h
```diff
@@ -255,6 +256,51 @@ row_log_estimate_work(
...
+log_tmp_block_encrypt(
+/*==================*/
+	const byte*		src_block,
+	ulint			size,
+	byte*			dst_block,
+	os_offset_t		offs,
+	ulint			space_id);
+
+/** Decrypt a temporary file block.
+@param[in]	src		block to encrypt or decrypt
```
percona-ysorokin: `block to decrypt`

storage/innobase/row/row0ftsort.cc
```diff
@@ -267,6 +268,31 @@ row_fts_psort_info_init(
...
+						ut_malloc_nokey(
+							block_size + 1024));
+
+				psort_info[j].crypt_block[i] =
+					static_cast<row_merge_block_t*>(
+						ut_align(
+							psort_info[j]
+							.crypt_alloc[i], 1024));
+
+				if (!psort_info[j].crypt_block[i]) {
```
percona-ysorokin: Is it OK to `ut_align()` on `NULL`s? If not, may be move this check just after `ut_malloc_nokey()`?

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,158 @@ struct row_log_t {
...
+{
+	size_t dst_len;
+	byte iv[MY_AES_BLOCK_SIZE];
+
+	memset(iv, 0, MY_AES_BLOCK_SIZE);
+
+	mach_write_to_8(iv, space_id);
+	mach_write_to_8(iv + 8, offs);
+
+	int res = my_aes_crypt(MY_AES_ECB,
```
percona-ysorokin, line 281: Does `my_aes_crypt()` allow to do inplace encryption?

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,158 @@ struct row_log_t {
...
+		ENCRYPTION_FLAG_ENCRYPT | ENCRYPTION_FLAG_NOPAD,
+		iv, MY_AES_BLOCK_SIZE, iv, &dst_len,
```
percona-ysorokin, line 283: Does `dst_len` need to be initialized with max out buffer size?

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,158 @@ struct row_log_t {
...
+	if (res != MY_AES_OK) {
+		ib::error() << "Unable to encrypt IV";
+		return false;
+	}
+
+	ut_ad(dst_len == MY_AES_BLOCK_SIZE);
+
+	res = my_aes_crypt(MY_AES_CBC,
+		ENCRYPTION_FLAG_ENCRYPT | ENCRYPTION_FLAG_NOPAD,
+		src_block, size, dst_block, &dst_len,
```
percona-ysorokin, line 296: The same question re `dst_len`.

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,158 @@ struct row_log_t {
...
+	byte iv[MY_AES_BLOCK_SIZE];
+
+	memset(iv, 0, MY_AES_BLOCK_SIZE);
+
+	mach_write_to_8(iv, space_id);
+	mach_write_to_8(iv + 8, offs);
+
+	int res = my_aes_crypt(MY_AES_ECB,
+		ENCRYPTION_FLAG_ENCRYPT | ENCRYPTION_FLAG_NOPAD,
+		iv, MY_AES_BLOCK_SIZE, iv, &dst_len,
```
percona-ysorokin, line 340: The same questions re inplace encryption and `dst_len`.

storage/innobase/row/row0log.cc
```diff
@@ -205,6 +215,158 @@ struct row_log_t {
...
+	if (res != MY_AES_OK) {
+		ib::error() << "Unable to encrypt IV";
+		return false;
+	}
+
+	ut_ad(dst_len == MY_AES_BLOCK_SIZE);
+
+	res = my_aes_crypt(MY_AES_CBC,
+		ENCRYPTION_FLAG_DECRYPT | ENCRYPTION_FLAG_NOPAD,
+		src_block, size, dst_block, &dst_len,
```
percona-ysorokin, line 353: The same question re `dst_len`.

mysql-test/suite/perfschema/r/show_sanity.result
gl-sergei: done

storage/innobase/include/row0log.h
gl-sergei: fixed this comment and corresponding one in .cc file

storage/innobase/row/row0ftsort.cc
gl-sergei: NULL remains NULL after alignment because it is already aligned. I see no issues here.

storage/innobase/row/row0log.cc
gl-sergei: Basically it is just a wrapper for `EVP_EncryptUpdate`, which does not allow `in` and `out` to partially overlap, but it does support in-place encryption. I am cheating here of course because I know that `iv` size equals to cipher block size and must be enough to accommodate the result, but I am doing it consciously.

storage/innobase/row/row0log.cc
gl-sergei: `dst_len` is output only. Here and below as well.

storage/innobase/row/row0ftsort.cc
gl-sergei: hmm.. however .. .`ut_align` implementation seems to disagree with me. It contains `ut_ad(ptr);`

gl-sergei: moved check before `ut_align`

gl-sergei:
All comments addressed, Jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1624/

storage/innobase/row/row0ftsort.cc
gl-sergei: BTW, upstream code just few lines above has the same problem. I did not touch it.

percona-ysorokin (approved):
LGTM

laurynas-biveinis (approved):

PR #1935, opened 2017-09-21
Add option to limit write batch size

george-lorch:
Upstream commit ID : fb-mysql-5.6.35/ad4cb203b72c2dc44ee56b6ab1dea653a52fdba

Summary:
Add a session variable `rocksdb_write_batch_max_bytes` which determines the maximum size of a rocksdb write batch. A value of zero means no limit. A write batch may exceed the limit by a bit because 2PC markers are not checked.

update-submodule: rocksdb

Reviewed By: gunnarku

Differential Revision: D4864740

fbshipit-source-id: 66acf6e

### DIFF
--- a/mysql-test/suite/rocksdb.sys_vars/t/rocksdb_write_batch_max_bytes_basic.test
+++ b/mysql-test/suite/rocksdb.sys_vars/t/rocksdb_write_batch_max_bytes_basic.test
@@ -1,0 +1,26 @@
+--source include/have_rocksdb.inc
+
+create table t (i int) engine=rocksdb;
+
+insert into t values (1), (2), (3), (4), (5);
+
+set session rocksdb_write_batch_max_bytes = 1000;
+
+insert into t values (1), (2), (3), (4), (5);
+
+set session rocksdb_write_batch_max_bytes = 10;
+
+--error ER_RDB_STATUS_GENERAL
+insert into t values (1), (2), (3), (4), (5);
+
+set session rocksdb_write_batch_max_bytes = 0;
+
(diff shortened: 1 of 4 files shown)


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1218/

mysql-test/suite/rocksdb.sys_vars/t/rocksdb_write_batch_max_bytes_basic.test
```diff
@@ -0,0 +1,26 @@
+--source include/have_rocksdb.inc
```
laurynas-biveinis, line 1: Ideally should go to the rocksdb main suite, and sys_vars should have the usual red tape test of session/global, ro/rw, type, range testing. I'm not asking you to make this change unless you feel this is necessary.

laurynas-biveinis (approved):

mysql-test/suite/rocksdb.sys_vars/t/rocksdb_write_batch_max_bytes_basic.test
george-lorch: Yeah, I kind of thought that too when I was touching up the test on import but left as-is for future merge. Like other tests, probably not a bad idea to create a JIRA issue to suggest and track the change/split, then send it upstream.

george-lorch: Created https://jira.percona.com/browse/MYR-189

PR #1940, opened 2017-10-02
Bug#1720810: after installation PS fails to run keyring_vault test suite

robgolebiowski:
After building and installing Percona Server - mtr run from installation
directory fails to run keyring_vault test suite.

The fix was to copy keyring_vault test directory to installation
directory on PS install.


laurynas-biveinis (approved):

percona-ysorokin (approved):
LGTM

PR #1941, opened 2017-10-02
Fix bug 1706356 5.7 (Adding index corrupts data in memory table)

Olernov:
When altering a heap table the data is being copied to a temporary table.
While creating this temporary table heap_prepare_hp_create_info function
is called, and it sets wrong value of hp_create_info.fixed_key_fieldnr.
This wrong value causes wrong behaviour of function heap_create called
after that.
This commit disables increase of fixed_key_fieldnr and fixes
bug LP 1706356. It also adds this bug to test case heap.test.
All other heap tests also pass successfully after this patch.

### DIFF
--- a/storage/heap/ha_heap.cc
+++ b/storage/heap/ha_heap.cc
@@ -856,15 +856,6 @@ heap_prepare_hp_create_info(TABLE *table_arg, bool internal_table,
       {
         fixed_data_size= next_field_pos;
       }
-
-
-      if (field->field_index >= fixed_key_fieldnr)
-      {
-        /*
-          Do not use seg->fieldnr as it's not reliable in case of temp tables
-        */
-        fixed_key_fieldnr= field->field_index + 1;
-      }
     }
   }
 
--- a/storage/heap/ha_heap.cc
+++ b/storage/heap/ha_heap.cc
@@ -663,7 +663,7 @@ heap_prepare_hp_create_info(TABLE *table_arg, bool internal_table,
 {
   uint key, parts, mem_per_row_keys= 0, keys= table_arg->s->keys;
   uint auto_key= 0, auto_key_type= 0;
-  uint fixed_key_fieldnr = 0, fixed_data_size = 0, next_field_pos = 0;
+  uint fixed_data_size = 0, next_field_pos = 0;
   uint column_idx, column_count= table_arg->s->fields;
   HP_COLUMNDEF *columndef;
   HP_KEYDEF *keydef;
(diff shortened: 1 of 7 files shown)


storage/heap/ha_heap.cc
```diff
@@ -856,15 +856,6 @@ heap_prepare_hp_create_info(TABLE *table_arg, bool internal_table,
       {
         fixed_data_size= next_field_pos;
       }
-
```
laurynas-biveinis, line 859: Deleting this code effectively makes fixed_key_fieldnr local var a constant zero, and then hp_create_info->fieked_key_fieldnr initialisation at the bottom of this function is always initialized to 0 too. Looking at how this field is used in heap_create, by code review, this would result in share->fixed_column_count / fixed_data_length to be initialized with too large values - can you please check?

laurynas-biveinis (changes requested):
Not critical (I'll have to re-apply the patch before merging), but if possible, please don't move submodule pointers.

storage/heap/ha_heap.cc
Olernov: No, if we consider creating and altering table T1 from my addition to heap.test, then share->fixed_column_count=2 and fixed_data_length=16 for both creating and altering the table. That seems reasonable. While debugging heap_create without the patch exposes fixed_column_count=4 and fixed_data_length=30 when altering the table, what is obviously wrong. 
Actually I could not quite figure out the meaning of field fixed_key_fieldnr and even suppose it's correct value for T1 case (1, 2 or smth else?) . Deleting code really makes it constant 0, so we may consider removing this var from code at all.  
It seems to me that looping through all columns in heap_create (line 136) starting from 0 should give us correct results, so is there any need in this field?

laurynas-biveinis: Indeed, I agree that loop at heap_create can just start from zero and that fixed_key_fieldnr can be simply deleted

Olernov: Deleted fixed_key_fieldnr but still could not handle those submodules to make the patch mergeable, sorry.

laurynas-biveinis (commented):
Testing looks good - do you want to port this patch to 5.5 and 5.6?

Olernov:
Sure, #1947 for 5.6.

Olernov:
#1948 for 5.5.

Olernov:
Re-applied the patch in #1949 where it has no conflicts with the base branch. This pull request may be closed.

PR #1942, opened 2017-10-05
Fix typo in m_string.h

tarqd:
Unfortunate typo


laurynas-biveinis:
Thanks! Feel free to submit to upstream MySQL too.

PR #1944, opened 2017-10-05
Ps 5.7 myr 180

george-lorch:


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1241/

mysql-test/suite/rocksdb.sys_vars/t/rocksdb_merge_tmp_file_removal_delay_ms_basic.test
```diff
@@ -0,0 +1,50 @@
...
+--source ../include/rocksdb_sys_var.inc
+
+DROP TABLE valid_values;
+DROP TABLE invalid_values;
+
+set session rocksdb_merge_buf_size=250;
+set session rocksdb_merge_combine_read_size=1000;
+set session rocksdb_merge_tmp_file_removal_delay_ms=1000;
+
+CREATE TABLE t1 (i INT, j INT, PRIMARY KEY (i)) ENGINE = ROCKSDB;
```
laurynas-biveinis: Should be split to a separate test in a rocksdb suite?

mysql-test/suite/rocksdb/t/flush_log_at_trx_commit.test
```diff
@@ -9,8 +9,8 @@ SET GLOBAL rocksdb_write_ignore_missing_column_families=true;
 
 create table aaa (id int primary key, i int) engine=rocksdb;
 
-SET GLOBAL rocksdb_flush_log_at_trx_commit=0;
---exec sleep 30
+SET GLOBAL rocksdb_flush_log_at_trx_commit=1;
+--exec sleep 5
```
laurynas-biveinis, line 13: What is the purpose of this and other sleeps in this testcase in the first place?

storage/rocksdb/ha_rocksdb.h
```diff
@@ -360,6 +372,8 @@ struct st_global_stats {
   ib_counter_t<ulonglong, 64, RDB_INDEXER> system_rows[ROWS_MAX];
 
   ib_counter_t<ulonglong, 64, RDB_INDEXER> queries[QUERIES_MAX];
+
+  ib_counter_t<ulonglong> covered_secondary_key_lookups;
```
laurynas-biveinis: IIRC that was one of recent upstream performance regressions, which has been fixed later, would be good to merge it by the next release at least

laurynas-biveinis (commented):
Some questions, go ahead with merge or address as you see fit

mysql-test/suite/rocksdb.sys_vars/t/rocksdb_merge_tmp_file_removal_delay_ms_basic.test
george-lorch: Yeah, probably should, sys_vars shouldn't be testing option behaviors. There may be another test already that I can just move this over into.

mysql-test/suite/rocksdb/t/flush_log_at_trx_commit.test
george-lorch: I believe the sleeps are to wait for the background timed flusher to wake/complete. Using debug variables or debug sync points would be ideal, but, the background flusher is down in rocksdb code which has no/difficult access to higher layer syncing tools.

storage/rocksdb/ha_rocksdb.h
george-lorch: Yes, good eye. I actually meant to pull that commit into this August series (it is in next few commits in September series) and forgot about it. (69512214a5d9b0888e20c8cd369c1fde996674b2 Fixing performance regression because of using ib_counter_t incorrectly) I'll go drag it into this series and update the PR.

george-lorch:
OK, I pulled in the perf regression commit and split the sysvar test. Mostly minor changes but ran another jenkins just to be sure http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1244/

mysql-test/suite/rocksdb/t/flush_log_at_trx_commit.test
laurynas-biveinis: Not necessarily: if there were a status counter how many times a background task was executed, it wouldn't be that hard to access and would allow release build testing too

laurynas-biveinis (approved):
Please log somewhere an issue for sleeps in mysql-test/suite/rocksdb/t/flush_log_at_trx_commit.test

george-lorch:
See https://github.com/facebook/mysql-5.6/issues/724 : Curious sleeps in write_sync.test

PR #1947, opened 2017-10-10
Fix bug 1706356 5.6 (Adding index corrupts data in memory table)

Olernov:
When altering a heap table the data is being copied to a temporary table.
While creating this temporary table heap_prepare_hp_create_info function
is called, and it sets wrong value of hp_create_info.fixed_key_fieldnr.
This wrong value causes wrong behaviour of function heap_create called
after that. This commit removes using of fixed_key_fieldnr from code.


Olernov:
Sorry, forgot to remove fixed_key_fieldnr from include/heap.h. Added one more commit.

laurynas-biveinis:
Merged in https://github.com/percona/percona-server/pull/1966

PR #1951, opened 2017-10-11
MYR-121 : Investigate missing FastCRC32 in rocksdb

george-lorch:
- Added detection and pass through options that activate compile in and runtime
  check of SSE42/FastCRC32. Unfortunately, the way Rocksdb is coded, the host
  that builds the binaries must support SSE42 in order for the runtime detection
  to get compiled in that switches between Fast and Slow CRC32.


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1248/
https://jira.percona.com/browse/BLD-847
https://jira.percona.com/browse/MYR-121

storage/rocksdb/CMakeLists.txt
```diff
@@ -50,6 +50,22 @@ IF(HAVE_SCHED_GETCPU)
...
+#include <cstdint>
+#include <nmmintrin.h>
+int main() {
+  volatile uint32_t x = _mm_crc32_u32(0, 0);
+}
+" HAVE_SSE42)
+IF (HAVE_SSE42)
+  ADD_DEFINITIONS(-DHAVE_SSE42)
+ELSE ()
+  MESSAGE(${MYROCKS_STATUS_MODE} "No SSE42 support found. Not building MyRocks")
```
laurynas-biveinis, line 66: Is "Not building MyRocks" message true?

EvgeniyPatlan: In our case we couldn't stop building it. We still need to build it but without -DHAVE_SSE42
In general I hope all our build machines should support it. Maybe the message should be - building without crc32 support

laurynas-biveinis: OK, then the message should read "Building MyRocks without SSE 4.2 support"

george-lorch: It ties into the -DWITH_ROCKSDB=0|1|2 functionality here. It will either be a warning and skip build of MyRocks or it will be a failure.

george-lorch:
According to Vadim, we must support FastCRC32 for MyRocks on all supported platforms. I think we have a problem if we still have CPUs around for building that are > 9 years old. My preliminary builds on jenkins showed MyRocks building successfully where expected. I suppose though it is possible that a mis-configured VM could 'mask' the CPU flags from the host system.

laurynas-biveinis (approved):

PR #1961, opened 2017-10-26
PS-249: Decide whether full JSON parser should be used

robgolebiowski:
with keyring_vault

It was decided that the vault_parser.cc is enough for our
needs to parse simple responses from Vault server.

In this commit a test for parsing garbage data was added.
All tests where run under valgrind. Valgrind did not show
any memory leakage or possible problems.

### DIFF
--- a/unittest/gunit/keyring_vault/vault_parser-t.cc
+++ b/unittest/gunit/keyring_vault/vault_parser-t.cc
@@ -4,6 +4,7 @@
 #include "vault_parser.h"
 #include "vault_key.h"
 #include "vault_base64.h"
+#include "my_rnd.h"
 
 #if defined(HAVE_PSI_INTERFACE)
 namespace keyring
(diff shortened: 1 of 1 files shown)


unittest/gunit/keyring_vault/vault_parser-t.cc
```diff
@@ -209,6 +210,18 @@ namespace keyring__vault_parser_unittest
     EXPECT_TRUE(vault_parser.parse_key_data(payload, &key));
   }
 
+  TEST_F(Vault_parser_test, ParsePayloadThatsGarbage)
+  {
+    uchar garbage[101000];
```
percona-ysorokin, line 215: This shouldn't be addressed, just out of curiosity, why 101000? :)

unittest/gunit/keyring_vault/vault_parser-t.cc
```diff
@@ -209,6 +210,18 @@ namespace keyring__vault_parser_unittest
...
+    my_rand_buffer(garbage, sizeof(garbage));
+    Secure_string payload(reinterpret_cast<char*>(garbage));
```
percona-ysorokin: There is a non-zero chance that `garbage` won't have any zero bytes in it. So, this is a potential memory overrun.
`Secure_string payload(reinterpret_cast<char*>(garbage), sizeof(garbage));` may be?

percona-ysorokin (changes requested):
LGTM with ` Secure_string` construction fixed

laurynas-biveinis:
I don't see any record of "It was decided ... " on PS-249?

unittest/gunit/keyring_vault/vault_parser-t.cc
robgolebiowski: No particular reason :)

unittest/gunit/keyring_vault/vault_parser-t.cc
robgolebiowski: nice catch

percona-ysorokin (approved):
LGTM

laurynas-biveinis (approved):

PR #1962, opened 2017-10-26
Ps 5.7 myr 197

george-lorch:


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1258/

george-lorch:
History into PS 5.7 tree has been inadvertently re-written, I need to recreate/re-write the branch in order to preserve the proper PS 5.7 history...stay tuned.

george-lorch:
OK, clean branch with no history issues force pushed and a new jenkins http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1259/

mysql-test/suite/rocksdb/t/bulk_load_unsorted_errors.test
```diff
@@ -0,0 +1,8 @@
+--source include/have_rocksdb.inc
+--source include/have_nodebug.inc
```
laurynas-biveinis, line 2: Why?

mysql-test/suite/rocksdb/t/skip_validate_tmp_table.test
```diff
@@ -0,0 +1,26 @@
+--source include/have_rocksdb.inc
+--source include/have_debug.inc
+
+# Write file to make mysql-test-run.pl expect the "crash", but don't restart the
+# server until it is told to
+--source include/expect_crash.inc
```
laurynas-biveinis: I don't see any crash-causing statement between here and restart_mysqld.inc ?

mysql-test/suite/rocksdb/t/skip_validate_tmp_table.test
```diff
@@ -0,0 +1,26 @@
...
+# This will append '#sql-test' to the end of new name
+set session debug="+d,gen_sql_table_name";
+rename table t1 to t2;
+set session debug= "-d,gen_sql_table_name";
+
+# Remove the corresponding .frm files
+--remove_files_wildcard $MYSQLTEST_VARDIR/mysqld.$_server_id/data/test *t1*.frm
+--remove_files_wildcard $MYSQLTEST_VARDIR/mysqld.$_server_id/data/test *t2*.frm
+
+# Restart the server with a table registered in RocksDB but does not have a .frm file
```
laurynas-biveinis, line 22: Does this test clean up after itself?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2197,18 +2270,25 @@ class Rdb_transaction_impl : public Rdb_transaction {
 
   rocksdb::Status get(rocksdb::ColumnFamilyHandle *const column_family,
                       const rocksdb::Slice &key,
-                      std::string *value) const override {
+                      rocksdb::PinnableSlice *const value) const override {
+    // clean PinnableSlice right begfore Get() for multiple gets per statement
```
laurynas-biveinis: "before"

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2985,6 +3116,54 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
                               tx->get_lock_count(), tx->get_write_count());
     }
   }
+
+  void populate_deadlock_buffer() {
+    auto dlock_buffer = rdb->GetDeadlockInfoBuffer();
+    m_data += "----------LATEST DETECTED DEADLOCKS----------\n";
+
+    for (auto path_entry : dlock_buffer) {
```
laurynas-biveinis: Doesn't this result in extra copies? Should this be const auto & ?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2985,6 +3116,54 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
...
+                          dl_info.victim_trx_id);
+      }
+      m_data += path_data;
+    }
+  }
+
+  std::vector<Rdb_deadlock_info> get_deadlock_info() {
+    std::vector<Rdb_deadlock_info> deadlock_info;
+    auto dlock_buffer = rdb->GetDeadlockInfoBuffer();
+    for (auto path_entry : dlock_buffer) {
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2959,7 +3040,57 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
...
+
+    txn_data.exclusive_lock = txn.m_exclusive;
+
+    return txn_data;
+  }
+
+  static Rdb_deadlock_info
+  get_dl_path_trx_info(const rocksdb::DeadlockPath &path_entry) {
+    Rdb_deadlock_info deadlock_info;
+
```
laurynas-biveinis: deadlock_info.path.reserve(path_entry.path.size()) ?

storage/rocksdb/CMakeLists.txt
```diff
@@ -123,6 +123,12 @@ INCLUDE_DIRECTORIES(
 ADD_DEFINITIONS(-DROCKSDB_PLATFORM_POSIX -DROCKSDB_LIB_IO_POSIX -DZLIB -DLZ4
   -DZSTD -DROCKSDB_SUPPORT_THREAD_LOCAL)
 
+IF (NOT "$ENV{WITH_TBB}" STREQUAL "")
+  SET(rocksdb_static_libs ${rocksdb_static_libs}
+  $ENV{WITH_TBB}/libtbb${PIC_EXT}.a)
+  ADD_DEFINITIONS(-DTBB)
```
laurynas-biveinis: rocksdb_use_clock_cache = TRUE requires this, so now we must update build deps, Jenkins, everything, to get TBB in the picture, at some correct version.

rocksdb_use_clock_cache is also missing any tests except for the sysvar test that does not exercise its "TRUE" value.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10833,6 +11099,91 @@ static void show_myrocks_vars(THD *thd, SHOW_VAR *var, char *buff) {
...
+static ulonglong
+io_stall_prop_value(const std::map<std::string, std::string> &props,
+                    const std::string &key) {
+  std::map<std::string, std::string>::const_iterator iter =
+      props.find("io_stalls." + key);
+  if (iter != props.end()) {
+    return std::stoull(iter->second);
+  } else {
+    DBUG_PRINT("warning",
+               ("RocksDB GetMapPropery hasn't returned key=%s", key.c_str()));
```
laurynas-biveinis: "Property" ?

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb/t/bulk_load_unsorted_errors.test
george-lorch: Not sure, came from upstream. I suspect though it was due to many debug assertions that were present in bulk load functionality that would be triggered by the test. Many of these have been removed by one of my submitted patches and replaced with client error propagation.

mysql-test/suite/rocksdb/t/skip_validate_tmp_table.test
george-lorch: Ahh, yeah, right. This is a mistake in updating the test when I imported it.

mysql-test/suite/rocksdb/t/skip_validate_tmp_table.test
george-lorch: Actually, no, it doesn't and should. What is interesting here though it that mtr is not detecting the 'dirt'. Probably because the .frm files aren't present anymore it doesn't see the table even though it does still exist within MyRocks. Probably a better thing to do here is to move the .frm files rather than deleting them, then move them back in order to properly cleanup.

storage/rocksdb/CMakeLists.txt
george-lorch: Yeah, I was actually suspicious of taking in that(those) patch(s) at all. TBB though seems is carried with glibc as evidenced that this actually did build on all of our builders. I don't think there is anything really to do there other than maybe also add conditional compilation of the option and its use in MyRocks and maybe fail to cmake MyRocks if it is missing.

Agreed that there is missing test but I'm really unsure of what can be validated other than the server can start with rocksdb_use_clock_cache=on and select @@global.rocksdb_use_clock_cache.

I am inclined to remove this entirely from PS but I suspect that this will be another of those edge case perf improvements that will be complained about if it is missing.

storage/rocksdb/ha_rocksdb.cc
george-lorch: There are lots of full copies in this code, but, this is for an I_S table used for debugging deadlocks, so performance isn't necessarily the focus. This was taken directly from the initial upstream implementation. I'll send an upstream PR for these so we can try to keep merging nightmares to a minimum.

mysql-test/suite/rocksdb/t/bulk_load_unsorted_errors.test
george-lorch: OK, I see the problem here. This functionality seems broken. There is a 'set' callback for the rocksdb_bulk_load_allow_unsorted. This callback calls my_error, which is not really a valid thing to do. When cleaning up the statement execution, debug builds assert if the status has already been set :

 375 void Diagnostics_area::set_ok_status(ulonglong affected_rows,
 376                                      ulonglong last_insert_id,
 377                                      const char *message_text)
 378 {
 379   DBUG_ENTER("set_ok_status");
 380   DBUG_ASSERT(! is_set());

So this is an upstream bug. They should be using a 'check' callback for the sysvar instead. I will make up a patch for this and send it upstream.

george-lorch: See https://github.com/facebook/mysql-5.6/pull/755

mysql-test/suite/rocksdb/t/skip_validate_tmp_table.test
george-lorch: Further digging seems that the test really does 'clean up' after itself. When MyRocks is restarted the final time with a table in its internal data dictionary that does not have a .frm file, it 'forgets' about that table, i.e. removes it from its data dictionary. If the test were to simply move the .frm away and then back, MyRocks will fail to start as it then detects that there is a .frm file but no entry in its data dictionary. So, while this is not 100% ideal, it seems to be about as close to a cleanup that can be done. The end result is that there will be 'orphaned' table data in the LSM that is never accessible.

george-lorch:
New jenkins for the things that were addressed http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1260/

storage/rocksdb/CMakeLists.txt
laurynas-biveinis: This is more like a "SSE4.2" situation in that a build w/o dep simply results in missing functionality, not a build failure, and a runtime error only. Thus builders succeeding does not tell anything about TBB. That missing test would, which can be as simple as possible (start an instance, run some RW workload) to run the code paths.

Fine if we can remove it but for all I know clock cache could be the future and not an edge case. Thus I don't think we can ignore it that easily, unfortunately.

george-lorch: Well, if you look at what I pushed, I made the lack of WITH_TBB a cmake failure in the same paradigm that follows the WITH_ROCKSDB=0|1|FORCE. Carefully reviewed the resulting jenkins jobs to see that the same jobs passed/failed as before. The missing test that was added in the push does just that, uses .opt file to start with the option on and selects the option to ensure it is on.

mysql-test/suite/rocksdb/t/skip_validate_tmp_table.test
laurynas-biveinis: This points to an operational issue in MyRocks re. .frm-less table recovery we will have to provide means to resolve sooner or later. Is this tracked somewhere?

For the testcase, maybe a comment stating the end state of the instance would help here?

mysql-test/suite/rocksdb/t/use_clock_cache.test
```diff
@@ -0,0 +1,3 @@
+--source include/have_rocksdb.inc
```
laurynas-biveinis: The test needs some block I/O so that they go through this clock cache.

Also it appears that the test does not fail in a TBB-less build, and what happens here is that NewClockCache returns nullptr which MyRocks happily takes and passes around. This suggests that, if block cache is requested (as it is in "if (!rocksdb_tbl_options->no_block_cache)"), then getting nullptr for it should be a hard error, not silent cache-less operation.

laurynas-biveinis (changes requested):

storage/rocksdb/CMakeLists.txt
laurynas-biveinis: Uhm CMake diff looks the same as in the previous iteration, am I missing something? I wrote the above without seeing the new test yet, but based on the code review. The test passing looks wrong to me, please see below a comment re. test.

mysql-test/suite/rocksdb/t/skip_validate_tmp_table.test
george-lorch: Not tracked anywhere. A test case comment clarifying the lack of cleanup seems reasonable.

storage/rocksdb/CMakeLists.txt
george-lorch: Crap, somehow, my revisions to CMakeLists.txt did not get pushed. I need to do this one over. The basic idea is to demand that TBB is available and warn/error on cmake based on WITH_ROCKSDB=0|1|FORCE

mysql-test/suite/rocksdb/t/use_clock_cache.test
george-lorch: No, a TBB less build should fail to compile completely as the definition of TBB specific code in rocksdb offers no compilation alternative for NewClockCache and would result in a basic compilation error. So, This code _requires_ TBB to compile, or not, no in-between.

laurynas-biveinis: Granted I'm looking at the RocksDB HEAD and not at the specific submodule pointer commit, yet, in clock_cache.h:

#if defined(TBB) && !defined(ROCKSDB_LITE)
#define SUPPORT_CLOCK_CACHE
#endif

In clock_cache.cc:

#ifndef SUPPORT_CLOCK_CACHE

namespace rocksdb {

std::shared_ptr<Cache> NewClockCache(size_t capacity, int num_shard_bits,
                                     bool strict_capacity_limit) {
  // Clock cache not supported.
  return nullptr;
}

}  // namespace rocksdb

#else

... implementation

george-lorch: Hmm, yes. I am not sure if I missed this in the diff at the point of the commit or if it was changed in rocksdb after. Either way, it is a mess for us and one that I personally do not wish to expend the time to figure out right now to get this series to pass review. We either take the change as-is, or we skip it entirely for now until some later time when we can properly evaluate and apply, then wait for some web celebrity to complain that we do not embrace the most modern edge case performance bits available. My preference is always stability and consistency across platforms before edge performance.

laurynas-biveinis: If fixing TBB is too much work, as it seems to be, that's fine, but what I find unacceptable is option being broken silently due to that missing nullptr check. Add that, remove the testcase, no need to patch CMake, file a MYR for TBB, and it's good to go

storage/rocksdb/CMakeLists.txt
george-lorch: I'm just going to pull this entirely and will revisit later if it becomes an issue

george-lorch:
new jenkins with the latest changes http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1261/

mysql-test/suite/rocksdb/t/skip_validate_tmp_table.test
laurynas-biveinis: A MYR/upstream issue for no means to recover or drop a table w/o FRM has to be filed, could be as brief as possible for now

laurynas-biveinis (approved):
Please also file a MYR for the missing clock cache import. It does not mean that we will have to import it, just that at this point until we know better we have to track what we did not import and why

george-lorch:
https://jira.percona.com/browse/MYR-199 : Missing clock cache import due to TBB requirements
https://github.com/facebook/mysql-5.6/issues/757 : Support .frm-less recovery

PR #1969, opened 2017-11-09
MYR-200 : Make default cf compression option to be LZ4

george-lorch:
- Changed default value of rocksdb_default_cf_options to ensure that the
  'default' column family is created with options necessary to compress all
  levels of th LSM tree with lz4.
- Re-recorded tests that expose the rocksdb_default_cf_options
- Altered test options for index_merge_rocksdb to disable compression because
  test relies on 'filler' data within a table that forces the table statistics
  into a certain shape to ensure a particular query plan is chosen. The
  compression alters this shape and changes the query plan which changes the
  focus of the test.

### DIFF
--- a/mysql-test/suite/rocksdb.sys_vars/r/rocksdb_default_cf_options_basic.result
+++ b/mysql-test/suite/rocksdb.sys_vars/r/rocksdb_default_cf_options_basic.result
@@ -1,7 +1,7 @@
 SET @start_global_value = @@global.ROCKSDB_DEFAULT_CF_OPTIONS;
 SELECT @start_global_value;
 @start_global_value
-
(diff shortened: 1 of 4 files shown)


george-lorch:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1281/

laurynas-biveinis (approved):
LGTM provided you confirmed that SSTs are indeed LZ4 compressed with this change

george-lorch:
Yeah, in fact I learned something new about the cf compression options where my compression tests were really not complete in that the did not set bottommost_compression. They really don't test compression though as much as they test that options can be set, rocksdb accepts them, and reports them back properly. You need to either push in data or use a tool like sstdump to see what is actually happening down in the rocksdb layer.

PR #1970, opened 2017-11-10
lp-1657737: Print locking transaction which cause "Lock wait timeout"…

vlad-lesin:
… errors into log file

The output format is the following:

2017-11-09T22:43:29.213401Z 5 [Note] InnoDB: Lock wait timeout info:
Requested thread id: 5
Requested trx id: 1289
Requested query: DELETE FROM t WHERE i = 1
Total blocking transactions count: 1
Blocking transaction number: 1
Blocking thread id: 4
Blocking query id: 38
Blocking trx id: 422021753699104

The execution process of blocking transaction can keep going on as opposed to
deadlock when all transactions, which take part in the deadlock, are blocked.
That is why some data in blocking transaction info can be changed(or have not
yet initialized) at the moment when we capture them. At the above example
blocking transaction has not yet initialized at the moment of capturing.

And that is why the info is captured before requested thread suspending, as
after it's waking up blocking transaction might has been already finished.

See also: https://github.com/percona/percona-server/pull/1971

http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2007/

### DIFF
--- a/mysql-test/suite/innodb/t/bug84563-master.opt
+++ b/mysql-test/suite/innodb/t/bug84563-master.opt
@@ -1,0 +1,1 @@
+--log-error=$MYSQLTEST_VARDIR/tmp/bug84563.err
--- a/mysql-test/suite/innodb/t/bug84563.test
+++ b/mysql-test/suite/innodb/t/bug84563.test
@@ -1,0 +1,37 @@
+--source include/have_innodb.inc
+--source include/count_sessions.inc
+
+SET @innodb_lock_wait_timeout_saved = @@GLOBAL.innodb_lock_wait_timeout;
+SET @innodb_print_lock_wait_timeout_info_saved =
+  @@GLOBAL.innodb_print_lock_wait_timeout_info;
+
+SET GLOBAL innodb_lock_wait_timeout = 1;
+SET GLOBAL innodb_print_lock_wait_timeout_info = true;
+
+CREATE TABLE t (i INT) ENGINE = InnoDB;
+INSERT INTO t (i) VALUES(1);
+
+START TRANSACTION;
+SELECT * FROM t WHERE i = 1 LOCK IN SHARE MODE;
+
+--connect(con0, localhost, root,,)
+START TRANSACTION;
+--error ER_LOCK_WAIT_TIMEOUT
+DELETE FROM t WHERE i = 1;
+COMMIT;
+--disconnect con0
+
+--connection default
+DELETE FROM t WHERE i = 1;
+COMMIT;
+
+--echo Grepping the error log for lock wait timeout info
+--let SEARCH_FILE= $MYSQLTEST_VARDIR/tmp/bug84563.err
+--let SEARCH_PATTERN= Lock wait timeout info:
+--source include/search_pattern_in_file.inc
+
+DROP TABLE t;
+SET GLOBAL innodb_lock_wait_timeout = @innodb_lock_wait_timeout_saved;
+SET GLOBAL innodb_print_lock_wait_timeout_info =
+  @innodb_print_lock_wait_timeout_info_saved;
+--source include/wait_until_count_sessions.inc
--- a/mysql-test/suite/sys_vars/r/all_vars.result
+++ b/mysql-test/suite/sys_vars/r/all_vars.result
@@ -14,5 +14,7 @@ left join t1 on variable_name=test_name where test_name is null ORDER BY variabl
 There should be *no* variables listed below:
 INNODB_FAKE_CHANGES
 INNODB_FAKE_CHANGES
+INNODB_PRINT_LOCK_WAIT_TIMEOUT_INFO
+INNODB_PRINT_LOCK_WAIT_TIMEOUT_INFO
 drop table t1;
 drop table t2;
--- a/mysql-test/suite/sys_vars/t/innodb_print_lock_wait_timeout_info.test
+++ b/mysql-test/suite/sys_vars/t/innodb_print_lock_wait_timeout_info.test
@@ -1,0 +1,93 @@
+
+# 2010-01-25 - Added
+#
+
+--source include/have_innodb.inc
+
+SET @start_global_value = @@global.innodb_print_lock_wait_timeout_info;
+SELECT @start_global_value;
+
+#
+# exists as global only
+#
+--echo Valid values are 'ON' and 'OFF'
+SELECT @@global.innodb_print_lock_wait_timeout_info in (0, 1);
+SELECT @@global.innodb_print_lock_wait_timeout_info;
+--error ER_INCORRECT_GLOBAL_LOCAL_VAR
+SELECT @@session.innodb_print_lock_wait_timeout_info;
+SHOW global variables LIKE 'innodb_print_lock_wait_timeout_info';
+SHOW session variables LIKE 'innodb_print_lock_wait_timeout_info';
+--disable_warnings
+SELECT * FROM information_schema.global_variables
+WHERE variable_name='innodb_print_lock_wait_timeout_info';
+SELECT * FROM information_schema.session_variables
+WHERE variable_name='innodb_print_lock_wait_timeout_info';
+--enable_warnings
+
+#
+# SHOW that it's writable
+#
+SET global innodb_print_lock_wait_timeout_info='OFF';
+SELECT @@global.innodb_print_lock_wait_timeout_info;
+--disable_warnings
+SELECT * FROM information_schema.global_variables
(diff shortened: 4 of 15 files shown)


mysql-test/suite/innodb/t/bug84563.test
```diff
@@ -0,0 +1,35 @@
...
+SET GLOBAL innodb_print_lock_wait_timeout_info = true;
+
+CREATE TABLE t (i INT) ENGINE = InnoDB;
+INSERT INTO t (i) VALUES(1);
+
+START TRANSACTION;
+SELECT * FROM t WHERE i = 1 LOCK IN SHARE MODE;
+
+--connect(con0, localhost, root,,)
+START TRANSACTION;
```
percona-ysorokin, line 18: may be, in this case, just
```
START TRANSACTION;
--error ER_LOCK_WAIT_TIMEOUT
DELETE FROM t WHERE i = 1;
```
without `--send / --reap`

mysql-test/suite/innodb/t/bug84563.test
```diff
@@ -0,0 +1,35 @@
+--source include/have_innodb.inc
+
```
percona-ysorokin: As you are using an additional connection add
```
...
```

mysql-test/suite/innodb/t/bug84563.test
```diff
@@ -0,0 +1,35 @@
...
+--connection con0
+COMMIT;
+
+--echo Grepping the error log for lock wait timeout info
+--let SEARCH_FILE= $MYSQLTEST_VARDIR/tmp/bug84563.err
+--let SEARCH_PATTERN= Lock wait timeout info:
+--source include/search_pattern_in_file.inc
+
+DROP TABLE t;
+SET GLOBAL innodb_lock_wait_timeout = default;
```
percona-ysorokin: Save these variables into xxx_saved and restore from them rather than from defaults

storage/innobase/lock/lock0wait.cc
```diff
@@ -185,6 +199,45 @@ lock_wait_table_reserve_slot(
...
 
 /***************************************************************//**
+Print lock wait timeout info to stderr. It's supposed this function
+is executed in trx's THD thread as it calls some non-thread-safe
+functions to get some info from THD. */
+void
+print_lock_wait_timeout(
+/*====================*/
+	trx_t		*trx,	/*!< in: requested trx */
+	blocking_trx_info		*blocking,	/*!< in: blocking info array */
```
percona-ysorokin: formatting

percona-ysorokin (changes requested):
Primarily test case comments
Re-running Jenkins build with single test case is OK.

mysql-test/suite/innodb/t/bug84563.test
vlad-lesin: fixed

mysql-test/suite/innodb/t/bug84563.test
vlad-lesin: Yes, sure. I used one more connection to get additional information from I_S and PFS, and forgot to remove this --send/--reap when this additional connection was removed. Fixed.

mysql-test/suite/innodb/t/bug84563.test
vlad-lesin: fixed

storage/innobase/lock/lock0wait.cc
vlad-lesin: Oh, I hate tabs. Fixed.

vlad-lesin:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2009/

mysql-test/suite/innodb/t/bug84563.test
```diff
@@ -0,0 +1,37 @@
+--source include/count_sessions.inc
```
percona-ysorokin: As per our conventions it's better to put lightweight `--source include/have_xxx.inc` includes before other includes which do non-trivial logic.
This is a minor comment which can be ignored

percona-ysorokin (approved):
LGTM with minor comment that can be ignored

mysql-test/suite/innodb/t/bug84563.test
vlad-lesin: fixed

mysql-test/suite/innodb/t/bug84563-master.opt
```diff
@@ -0,0 +1 @@
+--log-error=$MYSQLTEST_VARDIR/tmp/bug84563.err
```
laurynas-biveinis, line 1: This breaks the testcase with false positive when run with MTR --repeat option. The testcase should not use a custom error log path as an option, then restart in the testcase with the path, restart again with default path, grep & delete the custom path

mysql-test/suite/innodb/t/bug84563.test
```diff
@@ -0,0 +1,37 @@
+--source include/have_innodb.inc
```
laurynas-biveinis, line 1: This is a feature, not bugfix, consider naming as percona_print_lock_wait_timeout_trx

mysql-test/suite/sys_vars/r/all_vars.result
```diff
@@ -14,5 +14,7 @@ left join t1 on variable_name=test_name where test_name is null ORDER BY variabl
 There should be *no* variables listed below:
 INNODB_FAKE_CHANGES
 INNODB_FAKE_CHANGES
+INNODB_PRINT_LOCK_WAIT_TIMEOUT_INFO
```
laurynas-biveinis, line 17: The sys_vars test is named incorrectly, missing _basic

mysql-test/suite/sys_vars/t/innodb_print_lock_wait_timeout_info.test
```diff
@@ -0,0 +1,93 @@
+
+# 2010-01-25 - Added
```
laurynas-biveinis, line 2: No

storage/innobase/lock/lock0wait.cc
```diff
@@ -185,6 +199,45 @@ lock_wait_table_reserve_slot(
 }
 
 /***************************************************************//**
+Print lock wait timeout info to stderr. It's supposed this function
```
laurynas-biveinis, line 202: New InnoDB code uses 
/** Foo
@param 
*/
header style

storage/innobase/lock/lock0wait.cc
```diff
@@ -185,6 +199,45 @@ lock_wait_table_reserve_slot(
...
+is executed in trx's THD thread as it calls some non-thread-safe
+functions to get some info from THD. */
+void
+print_lock_wait_timeout(
+/*====================*/
+	trx_t *trx, /*!< in: requested trx */
```
laurynas-biveinis, line 208: const ref

laurynas-biveinis (commented):
Post-push fixes are needed

mysql-test/suite/sys_vars/t/innodb_print_lock_wait_timeout_info.test
vlad-lesin: removed

storage/innobase/lock/lock0wait.cc
vlad-lesin: fixed

storage/innobase/lock/lock0wait.cc
vlad-lesin: fixed

mysql-test/suite/innodb/t/bug84563-master.opt
vlad-lesin: Agree. But there is no need to restart the server once more with default error log path. It will be restarted automatically for the next test case (see server_need_restart() in mysql_test_run.pl)

mysql-test/suite/innodb/t/bug84563.test
vlad-lesin: ok

mysql-test/suite/innodb/t/bug84563-master.opt
laurynas-biveinis: Right, but you need to remove the custom error log path at the end, and the most portable way to do that is with file closed, for which server has to be restarted with a different path. But for us either way is fine.

laurynas-biveinis:
This has caused https://bugs.launchpad.net/percona-server/+bug/1738171

PR #1971, opened 2017-11-10
Lwt bug lp1657737 simpl 5.7 gca

vlad-lesin:
See https://github.com/percona/percona-server/pull/1970.

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1283/


percona-ysorokin (changes requested):
The same test case comments as in 5.6

vlad-lesin:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1285/

percona-ysorokin (approved):
LGTM with the same minor comment as in 5.6

laurynas-biveinis (commented):
Same as for 5.6

PR #1974, opened 2017-11-15

laurynas-biveinis:
…s before indexed columns)

usage. Fix by
cause;
so.

http://10.30.6.254/view/PS%205.5/job/percona-server-5.5-param/1593/


mysql-test/t/percona_heap_var.test
```diff
@@ -83,3 +84,63 @@ select count(*) from t1;
...
+
+CREATE TABLE t1 (a INT PRIMARY KEY, b BLOB, c INT, KEY c(c)) ENGINE=InnoDB;
+--error ER_TABLE_CANT_HANDLE_BLOB
+ALTER TABLE t1 ENGINE=MEMORY;
+DROP TABLE t1;
+
+CREATE TABLE t1 (a INT PRIMARY KEY, b TEXT, c INT, KEY c(c)) ENGINE=InnoDB;
+--error ER_TABLE_CANT_HANDLE_BLOB
+ALTER TABLE t1 ENGINE=MEMORY;
+DROP TABLE t1;
```
percona-ysorokin, line 146: I would also add checks for column reordering
```sql
CREATE TABLE t1 (a INT PRIMARY KEY, b BLOB) ENGINE=MEMORY;
--error ER_TABLE_CANT_HANDLE_BLOB
ALTER TABLE t1 MODIFY b BLOB FIRST;
```

mysql-test/r/percona_heap_var.result
```diff
@@ -192,3 +192,50 @@ count(*)
 10001
 set @@session.max_heap_table_size=default;
 drop table t1;
+#
+# Bug 1731483: MEMORY storage engine incorrectly allows BLOB columns before indexed columns
+#
+CREATE TABLE t1 (a BLOB, b INT PRIMARY KEY) ENGINE=MEMORY;
+ERROR 42000: The used table type doesn't support BLOB/TEXT columns
```
robgolebiowski, line 199: Could error message be more specific ? From this I would assume it is not possible at all to create BLOB/TEXT columns inside memory table

mysql-test/t/percona_heap_var.test
percona-ysorokin: And for composite keys:
```sql
--error ER_TABLE_CANT_HANDLE_BLOB
CREATE TABLE t1 (a INT, b BLOB, c INT, PRIMARY KEY(a, c)) ENGINE=MEMORY;

CREATE TABLE t1 (a INT, b BLOB, c INT) ENGINE=MEMORY;
--error ER_TABLE_CANT_HANDLE_BLOB
ALTER TABLE t1 ADD INDEX idx_composite (a, c);
--error ER_TABLE_CANT_HANDLE_BLOB
ALTER TABLE t1 ADD PRIMARY KEY (a, c);
```

robgolebiowski (changes requested):

storage/heap/ha_heap.cc
```diff
@@ -854,6 +854,15 @@ heap_prepare_hp_create_info(TABLE *table_arg, bool internal_table,
       {
         fixed_data_size= next_field_pos;
       }
+
+
+      if (field->field_index >= fixed_key_fieldnr)
```
percona-ysorokin, line 859: Will this work for the case when BLOB is between fields being indexed, like
```sql
CREATE TABLE t1 (a INT, b BLOB, c INT, PRIMARY KEY(a, c)) ENGINE=MEMORY
```

mysql-test/r/percona_heap_var.result
laurynas-biveinis: I have reused one of the existing messages, which is also used in the feature patch elsewhere. IMHO we should not add new error codes to 5.5 at this point

storage/heap/ha_heap.cc
laurynas-biveinis: Yes this code iterates through all the key parts. The testcase above for the composite keys covers it

robgolebiowski (approved):

percona-ysorokin (approved):
LGTM

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.5-param/1594/

PR #1975, opened 2017-11-15

laurynas-biveinis:
…s before indexed columns)

usage. Fix by
cause;
so.

(cherry picked from commit ec3ef03ac7bc2448e1499793aa8683d6c2020a28)

    http://10.30.6.254/view/PS%205.6/job/percona-server-5.6-param/2011/


percona-ysorokin (approved):
LGTM

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.6-param/2014/

PR #1976, opened 2017-11-15

laurynas-biveinis:
…s before indexed columns)

usage. Fix by
cause;
so.

In 5.7, extend the above to apply for JSON fields too.

(cherry picked from commit ec3ef03ac7bc2448e1499793aa8683d6c2020a28)
(cherry picked from commit 4e6e3e92bedcfb4d924cbf8b15801f562c6ff98b)

    http://jenkins.percona.com/job/mysql-5.7-param/1290/


percona-ysorokin (approved):
LGTM

laurynas-biveinis:
http://jenkins.percona.com/job/mysql-5.7-param/1294/

PR #1978, opened 2017-11-16
Bug1704195 5.7

laurynas-biveinis:
http://jenkins.percona.com/job/mysql-5.7-param/1292/


mysql-test/include/percona_nonflushing_analyze_debug.inc
```diff
@@ -0,0 +1,31 @@
+#
+# Test ANALYZE TABLE that does does flush table definition cache
```
robgolebiowski: Two does

mysql-test/include/percona_nonflushing_analyze_debug.inc
```diff
@@ -0,0 +1,31 @@
...
+send_eval SELECT * FROM $percona_nonflushing_analyze_table;
+
+--connection default
+
+SET DEBUG_SYNC="now WAIT_FOR idx_scan_in_progress";
+
+eval ANALYZE TABLE $percona_nonflushing_analyze_table;
+
+# With the bug fixed this should not block
+eval SELECT * FROM $percona_nonflushing_analyze_table;
```
robgolebiowski, line 22: How do we check that this statement is not blocking ?

sql/sql_admin.cc
```diff
@@ -1014,6 +1014,9 @@ static bool mysql_admin_table(THD* thd, TABLE_LIST* tables,
 
     if (table->table)
     {
+      const bool skip_flush=
```
robgolebiowski, line 1017: Just a sugestion - maybe different name for this variable? skip_flush is quite vage while we have clear reason why we skip flush. Maybe is_table_being_analized ?

robgolebiowski (changes requested):

percona-ysorokin (approved):
LGTM

george-lorch (approved):

mysql-test/include/percona_nonflushing_analyze_debug.inc
laurynas-biveinis: Without the change the testcase hangs here until timeout

sql/sql_admin.cc
laurynas-biveinis: Perhaps, but definitely not is_table_being_analyzed, because of the second part after &&: "(table->table->file->ha_table_flags() & HA_ONLINE_ANALYZE)"

PR #1979, opened 2017-11-16
Fix bug 1713700 (Assertion `inited == INDEX' failed).

Olernov:
When executing join_materialize_semijoin there is a possibility
of exceeding the size of  in-memory temporary table used to execute query.
In this case server converts it to ondisk table but misses index use
initialization because create_ondisk_from_heap function does not
perform it automatically.


dutow:
This was fixed in 5.7.22.

PR #1985, opened 2017-11-22
PS-256: keys created by keyring_vault MTR tests are not removed from

robgolebiowski:
Vault server

This fix changes how keys are removed from Vault. Instead of removing
one key after another, test:
 1) creates a mount point before running the tests.
 2) run actual test (which creates/removes keys in/from mount point
    created in step 1
 3) removes the mount point created in step 1


robgolebiowski:
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1311/

plugin/keyring_vault/tests/mtr/mount_point_service.inc
```diff
@@ -0,0 +1,62 @@
+# Creates or deletes secret mount point specified in keyring_vault configuration file
+# Mount point is :
+#  - created when MOUNT_POINT_SERVICE_OP is set to 1
```
laurynas-biveinis: Why not use "CREATE", "DELETE" or some other descriptive strings instead of 1 and 2?

laurynas-biveinis (changes requested):

plugin/keyring_vault/tests/mtr/mount_point_service.inc
robgolebiowski: The problem is where to define those CREATE and DELETE. Test includes mount_point_service.inc and it needs to define what operation it wants mount_point_service to perform. The perfect solution would be --let MOUNT_POINT_SERVICE_OP=CREATE and then source mount_point_service.inc, but the CREATE would need to be defined before this source operation and in every test that source mount_point_service.inc

laurynas-biveinis: Not CREATE but "CREATE" ?

robgolebiowski: ahh, right, changing

plugin/keyring_vault/tests/mtr/mount_point_service.inc
```diff
@@ -0,0 +1,62 @@
...
+    {
+      $secret_mount_point=$1;
+    }
+    elsif ($row =~ m/vault_ca[ ]*= [ ]*(.*)/)
+    {
+      $vault_ca=$1;
+    }
+  }
+  close(CONF_FILE);
+  if ($token eq "" || $vault_url eq "" || $secret_mount_point eq "")
```
percona-ysorokin, line 45: `|| $vault_ca eq ""`? Or is it optional?

robgolebiowski: it's optional

percona-ysorokin: Then you need to take into account this when you form `curl` command lines for both `CREATE` and `DELETE`
```perl
if($vault_ca)
{
    system(qq#curl -H "X-Vault-Token: $token" --cacert $vault_ca -X DELETE $vault_url/v1/sys/mounts/$secret_mount_point#);
}
else
{
    system(qq#curl -H "X-Vault-Token: $token" -X DELETE $vault_url/v1/sys/mounts/$secret_mount_point#);
}
```

plugin/keyring_vault/tests/mtr/mount_point_service.inc
robgolebiowski: Done

plugin/keyring_vault/tests/mtr/mount_point_service.inc
robgolebiowski: Done

robgolebiowski:
JR => https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1393/

percona-ysorokin (approved):
LGTM

plugin/keyring_vault/tests/mtr/key_rotation_qa.test
```diff
@@ -27,6 +27,13 @@ call mtr.add_suppression("\\[ERROR\\] InnoDB: Failed to decrpt encryption inform
 
 --source generate_default_conf_files.inc
 
+# Create mount points
+--let MOUNT_POINT_SERVICE_OP=CREATE
```
laurynas-biveinis, line 31: let FOO= vs let $BAR=
inconsistency on the dollar sign. By itself too minor to force a PR respin

laurynas-biveinis (approved):

laurynas-biveinis:
This has changed RocksDB submodule pointer while it is not RocksDB-related, probably broke the its builds

laurynas-biveinis:
https://bugs.launchpad.net/percona-server/+bug/1740350

PR #1986, opened 2017-11-23
BLD-862: Fixed compilation issues with gcc-7 for the 5.5 branch

dutow:
Most of the changes won't result in any behavior change - I had to reformat/move some comments and breaks so GCC's fallthrough parser finds them.

I also added breaks, with the reasoning explained in the commit message - most of the time, they were present in later versions.

The other changes are about fixing issues with sprintf (buffer length/truncation issues), and some uninitialized variables in the release build.

### DIFF
--- a/client/mysql.cc
+++ b/client/mysql.cc
@@ -2717,7 +2717,7 @@ You can turn off this feature to get a quicker startup with -A\n\n");
         mysql_free_result(fields);
         break;
       }
-      field_names[i][num_fields*2]= '\0';
+      field_names[i][num_fields*2]= NULL;
       j=0;
       while ((sql_field=mysql_fetch_field(fields)))
       {
--- a/extra/replace.c
+++ b/extra/replace.c
@@ -173,6 +173,7 @@ register char **argv[];
 	break;
       case 'V':
 	version=1;
+      // fallthrough
       case 'I':
       case '?':
 	help=1;					/* Help text written */
--- a/sql/sp_head.cc
+++ b/sql/sp_head.cc
@@ -150,7 +150,8 @@ sp_get_item_value(THD *thd, Item *item, String *str)
   case DECIMAL_RESULT:
     if (item->field_type() != MYSQL_TYPE_BIT)
       return item->val_str(str);
-    else {/* Bit type is handled as binary string */}
+    // fallthrough
+    /* Bit type is handled as binary string */
   case STRING_RESULT:
     {
       String *result= item->val_str(str);
@@ -2888,10 +2889,10 @@ sp_head::show_routine_code(THD *thd)
     */
     if (ip != i->m_ip)
     {
-      const char *format= "Instruction at position %u has m_ip=%u";
+      const char format[] = "Instruction at position %u has m_ip=%u";
       char tmp[sizeof(format) + 2*SP_INSTR_UINT_MAXLEN + 1];
 
-      sprintf(tmp, format, ip, i->m_ip);
+      snprintf(tmp, sizeof(tmp), format, ip, i->m_ip);
       /*
         Since this is for debugging purposes only, we don't bother to
         introduce a special error code for it.
--- a/sql/sql_partition.cc
+++ b/sql/sql_partition.cc
@@ -7637,7 +7637,7 @@ int get_part_iter_for_interval_cols_via_map(partition_info *part_info,
                                             PARTITION_ITERATOR *part_iter)
 {
   uint32 nparts;
-  get_col_endpoint_func  get_col_endpoint;
+  get_col_endpoint_func  get_col_endpoint= NULL;
   DBUG_ENTER("get_part_iter_for_interval_cols_via_map");
 
   if (part_info->part_type == RANGE_PARTITION)
--- a/storage/heap/hp_extra.c
+++ b/storage/heap/hp_extra.c
@@ -34,6 +34,7 @@ int heap_extra(register HP_INFO *info, enum ha_extra_function function)
   switch (function) {
   case HA_EXTRA_RESET_STATE:
     heap_reset(info);
+    break;
   case HA_EXTRA_NO_READCHECK:
     info->opt_flag&= ~READ_CHECK_USED;	/* No readcheck */
     break;
--- a/storage/innobase/include/data0type.ic
+++ b/storage/innobase/include/data0type.ic
@@ -477,7 +477,8 @@ dtype_get_fixed_size_low(
 #else /* !UNIV_HOTBACKUP */
 		return(len);
 #endif /* !UNIV_HOTBACKUP */
-		/* fall through for variable-length charsets */
+		// fallthrough 
+		// for variable-length charsets 
 	case DATA_VARCHAR:
 	case DATA_BINARY:
 	case DATA_DECIMAL:
--- a/storage/innobase/include/log0online.h
+++ b/storage/innobase/include/log0online.h
@@ -141,7 +141,11 @@ log_online_bitmap_iterator_next(
 
 /** Struct for single bitmap file information */
 struct log_online_bitmap_file_struct {
-	char		name[FN_REFLEN];	/*!< Name with full path */
+	/** Name with full path
+	    61 is a nice magic constant for the extra space needed for the sprintf
+	    template in the cc file
+	*/
+	char		name[FN_REFLEN+61];	/*!< Name with full path */
 	os_file_t	file;			/*!< Handle to opened file */
 	ib_uint64_t	size;			/*!< Size of the file */
 	ib_uint64_t	offset;			/*!< Offset of the next read,
--- a/storage/innobase/include/page0zip.ic
+++ b/storage/innobase/include/page0zip.ic
@@ -170,7 +170,7 @@ page_zip_rec_needs_ext(
 				ignored if zip_size == 0 */
 	ulint	zip_size)	/*!< in: compressed page size in bytes, or 0 */
 {
-	ut_ad(rec_size > comp ? REC_N_NEW_EXTRA_BYTES : REC_N_OLD_EXTRA_BYTES);
+	ut_ad(rec_size > (comp ? REC_N_NEW_EXTRA_BYTES : REC_N_OLD_EXTRA_BYTES));
 	ut_ad(ut_is_2pow(zip_size));
 	ut_ad(comp || !zip_size);
 
--- a/storage/innobase/log/log0online.c
+++ b/storage/innobase/log/log0online.c
@@ -501,9 +501,9 @@ log_online_make_bitmap_name(
 /*=========================*/
 	ib_uint64_t	start_lsn)	/*!< in: the start LSN name part */
 {
-	ut_snprintf(log_bmp_sys->out.name, FN_REFLEN, bmp_file_name_template,
-		    log_bmp_sys->bmp_file_home, bmp_file_name_stem,
-		    log_bmp_sys->out_seq_num, start_lsn);
+	ut_snprintf(log_bmp_sys->out.name, sizeof(log_bmp_sys->out.name), 
+            bmp_file_name_template, log_bmp_sys->bmp_file_home,
+            bmp_file_name_stem, log_bmp_sys->out_seq_num, start_lsn);
 }
 
 /*********************************************************************//**
--- a/storage/innobase/row/row0mysql.c
+++ b/storage/innobase/row/row0mysql.c
@@ -3627,7 +3627,8 @@ check_next_foreign:
 
 		row_mysql_handle_errors(&err, trx, NULL, NULL);
 
-		/* Fall through to raise error */
+		// fallthrough
+		// to raise error
 
 	default:
 		/* No other possible error returns */
@@ -4398,7 +4399,8 @@ loop:
 		fputs("  InnoDB: Warning: CHECK TABLE on ", stderr);
 		dict_index_name_print(stderr, prebuilt->trx, index);
 		fprintf(stderr, " returned %lu\n", ret);
-		/* fall through (this error is ignored by CHECK TABLE) */
+		// fallthrough
+		// (this error is ignored by CHECK TABLE)
 	case DB_END_OF_INDEX:
 func_exit:
 		mem_free(buf);
--- a/storage/innobase/row/row0purge.c
+++ b/storage/innobase/row/row0purge.c
@@ -407,7 +407,8 @@ row_purge_remove_sec_if_poss_leaf(
 				goto func_exit;
 			}
 		}
-		/* fall through (the index entry is still needed,
+		// fallthrough
+		/* (the index entry is still needed,
 		or the deletion succeeded) */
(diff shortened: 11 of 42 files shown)


percona-ysorokin:
@dutow A few stylistic / process-related comments:
We usually do 1 commit per PR. So, please squash your commits into one.
BTW, it's really great that you described everything in details - just put all those comments as sections of the commit message.

percona-ysorokin:
MySQL has coding standards. Although, I would say that they are questionable, but as we are a fork of MySQL Server, we need to stick to them. Moreover, InnoDB (everything in `storage/innobase/*`) has different standards than general SQL (everything else).
The most noticeable differences:

InnoDB uses tabs for indentation: 1 tab == 8 spaces.
Innodb has standard block indentation of 1 tab.
Innodb has opening curly braces `{` at the same line (for instance `if (cond) {`).
InnoDB has spaces before assignment operators/initialization (`int x = 8;`).
InnoDB does not like single line comments `// my comment`.
SQL uses spaces for indentation.
SQL has standard block indentation of 2 spaces.
SQL has opening curly braces `{` at anew line.
SQL has no spaces before assignment operators/initialization (`int x= 8;`).
SQL does not like single line comments `// my comment` although you may encounter them occasionally.


Everything else can be found in the code :)

percona-ysorokin:
Try to avoid reformatting of existing code (even when it's wrong). Out rule is to do minimal changes - this makes our life much easier when we merge from upstream.

storage/myisam/myisamchk.c
```diff
@@ -1365,7 +1365,7 @@ static void descript(MI_CHECK *param, register MI_INFO *info, char * name)
 	 key < share->state.header.uniques; key++, uniqueinfo++)
     {
       my_bool new_row=0;
-      char null_bit[8],null_pos[8];
+      char null_bit[8],null_pos[11];
```
percona-ysorokin, line 1368: This looks like a potential CV and should be reported to Oracle. (https://bugs.mysql.com/)

storage/innobase/log/log0online.c
```diff
@@ -98,7 +98,13 @@ static const char* bmp_file_name_stem = "ib_modified_log_";
 /** File name template for bitmap files.  The 1st format tag is a directory
 name, the 2nd tag is the stem, the 3rd tag is a file sequence number, the 4th
 tag is the start LSN for the file. */
+// There is simply no good way to handle truncation for this - if it happens,
+// it will break at more places than just this regex, and I don't think we
+// could handle it any better way than the default truncation
+#pragma GCC diagnostic push
```
percona-ysorokin: Not sure I understand the reasoning why we need this. But general comment - GCC is not the only compiler we support. Moreover, it doesn't look like MySQL code uses pragmas to suppress warnings

extra/replace.c
```diff
@@ -173,6 +173,7 @@ register char **argv[];
 	break;
       case 'V':
 	version=1;
+    // fallthrough
```
percona-ysorokin: indentation

storage/innobase/row/row0mysql.c
```diff
@@ -3627,7 +3627,8 @@ row_drop_table_for_mysql(
 
 		row_mysql_handle_errors(&err, trx, NULL, NULL);
 
-		/* Fall through to raise error */
+		// fallthrough
+        // to raise error
```
percona-ysorokin: indentation

storage/innobase/row/row0mysql.c
```diff
@@ -4398,7 +4399,8 @@ row_check_index_for_mysql(
 		fputs("  InnoDB: Warning: CHECK TABLE on ", stderr);
 		dict_index_name_print(stderr, prebuilt->trx, index);
 		fprintf(stderr, " returned %lu\n", ret);
-		/* fall through (this error is ignored by CHECK TABLE) */
+		// fallthrough
+        // (this error is ignored by CHECK TABLE)
```
percona-ysorokin: indentation

storage/innobase/row/row0purge.c
```diff
@@ -407,7 +407,8 @@ row_purge_remove_sec_if_poss_leaf(
 				goto func_exit;
 			}
 		}
-		/* fall through (the index entry is still needed,
+		// fallthrough
+        /* (the index entry is still needed,
```
percona-ysorokin: indentation

storage/innobase/include/data0type.ic
```diff
@@ -477,7 +477,8 @@ dtype_get_fixed_size_low(
 #else /* !UNIV_HOTBACKUP */
 		return(len);
 #endif /* !UNIV_HOTBACKUP */
-		/* fall through for variable-length charsets */
+		// fallthrough 
+        // for variable-length charsets
```
percona-ysorokin: indentation

sql/sp_head.cc
```diff
@@ -2888,10 +2888,10 @@ sp_head::show_routine_code(THD *thd)
     */
     if (ip != i->m_ip)
     {
-      const char *format= "Instruction at position %u has m_ip=%u";
+      const char format[] = "Instruction at position %u has m_ip=%u";
       char tmp[sizeof(format) + 2*SP_INSTR_UINT_MAXLEN + 1];
 
-      sprintf(tmp, format, ip, i->m_ip);
+      snprintf(tmp, sizeof(tmp), format, ip, i->m_ip);
```
percona-ysorokin, line 2895: This looks like another potential CV. Please report to Oracle

laurynas-biveinis:
Is this going to be upmerged or cherry-picked to higher versions? Please sync with @gl-sergei and @VasilyNemkov to avoid wasted duplicated work

percona-ysorokin:
@laurynas-biveinis Already did

storage/innobase/log/log0online.c
dutow: The reason is in the place where that format string is used.

GCC added a new warning, which tries to find possible `sprintf` overflows - and this is great, I found the `sprintf` changes you can see here thanks to that.

Its report is correct even in this case - we just can't do anything about it, and it's unlikely to happen.

This is the format string:

```
"%s%s%lu_%llu.xdb"
```

It's used with `snprintf` in a situation where:

* the output buffer is 512 byte long
* and the first parameter (for the first `%s`) is also a 512 byte long buffer

Based on these, GCC deduces that it's possible that the remaining parameters will have no space in the buffer, and since we correctly pass the same 512 as the length value for `snprintf`, the output will be truncated.

* I can't really change the size of the buffers - it isn't just 512, it uses a preprocessor definition for the maximum file name length. So it's logical for both of the buffers to use it, they are files
* Because it's also based on this definition, I can't calculate the maximum space which we could allocate for the first parameter, to truncate it in a different way, as I can't do arithmetic using the preprocessor to include the result of it inside a format string. Also, this would truncate it differently, but wouldn't remove the issue of the truncation.
* Disabling this warning globally probably isn't a good idea - it's useful in other cases.

So that's why I decided to disable it locally.

That leaves me with two options:

* disable it using pragmas. While these look like gcc specific, they will also work with clang, which is the only other compiler implementing this check
* disable it using some hack that will prevent GCC's analyzer picking up this change - for example by casting one of the parameters to a pointer instead of an array, so GCC wouldn't be able to guess it's size, and would ignore this statement.

As I considered the second option a hack, and I wanted to go with a solution that disables the warning, but does so explicitly, I went with the first option.

But if you see other options, I'm open to suggestion, I didn't really like this approach, I just couldn't find anything better.

laurynas-biveinis: The bmp_file_home buffer can be changed to 2 * FN_REFLEN + a bit, then add an ut_ad after the snprintf that the result length is < FN_REFLEN

dutow:
> Try to avoid reformatting of existing code (even when it's wrong). Out rule is to do minimal changes - this makes our life much easier when we merge from upstream.

I only reformatted code where I had to. GCC is very specific about the format and position of the fallthrough comments.

I did check my changes before submitting, and other than the innodb indentation problems (I completely missed that style - I thought it's just old unformatted code, ops), I still can't see anything that I don't have to do so we pass GCC's default fallthrough checks.

Other options include:

* Introducing a FALLTHROUGH macro, which is defined to an attribute on supported compilers, and to nothing on anywhere else (I wanted to avoid this, for example, so I didn't have to find a common include file or add includes, or anything like that)
* Changing the compiler options, so GCC's checker is more relaxed, and it accepts any comments as a reason for allowing a fallthrough (I decided against this because in some cases, that comment would be about something else)

storage/innobase/log/log0online.c
dutow: The bmp_file_home is the parameter which is too long, not the target buffer. If we want to change something, we should change log_online_bitmap_file_t.name, which I'm not sure if we should - it's header file is used at several other places, so it would require more investigation to see if it has any consequences.

laurynas-biveinis: Yes, it's the target buffer I had in mind. This is our code, not upstream, larger changes are OK here

extra/replace.c
dutow: Uh, the file is inconsistent here - at most places, it indents code using 2 spaces, but at this location, it uses tabs for some parts, and spaces for others... My original indendation matched the above line, but using spaces, not tabs, with a different tab size... I changed it to use the correct amount of spaces.

laurynas-biveinis:
Please handle the Launchpad bugs for the same: https://bugs.launchpad.net/percona-server/+bug/1729241, https://bugs.launchpad.net/percona-server/+bug/1733637. They need to be linked as duplicated, and, until we migrate to JIRA, it is the LP not BLD reference that is the canonical ID for the commit title

dutow:
I changed my branch based on the comments.

The jenkins builds are also passing:

* http://jenkins.percona.com/job/percona-server-5.5-param/1601/ - latest execution only on artful, because I had to fix the build config
* http://jenkins.percona.com/job/percona-server-5.5-param/1595/ - previous execution, without artful

client/mysql.cc
```diff
@@ -2717,7 +2717,7 @@ You can turn off this feature to get a quicker startup with -A\n\n");
         mysql_free_result(fields);
         break;
       }
-      field_names[i][num_fields*2]= '\0';
+      field_names[i][num_fields*2] = NULL;
```
laurynas-biveinis: No space before the assignment operator

sql/sql_partition.cc
```diff
@@ -7637,7 +7637,7 @@ int get_part_iter_for_interval_cols_via_map(partition_info *part_info,
                                             PARTITION_ITERATOR *part_iter)
 {
   uint32 nparts;
-  get_col_endpoint_func  get_col_endpoint;
+  get_col_endpoint_func  get_col_endpoint = NULL;
```
laurynas-biveinis: No space before the assignment operator

storage/heap/hp_extra.c
```diff
@@ -34,6 +34,7 @@ int heap_extra(register HP_INFO *info, enum ha_extra_function function)
   switch (function) {
   case HA_EXTRA_RESET_STATE:
     heap_reset(info);
+	// fallthrough
```
laurynas-biveinis: Isn't it an actual "break" missing here? Also formatting

storage/innobase/include/log0online.h
```diff
@@ -141,7 +141,11 @@ log_online_bitmap_iterator_next(
 
 /** Struct for single bitmap file information */
 struct log_online_bitmap_file_struct {
-	char		name[FN_REFLEN];	/*!< Name with full path */
+    /** Name with full path
```
laurynas-biveinis: Formatting

storage/innobase/include/page0zip.ic
```diff
@@ -170,7 +170,7 @@ page_zip_rec_needs_ext(
 				ignored if zip_size == 0 */
 	ulint	zip_size)	/*!< in: compressed page size in bytes, or 0 */
 {
-	ut_ad(rec_size > comp ? REC_N_NEW_EXTRA_BYTES : REC_N_OLD_EXTRA_BYTES);
```
laurynas-biveinis, line 173: What was the warning here? The issue here seems to be > operator having higher precedence than ?: one, not that either of REC_N_... constants must be non-zero

storage/innobase/log/log0online.c
```diff
@@ -501,9 +501,9 @@ log_online_make_bitmap_name(
 /*=========================*/
 	ib_uint64_t	start_lsn)	/*!< in: the start LSN name part */
 {
-	ut_snprintf(log_bmp_sys->out.name, FN_REFLEN, bmp_file_name_template,
-		    log_bmp_sys->bmp_file_home, bmp_file_name_stem,
-		    log_bmp_sys->out_seq_num, start_lsn);
+	ut_snprintf(log_bmp_sys->out.name, sizeof(log_bmp_sys->out.name),
```
laurynas-biveinis, line 504: Why the typecast?
Formatting
Add a debug assert that the result string fits into FN_REFLEN

storage/myisam/myisamchk.c
laurynas-biveinis: This is a a separate admin utility, thus I guess not a CV

laurynas-biveinis (changes requested):
Also please close one of the LP bugs as a duplicate of another, mention only the remaining one in the commit message.

storage/heap/hp_extra.c
dutow: Ops, you are right. I wasn't sure about it, as setting the flags during the reset also makes sense, but I checked now and in 8.0, there is a break here.

storage/innobase/log/log0online.c
dutow: It's snprintf, and I'm using sizeof - so it will fit into it.

The warning was generated by the format string, and for that, I can't add a simple assert - I can add the parts of the stringts together manually in the assert, but I'm not sure if that would be a good idea, it's too easy to forget when the pattern is changed.

The cast was just a remainder from a previous different fix for the warning, I removed it.

storage/innobase/include/page0zip.ic
dutow: You are right, that makes more sense I changed it.

dutow:
Updated Jenkins run after the changes: http://jenkins.percona.com/job/percona-server-5.5-param/1602/

storage/innobase/log/log0online.c
laurynas-biveinis: I specifically had in mind an assert that the formatted string < FN_REFLEN, not the bumped size of the buffer, but on second thought such assert is going to be useless anyway

laurynas-biveinis (approved):

laurynas-biveinis:
@dutow, please also update the LP bug: assign to yourself, per-series Fix Committed now, Fix Released once pushed to trunk, github PR links there

RoelVdP:
@dutow @percona-ysorokin @laurynas-biveinis please note this commit added a tokubackup submodule pointer - https://jira.percona.com/browse/PS-3778 - please fix asap.

PR #1987, opened 2017-11-23
lp1729241, BLD-861: Fixed compilation issues with gcc-7 for the 5.6 branch

dutow:
* Removed a statement from client/mysqlbinlog.cc which was never executed
* Corrected an empty string check in sql-common/client_authentication.cc
* Added a break to a case in sql/item_func.cc.
  Based on the code around this block, it looks logical that we do not wish to raise assertions for every string result - maybe only on error?
* Added overridden new and delete operators for Relay_log_info and Slave_worker because they require 64 byte alignment. These implementation currently only work on Linux, and they do contain a few lines of duplicated code.
* Added a missing break to sql/rpl_slave.cc.
  Based on the code around it, that's what makes sense.
* Fixed an uninitailized read during debug printing in sql/sql_cache.cc.
  This change was backported from 5.7.
* Null initialized a local pointer in sql/sql_partition.cc.
  As it otherwise could result in uninitailized reads in release builds.
  On debug builds, this case is guarded by an assert.
* Added a missing return to storage/innobase/row/row0log.cc
  This change was backported from 5.7.
* Corrected the length of a local sprintf buffer in storage/myisam/myisamchk.c.
* Returning the success status in tests/bug25714.c.
  This way the variable is also used in the release build, and the correctness of the execution can also be checked using the exit status of the program.
* Fixed two sprintf local buffer overflow in regex/debug.c.
* Default initialzied a local variable in storage/innobase/page/page0page.cc.
  As it otherwise could result in uninitailized reads in release builds.
  On debug builds, this case is guarded by an assert.
* Fixed some unused result or non-null warnings.
  These do not result in any behaviour change.
* Changed bool conversion to be more explict in some assertions.
  These do not result in any behaviour change.
* Added or changed fallthrough comments at many places so GCC's parser will be able to identify them.
  These do not result in any behaviour change.
* Moved a few breaks to different locations so GCC's parser will correctly identify them.
  These do not result in any behaviour change.

### DIFF
--- a/sql/rpl_rli.cc
+++ b/sql/rpl_rli.cc
@@ -2493,3 +2493,17 @@ void Relay_log_info::adapt_to_master_version(Format_description_log_event *fdle)
     }
   }
 }
+
+void* Relay_log_info::operator new(size_t request)
+{
+  void* ptr;
+  if (posix_memalign(&ptr, __alignof__(Relay_log_info), sizeof(Relay_log_info))) {
+    throw std::bad_alloc();
+  }
+  return ptr;
+}
+
+void Relay_log_info::operator delete(void * ptr) {
+  free(ptr);
+}
+
--- a/sql/rpl_rli_pdb.cc
+++ b/sql/rpl_rli_pdb.cc
@@ -1668,6 +1668,20 @@ void Slave_worker::do_report(loglevel level, int err_code, const char *msg,
   c_rli->va_report(level, err_code, buff_coord, msg, args);
 }
 
+void* Slave_worker::operator new(size_t request)
+{
+  void* ptr;
+  if (posix_memalign(&ptr, __alignof__(Slave_worker), sizeof(Slave_worker))) {
+    throw std::bad_alloc();
+  }
+  return ptr;
+}
+
+void Slave_worker::operator delete(void * ptr)
+{
+  free(ptr);
+}
+
 /**
    Function is called by Coordinator when it identified an event
    requiring sequential execution.
--- a/sql/sql_acl.cc
+++ b/sql/sql_acl.cc
@@ -2987,7 +2987,7 @@ bool change_password(THD *thd, const char *host, const char *user,
 
   mysql_mutex_assert_owner(&acl_cache->lock);
   table->use_all_columns();
-  DBUG_ASSERT(host != '\0');
+  DBUG_ASSERT(host != NULL && host[0] != '\0');
   table->field[MYSQL_USER_FIELD_HOST]->store(host, strlen(host),
                                              system_charset_info);
   table->field[MYSQL_USER_FIELD_USER]->store(user, strlen(user),
@@ -3424,7 +3424,7 @@ update_user_table(THD *thd, TABLE *table,
   if (!is_user_table_positioned)
   {
     table->use_all_columns();
-    DBUG_ASSERT(host != '\0');
+    DBUG_ASSERT(host != NULL && host[0] != '\0');
     table->field[MYSQL_USER_FIELD_HOST]->store(host, (uint) strlen(host),
 					       system_charset_info);
     table->field[MYSQL_USER_FIELD_USER]->store(user, (uint) strlen(user),
@@ -3573,7 +3573,7 @@ static int replace_user_table(THD *thd, TABLE *table, LEX_USER *combo,
   }
 
   table->use_all_columns();
-  DBUG_ASSERT(combo->host.str != '\0');
+  DBUG_ASSERT(combo->host.str != NULL && combo->host.str[0] != '\0');
   table->field[MYSQL_USER_FIELD_HOST]->store(combo->host.str,combo->host.length,
                                              system_charset_info);
   table->field[MYSQL_USER_FIELD_USER]->store(combo->user.str,combo->user.length,
@@ -3664,7 +3664,7 @@ static int replace_user_table(THD *thd, TABLE *table, LEX_USER *combo,
 
     old_row_exists = 0;
     restore_record(table,s->default_values);
-    DBUG_ASSERT(combo->host.str != '\0');
+    DBUG_ASSERT(combo->host.str != NULL && combo->host.str[0] != '\0');
     table->field[MYSQL_USER_FIELD_HOST]->store(combo->host.str,combo->host.length,
                                                system_charset_info);
     table->field[MYSQL_USER_FIELD_USER]->store(combo->user.str,combo->user.length,
@@ -12602,7 +12602,13 @@ private:
         filesize= ftell(key_file);
         fseek(key_file, 0, SEEK_SET);
         *key_text_buffer= new char[filesize+1];
-        (void) fread(*key_text_buffer, filesize, 1, key_file);
+        const size_t read_size = fread(*key_text_buffer, filesize, 1, key_file);
+        if (read_size != 1)
+        {
+          delete[] key_text_buffer;
+          key_text_buffer= NULL;
+          return true;
+        }
         (*key_text_buffer)[filesize]= '\0';
       }
       fclose(key_file);
--- a/storage/innobase/page/page0page.cc
+++ b/storage/innobase/page/page0page.cc
@@ -2832,7 +2832,7 @@ page_warn_strict_checksum(
 	ulint				space_id,
 	ulint				page_no)
 {
-	srv_checksum_algorithm_t	curr_algo_nonstrict;
+	srv_checksum_algorithm_t	curr_algo_nonstrict = srv_checksum_algorithm_t();
 	switch (curr_algo) {
 	case SRV_CHECKSUM_ALGORITHM_STRICT_CRC32:
 		curr_algo_nonstrict = SRV_CHECKSUM_ALGORITHM_CRC32;
--- a/storage/tokudb/tokudb_thread.h
+++ b/storage/tokudb/tokudb_thread.h
@@ -204,21 +204,21 @@ inline mutex_t::mutex_t(pfs_key_t key) {
     _owners = 0;
     _owner = _null_owner;
 #endif
-    int r = mysql_mutex_init(key, &_mutex, MY_MUTEX_INIT_FAST);
+    int  r MY_ATTRIBUTE((unused)) = mysql_mutex_init(key, &_mutex, MY_MUTEX_INIT_FAST);
     assert_debug(r == 0);
 }
 inline mutex_t::~mutex_t(void) {
 #ifdef TOKUDB_DEBUG
     assert_debug(_owners == 0);
 #endif
-    int r = mysql_mutex_destroy(&_mutex);
+    int  r MY_ATTRIBUTE((unused)) = mysql_mutex_destroy(&_mutex);
     assert_debug(r == 0);
 }
 inline void mutex_t::reinit(pfs_key_t key) {
 #ifdef TOKUDB_DEBUG
     assert_debug(_owners == 0);
 #endif
-    int r;
+    int  r MY_ATTRIBUTE((unused));
     r = mysql_mutex_destroy(&_mutex);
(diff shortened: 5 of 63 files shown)


sql/rpl_rli.cc
```diff
@@ -2493,3 +2493,19 @@ void Relay_log_info::adapt_to_master_version(Format_description_log_event *fdle)
     }
   }
 }
+
+void* Relay_log_info::operator new(size_t request)
```
laurynas-biveinis, line 2497: @gl-sergei has fixed this in a different way in XB (IIRC the alignment is not required in the first place)

sql/rpl_rli_pdb.cc
```diff
@@ -1668,6 +1668,22 @@ void Slave_worker::do_report(loglevel level, int err_code, const char *msg,
   c_rli->va_report(level, err_code, buff_coord, msg, args);
 }
 
+void* Slave_worker::operator new(size_t request)
```
laurynas-biveinis: Likewise

laurynas-biveinis (changes requested):

laurynas-biveinis:
This also looks like it has no relation with the 5.5 version of the same (if it's a cherry-pick, then please use the -x flag to record the source commit). Please check Confluence on the workflow

sql/rpl_rli.cc
dutow: He didn't fix this, he removed the alignment requirement from a struct. So that fixes the compilation error but results in a different alignment for these classes, even when allocated on the stack.

My version uses the correct, specified alignment, and I think that's better.

If we need to, I can introduce a helper function instead of these two small functions, which will also be able to handle other platforms - but I checked our supported platform list, and we only support linux.

(The alignment specifier is to align to cache lines, so I think it's there to increase the performance)

dutow:
It isn't a cherry-pick, because while it contains similar changes, it's also different: some changes present in the 5.5 version are missing here, and some new changes are added.

So while it has some relation, it isn't the same - so including the other commit message here wouldn't be a good idea. That message contains the changes I did for 5.5, and this message contains the changes I did for 5.6.

I'll edit this commit message to include the other's commit id after that's accepted.

sql/rpl_rli.cc
laurynas-biveinis: Upstream 8.0 has removed it too, IIRC because the alignment was not shown to have any perf improvement and was always broken (?). If your version is correct, should it be restored in XtraBackup too? Please reach consensus with @gl-sergei.

If this is the correct fix, then check whether this code is not compiled with -fno-exceptions and whether is not supposed to return NULL instead of throwing

laurynas-biveinis:
Our workflow is to upmerge or cherrypick (depending on whether merge is possible), and edit the commits, commit messages etc. as necessary.

sql/rpl_rli.cc
dutow: I didn't check 8.0 before - now I did, and they indeed removed it, but for a different reason: a sanitizer found that they used it without the correct alignment (as mentioned in the commit message: https://github.com/mysql/mysql-server/commit/ebc4a60a46a3077a20ce346dee07bdea2c936bfc ) - and my fix corrects that, after this change, it'll be correctly 64 bit aligned every time.

If it would be compiled without exceptions, that throw would generate an error - and it doesn't, so I think it isn't, but I'll check it to be sure.

dutow:
Updated the PR, the jenkins jobs are stillrunning, but most of them already succeeded: http://jenkins.percona.com/job/percona-server-5.6-param/2021/

sql/rpl_rli.cc
laurynas-biveinis: You are right, this is compiled with exceptions, in which case I am not sure what to do with the caller, which checks for NULL instead of catching the exception.

For 8.0, this will have to be merged with re-enabling the alignment or the fix will be lost

sql/rpl_rli.cc
```diff
@@ -2493,3 +2493,19 @@ void Relay_log_info::adapt_to_master_version(Format_description_log_event *fdle)
...
+{
+  // Note: this only work on linux, but since that's, all what we support 
+  void* ptr;
+  if(posix_memalign(&ptr, __alignof__(Relay_log_info), sizeof(Relay_log_info))) {
```
laurynas-biveinis: Space after if

sql/rpl_rli.cc
```diff
@@ -2493,3 +2493,19 @@ void Relay_log_info::adapt_to_master_version(Format_description_log_event *fdle)
     }
   }
 }
+
+void* Relay_log_info::operator new(size_t request)
+{
+  // Note: this only work on linux, but since that's, all what we support
```
laurynas-biveinis: Redundant comment

sql/rpl_rli.cc
```diff
@@ -2493,3 +2493,19 @@ void Relay_log_info::adapt_to_master_version(Format_description_log_event *fdle)
...
+  // Note: this only work on linux, but since that's, all what we support 
+  void* ptr;
+  if(posix_memalign(&ptr, __alignof__(Relay_log_info), sizeof(Relay_log_info))) {
+    throw std::bad_alloc();
+  }
+  return ptr;
+}
+
+void Relay_log_info::operator delete(void * ptr) {
+  // Just to be sure, in case for some reaseon the default delete defaults to something else
```
laurynas-biveinis: Redundant comment

sql/rpl_rli_pdb.cc
```diff
@@ -1668,6 +1668,22 @@ void Slave_worker::do_report(loglevel level, int err_code, const char *msg,
...
+{
+  // Note: this only work on linux, but since that's, all what we support
```
laurynas-biveinis: Same three comments

sql/rpl_rli.cc
dutow: Uh, I didn't check that. But that wouldn't work even with the standard new, as that also throws bad_alloc.

I could change it to return NULL, but the standard asks for bad_alloc in this case...

Maybe I could change the factory to catch the exception?

I could also try to submit this patch to mysql, maybe they would accept it if it passes the sanitizer tests?

sql/sql_acl.cc
```diff
@@ -12602,8 +12602,8 @@ class Rsa_authentication_keys
         filesize= ftell(key_file);
         fseek(key_file, 0, SEEK_SET);
         *key_text_buffer= new char[filesize+1];
-        (void) fread(*key_text_buffer, filesize, 1, key_file);
-        (*key_text_buffer)[filesize]= '\0';
+        const size_t read_size = fread(*key_text_buffer, filesize, 1, key_file);
```
laurynas-biveinis: Wouldn't it be more correct to delete the buffer on failure and return true?

storage/innobase/dict/dict0load.cc
```diff
@@ -1483,10 +1483,11 @@ dict_load_column_low(
 		goto err_len;
 	}
 
-	if (!column) {
+	if (!column && table) {
```
dutow: I just realized that this change causes failures, as unstable means that tests are failing, it's not just a warning.

I'll change back this line and instead of this change, modify the function declaration so that it accepts nulls.

storage/innobase/page/page0page.cc
```diff
@@ -2832,7 +2832,7 @@ page_warn_strict_checksum(
 	ulint				space_id,
 	ulint				page_no)
 {
-	srv_checksum_algorithm_t	curr_algo_nonstrict;
+	srv_checksum_algorithm_t	curr_algo_nonstrict = srv_checksum_algorithm_t();
```
laurynas-biveinis, line 2835: OK, but I wonder why any warning is produced in the first place - the switch below is exhaustive. Maybe ut_error is not annotated as "does not return"?

(but please don't spend non-trivial amount of time investigating)

storage/innobase/dict/dict0load.cc
```diff
@@ -1483,10 +1483,11 @@ dict_load_column_low(
...
 		dict_mem_table_add_col(table, heap, name, mtype,
 				       prtype, col_len);
-	} else {
+    }
+    if (!column) {
```
laurynas-biveinis: Is this really correct? Not "if (column)" ? What was the warning fixed?

storage/innobase/dict/dict0load.cc
```diff
@@ -1483,10 +1483,11 @@ dict_load_column_low(
 		goto err_len;
 	}
 
-	if (!column) {
+	if (!column && table) {
 		dict_mem_table_add_col(table, heap, name, mtype,
 				       prtype, col_len);
-	} else {
+    }
```
laurynas-biveinis: Wrong formatting

storage/tokudb/tokudb_thread.h
```diff
@@ -204,21 +204,21 @@ inline mutex_t::mutex_t(pfs_key_t key) {
     _owners = 0;
     _owner = _null_owner;
 #endif
-    int r = mysql_mutex_init(key, &_mutex, MY_MUTEX_INIT_FAST);
+    int  r __attribute__((__unused__)) = mysql_mutex_init(key, &_mutex, MY_MUTEX_INIT_FAST);
```
laurynas-biveinis: ```s/__attribute__/MY_ATTRIBUTE/g```
@georgelorchpercona , I am quite certain there was legacy/current JIRA issue re. this compilation warning, but I am unable to find it?

laurynas-biveinis (changes requested):

storage/innobase/dict/dict0load.cc
laurynas-biveinis: Yep, I had a related comment in my review too

storage/innobase/page/page0page.cc
dutow: The warning is later when the variable is used, and only appears in release mode. The ut_error macro expands to nothing in release builds, does nothing, and then the function used an uninitialized variable.

While it isn't necessarily better now, we at least have a consistent behavior, without using some memory garbage. I wasn't sure how we could implement a release-mode error here, as it's just an assertion failure even in debug builds.

sql/rpl_rli.cc
laurynas-biveinis: While returning NULL would "work", I don't like introducing broken code to be compatible with other broken code. While we are not a diverging fork and thus any patch has to be evaluated for its maintenance cost (which usually excludes cleanup patches, refactorings, etc), in this particular case catching the exception in the caller seems OK.

Re. upstream submission, a patch to replace NULL-checking with exception catching could be submitted, but it can take up to forever for them to merge it. For 8.0, restoring alignment is a performance patch, they usually want to see some evidence of performance impact

storage/tokudb/tokudb_thread.h
dutow: I've used ```__attribute__``` because it's in our own code, and in it's repository, tokudb uses this syntax.

So in the mysql repo, even for our own code, I should use the mysql conventions?

dutow:
New jenkins run, this time with successful test results: http://jenkins.percona.com/job/percona-server-5.6-param/2025/

I changed everything, except the attributes, I left those in their original state for now.

storage/tokudb/tokudb_thread.h
laurynas-biveinis: Yes, unless we have reasons to do otherwise. It's still MySQL code.

laurynas-biveinis (approved):
LGTM, Toku bits need @georgelorchpercona approval

george-lorch (approved):
LGTM

laurynas-biveinis:
This might have caused https://bugs.launchpad.net/percona-server/+bug/1739734, @dutow, please investigate

PR #1988, opened 2017-11-26
Implement bp:innodb-fts-ngram-ignore-stopword-list (A new InnoDB variable to control whether InnoDB FTS NGRAM parser should ignore stopword list)

percona-ysorokin:
(https://blueprints.launchpad.net/percona-server/+spec/innodb-fts-ngram-ignore-stopword-list)

lp:1679135 "Ignore INNODB_FT_DEFAULT_STOPWORD for ngram indexes"
(https://bugs.launchpad.net/percona-server/+bug/1679135)

Bug #84420 "stopwords and ngram indexes"
(https://bugs.mysql.com/bug.php?id=84420)

Thanks to Sveta Smirnova (sveta.smirnova@percona.com) for original
contribution.

Introduced new 'innodb_ft_ignore_ngram_check' InnoDB session/global variable
which can be set to 'ON' to instruct InnoDB Full Text Search NGRAM parser to
ignore stopword list.

Added new 'innodb_fts.percona_ft_ignore_ngram_check' MTR test case to check
stopword list ignoring logic.

Added new 'sys_vars.innodb_ft_ignore_ngram_check_basic' MTR test case to check
the behavior of the introduced variable.

### DIFF
--- a/mysql-test/suite/innodb_fts/t/percona_ft_ignore_stopwords.test
+++ b/mysql-test/suite/innodb_fts/t/percona_ft_ignore_stopwords.test
@@ -1,0 +1,121 @@
+--source include/have_innodb.inc
+
+if (`SELECT @@global.innodb_ft_min_token_size <> 3`)
+{
+  --skip this test requires innodb_ft_min_token_size to have default value (3)
+}
+
+--echo #
+--echo # bp:innodb-fts-ngram-ignore-stopword-list
+--echo # "A new InnoDB variable to control whether InnoDB FTS should ignore stopword list"
+--echo #
+--echo # lp:1679135 "Ignore INNODB_FT_DEFAULT_STOPWORD for ngram indexes"
+--echo # Bug #84420 "stopwords and ngram indexes"
+--echo #
+
+SET innodb_ft_ignore_stopwords = 0;
+
+CREATE TABLE articles (
+  id INT(10) UNSIGNED NOT NULL AUTO_INCREMENT,
+  body TEXT,
+  PRIMARY KEY (id),
+  FULLTEXT KEY ftx (body) WITH PARSER ngram
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+INSERT INTO articles(body) VALUES('ea');
+INSERT INTO articles(body) VALUES('east');
+INSERT INTO articles(body) VALUES('east area');
+INSERT INTO articles(body) VALUES('east job');
+INSERT INTO articles(body) VALUES('eastnation');
+INSERT INTO articles(body) VALUES('eastway, try try');
+INSERT INTO articles(body) VALUES('wa');
+INSERT INTO articles(body) VALUES('wast');
+INSERT INTO articles(body) VALUES('wast area');
+INSERT INTO articles(body) VALUES('wast job');
+INSERT INTO articles(body) VALUES('wastnation');
+INSERT INTO articles(body) VALUES('wastway, try try');
+INSERT INTO articles(body) VALUES('oz');
+INSERT INTO articles(body) VALUES('ozst');
+INSERT INTO articles(body) VALUES('ozst area');
+INSERT INTO articles(body) VALUES('ozst job');
+INSERT INTO articles(body) VALUES('ozstnation');
+INSERT INTO articles(body) VALUES('ozstway, try try');
+
+--echo # ngram parser checks
+--let $match_short_query = SELECT COUNT(*) FROM articles WHERE MATCH(body) AGAINST("ea" IN BOOLEAN MODE)
+
+--let $assert_text= 'Empty set (ngram, table created with ft_ignore_stopwords = 0 / query run with ft_ignore_stopwords = 0)'
+--let $assert_cond= [$match_short_query] = 0
+--source include/assert.inc
+
+SET innodb_ft_ignore_stopwords = 1;
+--let $assert_text= 'Empty set (ngram, table created with ft_ignore_stopwords = 0 / query run with ft_ignore_stopwords = 1)'
+--let $assert_cond= [$match_short_query] = 0
+--source include/assert.inc
+
+ALTER TABLE articles DROP INDEX ftx;
+ALTER TABLE articles ADD FULLTEXT KEY ftx(body) WITH PARSER ngram;
+--let $assert_text= '8 matches (ngram, table changed from ft_ignore_stopwords = 0 to 1 / query run with ft_ignore_stopwords = 1)'
+--let $assert_cond= [$match_short_query] = 8
+--source include/assert.inc
+
+SET innodb_ft_ignore_stopwords = 0;
+--let $assert_text= 'Empty set (ngram, table changed from ft_ignore_stopwords = 0 to 1 / query run with ft_ignore_stopwords = 0)'
+--let $assert_cond= [$match_short_query] = 0
+--source include/assert.inc
+
+ALTER TABLE articles DROP INDEX ftx;
+ALTER TABLE articles ADD FULLTEXT KEY ftx(body) WITH PARSER ngram;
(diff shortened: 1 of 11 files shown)


percona-ysorokin:
Jenkins build link
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1329

laurynas-biveinis (changes requested):
Needs a blueprint
It looks that the variable would be more usable as a session/global rather than global only

percona-ysorokin:
Variable renamed to `innodb_ft_ignore_ngram_check` to be more consistent with other FT control variables.
It has also been made session/global rather just global.
Updated Jenkins build link
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1348

mysql-test/suite/innodb_fts/t/percona_ft_ignore_ngram_check.test
```diff
@@ -0,0 +1,77 @@
...
+
+#
+# bp:innodb-fts-ngram-ignore-stopword-list
+# "A new InnoDB variable to control whether InnoDB FTS NGRAM parser should ignore stopword list"
+#
+# lp:1679135 "Ignore INNODB_FT_DEFAULT_STOPWORD for ngram indexes"
+# Bug #84420 "stopwords and ngram indexes"
+#
+
+SET innodb_ft_ignore_ngram_check = 0;
```
laurynas-biveinis: We need to rename this variable: the current name "innodb_ft_ignore_ngram_check" is not clear at all, does not mention anything about stopwords ("what ngram checks are ignored?")

innodb_ft_ngram_use_stopwords? (default TRUE)

storage/innobase/fts/fts0fts.cc
```diff
@@ -1224,11 +1224,13 @@ fts_tokenizer_word_get(
...
 	ut_ad(rw_lock_own(&cache->lock, RW_LOCK_X));
 
+	ut_ad(current_thd != NULL);
 	/* If it is a stopword, do not index it */
 	if (!fts_check_token(text,
 		    cache->stopword_info.cached_stopword,
 		    index_cache->index->is_ngram,
-		    index_cache->charset)) {
+		    index_cache->charset,
+		    thd_has_ft_ignore_ngram_check(current_thd))) {
```
laurynas-biveinis: Likewise for the function name

storage/innobase/fts/fts0fts.cc
```diff
@@ -4835,17 +4837,23 @@ or greater than fts_max_token_size.
 @param[in]	stopwords	stopwords rb tree
 @param[in]	is_ngram	is ngram parser
 @param[in]	cs		token charset
+@param[in]	skip		true if the check should be skipped
```
laurynas-biveinis, line 4840: This function is called for non ngram tokens too. What happens there if the variable is set?

storage/innobase/include/ha_prototypes.h
```diff
@@ -293,6 +293,13 @@ thd_lock_wait_timeout(
 /*==================*/
 	THD*	thd);	/*!< in: thread handle, or NULL to query
 			the global innodb_lock_wait_timeout */
+
+/** Is FT ignore ngram check set.
+@param thd Thread object
+@return true if ignore_ngram_check is set, false otherwise. */
+bool
```
laurynas-biveinis: warn_unused_result

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1850,6 +1854,16 @@ thd_lock_wait_timeout(
 	return(THDVAR(thd, lock_wait_timeout));
 }
 
+/** Is FT ignore ngram check set.
+@param thd Thread object
+@return true if ignore_ngram_check is set, false otherwise. */
+bool
+thd_has_ft_ignore_ngram_check(THD* thd)
```
laurynas-biveinis: Has to be renamed after the variable rename

storage/innobase/include/row0ftsort.h
```diff
@@ -117,6 +117,11 @@ struct fts_tokenize_ctx {
 	dfield_t		sort_field[FTS_NUM_FIELDS_SORT];
 						/*!< in: sort field */
 	fts_token_list_t	fts_token_list;
+	bool			ignore_ngram_check;
```
laurynas-biveinis: Has to be renamed after the system variable rename

laurynas-biveinis (changes requested):

storage/innobase/fts/fts0fts.cc
percona-ysorokin: This variable also affects non-NGRAM FTS. It has absolutely the same meaning - if it's enabled, FTS won't be checking if the current token is a stopword when building/updating an FTS index.
However, being a stopword doesn't just mean to be a one of the predefined words from the list. Tokens shorter than `innodb_ft_min_token_size`  or longer than `innodb_ft_max_token_size` are also considered stopwords.
In the case when we set our new variable to `ON`, `innodb_ft_min_token_size`/`innodb_ft_max_token_size` will be ignored meaning that very short and very long words will also be indexed.

Because of that, I also suggest to remove `ngram` part from the variable name.

mysql-test/suite/innodb_fts/t/percona_ft_ignore_ngram_check.test
percona-ysorokin: Agree that the name should include `stopwords` part. However, don't like the idea of reversing the value. My reasoning are the following:
If the default value is FALSE and we need this option to be enabled via command line, we can start `mysqld --innodb_ft_ignore_stopwords=ON` or just `mysqld --innodb_ft_ignore_stopwords`
whereas when the default is TRUE, the only way to change it at startup is
`mysqld --innodb_ft_ngram_use_stopwords=OFF` because the short form `mysqld --innodb_ft_ngram_use_stopwords` will mean no changes.

percona-ysorokin:
New variable renamed to `innodb_ft_ignore_stopwords`.
Extended MTR test cases for non-NGRAM parsers.
Jenkins build link
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1358/

mysql-test/suite/innodb_fts/t/percona_ft_ignore_stopwords.test
```diff
@@ -0,0 +1,121 @@
+--source include/have_innodb.inc
+
+if(`SELECT @@global.innodb_ft_min_token_size <> 3`)
```
laurynas-biveinis: Space after if

storage/innobase/handler/ha_innodb.cc
```diff
@@ -903,6 +903,10 @@ static MYSQL_THDVAR_STR(tmpdir,
   "Directory for temporary non-tablespace files.",
   innodb_tmpdir_validate, NULL, NULL);
 
+static MYSQL_THDVAR_BOOL(ft_ignore_stopwords, PLUGIN_VAR_OPCMDARG,
+  "Disable stop words checks for ngram indexes",
```
laurynas-biveinis: Outdated description: should mention FTS, should not mention ngram

laurynas-biveinis (changes requested):
Minor comments, full Jenkins is not required
The blueprint must be updated to describe non-ngram case too

percona-ysorokin:
New Jenkins link
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1362/

laurynas-biveinis (approved):
The blueprint still must be updated for non-ngram case. Strictly speaking the new variable description there is already not specific to ngrams, but given that the only use case there involves there, it's too easy to miss when documenting

percona-ysorokin:
Blueprint updated with the description of non-ngram use case.

PR #1989, opened 2017-11-28
lp1702330, PS55: unlock the mutex before returning from thread_group_close.

dutow:
Jenkins: http://jenkins.percona.com/job/percona-server-5.5-param/1603/

Some builds are failing because the change is based on an old commit.


laurynas-biveinis (approved):
For a Jenkins run, make sure the URL has the exact job

percona-ysorokin (approved):
LGTM

PR #1992, opened 2017-12-01
lp1729241, BLD-860: Fixed compilation issues with gcc-7 for the 5.7 branch

dutow:
This commit is based on 59eff2e4e870f3a87ade9d4b55dea5d47a2ba62b, which contains the comilation fixes for the 5.6 branch.

* Updated the PerconaFT and Percona-TokuBackup submodules with newer versions.
  The new version contain compilation fixes for those submodules.
* Corrected an empty string check in sql-common/client_authentication.cc.
* Changed a local sprintf buffer size in sql/sys_vars.h.
* Added a missing break to the mysqld parameter checks in sql/mysqld.cc.
* Modified an assertion in libbinlogevents/src/statement_events.cpp.
  In it's previous form itw as always true, and also caused integer-bool conversion warnings.
* Added breaks in mysqlxtest_src/mysqlx_protocol.cc.
  These do not change the flow of the program, they are only needed because the compiler was unable to detect throws in called functions.
* Added a break to a case in sql/item_func.cc.
  Based on the code around this block, it looks logical that we do not wish to raise assertions for every string result - maybe only on error?
* Fixed an incorrect if in sql/item_geofunc.cc.
  Half of the condition was missing.
* Added overridden new and delete operators for Relay_log_info and Slave_worker because they require 64 byte alignment.
  These implementation currently only work on Linux, and they do contain a few lines of duplicated code.
* Added a missing break to sql/rpl_slave.cc.
  Based on the code around it, that's what makes sense.
* Added a missing break to heap/hp_extra.c
  This change is backported from 8.0
* Null initialized a local pointer in sql/sql_partition.cc.
  As it otherwise could result in uninitailized reads in release builds.
  On debug builds, this case is guarded by an assert.
* Disabled fallthrough warnings for rocksdb, as it isn't our repository.
* Corrected the length of a local sprintf buffer in storage/myisam/myisamchk.c.
* Returning the success status in tests/bug25714.c.
  This way the variable is also used in the release build, and the correctness of the execution can also be checked using the exit status of the program.
* Fixed two sprintf local buffer overflow in regex/debug.c.
* Default initialzied a local variable in storage/innobase/page/page0page.cc.
  As it otherwise could result in uninitailized reads in release builds.
  On debug builds, this case is guarded by an assert.
* Fixed some unused result or non-null warnings.
  These do not result in any behaviour change.
* Changed bool conversion to be more explict in some assertions.
  These do not result in any behaviour change.
* Added or changed fallthrough comments at many places so GCC's parser will be able to identify them.
  These do not result in any behaviour change.
* Moved a few breaks to different locations so GCC's parser will correctly identify them.
  These do not result in any behaviour change.

### DIFF
--- a/libbinlogevents/src/statement_events.cpp
+++ b/libbinlogevents/src/statement_events.cpp
@@ -528,12 +528,9 @@ User_var_event(const char* buf, unsigned int event_len,
   size_t bytes_read= ((val + val_len) - start);
 #ifndef DBUG_OFF
   bool old_pre_checksum_fd= description_event->is_version_before_checksum();
-  bool checksum_verify= (old_pre_checksum_fd ||
-                         (description_event->footer()->checksum_alg ==
-                          BINLOG_CHECKSUM_ALG_OFF));
-  size_t data_written= (header()->data_written- checksum_verify);
-  BAPI_ASSERT(((bytes_read == data_written) ? 0 : BINLOG_CHECKSUM_LEN)||
-              ((bytes_read == data_written - 1) ? 0 : BINLOG_CHECKSUM_LEN));
+  const int checksum_length = (old_pre_checksum_fd || description_event->footer()->checksum_alg == BINLOG_CHECKSUM_ALG_OFF) ? 0 : BINLOG_CHECKSUM_LEN;
+  size_t data_written= (header()->data_written- checksum_length);
+  BAPI_ASSERT(bytes_read == data_written || (bytes_read == data_written - 1));
 #endif
     if ((header()->data_written - bytes_read) > 0)
     {
--- a/rapid/plugin/x/mysqlxtest_src/mysqlx_protocol.cc
+++ b/rapid/plugin/x/mysqlxtest_src/mysqlx_protocol.cc
@@ -159,7 +159,7 @@ bool mysqlx::parse_mysql_connstring(const std::string &connstring,
   return true;
 }
 
-static void throw_server_error(const Mysqlx::Error &error)
+MY_ATTRIBUTE((noreturn)) static void throw_server_error(const Mysqlx::Error &error)
 {
   throw Error(error.code(), error.msg());
 }
--- a/storage/innobase/row/row0log.cc
+++ b/storage/innobase/row/row0log.cc
@@ -2037,6 +2037,7 @@ row_log_table_apply_update(
 
 		When applying the subsequent ROW_T_DELETE, no matching
 		record will be found. */
+		// fallthrough
 	case DB_SUCCESS:
 		ut_ad(row != NULL);
 		break;
--- a/storage/rocksdb/CMakeLists.txt
+++ b/storage/rocksdb/CMakeLists.txt
@@ -34,6 +34,15 @@ IF ((CMAKE_CXX_COMPILER_ID STREQUAL GNU) AND
   RETURN ()
 ENDIF ()
 
+# Relax fallthrough warnings we have no control over, as it's in zstd
+IF(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  MY_CHECK_CXX_COMPILER_FLAG("-Wimplicit-fallthrough=0" HAVE_IMPLICIT_FALLTHROUGH)
+  IF(HAVE_IMPLICIT_FALLTHROUGH)
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wimplicit-fallthrough=0")
+    SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wimplicit-fallthrough=0")
+  ENDIF()
+ENDIF()
+
 # check that compiler supports cxx11 and set options
 INCLUDE (check_stdcxx11)
 IF (!HAVE_STDCXX11)
--- a/unittest/gunit/opt_range-t.cc
+++ b/unittest/gunit/opt_range-t.cc
@@ -351,7 +351,9 @@ void OptRangeTest::check_use_count(SEL_TREE *tree)
   {
     SEL_ARG *cur_range= tree->keys[i];
     if (cur_range != NULL)
+    {
       EXPECT_FALSE(cur_range->test_use_count(cur_range));
+    }
   }
 
   if (!tree->merges.is_empty())
--- a/client/mysqlbinlog.cc
+++ b/client/mysqlbinlog.cc
@@ -1317,6 +1317,7 @@ Exit_status process_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
       
       destroy_evt= TRUE;
     }
+    // fallthrough
           
     case binary_log::INTVAR_EVENT:
     {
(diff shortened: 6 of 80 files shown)


dutow:
Jenkins: http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1360/

Edit: restarted the jenkins job

dutow:
Changed the warning disabling code for rocksdb: it caused issues with older compilers, now we only disable it if the compiler supports the warning.

http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1361/

libbinlogevents/src/statement_events.cpp
```diff
@@ -532,8 +532,7 @@ User_var_event(const char* buf, unsigned int event_len,
                          (description_event->footer()->checksum_alg ==
                           BINLOG_CHECKSUM_ALG_OFF));
   size_t data_written= (header()->data_written- checksum_verify);
-  BAPI_ASSERT(((bytes_read == data_written) ? 0 : BINLOG_CHECKSUM_LEN)||
-              ((bytes_read == data_written - 1) ? 0 : BINLOG_CHECKSUM_LEN));
+  BAPI_ASSERT(bytes_read == data_written || bytes_read == data_written);
```
laurynas-biveinis: X == X || X == X ?
Also I don't quite understand the original intent of the assert, where is BINLOG_CHECKSUM_LEN supposed to go?

rapid/plugin/x/mysqlxtest_src/mysqlx_protocol.cc
```diff
@@ -557,6 +557,7 @@ void XProtocol::authenticate_mysql41(const std::string &user, const std::string
 
       case Mysqlx::ServerMessages::ERROR:
         throw_server_error(*static_cast<Mysqlx::Error*>(message.get()));
+        break;
```
laurynas-biveinis: How about marking throw_server_error as a noreturn function?
(although I am not sure why the compiler does not figure it out itself, is a static function in the same compilation unit)

storage/innobase/row/row0log.cc
```diff
@@ -2037,6 +2037,7 @@ row_log_table_apply_update(
 
 		When applying the subsequent ROW_T_DELETE, no matching
 		record will be found. */
+		return(DB_SUCCESS);
```
laurynas-biveinis: This changes the behavior of the function, are you sure?
Also space before (

storage/rocksdb/CMakeLists.txt
```diff
@@ -34,6 +34,15 @@ IF ((CMAKE_CXX_COMPILER_ID STREQUAL GNU) AND
   RETURN ()
 ENDIF ()
 
+# Relax fallthrough warnings we have no control over, as it's in zstd
```
laurynas-biveinis, line 37: @georgelorchpercona, we need to remember to remove this once the upstream fixes those warnings

unittest/gunit/opt_range-t.cc
```diff
@@ -351,7 +351,9 @@ void OptRangeTest::check_use_count(SEL_TREE *tree)
   {
     SEL_ARG *cur_range= tree->keys[i];
     if (cur_range != NULL)
+    {
```
laurynas-biveinis, line 354: What warning was fixed here?

laurynas-biveinis (changes requested):

libbinlogevents/src/statement_events.cpp
dutow: Ops, I wanted to write data_written -1 for the second part.

Based on the code (and comment) above, I think this assert wants to ensure that we had read the same amount as we have written, or maybe one byte more in some cases.

storage/innobase/row/row0log.cc
dutow: No, I messed up the conflict resolution somehow here - I fixed this locally and will push soon.

unittest/gunit/opt_range-t.cc
dutow: The macro below this contains an if-else without brackets, so it generated a warning about ambiguity.

libbinlogevents/src/statement_events.cpp
dutow: But now that I checked it again, maybe BINLOG_CHECKSUM_LEN should be included in this calculation, I'm not really sure :( I will try to test it.

Unfortunately, the original condition always evaluates to true... (or rather to BINLOG_CHECKSUM_LEN)

Edit: The 8.0 sources use true instead of the BINLOG_CHECKSUM_LEN, but the assert still always evaluates to true.

dutow: I ran the tests with my version, and it results in more failures - I could keep it in the state as it's in 8.0 (which compiles, but always evaluates to true), or maybe  I could simply remove it?

laurynas-biveinis: Looking at the 8.0 version - ```BAPI_ASSERT(((bytes_read == data_written) ? false : true) ||
((bytes_read == data_written - 1) ? false : true));``` - it is not obvious to me why would it always evaluate to true?

laurynas-biveinis:
If your final version shows unexpected Jenkins failures, compare with https://bugs.launchpad.net/bugs/1736028

dutow:
New Jenkins run (some variants still in progress): https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1373/

libbinlogevents/src/statement_events.cpp
dutow: If (bytes_read == data_written - 1) is true, then the first check has to be false, and since the ? : expression negates it, it results in a true result.
If (bytes_read == data_written) is true, the same happens in the other direction: the other check will be false, and the expression will negate it.

It's basically assert(a == X || a != X) in a bit more complicated way.

But I found out what this expression wants to check based on the 5.6 branch, I'm testing it currently.

dutow:
I updated the buffer length assertion the same way I changed it for the 5.6 build. I also changed the artful agents the same way as for the 5.6 build.

New jenkins link, just started: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1381/

Locally the rpl tests passed, those reported failures previously because of that assert.

Edit: changed the jenkins job, I made a typo in the project template.

rapid/plugin/x/mysqlxtest_src/mysqlx_protocol.cc
```diff
@@ -159,6 +159,8 @@ bool mysqlx::parse_mysql_connstring(const std::string &connstring,
   return true;
 }
 
+static void throw_server_error(const Mysqlx::Error &error) MY_ATTRIBUTE((noreturn));
+
 static void throw_server_error(const Mysqlx::Error &error)
```
laurynas-biveinis: Can't you just merge declaration and definition?

laurynas-biveinis (approved):
OK if Jenkins OK, with or without my suggested change

rapid/plugin/x/mysqlxtest_src/mysqlx_protocol.cc
dutow: Looks like I can, I just had to add MY_ATTRIBUTE earlier. I changed it.

PR #1997, opened 2017-12-06
lp:1722789 Intermediary slave does not log master changes with binlog…

vlad-lesin:
…_rows_query_log_events

Let's consider the following function:

int ha_blackhole::update_row(const uchar *old_data, uchar *new_data)
{
  DBUG_ENTER("ha_blackhole::update_row");
  THD *thd= ha_thd();
  if (is_slave_applier(thd) && thd->query().str == NULL)
    DBUG_RETURN(0);
  DBUG_RETURN(HA_ERR_WRONG_COMMAND);
}

In the case if this function is invoked from slave thread and it's
query string is empty, then it's supposed the function finishes successfully,
otherwise some error is returned. The same logic exists in some other blackhole
engine functions.

If "binlog_rows_query_log_events = 1" binary log contains events with queries.
The queries are ignored, but during "Rows_query_log_event" event execution
(see Rows_query_log_event::do_apply_event()) thread query string is filled with
the query from the event. And when then Update_rows_log_event::do_exec_row()
is executed, sql thread has query in thd->m_query_string, and this condition:

  is_slave_applier(thd) && thd->query().str == NULL

becomes false, what leads to errors during execution of blackhole engine
functions such as ha_blackhole::update_row(). Such errors are output to error
log and sql thread stops, as a result is't binary log is not filling.

The question is what does this
"is_slave_applier(thd) && thd->query().str == NULL" condition mean? I suppose
the condition means that the currently processed relay log event type is not
QUERY_EVENT.

That is why  I would propose the following fix:
replace "thd->query().str == NULL" with
"thd->rli_slave->rows_query_ev || thd->query() == NULL".

See also: https://github.com/percona/percona-server/pull/1998.
Testing: https://jenkins.percona.com/job/percona-server-5.6-param/2026/

### DIFF
--- a/storage/blackhole/ha_blackhole.cc
+++ b/storage/blackhole/ha_blackhole.cc
@@ -20,11 +20,18 @@
 #include "probes_mysql.h"
 #include "ha_blackhole.h"
 #include "sql_class.h"                          // THD, SYSTEM_THREAD_SLAVE_*
+#include "rpl_rli.h" // THD::rli_slave::rows_query_ev
 
-static bool is_slave_applier(THD *thd)
+static inline bool is_slave_applier(const THD &thd)
 {
-  return thd->system_thread == SYSTEM_THREAD_SLAVE_SQL ||
-    thd->system_thread == SYSTEM_THREAD_SLAVE_WORKER;
+  return thd.system_thread == SYSTEM_THREAD_SLAVE_SQL ||
+    thd.system_thread == SYSTEM_THREAD_SLAVE_WORKER;
+}
+
+static inline bool pretend_for_slave(const THD &thd)
+{
+  return is_slave_applier(thd) &&
+    (thd.rli_slave->rows_query_ev || thd.query() == NULL);
 }
 
 /* Static declarations for handlerton */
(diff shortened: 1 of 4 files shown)


storage/blackhole/ha_blackhole.cc
```diff
@@ -120,7 +121,8 @@ int ha_blackhole::update_row(const uchar *old_data, uchar *new_data)
 {
   DBUG_ENTER("ha_blackhole::update_row");
   THD *thd= ha_thd();
-  if (is_slave_applier(thd) && thd->query() == NULL)
+  if (is_slave_applier(thd) &&
```
laurynas-biveinis: Please extract this condition to a helper method

laurynas-biveinis (changes requested):
FWIW, the original condition might have been added for this (commit message of 43fb8633b0774f00dce6490fd93deebdfe4ab386):

"    It is necessary to only pretend this for the SQL thread, since
    a SELECT executed on the Blackhole engine will otherwise never
    return EOF, causing a livelock."

But it is not clear why it is not enough to check the thread type only

storage/blackhole/ha_blackhole.cc
```diff
@@ -20,6 +20,7 @@
 #include "probes_mysql.h"
 #include "ha_blackhole.h"
 #include "sql_class.h"                          // THD, SYSTEM_THREAD_SLAVE_*
+#include "rpl_rli.h"
```
laurynas-biveinis: Also please add a comment ```// THD::rli_slave::rows_query_ev```

storage/blackhole/ha_blackhole.cc
vlad-lesin: fixed

storage/blackhole/ha_blackhole.cc
vlad-lesin: fixed

vlad-lesin:
Ready for the next round.

laurynas-biveinis (approved):
Please also send this upstream

PR #1998, opened 2017-12-06
Rpl ibi lp 1722789 5.7 gca

vlad-lesin:
See https://github.com/percona/percona-server/pull/1997.
Testing:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1372/ - there were errors for some platforms due to jenkins nodes configuration update issues,
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1377/ - here is the same testing restart.


laurynas-biveinis (changes requested):
Same comments as for 5.6

laurynas-biveinis (approved):

PR #1999, opened 2017-12-06
lp-1733049: Blackhole : Specified key was too long; max key length is…

vlad-lesin:
… 1000 bytes

The maximum innodb key length is 3500 what is hardcoded in
ha_innobase::max_supported_key_length()). The maximum number of innodb indexes
is configured with MAX_INDEXES macro (see also MAX_KEY definition).

The same is currently implemented for blackhole storage engine.

See also: https://github.com/percona/percona-server/pull/2000, https://github.com/percona/percona-server/pull/2001.
Testing: https://jenkins.percona.com/job/percona-server-5.5-param/1604/

### DIFF
--- a/storage/blackhole/ha_blackhole.h
+++ b/storage/blackhole/ha_blackhole.h
@@ -20,6 +20,7 @@
 #include "thr_lock.h"                           /* THR_LOCK */
 #include "handler.h"                            /* handler */
 #include "table.h"                              /* TABLE_SHARE */
+#include "sql_const.h"                          /* MAX_KEY */
 
 /*
   Shared structure for correct LOCK operation
@@ -68,9 +69,9 @@ public:
             HA_READ_ORDER | HA_KEYREAD_ONLY);
   }
   /* The following defines can be increased if necessary */
-#define BLACKHOLE_MAX_KEY	64		/* Max allowed keys */
+#define BLACKHOLE_MAX_KEY	MAX_KEY		/* Max allowed keys */
 #define BLACKHOLE_MAX_KEY_SEG	16		/* Max segments for key */
-#define BLACKHOLE_MAX_KEY_LENGTH 1000
+#define BLACKHOLE_MAX_KEY_LENGTH	3500		/* Like in InnoDB */
   uint max_supported_keys()          const { return BLACKHOLE_MAX_KEY; }
   uint max_supported_key_length()    const { return BLACKHOLE_MAX_KEY_LENGTH; }
   uint max_supported_key_part_length() const { return BLACKHOLE_MAX_KEY_LENGTH; }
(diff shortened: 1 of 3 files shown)


storage/blackhole/ha_blackhole.h
```diff
@@ -20,7 +20,7 @@
 #include "thr_lock.h"                           /* THR_LOCK */
 #include "handler.h"                            /* handler */
 #include "table.h"                              /* TABLE_SHARE */
-
+#include "sql_const.h"
```
laurynas-biveinis: ```/* MAX_KEY */```

storage/blackhole/ha_blackhole.h
```diff
@@ -68,9 +68,9 @@ class ha_blackhole: public handler
             HA_READ_ORDER | HA_KEYREAD_ONLY);
   }
   /* The following defines can be increased if necessary */
-#define BLACKHOLE_MAX_KEY	64		/* Max allowed keys */
+#define BLACKHOLE_MAX_KEY	MAX_KEY		/* Max allowed keys */
 #define BLACKHOLE_MAX_KEY_SEG	16		/* Max segments for key */
-#define BLACKHOLE_MAX_KEY_LENGTH 1000
+#define BLACKHOLE_MAX_KEY_LENGTH 3500 /* Like in innodb */
```
laurynas-biveinis: Align the comment with ones above, s/innodb/InnoDB

mysql-test/t/blackhole-bug53588.test
```diff
@@ -0,0 +1,15 @@
+#
```
laurynas-biveinis: Either just bug53588.test, or just extend the existing main.blackhole test (make sure to use an --echo # header block like other bugs in that file and put the snippet above the "end of 5.5 tests" marker)

laurynas-biveinis (changes requested):
Minor comments

storage/blackhole/ha_blackhole.h
vlad-lesin: I have fixed it. But, honestly, I don't understand this rule. Ok, now this header file is used here only for the only constant. But, in the future somebody can use something declared in this header without even understanding it. And so what? Must the comment be changed? Why?

storage/blackhole/ha_blackhole.h
vlad-lesin: Aligned.

vlad-lesin:
Ready for the next review.

storage/blackhole/ha_blackhole.h
laurynas-biveinis: You are right, this is useful only if it survives possible future changes, which is hard, and better managed by tools like include-what-you-use. Let's consider this optional going forward.

laurynas-biveinis (approved):

laurynas-biveinis:
Please also send this upstream

PR #2000, opened 2017-12-06
Blackhole max key length 5.6 gca

vlad-lesin:
See: https://github.com/percona/percona-server/pull/1999, https://github.com/percona/percona-server/pull/2001
Testing: https://jenkins.percona.com/job/percona-server-5.6-param/2024/


laurynas-biveinis (changes requested):
Same comments as for 5.5

laurynas-biveinis (approved):

PR #2001, opened 2017-12-06
Blackhole max key length 5.7 gca

vlad-lesin:
See: https://github.com/percona/percona-server/pull/1999, https://github.com/percona/percona-server/pull/2000.
Testing: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1353/


laurynas-biveinis (changes requested):
Same comments as for 5.5

laurynas-biveinis (approved):

PR #2003, opened 2017-12-06
MYR-212 : fix rocksdb_block_cachecompressed_* names to be rocksdb_blo…

george-lorch:
…ck_cache_compressed_*

- Did just as title says, also fixed getupdatessince to get_updates_since and
  re-recorded rocksdb.rocksdb.test

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -10897,9 +10897,9 @@ struct rocksdb_status_counters_t {
   uint64_t bloom_filter_prefix_checked;
   uint64_t bloom_filter_prefix_useful;
   uint64_t number_reseeks_iteration;
-  uint64_t getupdatessince_calls;
-  uint64_t block_cachecompressed_miss;
-  uint64_t block_cachecompressed_hit;
+  uint64_t get_updates_since_calls;
+  uint64_t block_cache_compressed_miss;
+  uint64_t block_cache_compressed_hit;
   uint64_t wal_synced;
   uint64_t wal_bytes;
   uint64_t write_self;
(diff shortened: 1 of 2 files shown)


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1379/

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10897,9 +10897,9 @@ struct rocksdb_status_counters_t {
   uint64_t bloom_filter_prefix_checked;
   uint64_t bloom_filter_prefix_useful;
   uint64_t number_reseeks_iteration;
-  uint64_t getupdatessince_calls;
-  uint64_t block_cachecompressed_miss;
-  uint64_t block_cachecompressed_hit;
+  uint64_t get_updates_since_calls;
```
laurynas-biveinis, line 10900: getupdatessince was referred to as "GetUpdatesSince" in the MyRocks status var doc PR - is that an actual name in RocksDB? In which case it should be left as-is. And the request to fix cachecompressed was based on a premise that there is no any e.g. "CacheCompressed" in RocksDB?

george-lorch: All of these values are extracted from rocksdb via an API and identifiers which are 'mapped' into MySQL status variables. The identifiers for these are GET_UPDATES_SINCE_CALLS, BLOCK_CACHE_COMPRESSED_MISS, BLOCK_CACHE_COMPRESSED_HIT.

The text for the documentation was scraped from comments within the code and so could easily have been inconsistent.

laurynas-biveinis (approved):
Please merge after the 5.7 release branch is created

PR #2006, opened 2017-12-07
lp1736905: Fixing assertions modified during the artful compilation fix

dutow:
During the artful compilation fixes, an assertion checking that some pointers aren't equal to an empty string were corrected:
these asserts now checked that the pointers aren't NULL pointers and that they do not point to an empty string.

This more restrictive address caused assertation violations where it shouldn't be. This commit changes them back to the more relaxed behaviors of the original code: it only checks that the pointers aren't NULL pointers.

                                                                                                                                                  
Another assertion in log_event.cc had a similar issue, where the changed assertion didn't generate any warnings, but resulted in assertion errors.
Here the original assert always evaluated to true, and as such, had no function.                                                                  
Additional brackets were required to resolve the evaluation order, butt he original solution resulted in assertion failures.                      
This improved version correctly checks the original intention of the assertion.

### DIFF
--- a/sql/log_event.cc
+++ b/sql/log_event.cc
@@ -7838,18 +7838,9 @@ User_var_log_event(const char* buf, uint event_len,
     uint bytes_read= ((val + val_len) - start);
 #ifndef DBUG_OFF
     bool old_pre_checksum_fd= description_event->is_version_before_checksum();
+    const int checksum_length = (old_pre_checksum_fd || description_event->checksum_alg == BINLOG_CHECKSUM_ALG_OFF) ? 0 : BINLOG_CHECKSUM_LEN;
 #endif
-    DBUG_ASSERT((bytes_read == (data_written -
-                 ((old_pre_checksum_fd ||
-                  (description_event->checksum_alg ==
-                   BINLOG_CHECKSUM_ALG_OFF)) ?
-                 0 : BINLOG_CHECKSUM_LEN)))
-                ||
-                ((bytes_read == (data_written -1 -
-                 (old_pre_checksum_fd ||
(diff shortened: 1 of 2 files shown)


dutow:
Jenkins: https://jenkins.percona.com/job/percona-server-5.6-param/2028/

laurynas-biveinis (commented):
The code looks OK, but Jenkins run results not, especially Artful ones, but all platforms on Debug. Is this something caused by this PR or we missed a regression in one of the previous ones?

dutow:
I've missed one of the hosts assertions, and main.grant still failed. Just pushed the change now and restarted the failed jenkins builds: https://jenkins.percona.com/job/percona-server-5.6-param/2029/

Edit: for the artful failures, those are caused by missing packages required by the tests. I'll edit the config and rerun them again.

percona-ysorokin:
I think we should also add a new MTR test case as, surprisingly enough, this simple situation wasn't covered by any existing test from the whole suite.

dutow:
@percona-ysorokin it was, I just missed it in the jenkins report page :( For some reason my previous execution didn't report that several test failed in the summary page, only in the build log - which I didn't check...

dutow:
Logs for artful: https://jenkins.percona.com/job/percona-server-5.6-param/2030/

* Previously there was a failure on jessie, which disappeared now... I have no idea why, I wasn't able to reproduce that failure locally so far.
* Artful reports an unstable result, but if you check the logs, it has no failures / errors in it - that warning is only generated by the junit reporter, because for some reason junit.xml isn't an xml at all

laurynas-biveinis (approved):

PR #2008, opened 2017-12-07
PS-260: The mysqld release build crashed when loading the world database

dutow:
btr_cur_optimistic_insert called
btr_cur_ins_lock_and_undo which called
trx_undo_report_row_operation

Because these calls had no conditions on the thr variable, and trx_undo_report_row_operation marked thr as non null, the optimizer assumed it to be non null in other conditions too.

Based on this, the compiler optimized out some thr && conditions, resulting in null pointer crashes.

trx_undo_report_row_operation can handle null thr values with the correct flags, and this attribute was already removed in the 5.7 branch.

### DIFF
--- a/storage/innobase/include/trx0rec.h
+++ b/storage/innobase/include/trx0rec.h
@@ -233,7 +233,7 @@ trx_undo_report_row_operation(
 					inserted undo log record,
 					0 if BTR_NO_UNDO_LOG
 					flag was specified */
-	MY_ATTRIBUTE((nonnull(3,4,10), warn_unused_result));
+	MY_ATTRIBUTE((nonnull(4,10), warn_unused_result));
 /******************************************************************//**
 Copies an undo record to heap. This function can be called if we know that
 the undo log record exists.


percona-ysorokin:
A simple MTR test case would be a nice addition to this fix.
It can be as simple as trying to import `world.sql` into the database and make sure it doesn't crash.

mysql-test/t/world_database.test
```diff
@@ -0,0 +1,17 @@
+--source include/have_innodb.inc
```
laurynas-biveinis: While this bug needs a testcase, just trying to load world database is not a very specific one. Is it possible to reduce it?

The testcase should go the InnoDB suite, and probably named percona_ps_260.test or similar

laurynas-biveinis (changes requested):
The trx0rec.h bit can be committed separately to unblock the release

mysql-test/t/world_database.test
dutow: During my tests, it was consistently the 181st city that crashed the server, and that's the first table that was created/filled. I could reduce the sql file to include only entries up to that point, but I don't know if that would make it more specific?

dutow:
I updated the PR to only include the fix.

laurynas-biveinis (approved):
Please also report the issue upstream (after the testcase and release work)

PR #2015, opened 2017-12-11
lp1716844: Escaping control characters in the audit log

dutow:
Implemented escaping for the JSON output format, and updated the test suite.

The other output formats are left unchanged, as they do not have a well defined method for escaping these characters.
XML does support control characters in version 1.1, but most tools only understand 1.0, and our output is 1.0.

### DIFF
--- a/mysql-test/include/audit_log_events.inc
+++ b/mysql-test/include/audit_log_events.inc
@@ -6,6 +6,42 @@ CREATE TABLE t1
              (c1 INT,
               c2 CHAR(20));
 INSERT INTO t1 VALUES (1,'a'),(2,'b'),(3,'c');
+-- if ($test_control_chars) {
+INSERT INTO `t1` VALUES (4,NULL);
+# can't add the zero ascii character, as that's a syntax error in MySQL
+INSERT INTO `t1` VALUES (6,'');
+INSERT INTO `t1` VALUES (7,'');
+INSERT INTO `t1` VALUES (8,'');
+INSERT INTO `t1` VALUES (9,'');
+INSERT INTO `t1` VALUES (10,'');
+INSERT INTO `t1` VALUES (11,'');
+INSERT INTO `t1` VALUES (12,'');
+INSERT INTO `t1` VALUES (13,'');
+INSERT INTO `t1` VALUES (14,'	');
+INSERT INTO `t1` VALUES (15,'
+');
+INSERT INTO `t1` VALUES (16,'');
+INSERT INTO `t1` VALUES (17,'');
+INSERT INTO `t1` VALUES (18,'');
+INSERT INTO `t1` VALUES (19,'');
+INSERT INTO `t1` VALUES (20,'');
+INSERT INTO `t1` VALUES (21,'');
+INSERT INTO `t1` VALUES (22,'');
+INSERT INTO `t1` VALUES (23,'');
+INSERT INTO `t1` VALUES (24,'');
+INSERT INTO `t1` VALUES (25,'');
+INSERT INTO `t1` VALUES (26,'');
+INSERT INTO `t1` VALUES (27,'');
+INSERT INTO `t1` VALUES (28,'');
+INSERT INTO `t1` VALUES (29,'');
+INSERT INTO `t1` VALUES (30,'');
+INSERT INTO `t1` VALUES (31,'');
+INSERT INTO `t1` VALUES (32,'');
+INSERT INTO `t1` VALUES (33,'');
+INSERT INTO `t1` VALUES (34,'');
+INSERT INTO `t1` VALUES (35,'');
+INSERT INTO `t1` VALUES (36,'');
+-- }
 SELECT * FROM t1;
 --error ER_NO_SUCH_TABLE
 SELECT * FROM t2;
--- a/mysql-test/t/audit_log_json.test
+++ b/mysql-test/t/audit_log_json.test
@@ -7,6 +7,7 @@ SET GLOBAL audit_log_flush=ON;
 --remove_file $MYSQLD_DATADIR/test_audit.log
 SET GLOBAL audit_log_flush=ON;
 
+--let $test_control_chars=1;
 --source include/audit_log_events.inc
 
 --move_file $MYSQLD_DATADIR/test_audit.log $MYSQLD_DATADIR/test_audit_json.log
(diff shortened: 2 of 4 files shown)


plugin/audit_log/audit_log.c
```diff
@@ -238,6 +271,39 @@ void json_escape(const char *in, size_t *inlen, char *out, size_t *outlen)
     { '\b',  2, "\\b" },
     { '\f',  2, "\\f" },
     { '\t',  2, "\\t" },
+    // escaping basic ASCII control characters
+    // skipping 0: it can't be handled by the current algorithm
```
percona-ysorokin: The same comment, plus
`\b` and `\f`

plugin/audit_log/audit_log.c
```diff
@@ -219,6 +219,39 @@ void xml_escape(const char *in, size_t *inlen, char *out, size_t *outlen)
...
+    // skipping 0: it can't be represented in a valid XML document
+    { 1,  8, "\\u#x0001" },
+    { 2,  8, "\\u#x0002" },
+    { 3,  8, "\\u#x0003" },
+    { 4,  8, "\\u#x0004" },
+    { 5,  8, "\\u#x0005" },
+    { 8,  8, "\\u#x0006" },
+    { 7,  8, "\\u#x0007" },
+    { 8,  8, "\\u#x0008" },
+    { 9,  8, "\\u#x0009" },
```
percona-ysorokin: Is this OK that we have duplicates in this table?
`'\t'` == `9`
`'\n'` == `10`
`'\r'` == `13`

dutow: I didn't want to delete the original entries (to reduce the chance of merge conflicts later), but I also thought that adding every entry to this list would be clearer than skipping those records.

Technically, it won't cause any problems, as the escape code will stop as soon as it finds an entry in this list. If I remove one of the variants, it could end up a bit faster, but now we are dealing with quite larger arrays than before (around 40 entries vs the 7-8 original). 

If this is a performance critical part of the server (I thought it isn't that critical, as it looks like a verbose logging feature), I could implement this differently, but that wouldn't be just the adding of a few lines to the existing escaping code, but changing parts of the actual escape loop.

laurynas-biveinis: This is fully our code. Merge conflicts should not be a concern

plugin/audit_log/audit_log.c
```diff
@@ -219,6 +219,39 @@ void xml_escape(const char *in, size_t *inlen, char *out, size_t *outlen)
...
     { '\t', 5, "&#9;" },
     { '"',  6, "&quot;" },
+    // escaping basic ASCII control characters
+    // skipping 0: it can't be represented in a valid XML document
+    { 1,  8, "\\u#x0001" },
+    { 2,  8, "\\u#x0002" },
+    { 3,  8, "\\u#x0003" },
+    { 4,  8, "\\u#x0004" },
+    { 5,  8, "\\u#x0005" },
+    { 8,  8, "\\u#x0006" },
```
laurynas-biveinis: 6 ?

mysql-test/include/audit_log_events.inc
```diff
@@ -6,6 +6,8 @@ CREATE TABLE t1
              (c1 INT,
               c2 CHAR(20));
 INSERT INTO t1 VALUES (1,'a'),(2,'b'),(3,'c');
+INSERT INTO `t1` VALUES (4,NULL);
+INSERT INTO `t1` VALUES (5,'541S?');
```
laurynas-biveinis: Probably wouldn't hurt to test the escaped character set exhaustively.

laurynas-biveinis (changes requested):

mysql-test/include/audit_log_events.inc
dutow: The tests currently are far from exhaustive (for example, the actual xml or json output isn't tested, only to the point that it is valid), these statements here only verify the SQL statements that were executed when creating the database.

So unless I also add some actual output verification, exhaustive testing wouldn't accomplish much more. Should I do that?

laurynas-biveinis: It's just 31 characters - if sorted in input, recording the manual outputs of \01 \02 ... etc would be enough

plugin/audit_log/audit_log.c
dutow: In this case, I updated the PR based on my original idea: I introduced a new array for the control characters, so we won't end up in a 40 element loop for each character.

plugin/audit_log/audit_log.c
dutow: Fixed.

dutow:
Updates:

* I removed the xml escapes, as that's only supported by XML 1.1, and we are generating 1.0 output. Many tools only support 1.0, so it's probably a good idea to keep using that.
* I added a more extensive test for the json output
* Control character tests are only included in the json output                                     
* Added unsigned casts to a few places, as signed chars caused bugs

dutow:
Jenkins run: https://jenkins.percona.com/job/percona-server-5.6-param/2034/

mysql-test/t/audit_log_json.test
```diff
@@ -7,17 +7,41 @@ SET GLOBAL audit_log_flush=ON;
 --remove_file $MYSQLD_DATADIR/test_audit.log
 SET GLOBAL audit_log_flush=ON;
 
+--let $TEST_CONTROL_CHARS=1;
```
laurynas-biveinis: Uppercase vars in MTR become environment vars too, lowercase ones don't. Since you don't need an env var here (e.g. for interacting with inline Perl), use lowercase. This comment on its own would be too minor to force a re-spin of this PR

plugin/audit_log/audit_log.c
```diff
@@ -186,16 +187,29 @@ void escape_buf(const char *in, size_t *inlen, char *out, size_t *outlen,
   while ((in < inend) && (out < outend))
   {
     replaced= FALSE;
-    for (rule= escape_rules; rule->character; rule++)
-    {
-      if (*in == rule->character)
-      {
+    if ((unsigned char)(*in) < 32) {
```
laurynas-biveinis: Why a typecast is needed here?

plugin/audit_log/audit_log.c
```diff
@@ -186,16 +187,29 @@ void escape_buf(const char *in, size_t *inlen, char *out, size_t *outlen,
...
+      if (control_escape_rules[(unsigned int)*in].character) {
```
laurynas-biveinis: Likewise. If this one is because of sign mismatch, is it possible to make the ```in``` arg ```const unsigned char *``` instead?

plugin/audit_log/audit_log.c
```diff
@@ -186,16 +187,29 @@ void escape_buf(const char *in, size_t *inlen, char *out, size_t *outlen,
...
         replaced= TRUE;
-        break;
+      }
+    } else
+    {
+      for (rule= other_escape_rules; rule->character; rule++)
+      {
+        if (*in == rule->character)
+        {
+          if ((outend - out) < (int) rule->length)
```
laurynas-biveinis: Why a typecast here? If it has to stay, wouldn't it be more exact to use ptrdiff_t?

mysql-test/t/audit_log_json.test
```diff
@@ -7,17 +7,41 @@ SET GLOBAL audit_log_flush=ON;
...
+          my @query_chars = sort($query =~ /./sg);
+          my $minimum_character = ord($query_chars[0]);
+          if ($minimum_character == 1) {
+              $found_1st_control_char = 1;
+          }
+          if ($found_1st_control_char && $control_char_count < 31) {
+              $control_char_count = $control_char_count + 1;
+              my $expected = $last_control_char + 1;
+              if ($expected != $minimum_character) {
+                   print "Incorrent control character in output: Expected $expected, got $minimum_character\n";
```
laurynas-biveinis: "Incorrect"

laurynas-biveinis (changes requested):

plugin/audit_log/audit_log.c
dutow: Because by default char is signed in gcc. Even if I specify the parameter as unsigned, I have to add this cast somewhere.

plugin/audit_log/audit_log.c
dutow: I could eliminate this cast by specifying less than space, but I'm not sure if that would be clearer.

dutow:
Updated the PR: 
* fixed the typo
* removed some duplicated code I added (more refactoring)
* changed one of the casts (int->ptrdiff_t)

laurynas-biveinis:
Have you tested on Jenkins again?

dutow:
Yes: https://jenkins.percona.com/job/percona-server-5.6-param/2037/

(I had to change a variable declaration, because we don't compile with C99 on older platforms)

laurynas-biveinis (approved):

laurynas-biveinis:
@dutow, we usually merge when when all version PRs are ready, in this case the 5.7 one.

dutow:
Ops, I completely forgot about that one :( I'll open the 5.7 PR too, it's basically the same change.

PR #2016, opened 2017-12-11
Null-merge branch 5.6 (commit ed2382f) into 5.7

percona-ysorokin:
This is a null merge after merging upstream 5.6.38 and 5.7.20.


percona-ysorokin:
Self-approving

PR #2022, opened 2017-12-14
PS-3767: Fixed clang 4/5 warnings.

dutow:
* Removed unneccessary parantheses from conditions
* Disabled the tautological constant out of range warning (static asserts would require c++11)
* Fixed a const char* -> char* conversion
* Changed null checks of arrays to compile time assertions that they are arrays
* Added some explicit casts
* Removed unused functions / variables
* Marked functions which are only used in some configurations as possibly unused

### DIFF
--- a/cmd-line-utils/readline/history.c
+++ b/cmd-line-utils/readline/history.c
@@ -211,14 +211,14 @@ history_get (offset)
 
 HIST_ENTRY *
 alloc_history_entry (string, ts)
-     char *string;
+     const char *string;
      char *ts;
 {
   HIST_ENTRY *temp;
 
   temp = (HIST_ENTRY *)xmalloc (sizeof (HIST_ENTRY));
 
-  temp->line = string ? savestring (string) : string;
+  temp->line = string ? savestring (string) : NULL;
   temp->data = (char *)NULL;
   temp->timestamp = ts;
 
--- a/extra/yassl/include/openssl/crypto.h
+++ b/extra/yassl/include/openssl/crypto.h
@@ -19,7 +19,7 @@
 
 /* crypto.h for openSSL */
 
-#ifndef ysSSL_crypto_h__
+#ifndef yaSSL_crypto_h__
 #define yaSSL_crypto_h__
 
 #ifdef YASSL_PREFIX
--- a/libmysql/libmysql.c
+++ b/libmysql/libmysql.c
@@ -1379,7 +1379,8 @@ void set_stmt_errmsg(MYSQL_STMT *stmt, NET *net)
   DBUG_ASSERT(stmt != 0);
 
   stmt->last_errno= net->last_errno;
-  if (net->last_error && net->last_error[0])
+  compile_time_assert(sizeof(net->last_error) / sizeof(void*) > 1);
+  if (net->last_error[0])
     strmov(stmt->last_error, net->last_error);
   strmov(stmt->sqlstate, net->sqlstate);
 
--- a/sql/item.cc
+++ b/sql/item.cc
@@ -7035,7 +7035,9 @@ bool Item_default_value::fix_fields(THD *thd, Item **items)
   }
   if (!(def_field= (Field*) sql_alloc(field_arg->field->size_of())))
     goto error;
-  memcpy(def_field, field_arg->field, field_arg->field->size_of());
+  memcpy(static_cast<void*>(def_field),
+         static_cast<void*>(field_arg->field),
+         field_arg->field->size_of());
   def_field->move_field_offset((my_ptrdiff_t)
                                (def_field->table->s->default_values -
                                 def_field->table->record[0]));
@@ -7172,7 +7174,9 @@ bool Item_insert_value::fix_fields(THD *thd, Item **items)
     Field *def_field= (Field*) sql_alloc(field_arg->field->size_of());
     if (!def_field)
       return true;
-    memcpy(def_field, field_arg->field, field_arg->field->size_of());
+    memcpy(static_cast<void*>(def_field),
+           static_cast<void*>(field_arg->field),
+           field_arg->field->size_of());
     def_field->move_field_offset((my_ptrdiff_t)
                                  (def_field->table->insert_values -
                                   def_field->table->record[0]));
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -1236,8 +1236,9 @@ bool LOGGER::slow_log_print(THD *thd, const char *query, uint query_length,
     }
 
     /* fill in user_host value: the format is "%s[%s] @ %s [%s]" */
+    compile_time_assert(sizeof(sctx->priv_user) / sizeof(void*) > 1);
     user_host_len= (strxnmov(user_host_buff, MAX_USER_HOST_SIZE,
-                             sctx->priv_user ? sctx->priv_user : "", "[",
+                             sctx->priv_user, "[",
                              sctx->user ? sctx->user : (thd->slave_thread ? "SQL_SLAVE" : ""), "] @ ",
                              sctx->get_host()->length() ?
                              sctx->get_host()->ptr() : "", " [",
@@ -2240,7 +2241,7 @@ static int find_uniq_filename(char *name, ulong *next, bool need_next)
   my_dirend(dir_info);
 
   /* check if reached the maximum possible extension number */
-  if ((max_found == MAX_LOG_UNIQUE_FN_EXT))
+  if (max_found == MAX_LOG_UNIQUE_FN_EXT)
   {
     sql_print_error("Log filename extension number exhausted: %06lu. \
 Please fix this by archiving old logs and \
--- a/sql/log_event.cc
+++ b/sql/log_event.cc
@@ -1164,7 +1164,7 @@ failed my_b_read"));
   Log_event *res=  0;
 #ifndef max_allowed_packet
   THD *thd=current_thd;
-  uint max_allowed_packet= thd ? slave_max_allowed_packet:~(ulong)0;
+  ulong max_allowed_packet= thd ? slave_max_allowed_packet:~(ulong)0;
 #endif
 
   if (data_len > max_allowed_packet)
@@ -2400,8 +2400,9 @@ bool Query_log_event::write(IO_CACHE* file)
       user= thd->get_invoker_user();
       host= thd->get_invoker_host();
     }
-    else if (thd->security_ctx->priv_user)
+    else
     {
+      compile_time_assert(sizeof(thd->security_ctx->priv_user) / sizeof(void*) > 1);
       Security_context *ctx= thd->security_ctx;
 
       user.length= strlen(ctx->priv_user);
--- a/sql/slave.cc
+++ b/sql/slave.cc
@@ -4109,8 +4109,8 @@ static int queue_event(Master_info* mi,const char* buf, ulong event_len)
 
        TODO: handling `when' for SHOW SLAVE STATUS' snds behind
     */
-    if ((memcmp(mi->master_log_name, hb.get_log_ident(), hb.get_ident_len())
-         && mi->master_log_name != NULL)
+    compile_time_assert(sizeof(mi->master_log_name) / sizeof(void*) > 1);
+    if (memcmp(mi->master_log_name, hb.get_log_ident(), hb.get_ident_len())
         || mi->master_log_pos != hb.log_pos)
     {
       /* missed events of heartbeat from the past */
@@ -4368,7 +4368,8 @@ static int connect_to_master(THD* thd, MYSQL* mysql, Master_info* mi,
     mysql_options(mysql, MYSQL_PLUGIN_DIR, opt_plugin_dir_ptr);
 
   /* we disallow empty users */
-  if (mi->user == NULL || mi->user[0] == 0)
+  compile_time_assert(sizeof(mi->user) / sizeof(void*) > 1);
+  if (mi->user[0] == 0)
   {
     mi->report(ERROR_LEVEL, ER_SLAVE_FATAL_ERROR,
                ER(ER_SLAVE_FATAL_ERROR),
@@ -4504,8 +4505,8 @@ MYSQL *rpl_connect_master(MYSQL *mysql)
   /* This one is not strictly needed but we have it here for completeness */
   mysql_options(mysql, MYSQL_SET_CHARSET_DIR, (char *) charsets_dir);
 
-  if (mi->user == NULL
-      || mi->user[0] == 0
+  compile_time_assert(sizeof(mi->user) / sizeof(void*) > 1);
+  if (mi->user[0] == 0
       || io_slave_killed(thd, mi)
       || !mysql_real_connect(mysql, mi->host, mi->user, mi->password, 0,
                              mi->port, 0, 0))
--- a/sql/sql_acl.cc
+++ b/sql/sql_acl.cc
@@ -172,6 +172,7 @@ TABLE_FIELD_TYPE mysql_db_table_fields[MYSQL_DB_FIELD_COUNT] = {
   }
 };
 
+#ifndef NO_EMBEDDED_ACCESS_CHECKS
 static const
 TABLE_FIELD_TYPE mysql_user_table_fields[MYSQL_USER_FIELD_COUNT] = {
   {
@@ -553,11 +554,13 @@ TABLE_FIELD_TYPE mysql_tables_priv_table_fields[MYSQL_TABLES_PRIV_FIELD_COUNT] =
     { C_STRING_WITH_LEN("utf8") }
   }
 };
+#endif // NO_EMBEDDED_ACCESS_CHECKS
 
 
 const TABLE_FIELD_DEF
   mysql_db_table_def= {MYSQL_DB_FIELD_COUNT, mysql_db_table_fields};
 
+#ifndef NO_EMBEDDED_ACCESS_CHECKS
 const TABLE_FIELD_DEF
   mysql_user_table_def= {MYSQL_USER_FIELD_COUNT, mysql_user_table_fields};
 
@@ -576,6 +579,7 @@ const TABLE_FIELD_DEF
 const TABLE_FIELD_DEF
   mysql_tables_priv_table_def= {MYSQL_TABLES_PRIV_FIELD_COUNT,
                                 mysql_tables_priv_table_fields};
+#endif // NO_EMBEDDED_ACCESS_CHECKS
 
 static LEX_STRING native_password_plugin_name= {
   C_STRING_WITH_LEN("mysql_native_password")
@@ -5719,7 +5723,8 @@ bool check_grant_db(THD *thd,const char *db)
   bool error= TRUE;
   size_t copy_length;
 
-  copy_length= (size_t) (strlen(sctx->priv_user ? sctx->priv_user : "") +
+  compile_time_assert(sizeof(sctx->priv_user) / sizeof(void*) > 1);
+  copy_length= (size_t) (strlen(sctx->priv_user) +
                  strlen(db ? db : "")) + 1; /* Added 1 at the end to avoid  
                                                buffer overflow at strmov()*/
 
@@ -8623,7 +8628,7 @@ void fill_effective_table_privileges(THD *thd, GRANT_INFO *grant,
   DBUG_PRINT("enter", ("Host: '%s', Ip: '%s', User: '%s', table: `%s`.`%s`",
                        sctx->priv_host, (sctx->get_ip()->length() ?
                        sctx->get_ip()->ptr() : "(NULL)"),
-                       (sctx->priv_user ? sctx->priv_user : "(NULL)"),
+                       sctx->priv_user,
                        db, table));
   /* --skip-grants */
   if (!initialized)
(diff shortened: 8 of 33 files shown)


cmd-line-utils/readline/history.c
```diff
@@ -211,14 +211,14 @@ history_get (offset)
...
-     char *string;
+     const char *string;
      char *ts;
 {
   HIST_ENTRY *temp;
 
   temp = (HIST_ENTRY *)xmalloc (sizeof (HIST_ENTRY));
 
-  temp->line = string ? savestring (string) : string;
+  temp->line = string ? savestring (string) : NULL;
```
laurynas-biveinis, line 221: What was the warning here?

libmysql/libmysql.c
```diff
@@ -1379,7 +1379,8 @@ void set_stmt_errmsg(MYSQL_STMT *stmt, NET *net)
   DBUG_ASSERT(stmt != 0);
 
   stmt->last_errno= net->last_errno;
-  if (net->last_error && net->last_error[0])
+  compile_time_assert(sizeof(net->last_error)/sizeof(void*) > 1);
```
laurynas-biveinis: Spaces around /

sql/item.cc
```diff
@@ -7035,7 +7035,7 @@ bool Item_default_value::fix_fields(THD *thd, Item **items)
   }
   if (!(def_field= (Field*) sql_alloc(field_arg->field->size_of())))
     goto error;
-  memcpy(def_field, field_arg->field, field_arg->field->size_of());
+  memcpy((void*)def_field, (void*)field_arg->field, field_arg->field->size_of());
```
laurynas-biveinis: What was the warning fixed here?

sql/log.cc
```diff
@@ -1236,8 +1236,9 @@ bool LOGGER::slow_log_print(THD *thd, const char *query, uint query_length,
     }
 
     /* fill in user_host value: the format is "%s[%s] @ %s [%s]" */
+    compile_time_assert(sizeof(sctx->priv_user)/sizeof(void*)>1);
```
laurynas-biveinis: Spaces around operators

sql/log_event.cc
```diff
@@ -1164,7 +1164,7 @@ failed my_b_read"));
   Log_event *res=  0;
 #ifndef max_allowed_packet
   THD *thd=current_thd;
-  uint max_allowed_packet= thd ? slave_max_allowed_packet:~(ulong)0;
+  uint max_allowed_packet= thd ? slave_max_allowed_packet:~(uint)0;
```
laurynas-biveinis: Spaces around :

sql/log_event.cc
```diff
@@ -2400,8 +2400,9 @@ bool Query_log_event::write(IO_CACHE* file)
       user= thd->get_invoker_user();
       host= thd->get_invoker_host();
     }
-    else if (thd->security_ctx->priv_user)
+    else
     {
+      compile_time_assert(sizeof(thd->security_ctx->priv_user)/sizeof(void*)>1);
```
laurynas-biveinis: Spaces around operators

sql/slave.cc
```diff
@@ -4099,8 +4099,8 @@ static int queue_event(Master_info* mi,const char* buf, ulong event_len)
 
        TODO: handling `when' for SHOW SLAVE STATUS' snds behind
     */
-    if ((memcmp(mi->master_log_name, hb.get_log_ident(), hb.get_ident_len())
-         && mi->master_log_name != NULL)
+    compile_time_assert(sizeof(mi->master_log_name)/sizeof(void*)>1);
```
laurynas-biveinis: Likewise

sql/slave.cc
```diff
@@ -4358,7 +4358,8 @@ static int connect_to_master(THD* thd, MYSQL* mysql, Master_info* mi,
     mysql_options(mysql, MYSQL_PLUGIN_DIR, opt_plugin_dir_ptr);
 
   /* we disallow empty users */
-  if (mi->user == NULL || mi->user[0] == 0)
+  compile_time_assert(sizeof(mi->user)/sizeof(void*)>1);
```
laurynas-biveinis: Likewise

sql/slave.cc
```diff
@@ -4494,8 +4495,8 @@ MYSQL *rpl_connect_master(MYSQL *mysql)
   /* This one is not strictly needed but we have it here for completeness */
   mysql_options(mysql, MYSQL_SET_CHARSET_DIR, (char *) charsets_dir);
 
-  if (mi->user == NULL
-      || mi->user[0] == 0
+  compile_time_assert(sizeof(mi->user)/sizeof(void*)>1);
```
laurynas-biveinis: Likewise

sql/sql_acl.cc
```diff
@@ -558,22 +558,22 @@ TABLE_FIELD_TYPE mysql_tables_priv_table_fields[MYSQL_TABLES_PRIV_FIELD_COUNT] =
 const TABLE_FIELD_DEF
   mysql_db_table_def= {MYSQL_DB_FIELD_COUNT, mysql_db_table_fields};
 
-const TABLE_FIELD_DEF
+__attribute__((unused)) const TABLE_FIELD_DEF
```
laurynas-biveinis: In what configurations this gets skipped? NO_EMBEDDED_ACCESS_CHECKS? Can't you just use the same ifdef here instead of marking as unused?

If fixed in the current way, then MY_ATTRIBUTE everywhere

sql/sql_acl.cc
```diff
@@ -5719,7 +5719,8 @@ bool check_grant_db(THD *thd,const char *db)
   bool error= TRUE;
   size_t copy_length;
 
-  copy_length= (size_t) (strlen(sctx->priv_user ? sctx->priv_user : "") +
+  compile_time_assert(sizeof(sctx->priv_user)/sizeof(void*)>1);
```
laurynas-biveinis: Spaces around operators

sql/sql_acl.cc
```diff
@@ -10397,10 +10392,10 @@ acl_authenticate(THD *thd, uint connect_errors, uint com_change_user_pkt_len)
   {
     if (strcmp(mpvio.auth_info.authenticated_as, mpvio.auth_info.user_name))
     {
+      compile_time_assert(sizeof(mpvio.auth_info.authenticated_as)/sizeof(void*)>1);
```
laurynas-biveinis: Likewise

sql/sql_audit.cc
```diff
@@ -120,7 +120,7 @@ static audit_handler_t audit_handlers[] =
   general_class_handler, connection_class_handler
 };
 
-static const uint audit_handlers_count=
+__attribute__((unused)) static const uint audit_handlers_count=
```
laurynas-biveinis: Same comments as for the previous unused attribute

sql/sql_audit.cc
```diff
@@ -345,8 +345,8 @@ int initialize_audit_plugin(st_plugin_int *plugin)
 {
   st_mysql_audit *data= (st_mysql_audit*) plugin->plugin->info;
   
-  if (!data->class_mask || !data->event_notify ||
-      !data->class_mask[0])
+  compile_time_assert(sizeof(data->class_mask)/sizeof(void*)>0);
```
laurynas-biveinis: Spaces

sql/sql_parse.cc
```diff
@@ -4031,8 +4031,8 @@ case SQLCOM_PREPARE:
     LEX_USER *grant_user= get_current_user(thd, lex->grant_user);
     if (!grant_user)
       goto error;
-    if ((thd->security_ctx->priv_user &&
-	 !strcmp(thd->security_ctx->priv_user, grant_user->user.str)) ||
+    compile_time_assert(sizeof(thd->security_ctx->priv_user)/sizeof(void*)>1);
```
laurynas-biveinis: Spaces

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1938,6 +1938,7 @@ trx_is_registered_for_2pc(
 
 /*********************************************************************//**
 Note that a transaction owns the prepare_commit_mutex. */
+__attribute__((__unused__))
```
laurynas-biveinis: Is this also used in some configs only?

laurynas-biveinis (changes requested):
- I have reviewed the diff itself, but in general, if any of those warnings are fixed in 5.6-8.0, please take those fixes instead unless they have issues (minor ones are acceptable)
- Please file a bug and mention it in the commit message title

laurynas-biveinis:
If taking an upstream fix from a higher version, and that fix commit scope matches, cherry picks are OK (then this PR can have e.g. several cherry picks and own commit on the top for the rest)

cmd-line-utils/readline/history.c
dutow: This is part of the other change in this file: string was a char*, and not a const char*. And line is simply a char* (and has to stay that), so we can't assign string directly to it, even when we know it's null.

sql/item.cc
dutow: memcpy is used to copy a class with virtual functions, and clang generates the warning that in this case, it'll copy the vtblptr. Which is fine, but this is the way the warning can be silenced.

sql/sql_acl.cc
dutow: MY_ATTRIBUTE was introduced in 5.6, 5.5 used simply attribute (there is an ifdef for it in include/my_attribute.h)

Yes, it's that definition, I could add that ifdef here too.

storage/innobase/handler/ha_innodb.cc
dutow: Yes, only if EXTENDED_FOR_COMMIT_ORDERED

sql/item.cc
laurynas-biveinis: Isn't memcpy on such object undefined behavior, i.e. not fine? What do higher versions do?

sql/sql_acl.cc
laurynas-biveinis: Please use that definition (but my comment of taking higher version fixes, if available, takes precedence)

storage/innobase/handler/ha_innodb.cc
laurynas-biveinis: OK, let's use it then. (I suspect that this define is ours, not upstream, and we could remove it, but 5.5 is a mature release next in line for EOL, so maybe let's just fix the warning)

sql/item.cc
dutow: It is, but I think in practice it works in every implementation. 

They refactored this part, by introducing a pure virtual clone method in the Field class, and implementing it in every derived class.

laurynas-biveinis: OK, that's too much to backport, let's just add the typecasts. Please use C++ style for them

dutow:
Updated the PR.

I was only able to use one of the upstream commits - the other is focused on warning fixes too, but there were more conflicts than useful changes in it.

cmake/maintainer.cmake
```diff
@@ -31,7 +31,7 @@ SET(MY_CXX_WARNING_FLAGS
 # Extra warning flags for Clang++
 IF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   SET(MY_CXX_WARNING_FLAGS
-      "${MY_CXX_WARNING_FLAGS} -Wno-null-conversion -Wno-unused-private-field")
+      "${MY_CXX_WARNING_FLAGS} -Wno-null-conversion -Wno-unused-private-field -Wno-tautological-constant-out-of-range-compare")
```
laurynas-biveinis: On a second though, please add a comment here so this is removed in an eventual merge to 8.0. This is not required if 5.6 or 5.7 already don't need this

sql/log.cc
```diff
@@ -1236,8 +1236,9 @@ bool LOGGER::slow_log_print(THD *thd, const char *query, uint query_length,
     }
 
     /* fill in user_host value: the format is "%s[%s] @ %s [%s]" */
+    compile_time_assert(sizeof(sctx->priv_user) / sizeof(void*)>1);
```
laurynas-biveinis: Spaces around >

sql/log_event.cc
```diff
@@ -2400,8 +2400,9 @@ bool Query_log_event::write(IO_CACHE* file)
       user= thd->get_invoker_user();
       host= thd->get_invoker_host();
     }
-    else if (thd->security_ctx->priv_user)
+    else
     {
+      compile_time_assert(sizeof(thd->security_ctx->priv_user) / sizeof(void*)>1);
```
laurynas-biveinis: Likewise

sql/slave.cc
```diff
@@ -4099,8 +4099,8 @@ static int queue_event(Master_info* mi,const char* buf, ulong event_len)
 
        TODO: handling `when' for SHOW SLAVE STATUS' snds behind
     */
-    if ((memcmp(mi->master_log_name, hb.get_log_ident(), hb.get_ident_len())
-         && mi->master_log_name != NULL)
+    compile_time_assert(sizeof(mi->master_log_name) / sizeof(void*)>1);
```
laurynas-biveinis: Likewise

sql/slave.cc
```diff
@@ -4358,7 +4358,8 @@ static int connect_to_master(THD* thd, MYSQL* mysql, Master_info* mi,
     mysql_options(mysql, MYSQL_PLUGIN_DIR, opt_plugin_dir_ptr);
 
   /* we disallow empty users */
-  if (mi->user == NULL || mi->user[0] == 0)
+  compile_time_assert(sizeof(mi->user) / sizeof(void*)>1);
```
laurynas-biveinis: Likewise

sql/slave.cc
```diff
@@ -4494,8 +4495,8 @@ MYSQL *rpl_connect_master(MYSQL *mysql)
   /* This one is not strictly needed but we have it here for completeness */
   mysql_options(mysql, MYSQL_SET_CHARSET_DIR, (char *) charsets_dir);
 
-  if (mi->user == NULL
-      || mi->user[0] == 0
+  compile_time_assert(sizeof(mi->user) / sizeof(void*)>1);
```
laurynas-biveinis: Likewise

sql/sql_acl.cc
```diff
@@ -576,6 +577,7 @@ const TABLE_FIELD_DEF
 const TABLE_FIELD_DEF
   mysql_tables_priv_table_def= {MYSQL_TABLES_PRIV_FIELD_COUNT,
                                 mysql_tables_priv_table_fields};
+#endif //NO_EMBEDDED_ACCESS_CHECKS
```
laurynas-biveinis: Space after //

sql/sql_acl.cc
laurynas-biveinis: Comment still stands

sql/sql_acl.cc
```diff
@@ -10397,10 +10394,10 @@ acl_authenticate(THD *thd, uint connect_errors, uint com_change_user_pkt_len)
   {
     if (strcmp(mpvio.auth_info.authenticated_as, mpvio.auth_info.user_name))
     {
+      compile_time_assert(sizeof(mpvio.auth_info.authenticated_as) / sizeof(void*)>1);
```
laurynas-biveinis: Spaces around >

sql/sql_audit.cc
```diff
@@ -120,9 +120,10 @@ static audit_handler_t audit_handlers[] =
   general_class_handler, connection_class_handler
 };
 
+#if !defined(DBUG_OFF) && !defined(_lint)
```
laurynas-biveinis, line 123: Why _lint too?

sql/sql_audit.cc
```diff
@@ -345,8 +346,8 @@ int initialize_audit_plugin(st_plugin_int *plugin)
 {
   st_mysql_audit *data= (st_mysql_audit*) plugin->plugin->info;
   
-  if (!data->class_mask || !data->event_notify ||
-      !data->class_mask[0])
+  compile_time_assert(sizeof(data->class_mask) / sizeof(void*)>0);
```
laurynas-biveinis: Spaces around >

sql/sql_parse.cc
```diff
@@ -4031,8 +4031,8 @@ case SQLCOM_PREPARE:
     LEX_USER *grant_user= get_current_user(thd, lex->grant_user);
     if (!grant_user)
       goto error;
-    if ((thd->security_ctx->priv_user &&
-	 !strcmp(thd->security_ctx->priv_user, grant_user->user.str)) ||
+    compile_time_assert(sizeof(thd->security_ctx->priv_user) / sizeof(void*)>1);
```
laurynas-biveinis: Likewise

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1947,6 +1948,7 @@ trx_owns_prepare_commit_mutex_set(
 	ut_a(trx_is_registered_for_2pc(trx));
 	trx->owns_prepare_mutex = 1;
 }
+#endif //EXTENDED_FOR_COMMIT_ORDERED
```
laurynas-biveinis: Space after //

laurynas-biveinis (changes requested):

cmake/maintainer.cmake
dutow: This is only required for some assertions: the assertion checks that the actual value (which has an enum type) returned by a function is less than 256. Clang sees that the largest value of the enum is smaller than 256 (as it should be), and reports a warning because of this.

I didn't see this warning in later versions, so  I think these asserts were removed or changed.

sql/sql_audit.cc
dutow: I copied this ifdef from the definition of the assert macro, which uses the variable.

dutow:
New jenkins run: https://jenkins.percona.com/job/percona-server-5.5-param/1607/

cmake/maintainer.cmake
laurynas-biveinis: I misunderstood the PR description: I thought it is the compile_time_assert itself which caused warnings here.

In this case it is those higher version fixes we should take, unless they are intertwined with other major works. It is also an option to remove the tautological asserts.

dutow: That's a similar, but different warnings, and that was something I could fix with compile-time assertions. 

I wanted to avoid the removal of these asserts, as they are only tautological until they do work - the moment they break, we wouldn't receive this warning. (when somebody adds a value greater than 256 to these enums)

dutow: I updated the PR with a version where I removed the two related asserts (unireg.cc and ha_innodb.cc)

laurynas-biveinis: Would a cast, inside the assert, to a wider type before comparing work?

sql/unireg.cc
```diff
@@ -924,7 +924,6 @@ static bool pack_fields(File file, List<Create_field> &create_fields,
     recpos= field->offset+1 + (uint) data_offset;
     int3store(buff+5,recpos);
     int2store(buff+8,field->pack_flag);
-    DBUG_ASSERT(field->unireg_check < 256);
```
laurynas-biveinis, line 946: IMHO what this is trying to say is compile_time_assert(sizeof() == 1)

laurynas-biveinis (approved):

sql/unireg.cc
dutow: Yes, that would be the intention, but that check would fail - because the size of that field is 4 byte, not 1 - it just never contains a value higher than 256.

percona-ysorokin:
@dutow Please also double check if there are no build issues with clang when both `-DCMAKE_BUILD_TYPE=RelWithDebInfo` and `-DMYSQL_MAINTAINER_MODE=ON` are enabled.

dutow:
@percona-ysorokin The relwithdebinfo errors aren't clang specific, I'll submit those in a separate PR.

dutow:
A small update:
* rebased to latest 5.5 head
* added the llvm-dev package to the travis config
* fixed a header guard in the bundled ssl library

With these, the travis check is green.

:q
```diff
@@ -0,0 +1,77 @@
+# Ubuntu 14.04
```
laurynas-biveinis: File named ":q"? Switch to Emacs ASAP!

extra/yassl/include/openssl/crypto.h
```diff
@@ -19,7 +19,7 @@
 
 /* crypto.h for openSSL */
 
-#ifndef ysSSL_crypto_h__
+#ifndef yaSSL_crypto_h__
```
laurynas-biveinis, line 22: Please report upstream if present in 5.6 at least

laurynas-biveinis (changes requested):
Would be good if not that vi thing... :)

laurynas-biveinis:
Now that we have switched, s/LP/JIRA in the commit message

extra/yassl/include/openssl/crypto.h
dutow: It isn't. The yassl versions is different in the 3 versions, and I've only seen (different) issues in 5.5 and 5.7

:q
dutow: Ops, removed :)

dutow:
Commit message / pr name edited, file removed.

laurynas-biveinis (approved):

PR #2025, opened 2017-12-14
Fixed indentation on the release notes

hrvojem:
Fixed missing links to encryption docs from release notes
Had split MyRocks variables docs as it was too big for Drupal to import


laurynas-biveinis (approved):
Could be self-approved

PR #2026, opened 2017-12-14
MYR-179 : Basic Sysbench run fails with basic MyROCKS install due to …

george-lorch:
…lack of open files

- Discussed behavior changes with upstream, previous attempts to fix this
  uncovered a 'need' to be able to pass '-1' through to rocksdb for those
  users that want to or can do extreme tuning of their environment.
- New behavior is as follows:
  * -2 : 'auto-tune', silently set rocksdb_max_open_files to 1/2 of mysql open_files_limit
  * -1 : infinite, pass -1 to rocksdb
  * >= 0 and <= open_files_limit : take value as-is
  * > open_files_lmit : set rocksdb_max_open_files to 1/2 of mysql open_files_limit and emit a warning to the mysqld error log.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -508,6 +508,7 @@ static std::unique_ptr<rocksdb::DBOptions> rdb_init_rocksdb_db_options(void) {
   o->listeners.push_back(std::make_shared<Rdb_event_listener>(&ddl_manager));
   o->info_log_level = rocksdb::InfoLogLevel::INFO_LEVEL;
   o->max_subcompactions = DEFAULT_SUBCOMPACTIONS;
+  o->max_open_files = -2; // auto-tune to 50% open_files_limit
 
   o->concurrent_prepare = true;
   o->manual_wal_flush = true;
@@ -854,7 +855,8 @@ static MYSQL_SYSVAR_BOOL(
 static MYSQL_SYSVAR_INT(max_open_files, rocksdb_db_options->max_open_files,
                         PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_READONLY,
                         "DBOptions::max_open_files for RocksDB", nullptr,
-                        nullptr, 1000, /* min */ -1, /* max */ INT_MAX, 0);
+                        nullptr, rocksdb_db_options->max_open_files,
+                        /* min */ -2, /* max */ INT_MAX, 0);
 
 static MYSQL_SYSVAR_ULONG(max_total_wal_size,
                           rocksdb_db_options->max_total_wal_size,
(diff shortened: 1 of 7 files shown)


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1391/

Corresponding upstream pull request that has been tentatively approved : https://github.com/facebook/mysql-5.6/pull/771

This differs slightly from what was discussed on bi-weekly meeting due to fear/possibility that rocksdb may also treat '0' as a special case so we should not use it to mean auto-tune for MyRocks. '-2' is guaranteed to be free to mean 'auto-tune'.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -508,6 +508,7 @@ static std::unique_ptr<rocksdb::DBOptions> rdb_init_rocksdb_db_options(void) {
   o->listeners.push_back(std::make_shared<Rdb_event_listener>(&ddl_manager));
   o->info_log_level = rocksdb::InfoLogLevel::INFO_LEVEL;
   o->max_subcompactions = DEFAULT_SUBCOMPACTIONS;
+  o->max_open_files = -2; // auto-tune to 50% open_files_limit
```
laurynas-biveinis, line 511: Magic number in the code, introduce some constexpr

laurynas-biveinis (approved):
The numeric sysvar now has a couple of magic values for users to remember. I'd convert the sysvar to a string one, assign string names to the current -2 and -1, attempt to convert the string to number otherwise.

In any case the current PR is good to go, addressing the code review comment and the above may happen later.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Ack and not a bad idea at all @laurynas-biveinis . The switch to '-2' over '0' was a last second change due to some speculation and guessing on rocksdb expectations. As long as the UI is 'locked in' at this point for GA is the main thing. I can/will come back and make the code more correct to fit this immediately. I have https://jira.percona.com/browse/DOC-661 opened now to ensure that this behavior is clear.

laurynas-biveinis: If you agree on the string sysvar, then it depends whether the change can be made before the GA or afterwards. In one case the -2 -1 numeric options may be removed, in the other not, and it's fine both ways.

george-lorch: Right now, I am resigned to leave as-is since there is not much more time for re-test and re-review with holidays and such. I _believe_ that a change such as you suggest would be invisible to the UI and would make forward/backward compatibility a non-issue. My main concern is changing the expectation or behavior of the UI over a code completeness at this stage combined with our compressed timeline. Switching to a string right now would introduce a bit more code that would need testing and more docs and more chances of breaking something. We can debate if there is any left tomorrow standup.

laurynas-biveinis: Yes it's OK to change later, absolutely no reason to rush this

PR #2032, opened 2017-12-22
lp1716844: Escaping control characters in the audit log

dutow:
Implemented escaping for the JSON output format, and updated the test suite.

The other output formats are left unchanged, as they do not have a well defined method for escaping these characters.
XML does support control characters in version 1.1, but most tools only understand 1.0, and our output is 1.0.

This is exactly the same as #2015, just a few files were moved between the two branches.


dutow:
Jenkins in progress: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1405/

laurynas-biveinis (approved):

dutow:
I had to restart the tests - rocks causes compilation errors, but looks unrelated to this change.

https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1406/

PR #2034, opened 2017-12-26
Add support to log queries exceeding a certain number of rows returned

fipar:
This PR is a reference implementation for PS-284

### DIFF
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -63,7 +63,7 @@ char *audit_log_syslog_ident;
 char default_audit_log_syslog_ident[] = "percona-audit";
 ulong audit_log_syslog_facility= 0;
 ulong audit_log_syslog_priority= 0;
-ulonglong audit_log_num_rows_threshold= 100000;
+ulonglong audit_log_num_rows_threshold= 0;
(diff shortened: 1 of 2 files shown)


plugin/audit_log/audit_log.c
```diff
@@ -1213,6 +1225,25 @@ static MYSQL_SYSVAR_ULONGLONG(buffer_size, audit_log_buffer_size,
   "if FILE handler is used.",
   NULL, NULL, 1048576UL, 4096UL, ULONGLONG_MAX, 4096UL);
 
+static
```
laurynas-biveinis: Unused (and not required it seems)

fipar: You are correct. You've caught me changing something I don't fully understand ;) 

I thought the audit_log_num_rows_threshold_update would be needed as a hook so that the var can be changed, but it can be change even with the function removed. 
I'll push an update soon. Also, I only meant to provide the PR as a reference implementation for PS-284, so that the feature request is more clear. My goal is that hopefully this functionality makes it into the plugin because I have a client that needs it, but if you think the feature is good you may prefer to do your own implementation instead.

dutow:
Hello!

PR #2040, opened 2017-12-28
PS-3767: Fixed clang 4/5 warnings for PS 5.7.

dutow:
* Initialized uninitailized variables
* Marked debug variables as possibly unused
* Marked the functions generated by the define_xdr_funcs macro possibly unused. This macro generates several functions, and only some of them is used.
* Removed the unused MYSQLX_PORT global constant
* Corrected two unsigned/signed std::max
* Disabled the unused variable warning for rocksdb, as it's in 3rd party code
* Removed unneccessary assertions from rocksdb (asserting that a non null parameter is non null)
* Added a missing override specifier
* Added additional braces to a struct initalizer
* Changed how optimization is disabled for a function

### DIFF
--- a/client/mysql.cc
+++ b/client/mysql.cc
@@ -1430,7 +1430,7 @@ int main(int argc,char *argv[])
     "statement.\n");
   put_info(buff,INFO_INFO);
 
-  uint protocol, ssl_mode;
+  uint protocol= 0, ssl_mode= 0;
   if (!mysql_get_option(&mysql, MYSQL_OPT_PROTOCOL, &protocol) &&
       !mysql_get_option(&mysql, MYSQL_OPT_SSL_MODE, &ssl_mode))
   {
--- a/client/mysqladmin.cc
+++ b/client/mysqladmin.cc
@@ -1093,6 +1093,7 @@ static int execute_commands(MYSQL *mysql,int argc, char **argv)
       /* Warn about password being set in non ssl connection */
 #if defined(HAVE_OPENSSL) && !defined(EMBEDDED_LIBRARY)
       uint ssl_mode;
+      ssl_mode= 0;
       if (!mysql_get_option(mysql, MYSQL_OPT_SSL_MODE, &ssl_mode) &&
           ssl_mode <= SSL_MODE_PREFERRED)
       {
--- a/mysys_ssl/yassl.cc
+++ b/mysys_ssl/yassl.cc
@@ -107,6 +107,7 @@ static int EVP_CipherInit_ex(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *cipher,
   return 1;
 }
 
+#ifndef NDEBUG
 static int EVP_CIPHER_CTX_key_length(const EVP_CIPHER_CTX *ctx)
 {
   return ctx->key_len;
@@ -116,6 +117,7 @@ static int EVP_CIPHER_CTX_iv_length(const EVP_CIPHER_CTX *ctx)
 {
   return ctx->flags & EVP_CIPH_ECB_MODE ? 0 : TaoCrypt::AES::BLOCK_SIZE;
 }
+#endif
 
 static void do_whole_blocks(EVP_CIPHER_CTX *ctx, uchar *out, int *outl,
                             const uchar *in, int inl)
--- a/rapid/plugin/group_replication/libmysqlgcs/src/bindings/xcom/gcs_xcom_state_exchange.cc
+++ b/rapid/plugin/group_replication/libmysqlgcs/src/bindings/xcom/gcs_xcom_state_exchange.cc
@@ -300,8 +300,10 @@ void Gcs_xcom_state_exchange::reset_with_flush()
 
 void Gcs_xcom_state_exchange::reset()
 {
+#ifndef NDEBUG
   Gcs_xcom_communication_interface *binding_broadcaster=
     static_cast<Gcs_xcom_communication_interface *>(m_broadcaster);
+#endif
   assert(binding_broadcaster->number_buffered_messages() == 0);
 
   m_configuration_id= null_synode;
--- a/sql/item_geofunc.cc
+++ b/sql/item_geofunc.cc
@@ -1867,7 +1867,7 @@ append_geometry(Geometry::wkb_parser *parser, Json_object *geometry,
         }
         else
         {
-          bool result;
+          bool result= false;
           Json_array *points= new (std::nothrow) Json_array();
           if (points == NULL || collection->append_alias(points))
             return true;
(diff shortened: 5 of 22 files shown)


client/mysql.cc
```diff
@@ -1430,7 +1430,7 @@ int main(int argc,char *argv[])
     "statement.\n");
   put_info(buff,INFO_INFO);
 
-  uint protocol, ssl_mode;
+  uint protocol = 0, ssl_mode = 0;
```
laurynas-biveinis: MySQL coding style: no space before the assignment operator

client/mysqladmin.cc
```diff
@@ -1093,6 +1093,7 @@ static int execute_commands(MYSQL *mysql,int argc, char **argv)
       /* Warn about password being set in non ssl connection */
 #if defined(HAVE_OPENSSL) && !defined(EMBEDDED_LIBRARY)
       uint ssl_mode;
+      ssl_mode = 0;
```
laurynas-biveinis: Likewise, and initialize at declaration

sql/item_geofunc.cc
```diff
@@ -1867,7 +1867,7 @@ append_geometry(Geometry::wkb_parser *parser, Json_object *geometry,
         }
         else
         {
-          bool result;
+          bool result = false;
```
laurynas-biveinis: Likewise

storage/innobase/include/ut0new.h
```diff
@@ -663,8 +663,8 @@ class ut_allocator {
 
 		/* e.g. "btr0cur", derived from "/path/to/btr0cur.cc" */
 		char		keyname[FILENAME_MAX];
-		const size_t	len = ut_basename_noext(file, keyname,
-							sizeof(keyname));
+		const size_t	len MY_ATTRIBUTE((unused)) =
```
laurynas-biveinis, line 666: Why? It's used in ut_a below, which should not be compiled away in any config

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4210,7 +4210,6 @@ longlong ha_rocksdb::update_hidden_pk_val() {
 
 /* Get the id of the hidden pk id from m_last_rowkey */
 int ha_rocksdb::read_hidden_pk_id_from_rowkey(longlong *const hidden_pk_id) {
```
laurynas-biveinis: IMHO we should remove attribute nonnull instead. Its abuses with nullptrs passed will not be caught by compiler nor generated code, leading to nasty "miscompilations", we had two instances of that in the core server. @georgelorchpercona  ?

laurynas-biveinis (changes requested):
MyRocks changes can be reviewed by either me or @georgelorchpercona, but in this case he has to agree with the plan re. nonnulls first.

client/mysqladmin.cc
dutow: I wanted to do that, but I couldn't - this part of the function is inside a goto jump (from a later part to an earlier label), and initializing within it is a compilation error.

storage/innobase/include/ut0new.h
dutow: ut_a is defined to be assert(expr) here, which isn't compiled in release builds.

laurynas-biveinis: I see, this must be innochecksum then

dutow:
Updated the PR:
* Added the toku changes (this PR also requires the PerconaFT submodule changes, as the 5.6 branch)
* Removed the non null attributes instaed of the asserts in myrocks

The clang reldebinfo build wasn't fixed, as that build also fails with gcc, and will be addressed in a separate PR.

rapid/plugin/group_replication/libmysqlgcs/src/bindings/xcom/gcs_xcom_state_exchange.cc
```diff
@@ -300,7 +300,7 @@ void Gcs_xcom_state_exchange::reset_with_flush()
 
 void Gcs_xcom_state_exchange::reset()
 {
-  Gcs_xcom_communication_interface *binding_broadcaster=
+  Gcs_xcom_communication_interface *binding_broadcaster MY_ATTRIBUTE((unused)) =
```
laurynas-biveinis: Why not #ifndef NDEBUG instead?

sql/item_geofunc.cc
laurynas-biveinis: Still not fixed

laurynas-biveinis (changes requested):

dutow:
A small update:

* fixed the previous comments
* rebased to latest 5.7 head
* added the llvm-dev package to the travis config
* fixed two unused function in yassl (for the bundled build)

With these changes, the travis check would be green... except that there are still some errors with gcc, and that it timeouts. Sometimes you can see 1 or 2 green build in it.

mysys_ssl/yassl.cc
```diff
@@ -107,12 +107,12 @@ static int EVP_CipherInit_ex(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *cipher,
   return 1;
 }
 
-static int EVP_CIPHER_CTX_key_length(const EVP_CIPHER_CTX *ctx)
+MY_ATTRIBUTE((unused)) static int EVP_CIPHER_CTX_key_length(const EVP_CIPHER_CTX *ctx)
```
laurynas-biveinis: Please upstream (there consider dropping if they are indeed unused?)

testclients/bug25714.c
```diff
@@ -13,64 +13,52 @@
    along with this program; if not, write to the Free Software
    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA */
 
+#include <assert.h>
```
laurynas-biveinis: Non-minimum diff - looks like this file was clang-format'ed, some changes go against MySQL coding standard

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11463,12 +11463,10 @@ const char *get_rdb_io_error_string(const RDB_IO_ERROR_TYPE err_type) {
   }
 }
 
-// In case of core dump generation we want this function NOT to be optimized
```
laurynas-biveinis: Why removed the comment?

laurynas-biveinis (changes requested):

mysys_ssl/yassl.cc
dutow: I changed it to ifndef ndebug

dutow:
Updated based on comments.

laurynas-biveinis (approved):

PR #2048, opened 2018-01-03
MYR-221 : handle_fatal_signal (sig=11) in myrocks::Rdb_dict_manager::…

george-lorch:
…get_value when querying I_S when MyRocks is not loaded, but rocksdb_global_info is

- Core issue is that rocksdb i_s plugins were not aware of initialized status of
  the engine plugin and always assumed that the engine or parts of it were
  available.
- Fix was to adapt similar construct as TokuDB that maintains a protected flag
  to indicate if/when the engine is fully initialized and able to respond to
  whatever the i_s plugins need. Synchronization is important here to prevent
  race conditions where one session may be installing/uninstalling the engine
  and others may be accessing the i_s plugins.
- Created a new basic i_s plugin test that includes the MYR-221 test sequence.
- For all i_s plugins that had an existing test, include the new basic i_s
  test first.
- For all i_s plugins that were missing tests, created new basic tests that
  included the new i_s basic test.
- Added some missing cleanup to the rocksdb_init_func that would have left some
  resources behind in the case of an initialization error.


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1418/

storage/rocksdb/ha_rocksdb.h
```diff
@@ -1386,4 +1386,72 @@ struct Rdb_inplace_alter_ctx : public my_core::inplace_alter_handler_ctx {
   Rdb_inplace_alter_ctx(const Rdb_inplace_alter_ctx &);
   Rdb_inplace_alter_ctx &operator=(const Rdb_inplace_alter_ctx &);
 };
+
+/*
+  Helper class to control access/init to handlerton instance.
+  Contains a flag that is set if the handlerton is in an initialized, usable
+  state, plus a reader-write lock to protect it without serializing reads.
```
laurynas-biveinis: "writer"

storage/rocksdb/ha_rocksdb.h
```diff
@@ -1386,4 +1386,72 @@ struct Rdb_inplace_alter_ctx : public my_core::inplace_alter_handler_ctx {
...
+
+  void unlock() {
+    mysql_rwlock_unlock(&m_rwlock);
+  }
+
+  /*
+    Must be called with either a read or write lock held, unable to enforce
+    behavior as mysql_rwlock has no means of determining if a thread has a lock
+  */
+  bool initialized() {
```
laurynas-biveinis: const

storage/rocksdb/ha_rocksdb.h
```diff
@@ -1386,4 +1386,72 @@ struct Rdb_inplace_alter_ctx : public my_core::inplace_alter_handler_ctx {
...
+    return m_initialized;
+  }
+
+  /*
+    Must be called with only a write lock held, unable to enforce behavior as
```
laurynas-biveinis, line 1449: Can be implemented with a thread_local flag set/cleared on a lock_write/unlock, but not sure if worth it

storage/rocksdb/ha_rocksdb_proto.h
```diff
@@ -90,4 +90,8 @@ Rdb_dict_manager *rdb_get_dict_manager(void)
 class Rdb_ddl_manager;
 Rdb_ddl_manager *rdb_get_ddl_manager(void)
     MY_ATTRIBUTE((__warn_unused_result__));
+
+struct Rdb_hton_init_state;
+Rdb_hton_init_state *rdb_get_hton_init_state(void)
```
laurynas-biveinis, line 95: Why not a reference instead?

mysql-test/suite/rocksdb/t/i_s_global_info.test
```diff
@@ -0,0 +1,15 @@
...
+
+SHOW CREATE TABLE INFORMATION_SCHEMA.ROCKSDB_GLOBAL_INFO;
+
+# Test that MYR-221 is addressed and that proper error is returned instead of a
+# crash when the SE is unloaded but the I_S plugin is still present
+
+UNINSTALL PLUGIN ROCKSDB;
+
+--error ER_PLUGIN_IS_NOT_LOADED
+SELECT * FROM INFORMATION_SCHEMA.ROCKSDB_GLOBAL_INFO;
```
laurynas-biveinis: I'd try querying the rest of the I_S tables too here (no need for matching SHOW CREATE TABLE above, in fact I'd consider removing the existing one too)

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb/t/i_s_global_info.test
george-lorch: Well, this one was one guaranteed to crash where not all would, and it did not yet have some basic test for the table so I thought I would combine them. Really, this test should probably be executed on each i_s table for all existing tests, creating new tests for those missing i_s are well outside the scope.

storage/rocksdb/ha_rocksdb.h
george-lorch: True, and I considered that, but seemed to be well over the reinvent the wheel. I could have easily just wrapped the rwlock alone and kept the flag apart but I thought combining them, even in this 'broken' fashion made more sense from a readibility and scoping standpoint.

mysql-test/suite/rocksdb/t/i_s_global_info.test
laurynas-biveinis: It's just that the added code for all the I_S tables needs to be checked, for coverage, copy-paste errors and the like.

But I have noticed the name of the testcase only now, so my suggestion would need a second testcase at least, not sure if it's worth the trouble

storage/rocksdb/ha_rocksdb_proto.h
george-lorch: I considered it and was torn. My initial implementation was a ref only and I changed it. I was trying to stay sort of inline with existing MyRocks semantics. There is a difference though, all of the existing MyRocks constructs that follow this model are dynamic whereas this is pretty much a static singleton. Really I was considering what upstream would be more inclined to take in and trying to fit that.

mysql-test/suite/rocksdb/t/i_s_global_info.test
george-lorch: Yeah, agreed, and maybe this is the opportunity to at least introduce the minimal tests for each i_s table that does just this check if no test exists and append this to any existing tests.

storage/rocksdb/ha_rocksdb.h
```diff
@@ -1386,4 +1386,72 @@ struct Rdb_inplace_alter_ctx : public my_core::inplace_alter_handler_ctx {
...
+      else
+        m_state.lock_read();
+    }
+    ~Scoped_lock() {
+      m_state.unlock();
+    }
+  private:
+    Rdb_hton_init_state& m_state;
+  };
+
```
vlad-lesin: Please, also disable copy constructor and assign operator for Rdb_hton_init_state::Scoped_lock.

george-lorch: Good catch, will prevent shenanigans and misuse later on if someone were to misunderstand the purpose.

george-lorch:
OK, did a nice bit of needed work on the i_s tests and factored out common i_s test behavior into a new i_s_basic.inc
Touched up some of the code bits talked about and a new jenkins here https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1419/

mysql-test/suite/rocksdb/include/i_s_basic.inc
```diff
@@ -0,0 +1,31 @@
...
+#    are common to all of the MyRocks information_schema plugins.
+#
+# USAGE
+#
+#    let $i_s_plugin=<plugin_table_name>
+
+if (!$i_s_plugin)
+{
+  --echo i_s_plugin not specified to suite/rocksdb/include/i_s_basic.inc
+  --die i_s_plugin not specified to suite/rocksdb/include/i_s_basic.inc
```
laurynas-biveinis: Just --die should be enough?

laurynas-biveinis (approved):
--echo/--die comment addressing up to you

mysql-test/suite/rocksdb/include/i_s_basic.inc
george-lorch: Well, maybe it is an older/deprecated concept. I looked in some of the mysql-test/include files for some idiom to use and most did it this way, which I thought was odd, but I went with it. It is a trivial thing/change and really only there to ensure that a consumer of the .inc is using it properly.

PR #2049, opened 2018-01-04
Added .travis.yml for Travis CI

inikep:
This script adds an option of using free TravisCI.org services as
additional tests for percona-server. To use the script each GitHub
user have to create a Travis account and enable testing for his/her fork
of percona-server. Tests are performed on Ubuntu 14.04 LTS (Trusty Tahr).

This script starts 9 Travis jobs. A job number 1 checks if the last commit
is valid with clang-format. The rest of jobs builds precona-server using
the latest versions of gcc-4.8, gcc-5, gcc-6, gcc-7, clang-4.0, clang-5.0
with both `RelWithDebInfo` and `Debug` compilation. There are 2 additional
jobs for gcc-7 and clang-5.0 that use `-DWITH_SSL=bundled` instead of
`-DWITH_SSL=system`

Jobs with a number >= 6 are done only for a pull request to speed up testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,77 @@
+# Ubuntu 14.04
+dist: trusty
+sudo: required
+language: c
+matrix:
+  include:
+    - env: COMMAND=clang-test
+      script:
+        - wget https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/clang-format-diff.py
+        - chmod a+x clang-format-diff.py
+        - git diff -U0 --no-color HEAD^1 | ./clang-format-diff.py -p1 >_GIT_DIFF
+        - '[ ! -s _GIT_DIFF ] && echo The last git commit is clang-formatted || cat _GIT_DIFF'
+
+    - env: GCC=gcc-4.8   CXX=g++-4.8     LIBTYPE=system
+    - env: GCC=gcc-7     CXX=g++-7       LIBTYPE=system  PACKAGES=g++-7     PPA=ubuntu-toolchain-r/test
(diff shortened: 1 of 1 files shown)


dutow:
> A job number 1 checks if the last commit is valid with clang-format

I don't see any clang format config file in the 5.5 branch, or any settings specified in the script. How is it configured?

And I'm not 100% sure about travis, but the default clang version in trusty is 3.5, so it's possible that the diff script uses a quite old clang-format. As we have newer versions of the compiler there, wouldn't it be a better idea to use the clang-format binary from 5.0 explicitly? (and also, to use the diff script from the 5.0 branch instead of the one in the llvm trunk?)

inikep:
The logic behind clang-format is here:
https://github.com/inikep/percona-server/blob/b7156b801b52486279854cb46e7d4d12ae1ae7e2/.travis.yml#L25-L28

inikep:
I checked and Travis has `clang-format version 5.0.0 (tags/RELEASE_500/final)`. But we need a good clang-format file before we can rely on results from Travis.

inikep:
I changed the link to
https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/clang-format-diff.py

laurynas-biveinis (approved):
Do you have a sample TravisCI run link?

laurynas-biveinis:
You can also take a look at MariaDB Travis config for some ideas

inikep:
This is a sample TravisCI run for 5.5:
https://travis-ci.org/inikep/percona-server/builds/325066332

inikep:
I checked MariaDB Travis config. The main differences:
- they are doing macOS tests (and Windows tests with AppVeyor)
- they are doing `mtr` tests (the tests are divided among different jobs and compilers)
- they are using `ccache`
- sent notifications after build using irc

percona-ysorokin (approved):
LGTM

PR #2054, opened 2018-01-05
MYR-207 : Provide a way to force build MyRocks without FastCRC32 if h…

george-lorch:
…ost doesn't support it

- Added new cmake flag ALLOW_NO_SSE42 that can be specified via
  -DALLOW_NO_SSE42=1 which will allow MyRocks to compile if SSE 4.2 is not
  detected.
- Changed warning to error if SSE 4.2 is not detected and no ALLOW_NO_SSE42is
  specified.
- Fixed logic for debug builds in HAVE_SSE42 CHECK_CXX_SOURCE_COMPILES where
  debug builds emit a warning as error "x is unused"

### DIFF
--- a/storage/rocksdb/CMakeLists.txt
+++ b/storage/rocksdb/CMakeLists.txt
@@ -66,13 +66,17 @@ CHECK_CXX_SOURCE_COMPILES("
 #include <cstdint>
 #include <nmmintrin.h>
 int main() {
-  volatile uint32_t x = _mm_crc32_u32(0, 0);
+  volatile uint32_t x __attribute__((unused)) = _mm_crc32_u32(0, 0);
 }
 " HAVE_SSE42)
 IF (HAVE_SSE42)
   ADD_DEFINITIONS(-DHAVE_SSE42)
 ELSE ()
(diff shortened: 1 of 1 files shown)


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1428/

laurynas-biveinis (approved):
Is "x = x" warning-clean, no "variable set but not used"?
I'd have gone with unused attribute, but if the assignment works the same, then LGTM

george-lorch:
Yeah, actually the initial problem is that x was unused, in debug build with warn-as-error, this caused the code snipped to fail to compile. This solves that and doesn't seem to be emitting any other warnings but I suppose that a really smart compiler could see this and eventually start emitting more warnings/errors as you say, 'set but not used'. I don't know why I didn't think of attribute, this is exactly what it exists for. I'll switch it out and run another jenkins, then merge.

george-lorch:
New jenkins with __attribute__((unused)) in the cmake code snippit. https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1431/

Merging...

PR #2056, opened 2018-01-05
Fixed DOC-669 by adding variable description and details from FAQ wri…

fiowro:
…tten by Vlad Lesin


laurynas-biveinis (changes requested):
For something introduced in mid-series, we have "introduced in version x.y.z-t" info, please check other variables for examples. Otherwise LGTM

PR #2058, opened 2018-01-08
MYR-223 : MyRocks alter table with string indexed column to non-binar…

george-lorch:
…y collation works

- Imported patch from MariaDB MDEV-14293 and touched up to use MySQL/Percona
  push_warning_printf. Patch changes error of using inefficient collation
  to a warning.
- Re-worked impacted tests to remove expected error and re-recorded to capture
  new SQL warnings.

### DIFF
--- a/mysql-test/suite/rocksdb/t/collation.test
+++ b/mysql-test/suite/rocksdb/t/collation.test
@@ -4,74 +4,73 @@
 --source include/have_rocksdb.inc
 
 # ci non-indexed column is allowed
-CREATE TABLE t1 (id INT primary key, value varchar(50), value2 varbinary(50), value3 text) engine=rocksdb charset utf8;
+CREATE TABLE t1 (id INT PRIMARY KEY, value VARCHAR(50), value2 VARBINARY(50), value3 TEXT) ENGINE=ROCKSDB CHARSET UTF8;
 DROP TABLE t1;
 
 # ci indexed column is not allowed
---error ER_UNSUPPORTED_COLLATION
-CREATE TABLE t1 (id INT primary key, value varchar(50), value2 varbinary(50), value3 text, index(value)) engine=rocksdb charset utf8;
---error ER_UNSUPPORTED_COLLATION
-CREATE TABLE t1 (id INT primary key, value varchar(50), value2 varbinary(50), value3 text, index(value3(50))) engine=rocksdb charset utf8;
+CREATE TABLE t1 (id INT PRIMARY KEY, value VARCHAR(50), value2 VARBINARY(50), value3 TEXT, INDEX(value)) ENGINE=ROCKSDB CHARSET UTF8;
+DROP TABLE t1;
+CREATE TABLE t1 (id INT PRIMARY KEY, value VARCHAR(50), value2 VARBINARY(50), value3 TEXT, INDEX(value3(50))) ENGINE=ROCKSDB CHARSET UTF8;
+DROP TABLE t1;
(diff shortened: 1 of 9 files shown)


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1432/

mysql-test/suite/rocksdb/t/collation.test
```diff
@@ -38,40 +38,39 @@ DROP TABLE t1;
...
 CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
 DROP TABLE t1;
---error ER_UNSUPPORTED_COLLATION
 CREATE TABLE u1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
+DROP TABLE u1;
 
 # test allowing alters to create temporary tables
 SET GLOBAL rocksdb_strict_collation_exceptions='t1';
-CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb;
+CREATE TABLE t1 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
```
laurynas-biveinis: Maybe make the keyword case consistent too at the same time

mysql-test/suite/rocksdb/t/collation.test
```diff
@@ -38,40 +38,39 @@ DROP TABLE t1;
...
 ALTER TABLE t1 AUTO_INCREMENT=1;
 DROP TABLE t1;
---error ER_UNSUPPORTED_COLLATION
-CREATE TABLE t2 (id INT primary key, value varchar(50), index(value)) engine=rocksdb;
-CREATE TABLE t2 (id INT primary key, value varchar(50)) engine=rocksdb;
---error ER_UNSUPPORTED_COLLATION
+CREATE TABLE t2 (id INT primary key, value varchar(50), index(value)) engine=rocksdb charset utf8;
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/t/collation.test
```diff
@@ -38,40 +38,39 @@ DROP TABLE t1;
...
+DROP TABLE t2;
+CREATE TABLE t2 (id INT primary key, value varchar(50)) engine=rocksdb charset utf8;
```
laurynas-biveinis: Likewise

laurynas-biveinis (approved):

mysql-test/suite/rocksdb/t/collation.test
george-lorch: Yeah, true, this kind of stuff drives me nuts because the end result is that I end up re-working the entire test case to be consistent one way or another, then it destroys my ability to do clean merges :-|

laurynas-biveinis: ugh. Do you what you have to do to stay sane

george-lorch: Looking at the diffs on this test, we are pretty far diverged on the entire case as it is, so now I think fixing case consistency will not introduce any more merge issue than already exist in this test/result. So yeah, ignore my initial comment, I'll fix all of the case inconsistencies in this test with this PR and send it upstream at the next opportunity.

george-lorch:
New jenkins https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1437/

laurynas-biveinis (approved):

PR #2059, opened 2018-01-08
Fixed DOC-668 by further documenting encryption of binlog and relay l…

fiowro:
…ogs.

Also fixes DOC-674 (mysqlbinlog decryption limitations)

### DIFF
--- a/doc/source/management/data_at_rest_encryption.rst
+++ b/doc/source/management/data_at_rest_encryption.rst
@@ -80,6 +80,46 @@ Trying to add unencrypted table to this table space will result in an error:
   |Percona XtraBackup| currently doesn't support backup of encrypted general
   tablespaces.
 
+Binary log encryption
+=====================
+
+A new option, implemented since |Percona Server| :rn:`5.7.20-19`, is
+encryption of binary and relay logs, triggered by the
+:variable:`encrypt_binlog` variable.
+
+Besides turning :variable:`encrypt_binlog` ``ON``, this feature requires both
+`master_verify_checksum
+<https://dev.mysql.com/doc/refman/5.7/en/replication-options-binary-log.html#sysvar_master_verify_checksum>`_
+and `binlog_checksum
+<https://dev.mysql.com/doc/refman/5.7/en/replication-options-binary-log.html#sysvar_binlog_checksum>`_
+variables to be turned ``ON``.
+
+While replicating, master sends the stream of decrypted binary log events to a
+slave (SSL connections can be set up to encrypt them in transport). That said,
+masters and slaves use separate keyring storages and are free to use differing
+keyring plugins.
+
+Dumping of encrypted binary logs involves decryption, and can be done using
+``mysqlbinlog`` with ``--read-from-remote-server`` option.
+
+.. note:: Taking into account that ``--read-from-remote-server`` option  is only
+   relevant to binary logs, encrypted relay logs can not be dumped/decrypted
+   in this way.
+
+System Variables
+----------------
+
+.. variable:: encrypt_binlog
+
+  :version 5.7.20-19: Implemented
+  :cli: ``--encrypt-binlog``
+  :dyn: No
+  :scope: Global
+  :vartype: Boolean
+  :default: ``OFF``
+
+The variable turns on binary and relay logs encryption.
+
 .. _keyring_vault_plugin:
 
 Keyring Vault plugin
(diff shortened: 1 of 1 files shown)


doc/source/management/data_at_rest_encryption.rst
```diff
@@ -80,6 +80,54 @@ Trying to add unencrypted table to this table space will result in an error:
...
+server UUIDs, which makes multi-source replication possible.
+The encryption key is always automatically rotated at the start of each
+binary log file. Therefore configuring the automatic purging of old binary
+logs ensures that old key versions are not in use. 
+``PURGE BINARY LOGS`` and ``RESET MASTER`` commands delete obsolete binary log 
+key versions from the keyring, and obsolete relay log key versions are deleted 
+together with their corresponding relay logs. This scheme is independent from 
+the master instance key and its rotation. 
+
+While replication, master sends the stream of decrypted binary log events to a
```
laurynas-biveinis: s/replication/replicating

doc/source/management/data_at_rest_encryption.rst
```diff
@@ -80,6 +80,54 @@ Trying to add unencrypted table to this table space will result in an error:
...
+``PURGE BINARY LOGS`` and ``RESET MASTER`` commands delete obsolete binary log 
+key versions from the keyring, and obsolete relay log key versions are deleted 
+together with their corresponding relay logs. This scheme is independent from 
+the master instance key and its rotation. 
+
+While replication, master sends the stream of decrypted binary log events to a 
+slave (SSL connections can be set up to encrypt them in transport). That said, 
+masters and slaves are free to use differing keyring plugins.
+
+Decryption of binary logs can be done using ``mysqlbinlog`` with
```
laurynas-biveinis: Maybe "Dumping of encrypted binary logs", if yes, then the next note should be changed accordingly too

laurynas-biveinis (changes requested):
@robgolebiowski should review too

doc/source/management/data_at_rest_encryption.rst
```diff
@@ -80,6 +80,55 @@ Trying to add unencrypted table to this table space will result in an error:
...
+
+Besides turning :variable:`encrypt_binlog` ``ON``, this feature requires both
+`master_verify_checksum
+<https://dev.mysql.com/doc/refman/5.7/en/replication-options-binary-log.html#sysvar_master_verify_checksum>`_
+and `binlog_checksum
+<https://dev.mysql.com/doc/refman/5.7/en/replication-options-binary-log.html#sysvar_binlog_checksum>`_
+variables to be turned ``ON``.
+
+Distinct keys are used for binary and relay logs, and key identifiers include
+server UUIDs, which makes multi-source replication possible.
```
robgolebiowski: This is not true or not yet true - depends on what we decide. The same key is used for encrypting both binlog and relay logs - this is how MariaDB implementation works and I do not see gain in having separate keys for relay log and binlog log. The key name does not contain UUID as each server has to use separate keyring storage (file for keyring_file and secret_mount_point for keyring_vault). The documentation needs updating in those regards.

doc/source/management/data_at_rest_encryption.rst
```diff
@@ -80,6 +80,55 @@ Trying to add unencrypted table to this table space will result in an error:
...
+The encryption key is always automatically rotated at the start of each
+binary log file. Therefore configuring the automatic purging of old binary
+logs ensures that old key versions are not in use.
+``PURGE BINARY LOGS`` and ``RESET MASTER`` commands delete obsolete binary log
+key versions from the keyring, and obsolete relay log key versions are deleted
+together with their corresponding relay logs. This scheme is independent from
+the master instance key and its rotation.
+
```
robgolebiowski: Key rotation is not yet implemented. In current state the same key is used always. In the next iteration when key versioning gets implemented it will be possible to rotate binlog key with use of SQL function. In further iterationx we will add PURGE BINARY LOGS and RESET MASTER support together with rotating key on start of each binary log file (this is however problematic as might cause binlog to grow rapidly and cause some lags in case keyring_vault - each rotation will require connecting to Vault server).

doc/source/management/data_at_rest_encryption.rst
```diff
@@ -80,6 +80,55 @@ Trying to add unencrypted table to this table space will result in an error:
...
+slave (SSL connections can be set up to encrypt them in transport). That said,
+masters and slaves are free to use differing keyring plugins.
+
+Dumping of encrypted binary logs involves decryption, and can be done using
+``mysqlbinlog`` with ``--read-from-remote-server`` option.
+
+.. note:: Taking into account that ``--read-from-remote-server`` option  is only
+   relevant to binary logs, encrypted relay logs can not be dumped/decrypted
+   in this way.
+
```
robgolebiowski, line 108: There is a request from PZ to document "how can I extract Encrypted binglogs to do some advanced point in time recovery and diagnostics ?" @laurynas-biveinis - this should be part of documentation or rather a blog post ?

robgolebiowski (changes requested):
Things needs to be updated in documentation to - this is task for me - once we make sure we are on the same page with @laurynas-biveinis.

doc/source/management/data_at_rest_encryption.rst
laurynas-biveinis: Both, eventually.

This PR should not be help until it's fully complete: only that the errors are fixed, then further content should be added in follow-ups.

robgolebiowski (approved):

PR #2064, opened 2018-01-10
Bug: PS-264 handle_fatal_signal (sig=11) in

robgolebiowski:
keyring::Vault_curl::write_key | plugin/keyring_vault/vault_curl.cc:256

The problem itself was in CURL library, sometimes it fails to
re-initialize CURL session properly. The CURL documentation is vague in
describing the issue. The fix was to reinitialize curl session per
every operation and do not try to reuse the existing curl session.

Also both keyrings (keyring_file and keyring_vault) export a lot of
variables. Scope of those variables were limited to library scope. This
should guard against any dynamic linking problems when both keyrings
get installed on PS.

### DIFF
--- a/plugin/keyring_vault/vault_curl.cc
+++ b/plugin/keyring_vault/vault_curl.cc
@@ -1,5 +1,6 @@
 #include <my_global.h>
 #include <boost/scope_exit.hpp>
+#include <boost/core/noncopyable.hpp>
 #include <algorithm>
 #include "my_rdtsc.h"
 #include "sql_error.h"
@@ -37,6 +38,21 @@ class Thd_wait_end_guard
         was_thd_wait_started = false;
       }
     }
+};
+
+class Curl_session_guard : private boost::noncopyable
+{
+public:
+  Curl_session_guard(CURL *curl)
+    : curl(curl)
+  {}
+  ~Curl_session_guard()
+  {
+    if (curl != NULL)
+      curl_easy_cleanup(curl);
+  }
+private:
+  CURL *curl;
 };
 
 static size_t write_response_memory(void *contents, size_t size, size_t nmemb, void *userp)
@@ -131,23 +147,9 @@ bool Vault_curl::init(const Vault_credentials &vault_credentials)
   return false;
 }
 
-bool Vault_curl::reset_curl_session()
+bool Vault_curl::setup_curl_session(CURL *curl)
 {
   CURLcode curl_res = CURLE_OK;
-  if (unlikely(curl == NULL))
-  {
-    curl = curl_easy_init();
-    if (curl == NULL)
-    {
-      logger->log(MY_ERROR_LEVEL, "Could not create CURL session");
-      return true;
-    }
-    return false; 
-  }
-  else
-  {
-    curl_easy_reset(curl);
-  }
   read_data_ss.str("");
   read_data_ss.clear();
   curl_errbuf[0] = '\0';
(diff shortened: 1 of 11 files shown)


robgolebiowski:
JR => https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1436/

laurynas-biveinis (approved):
Should keyring_file exporting too much of symbols be reported upstream?

plugin/keyring_vault/vault_keyring.cc
```diff
@@ -13,32 +13,13 @@ using keyring::Vault_keys_container;
...
-}
-
-static void cleanup_curl()
-{
-  if (curl != NULL)
-    curl_easy_cleanup(curl);
-  curl_global_cleanup();
-}
 
 static bool reset_curl()
```
percona-ysorokin, line 19: Do we still need to call `reset_curl()` from `check_keyring_file_data()` (when changing `keyring_vault_config`)?

plugin/keyring_vault/vault_curl.cc
```diff
@@ -39,6 +39,21 @@ class Thd_wait_end_guard
     }
 };
 
+class Curl_session_guard
```
percona-ysorokin: noncopyable

unittest/gunit/keyring_vault/vault_io-t.cc
```diff
@@ -608,14 +606,14 @@ int main(int argc, char **argv) {
...
-  if (keyring__vault_io_unittest::curl == NULL)
+  CURL *curl = curl_easy_init();
+  if (curl == NULL)
   {
     std::cout << "Could not initialize CURL session" << std::endl;
     return 1; 
   }
   keyring__vault_io_unittest::logger = new keyring::Mock_logger();
-  keyring::Vault_mount vault_mount(keyring__vault_io_unittest::curl, keyring__vault_io_unittest::logger);
+  keyring::Vault_mount vault_mount(curl, keyring__vault_io_unittest::logger);
```
percona-ysorokin: add `curl_easy_cleanup(curl)` before each error return:
return 2;
return 3;
return 4;

unittest/gunit/keyring_vault/vault_keys_container-t.cc
```diff
@@ -788,14 +787,14 @@ int main(int argc, char **argv)
...
   {
     std::cout << "Could not initialize CURL session" << std::endl;
     return 1; 
   }
   keyring__vault_keys_container_unittest::logger = new keyring::Mock_logger();
-  keyring::Vault_mount vault_mount(keyring__vault_keys_container_unittest::curl, keyring__vault_keys_container_unittest::logger);
+  keyring::Vault_mount vault_mount(curl, keyring__vault_keys_container_unittest::logger);
 
   if (generate_credential_file(keyring__vault_keys_container_unittest::credential_file_url, CORRECT, keyring__vault_keys_container_unittest::uuid))
   {
```
percona-ysorokin, line 808: Likewise
add curl_easy_cleanup(curl) before each error return:
return 2;
return 3;
return 4;

plugin/keyring_vault/vault_curl.h
```diff
@@ -38,15 +37,14 @@ class Vault_curl : public IVault_curl, private boost::noncopyable
...
-  bool reset_curl_session();
+  bool setup_curl_session(CURL *curl);
   std::string get_error_from_curl(CURLcode curl_code);
   bool encode_key_signature(const Vault_key &key, Secure_string *encoded_key_signature);
   bool get_key_url(const Vault_key &key, Secure_string *key_url);
 
   ILogger *logger;
   Secure_string token_header;
   Secure_string vault_url;
-  CURL *curl;
```
percona-ysorokin, line 49: Were there any problems with keeping `curl` as a class member and calling `curl_easy_reset(curl)` before each request?

percona-ysorokin (changes requested):

plugin/keyring_vault/vault_curl.h
robgolebiowski: Yes. For some reason CURL is not coping very well with repetitive new connections using the same CURL object. There were quite a few "CURL not initialized correctly" errors in error logs - which by its own should not be a problem - CURL should just stop working, but under stress testing it caused CURL to crash. I think it should be safer this way to just use fresh CURL object per each operation than make yet another try to reuse CURL object in a way that will not cause a crash. The reason is also that Centos6 is using very old CURL - from 2008 and it was a source of lots of troubles and I do not think it is a reliable library.

plugin/keyring_vault/vault_keyring.cc
robgolebiowski: Yes, this is needed for Centos6. For some reason without this CURL (from 2008) just crashes. I tried to debug this version of CURL lib, but it was nothing obvious and I did not have more time to spend on this.

robgolebiowski:
JR (2nd CR) => https://jenkins.percona.com/job/mysql-5.7-param/1440/

percona-ysorokin (approved):
LGTM

robgolebiowski:
@laurynas-biveinis I am not sure if they will be interested in limiting the number of variables, as they do not do it for any other plugin. Having said that I do not think there is any harm in suggesting it :)

PR #2065, opened 2018-01-11
PS-310: Added support for building PS 5.5 with clang and MSAN.

dutow:
This change adds the option to build PS 5.5 with MSAN, but doesn't try to fix any actual issues in the main MySQL code.
    
    The only issue fixed is an uninitailized read in the mysqltest binary - without that, the test suite can't be started.
    
    After these changes, running the MTR tests is possible, but there are still some remaining errors.
    To build with MSAN:
    
    * use an instrumented libcxx (with msan)
    * use the bundled zlib and libssl
    
    And set the following environment variables before running CMake:
    
    * CFLAGS: -fsanitize=memory
    * CXXFLAGS: -fsanitize=memory -Wno-unused-command-line-argument -nostdinc++ -stdlib=libc++
    * CC, CXX to clang
    * CPLUS_INCLUDE_PATH, C_INCLUDE_PATH, LIBRARY_PATH, LD_LIBRARY_PATH to use the instrumented libcxx
    
    the LD_LIBRARY_PATH setting is also required for running any of the binaries, otherwise MSAN will report false positives.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,28 @@ IF (WITH_ASAN)
   ENDIF()
 ENDIF()
 
+OPTION(WITH_MSAN "Enable memory sanitizer" OFF)
+IF(WITH_MSAN)
+  # TLS model is required because otherwise MSAN has issues with thread local
+  # variable initialization in dynamic libraries
+  SET(MSAN_FLAGS "-fsanitize=memory -fsanitize-memory-track-origins=2 -ftls-model=initial-exec -fPIC -fno-omit-frame-pointer -fsanitize-blacklist=${CMAKE_CURRENT_SOURCE_DIR}/cmake/msan-blacklist.txt")
+  MY_CHECK_C_COMPILER_FLAG("-fsanitize=memory" HAVE_C_MSANITIZE)
+  MY_CHECK_CXX_COMPILER_FLAG("-fsanitize=memory" HAVE_CXX_MSANITIZE)
+  IF(HAVE_C_MSANITIZE AND HAVE_CXX_MSANITIZE)
+    SET(CMAKE_C_FLAGS_DEBUG
+      "${CMAKE_C_FLAGS_DEBUG} ${MSAN_FLAGS} -O1")
+    SET(CMAKE_C_FLAGS_RELWITHDEBINFO
+      "${CMAKE_C_FLAGS_RELWITHDEBINFO} ${MSAN_FLAGS}")
+    SET(CMAKE_CXX_FLAGS_DEBUG
+      "${CMAKE_CXX_FLAGS_DEBUG} ${MSAN_FLAGS} -O1")
+    SET(CMAKE_CXX_FLAGS_RELWITHDEBINFO
+      "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} ${MSAN_FLAGS}")
+    SET(WITH_MSAN_OK 1)
+  ELSE()
+    MESSAGE(FATAL_ERROR "Do not know how to enable memory sanitizer.")
+  ENDIF()
+ENDIF()	
+
 # Always enable debug sync for debug builds.
 SET(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DENABLED_DEBUG_SYNC")
 SET(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -DENABLED_DEBUG_SYNC")
--- a/client/mysqltest.cc
+++ b/client/mysqltest.cc
@@ -10321,10 +10321,10 @@ int insert_pointer_name(reg1 POINTER_ARRAY *pa,char * name)
     if (!(pa->typelib.type_names=(const char **)
 	  my_malloc(((PC_MALLOC-MALLOC_OVERHEAD)/
 		     (sizeof(char *)+sizeof(*pa->flag))*
-		     (sizeof(char *)+sizeof(*pa->flag))),MYF(MY_WME))))
+		     (sizeof(char *)+sizeof(*pa->flag))),MYF(MY_WME | MY_ZEROFILL))))
       DBUG_RETURN(-1);
     if (!(pa->str= (uchar*) my_malloc((uint) (PS_MALLOC-MALLOC_OVERHEAD),
-				     MYF(MY_WME))))
+				     MYF(MY_WME | MY_ZEROFILL))))
     {
       my_free(pa->typelib.type_names);
       DBUG_RETURN (-1);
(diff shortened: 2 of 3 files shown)


dutow:
No jenkins runs yet: this changeset doesn't compile without #2022. I will rebase the PR after that's merged.

Adding msan to jenkins will also take some time because of the additional instrumented libcxx requirement, but I also created docker images for easy testing on any computer.

CMakeLists.txt
```diff
@@ -260,6 +260,26 @@ IF (WITH_ASAN)
...
+  IF(HAVE_C_MSANITIZE AND HAVE_CXX_MSANITIZE)
+    SET(CMAKE_C_FLAGS_DEBUG
+      "${CMAKE_C_FLAGS_DEBUG} ${MSAN_FLAGS} -O1")
+    SET(CMAKE_C_FLAGS_RELWITHDEBINFO
+      "${CMAKE_C_FLAGS_RELWITHDEBINFO} ${MSAN_FLAGS}")
+    SET(CMAKE_CXX_FLAGS_DEBUG
+      "${CMAKE_CXX_FLAGS_DEBUG} ${MSAN_FLAGS} -O1")
+    SET(CMAKE_CXX_FLAGS_RELWITHDEBINFO
+      "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} ${MSAN_FLAGS}")
+    SET(WITH_ASAN_OK 1)
```
laurynas-biveinis: ASAN?

client/mysqltest.cc
```diff
@@ -10318,17 +10318,20 @@ int insert_pointer_name(reg1 POINTER_ARRAY *pa,char * name)
...
+         (sizeof(char *)+sizeof(*pa->flag)));
     if (!(pa->typelib.type_names=(const char **)
-	  my_malloc(((PC_MALLOC-MALLOC_OVERHEAD)/
-		     (sizeof(char *)+sizeof(*pa->flag))*
-		     (sizeof(char *)+sizeof(*pa->flag))),MYF(MY_WME))))
+	       my_malloc(type_names_size, MYF(MY_WME))))
       DBUG_RETURN(-1);
+    memset(pa->typelib.type_names, 0, type_names_size);
     if (!(pa->str= (uchar*) my_malloc((uint) (PS_MALLOC-MALLOC_OVERHEAD),
 				     MYF(MY_WME))))
```
laurynas-biveinis: Likewise

laurynas-biveinis (changes requested):
Please file a JIRA issue

client/mysqltest.cc
dutow: Likewise to what?

client/mysqltest.cc
```diff
@@ -10318,17 +10318,20 @@ int insert_pointer_name(reg1 POINTER_ARRAY *pa,char * name)
...
   if (! pa->typelib.count)
   {
+    const size_t type_names_size = ((PC_MALLOC-MALLOC_OVERHEAD)/
+         (sizeof(char *)+sizeof(*pa->flag))*
+         (sizeof(char *)+sizeof(*pa->flag)));
     if (!(pa->typelib.type_names=(const char **)
-	  my_malloc(((PC_MALLOC-MALLOC_OVERHEAD)/
-		     (sizeof(char *)+sizeof(*pa->flag))*
-		     (sizeof(char *)+sizeof(*pa->flag))),MYF(MY_WME))))
+	       my_malloc(type_names_size, MYF(MY_WME))))
```
laurynas-biveinis: Make my_malloc a calloc here by oring a MY_ZEROFILL flag

client/mysqltest.cc
laurynas-biveinis: To the comment above which for some reason is gone :) Reposted

dutow:
I updated the pull request.

After these changes, the remaining msan failures in the 5.5 mtr tests are:

* main.mysql_client_test - uninitailized reads in the client libraries. Could also be a problem with the tests, as this is the only occurence, but needs more investigation
* main.bug80134 - msan repoprts that it can't allocate several terabytes of memory. This could be a bug in msan, but I'm not yet 100% sure
* main.myisampack - reports uninitialized reads in the packing code, I couldn't figure it out so far

sql/filesort.cc
```diff
@@ -226,7 +226,7 @@ ha_rows filesort(THD *thd, TABLE *table, SORT_FIELD *sortorder, uint s_length,
       }
       if ((table_sort.sort_keys=
            make_char_array(table_sort.sort_keys,
-                           param.keys, param.rec_length, MYF(0))))
+                           param.keys, param.rec_length, MYF(MY_ZEROFILL))))
```
dutow: A bit more info here: this doesn't result in taking any incorrect paths - it's only about copying random memory garbage around, and I think we also write it to the disk sometimes.

If we say that this is ok to do, I could mark these allocations as initialized instead of really initializing them.

cmake/msan-blacklist.txt
```diff
@@ -0,0 +1 @@
+fun:longest_match
```
laurynas-biveinis: This is not mentioned in commit message?

sql/filesort.cc
laurynas-biveinis: Unless there is a noticeable performance hit, better to initialize. BTW, there used to be Valgrind errors on "sycall pointing to uninit bytes" which match the description of I/O on uninit data here.

cmake/msan-blacklist.txt
dutow: I mentioned it now (disabling a false positive in zlib)

dutow:
Another update:

* main.bug80134 isn't an issue. Testcase requires an envinoment variable: `MSAN_OPTIONS="allocator_may_return_null=1"`
* main.myisampack fixed - at least at the occurence repoprted by the testcase, but I suspect that there could be other similar issues (caused by an "optimized" huffman decoder, which expects that it can read 3 bytes after the buffer's end, but sometimes the memory there wasn't initialized)
* main.mysql_client_test partially fixed: the test now runs to a point, but then in reports an error: last_insert_id returns an unexpected value, and the record really has a differnt ID compared to the non-msan build, which passes with the same codebase...

And discovered another issue: the terminfo library has no support in msan, so calls involving that reports errors. This means that the `mysql` command line tool can't be started up simply in a terminal, but when used with a file redirect to stdin, it works.

dutow:
Updated again: based on a slack discussion with Yura, I removed actual bugfixes and will submit them as separate PRs. This PR is now only about:

* adding the WITH_MSAN option
* running the mtr tests with msan

percona-ysorokin (approved):
LGTM

laurynas-biveinis (approved):

dutow:
Jenkins: https://jenkins.percona.com/job/percona-server-5.5-param/1640/

PR #2066, opened 2018-01-11
PS-1525: Adding the innodb_encrypt_tables system variable.

dutow:
Possible values: ON, OFF, FORCE.

If an InnoDb table is created with this set to ON or FORCE without
any encryption setting specified, it will be created encrypted, and
the encryption setting will show up in it's creation info.


storage/innobase/handler/handler0alter.cc
```diff
@@ -556,11 +556,11 @@ ha_innobase::check_if_supported_inplace_alter(
...
 	/* We don't support change encryption attribute with
 	inplace algorithm. */
-	char*	old_encryption = this->table->s->encrypt_type.str;
+	const bool currently_encrypted =
+		m_prebuilt->table->flags2 & DICT_TF2_ENCRYPTION;
 	char*	new_encryption = altered_table->s->encrypt_type.str;
 
-	if (Encryption::is_none(old_encryption)
-	    != Encryption::is_none(new_encryption)) {
+	if (currently_encrypted != !Encryption::is_none(new_encryption)) {
```
dutow: I'm quite sure that I'll have to add another check here: with this, currently we allow inplace alterations for non encrypted tables, even when innodb_encrypt_tables is FORCE. This is consistent with mariadb, but I don't think it's logical. Currently waiting for some opinions about this.

mysql-test/suite/innodb/t/encryption_force.test
```diff
@@ -0,0 +1,107 @@
...
+call mtr.add_suppression("\\[Error\\] InnoDB: Can't generate new master key for tablespace encryption, please check the keyring plugin is loaded.");
+call mtr.add_suppression("\\[Error\\] InnoDB: Encryption can't find master key, please check the keyring plugin is loaded.");
+call mtr.add_suppression("\\[ERROR\\] Function 'keyring_file' already exists");
+call mtr.add_suppression("\\[ERROR\\] Couldn't load plugin named 'keyring_file' with soname 'keyring_file.*'.");
+call mtr.add_suppression("Plugin keyring_file reported");
+
+# Restart the server with keyring loaded
+--let $restart_parameters="$keyring2_restart_param"
+--let $restart_hide_args= 1
+--source include/restart_mysqld.inc
```
laurynas-biveinis: This is the sole restart in the testcase - can it be performed by the testcase options file?

mysql-test/suite/innodb/t/encryption_force.test
```diff
@@ -0,0 +1,107 @@
+-- source include/have_innodb.inc
```
laurynas-biveinis: I'd name "percona_force_encryption" so that it's immediately visible in the runs as a Percona feature testcase

mysql-test/suite/sys_vars/r/all_vars.result
```diff
@@ -15,6 +15,8 @@ left join t1 on variable_name=test_name where test_name is null ORDER BY variabl
 There should be *no* variables listed below:
 DISABLED_STORAGE_ENGINES
 DISABLED_STORAGE_ENGINES
+INNODB_ENCRYPT_TABLES
```
laurynas-biveinis: Please review this testcase, specifically the "There should be *no* variables listed below:" comment above :)

As you create sys_vars.innodb_encrypt_tables_basic, move the sysvar tests from the feature testcase

mysql-test/suite/innodb/t/encryption_force.test
```diff
@@ -0,0 +1,107 @@
...
+
+insert into t1_default_explicit_on values (42);
+insert into t2_default_explicit_off values (42);
+insert into t3_on_default values (42);
+insert into t4_on_explicit_off values (42);
+insert into t5_off_explicit_off values (42);
+insert into t6_force_default values (42);
+
+--error ER_WRONG_VALUE_FOR_VAR
+set global innodb_encrypt_tables='FOO';
```
laurynas-biveinis: Move this elsewhere (see the comment at sys_vars.all_vars.result)

mysql-test/suite/innodb/t/encryption_force.test
```diff
@@ -0,0 +1,107 @@
...
+alter table t1_default_explicit_on encryption="N";
+alter table t2_default_explicit_off encryption="Y";
+--error ER_INVALID_ENCRYPTION_OPTION
+alter table t2_default_explicit_off encryption="N";
+alter table t3_on_default;
+--error ER_INVALID_ENCRYPTION_OPTION
+alter table t4_on_explicit_off encryption="N";
+
+set global innodb_encrypt_tables='ON';
+--error 1846
```
laurynas-biveinis: s/1846/ER_something/g

storage/innobase/handler/ha_innodb.cc
```diff
@@ -2681,6 +2681,22 @@ Encryption::is_none(const char* algorithm)
...
+/** Check if the NO algorith was explicitly specified.
+@param[in]      algorithm       Encryption algorithm to check
+@return true if no algorithm explicitly requested */
+bool
+Encryption::none_specified(const char* algorithm)
+{
+	/* NULL is the same as NONE */
+	if (algorithm != NULL && (
+	    innobase_strcasecmp(algorithm, "n") == 0
+	    || innobase_strcasecmp(algorithm, "") == 0)) {
```
laurynas-biveinis: maybe just algorithm[0] == '\0'

storage/innobase/handler/ha_innodb.cc
```diff
@@ -11080,7 +11096,11 @@ create_table_info_t::create_table_def()
 
 		const char*	encrypt = m_create_info->encrypt_type.str;
 
-		if (!Encryption::is_none(encrypt)) {
+		const bool default_to_encrypt =
+			!Encryption::none_specified(encrypt)
```
laurynas-biveinis: For reader very hard to tell the difference between "none_specified" and "is_none", consider renaming

storage/innobase/handler/ha_innodb.cc
```diff
@@ -11727,6 +11747,14 @@ create_table_info_t::create_option_encryption_is_valid() const
 	bool table_is_encrypted =
 		!Encryption::is_none(m_create_info->encrypt_type.str);
 
+	if(srv_encrypt_tables == 2
```
laurynas-biveinis: Space after if

laurynas-biveinis: Introduce enum for 0, 1, 2 here

storage/innobase/handler/ha_innodb.cc
```diff
@@ -11989,6 +12017,11 @@ ha_innobase::update_create_info(
 	    && create_info->tablespace == NULL) {
 		create_info->tablespace = m_prebuilt->table->tablespace;
 	}
+	if((m_prebuilt->table->flags2 & DICT_TF2_ENCRYPTION)
```
laurynas-biveinis: Space after if

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21005,6 +21038,19 @@ static MYSQL_SYSVAR_ULONG(autoextend_increment,
...
   NULL, NULL, 64L, 1L, 1000L, 0);
 
+static const char* srv_encrypt_tables_names[] = { "OFF", "ON", "FORCE", 0 };
+static TYPELIB srv_encrypt_tables_typelib = {
+	array_elements(srv_encrypt_tables_names)-1, 0, srv_encrypt_tables_names,
+	NULL
+};
+static MYSQL_SYSVAR_ENUM(encrypt_tables, srv_encrypt_tables,
+			 PLUGIN_VAR_OPCMDARG,
+			 "Enable encryption for tables. ",
```
laurynas-biveinis: Describe the effect of the three values

storage/innobase/handler/handler0alter.cc
laurynas-biveinis: Consider "==" instead of "!= !"

laurynas-biveinis (changes requested):
Please file a JIRA issue and use PS-XXX in the commit message header

dutow:
Everything's fixed, I just tried to do a successful jenkins run... but wasn't able, as usual for 5.7.

So this is it: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1503/

Most platforms have warnings/network errors, bun two builds succeeded without any warnings.

mysql-test/suite/innodb/t/percona_force_encryption.test
```diff
@@ -0,0 +1,96 @@
...
+set global innodb_encrypt_tables='FORCE';
+
+insert into t1_default_explicit_on values (42);
+insert into t2_default_explicit_off values (42);
+insert into t3_on_default values (42);
+insert into t4_on_explicit_off values (42);
+insert into t5_off_explicit_off values (42);
+insert into t6_force_default values (42);
+
+--error ER_WRONG_VALUE_FOR_VAR
```
laurynas-biveinis: Remove, now covered by the sys_vars test

mysql-test/suite/innodb/t/percona_force_encryption.test
```diff
@@ -0,0 +1,96 @@
...
+alter table t1_default_explicit_on encryption="Y", ALGORITHM=INPLACE;
+alter table t1_default_explicit_on encryption="N";
+alter table t2_default_explicit_off encryption="N";
+--error ER_ALTER_OPERATION_NOT_SUPPORTED_REASON
+alter table t3_on_default encryption="N", ALGORITHM=INPLACE;
+alter table t3_on_default encryption="Y", ALGORITHM=INPLACE;
+alter table t3_on_default encryption="N";
+alter table t4_on_explicit_off ALGORITHM=INPLACE;
+alter table t4_on_explicit_off encryption="N";
+--error 1846
```
laurynas-biveinis: s/1846/ER_...

mysql-test/suite/innodb/t/percona_force_encryption.test
```diff
@@ -0,0 +1,96 @@
...
+select * from t6_force_default;
+
+drop table t1_default_explicit_on;
+drop table t2_default_explicit_off;
+drop table t3_on_default;
+drop table t4_on_explicit_off;
+drop table t5_off_explicit_off;
+drop table t6_force_default;
+
+#Cleanup
```
laurynas-biveinis: Space after #

storage/innobase/handler/ha_innodb.cc
```diff
@@ -2681,6 +2681,21 @@ Encryption::is_none(const char* algorithm)
 	return(false);
 }
 
+/** Check if the NO algorith was explicitly specified.
```
laurynas-biveinis: s/algorith/algorithm

storage/innobase/handler/ha_innodb.cc
```diff
@@ -2681,6 +2681,21 @@ Encryption::is_none(const char* algorithm)
...
+@param[in]      algorithm       Encryption algorithm to check
+@return true if no algorithm explicitly requested */
+bool
+Encryption::none_explicitly_specified(const char* algorithm)
+{
+	/* NULL is the same as NONE */
+	if (algorithm != NULL &&
```
laurynas-biveinis: {
return (algorithm != NULL && innobase_strcmp() == 0);
}

storage/innobase/include/os0file.h
```diff
@@ -448,12 +448,18 @@ struct Encryption {
...
-        /** Check if the string is "empty" or "none".
-        @param[in]      algorithm       Encryption algorithm to check
-        @return true if no algorithm requested */
+	/** Check if the string is "" or "n".
+	@param[in]      algorithm       Encryption algorithm to check
+	@return true if no algorithm requested */
 	static bool is_none(const char* algorithm)
 		MY_ATTRIBUTE((warn_unused_result));
 
+	/** Check if the NO algorithm was explicitly speficied.
```
laurynas-biveinis: "specified"

laurynas-biveinis (changes requested):
Previous review comment "Please file a JIRA issue and use PS-XXX in the commit message header" still stands

dutow:
New jenkins run: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1510/

I also deleted a comment from `none_explicitly_specified`, which was misleading (I copied that from `is_none`, but it wasn't true for the new function)

mysql-test/suite/innodb/t/percona_force_encryption.test
```diff
@@ -0,0 +1,93 @@
+-- source include/have_innodb.inc
```
laurynas-biveinis: If you have to re-push due to other review comments, then no space between -- and source

mysql-test/suite/innodb/t/percona_force_encryption.test
```diff
@@ -0,0 +1,93 @@
...
+-- source include/not_embedded.inc
+-- source include/have_partition.inc
```
laurynas-biveinis: Partitioning is not used in the testcase - adding such tests might be a good idea

mysql-test/suite/sys_vars/t/innodb_encrypt_tables_basic.test
```diff
@@ -0,0 +1,27 @@
...
+--error ER_WRONG_TYPE_FOR_VAR
+SET GLOBAL innodb_encrypt_tables=1e1;
+--error ER_WRONG_VALUE_FOR_VAR
+SET GLOBAL innodb_encrypt_tables='foo';
+
+SELECT @@GLOBAL.innodb_encrypt_tables;
+
+# If we are lacking permissions, then we cannot restore the startup value
+# at the end.
+--source include/restart_mysqld.inc
```
laurynas-biveinis: ? If we are lacking permissions, then we could not switch away from the startup value neither. And how can the MTR MySQL user lack permissions?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -11989,6 +12018,11 @@ ha_innobase::update_create_info(
 	    && create_info->tablespace == NULL) {
 		create_info->tablespace = m_prebuilt->table->tablespace;
 	}
+	if ((m_prebuilt->table->flags2 & DICT_TF2_ENCRYPTION)
+			&& create_info->encrypt_type.length == 0) {
+		static const LEX_STRING yes = { C_STRING_WITH_LEN("Y") };
+		create_info->encrypt_type = yes;
```
laurynas-biveinis: Use of stack variable after scope? Just in case, please grep, such LEX_STRING/LEX_CSTRING might be already available elsewhere for reuse

laurynas-biveinis (changes requested):

mysql-test/suite/sys_vars/t/innodb_encrypt_tables_basic.test
dutow: I based this on another testcase, and this was included in it - I assumed it made sense. I'll remove it then.

storage/innobase/handler/ha_innodb.cc
dutow: I don't see a stack variable? I only see a static const.
But it would also work with a stack variable, because the assignment calls the copy constructor, and the struct only contains a const char pointer to a string literal, and an integer (the length)

laurynas-biveinis: Right, sorry

dutow:
Comments fixed.
New jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1530/

mysql-test/suite/innodb/t/percona_force_encryption.test
```diff
@@ -0,0 +1,92 @@
...
+
+drop table t1_default_explicit_on;
+drop table t2_default_explicit_off;
+drop table t3_on_default;
+drop table t4_on_explicit_off;
+drop table t5_off_explicit_off;
+drop table t6_force_default;
+
+# Cleanup
+set global innodb_encrypt_tables='OFF';
```
percona-ysorokin: save the value of the `@@global.innodb_encrypt_tables` system variable at the beginning of the test and restore it here.

mysql-test/suite/perfschema/r/show_sanity.result
```diff
@@ -412,6 +412,7 @@ SHOW_MODE	SOURCE	VARIABLE_NAME
 5.6	I_S.SESSION_VARIABLES	INNODB_COMPRESSED_COLUMNS_THRESHOLD
 5.6	I_S.SESSION_VARIABLES	INNODB_COMPRESSED_COLUMNS_ZIP_LEVEL
 5.6	I_S.SESSION_VARIABLES	INNODB_DEADLOCK_DETECT
+5.6	I_S.SESSION_VARIABLES	INNODB_ENCRYPT_TABLES
```
percona-ysorokin: The proper way to fix this is to extend `show_sanity.test` itself with a new known global variable rather than re-recording `.result` file.

mysql-test/suite/perfschema/r/show_sanity.result
```diff
@@ -441,6 +442,7 @@ SHOW_MODE	SOURCE	VARIABLE_NAME
...
```
percona-ysorokin: Likewise

sql/sql_show.cc
```diff
@@ -1923,7 +1923,11 @@ int store_create_info(THD *thd, TABLE_LIST *table_list, String *packet,
     if (table->s->encrypt_type.length)
     {
       packet->append(STRING_WITH_LEN(" ENCRYPTION="));
-      append_unescaped(packet, share->encrypt_type.str, share->encrypt_type.length);
+      append_unescaped(packet, table->s->encrypt_type.str, table->s->encrypt_type.length);
```
percona-ysorokin: Isn't `share` is just a synonym for `table->s` here?

sql/sql_show.cc
```diff
@@ -1923,7 +1923,11 @@ int store_create_info(THD *thd, TABLE_LIST *table_list, String *packet,
...
+    } else if (create_info.encrypt_type.length)
```
percona-ysorokin: Why do we need another branch here? Shouldn't `table->s->encrypt_type` be always populated with actual encryption type value?

percona-ysorokin (changes requested):

sql/sql_show.cc
dutow: It's... similar, but not exactly the same. table->s is based on the sql parser. If the option was specified in the SQL query, it is in it.

But if a table is encrypted because of the variable, only the innodb flag is set, the information during the parsing is unchanged.

I found no good way to change it, but the `create_info` structure can be altered by the virtual `update_create_info` function, which I modified for innodb.

sql/sql_show.cc
dutow: It is , but the condition used the other name - I thought it would be clearer to use the same on both places.

mysql-test/suite/innodb/t/percona_force_encryption.test
laurynas-biveinis: Then probably "select @@innodb_encrypt_tables;" at line 10 is redundant too

dutow:
Test cases updated.

mysql-test/suite/perfschema/t/show_sanity.test
```diff
@@ -760,7 +760,8 @@ insert into test.sanity values
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "VERSION_COMMENT"),
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "VERSION_COMPILE_MACHINE"),
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "VERSION_COMPILE_OS"),
-  ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "REQUIRE_SECURE_TRANSPORT")
+  ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "REQUIRE_SECURE_TRANSPORT"),
```
laurynas-biveinis: This list is maintained sorted

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21005,6 +21039,22 @@ static MYSQL_SYSVAR_ULONG(autoextend_increment,
...
   NULL, NULL, 64L, 1L, 1000L, 0);
 
+static TYPELIB srv_encrypt_tables_typelib = {
+	array_elements(srv_encrypt_tables_names)-1, 0, srv_encrypt_tables_names,
+	NULL
+};
+static MYSQL_SYSVAR_ENUM(encrypt_tables, srv_encrypt_tables,
+			 PLUGIN_VAR_OPCMDARG,
+			 "Enable encryption for tables. "
+			 "When turned ON, the default encryption setting is yes when it isn't "
```
laurynas-biveinis: Maybe "When turned ON, all tables are created encrypted unless specified otherwise."

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21005,6 +21039,22 @@ static MYSQL_SYSVAR_ULONG(autoextend_increment,
...
+			 "specified. When it's set to FORCE, only encrypted tables can be created."
+			 "The FORCE setting also disables non inplace alteration of unencryted tables,"
```
laurynas-biveinis: "unencryted"
Also maybe "The FORCE setting also disables ALTER TABLE operations for unencrypted tables that would cause table rebuilds"

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21005,6 +21039,22 @@ static MYSQL_SYSVAR_ULONG(autoextend_increment,
...
+			 " without encryting them in the process.",
```
laurynas-biveinis: "encryting"

laurynas-biveinis (changes requested):
Minor comments only, I suggest to wait for @percona-ysorokin review before addressing/testing them

sql/sql_show.cc
percona-ysorokin: Still not sure why we have unsynchronized data here. A handler for `SHOW CREATE TABLE` (`mysqld_show_create()`) before calling `store_create_info()` function calls `open_tables()` - a function which should populate all the table attributes (including encryption flag). This `open_tables()` is used all over the code (not only for `SHOW CREATE TABLE`) and if it doesn't do this, we have a problem.

mysql-test/suite/innodb/t/percona_force_encryption.test
```diff
@@ -0,0 +1,92 @@
...
+select * from t5_off_explicit_off;
+select * from t6_force_default;
+
+drop table t1_default_explicit_on;
+drop table t2_default_explicit_off;
+drop table t3_on_default;
+drop table t4_on_explicit_off;
+drop table t5_off_explicit_off;
+drop table t6_force_default;
+
```
percona-ysorokin: Please, also add tests for how `CREATE TABLE LIKE <existing_table>` works for different combinations of `innodb_encrypt_tables` and encryption attribute of `<existing_table>`.

percona-ysorokin (changes requested):

percona-ysorokin:
Another big question is whether `ENCRYPTION` attribute should be replicated depending on the value of `innodb_encrypt_tables` on the slave?

mysql-test/suite/innodb/t/percona_force_encryption.test
dutow: I added two new tables, their name ending with _dup

dutow:
Updated PR:
* added the adjust_create_info_for_frm function to the handles class. This function allows any storage engine to modify the table definition before it's written into the frm file
* rebased to latest 5.7 head

mysql-test/suite/innodb/t/percona_force_encryption.test
laurynas-biveinis: There are no DROP TABLE for _dup tables?

mysql-test/suite/innodb/t/percona_force_encryption.test
```diff
@@ -0,0 +1,96 @@
+--source include/have_innodb.inc
```
laurynas-biveinis, line 1: Since the feature goes through .FRM, its cold reading should be tested: add FLUSH TABLES, query the same tables again

storage/innobase/handler/ha_innodb.cc
```diff
@@ -11958,6 +11983,20 @@ create_table_info_t::create_options_are_invalid()
 	return(ret);
 }
 
+void
+ha_innobase::adjust_create_info_for_frm(
+/*============================*/
+	HA_CREATE_INFO*	create_info)	/*!< in/out: create info */
```
laurynas-biveinis: New-InnoDB doxygen style header comment

laurynas-biveinis (changes requested):

dutow:
PR updated.

mysql-test/suite/innodb/t/percona_force_encryption.test
```diff
@@ -0,0 +1,111 @@
...
+
+call mtr.add_suppression("\\[Error\\] InnoDB: Can't generate new master key for tablespace encryption, please check the keyring plugin is loaded.");
+call mtr.add_suppression("\\[Error\\] InnoDB: Encryption can't find master key, please check the keyring plugin is loaded.");
+call mtr.add_suppression("\\[ERROR\\] Function 'keyring_file' already exists");
+call mtr.add_suppression("\\[ERROR\\] Couldn't load plugin named 'keyring_file' with soname 'keyring_file.*'.");
+call mtr.add_suppression("Plugin keyring_file reported");
+
+let $initial_encrypt_tables_value = `select @@innodb_encrypt_tables`;
+
+create table t1_default_explicit_on (a int) engine=innodb encryption="Y";
```
laurynas-biveinis, line 16: I'm sorry if this was discussed already, but what about partitioned tables?

dutow: They should work the same way as when you specify the encrypt flag manually - this change only adds the encrypt flag to the create table statement if it's not present, and reports an error if it is no and we want to farce encryption.

I'll add a few tests with them.

dutow:
Jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1612/

mysql-test/suite/innodb/t/percona_force_encryption.test
dutow: Added test cases with partitions / tablespaces.

laurynas-biveinis (approved):

storage/innobase/handler/ha_innodb.cc
```diff
@@ -11958,6 +11983,19 @@ create_table_info_t::create_options_are_invalid()
 	return(ret);
 }
 
+void
+ha_innobase::adjust_create_info_for_frm(
+	HA_CREATE_INFO*	create_info)
+{
+		if (create_info->encrypt_type.length == 0
+			&& create_info->encrypt_type.str == NULL
+			&& srv_encrypt_tables != 0
```
percona-ysorokin: `srv_encrypt_tables != SRV_ENCRYPT_TABLES_OFF`

storage/innobase/handler/ha_innodb.cc
```diff
@@ -11958,6 +11983,19 @@ create_table_info_t::create_options_are_invalid()
...
+			) {
+				static const LEX_STRING yes = { C_STRING_WITH_LEN("Y") };
```
percona-ysorokin: Function-level static variables/constants are not potentially thread-safe in c++98 (even for POD data types). Please, move to the global level or class level
`/*static*/ const LEX_STRING ha_innobase::yes = { C_STRING_WITH_LEN("Y") };`

percona-ysorokin (changes requested):

dutow:
Updated.

percona-ysorokin (approved):
LGTM

dutow:
Jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1646/

The centos6 build failed because there was no space on the disk.

PR #2069, opened 2018-01-15
Fix attempt for DOC-444: note on dbl-quotes and semicolon in early-plugin-load

fiowro:
...the only variable of this family which appears in percona-server documentation. Discussion in comments for this issue had settled down with the following statements: "loading plugins is described in the upstream docs" (i), "the problem of semicolon with or without quotes is related to bash scripting and not percona-server" (ii), and "MyRocks and TokuDB installation instructions propose using ps-admin, not command line options" (iii).  So perhaps this is the only place such comment is reasonable to reside.

### DIFF
--- a/doc/source/management/data_at_rest_encryption.rst
+++ b/doc/source/management/data_at_rest_encryption.rst
@@ -145,6 +145,13 @@ option:
 It should be loaded this way to be able to facilitate recovery for encrypted
 tables.
 
+.. warning::
+
+  If server should be started with several plugins loaded early,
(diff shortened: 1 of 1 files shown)


laurynas-biveinis (changes requested):
Not sure, but s/semicolon/semicolons/g ?

fiowro:
Well, plural form of semicolon is certainly good in this case, so switched to it :)

laurynas-biveinis (approved):

PR #2070, opened 2018-01-16
Fix bug 1741647 (5.7.20-19 build error on macOS)

laurynas-biveinis:
macOS build has been broken by
1) memset_s use has been introduced in core server with incorrect
CMake portability check
2) keyring_vault and keyring_file symbol visibility has been limited
with a version script, which is not supported by macOS linker.

The first issue is fixed trivially, the second one by limiting the
version script use to Linux, if no sanitizers are enabled, to match
the existing behavior in libmysql.

https://jenkins.percona.com/job/mysql-5.7-param/1461/


plugin/keyring_vault/CMakeLists.txt
```diff
@@ -49,8 +48,10 @@ MYSQL_ADD_PLUGIN(keyring_vault
                  MODULE_ONLY
                  MODULE_OUTPUT_NAME "keyring_vault")
 
-get_property(keyring_vault_link_flags TARGET keyring_vault PROPERTY LINK_FLAGS)
-set_property(TARGET keyring_vault PROPERTY LINK_FLAGS "${keyring_vault_link_flags} -Wl,--version-script=${CMAKE_SOURCE_DIR}/plugin/keyring_vault/keyring_vault.version")
+IF(LINK_FLAG_NO_UNDEFINED)
```
robgolebiowski: Maybe a word of comment that LINK_FLAG_NO_UNDEFINED is defined for Linux system only ? What if confusing about this flag that only libmysql and libmysqld actually use it, all other plugins export undefined symbols. This is optional comment and might be skipped.

robgolebiowski (approved):
LGTM, with one optional change.

plugin/keyring_vault/CMakeLists.txt
laurynas-biveinis: As discussed on Slack, pushing with "# We limit symbols exported on Linux to only those required by server." added

PR #2072, opened 2018-01-16
Ps 257 5.6

laurynas-biveinis:
https://jenkins.percona.com/job/percona-server-5.6-param/2044/


laurynas-biveinis:
I'll assume @robgolebiowski approval of 5.5 and 5.7 versions extends to 5.6 too

PR #2077, opened 2018-01-18
PS-3780: Fixing release errors with GCC7

dutow:
* Changed the type of m_gtid_purged from an enum to an ulong, as it's used with TYPELIB.
* Removed basically broken c style casts of m_gtid_purged to ulong*
* Changed C++ standard handling in the CMake files from hand written CXXFLAGS to the standard variables
  This is required because modern compilers default to C++14,which results in a long list of deprecation
  warnings for MySQL code, and the previous solution only set it explicitly to C++03 for debug builds -
  release builds compiled with C++14 mode, which could even cause semantic differences.
* Fixed an snprintf call in update_timestamp
* Moved update_timestamp to the cpp file from the header, as my_snprintf isn't available in the header
* Added a missing mutex unlock in an execution path which should never happen
* Fixed an unused variable in the release buld of a test, and made the tests's return code the result

### DIFF
--- a/client/dump/sql_formatter_options.h
+++ b/client/dump/sql_formatter_options.h
@@ -56,7 +56,12 @@ public:
   bool m_suppress_create_database;
   bool m_timezone_consistent;
   bool m_skip_definer;
-  enum enum_gtid_purged_mode m_gtid_purged;
+  // Storing m_gtid_purged as an ulong, because its setter code treats it as a
+  // long variable.
+  // Storing it as an enum causes issues on some systems beacuse of endianness,
+  // and only works because of the alignment of the next pointer on 64 bit
+  // systems.
+  ulong m_gtid_purged;
   const Mysql_chain_element_options* m_mysql_chain_element_options;
 
   const TYPELIB* get_gtid_purged_mode_typelib()
--- a/sql/rpl_reporting.h
+++ b/sql/rpl_reporting.h
@@ -101,24 +101,7 @@ public:
 
     }
 
-    void update_timestamp()
-    {
-      struct tm tm_tmp;
-      struct tm *start;
-
-      skr= my_time(0);
-      localtime_r(&skr, &tm_tmp);
-      start=&tm_tmp;
-
-      sprintf(timestamp, "%02d%02d%02d %02d:%02d:%02d",
-              start->tm_year % 100,
-              start->tm_mon+1,
-              start->tm_mday,
-              start->tm_hour,
-              start->tm_min,
-              start->tm_sec);
-      timestamp[15]= '\0';
-    }
+    void update_timestamp();
 
     /** Error code */
     uint32 number;
--- a/storage/rocksdb/CMakeLists.txt
+++ b/storage/rocksdb/CMakeLists.txt
@@ -1,4 +1,5 @@
 # TODO: Copyrights
+SET(CMAKE_CXX_STANDARD 11)
 
 # by default MyRocks is not built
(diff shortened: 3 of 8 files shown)


client/dump/sql_formatter_options.h
```diff
@@ -56,7 +56,11 @@ class Sql_formatter_options
   bool m_suppress_create_database;
   bool m_timezone_consistent;
   bool m_skip_definer;
-  enum enum_gtid_purged_mode m_gtid_purged;
+  // Storing m_gtid_purded as an ulong, because the it's setert code
```
laurynas-biveinis: "purded", "setert"

laurynas-biveinis (changes requested):
Meanwhile, MySQL 5.7.21 came out with the following in the release notes: "MySQL did not compile with GCC 7. (Bug #25643811, Bug #26825211)" - please reconcile, commits ba1a99c5cd7 and acad90d206f.

client/dump/sql_formatter_options.h
dutow: Uh, I thought I edited this comment. Fixed.

dutow:
Updated the PR, as it had issues on some debian/centos versions.

The main issue was with enabling the gnu extensions for building tokudb, I disabled that in `check_stdcxx11`

Jenkins (still running, but it got past the previous errors: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1528/

sql/rpl_reporting.h
```diff
@@ -101,24 +101,7 @@ class Slave_reporting_capability
...
-    void update_timestamp()
-    {
-      struct tm tm_tmp;
-      struct tm *start;
-
-      skr= my_time(0);
-      localtime_r(&skr, &tm_tmp);
-      start=&tm_tmp;
-
-      sprintf(timestamp, "%02d%02d%02d %02d:%02d:%02d",
```
laurynas-biveinis, line 113: What was fixed there?

storage/rocksdb/CMakeLists.txt
```diff
@@ -34,6 +35,13 @@ IF ((CMAKE_CXX_COMPILER_ID STREQUAL GNU) AND
   RETURN ()
 ENDIF ()
 
+# Relax unused variable warnings for release modes
```
laurynas-biveinis, line 38: Why?

sql/rpl_reporting.h
dutow: With sprintf, gcc belives that this could overflow (it can't, as it's 15 chars max + null terminator). So I changed it to my_snprintf, which silenced the warning, but that function isn't available in the header, so I moved the implementation to the cpp file.

(and this is a point where mysql fixed this differently: they simply increased the buffer size to 64 bytes... which just makes no sense)

storage/rocksdb/CMakeLists.txt
dutow: This is also part of the clang 5.7 PR too: because it's in a submodule owned by facebook, and it contains unused variables in release mode. So we'll have to silence the warning until it's fixed.

laurynas-biveinis (approved):

laurynas-biveinis:
Ah, I missed pointing out that there should be a bug report which should be referenced in the commit message. Might be too late depending on what's the status of GCC 7 fixes for 5.5 and 5.6.

dutow:
Created a jira ticket for this / updated the commit message

laurynas-biveinis (approved):

PR #2079, opened 2018-01-22
Fix bug 1660628 (-Wmaybe-uninitialized sql_planner.cc compilation war…

inikep:
…nings)

Disable warnings for save_to_position() in sql/sql_planner.cc, as the
uninitialization itself is intentional.

It is improved version of a bug fix from Laurynas:
https://github.com/laurynas-biveinis/percona-server/commit/000e5c8a94ef7d18ccec22040d5e9ad1e1e2ea72


inikep:
Jenkins tests:
https://jenkins.percona.com/job/percona-server-5.6-param/2052/

sql/sql_planner.cc
```diff
@@ -118,14 +118,6 @@ class Loose_scan_opt
       but initializing them would cause a 2% CPU time loss in a 20-table plan
       search. So we initialize only if warnings would stop the build.
```
laurynas-biveinis: "So we initialize only if warnings would stop the build" comment becomes outdated and needs removing

laurynas-biveinis (approved):
OK to push after minor comment correction and having prepared 5.7 branch (which is a null merge?)

percona-ysorokin (approved):
LGTM

PR #2080, opened 2018-01-22
PS-276: Avoid long buffer wait times on busy servers

dutow:
This was caused by buf_flush_wait_flushed, which implemented waiting
for flushing the latest LSN the following way:

For every buffer pool:
 * check latest LSN
 * wait until the LSN is flushed

With multiple buffer pools, when other sessions also resulted in
buffer writes, this resulted in the later buffers waiting for
more recent flushes.

To make sure that the function only waits to the point of the
latest LSNs at the moment of its calls, the implementation was
changed to:

For every buffer pool:
* check the latest LSN
For every buffer pool:
* wait until the previuosly checked/stored LSN is flushed

### DIFF
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -2091,10 +2091,10 @@ void
 buf_flush_wait_flushed(
 	lsn_t		new_oldest)
 {
+	lsn_t		instance_newest[MAX_BUFFER_POOLS];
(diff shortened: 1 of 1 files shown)


laurynas-biveinis (approved):

dutow:
Jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1511/

dutow:
A minor change: fixed a few typos in the commit message / PR description

PR #2083, opened 2018-01-23
Fix bug 1733781 (sql/handler.cc:3008: int handler::ha_close(): Assert…

laurynas-biveinis:
…ion `m_lock_type == 2' failed)

The issue is that temporary table locks are not properly released
before closing their tables, for example, if a temp table is opened,
locked, and a format description binary log event is processed, which
forces closing it. This was introduced by our fix for bug 1953904 /
https://bugs.mysql.com/bug.php?id=83003 not integrating the changes of
a corresponding upstream bugfix [1] properly. Fix by introducing a
mysql_lock_remove call in a code path introduced by our fix to 83003
to mirror the upstream fix.

[1]:

d46c1c50885f39e204ea589cd472a67e4b81013e
Author: Dyre Tjeldvoll <Dyre.Tjeldvoll@oracle.com>
AuthorDate: Fri Jan 23 10:36:58 2015 +0100
Commit: Dyre Tjeldvoll <Dyre.Tjeldvoll@oracle.com>
CommitDate: Wed Jan 28 09:40:41 2015 +0100

Parent: 29e683ff841 Disbaled testcases which are failing due to Bug#20415775, BUG#20427694.
Containing: 5.7 8.0 binlog_no_sync_before_close bug1540338-5.7 bug1553883 bug1583589-5.7 bug1618718-5.7 (35 more)
Follows: mysql-5.7.5 (1751)
Precedes: mysql-5.7.6 (474)

BUG#19894987: HANDLE_FATAL_SIGNAL (SIG=11) IN HA_INDEX_OR_RND_END | SQL/HANDLER.H:1999

Execution of certain BINLOG statements while having temporary tables
open by HANDLER statements caused server crash.

Execution of BINLOG statements for Start/Format_description events
may close the connection's temporary tables. If a handler had been associated
with a temporary table, the BINLOG statement's call to close_temporary_tables()
would result in dangling pointers to the now freed TABLE in
thd->handler_table_hash. Similarly, a dangling pointer would also be left if
there was a lock on the temporary table. Running with valgrind showed that
freed memory was being accessed.

Solution: Properly close handlers and remove locks associated with temporary
tables being closed by the BINLOG statement.

https://jenkins.percona.com/job/mysql-5.7-param/1513/

I have considered adding the testcase to lower versions too, but it needs server-id=0, which is only available starting with 5.7.


mysql-test/suite/binlog/t/percona_bug1733781.test
```diff
@@ -0,0 +1,17 @@
...
+--echo #
+
+--let $assert_text= server-id must be zero
+--let $assert_cond= @@server_id = 0
+--source include/assert.inc
+
+CREATE TEMPORARY TABLE t4(f int) ENGINE=INNODB;
+LOCK TABLES t4 WRITE;
+# Format description event that forces closing of open temp tables
+BINLOG ' SOgWTg8BAAAAbgAAAHIAAAAAAAQANS42LjMtbTUtZGVidWctbG9nAAAAAAAAAAAAAAAAAAAAAAAA AAAAAAAAAAAAAAAAAABI6BZOEzgNAAgAEgAEBAQEEgAAVgAEGggAAAAICAgCAAAAAAVAYI8=';
```
percona-ysorokin: In one of the tests I did (`main.bug75721`), changing `BINLOG ' SOgWTg8B` to `BINLOG ' SOgWTg8C` (changing 8th character) was enough to make `server-id=1`. So you can potentially get rid of `.opt` file and add this test to 5.6 as well.

percona-ysorokin (approved):
LGTM
(with minor suggestion for 5.6)

mysql-test/suite/binlog/t/percona_bug1733781.test
laurynas-biveinis: This is great - it crashed 5.6 (but not 5.5) too!

laurynas-biveinis:
https://jenkins.percona.com/job/mysql-5.7-param/1521/

percona-ysorokin (approved):
LGTM

PR #2090, opened 2018-01-24
Post-push fix for bug 1689268: include the missing plugin ABI check bits

laurynas-biveinis:
https://jenkins.percona.com/job/mysql-5.7-param/1520/


laurynas-biveinis:
Self-approving to fix trunk

PR #2091, opened 2018-01-24
Add testcase for bug 1733781 (sql/handler.cc:3008: int handler::ha_cl…

laurynas-biveinis:
…ose(): Assertion `m_lock_type == 2' failed.)

Add a regression testcase for a higher-version-only bug

https://jenkins.percona.com/job/percona-server-5.5-param/1626/


percona-ysorokin (approved):
LGTM

PR #2097, opened 2018-01-25
PS-2414: keyring_vault timeout MTR tests always producing warnings

robgolebiowski:
Suppressed warning "File '' not found". This is a correct warning
generated when keyring_vault is loaded without correct path to
configuration file. In this case the path is null, i.e. ''.

(fix for keyring_vault_timeout)


robgolebiowski:
self-approved

PR #2098, opened 2018-01-26
TDB-63: UPSERT/NOAR

vlad-lesin:
The detailed feature explanation can be found here:
https://www.percona.com/blog/2013/02/12/fast-updates-with-tokudb/

The current implementation contains the following changes:
- NOAR keyword is removed
- Instead of NOAR there are two separate system variables to turn on/off fast
update/upsert modes correspondingly:
    tokudb_enable_fast_update,
    tokudb_enable_fast_upsert.
- tokudb_disable_slow_update and tokudb_disable_slow_upsert system variables
are removed.

Testing: https://jenkins.percona.com/job/percona-server-5.6-param/2062/
See also: https://github.com/percona/percona-server/pull/2099


mysql-test/suite/tokudb/include/setup_fast_update_upsert.inc
```diff
@@ -0,0 +1,4 @@
+disable_warnings;
```
laurynas-biveinis: How is this setting up fast update / upsert?

mysql-test/suite/tokudb/t/disabled.def
```diff
@@ -2,27 +2,27 @@ mvcc-19: tokutek
...
-fast_update_uint_bounds : https://tokutek.atlassian.net/browse/DB-871
-fast_update_varchar : https://tokutek.atlassian.net/browse/DB-871
-fast_upsert_bin_pad : https://tokutek.atlassian.net/browse/DB-871
-fast_upsert_char : https://tokutek.atlassian.net/browse/DB-871
-fast_upsert_deadlock : https://tokutek.atlassian.net/browse/DB-871
-fast_upsert_int : https://tokutek.atlassian.net/browse/DB-871
-fast_upsert_key : https://tokutek.atlassian.net/browse/DB-871
-fast_upsert_sqlmode : https://tokutek.atlassian.net/browse/DB-871
-fast_upsert_values : https://tokutek.atlassian.net/browse/DB-871
+#fast_update_binlog_mixed : https://tokutek.atlassian.net/browse/DB-871
```
laurynas-biveinis: Remove don't comment out

mysql-test/suite/tokudb/t/fast_update_blobs_with_varchar.py
```diff
@@ -9,9 +9,9 @@ def main():
     print "source include/have_tokudb.inc;"
     print "source include/have_innodb.inc;"
     print "set default_storage_engine='tokudb';"
-    print "disable_warnings;"
     print "drop table if exists t;"
```
laurynas-biveinis: Drop as well

sql/handler.cc
```diff
@@ -7998,7 +7998,26 @@ int handler::ha_delete_row(const uchar *buf)
   return 0;
 }
 
+int handler::ha_fast_update(THD *thd,
```
laurynas-biveinis: Add header comment (but see handler.h comments first)

sql/handler.cc
```diff
@@ -7998,7 +7998,26 @@ int handler::ha_delete_row(const uchar *buf)
...
+  int error= fast_update(thd, update_fields, update_values, conds);
+  if (error == 0)
+    mark_trx_read_write();
+  return error;
+}
 
+int handler::ha_upsert(THD *thd,
+                       List<Item> &update_fields,
+                       List<Item> &update_values)
+{
```
laurynas-biveinis: Likewise

sql/handler.h
```diff
@@ -3145,7 +3145,27 @@ class handler :public Sql_alloc
 public:
  /* End of On-line/in-place ALTER TABLE interface. */
 
+  // Offload an update to the storage engine.
```
laurynas-biveinis: Convert comment to MySQL doxygen style

sql/handler.h
```diff
@@ -3145,7 +3145,27 @@ class handler :public Sql_alloc
...
+  // Returns 0 if the storage engine handled the update.
+  // Returns ENOTSUP if the storage engine can not handle the update and the slow update code path
+  // should continue executing the update.
+  // Returns an error if the update should be terminated.
+  int ha_fast_update(THD *thd, List<Item> &update_fields, List<Item> &update_values, Item *conds);
+
+  // Offload an upsert to the storage engine.
```
laurynas-biveinis: Likewise

sql/handler.h
```diff
@@ -3145,7 +3145,27 @@ class handler :public Sql_alloc
...
+  // Expects the row to be stored in record[0].
+  // Returns 0 if the storage engine handled the upsert.
+  // Returns ENOTSUP if the storage engine can not handle the upsert and the slow insert code path
+  // should continue executing the insert.
+  // Returns an error if the insert should be terminated.
+  int ha_upsert(THD *thd, List<Item> &update_fields, List<Item> &update_values);
+private:
+  virtual int fast_update(THD *thd, List<Item> &update_fields, List<Item> &update_values, Item *conds)
```
laurynas-biveinis: Add comment

sql/handler.h
```diff
@@ -3145,7 +3145,27 @@ class handler :public Sql_alloc
...
+  { return ENOTSUP; }
+  virtual int upsert(THD *thd, List<Item> &update_fields, List<Item> &update_values)
```
laurynas-biveinis: Add comment

sql/item.h
```diff
@@ -4051,6 +4051,7 @@ class Item_insert_value : public Item_field
   bool eq(const Item *item, bool binary_cmp) const;
   bool fix_fields(THD *, Item **);
   virtual void print(String *str, enum_query_type query_type);
+  enum Type type() const { return INSERT_VALUE_ITEM; }
```
laurynas-biveinis, line 4054: For 5.7, wait until the 5.7.21 upstream merge and rebase, it introduced the same. For 5.6 it would apply as well but due to GCA it will have to be resolved at merge time

sql/sql_insert.cc
```diff
@@ -1071,6 +1071,7 @@ bool mysql_insert(THD *thd,TABLE_LIST *table_list,
       error= 1;
       break;
     }
+
```
laurynas-biveinis: Spurious

storage/tokudb/CMakeLists.txt
```diff
@@ -103,7 +103,6 @@ INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR}/${TOKU_FT_DIR_NAME})
 INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR}/${TOKU_FT_DIR_NAME}/buildheader)
 INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR}/${TOKU_FT_DIR_NAME}/portability)
 
-
```
laurynas-biveinis: Spurious

laurynas-biveinis (changes requested):
- MySQL has HA_READ_BEFORE_WRITE_REMOVAL - is it possible to use that instead of patching server layer? My server layer review is partial, due to this question
- fast_update_blobs.result etc: the testcases are crazy and unworkable with some execution trace being recorded with tens of thousands of lines of SQL, resulting in total diff here of 200K. Please discuss with @georgelorchpercona  and either adjust the generators to insert/update 10 not 10000 rows in a single loop, or if that is required for some otherwise untested code paths, --disable-query-log/--disable-result-log each loop body and add --echo # Inserting billions of trillions of rows to header instead

vlad-lesin:
> MySQL has HA_READ_BEFORE_WRITE_REMOVAL - is it possible to use that instead of patching server layer? My server layer review is partial, due to this question

Good point. But I am not sure it's reasonable because of two things:

1)
Consider this example from handler.h which explains how to use this feature:
```cpp
  Example:
  UPDATE a=1 WHERE pk IN (<keys>)

  mysql_update()
  {
    if (<conditions for starting read removal>)
      start_read_removal()
      -> handler returns true if read removal supported for this table/query

    while(read_record("pk=<key>"))
      -> handler returns fake row with column "pk" set to <key>

      ha_update_row()
      -> handler sends write "a=1" for row with "pk=<key>"

    end_read_removal()
    -> handler returns the number of rows actually written
  }
``` 

The signature of ha_update_row() is the following:
```cpp
int handler::ha_update_row(const uchar *old_data, uchar *new_data);
```

It's supposed that there are two rows, the first row is with old data and the second row is with new data. But "fast update" message contains not only data to update, but also conditions for update. These conditions are passed to mysql_update as a tree, and they are also passed to tokudb engine fast update function which packs them to message.

This is the signature of the function for fast update:
```cpp
int ha_tokudb::fast_update(
    THD* thd,
    List<Item>& update_fields,
    List<Item>& update_values,
    Item* conds) ;
```

We can see "conds" there. And the question is how to pass the conditions to tokudb handler without changing server code?

2) The implementation contains not only "update" optimization, but also "insert", what is not covered with HA_READ_BEFORE_WRITE_REMOVAL case.

The server code changes itself are not very big. The big diff caused by code block shifting with the following reformatting. For exampe, the code in mysql_update is changed like the following.
before:
```cpp
int mysql_update(...)
{
...
  while (!(error=info.read_record(&info)) && !thd->killed)
  {
  ...
  }
...
}
```
after:
```cpp
int mysql_update(...)
{
...
  error= table->file->ha_fast_update(thd, fields, values, conds);
  if (error == 0)
    error= -1; // error < 0 means really no error at all (see below)
  else if (error != ENOTSUP) {
    error= 1;
  }
  else
    while (!(error=info.read_record(&info)) && !thd->killed)
    {
    ...
    }
...
}
```

laurynas-biveinis:
Understood re. HA_READ_BEFORE_WRITE_REMOVAL - please add a short comment somewhere in the feature that "We looked into HA_READ_BEFORE_WRITE_REMOVAL and we cannot use it"

vlad-lesin:
New testing: https://jenkins.percona.com/job/percona-server-5.6-param/2082/

vlad-lesin:
Ready for the next review round.

mysql-test/suite/tokudb/include/setup_fast_update_upsert.inc
```diff
@@ -0,0 +1,2 @@
+set session sql_mode=(select replace(@@sql_mode,'STRICT_TRANS_TABLES',''));
```
laurynas-biveinis, line 7: Previous question still standing: how is this setting up fast update/upserts? Why STRICT_TRANS/ALL_TABLES have to be removed?

mysql-test/suite/tokudb/r/fast_update_binlog_mixed.result
```diff
@@ -1,17 +1,25 @@
...
 show variables like 'server_id';
 Variable_name	Value
 server_id	1
+set session sql_mode=(select replace(@@sql_mode,'STRICT_TRANS_TABLES',''));
+set session sql_mode=(select replace(@@sql_mode,'STRICT_ALL_TABLES',''));
+set session tokudb_enable_fast_update=1;
 show variables like 'binlog_format';
 Variable_name	Value
-binlog_format	MIXED
+binlog_format	STATEMENT
```
laurynas-biveinis: Disagrees with the testcase name

mysql-test/suite/tokudb/t/fast_update_binlog_mixed.test
```diff
@@ -5,13 +5,17 @@ source include/have_tokudb.inc;
 set default_storage_engine='tokudb';
 create table tt (id int primary key, x int);
 
-set session tokudb_disable_slow_upsert=1;
-
 connection master;
+source ../include/setup_fast_update_upsert.inc;
+set session tokudb_enable_fast_update=1;
```
laurynas-biveinis: Perhaps this set session var set should go to that include file above

mysql-test/suite/tokudb/t/fast_update_blobs.py
```diff
@@ -9,18 +9,23 @@ def main():
     print "source include/have_tokudb.inc;"
```
laurynas-biveinis: The generated testfiles are still too large (as diff says +80K added lines) - please discuss with @georgelorchpercona to reduce the tests to as few SQL statements as possible. Even better if the python generators could be completely removed

mysql-test/suite/tokudb/t/fast_update_blobs_fixed_varchar.py
```diff
@@ -9,19 +9,24 @@ def main():
     print "source include/have_tokudb.inc;"
```
laurynas-biveinis: Likewise

mysql-test/suite/tokudb/t/fast_update_blobs_with_varchar.py
```diff
@@ -9,19 +9,24 @@ def main():
     print "source include/have_tokudb.inc;"
```
laurynas-biveinis: Likewise

mysql-test/suite/tokudb/t/fast_update_varchar.py
```diff
@@ -9,16 +9,21 @@ def main():
     print "source include/have_tokudb.inc;"
```
laurynas-biveinis: Likewise

storage/tokudb/ha_tokudb.h
```diff
@@ -1037,15 +1037,28 @@ class ha_tokudb : public handler {
...
-#if TOKU_INCLUDE_UPSERT
 private:
-    int fast_update(THD *thd, List<Item> &update_fields, List<Item> &update_values, Item *conds);
-    bool check_fast_update(THD *thd, List<Item> &update_fields, List<Item> &update_values, Item *conds);
-    int send_update_message(List<Item> &update_fields, List<Item> &update_values, Item *conds, DB_TXN *txn);
-    int upsert(THD *thd, List<Item> &update_fields, List<Item> &update_values);
-    bool check_upsert(THD *thd, List<Item> &update_fields, List<Item> &update_values);
-    int send_upsert_message(THD *thd, List<Item> &update_fields, List<Item> &update_values, DB_TXN *txn);
-#endif
+    int fast_update(THD *thd,
```
laurynas-biveinis: Every single one of these should probably be MY_ATTRIBUTE((warn_unused_result))

laurynas-biveinis (changes requested):
Please check mysql-test/suite/tokudb/r/fast_upsert_bin_pad.result test for spurious non-ASCII characters, why github is showing it as "binary" file

sql/sql_update.cc
```diff
@@ -740,183 +740,190 @@ int mysql_update(THD *thd,
...
-    thd->inc_examined_row_count(1);
-    bool skip_record;
-    if (!select || (!select->skip_record(thd, &skip_record) && !skip_record))
+  error= table->file->ha_fast_update(thd, fields, values, conds);
+  if (error == 0)
+    error= -1; // error < 0 means really no error at all (see below)
+  else if (error != ENOTSUP) {
+    error= 1;
+  }
+  else
```
laurynas-biveinis, line 749: Consider replacing "else" with a goto to jump over the legacy code to avoid its reindentation

mysql-test/suite/tokudb/t/fast_update_blobs_fixed_varchar.py
vlad-lesin: Sorry, I did not understand, "likewise" what? Could you please clarify?

mysql-test/suite/tokudb/t/fast_update_blobs.py
vlad-lesin: The tests check for the different combinations of data types. How it can be reduced? I could, for example, try to use loops inside of test case instead of python test generator. But loops in mtr are not well-designed, we can't iterate through arrays, for example. And, despite, I met some problems when I tried to use nested loops, I am not sure it is currently fixed. I could play with temporary tables or strings to emulate arrays, I could try to use pure SQL to implement enumeration, the but I am not sure the result code will be more readable than the initial python generator code.

vlad-lesin: BTW, random string generator will look weird too on SQL.

vlad-lesin: Apart from this each test has the comment which explains the test is generated, what means we should look at the test generator instead of trying to read the test text itself to understand what exactly is tested. The pattern of each test is the same - insert or change the data both in TokuDB and InnoDB tables, and compare their content, so the result file itself is not important for understanding if the test passed or failed, what is important is the result of tables content comparison. I could, for example, rewrite one of the tests using SQL, and the post here it's text to decide if it's reasonable to do the same with the rest of the tests.

mysql-test/suite/tokudb/t/fast_update_blobs_fixed_varchar.py
laurynas-biveinis: This points to "The generated testfiles are still too large (as diff says +80K added lines) ... "

vlad-lesin: Ok, I got it. The comment is not about the certain line, but about the whole file.

mysql-test/suite/tokudb/include/setup_fast_update_upsert.inc
vlad-lesin: Because it's supposed that for strict mode some additional checks must be provided, what can lead to necessity to read some data from storage, while the general optimization for fast update and upsert is that there are no read operations during insert/update execution.

mysql-test/suite/tokudb/r/fast_update_binlog_mixed.result
vlad-lesin: Those lines are for the slave, not for the master, so the test name corresponds to what the test does. But those checks are, actually, needless, because all necessary checks are performed in 'have_binlog_format*.inc" files. So I have just removed it.

mysql-test/suite/tokudb/t/fast_update_binlog_mixed.test
vlad-lesin: No, because different tests require different sets of these variables.

mysql-test/suite/tokudb/t/fast_update_blobs.py
vlad-lesin: I have rewrote python generators in perl and embedded them into mtr tests.

mysql-test/suite/tokudb/t/fast_update_blobs_fixed_varchar.py
vlad-lesin: See new version of ex-'python generated' tests,

mysql-test/suite/tokudb/t/fast_update_blobs_with_varchar.py
vlad-lesin: Likewise

mysql-test/suite/tokudb/t/fast_update_varchar.py
vlad-lesin: Likewise

storage/tokudb/ha_tokudb.h
vlad-lesin: I think, this is not a good idea to use compiler-specific keywords without strict necessity.

sql/sql_update.cc
vlad-lesin: Using "goto" is not a good style unless it gives an advantage over other solutions. In our case I would prefer just not to do reindentation for "else" code block.

vlad-lesin:
> Please check mysql-test/suite/tokudb/r/fast_upsert_bin_pad.result test for spurious non-ASCII characters, why github is showing it as "binary" file

That test uses 'binary' mysql type, the corresponding field values are right-padded with '\0' to the specified length. 'select' returns values with pad bytes.

mysql-test/suite/tokudb/include/setup_fast_update_upsert.inc
laurynas-biveinis: This needs to be documented in a comment

mysql-test/suite/tokudb/t/fast_update_binlog_mixed.test
laurynas-biveinis: Then that file name might be misleading, because it does not "set up" fast updates and upserts: it just prepares for subsequent setup. But I don't have a suggestion for a better name, maybe just "relax sql modes" or similar

storage/tokudb/ha_tokudb.h
laurynas-biveinis: It will give better diagnostics and will be defined to nothing on unsupporting compilers. No functionality or portability is lost, and the code base already contains many of them.

sql/sql_update.cc
laurynas-biveinis: It will not survive automated clang format checks. Goto would.

vlad-lesin: There are a lot of places even in this function, which will not pass clang-format checks. The previous version was indented so that it would pass the check, but this is not good for our upstream merge process. The both solutions('goto' and not-indented 'else')  are bad from code readability perspective. But using 'goto' in the place where there is no need to use it only to pass clang-format check is quite strange solution from my point of view.

storage/tokudb/ha_tokudb.h
vlad-lesin: Ok, there is no portability issues in our code, what is good. But I try to understand the pattern of this attribute usage in our code. gcc documentation says:

"This is useful for functions where not checking the result is either a security problem or always a bug, such as realloc."

Each function which returns error code should be with this attribute because not checking an error code is potential bug, also functions which return the result of check or test something, what suits to my case, or return some class member value, etc.

If so, then why, for example, there is no such attribute for thd_get_ft_query_extra_word_chars() in b7cb58, or for System_keys_container::is_system_key_without_version() in 0d4ef6, or in tons of another code? What is the difference between this and those cases?

From my point of view this attribute is for quite strict cases. Otherwise we should include it on our codestyle guide and use it for every function(with rare exceptions), which returns any value.

laurynas-biveinis: Basically every function whose return value should be checked should carry it, so I agree re. including in the codestyle guide. Those sample cases should have added it too, but missed it. I plan to do a batch add on e.g. 8.0 port.

vlad-lesin: BTW, there is "nodiscard" attribute in C++17. So the common direction is to include the ability to specify if  returned value can be ignored or not during function declaration into the language standard. Maybe it would be reasonable for us to create the corresponding macros and use it instead of MY_ATTRIBUTE for this purpose?

Something like
```
#if __has_cpp_attribute(nodiscard)
#define NODISCARD [[nodiscard]]
#elif __has_cpp_attribute(gnu::warn_unused_result)
#define NODISCARD [[gnu::warn_unused_result]]
#else
#define NODISCARD
#endif
```

gnu::warn_unused_result is supported both in gcc and clang.

vlad-lesin:
https://jenkins.percona.com/job/percona-server-5.6-param/2098/

storage/tokudb/ha_tokudb.h
laurynas-biveinis: This can work, but I am not sure whether existing uses in our patch should be converted, which would seem to be a refactoring that is slightly hard to justify.

vlad-lesin: Following this logic, if we are going to include using of MY_ATTRIBUTE((warn_unused_result)) in our code style sheet, then refactoring of our existing code is required too.

I think we could use the same rule as we use for indentation. We reindent only the code which is currently fixed for our purposes.

vlad-lesin:
There is one more thing to discuss. Originally the feature was under conditional compilation which I removed. Now I think the conditional compilation could be useful, we could easily remove the feature. It can be used, for example, for step-by-step porting to 8.0.

storage/tokudb/ha_tokudb.h
laurynas-biveinis: I agree

laurynas-biveinis:
Unless both sides of conditional compilation have been tested recently, it's possibly bitrotten and can be just dropped (if e.g. 8.0 work needs it, the effort to add it there could be comparable with the effort to fix it)

vlad-lesin:
Compilation test: https://jenkins.percona.com/job/percona-server-5.6-param/2101/

laurynas-biveinis (approved):

mysql-test/mysql-test-run.pl
```diff
@@ -3827,6 +3827,7 @@ ($)
 {
   my $tinfo= shift;
 
+  $ENV{'MYSQL_CURRENT_TEST_DIR'} = dirname($tinfo->{'path'});
```
laurynas-biveinis, line 3830: Is this actually used?

vlad-lesin: Yes, there are some parts in generated tests where there is no need to substitute any parameters. Therefore there is no need to use perl 'print' to generate these absolutely the same parts for each set of parameters. So I moved these parts in separate *.inc files and just inserted '--source path_to_inc_file'" in generated file. But I need to know full path_to_inc_file because the generated test will be executed in temp dir, I can't use relative path there. I could, for example, just use mysql-test directory and add predefined path inside of this directory. But I don't like this idea, the tests can be moved in another subdirectory, the suite can be renamed etc., that's why I decided to introduce new environment variable. That does not require big changes, just one line, so it's feasible.

PR #2100, opened 2018-01-29
PS260 testcase

dutow:
Without the previous commit fixing it, loading the world database crashed at the 181th city.

### DIFF
--- a/mysql-test/suite/innodb/r/percona_bug_ps260.result
+++ b/mysql-test/suite/innodb/r/percona_bug_ps260.result
@@ -1,0 +1,5 @@
+#
+# Loading the MySQL World sample database
+# PS-260: An optimizer bug caused a crash when loading this SQL
+#
+# Loading data
--- a/mysql-test/suite/innodb/t/percona_bug_ps260.test
+++ b/mysql-test/suite/innodb/t/percona_bug_ps260.test
@@ -1,0 +1,226 @@
+--source include/have_innodb.inc
+
+--echo #
+--echo # Loading the MySQL World sample database
+--echo # PS-260: An optimizer bug caused a crash when loading this SQL
+--echo #
+
+--disable_query_log
+--disable_result_log
+--echo # Loading data
+drop database if exists world;
+create database world;
+use world;
+CREATE TABLE `city` (
+  `ID` INT(11) NOT NULL AUTO_INCREMENT,
+  `Name` CHAR(35) NOT NULL DEFAULT '',
+  `CountryCode` CHAR(3) NOT NULL DEFAULT '',
+  `District` CHAR(20) NOT NULL DEFAULT '',
+  `Population` INT(11) NOT NULL DEFAULT '0',
+  PRIMARY KEY (`ID`),
+  KEY `CountryCode` (`CountryCode`)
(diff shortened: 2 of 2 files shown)


mysql-test/std_data/percona_bug_ps260_world200.sql
```diff
@@ -0,0 +1,246 @@
...
+INSERT INTO `city` VALUES (191,'Hamilton','BMU','Hamilton',1200);
+INSERT INTO `city` VALUES (192,'Thimphu','BTN','Thimphu',22000);
+INSERT INTO `city` VALUES (193,'Santa Cruz de la Sierra','BOL','Santa Cruz',935361);
+INSERT INTO `city` VALUES (194,'La Paz','BOL','La Paz',758141);
+INSERT INTO `city` VALUES (195,'El Alto','BOL','La Paz',534466);
+INSERT INTO `city` VALUES (196,'Cochabamba','BOL','Cochabamba',482800);
+INSERT INTO `city` VALUES (197,'Oruro','BOL','Oruro',223553);
+INSERT INTO `city` VALUES (198,'Sucre','BOL','Chuquisaca',178426);
+INSERT INTO `city` VALUES (199,'Potos�','BOL','Potos�',140642);
+INSERT INTO `city` VALUES (200,'Tarija','BOL','Tarija',125255);
```
laurynas-biveinis: Is this the reduced set of INSERTs?

mysql-test/t/percona_bug_ps260.test
```diff
@@ -0,0 +1,17 @@
...
+--echo # Loading the MySQL World sample database
+--echo # PS-260: An optimizer bug caused a crash when loading this SQL
+--echo #
+
+--disable_query_log
+--disable_result_log
+--echo # Loading data
+drop database if exists world;
+create database world;
+--exec $MYSQL  world 2>&1 < "./std_data/percona_bug_ps260_world200.sql"
```
laurynas-biveinis: Inline the .sql file here instead of loading it

laurynas-biveinis (changes requested):
In general, for upstream bugs, we use upstream MTR file names (check Confluence), but here upstream-ness is not proved yet, so OK as-is

mysql-test/std_data/percona_bug_ps260_world200.sql
dutow: It is. It only contains 200 cities, we need them up to 181 for the crash.

mysql-test/t/percona_bug_ps260.test
dutow: Updated.

mysql-test/t/percona_bug_ps260.test
```diff
@@ -0,0 +1,226 @@
+--source include/have_innodb.inc
```
laurynas-biveinis: Sorry for missing this earlier: Since the crash is inside of InnoDB, it should go to the InnoDB suite. No need for Jenkins re-test and another review round, just move, test locally, and merge.

laurynas-biveinis (approved):

dutow:
https://jenkins.percona.com/job/percona-server-5.6-param/2074/

PR #2103, opened 2018-01-29
PS-3788: partially dynamic @version variable using support of @version_suffix

inikep:
…nd @version_comment variables.

@version_comment: convert this variable from a global read-only to a global read-write variable.

@version: introduce new global read-write string variable @version_suffix. If @version_suffix is set to a string, for example, "-foo", then @version variable is truncated after the version number and @version_suffix is appended, resulting in e.g. "5.7.13-foo". Note that an empty string provides a way to leave only the version number in @version.

### DIFF
--- a/mysql-test/suite/sys_vars/t/version_comment_basic.test
+++ b/mysql-test/suite/sys_vars/t/version_comment_basic.test
@@ -35,9 +35,26 @@ SELECT COUNT(@@GLOBAL.version_comment);
 #   Check if Value can set                                         #
 ####################################################################
 
---error ER_INCORRECT_GLOBAL_LOCAL_VAR
+--error ER_WRONG_TYPE_FOR_VAR
 SET @@GLOBAL.version_comment=1;
---echo Expected error 'Read only variable'
+--echo Expected error 'Incorrect argument type to variable'
+
+# save the start value
+SET @saved_version_comment = @@global.version_comment;
+
+SET GLOBAL version_comment = DEFAULT;
+
+SET GLOBAL version_comment = '';
+SELECT @@global.version_comment;
+
+SET GLOBAL version_comment = 'my_comment';
+SELECT @@global.version_comment;
+
+SET GLOBAL version_comment = 'my_comment2';
+SELECT @@global.version_comment;
+
+# restore the start value
+SET GLOBAL version_comment = @saved_version_comment;
 
 SELECT COUNT(@@GLOBAL.version_comment);
 --echo 1 Expected
--- a/mysql-test/suite/sys_vars/t/version_suffix_basic.test
+++ b/mysql-test/suite/sys_vars/t/version_suffix_basic.test
@@ -1,0 +1,96 @@
+####################################################################
+#   Displaying default value                                       #
+####################################################################
+SELECT COUNT(@@GLOBAL.version_suffix);
+--echo 1 Expected
+
+
+####################################################################
+#   Check if Value can set and if @@version is set according to    #
+#   @@version_suffix                                               #
+####################################################################
+
+--error ER_WRONG_TYPE_FOR_VAR
+SET @@GLOBAL.version_suffix=1;
+--echo Expected error 'Incorrect argument type to variable'
+
+# save the start value
+SET @saved_version_suffix = @@global.version_suffix;
+
+SET GLOBAL version_suffix = DEFAULT;
+
+SET GLOBAL version_suffix = '';
+SET @version_base = @@global.version;
+SELECT @@global.version_suffix;
+
+SET GLOBAL version_suffix = '-my_version';
+SELECT @@global.version_suffix;
+--let $assert_text= Strings should be equal
+--let $assert_cond= STRCMP(@@global.version, CONCAT(@version_base, @@global.version_suffix)) = 0
+--source include/assert.inc
+
+SET GLOBAL version_suffix = '-my_version2';
+SELECT @@global.version_suffix;
+--let $assert_text= Strings should be equal
+--let $assert_cond= STRCMP(@@global.version, CONCAT(@version_base, @@global.version_suffix)) = 0
+--source include/assert.inc
+
+# restore the start value
+SET GLOBAL version_suffix = @saved_version_suffix;
+
+SELECT COUNT(@@GLOBAL.version_suffix);
+--echo 1 Expected
+
+
+
+
+#################################################################
+# Check if the value in GLOBAL Table matches value in variable  #
+#################################################################
+
+SELECT @@GLOBAL.version_suffix = VARIABLE_VALUE
+FROM INFORMATION_SCHEMA.GLOBAL_VARIABLES
+WHERE VARIABLE_NAME='version_suffix';
+--echo 1 Expected
+
+SELECT COUNT(@@GLOBAL.version_suffix);
+--echo 1 Expected
+
+SELECT COUNT(VARIABLE_VALUE)
+FROM INFORMATION_SCHEMA.GLOBAL_VARIABLES 
+WHERE VARIABLE_NAME='version_suffix';
+--echo 1 Expected
+
+
+
+################################################################################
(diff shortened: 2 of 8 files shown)


sql/mysqld.cc
```diff
@@ -9598,6 +9599,7 @@ static void set_server_version(void)
 #endif
   if (opt_log || opt_slow_log || opt_bin_log)
     strmov(end, "-log");                        // This may slow down system
+  strcpy(server_version_suffix, server_version + strlen(MYSQL_SERVER_VERSION));
```
laurynas-biveinis: Check for overflow

sql/mysqld.h
```diff
@@ -609,6 +609,7 @@ extern const char *mysql_real_data_home_ptr;
 extern ulong thread_handling;
 extern MYSQL_PLUGIN_IMPORT char  *mysql_data_home;
 extern "C" MYSQL_PLUGIN_IMPORT char server_version[SERVER_VERSION_LENGTH];
+extern "C" MYSQL_PLUGIN_IMPORT char server_version_suffix[SERVER_VERSION_LENGTH];
```
laurynas-biveinis, line 612: Why it has to be extern "C" MYSQL_PLUGIN_IMPORT?

laurynas-biveinis (changes requested):
- This looks like correct design. I'd also consider making "version" dynamic internally only, and proving an check hook for the sys var that blocks all the actual user attempts to modify it with "invalid value", so that only version_suffix has effects. This is to prevent users from changing the version string completely, which includes version number, which then goes out of sync on upgrades.
- Please move the LP blueprint to JIRA if not already done so, and reference it in the commit message header

sql/sys_vars.h
```diff
@@ -550,6 +550,53 @@ class Sys_var_charptr: public sys_var
...
+    char **version_ptr= reinterpret_cast<char**> (
+                               Sys_var_charptr::global_value_ptr(thd, base));
+    if (version_ptr == NULL)
+      return NULL;
+
+    sys_var *suffix_var= find_sys_var(thd, STRING_WITH_LEN("version_suffix"));
+    if (suffix_var == NULL)
+      return reinterpret_cast<uchar*> (version_ptr);
+
+    char** suffix_ptr= reinterpret_cast<char**> (suffix_var->value_ptr(thd,
```
percona-ysorokin, line 579: Can there be a race condition if another thread modifies `version_suffix` at the same time?

sql/mysqld.cc
inikep: `server_version` has `SERVER_VERSION_LENGTH` bytes (and it's not checked for overflow) 
`server_version_suffix` is just a substring of `server_version` so if `server_version` is correct then `server_version_suffix` will not be overflowed
Should I change it anyway?

sql/sys_vars.cc
```diff
@@ -3463,15 +3463,21 @@ static Sys_var_mybool Sys_timed_mutexes(
...
+static Sys_var_version Sys_version(
        "version", "Server version",
        READ_ONLY GLOBAL_VAR(server_version_ptr), NO_CMD_LINE,
-       IN_SYSTEM_CHARSET, DEFAULT(server_version));
+       IN_SYSTEM_CHARSET, DEFAULT(MYSQL_SERVER_VERSION));
+
+static char *server_version_suffix_ptr;
+static Sys_var_charptr Sys_version_suffix(
+       "version_suffix", "version_suffix",
+       GLOBAL_VAR(server_version_suffix_ptr), NO_CMD_LINE,
```
percona-ysorokin, line 3474: What's the size limit of `version_suffix`?
What happens if its more than `SERVER_VERSION_LENGTH`?

sql/mysqld.h
inikep: Without it there is
```
CMakeFiles/handlersocket.dir/handlersocket/handlersocket.cpp.o:(.bss+0x0): multiple definition of `server_version_suffix'
```

percona-ysorokin (changes requested):

percona-ysorokin:
A new MTR test case is needed (at least in `sys_var` test suite) as we introduce a new system variable.

inikep:
@laurynas-biveinis 
1. `Sys_var_version Sys_version` is still `READ_ONLY` for users.
2. Jira ticked created at https://jira.percona.com/browse/PS-3788

sql/sys_vars.cc
inikep: `version_suffix` has no limits but now `version` has limit of `SERVER_VERSION_LENGTH` and only first `SERVER_VERSION_LENGTH - 1 - strlen(MYSQL_SERVER_VERSION)` bytes of `version_suffix` will be concatenated what is checked in:
```
    if (suffix_ptr_len + version_ptr_len + 1 > SERVER_VERSION_LENGTH)
      suffix_ptr_len = SERVER_VERSION_LENGTH - version_ptr_len - 1;
```

sql/sys_vars.h
inikep: This commit does not change anything in update of variables. They are modified using
`sys_var::update(THD *thd, set_var *var)`
which uses
```
    AutoWLock lock1(&PLock_global_system_variables);
    AutoWLock lock2(guard);
```

mysql-test/suite/sys_vars/t/version_comment_basic.test
```diff
@@ -35,9 +35,24 @@ SELECT COUNT(@@GLOBAL.version_comment);
...
+
+SET @@GLOBAL.version_comment='my_comment';
+SELECT @@version_comment;
+
+SET GLOBAL version_comment='my_comment2';
+SELECT @@version_comment;
+
+SET @@GLOBAL.version_comment=DEFAULT;
+SELECT STRCMP(@@version_comment, @version_comment_backup);
+--echo 0 Expected
```
laurynas-biveinis: Just restore from backup (then MTR --mysqld=--version-comment=notdefaultvalue works better)

mysql-test/suite/sys_vars/t/version_suffix_basic.test
```diff
@@ -0,0 +1,125 @@
...
+################## mysql-test\t\version_suffix_basic.test ####################
+#                                                                             #
+# Variable Name: version_suffix                                              #
+# Scope: Global                                                               #
+# Access Type: Static                                                         #
+# Data Type: String                                                           #
+#                                                                             #
+#                                                                             #
+# Creation Date: 2008-02-07                                                   #
+# Author : Sharique Abdullah                                                      #
```
laurynas-biveinis: Please remove all the boilerplate comments and also those BS_STVARS_054_01 codes that have no meaning for us. A ~20 line testcase is fine here

laurynas-biveinis (changes requested):
In commit message, descriptions etc: system variables are referenced by a double @@ not single @ (on github maybe try to avoid pinging the usernames inadvertently too)

inikep:
Changes:
1. Improved MTR tests
2. descriptions referenced by a double @@ not single @ 
3. Added in `set_server_version(void)`: 
```
DBUG_ASSERT(end < server_version + SERVER_VERSION_LENGTH);
```

laurynas-biveinis (approved):

sql/mysqld.cc
```diff
@@ -9597,7 +9598,10 @@ static void set_server_version(void)
     end= strmov(end, "-debug");
 #endif
   if (opt_log || opt_slow_log || opt_bin_log)
-    strmov(end, "-log");                        // This may slow down system
+    end= strmov(end, "-log");                        // This may slow down system
+
+  DBUG_ASSERT(end <= server_version + SERVER_VERSION_LENGTH);
```
dutow: Shouldn't this assertion use just a <?

strmov/strxmov return a pointer to the terminating null character, so if it equals, it's already a one byte overflow.

inikep: Fixed, thanks

PR #2104, opened 2018-01-29
PS-3778: removing unneccessary broken submodule

dutow:
This was accidentally added in 0620870190f546bc5531ee2e2c697d61fa758042


laurynas-biveinis (approved):
s/brokes/broken ?

PRs like this do not need wait for a review and can be self-approved.

PR #2106, opened 2018-01-29
PS-259 : Issue during valgrind-yassl binary build

george-lorch:
PS-1121 : LP #1729321: Tokudb does not build when WITH_VALGRIND=1
- PerconaFT recognizes USE_VALGRIND and not WITH_VALGRIND. Added minor check in
  TokuDB CMakeLists.txt that setc USE_VALGRIND if WITH_VALGRIND is set.


george-lorch:
Regular jenkins to prove it does not break standard builds https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2071/

laurynas-biveinis (approved):

PR #2108, opened 2018-01-29
TDB-48 : mtr Valgrind errors on LZMA

george-lorch:
- Added mtr Valgrind suppression for PerconaFT LZMA lz_encoder_prepare

### DIFF
--- a/mysql-test/valgrind.supp
+++ b/mysql-test/valgrind.supp
@@ -915,6 +915,13 @@
    fun:tls_get_addr_tail
 }
 
+{
+   lzma encoder prepare bug
+   Memcheck:Cond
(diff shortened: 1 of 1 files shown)


george-lorch:
Here is a 5.7 jenkins valgrind param run to prove that the added suppression does not break anything and seems to work https://jenkins.percona.com/view/5.7/job/mysql-5.7-valgrind-param/90/ Skipped 5.6 as the 5.6 valgrind param is extremely out of date and hasn't been run in 3 years 9 months.

laurynas-biveinis (approved):
LGTM assuming the suppression has been tested on 5.6 locally

PR #2109, opened 2018-01-29
Ps 5.7 tdb 48

george-lorch:


george-lorch:
Here is a 5.7 jenkins valgrind param run to prove that the added suppression does not break anything and seems to work https://jenkins.percona.com/view/5.7/job/mysql-5.7-valgrind-param/90/

laurynas-biveinis (approved):

PR #2111, opened 2018-01-30
PS-296: mysqld crashing with subselects and long keys

dutow:
The heap-dynamic-rows patch limited the maximum key size of strings to 1000.
This worked fine in earlier versions, where the maximum key length was below 1000.
In 5.7, this was increased to 3072, which resulted in keys with 999 or more characters crashing in some operations.
The latest dynamic rows patch doesn't contain this limitation anymore, looks like it can be safely removed.

Note that this change doesn't fix the issue completely:
varchar/varbinary keys at least 3071 character long will still crash - in practice, this means a maximum length of 3071 or 3072 chararcters, as that's the current maximum.
This same bug is also present in mysql server.

### DIFF
--- a/include/heap.h
+++ b/include/heap.h
@@ -42,7 +42,7 @@ extern "C" {
 #define HP_MAX_KEY                  MAX_INDEXES         /* Max allowed keys */
 #endif
 
-#define HP_MAX_KEY_LENGTH           1000            /* Max length in bytes */
+#define HP_MAX_KEY_LENGTH           MAX_KEY_LENGTH      /* Max length in bytes */
 
 /* defines used by heap-funktions */
 
--- a/mysql-test/t/percona_heap_bug_ps296.test
+++ b/mysql-test/t/percona_heap_bug_ps296.test
@@ -1,0 +1,59 @@
+--echo PS-296: inconsistent maximum key lenghts caused a crash in the query optimizer
+--source include/have_innodb.inc
+
+# Test that we can use long strings in memory tables
+CREATE TABLE big_ok (id VARBINARY(3072) NOT NULL, PRIMARY KEY (id)) ENGINE=MEMORY;
+CREATE TABLE big_ok2 (id VARBINARY(1072) NOT NULL, id2 VARBINARY(2000), PRIMARY KEY (id, id2)) ENGINE=MEMORY;
+--error ER_TOO_LONG_KEY
+CREATE TABLE big_fails (id VARBINARY(3073) NOT NULL, PRIMARY KEY (id)) ENGINE=MEMORY;
+--error ER_TOO_LONG_KEY
+CREATE TABLE big_fails (id VARBINARY(1073) NOT NULL, id2 VARBINARY(2000), PRIMARY KEY (id, id2)) ENGINE=MEMORY;
+
+SET @@session.max_sort_length = 4000;
+INSERT INTO big_ok VALUES (CONCAT(REPEAT('a', 3071), 'c'));
+INSERT INTO big_ok VALUES (CONCAT(REPEAT('a', 3071), 'b'));
+INSERT INTO big_ok VALUES (CONCAT(REPEAT('a', 3071), 'a'));
+
+SELECT RIGHT(id, 5) FROM big_ok FORCE INDEX (PRIMARY) ORDER BY id ASC LIMIT 10;
+SELECT RIGHT(id, 5) FROM big_ok IGNORE INDEX (PRIMARY) ORDER BY id ASC LIMIT 10;
+
+INSERT INTO big_ok2 VALUES (REPEAT('a', 1072), CONCAT(REPEAT('a', 1999), 'c'));
+INSERT INTO big_ok2 VALUES (REPEAT('a', 1072), CONCAT(REPEAT('a', 1999), 'b'));
+INSERT INTO big_ok2 VALUES (REPEAT('a', 1072), CONCAT(REPEAT('a', 1999), 'a'));
+
+SELECT RIGHT(id2, 5) FROM big_ok2 FORCE INDEX (PRIMARY) ORDER BY id ASC, id2 ASC LIMIT 10;
+SELECT RIGHT(id2, 5) FROM big_ok2 IGNORE INDEX (PRIMARY) ORDER BY id ASC, id2 ASC LIMIT 10;
+
+# Test that materialized queries do not crash
+
+CREATE TABLE a (id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY);
+INSERT INTO a () VALUES ();
+delimiter |;
+CREATE PROCEDURE dowhile()
+BEGIN
+  DECLARE v1 INT DEFAULT 1;
+  WHILE v1 < 15 DO 
+    INSERT INTO a SELECT NULL FROM a;
+    SET v1 = v1 + 1;
+  END WHILE;
+END;
+|
+delimiter ;|
+CREATE TABLE big1 (id VARBINARY(3070) NOT NULL, PRIMARY KEY (id)) ENGINE=InnoDB;
+CREATE TABLE big2 (id VARBINARY(3070) NOT NULL, PRIMARY KEY (id)) ENGINE=InnoDB;
+--disable_result_log
+CALL dowhile();
+INSERT INTO big1 SELECT * FROM a;
+INSERT INTO big2 SELECT * FROM a;
+--enable_result_log
+
+SET session optimizer_switch='duplicateweedout=off,materialization=on';
+EXPLAIN SELECT * FROM big1 WHERE id IN (SELECT id FROM big2 IGNORE INDEX (PRIMARY)) ORDER BY id ASC LIMIT 10;
+SELECT * FROM big1 WHERE id IN (SELECT id FROM big2 IGNORE INDEX (PRIMARY)) ORDER BY id ASC LIMIT 10;
+
+DROP TABLE big_ok;
+DROP TABLE big_ok2;
+DROP TABLE big1;
+DROP TABLE big2;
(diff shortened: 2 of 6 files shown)


mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,36 @@
+
```
laurynas-biveinis, line 3: Add header block --echo # / --echo # PS-296: blah blah / --echo #

mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,36 @@
...
+
+--disable_query_log
```
laurynas-biveinis: Why?

mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,36 @@
...
+--disable_result_log
+create table a (id int unsigned not null auto_increment primary key);
+delimiter |;
```
laurynas-biveinis: One row too soon?

mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,36 @@
...
+insert into a () values ();
+CREATE PROCEDURE dowhile()
```
laurynas-biveinis: Inconsistent SQL keyword case

mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,36 @@
...
+BEGIN
+ DECLARE v1 INT DEFAULT 1;
+ while v1 < 15 do 
+insert into a select null from a;
```
laurynas-biveinis: Indentation

mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,36 @@
...
+CREATE TABLE `big2` (`id` varbinary(999) NOT NULL, PRIMARY KEY (`id`)) ENGINE=InnoDB;
+insert into big1 select * from a;
+insert into big2 select * from a;
+set session optimizer_switch='duplicateweedout=off,materialization=on';
+--enable_result_log
+--enable_query_log
+
+select * from big1 where id in (select id from big2 ignore index (primary)) order by id asc limit 10;
+
+--disable_query_log
```
laurynas-biveinis: Why?

mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,36 @@
...
+call dowhile();
+CREATE TABLE `big1` (`id` varbinary(999) NOT NULL,PRIMARY KEY (`id`)) ENGINE=InnoDB;
+CREATE TABLE `big2` (`id` varbinary(999) NOT NULL, PRIMARY KEY (`id`)) ENGINE=InnoDB;
+insert into big1 select * from a;
+insert into big2 select * from a;
+set session optimizer_switch='duplicateweedout=off,materialization=on';
+--enable_result_log
+--enable_query_log
+
+select * from big1 where id in (select id from big2 ignore index (primary)) order by id asc limit 10;
```
laurynas-biveinis: add explain select for the same too

laurynas-biveinis: so this is due to a temp MEMORY table that stores VARBINARY. How do we end up getting one? Dynamic varchar patch did not enable them for the query optimizer use. Does the upstream use MEMORY table too?

laurynas-biveinis (changes requested):

mysql-test/t/percona_heap_bug_ps296.test
dutow: Yes, it's the same. And I don't see anything in the query code that would disable the memory engine for string joins - the heap engine's maximum key length is used both in upstream and our code for this join, even for longer strings.

mysql-test/t/percona_heap_bug_ps296.test
dutow: I thought that it would be good if the result file only focused on the statements which are important for the crash.

mysql-test/t/percona_heap_bug_ps296.test
dutow: I'll remove both this and the later disabling instructions - but keeping the result log disabled is probably a good idea here, because that stored procedure inserts a few thousands records.

mysql-test/t/percona_heap_bug_ps296.test
dutow: ops, I deleted this comment, as it wasn't true - local changes resulted in some incorrect results

mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,31 @@
+-- echo PS-296: inconsistent maximum key lenghts caused a crash in the query optimizer
+
+--disable_result_log
```
percona-ysorokin: Add `--source include/have_innodb.inc`

mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,31 @@
...
+insert into a () values ();
+delimiter |;
+create procedure dowhile()
+begin
+  declare v1 int default 1;
+  while v1 < 15 do 
+  insert into a select null from a;
+  set v1 = v1 + 1;
+  end while;
+  end;
```
percona-ysorokin: indentation

mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,31 @@
...
+
+--disable_result_log
+create table a (id int unsigned not null auto_increment primary key);
+insert into a () values ();
+delimiter |;
+create procedure dowhile()
+begin
+  declare v1 int default 1;
+  while v1 < 15 do 
+  insert into a select null from a;
```
percona-ysorokin: indentation for the inner `while` block

mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,31 @@
+-- echo PS-296: inconsistent maximum key lenghts caused a crash in the query optimizer
+
+--disable_result_log
+create table a (id int unsigned not null auto_increment primary key);
```
percona-ysorokin: Please, also consider uppercase for SQL keywords - this will make the testcase more readable.

percona-ysorokin (changes requested):

mysql-test/t/percona_heap_bug_ps296.test
dutow: Done

mysql-test/t/percona_heap_bug_ps296.test
dutow: Done

mysql-test/t/percona_heap_bug_ps296.test
dutow: Done

mysql-test/t/percona_heap_bug_ps296.test
dutow: Done

dutow:
PR updated.

mysql-test/t/percona_heap_bug_ps296.test
```diff
@@ -0,0 +1,32 @@
+-- echo PS-296: inconsistent maximum key lenghts caused a crash in the query optimizer
+--source include/have_innodb.inc
+
+--disable_result_log
```
laurynas-biveinis: disable result log only before the loop calls

include/heap.h
```diff
@@ -42,7 +42,7 @@ extern "C" {
 #define HP_MAX_KEY                  MAX_INDEXES         /* Max allowed keys */
 #endif
 
-#define HP_MAX_KEY_LENGTH           1000            /* Max length in bytes */
+#define HP_MAX_KEY_LENGTH           MAX_KEY_LENGTH      /* Max length in bytes */
```
laurynas-biveinis, line 45: I'm sorry for not commenting this earlier, but this is a functionality change for MEMORY SE, thus needs at least the most basic test: CREATE TABLE w/ index lengths > 1000 < new maximum, check insert, reads using that index, other reads. Check that CREATE TABLE w/ index lengths > newmax fails OK. Check that composite key with key parts > 1000 < newmax, sum < newmax work, sum > newmax fail.

laurynas-biveinis (changes requested):

include/heap.h
dutow: Uh, we have another issue there, and I was able to reproduce this in upstream, both 5.7 and 8.0:

varchar and varbinary ordering fails in memory tables if the key is longer than 1022 characters. (as the length adds 2 more bytes, as would be my guess)

As this is present in the upstream too, and it's a different bug than the crash, shouldn't we handle it as a different issue?

dutow: I figured this out, and added a test (with a single key for now). The issue is that the max_sort_length variable defaults to 1024, while the max key length is 3072.

For now, I changed the max_sort_length value for the test sessions, but maybe it would be a good idea to change its default to be 3072 too - otherwise a query's result will depend on the sorting algorithm used.

laurynas-biveinis (approved):

percona-ysorokin (approved):
LGTM

dutow:
Made two small adjustments:
* rerecorded a test result - it had warnings about too long keys, which now aren't there (these warnings were ignored by the test)
* forgot to drop a table in the test for the bug

jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1647/

laurynas-biveinis (approved):

PR #2114, opened 2018-01-30
[PS-2046] Update package description

EvgeniyPatlan:


EvgeniyPatlan:
self approved as the same PR is already merged into 5.6

PR #2120, opened 2018-01-30
[PS-3784] ps_mysqld_helper.sh script errors

EvgeniyPatlan:


EvgeniyPatlan:
self approved as it was already approved

PR #2123, opened 2018-01-31
BLD-933: CMake option to force build and fail it if the library is not

robgolebiowski:
found

Adds a new cmake variable WITH_KEYRING_VAULT. Allowed values are:
FORCE,ON,1,OFF,0

When cmake is run with -DWITH_KEYRING_VAULT=FORCE and compilation of
keyring_vault plugin is not possible, cmake aborts. Setting
WITH_KEYRING_VAULT to FORCE and providing CURL library which depends on
RTMP library will cause keyring vault unitest to not be build, but
overall cmake configuration will not be aborted. A warning message will
be generated.

When cmake is run with -DWITH_KEYRING_VAULT=ON and compilation of
keyring_vault plugin is not possible, a warning message is generated,
but the cmake configuration is not aborted.

### DIFF
--- a/plugin/keyring_vault/CMakeLists.txt
+++ b/plugin/keyring_vault/CMakeLists.txt
@@ -1,29 +1,27 @@
+IF (NOT DEFINED WITH_KEYRING_VAULT)
+  SET (WITH_KEYRING_VAULT 1)
+ENDIF()
+IF (NOT WITH_KEYRING_VAULT)
+  MESSAGE (STATUS "Not building keyring_vault plugin")
+  RETURN()
+ENDIF()
+
 INCLUDE(CheckFunctionExists)
 INCLUDE(curl)
 
 SET(CMAKE_EXTRA_INCLUDE_FILES string.h)
 SET(CMAKE_EXTRA_INCLUDE_FILES)
 
-MACRO (CHECK_IF_LIB_FOUND lib_name project_name)
+MACRO (CHECK_IF_LIB_FOUND lib_name project_name status_mode)
   SET (lib_found_variable_name "${lib_name}_FOUND")
   IF (NOT DEFINED ${lib_found_variable_name} OR NOT ${${lib_found_variable_name}})
-    message("Not building ${project_name}, could not find library: ${lib_name}")
+    message(${status_mode} "Not building ${project_name}, could not find library: ${lib_name}")
     RETURN()
   ENDIF()
 ENDMACRO()
 
-MACRO (CHECK_IF_CURL_DEPENDS_ON_RTMP project_name)
-  EXECUTE_PROCESS(COMMAND ldd ${CURL_LIBRARY}
-                  COMMAND grep rtmp
-                  OUTPUT_VARIABLE CURL_DEPENDS_ON_RTMP)
-  IF (NOT CURL_DEPENDS_ON_RTMP STREQUAL "")
-    message("Not building ${project_name}. The supplied CURL library depends on rtmp library.
-Please provide CURL library that does not depend on rtmp library to build keyring_vault unittests.")
-    RETURN()
-  ENDIF()
-ENDMACRO()
-
-CHECK_IF_LIB_FOUND(CURL "keyring_vault")
+CHECK_IF_LIB_FOUND(CURL "keyring_vault" FATAL_ERROR)
+MESSAGE (STATUS "Building keyring_vault plugin")
 
 INCLUDE_DIRECTORIES(${BOOST_PATCHES_DIR})
 INCLUDE_DIRECTORIES(SYSTEM ${BOOST_INCLUDE_DIR} ${CURL_INCLUDE_DIRS})
--- a/unittest/gunit/keyring_vault/CMakeLists.txt
+++ b/unittest/gunit/keyring_vault/CMakeLists.txt
@@ -1,11 +1,28 @@
 INCLUDE(curl)
 
-IF(NOT GMOCK_FOUND)
+IF (NOT DEFINED WITH_KEYRING_VAULT)
+  SET (WITH_KEYRING_VAULT 1)
+ENDIF()
+IF(NOT GMOCK_FOUND OR NOT WITH_KEYRING_VAULT)
+    MESSAGE (STATUS "Not building keyring_vault unit tests")
     RETURN()
 ENDIF()
 
-CHECK_IF_LIB_FOUND(CURL "keyring_vault unit tests")
+MACRO (CHECK_IF_CURL_DEPENDS_ON_RTMP project_name)
+  EXECUTE_PROCESS(COMMAND ldd ${CURL_LIBRARY}
+                  COMMAND grep rtmp
+                  OUTPUT_VARIABLE CURL_DEPENDS_ON_RTMP)
+  IF (NOT CURL_DEPENDS_ON_RTMP STREQUAL "")
+    message(WARNING "Not building ${project_name}. The supplied CURL library depends on rtmp library.
+Please provide CURL library that does not depend on rtmp library to build keyring_vault unittests.")
+    RETURN()
+  ENDIF()
+ENDMACRO()
+
+CHECK_IF_LIB_FOUND(CURL "keyring_vault unit tests" FATAL_ERROR)
 CHECK_IF_CURL_DEPENDS_ON_RTMP("keyring_vault unit tests")
+
+message(STATUS "Building keyring_vault unittests")
 
 INCLUDE_DIRECTORIES(
         SYSTEM ${GMOCK_INCLUDE_DIRS}


robgolebiowski:
JR with -DWITH_KEYRING_VAULT=FORCE : https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1556/
JR with -DWITH_KEYRING_VAULT=ON : https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1557/
JR without -DWITH_KEYRING_VAULT=OFF : https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1558/

plugin/keyring_vault/CMakeLists.txt
```diff
@@ -1,3 +1,14 @@
+IF (NOT WITH_KEYRING_VAULT)
+  MESSAGE (STATUS "Not building with keyring_vault plugin, you must specify -DWITH_KEYRING_VAULT=1|ON|FORCE to cmake in order to build with keyring_vault.")
```
laurynas-biveinis: The message suggests that the plugin will not be attempted to be built if the option is not specified?

plugin/keyring_vault/CMakeLists.txt
```diff
@@ -7,23 +18,13 @@ SET(CMAKE_EXTRA_INCLUDE_FILES)
...
-ENDMACRO()
-
-MACRO (CHECK_IF_CURL_DEPENDS_ON_RTMP project_name)
-  EXECUTE_PROCESS(COMMAND ldd ${CURL_LIBRARY}
-                  COMMAND grep rtmp
-                  OUTPUT_VARIABLE CURL_DEPENDS_ON_RTMP)
-  IF (NOT CURL_DEPENDS_ON_RTMP STREQUAL "")
-    message("Not building ${project_name}. The supplied CURL library depends on rtmp library.
-Please provide CURL library that does not depend on rtmp library to build keyring_vault unittests.")
+    message(${KEYRING_VAULT_STATUS_MODE} "Not building ${project_name}, could not find library: ${lib_name}")
```
dutow: This macro has a generic name (`CHECK_IF_LIB_FOUND`), takes a `project_name` argument (called only with `"keyring_vault"`), but the message uses the fixed `KEYRING_VAULT_STATUS_MODE` variable. 

Isn't this a bit inconsistent/confusing? Especially since macros are available in the parents cmakelists too after their definition.

plugin/keyring_vault/CMakeLists.txt
robgolebiowski: I did it taking RocksDB as the model and I got suggested that MyRocksDB is not build by default, there is this comment:  "# by default MyRocks is not built", but on the closer look it turns out it actually is. So keyring_vault should be built by default ? Should I also include variables like WITHOUT_KEYRING_VAULT ?

plugin/keyring_vault/CMakeLists.txt
robgolebiowski: Done, made status a parameter.

plugin/keyring_vault/CMakeLists.txt
laurynas-biveinis: Yes, I believe it should be attempted to be built by default. Not sure re. WITHOUT_ variable, I wouldn't miss it, but what about others? @percona-ysorokin

percona-ysorokin: Guys, I am strictly against any kind of implicit actions in the build process and silently skipping keyring_vault plugin, keyring_vault unit tests or both is one of them. From my own experience during the latest upstream merge, it turned out that everything was "OK" with 'keyring_vault' plugin only because it wasn't built at all on 50% of Jenkins nodes - it was silently skipped there. However, there were actual code errors.

So, my suggestion is to have only 2 options `-DWITH_KEYRING_VAULT=ON | OFF`.
When `OFF` - the plugin and its unit tests are never built.
When 'ON' - will attempt to build both the plugin and its unit tests and will generate an error and stop if some dependencies cannot be resolved or there is an actual compilation/linking error.

I also suggest the default to be `ON`.

unittest/gunit/keyring_vault/CMakeLists.txt
```diff
@@ -1,12 +1,25 @@
 INCLUDE(curl)
 
-IF(NOT GMOCK_FOUND)
+IF(NOT GMOCK_FOUND OR NOT WITH_KEYRING_VAULT)
     RETURN()
 ENDIF()
 
-CHECK_IF_LIB_FOUND(CURL "keyring_vault unit tests")
+MACRO (CHECK_IF_CURL_DEPENDS_ON_RTMP project_name)
+  EXECUTE_PROCESS(COMMAND ldd ${CURL_LIBRARY}
```
percona-ysorokin, line 12: BTW, @robgolebiowski 
I saw this message `The supplied CURL library depends on rtmp library` on Ubuntu Artful with one of the most recent `curl`. The plugin itself was built there but not the unit tests. May be this check is not accurate for Artful.

dutow: Curl on artful depends on the rtmp library, my local builds also have this issue:

curl -V
curl 7.55.1 (x86_64-pc-linux-gnu) libcurl/7.55.1 OpenSSL/1.0.2g zlib/1.2.11 libidn2/2.0.2 libpsl/0.18.0 (+libidn2/2.0.2) librtmp/2.3
Release-Date: 2017-08-14
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp smb smbs smtp smtps telnet tftp 
Features: AsynchDNS IDN IPv6 Largefile GSS-API Kerberos SPNEGO NTLM NTLM_WB SSL libz TLS-SRP UnixSockets HTTPS-proxy PSL

plugin/keyring_vault/CMakeLists.txt
robgolebiowski: This could be done for keyring_vault plugin, but not for the unittests. The reason being that unittests are only built when CURL does not depend on RTMP library. In case of unittests it has to be always a warning if they cannot be built. It is not a big problem, as unittests passing on just one jenkins machine is enough (cento6  currently). The rest of the machines run MTR tests.

unittest/gunit/keyring_vault/CMakeLists.txt
robgolebiowski: rtmp does not depend on the version of curl. It is just built with rtmp on some machines and on some it is not. From what I am aware of - rtmp has been built in with curl for quite some time.

plugin/keyring_vault/CMakeLists.txt
laurynas-biveinis: I agree with @percona-ysorokin. This will give us extra Jenkins config pain (this platform should build that, that platform should not build this etc), but it's the best way to ensure we don't silently break non-universally built parts

robgolebiowski: @laurynas-biveinis, @percona-ysorokin. Do we agree that keyring_vault options should just be ON/OFF. ON - causing cmake to stop if keyring_vault cannot be built, unittests when they cannot be built should just generate a warning ? This is because CURL that comes with the platform usually has rtmp dependency. If we stop build for unititests this means that all jenkins machine will need to compile a custom CURL without rtmp (there is a configuration switch for that). Even if this is doable then would not be better to have MTR tests run on the actual CURL that later will be used in production (the distribution default) ?

dutow: Just an idea: can't we add a WITH_CURL=bundled/system option, similar to the other 3rd parties? As I see, curl is also built with cmake, so this shouldn't be too hard to do.

robgolebiowski: @dutow bundled means built in into PS. For WITH_SSL we use bundled for building with YASSL that is shipped with PS. It is already possible to built with custom CURL by providing its path to WITH_CURL=, without the custom path it will just be built with the system CURL.

percona-ysorokin: So, let me summarize what we've agreed on.
A new `-DWITH_KEYRING_VAULT=ON | OFF` will be introduced to Percona Server build options.
The default is `ON`.

When `ON`, we will try to build `keyring_vault` plugin. If there is a missing dependency, Cmake will return a fatal error. We will also try to build 'keyring_vault' unit tests. It should also generate a fatal error on a missing dependency. The only exception is that it may generate a warning and skip building 'keyring_vault' unit tests when it finds a system `curl` library which is build with `RTMP` support.

When `OFF`, neither `keyring_vault` plugin nor `keyring_vault` unit tests are built.

plugin/keyring_vault/CMakeLists.txt
```diff
@@ -1,29 +1,27 @@
+IF (NOT DEFINED WITH_KEYRING_VAULT)
```
percona-ysorokin, line 1: Commit comment still mentions `FORCE` as a valid option for `WITH_KEYRING_VAULT`

percona-ysorokin (approved):
LGTM with comment regarding commit message addressed.

laurynas-biveinis (approved):

PR #2125, opened 2018-02-01
PS-311: On server shutdown threadpool is crashing in debug mode due to

robgolebiowski:
race condition

Moved calling set_my_sys_thread_var before signaling conditional variable
THR_COND_threads. This is variable on which my_thread_global_end is waiting.
After it is signalled THR_KEY_mysys_initialized is set to FALSE in my_thread_global_end.
There is a debug assertion that THE_KEY_mysys_initialized == TRUE in
set_my_sys_thread_var.

Moved my_init_signals, which sets signal mask to block all handleable kill signals
- which are later serviced by signal_hand thread. my_init_singals was
moved before PFS initialization. PFS thread can be affected by "handleable kill"
signals unless they have singal mask set.

### DIFF
--- a/mysys/my_thr_init.c
+++ b/mysys/my_thr_init.c
@@ -354,7 +354,14 @@ void my_thread_end()
       tmp->dbug= NULL;
     }
     free(tmp);
-
+    /*
+     set_mysys_thread_var (NULL) must be called before signaling conditional variable
+     THR_COND_threads. This is variable on which my_thread_global_end is waiting.
+     After it is signalled THR_KEY_mysys_initialized is set to FALSE in my_thread_global_end.
+     There is a debug assertion that THE_KEY_mysys_initialized == TRUE in
+     set_my_sys_thread_var.
+    */
+    set_mysys_thread_var(NULL);
     /*
       Decrement counter for number of running threads. We are using this
       in my_thread_global_end() to wait until all threads have called
@@ -367,7 +374,6 @@ void my_thread_end()
       mysql_cond_signal(&THR_COND_threads);
     mysql_mutex_unlock(&THR_LOCK_threads);
   }
-  set_mysys_thread_var(NULL);
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


robgolebiowski:
JR => https://jenkins.percona.com/job/mysql-5.7-param/1561/

sql/mysqld.cc
```diff
@@ -4781,6 +4781,7 @@ int mysqld_main(int argc, char **argv)
   sys_var_init();
   ulong requested_open_files;
   adjust_related_options(&requested_open_files);
+  my_init_signals();
```
laurynas-biveinis: The change looks correct by code review, but how come SIGTERM does not kill a server with a PFS thread running then currently? Or it does? What is the relation between this change and the current bug?

laurynas-biveinis: Please add a short comment "moved signal initialization here so that PFS thread inherits signal mask or similar"

laurynas-biveinis (changes requested):
Both changes look like they would apply to upstream as well?

sql/mysqld.cc
robgolebiowski: Sorry, I have not made myself clear about this. The problem is that after fixing this problem, an other crash happened - due to different reason. I have posted a comment in the bug description. Please have a look and tell me if this make more sense now.

laurynas-biveinis: Thanks - please update commit message too. But my question still stands: it looks like the current server should crash too without this patch, upstream, w/o threadpool, if it's the PFS thread that gets the SIGTERM?

robgolebiowski: This is due to the fact that error handling thread (function start_signal_handler) is called long after threadpool gets initialized. Basing on your comment in code - theadpool initialization is heavy which means that PFS threads have to survive longer period of time without any signal handling - as they do not get signal mask set.

mysys/my_thr_init.c
```diff
@@ -354,7 +354,7 @@ void my_thread_end()
       tmp->dbug= NULL;
     }
     free(tmp);
-
+    set_mysys_thread_var(NULL);
```
percona-ysorokin: Please, also add a comment here regarding why `set_mysys_thread_var(NULL)` should be called before signalling. The description form the commit message is precise enough.

percona-ysorokin (approved):
LGTM

laurynas-biveinis (approved):
LGTM with comments added

PR #2127, opened 2018-02-05
Fix PS-3794 (MTR test main.percona_show_temp_tables_stress does not w…

laurynas-biveinis:
…ait for events to start)

Synchronize the event scheduler-spawned worker threads with the main
connection thread by waiting for them to start up. At the same time
wait at the end of the testcase for the event scheduler to finish
stopping.

https://jenkins.percona.com/job/percona-server-5.5-param/1630/


percona-ysorokin (approved):
LGTM

PR #2130, opened 2018-02-05
PS-3788: (PS 5.7 partially dynamic @version variable using support of @version_suffix 

inikep:
Tested at https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1576/


sql/sys_vars.h
```diff
@@ -840,34 +840,54 @@ class Sys_var_charptr: public sys_var
 
 class Sys_var_version : public Sys_var_charptr
 {
+private:
+  char withsuffix[SERVER_VERSION_LENGTH];
```
laurynas-biveinis, line 844: s/char/uchar/g ?

sql/sys_vars.h
```diff
@@ -840,34 +840,54 @@ class Sys_var_charptr: public sys_var
...
+
+    memcpy(withsuffix, *version_ptr, version_ptr_len);
+    memcpy(withsuffix + version_ptr_len, *suffix_ptr, suffix_ptr_len);
+    withsuffix[suffix_ptr_len + version_ptr_len] = 0;
+
+    return reinterpret_cast<uchar*> (&withsuffix_ptr);
   }
 };
 
-
```
laurynas-biveinis: Spurious change

laurynas-biveinis (changes requested):

sql/sys_vars.h
inikep: I do not understand. What do you mean?
`char` is on purpose as is minimizes casts for `memcpy` and other.

laurynas-biveinis: Ack, OK then. My suggestion was in order to remove some of casts with uchar, please keep whichever results in fewer casts.

PR #2131, opened 2018-02-05
PS-3793: added osx travis config (PS5.5)

dutow:
* Added OSX to the build matrix.
* Fixed issues with older bash versions on OSX.
* Made maintainer_mode optional, as the OSX build results in some warnings, and fixing them would take time.
  As it's not an officially supported platform, warnings are ok for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@
 dist: trusty
 sudo: required
 language: c
+env:
+  global:
+    - MAINTAINER_MODE=ON
 matrix:
   include:
     - env: COMMAND=clang-test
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,11 +23,14 @@ matrix:
     - env: GCC=gcc-6     CXX=g++-6       LIBTYPE=system  PACKAGES=g++-6     PPA=ubuntu-toolchain-r/test
     - env: GCC=clang-4.0 CXX=clang++-4.0 LIBTYPE=system  PACKAGES="clang-4.0 llvm-4.0-dev" PPA=ubuntu-toolchain-r/test LLVM=llvm-toolchain-trusty-4.0
     - env: GCC=clang-5.0 CXX=clang++-5.0 LIBTYPE=bundled CMAKE_OPT="-DWITH_PAM=1" PACKAGES="clang-5.0 llvm-5.0-dev" LLVM=llvm-toolchain-trusty-5.0
+    - os: osx
+      osx_image: xcode9.3
```
laurynas-biveinis: XCode 9.3 is beta ATM, IMHO we should use 9.2 here

.travis.yml
```diff
@@ -20,11 +23,14 @@ matrix:
...
+      env: LIBTYPE=bundled GCC=clang CXX=clang++ MAINTAINER_MODE=OFF
```
laurynas-biveinis: Possible to do brew install openssl@1.0 and then LIBTYPE=system for faster builds. CMake needs -DWITH_SSL=system -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl/ then

laurynas-biveinis: GCC=clang CXX=clang++ should be redundant

laurynas-biveinis (changes requested):

.travis.yml
dutow: I think without setting them, the variables were empty, but I'll check it again.

dutow: The question is, which would be more realistic/better? For the other builds, we have builds wiith bundled and system libraries, but with osx, I wanted to avoid that, as the 5.7 build is slow anyway.

I went with bundled, and didn't even try it with the system version, because I though that uses a larger percentage of the codebase - but I'm not using osx for development, so you probably have a better idea about this.

laurynas-biveinis: There is no point in covering larger percentage of codebase since no official support: the last thing I want is to fix platform bugs in shipped 3rd party libraries, which we don't do much on Linux neither. Hence, build speed and not build breaking due to 3rd party libs is more important here, IMHO

inikep: This seems to be `clang-4.0` issue, not macOS issue.
I have the same error as macOS on Ubuntu with `clang-4.0` at https://travis-ci.org/inikep/percona-server/jobs/337929558:
```
error: invalid value 'gnu++03' in '-std=gnu++03'
make[2]: *** [sql/CMakeFiles/gen_lex_hash.dir/gen_lex_hash.cc.o] Error 1
make[1]: *** [sql/CMakeFiles/gen_lex_hash.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....```
```

dutow: @inikep - right, with the changes laurynas suggested in the 5.7 review, this is also fixed

@laurynas-biveinis - I tested this, and looks like using brew / the system libraries doesn't result in a speed improvement - maybe because the brew install task itself takes time. On the other hand, it complicates the config because of the required additional cmake option.

.travis.yml
dutow: Changed.

laurynas-biveinis (approved):

PR #2132, opened 2018-02-05
PS-3793: added osx travis config (5.6)

dutow:
* Added OSX to the build matrix.
* Fixed issues with older bash versions on OSX.
* Made maintainer_mode optional, as the OSX build results in some warnings, and fixing them would take time.
  As it's not an officially supported platform, warnings are ok for now.
* Made tokudb optional, as the OSX build doesn't support it

(cherry picked from commit 568ab7d75834ea909dacd94a1134226827d5de7b)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@
 dist: trusty
 sudo: required
 language: c
+env:
+  global:
+    - MAINTAINER_MODE=ON
+    - WITHOUT_TOKUDB=OFF
 matrix:
   include:
     - env: COMMAND=clang-test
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,11 +24,14 @@ matrix:
     - env: GCC=gcc-6     CXX=g++-6       LIBTYPE=system  CMAKE_OPT="-DWITH_READLINE=system" PACKAGES=g++-6     PPA=ubuntu-toolchain-r/test
     - env: GCC=clang-4.0 CXX=clang++-4.0 LIBTYPE=system  CMAKE_OPT="-DWITH_READLINE=system" PACKAGES="clang-4.0 llvm-4.0-dev" PPA=ubuntu-toolchain-r/test LLVM=llvm-toolchain-trusty-4.0
     - env: GCC=clang-5.0 CXX=clang++-5.0 LIBTYPE=bundled CMAKE_OPT="-DWITH_PAM=1"           PACKAGES="clang-5.0 llvm-5.0-dev" LLVM=llvm-toolchain-trusty-5.0
+    - os: osx
+      osx_image: xcode9.3
+      env: LIBTYPE=bundled GCC=clang CXX=clang++ MAINTAINER_MODE=OFF WITHOUT_TOKUDB=ON
```
laurynas-biveinis: Difference from 5.5 review comments: the OpenSSL system CMake option becomes -DWITH_SSL=/usr/local/open/openssl/, and also possible to do brew install libevent and -DWITH_LIBEVENT=system

laurynas-biveinis (changes requested):

.travis.yml
dutow: As I mentioned for the 5.5 pr: I didn't see any consistent speedup, because brew also takes time. Also, the added option would make the travis config more complex.

Otherwise I updated the PR based on the changes in the 5.5 version.

laurynas-biveinis (approved):

PR #2133, opened 2018-02-05
PS-3793: OSX travis config and build fixes (PS5.7)

dutow:


.travis.yml
```diff
@@ -20,11 +25,14 @@ matrix:
     - env: GCC=gcc-6     CXX=g++-6       LIBTYPE=system  CMAKE_OPT="-DWITH_READLINE=system" PACKAGES=g++-6     PPA=ubuntu-toolchain-r/test
     - env: GCC=clang-4.0 CXX=clang++-4.0 LIBTYPE=system  CMAKE_OPT="-DWITH_READLINE=system" PACKAGES="clang-4.0 llvm-4.0-dev" PPA=ubuntu-toolchain-r/test LLVM=llvm-toolchain-trusty-4.0
     - env: GCC=clang-5.0 CXX=clang++-5.0 LIBTYPE=bundled CMAKE_OPT="-DWITH_PAM=1"           PACKAGES="clang-5.0 llvm-5.0-dev" LLVM=llvm-toolchain-trusty-5.0
+    - os: osx
+      osx_image: xcode9.3
+      env: LIBTYPE=bundled GCC=clang CXX=clang++ MAINTAINER_MODE=OFF WITHOUT_TOKUDB=ON WITHOUT_ROCKSDB=ON
```
laurynas-biveinis: In addition to what's suggested on the 5.6 review, it's also possible to do brew install lz4 mecab mecab-ipadic protobuf, and use system for all of those.

cmake/build_configurations/compiler_options.cmake
```diff
@@ -29,7 +29,10 @@ SET(CMAKE_CXX_STANDARD 98)
  
 # Compiler options
 IF(UNIX)  
-  SET(CMAKE_CXX98_EXTENSION_COMPILE_OPTION -std=gnu++03)
+  IF(NOT APPLE)
```
laurynas-biveinis: In theory it's possible to use a recent GCC on macOS, thus I'd try NOT (APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang"). But maybe it's even possible to get a "community" clang version there as well, so maybe it should be a direct CMake compiler runtime check whether the option is accepted. If this is not overengineering.

laurynas-biveinis (changes requested):

cmake/build_configurations/compiler_options.cmake
dutow: It's also possible with travis, it has a gcc7 profile for example. For this, my starting point was our slack discussion, with the conclusion that we only need clang.

I'll change it to a runtime check then - and this also affects all 3 branches.

laurynas-biveinis: ack, sorry for not being clear on Slack

dutow: Updated it, and this also fixed the clang4 issue.

laurynas-biveinis (approved):
(note when pushing that now Travis config conflicts with mecab additions)

dutow:
I rebased to the latest 5.7 head and fixed the conflicts. Also disabled the mecab setting for osx, as it causes build errors there.

laurynas-biveinis:
@dutow, please merge

PR #2134, opened 2018-02-06
Fix PS-3798 (MTR test innodb.percona_extended_innodb_status fails if …

laurynas-biveinis:
…InnoDB status contains unquoted special characters)

The testcase performed

--let $status = query_get_value(SHOW ENGINE INNODB STATUS, Status, 1)
eval INSERT INTO t VALUES("$status");

which does not perform any SQL quoting on $status, breaking the INSERT
statement in case it contains quote characters. Fix by redirecting
status output to a temporary file and searching for a pattern there
instead.

https://jenkins.percona.com/job/percona-server-5.5-param/1631/


percona-ysorokin (approved):
LGTM

PR #2138, opened 2018-02-07
PS-3802: Travis CI: Separate jobs for Debug and RelWithDebInfo (for 5.7)

inikep:
Fix 50 min timeout what is not enough for Travis CI

### DIFF
--- a/storage/innobase/include/srv0srv.h
+++ b/storage/innobase/include/srv0srv.h
@@ -197,10 +197,6 @@ at a time */
 extern ib_mutex_t	page_zip_stat_per_index_mutex;
 /* Mutex for locking srv_monitor_file. Not created if srv_read_only_mode */
 extern ib_mutex_t	srv_monitor_file_mutex;
-
-/* prototypes for new functions added to ha_innodb.cc */
-bool	innobase_get_slow_log();
-
 /* Temporary file for innodb monitor output */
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -11,79 +12,83 @@ matrix:
...
-    -DWITH_SSL=$LIBTYPE
-    -DWITH_ZLIB=$LIBTYPE
-    -DWITH_LIBEVENT=$LIBTYPE
     -DENABLE_DOWNLOADS=1
     -DDOWNLOAD_BOOST=1
     -DWITH_BOOST=../deps
     -DWITHOUT_TOKUDB=ON
     -DWITHOUT_ROCKSDB=ON
+    -DWITH_RAPID=OFF
      $CMAKE_OPT
```
percona-ysorokin: Please, add `-DWITH_PAM=ON` unconditionally here and remove it from `CMAKE_OPT` in the matrix.
`-DWITH_PAM` just enables an additional code tree to be built, it's not like '-DWITH_EMBEDDED_SERVER` which is used in a number of `#ifdef`s. So, it just make sense to always enable it.

percona-ysorokin: Depending on the status of #2123 it also make sense to add `-DWITH_KEYRING_VAULT=ON` here.

PR #2139, opened 2018-02-07
TDB-89 : Row count and cardinality drops to zero after bug #1607300 i…

george-lorch:
…s fixed

TDB-107 : tsan detected a data race in the minicron between the 'do' and 'change period' functions
- Advances the git submodule commit pointer for PerconaFT to pick up fixes for
  these two issues.
- Will be cherry picked forwward.


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2081/

george-lorch:
Self approving...

PR #2141, opened 2018-02-08
PS-314: issue with 65536+ threads and mdl locks

dutow:
MDL uses the LF_PINS structure for maintaining a global map, which had a 16 bit limitations.
After reaching 65535 threads, new threads couldn't access the MDL structures until older threads disconnected.
This limitation also caused assertions in the debug builds, stopping the server.

This patch changes the 16 bit limitation of the LF_PINS structure to 32 bit.
While this solves the problem, it does so with a high memory cost:
as the LF_PINS structure uses the LF_DYNARRAY for storing its data, a 1GB array will be allocated when reaching 65536+256+1 threads.
99% of that memory will never be used, as MySQL has a connection limit of 100000, and this array would be enough for 4 billion.
As servers with 65k+ connections will use more than 40GB of address space anyway, this shouldn't be an issue, especially as most of it will never be accessed..

### DIFF
--- a/include/lf.h
+++ b/include/lf.h
@@ -59,19 +59,19 @@ typedef struct {
   lf_pinbox_free_func *free_func;
   void *free_func_arg;
   uint free_ptr_offset;
-  uint32 volatile pinstack_top_ver;         /* this is a versioned pointer */
-  uint32 volatile pins_in_array;            /* number of elements in array */
+  uint64 volatile pinstack_top_ver;         /* this is a versioned pointer */
+  uint64 volatile pins_in_array;            /* number of elements in array */
 } LF_PINBOX;
 
 typedef struct st_lf_pins {
   void * volatile pin[LF_PINBOX_PINS];
   LF_PINBOX *pinbox;
   void  *purgatory;
-  uint32 purgatory_count;
-  uint32 volatile link;
+  uint64 purgatory_count;
+  uint64 volatile link;
 /* we want sizeof(LF_PINS) to be 64 to avoid false sharing */
-#if SIZEOF_INT*2+SIZEOF_CHARP*(LF_PINBOX_PINS+2) != 64
-  char pad[64-sizeof(uint32)*2-sizeof(void*)*(LF_PINBOX_PINS+2)];
+#if 2*8+SIZEOF_CHARP*(LF_PINBOX_PINS+2) != 64
+  char pad[64-sizeof(uint64)*2-sizeof(void*)*(LF_PINBOX_PINS+2)];
 #endif
 } LF_PINS;
 
--- a/mysys/lf_alloc-pin.c
+++ b/mysys/lf_alloc-pin.c
@@ -92,13 +92,13 @@
   Pins are given away from a "pinbox". Pinbox is stack-based allocator.
   It used dynarray for storing pins, new elements are allocated by dynarray
   as necessary, old are pushed in the stack for reuse. ABA is solved by
-  versioning a pointer - because we use an array, a pointer to pins is 16 bit,
-  upper 16 bits are used for a version.
+  versioning a pointer - because we use an array, a pointer to pins is 32 bit,
+  upper 32 bits are used for a version.
 */
 #include "lf.h"
 #include "mysys_priv.h" /* key_memory_lf_node */
 
-#define LF_PINBOX_MAX_PINS 65536
+#define LF_PINBOX_MAX_PINS (65536ULL*65536ULL)
 
 static void lf_pinbox_real_free(LF_PINS *pins);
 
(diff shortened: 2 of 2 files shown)


include/lf.h
```diff
@@ -59,19 +59,19 @@ typedef struct {
...
   LF_PINBOX *pinbox;
   void  *purgatory;
-  uint32 purgatory_count;
-  uint32 volatile link;
+  uint64 purgatory_count;
+  uint64 volatile link;
 /* we want sizeof(LF_PINS) to be 64 to avoid false sharing */
-#if SIZEOF_INT*2+SIZEOF_CHARP*(LF_PINBOX_PINS+2) != 64
-  char pad[64-sizeof(uint32)*2-sizeof(void*)*(LF_PINBOX_PINS+2)];
+#if SIZEOF_INT*4+SIZEOF_CHARP*(LF_PINBOX_PINS+2) != 64
```
laurynas-biveinis: Is the SIZEOF_INT dependency correct here, and why it's needed in the first place when we use exact-width type? I.e. can SIZEOF_INT*4 can be replaced with 16?

laurynas-biveinis (commented):
I guess adding a (even --big-test) testcase is not that easy?

include/lf.h
dutow: It can be, I'll change it.

dutow:
For a test case: I'm not sure how we could add it

* it needs a python (or some other) script which establishes 65k+ connections
* it has to change some sysctl settings/process limits (which requires root privileges)

laurynas-biveinis:
MTR can do connections in a loop, the testcase could assume that sysctls are set properly or skip itself

include/lf.h
percona-ysorokin: I would also add a compile-time assertion that `(8*2 + SIZEOF_CHARP*(LF_PINBOX_PINS+2) < 64)`

mysys/lf_alloc-pin.c
```diff
@@ -92,13 +92,13 @@
...
-  versioning a pointer - because we use an array, a pointer to pins is 16 bit,
-  upper 16 bits are used for a version.
+  versioning a pointer - because we use an array, a pointer to pins is 32 bit,
+  upper 32 bits are used for a version.
 */
 #include "lf.h"
 #include "mysys_priv.h" /* key_memory_lf_node */
 
-#define LF_PINBOX_MAX_PINS 65536
+#define LF_PINBOX_MAX_PINS (65536l*65536l)
```
percona-ysorokin: On 32-bit platforms long is 32-bit wide so `1^32` won't fit into `long`. So `#define LF_PINBOX_MAX_PINS (65536ULL*65536ULL)`

mysys/lf_alloc-pin.c
```diff
@@ -153,7 +152,7 @@ LF_PINS *lf_pinbox_get_pins(LF_PINBOX *pinbox)
     if (!(pins= top_ver % LF_PINBOX_MAX_PINS))
     {
       /* the stack of free elements is empty */
-      pins= my_atomic_add32((int32 volatile*) &pinbox->pins_in_array, 1)+1;
+      pins= my_atomic_add64((int64 volatile*) &pinbox->pins_in_array, 1)+1;
```
percona-ysorokin, line 155: General question: won't 64-bit atomic operations on x86 be a big performance degradation? Moreover, I remember some bugs related to implementing double CAS operations in old GCCs.
May be we should consider making this change for x64 only, but leave original implementation for x86? We will just have to decrease `connection_limit` for x86 from `100000` to `65536-<number_of_admin_connections>`.

include/lf.h
dutow: I updated this part.

I also didn't add any mtr test yet - based on Friday's meeting.

mysys/lf_alloc-pin.c
laurynas-biveinis: Very strictly speaking the constants have to be UL (signed long guaranteed max is 2B only)

laurynas-biveinis (approved):
The only comment is too minor to require addressing

percona-ysorokin (changes requested):

include/lf.h
dutow: There is a compile time assert in the c file about the size of the struct, which is basically the same.

mysys/lf_alloc-pin.c
dutow: That's possible too. I was thinking about something like this, but then I've seen that there are 64 bit atomic operations at other places in the code, so I went with the simpler approach.

Even if we do not change the connection limit, the 32 bit client will run out of address space way before it reaches this number.

laurynas-biveinis: It's OK to ignore x86 performance here (correctness not so much, but I don't recall any recent bugs in 32 bit CMPXCHG8B output)

mysys/lf_alloc-pin.c
dutow: Updated.

dutow:
I updated the maximum size.

laurynas-biveinis (approved):

percona-ysorokin (approved):
LGTM

dutow:
Jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1626/
Usual random failures, I can't see any test that fails on every platform because of this.

PR #2146, opened 2018-02-12
Fix PS-3813 (Leftover function declaration innobase_get_slow_log)

laurynas-biveinis:
Remove the leftover declaration.

https://jenkins.percona.com/job/mysql-5.7-param/1604/


laurynas-biveinis:
Trivial, self-approving

PR #2149, opened 2018-02-12
PS-3815: optnone is ignored by GCC

dutow:
The optnone attirbute is ignored by gcc, and the funcsion remains optimized.

This commit disables optimization with different attributes correctly for both gcc and clang.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -11929,7 +11929,11 @@ const char *get_rdb_io_error_string(const RDB_IO_ERROR_TYPE err_type) {
 // so that we can capture as much data as possible to debug the root cause
 // more efficiently.
 #if defined(DBUG_OFF)
+#ifdef __clang__
 MY_ATTRIBUTE((optnone))
+#else
+MY_ATTRIBUTE((optimize("O0")))
+#endif
 #endif
 void rdb_handle_io_error(const rocksdb::Status status,
                          const RDB_IO_ERROR_TYPE err_type) {


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11929,7 +11929,11 @@ const char *get_rdb_io_error_string(const RDB_IO_ERROR_TYPE err_type) {
 // so that we can capture as much data as possible to debug the root cause
 // more efficiently.
 #if defined(DBUG_OFF)
+#ifdef __clang__
 MY_ATTRIBUTE((optnone))
+#else
+MY_ATTRIBUTE((optimize("O0")))
```
laurynas-biveinis, line 11935: What is "This attribute should be used for debugging purposes only. It is not suitable in production code." trying to say over at https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes ?

I don't think that makes it unusable here, but something to keep in mind if this will need revisiting again

laurynas-biveinis (approved):

storage/rocksdb/ha_rocksdb.cc
dutow: I googled about this:

* originally we used pragma optimize here, which has the same issues (uses the same code)
* the issue according to some other projects is that using either this attribute or the pragmas and lto results in compiler crashes with some specific gcc versions. As we do not see that anywhere, it should work for us.

dutow:
Jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1612/

PR #2163, opened 2018-02-14
DOC-694 - Change rocksdb_bytes_per_sync and rocksdb_wal_bytes_per_syn…

fiowro:
…c to be dynamic


hrvojem (changes requested):
Please add version info, so users are aware when the change took place. 
Add this info to release notes as well.

hrvojem (approved):
LGTM

PR #2166, opened 2018-02-14
Fix PS-3812 (Revert PS-3360 (Busy server prefers LRU flushing over fl…

laurynas-biveinis:
…ush list flushing too strongly))

Once the main cleaner loop was performing both LRU and flush list
flushing, it used to skip the flush list part if the LRU part took too
long. This broke sync flushes and PS-3360 fixed this by not skipping
the rest of iteration. This became meaningless with MT LRU flusher -
which removed LRU flushing from the cleaner loop - and left no-op code
and outdated comment. Removed.

https://jenkins.percona.com/job/mysql-5.7-param/1614/


george-lorch (approved):
I am no expert in this code, but, the reason given and the logic change looks correct.

PR #2167, opened 2018-02-14
add info about renaming and deprecation of the rocksdb_concurrent_pre…

fiowro:
…pare variable along with DOC-698


george-lorch (approved):
+1

PR #2170, opened 2018-02-14
MYR-126 : rocksdb.concurrent_alter unstable

george-lorch:
- Test uses multiple instances of mysqlslap with multiple threads each, which
  occasionally causes mysqld to emit "[Warning] Too many connections" to the
  error log.
- Fixed this in two ways, one was to raise the servers max_connections to 500.
  The other was to suppress the warning as it means nothing to the test.
- Also added big_test as a single instance of this test can overtake an entire
  system for the duration of the test.

### DIFF
--- a/mysql-test/suite/rocksdb/r/concurrent_alter.result
+++ b/mysql-test/suite/rocksdb/r/concurrent_alter.result
@@ -1,3 +1,6 @@
+set @orig_max_connections=@@global.max_connections;
+set @@global.max_connections=500;
+call mtr.add_suppression("Too many connections");
 CREATE DATABASE mysqlslap;
 use mysqlslap;
 CREATE TABLE a1 (a int, b int) ENGINE=ROCKSDB;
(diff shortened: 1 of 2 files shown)


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1618/

mysql-test/suite/rocksdb/r/concurrent_alter.result
```diff
@@ -1,3 +1,6 @@
+set @orig_max_connections=@@global.max_connections;
+set @@global.max_connections=500;
+call mtr.add_suppression("Too many connections");
```
laurynas-biveinis, line 3: I would leave the warning unsuppressed. 500 should be plenty, and if the warning is still seen, this would indicate something else needing investigating going on.

laurynas-biveinis (changes requested):

george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1640/

george-lorch:
OK, removed suppression, tests still look stable with only bumping max_connections.

laurynas-biveinis (approved):

PR #2179, opened 2018-02-19
Fix PS-3843 (Reconcile EINTR handling after WL#7696)

laurynas-biveinis:
Remove EINTR handling code from os_file_create_simpl_func and make
os_file_get_last_error_low return OS_FILE_AIO_INTERRUPTED for every
EINTR, regardless of srv_use_native_aio value.

OS_FILE_AIO_INTERRUPTED should be renamed to remove "AIO", but this is
not done in order to keep diff minimal.

https://jenkins.percona.com/job/mysql-5.7-param/1643/

### DIFF
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -3186,9 +3186,7 @@ os_file_get_last_error_low(
 		}
 		break;
 	case EINTR:
-		if (srv_use_native_aio) {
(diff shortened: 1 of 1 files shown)


george-lorch (approved):
This looks good to me, assuming that we can be assured that the upstream WL changes are coming in the next merge to properly handle the EINTR case.

laurynas-biveinis:
This upstream WL was merged in 5.7.9 or so

PR #2182, opened 2018-02-19
MYR-234 : lower default value of rocksdb_max_row_locks to 1 million

george-lorch:
- To help prevent OOM on bulk/large operations and return an error to client
  before server is killed as RocksDB keep all uncommitted changes in memory
  until commit. This is only a temporary fix to try to prevent server kills.
  Longer term fix is in the works upstream.
- Reduced rocksdb_max_row_locks from 1024 * 1024 * 1024 (~ 1 billion) to
  1024 * 1024 (1 million) and re-recorded tests that need to be adjusted.


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1641/

laurynas-biveinis (approved):

PR #2183, opened 2018-02-19
Skip trailing space bytes for non-unpackable fields

george-lorch:
Upstream commit ID : fb-mysql-5.6.35/f3d36951a62aa1a00a5a10083fb77e9919022e03
Cherry picked directly.

Summary:
We always generate unpack info bytes that indicate how much trailing space we need, even when the field can never be unpacked, because it is not implemented. When skipping these fields however, we only skip the key portion, and not the value portion which corrupts data for code that reads unpack_info downstream. Fix by skipping these bytes.

Closes #784
Closes https://github.com/facebook/mysql-5.6/pull/785

Differential Revision: D6699483

fbshipit-source-id: fd5c9d6

### DIFF
--- a/mysql-test/suite/rocksdb/t/type_varchar.test
+++ b/mysql-test/suite/rocksdb/t/type_varchar.test
@@ -71,3 +71,12 @@ select 'email_i' as index_name, count(*) AS count from t force index(email_i);
 drop table t;
 set global rocksdb_checksums_pct = @save_rocksdb_checksums_pct;
 set session rocksdb_verify_row_debug_checksums = @save_rocksdb_verify_row_debug_checksums;
+
+# Issue #784 - Skip trailing space bytes for non-unpackable fields
+
+create table t (h varchar(31) character set utf8 collate utf8_bin not null, i varchar(19) collate latin1_bin not null, primary key(i), key(h)) engine=rocksdb;
+insert into t(i,h) values('a','b');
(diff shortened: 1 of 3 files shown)


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1642/

mysql-test/suite/rocksdb/t/type_varchar.test
```diff
@@ -71,3 +71,14 @@ select 'email_i' as index_name, count(*) AS count from t force index(email_i);
 drop table t;
 set global rocksdb_checksums_pct = @save_rocksdb_checksums_pct;
 set session rocksdb_verify_row_debug_checksums = @save_rocksdb_verify_row_debug_checksums;
+
+# Issue #784 - Skip trailing space bytes for non-unpackable fields
+
+drop table if exists t;
```
laurynas-biveinis: Remove

laurynas-biveinis (changes requested):
Trivial comment, pre-approved

mysql-test/suite/rocksdb/t/type_varchar.test
george-lorch: Ahh, yeah, missed that (obviously). Will remove it, re-record and retest locally (skip jenkins) before merging. Thanks!

PR #2184, opened 2018-02-20
PS-1096: destroy PSI objects when removing thread from pool

dutow:
Previous versions used my_thread_end, which was changed to some other code in 5.7. The call to free up the psi resources was missing from it.


dutow:
Jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1656/

laurynas-biveinis (approved):

PR #2185, opened 2018-02-20
PS-3848: macOS build broken, undefined symbol __db_keyword_, referenced

robgolebiowski:
from binary_log::Log_event_header::Log_event_header

binlog_event.cpp as a part of libbinlogevent should not depend on
mtr debug variables and thus on my_dbug.h. Created
binary_log_debug::debug_expect_unknown_event variable which value
depends on the value of mtr debug variable expect_Unknown_event.
This is the correct way to make binlog_event.cpp not depend on
my_dbug.h


laurynas-biveinis (approved):

robgolebiowski:
JR => https://jenkins.percona.com/job/mysql-5.7-param/1657/

PR #2187, opened 2018-02-20
Ps 5.7 myr 160

george-lorch:


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1651/

laurynas-biveinis (commented):
The testsuite bit is approved, re. shutdowns, I think you have seen errors with 600 too? This might indicate an actual stall and warrants further investigation before bumping then

george-lorch:
re: stalls, yes, I am quite concerned about this, see PS-3846. We can discuss via another venue.

PR #2190, opened 2018-02-21
PS-1095: Fixed a possible integer overflow in threadpool_unix PS 5.7

dutow:


mysql-test/t/percona_bug_ps1095.test
```diff
@@ -0,0 +1,16 @@
+# Bug PS-1095: Issues then thread_pool_stall_timeout is above 2.1 seconds
+
+-- source include/have_pool_of_threads.inc
+
+CREATE TABLE t1 (i int);
+
+--sleep 9
+
+# The server should be disconnected by the timer thread
```
laurynas-biveinis: Don't assume that in our Jenkins the timer thread will be scheduled to be run at least once, but synchronize: replace --error/drop table with wait_until_disconnected.inc

sql/threadpool_unix.cc
```diff
@@ -568,7 +568,7 @@ static void* timer_thread(void *param)
     struct timespec ts;
     int err;
 
-    set_timespec_nsec(&ts, timer->tick_interval*1000000);
+    set_timespec_nsec(&ts, timer->tick_interval*1000000ULL);
```
laurynas-biveinis: Spaces around * since you are touching this line

laurynas-biveinis (changes requested):
Consider adding testcase to 5.5/5.6 too

dutow:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1664/

laurynas-biveinis (approved):

PR #2197, opened 2018-02-26
PS-290: Crash in the unsafe query warning checks with views (5.7)

dutow:
In 5.7, the table pointer is null on updatable views, tables are
listed in the view_tables list.


mysql-test/t/percona_bug_ps290.test
```diff
@@ -0,0 +1,9 @@
+# PS-290: Unsafe query warning improvements caused a crash
```
laurynas-biveinis: The test should go to binlog suite, and SET binlog_format replaced with have_binlog_format_statement.inc

laurynas-biveinis (changes requested):
I am not sure re. its status in the lower versions: LP said "New", i.e. not checked there yet, and JIRA says "Not 5.5, Not 5.6", i.e. checked and found to not apply there. Is this correct?

In any case, please add the testcase to the lower versions too

sql/sql_base.cc
```diff
@@ -10254,7 +10254,11 @@ inline Join_node::Join_node(const TABLE_LIST* table,
                             Item* cond, const ORDER* order)
 {
   max_sort_length= current_thd->variables.max_sort_length;
-  add_table(table->table);
+  if (table->table) {
+    add_table(table->table);
+  } else {
```
laurynas-biveinis: }
else
{

dutow:
Yes, I checked it, and it's 5.7 only.

dutow:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1677/

laurynas-biveinis (commented):
Same as with 5.5 re. testcase, code part looks good

PR #2202, opened 2018-02-27
PS-290: Added testcase (5.5)

dutow:


dutow:
https://jenkins.percona.com/view/PS%205.5/job/percona-server-5.5-param/1633/

mysql-test/suite/binlog/t/percona_bug_ps290-master.opt
```diff
@@ -0,0 +1,2 @@
+--log-bin
```
laurynas-biveinis: Is the option file required?

laurynas-biveinis (commented):
Approved if the option file is indeed required, otherwise approved with it removed

mysql-test/suite/binlog/t/percona_bug_ps290-master.opt
dutow: No, looks like it's not needed, I removed the file.

laurynas-biveinis (approved):

PR #2219, opened 2018-03-07
TDB-114 : pre-8.0 : Change use of MySQL HASH to unordered_map

george-lorch:
- Harmless change that changes the container of currently opened TOKUDB_SHARE
  instances from themysql HASH to std::unordered_map to prep for the coming 8.0
  change that alters the API to the HASH in ways we just don't want to deal
  with.

### DIFF
--- a/storage/tokudb/ha_tokudb.cc
+++ b/storage/tokudb/ha_tokudb.cc
@@ -34,7 +34,7 @@ Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved.
 pfs_key_t ha_tokudb_mutex_key;
 pfs_key_t num_DBs_lock_key;
 
-HASH TOKUDB_SHARE::_open_tables;
+std::unordered_map<std::string, TOKUDB_SHARE*> TOKUDB_SHARE::_open_tables;
 tokudb::thread::mutex_t TOKUDB_SHARE::_open_tables_mutex;
 
 static const char* ha_tokudb_exts[] = {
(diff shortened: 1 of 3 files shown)


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2099/

george-lorch:
Re-pushed a re-commit in order to properly credit Rik Prohaska, code is same, only commit id, author, and message has been changed.

storage/tokudb/ha_tokudb.cc
```diff
@@ -152,17 +152,15 @@ void TOKUDB_SHARE::hash_free_element(TOKUDB_SHARE* share) {
...
-        (my_hash_get_key)hash_get_key,
-        (my_hash_free_key)hash_free_element, 0);
+    assert_always(_open_tables.size() == 0);
 }
 void TOKUDB_SHARE::static_destroy() {
-    my_hash_free(&_open_tables);
+    for (auto it = _open_tables.cbegin(); it != _open_tables.cend(); ) {
+        TOKUDB_TRACE("_open_tables %s %p", it->first.c_str(), it->second);
+        hash_free_element(it->second);
+        it = _open_tables.erase(it);
```
laurynas-biveinis: Can you just iterate over the contents here and do clear() after the loop?

storage/tokudb/ha_tokudb.cc
```diff
@@ -218,11 +216,12 @@ TOKUDB_SHARE* TOKUDB_SHARE::get_share(const char* table_name,
                                       THR_LOCK_DATA* data,
                                       bool create_new) {
     mutex_t_lock(_open_tables_mutex);
-    int error = 0;
-    uint length = (uint)strlen(table_name);
-    TOKUDB_SHARE* share = (TOKUDB_SHARE*)my_hash_search(
-        &_open_tables, (uchar*)table_name, length);
-
+    auto it = _open_tables.find(std::string(table_name));
```
laurynas-biveinis: Extract std::string(table_name) to a variable

storage/tokudb/hatoku_defines.h
```diff
@@ -60,6 +60,9 @@ Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved.
 #include "misc.h"
 #endif
 
+#include <string>
```
laurynas-biveinis, line 63: I guess there are reasons why not ha_tokudb.h itself includes these?

laurynas-biveinis (changes requested):

storage/tokudb/hatoku_defines.h
george-lorch: Yeah, geneally Toku tries to have all 'outside' includes in hatoku_defines.h, particularly anything mysql (which this is admittedly not) to make sure they are in proper order as there are problems with including various things from the server if not done in exactly the correct order :(

storage/tokudb/ha_tokudb.cc
george-lorch: True, and actually, this shows another minor problem, there is no need for hash_free_element, the code should just do
`it->second->destroy();
delete it->second;`

george-lorch:
OK, fixed up, found a little more vestigial code that could be removed as a result of the removal of the HASH, jenkins here : https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2102/

storage/tokudb/ha_tokudb.cc
```diff
@@ -139,30 +139,18 @@ static void free_key_and_col_info (KEY_AND_COL_INFO* kc_info) {
...
 void TOKUDB_SHARE::static_destroy() {
-    my_hash_free(&_open_tables);
+    for (auto it = _open_tables.cbegin(); it != _open_tables.cend(); it++) {
+        TOKUDB_TRACE("_open_tables %s %p", it->first.c_str(), it->second);
+        TOKUDB_SHARE* share = it->second;
+        share->destroy();
+        delete share;
+    }
+    _open_tables.clear();
+    assert_always(_open_tables.size() == 0);
```
laurynas-biveinis, line 153: I wouldn't bother with this assert right after clear call, but if it's there, it can stay

laurynas-biveinis (approved):

PR #2225, opened 2018-03-12
PS-310: Added support for building PS 5.6 with clang and MSAN

dutow:
This change adds the option to build PS 5.6 with MSAN, but doesn't try to fix any actual issues in the main MySQL code.

The only issue fixed is an uninitailized read in the mysqltest binary - without that, the test suite can't be started.

After these changes, running the MTR tests is possible, but there are still some remaining errors.
To build with MSAN:

* use an instrumented libcxx (with msan)
* use the bundled zlib and libssl

And set the following environment variables before running CMake:

* CFLAGS: -fsanitize=memory
* CXXFLAGS: -fsanitize=memory -Wno-unused-command-line-argument -nostdinc++ -stdlib=libc++
* CC, CXX to clang
* CPLUS_INCLUDE_PATH, C_INCLUDE_PATH, LIBRARY_PATH, LD_LIBRARY_PATH to use the instrumented libcxx

the LD_LIBRARY_PATH setting is also required for running any of the binaries, otherwise MSAN will report false positives.

(cherry picked from commit efb02379116a56af2c91933cd36c0cd79a6e7cf0)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -291,6 +291,28 @@ IF (WITH_ASAN)
   ENDIF()
 ENDIF()
 
+OPTION(WITH_MSAN "Enable memory sanitizer" OFF)
+IF(WITH_MSAN)
+  # TLS model is required because otherwise MSAN has issues with thread local
+  # variable initialization in dynamic libraries
+  SET(MSAN_FLAGS "-fsanitize=memory -fsanitize-memory-track-origins=2 -ftls-model=initial-exec -fPIC -fno-omit-frame-pointer -fsanitize-blacklist=${CMAKE_CURRENT_SOURCE_DIR}/cmake/msan-blacklist.txt")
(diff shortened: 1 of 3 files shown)


percona-ysorokin (approved):
LGTM

dutow:
Jenkins: https://jenkins.percona.com/job/percona-server-5.6-param/2107/

laurynas-biveinis:
Why this has been merged without 5.7 version being ready?

dutow:
5.7 already includes the WITH_MSAN flag, so technically we can already build it with MSAN.

Running however is a different question, as it requires a patched clang, because it uses some functions in glibc which aren't handled correctly by the current MSAN.

laurynas-biveinis:
Sorry, then it's OK, if it's a "null merge" to 5.7

PR #2226, opened 2018-03-12
TDB-130: alter TokuDB table comment rebuild whole engine data

fungo:
Alter TokuDB table comment should not using copying algorithm and
rebuild whole engine data, only FRM definition should be changed.

Make `ha_tokudb::check_if_supported_inplace_alter()` return
`HA_ALTER_INPLACE_EXCLUSIVE_LOCK` while altering comment, and mysql
server will invoke `ha_tokudb::inplace_alter_table` which will do
nothing for altering comment.

### DIFF
--- a/storage/tokudb/ha_tokudb_alter_56.cc
+++ b/storage/tokudb/ha_tokudb_alter_56.cc
@@ -248,6 +248,40 @@ static bool only_flags(ulong bits, ulong mask) {
     return (bits & mask) != 0 && (bits & ~mask) == 0;
 }
 
+// Table create options that should be ignored by TokuDB
+// There are totally 25 create options defined by mysql server(see handler.h),
+// and only 4 options will touch engine data, either rebuild engine data or
+// just update meta info:
+//   1. HA_CREATE_USED_AUTO        update auto_inc info
+//   2. HA_CREATE_USED_CHARSET     rebuild table if contains character columns
+//   3. HA_CREATE_USED_ENGINE      rebuild table
+//   4. HA_CREATE_USED_ROW_FORMAT  update compression method info
+//
+// All the others are either not supported by TokuDB or no need to
+// touch engine data.
+static const uint32_t TOKUDB_IGNORED_ALTER_CREATE_OPTION_FIELDS =
+    HA_CREATE_USED_RAID |              // deprecated field
+    HA_CREATE_USED_UNION |             // for MERGE table
+    HA_CREATE_USED_INSERT_METHOD |     // for MERGE table
+    HA_CREATE_USED_MIN_ROWS |          // for MEMORY table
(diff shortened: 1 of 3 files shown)


laurynas-biveinis (commented):
@georgelorchpercona, this LGTM, but up to you to review (I'd make the testcase use assert.inc, but that's a minor comment)

fungo:
@laurynas-biveinis thanks for your advice, the testcase has been updated using assert.inc.

storage/tokudb/ha_tokudb_alter_56.cc
```diff
@@ -503,6 +503,14 @@ enum_alter_inplace_result ha_tokudb::check_if_supported_inplace_alter(
...
                 result = HA_ALTER_INPLACE_EXCLUSIVE_LOCK;
             }
+        } else {
+            // other alter create options could be categorized in two types:
+            // a) options such as KEY_BLOCK_SIZE, BLOCK_FORMAT are not
+            //    supported by TokuDB
+            // b) options such as HA_CREATE_USED_COMMENT, HA_CREATE_USED_DEFAULT_CHARSET
+            //    do not need rebuild engine data
+            // and they all should be ignored
+            result = HA_ALTER_INPLACE_NO_LOCK_AFTER_PREPARE;
```
george-lorch: I would prefer that this is more explicit in checking explicitly for the cases where types can be ignored as new types can and will be added in the future, and will automatically be ignored rather than being handled (fully rebuilt) properly.

fungo: The current logic is try to keep same as InnoDB while handling `Alter_inplace_info::CHANGE_CREATE_OPTION` option, check [innobase_need_rebuild()](https://github.com/percona/percona-server/blob/5.6/storage/innobase/handler/handler0alter.cc#L207-L214).

I agree with you there should be more explicit, prudence is always good in code :)

I'll summit a updated version.

george-lorch: Thanks @fungo, we very much appreciate your contribution.

fungo: @georgelorchpercona happy to contribute and make TokuDB better :)

A updated version has been pushed.

The new patch checks every create option explicitly both in code and test case.

george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2113/

george-lorch:
See https://github.com/percona/percona-server/pull/2284 for GCA based branch

george-lorch:
@fungo thanks for the contribution. Closing this pr as I merged your change in via https://github.com/percona/percona-server/pull/2284 and https://github.com/percona/percona-server/pull/2285

PR #2228, opened 2018-03-13
TDB-137 : TokuDB 5.6 missing MRR tests

george-lorch:
- TokuDB 5.6 supports the Multi-Range Read optimization but has no tests to
  verify it. This adds the basic MRR test taken from mysql-test/t/inodb_mrr.test
  with some minor modifications.

### DIFF
--- a/mysql-test/suite/tokudb/r/tokudb_mrr.result
+++ b/mysql-test/suite/tokudb/r/tokudb_mrr.result
@@ -1,0 +1,706 @@
+set optimizer_switch='mrr=on,mrr_cost_based=off';
+set default_storage_engine=TokuDB;
+create table t1(a int);
+show create table t1;
+Table	Create Table
+t1	CREATE TABLE `t1` (
+  `a` int(11) DEFAULT NULL
+) ENGINE=TokuDB DEFAULT CHARSET=latin1
+insert into t1 values (0),(1),(2),(3),(4),(5),(6),(7),(8),(9);
+create table t2(a int);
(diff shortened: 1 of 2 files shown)


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2108/

laurynas-biveinis (changes requested):
How did you choose which of the MRR tests to adopt? These look like InnoDB-specific MRR tests, while there is include/mrr_tests.inc, which can be reused without copy-pasting. Please review `t/innodb_mrr.test`, `t/innodb_mrr_all.test`, `t/innodb_mrr_*.test`, `t/myisam_mrr*.test`, perhaps same should be adopted here.

george-lorch:
This does include mrr_all, plus it copy-paste from innodb specific to tokudb specific test that tests two very specific cases. All include the same basic test of mysql-test/include/mrr_tests.inc. Innodb only adds two specific coded tests that are changed from innodb to tokudb. I only did not add the mrr_cost test, which is not a bad idea but seems very redundant as it is more intentioned to test the optimizer behavior and not how engines react/respond as engine participation is identical. I can add it but IMHO it is a useless addition for this purpose and only extends runtime.

george-lorch:
As discussed on standup call, test is OK but one of the InnoDB specific additions is not necessary. I removed that but left the modified gap lock test in place as it does apply to TokuDB even though the gap lock behavior is slightly different. No new jenkins as this is just a removal of some test code and re-record. Verified it passes locally. Merging now as these changes are nneded/blocking the completion of 5.7/pre-8.0 work.

PR #2232, opened 2018-03-15
TDB-128 : Make .frm discovery optional at compile time

george-lorch:
- Based on idea from "rik prohaska <prohaska7@gmail.com>", make discovery
  functionality a compile time option to allow easier initial porting to 8.0
- Introduced new macro flag TOKU_INCLUDE_DISCOVER_FRM that when defined, causes
  all .frm discovery to be compiled in, otherwise it is compiled out.

### DIFF
--- a/storage/tokudb/ha_tokudb.cc
+++ b/storage/tokudb/ha_tokudb.cc
@@ -1645,7 +1645,8 @@ int ha_tokudb::initialize_share(const char* name, int mode) {
         goto exit;
     }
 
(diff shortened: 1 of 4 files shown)


george-lorch:
I don't have a 5.7 merge of this yet as I want to wait until the conditional compilation stuff lands there first to avoid a merge conflict.

https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2109/

laurynas-biveinis (approved):
Same comment as for TDB-139

PR #2235, opened 2018-03-15
PS 5.6 clang 6 fixes

dutow:


percona-ysorokin (approved):
LGTM with the same minor comments as in 5.5

dutow:
https://jenkins.percona.com/job/percona-server-5.6-param/2111/

There is one mtr test failing on jenkins, I tested that and it happens without these commits too

PR #2237, opened 2018-03-16
PS-3898: Include array in rdb_datadic

dutow:
It is used in this file, but never included. With libstdc++, the header works beacuse array is included internally by one of the other headers.


dutow:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1727/

george-lorch (approved):

peterschwaller:
touch

PR #2240, opened 2018-03-16
PS-3899: silence an uninitialized read in zlib

dutow:
According to the library developers, this is an optimization which while reads these bytes, doesn't use them for anything, and should be ignored.

(cherry picked from commit 2096be4b643e2047fa33c2dbd12c73d820fdae18)
(cherry picked from commit 43f99a0eacd2a6e0271482d79e544c9d518268d6)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -332,7 +332,7 @@ ENDIF()
 
 OPTION(WITH_MSAN "Enable memory sanitizer" OFF)
 IF(WITH_MSAN)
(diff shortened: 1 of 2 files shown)


dutow:
Compared to the changes for 5.5 and 5.6, in this I also added a reference to the blacklist file to the CMakeLists - as it wasn't part of the original options.

laurynas-biveinis (approved):

peterschwaller:
touch

PR #2245, opened 2018-03-19
Ps 5.7 tdb 128

george-lorch:


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1731/

george-lorch:
5.6 reviewed and merged already. Passed jenkins so self approving and merging.

PR #2251, opened 2018-03-20
PS-3901 (5.6): Fix -DWITHOUT_xxx_STORAGE_ENGINE=ON for some plugins

inikep:

PR #2250, opened 2018-03-20
PS-3901 (5.5): Fix -DWITHOUT_xxx_STORAGE_ENGINE=ON for some plugins

inikep:

PR #2252, opened 2018-03-20
PS-3901 (5.7): Fix -DWITHOUT_xxx_STORAGE_ENGINE=ON for some plugins

inikep:


inikep:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1734/

laurynas-biveinis (changes requested):
Same as for #2250 / #2251, and please also check whether 5.7 actually needs the InnoDB change, because InnoDB is a mandatory storage engine starting with 5.7.

inikep:
The commit message is updated. InnoDB doesn't need that change but I left it for symmetry and easier merges.

laurynas-biveinis:
We should not have "no-op" patches. It makes merges easier from 5.6 to 5.7 but harder from 5.7 to 8.0, which has higher weight.

inikep:
I removed the `no-op` part.

laurynas-biveinis (commented):
Same comment as for #2251

laurynas-biveinis (approved):
Please upstream and PR to 8.0 seed branch as appropriate

inikep:
For 5.7 `PERFSCHEMA` is `MANDATORY` and `PARTITION` is protected from using both `WITH_PARTITION_STORAGE_ENGINE=ON` and `WITHOUT_PARTITION_STORAGE_ENGINE=ON`.
This PR is probably a no-op (I didn't check all plugins for 5.7).

PR #2255, opened 2018-03-21
TDB-148 : Make UPSERT/NOAR optional at compile time

george-lorch:
- UPSERT/NOAR functionality requires server side patches to funvtion correctly.
  For 8.0, these initially will not be present, and thus we will have a bunch
  of dead-code present, and get compiler warning/errors from -Wunused-function.
- Added in proper #if defined(TOKU_INCLUDE_UPSERT && TOKU_INCLUDE_UPSERT around
  all new functionality.


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1736/

laurynas-biveinis (approved):

PR #2256, opened 2018-03-22
TDB-128 : Make .frm discovery optional at compile time

george-lorch:
- The original patch that introduced TOKU_INCLUDE_WRITE_FRM_DATA was incorrect
  and incomplete. It failed to properly isolate all instances of storing .frm
  data within TokuDB status (metadata) files and so in all cases, TokuDB would
  perform this regardless of the definition of the macro.
- The second patch which introduced TOKU_INCLUDE_DISCOVER_FRM was also incorrect
  in that it extended beyond just guarding the functionality of using the stored
  .frm data to provide MySQL table discovery functionality and guarded some, but
  not all, of the 'WRITE' functionaloty.
- This patch corrects all of these incorrect behaviors and properly isolated the
  two different, but related areas of 'WRITE' and 'DISCOVER'.
- Due to the timing and logistics of merging this within the GCA workflow, it is
  not GCA but will be a cherry-pick to 5.7

### DIFF
--- a/storage/tokudb/ha_tokudb.cc
+++ b/storage/tokudb/ha_tokudb.cc
@@ -1631,7 +1631,7 @@ int ha_tokudb::initialize_share(const char* name, int mode) {
         goto exit;
     }
 
-#if defined(TOKU_INCLUDE_DISCOVER_FRM) && TOKU_INCLUDE_DISCOVER_FRM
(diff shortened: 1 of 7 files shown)


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2130/

laurynas-biveinis (approved):

PR #2260, opened 2018-03-23
PS-3921: Enable RocksDB and TokuDB in Travis config for PS 5.7

inikep:
Set default values in Travis config to:
    - WITHOUT_TOKUDB=OFF
    - WITHOUT_ROCKSDB=OFF


.travis.yml
```diff
@@ -17,8 +17,8 @@ env:
     - CCACHE_MAXSIZE=500Mi
     - CCACHE_CPP2=1
     - MAINTAINER_MODE=ON
-    - WITHOUT_TOKUDB=ON
-    - WITHOUT_ROCKSDB=ON
+    - WITHOUT_TOKUDB=OFF
```
laurynas-biveinis, line 20: Maybe inverted WITH_TOKUDB=ON WITH_ROCKSDB=ON would be more readable?

laurynas-biveinis (approved):
LGTM with or without the suggested change

.travis.yml
inikep: I will change it in the future because it should be done staring from 5.6.

PR #2263, opened 2018-03-26
PS-3925: Use strncmp instead of memcmp in log_in_use (5.5)

dutow:


dutow:
https://jenkins.percona.com/job/percona-server-5.5-param/

laurynas-biveinis (approved):

PR #2271, opened 2018-03-28
PS-3928: Fix for complex case insensitive full text queries

dutow:
These queries tried to validate the result by matching it back to the original result, converted to lower case form.
As queries using binary collections weren't converted to lowercase, these validations failed when the query contained any upper case characters.
This caused ngram queries which contained upper case characters and were longer than the ngram length to return no results.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -7173,7 +7173,9 @@ innobase_fts_text_case_cmp(
 	const fts_string_t*	s2 = (const fts_string_t*) p2;
 	ulint			newlen;
 
-	my_casedn_str(charset, (char*) s2->f_str);
+	if (!my_binary_compare(charset)) {
+		my_casedn_str(charset, (char*) s2->f_str);
+	}
 
 	newlen = strlen((const char*) s2->f_str);
 
--- a/mysql-test/suite/innodb_fts/r/bug78048.result
+++ b/mysql-test/suite/innodb_fts/r/bug78048.result
@@ -1,0 +1,48 @@
+create table `ngram_simple` (
+`i` int(11) not null auto_increment,
+`txt` text collate utf8mb4_bin not null,
+primary key (`i`),
+fulltext key `fx_txts` (`txt`) ) engine=InnoDB auto_increment=10 default charset=utf8mb4 collate=utf8mb4_bin;
+insert into ngram_simple (txt) values ('CompP&C01');
+insert into ngram_simple (txt) values ('CompP&C02');
+insert into ngram_simple (txt) values ('CompP&C03');
+insert into ngram_simple (txt) values ('CompP&C04');
+insert into ngram_simple (txt) values ('CompP&C05');
+insert into ngram_simple (txt) values ('CompP&C06');
+insert into ngram_simple (txt) values ('CompP&c04');
+insert into ngram_simple (txt) values ('abc*efg');
+insert into ngram_simple (txt) values ('abc&efg');
+insert into ngram_simple (txt) values ('abC&efGh');
+select * from ngram_simple where match(txt) against ('abc' in boolean mode);
+i	txt
+17	abc*efg
+18	abc&efg
+select * from ngram_simple where match(txt) against ('abC' in boolean mode);
+i	txt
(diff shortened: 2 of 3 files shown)


dutow:
Jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1741/

mysql-test/suite/innodb_fts/t/percona_bug_ps3928.test
```diff
@@ -0,0 +1,49 @@
...
+alter table ngram_simple drop key fx_txts;
+alter table ngram_simple add fulltext key `fx_txts` (`txt`) with parser ngram;
+optimize table ngram_simple;
+
+select * from ngram_simple where match(txt) against ('abc' in boolean mode);
+select * from ngram_simple where match(txt) against ('abC' in boolean mode);
+select * from ngram_simple where match(txt) against ('C04' in boolean mode);
+select * from ngram_simple where match(txt) against ('c04' in boolean mode);
+
+alter table ngram_simple drop key fx_txts;
```
percona-ysorokin: Hasn't this (non-ngram case) already been tested when the table was created?

dutow: It was. I only kept it in the test because this is how the original test case in the ticket / mysql bug report looked like, it probably doesn't test anything new.

percona-ysorokin (approved):
LGTM

storage/innobase/fts/fts0que.cc
```diff
@@ -4096,7 +4096,7 @@ fts_query(
 	/* For binary collations, a case sensitive search is
 	performed. Hence don't convert to lower case. */
 	if (my_binary_compare(charset)) {
-	memcpy(lc_query_str, query_str, query_len);
+		memcpy(lc_query_str, query_str, query_len);
```
laurynas-biveinis: Is this our code? If not, then just drop (OK to include in eventual OCA submission)

storage/innobase/handler/ha_innodb.cc
```diff
@@ -7173,7 +7173,9 @@ innobase_fts_text_case_cmp(
 	const fts_string_t*	s2 = (const fts_string_t*) p2;
 	ulint			newlen;
 
-	my_casedn_str(charset, (char*) s2->f_str);
+	if(!my_binary_compare(charset)) {
```
laurynas-biveinis: Space after if

mysql-test/suite/innodb_fts/t/percona_bug_ps3928.test
```diff
@@ -0,0 +1,49 @@
+# Bug PS-3928: Complex case insensitive full text queries returned no results
```
laurynas-biveinis: Use upstream bug id for testcase naming and in the comments inside (see Confluence)

mysql-test/suite/innodb_fts/t/percona_bug_ps3928.test
```diff
@@ -0,0 +1,49 @@
...
+alter table ngram_simple add fulltext key `fx_txts` (`txt`) with parser ngram;
+optimize table ngram_simple;
+
+select * from ngram_simple where match(txt) against ('abc' in boolean mode);
+select * from ngram_simple where match(txt) against ('abC' in boolean mode);
+select * from ngram_simple where match(txt) against ('C04' in boolean mode);
+select * from ngram_simple where match(txt) against ('c04' in boolean mode);
+
+alter table ngram_simple drop key fx_txts;
+alter table ngram_simple add fulltext key `fx_txts` (`txt`);
```
laurynas-biveinis: Same question Yura had
(now this extra testing seems to be completely superfluous)

laurynas-biveinis (changes requested):

storage/innobase/fts/fts0que.cc
dutow: No, it's part of the previous fix for the issue.

dutow:
Updated.
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1747/

laurynas-biveinis (approved):
Please upstream / 8.0 seed

PR #2273, opened 2018-03-28
PS-3773 : LP #1744430: "Incorrect key file for table" frequently for …

george-lorch:
…tokudb

- Added new logging when an index inconsistency is discovered between any SK
  and the PK, specifically when a key in an SK can not find the corresponding
  record in the PK.
- Added DBUG_EXECUTE_IF point to fake return a HA_ERR_CRASHED/ER_NOT_KEYFILE from
  various index lookup calls and implemented basic test to ensure new logging
  actualy occurs.

### DIFF
--- a/mysql-test/suite/tokudb.bugs/t/PS-3773.test
+++ b/mysql-test/suite/tokudb.bugs/t/PS-3773.test
@@ -1,0 +1,26 @@
+--source include/have_tokudb.inc
+--source include/have_debug.inc
+
+--let SEARCH_FILE=$MYSQLTEST_VARDIR/tmp/tokudb.bugs.PS-3773.log
+--let $restart_parameters="restart: --log-error=$SEARCH_FILE"
+--source include/restart_mysqld.inc
+
+CREATE TABLE t1(a INT, b INT, c INT, PRIMARY KEY(a), KEY(b)) ENGINE=TokuDB;
+SET tokudb_auto_analyze=0;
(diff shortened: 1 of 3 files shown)


george-lorch:
Jenkins https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2134/
Two test failures were due to me not placing the most obvious have_tokudb.inc in the test which I later added and pushed.

ps : will follow with a 5.7 PR one this is approved in case there is any discussion.

mysql-test/suite/tokudb.bugs/t/PS-3773.test
```diff
@@ -0,0 +1,26 @@
...
+SET GLOBAL debug = "+d,tokudb_fake_db_notfound_error_in_read_full_row";
+--error ER_NOT_KEYFILE
+SELECT * FROM t1 WHERE b = 2;
+
+DROP TABLE t1;
+
+--let SEARCH_PATTERN=ha_tokudb::read_full_row on table
+--source include/search_pattern_in_file.inc
+
+--remove_file $SEARCH_FILE
```
laurynas-biveinis: Better to avoid removing in-use file - remove after restart instead

laurynas-biveinis (changes requested):

george-lorch:
Done...

laurynas-biveinis (approved):

PR #2277, opened 2018-03-30
TDB-110 : Prepare TokuDB for 8.0

george-lorch:
TDB-118 : PerconaFT needs -fexceptions flag for ftcxx to compile under 8.0
TDB-84 : PerconaFT fix -Werror=undef
TDB-85 : PerconaFT fails to build in 8.0 due use of MYSQL_VERSION_* that is actually not passed down
TDB-172 : PerconaFT needs different libraries to link to 8.0 debug sync and PFS

- Advancing the git submodule commit pointer to pick up these changes from FT.
- This branch will be cherry-picked forward to 5.7


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2138/

laurynas-biveinis (approved):
Not sure what's there to review in a PR like that, should be self-approved

george-lorch:
True @laurynas-biveinis, I was thinking the same thing but did not want to appear to 'sneak something in'. It is possible though that there could be a problem in that the commit pointer isn't advancing to the proper/intended commit, or that there is some issue in one of the jenkins jobs related to this new combination of TokuDB + FT that was not addressed correctly, etc... Just being a bit prudent here is all.

PR #2279, opened 2018-03-30
TDB-141 : Fix various places where macros are redefined or used when …

george-lorch:
…undefined

- Fixed up some macros to use #if defined(xxx) &&xxx semantics with #endif comment


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2136/

laurynas-biveinis (approved):

PR #2284, opened 2018-04-02
TDB-130: alter TokuDB table comment rebuild whole engine data

george-lorch:
Alter TokuDB table comment should not using copying algorithm and
rebuild whole engine data, only FRM definition should be changed.

Make `ha_tokudb::check_if_supported_inplace_alter()` return
`HA_ALTER_INPLACE_EXCLUSIVE_LOCK` while altering comment, and mysql
server will invoke `ha_tokudb::inplace_alter_table` which will do
nothing for altering comment.

Besides table comment, there are many other create options that
should be ignored by TokuDB, and they are all handled in this patch.

### DIFF
--- a/storage/tokudb/ha_tokudb_alter_56.cc
+++ b/storage/tokudb/ha_tokudb_alter_56.cc
@@ -248,6 +248,40 @@ static bool only_flags(ulong bits, ulong mask) {
     return (bits & mask) != 0 && (bits & ~mask) == 0;
 }
 
+// Table create options that should be ignored by TokuDB
+// There are 25 total create options defined by mysql server (see handler.h),
+// and only 4 options will touch engine data, either rebuild engine data or
+// just update meta info:
+//   1. HA_CREATE_USED_AUTO        update auto_inc info
+//   2. HA_CREATE_USED_CHARSET     rebuild table if contains character columns
+//   3. HA_CREATE_USED_ENGINE      rebuild table
+//   4. HA_CREATE_USED_ROW_FORMAT  update compression method info
+//
+// All the others are either not supported by TokuDB or no need to
+// touch engine data.
+static constexpr uint32_t TOKUDB_IGNORED_ALTER_CREATE_OPTION_FIELDS =
+    HA_CREATE_USED_RAID |              // deprecated field
+    HA_CREATE_USED_UNION |             // for MERGE table
+    HA_CREATE_USED_INSERT_METHOD |     // for MERGE table
+    HA_CREATE_USED_MIN_ROWS |          // for MEMORY table
+    HA_CREATE_USED_MAX_ROWS |          // for NDB table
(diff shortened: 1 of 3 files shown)


george-lorch:
This is a GCA based branch of https://github.com/percona/percona-server/pull/2226

https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2113/

storage/tokudb/ha_tokudb_alter_56.cc
```diff
@@ -248,6 +248,40 @@ static bool only_flags(ulong bits, ulong mask) {
     return (bits & mask) != 0 && (bits & ~mask) == 0;
 }
 
+// Table create options that should be ignored by TokuDB
+// There are totally 25 create options defined by mysql server(see handler.h),
```
laurynas-biveinis: Space after "server", and I think "are totally 25" -> "are 25 ... in total" but George should know better

storage/tokudb/ha_tokudb_alter_56.cc
```diff
@@ -248,6 +248,40 @@ static bool only_flags(ulong bits, ulong mask) {
...
+// and only 4 options will touch engine data, either rebuild engine data or
+// just update meta info:
+//   1. HA_CREATE_USED_AUTO        update auto_inc info
+//   2. HA_CREATE_USED_CHARSET     rebuild table if contains character columns
+//   3. HA_CREATE_USED_ENGINE      rebuild table
+//   4. HA_CREATE_USED_ROW_FORMAT  update compression method info
+//
+// All the others are either not supported by TokuDB or no need to
+// touch engine data.
+static const uint32_t TOKUDB_IGNORED_ALTER_CREATE_OPTION_FIELDS =
```
laurynas-biveinis: constexpr

laurynas-biveinis (approved):
Minor comments, not sure if worth a respin

george-lorch:
yeah, I think constexpr is worth a touch up and will take care of the comment along the way.

storage/tokudb/ha_tokudb_alter_56.cc
```diff
@@ -248,6 +248,40 @@ static bool only_flags(ulong bits, ulong mask) {
     return (bits & mask) != 0 && (bits & ~mask) == 0;
 }
 
+// Table create options that should be ignored by TokuDB
+// There are 25 total create options defined by mysql server(see handler.h),
```
laurynas-biveinis: Still no space between "server" and "(", but, meh :)

laurynas-biveinis (approved):

storage/tokudb/ha_tokudb_alter_56.cc
george-lorch: Gah, sorry about that, I (obviously) missed that part of your comment. It's not a problem, I'll fix it and re-push after a local mtr cycle, then merge without needing more review.

PR #2286, opened 2018-04-03
PS-235: query_response_time_session_stats variable (5.6)

dutow:


laurynas-biveinis (approved):

plugin/query_response_time/tests/mtr/session_toggle.test
```diff
@@ -0,0 +1,106 @@
...
+--echo # Tests for query_response_time_session_stats variable
+--echo #
+
+--source include/count_sessions.inc
+
+SET @saved_query_response_time_stats = @@GLOBAL.query_response_time_stats;
+SET @saved_query_response_time_flush = @@GLOBAL.query_response_time_flush;
+SET GLOBAL query_response_time_flush= 'ON';
+
+SELECT SUM(COUNT) FROM INFORMATION_SCHEMA.QUERY_RESPONSE_TIME;
```
percona-ysorokin, line 13: Here and everywhere, please replace direct `SELECT SUM(COUNT) FROM INFORMATION_SCHEMA.QUERY_RESPONSE_TIME` with
```
--let $assert_text= ...
--let $assert_cond="[SELECT SUM(COUNT) FROM INFORMATION_SCHEMA.QUERY_RESPONSE_TIME]" = XXX
--source include/assert.inc
```
specifying expected value XXX explicitly.

percona-ysorokin (changes requested):
Requested changes in MTR test case only.
Jenkins re-run with single test case is OK.

plugin/query_response_time/tests/mtr/session_toggle.test
dutow: I tried this, but looks like that the assert include itself executes a large, variable number of queries. With that, instead of simple comments and easy to follow numbers we'll get rapidly increasing counts, which are quite impossible to follow.

In the current version, I can see why the numbers start as 0/2/3. With the asserts, it would be 0/8/36, based on the  implementation of the include files, which could possibly change with unrelated commits.

percona-ysorokin: Oh, did not think about that. Changing to approved.

percona-ysorokin (approved):
LGTM

dutow:
Jenkins: https://jenkins.percona.com/job/percona-server-5.6-param/2141/

PR #2290, opened 2018-04-04
TDB-115 : Fix -Wunused-parameter

george-lorch:
- Many places where API is pre-defined either by FT or by server and parameters
  are unused needed annotation with TOKUDB_UNUSED.
- Many internally used functions had extra, unused parameters added and were
  cleaned up to remove the parameters rather than proliferating TOKUDB_UNUSED.


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2142/

laurynas-biveinis (approved):

PR #2294, opened 2018-04-05
PS-3919: Assertion when analyze table performs a binlog rotation

dutow:
This commit:
* Adds a minimal testcase
* Backports the fix for this issue from 8.0, which can be found as part of two huge commits:
 * Disabling autocommit/analyze read lock for mysql_admin_table from 0a10fc8b5ca0
 * Disable_autocommit_guard RAII class from 31350e8ab151 (used by the otherr backported change)


mysql-test/t/percona_bug_ps3919-master.opt
```diff
@@ -0,0 +1 @@
+--log_bin=binlog --gtid_mode=ON --enforce_gtid_consistency=ON --sql_mode=ONLY_FULL_GROUP_BY
```
laurynas-biveinis: Binlog suite?

laurynas-biveinis: Why does it need sql_mode?

sql/sql_admin.cc
```diff
@@ -281,6 +281,13 @@ static bool mysql_admin_table(THD* thd, TABLE_LIST* tables,
                                                             HA_CHECK_OPT *),
                               int (view_operator_func)(THD *, TABLE_LIST*))
 {
+  /*
+    Prevent InnoDB from automatically committing InnoDB
+    transaction each time data-dictionary tables are closed after
```
laurynas-biveinis: This comment is not correct for 5.7

sql/sql_admin.cc
```diff
@@ -1211,6 +1220,13 @@ bool Sql_cmd_analyze_table::execute(THD *thd)
   if (check_table_access(thd, SELECT_ACL | INSERT_ACL, first_table,
                          FALSE, UINT_MAX, FALSE))
     goto error;
+
+  DBUG_EXECUTE_IF("simulate_analyze_table_lock_wait_timeout_error",
```
laurynas-biveinis: Not used by the testcase

laurynas-biveinis (changes requested):
Does the testcase pass on (either our or upstream) 8.0 branch as expected?

mysql-test/t/percona_bug_ps3919-master.opt
dutow: Moved it there.

dutow: I'm not sure - but without that parameter, the crash didn't happen. The original bug report mentioned several other parameters, and I was able to remove most of them, but not this one.

sql/sql_admin.cc
dutow: Removed it.

sql/sql_admin.cc
dutow: Removed it.

dutow:
For 8.0: yes, I tested the test case on our 8.0 branch, and it passed - I backported the changed from 8.0 after noticing that.

dutow:
Jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1809/

laurynas-biveinis (approved):

PR #1462, opened 2017-03-02
BLD-591 - ps-admin.sh script for RocksDB/TokuDB and plugins easy install

tplavcic:
**Ticket:** BLD-591

What was done:
- changed some specific options from ps_tokudb_admin like
`--enable` -> `--enable-tokudb,` `--enable-backup` -> `--enable-tokubackup` and similar, and also changed variable names to not be tokudb specific
- added setup of RocksDB
- added setup of plugins: Audit Log, PAM, Query Response Time, MYSQLX
- added `--config-file` option
- added into CMakeLists.txt that the script is installed only if built with -DWITH_ROCKSDB since currently we don't want this script to be shipped

**TEST:**
**RocksDB:**
```
plavi@zoidberg  ~/sandboxes/binaries/5.7.17/bin  sudo ./percona_server_admin --enable-rocksdb --user=msandbox --password=msandbox --socket=/tmp/mysql_sandbox5717.sock                                                            
Checking RocksDB engine plugin status...
INFO: RocksDB engine plugin is not installed.

Installing RocksDB engine...
INFO: Successfully installed RocksDB engine plugin.

mysql [localhost] {msandbox} ((none)) > show engines;
+--------------------+-----------+----------------------------------------------------------------------------+----------------+--------+--------------+
| Engine             | Support   | Comment                                                                    | Transactions   | XA     | Savepoints   |
|--------------------+-----------+----------------------------------------------------------------------------+----------------+--------+--------------|
| ROCKSDB            | YES       | RocksDB storage engine                                                     | YES            | YES    | YES          |
+--------------------+-----------+----------------------------------------------------------------------------+----------------+--------+--------------+

+-----------------------------+----------+--------------------+---------------+-----------+
| Name                        | Status   | Type               | Library       | License   |
|-----------------------------+----------+--------------------+---------------+-----------|
| ROCKSDB                     | ACTIVE   | STORAGE ENGINE     | ha_rocksdb.so | GPL       |
| ROCKSDB_CFSTATS             | ACTIVE   | INFORMATION SCHEMA | ha_rocksdb.so | GPL       |
| ROCKSDB_DBSTATS             | ACTIVE   | INFORMATION SCHEMA | ha_rocksdb.so | GPL       |
| ROCKSDB_PERF_CONTEXT        | ACTIVE   | INFORMATION SCHEMA | ha_rocksdb.so | GPL       |
[...]

PR #2296, opened 2018-04-05
Fix for DOC-736, renaming ps_tokudb_admin and its options

fiowro:
(based on info from https://github.com/percona/percona-server/pull/1462)


george-lorch:
@tplavcic can you review for accuracy?

tplavcic (approved):
LGTM

PR #2298, opened 2018-04-05
Ps 5.7 tdb 110

george-lorch:


george-lorch:
This PR has a few extra committs after the 5.6 merge of TDB-183 ans TDB-141. Each commit removes some section of dead code that would have been previously ifdef'ed out for either MySQL 5.5 or MariaDB.

https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1763/

laurynas-biveinis (approved):

PR #2306, opened 2018-04-08
Fix for DOC-658 mentioning that innodb-force-recovery>0 prevents…

fiowro:
… to modify system tables

### DIFF
--- a/doc/source/reliability/innodb_corrupt_table_action.rst
+++ b/doc/source/reliability/innodb_corrupt_table_action.rst
@@ -31,6 +31,14 @@ see `Forcing InnoDB Recovery
 <https://dev.mysql.com/doc/refman/5.5/en/forcing-innodb-recovery.html>`_
 from the MySQL Reference Manual.
 
+.. note::
+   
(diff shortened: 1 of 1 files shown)


laurynas-biveinis (changes requested):
I am not sure where would be the right place to document this. innodb_corrupt_table_action is not necessarily the one - innodb_corrupt_table_action and innodb_force_recovery are two different things, and the use cases described (MyRocks installation/uninstallation etc.) will not be found through innodb_corrupt_table_action documentation.

IMHO in the present state/scope of our manual it does not really belong here, but a bug should be filed upstream to improve their docs.

fiowro:
I'd say this file is currently the only place in docs mentioning innodb_force_recovery.

laurynas-biveinis:
Yes, but the fact that it's the only place is not sufficient to make it the right place. How would someone trying to install a plugin with innodb_force_recovery consider innodb_corrupt_table_action docs?

laurynas-biveinis:
What's the status of this?

PR #2307, opened 2018-04-11
TDB-188 : TDB-114 (Change use of MySQL HASH to unordered_map) introdu…

george-lorch:
…ces memory leak

- TDB-188 eliminated the HASH callback that releases a TOKUDB_SHARE without
  adding that logic back into the drop_share method. Added destruction logic
  back.
- Since this commit is fixing an issue that has not yet been released, the GCA
  position is behind the place where the issue was introduced, therefore, this
  commit is being made on the HEAD and will be cherry picked forward to 5.7.


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2150/

laurynas-biveinis (approved):
The commit message probably meant TDB-114, not TDB-188 in the first list item

george-lorch:
Gah, yes, you are correct. I will re-write the commit message before merging.

PR #2308, opened 2018-04-11
TDB-188 : TDB-114 (Change use of MySQL HASH to unordered_map) introdu…

george-lorch:
…ces memory leak

- TDB-188 eliminated the HASH callback that releases a TOKUDB_SHARE without
  adding that logic back into the drop_share method. Added destruction logic
  back.
- Since this commit is fixing an issue that has not yet been released, the GCA
  position is behind the place where the issue was introduced, therefore, this
  commit is being made on the HEAD and will be cherry picked forward to 5.7.

(cherry picked from commit ef33a418371017cbe745ceafa0dfe7f2ff99c175)


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1770/

laurynas-biveinis (approved):

PR #2310, opened 2018-04-11
PS-3940: key_block_size shouldn't be set automatically for the heap engine (5.5)

dutow:
Key_block_size was set automatically by the improved heap storage engine, which caused:
* warnings (5.5/5.6) or errors (5.7) when changing the engine type to innodb
* constantly growing key_block_size during alter operations

With this change, key_block_size won't be set unless the user specifies a custom value.
* switching to innodb will only result in an error if the user manually specified a key_block_size before, which is expected.
* key_block_size was growing because max_chuck_size is set based on the specified key_block_size, and then key_block_size was set to max_chunk_size plus an overhead value, increasing it every time.
  by leaving out the second step, it remains stable.

Note that key_block_size was set so the chunk size is visible even when calculated automatically.
As it is always set to 256 unless the user specifies something else, this didn't provide too much information.

### DIFF
--- a/mysql-test/t/percona_heap_var.test
+++ b/mysql-test/t/percona_heap_var.test
@@ -4,57 +4,56 @@
 --source include/have_innodb.inc
 
 --disable_warnings
-drop table if exists t1;
+DROP TABLE IF EXISTS t1;
 --enable_warnings
 
 set @@session.max_heap_table_size=16*1024*1024;
 
 --error 1234
-create table t1 (a int not null, b varchar(400), c int, primary key (a), key (c)) engine=heap comment="testing heaps" key_block_size=128;
+CREATE TABLE t1 (a INT NOT NULL, b VARCHAR(400), c INT, PRIMARY KEY (a), KEY (c)) ENGINE=HEAP COMMENT="testing heaps" KEY_BLOCK_SIZE=128;
 
 --error 1234
-create table t1 (a int not null, b int, c varchar(400), primary key (a), key (b)) engine=heap comment="testing heaps" key_block_size=4;
+CREATE TABLE t1 (a INT NOT NULL, b INT, c VARCHAR(400), PRIMARY KEY (a), KEY (b)) ENGINE=HEAP COMMENT="testing heaps" KEY_BLOCK_SIZE=4;
 
-create table t1 (a int not null, b int, c varchar(400), d varchar(400), primary key (a), key (b)) engine=heap comment="testing heaps" key_block_size=24;
-
---replace_column 6 X 7 X 8 X 9 X 10 X 11 X 12 X 13 X 14 X
---query_vertical show table status like "t1"
-
-insert into t1 values (1,1,'012',NULL), (2,2,'0123456789',NULL), (3,3,'012345678901234567890123456789',NULL), (4,4,NULL,'0123456789012345678901234567890123456789012345678901234567890123456789');
-select * from t1;
-
-delete from t1 where a = 3;
-select * from t1;
-
-insert into t1 values (5,5,NULL,'0123'), (6,6,NULL,'0123');
-select * from t1;
-
-update t1 set c = '012345678901234567890123456789' where a = 2;
-select * from t1;
-
-update t1 set c = '0123456789' where a = 2;
-select * from t1;
-
-insert into t1 values (7,7,'0123',NULL), (8,8,'0123',NULL);
-select * from t1;
+CREATE TABLE t1 (a INT NOT NULL, b INT, c VARCHAR(400), d VARCHAR(400), PRIMARY KEY (a), KEY (b)) ENGINE=HEAP COMMENT="testing heaps" KEY_BLOCK_SIZE=24;
 
 --replace_column 6 X 7 X 8 X 9 X 10 X 11 X 12 X 13 X 14 X
---query_vertical show table status like "t1"
-alter table t1 key_block_size = 0;
---replace_column 6 X 7 X 8 X 9 X 10 X 11 X 12 X 13 X 14 X
---query_vertical show table status like "t1"
-alter table t1 row_format = dynamic;
---replace_column 6 X 7 X 8 X 9 X 10 X 11 X 12 X 13 X 14 X
---replace_regex /KEY_BLOCK_SIZE=[[:digit:]]+/KEY_BLOCK_SIZE=X/
---query_vertical show table status like "t1"
-alter table t1 key_block_size = 128, max_rows = 10001;
---replace_column 6 X 7 X 8 X 9 X 10 X 11 X 12 X 13 X 14 X
---query_vertical show table status like "t1"
+--query_vertical SHOW TABLE STATUS LIKE "t1"
 
-select * from t1;
+INSERT INTO t1 VALUES (1,1,'012',NULL), (2,2,'0123456789',NULL), (3,3,'012345678901234567890123456789',NULL), (4,4,NULL,'0123456789012345678901234567890123456789012345678901234567890123456789');
+SELECT * FROM t1;
 
-delete from t1;
-select * from t1;
+DELETE FROM t1 WHERE a = 3;
+SELECT * FROM t1;
+
+INSERT INTO t1 VALUES (5,5,NULL,'0123'), (6,6,NULL,'0123');
+SELECT * FROM t1;
+
+UPDATE t1 SET c = '012345678901234567890123456789' WHERE a = 2;
+SELECT * FROM t1;
+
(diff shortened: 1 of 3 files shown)


mysql-test/t/percona_bug_ps3940.test
```diff
@@ -0,0 +1,17 @@
+# PS-3940: Warnings/errors when switching from the memory engine to innodb
```
laurynas-biveinis: No need to add a new test file - extend the most suitable one of percona_heap_* tests

sql/sql_table.cc
```diff
@@ -6820,6 +6820,14 @@ bool mysql_alter_table(THD *thd,char *new_db, char *new_name,
   else
     create_info->data_file_name=create_info->index_file_name=0;
 
+  if (old_db_type != new_db_type && old_db_type->db_type == DB_TYPE_HEAP)
```
laurynas-biveinis: Looks like leaky abstraction here - is it possible to make the change in the MEMORY SE layer instead?

laurynas-biveinis (changes requested):

sql/sql_table.cc
dutow: I tried to do that in a few different ways, but couldn't do it. It would require a before/during alter hook in the (current) storage engine (so the memory se would be able to change the key_block_size when the engine is changed), and there isn't such a thing in the code - only the validation step, with the new storage engine, where innodb raises the warnings/errors.

laurynas-biveinis: Ack, then only the testcase comment needs addressing

dutow: Correction: looks like the change which is required for PS-3943 also fixes this issue, without this change here. I updated my commit based on that, and I'll also update the test cases tomorrow.

This updated version solves both PS-3940 and PS-3943.

dutow:
II also want to mention that we could do this in a different way too:

* keeping my original change for PS-3940 (resetting the key_block_size setting when changing the engine type)
* and instead of deleting the lines which set the key_block_size, correcting them to use the value without the additional overhead, essentially stabilizing the value for PS-3943.

That was my original solution, and I wanted to submit the second part as another pull request, but I realized that this is also doable, simpler, and has a more logical behavior.

dutow:
Test case updated: I moved the 3 new tests into the percona_heap_var mtr test.

laurynas-biveinis:
Is it ready for review?

dutow:
It is - after my last comment I only rebased it to the latest 5.5 head, so it doesn't conflict with another heap storage change.

mysql-test/t/percona_heap_var.test
```diff
@@ -165,3 +165,44 @@ ALTER TABLE t1 ADD PRIMARY KEY (a, c);
 --error ER_TABLE_CANT_HANDLE_BLOB
 ALTER TABLE t1 ADD INDEX idx (c, a);
 DROP TABLE t1;
+
+create table `t1` (
```
laurynas-biveinis: Please keep the SQL keyword case consistent throughout the testcase

mysql-test/t/percona_heap_var.test
```diff
@@ -165,3 +165,44 @@ ALTER TABLE t1 ADD PRIMARY KEY (a, c);
...
+  `id` int(11) not null auto_increment,
+  `json_column` text,
+  primary key (`id`)
+) engine=memory;
+show create table t1;
+
+# PS-3940: Engine can be changed to innodb and back
```
laurynas-biveinis:
```
--echo #
--echo # PS-3940: Foo bar ...
--echo #
```
so that output sections are clearly separated in result file too

mysql-test/t/percona_heap_var.test
```diff
@@ -165,3 +165,44 @@ ALTER TABLE t1 ADD PRIMARY KEY (a, c);
...
+alter table  t1 engine=innodb;
+alter table  t1 engine=memory;
+
+# PS-3943: Adding additional columns won't change the key_block size
```
laurynas-biveinis:
```
--echo #
--echo # PS-3943: Foo bar ...
--echo #
```
so that output sections are clearly separated in result file too

mysql-test/t/percona_heap_var.test
```diff
@@ -165,3 +165,44 @@ ALTER TABLE t1 ADD PRIMARY KEY (a, c);
...
+alter table t1 modify json_column blob;
+alter table t1 modify json_column text;
+show create table t1;
+
+alter table t1 modify json_column blob;
+alter table t1 modify json_column text;
+show create table t1;
+
+alter table t1 modify json_column blob;
```
laurynas-biveinis: Why 3x ?

mysql-test/t/percona_heap_var.test
```diff
@@ -165,3 +165,44 @@ ALTER TABLE t1 ADD PRIMARY KEY (a, c);
...
+alter table t1 modify json_column text;
+show create table t1;
+
+drop table t1;
+
+# PS-3943: Explicit key_block_size is kept and doesn't change
```
laurynas-biveinis: Also 3943?

laurynas-biveinis (changes requested):
The source change looks correct, but not necessarily complete: I disagree with the reasoning "* switching to innodb will only result in an error if the user manually specified a key_block_size before, which is expected." -

I don't think this is expected because key_block_size has different semantics in different SEs. If someone specified key_block_size for MEMORY SE, it's purely for MEMORY reasons - to specify the size of dynamic block. If one is to change the engine of such table, the old KEY_BLOCK_SIZE loses its meaning on the change. 

Is there a way to change the engine then with this patch then?

laurynas-biveinis:
Please also list all the issues fixed in the commit message header ("PS-X PS-Y PS-Z several MEMORY bugs fixed")

dutow:
It is, I'll update the test case to include an example.

Basically if the user specified a key_block_size for the memory se, he has to specify key_block_size=0, or a valid value when switching to innodb - if he does that, it works. That's also how it worked before, the only difference is that now he doesn't have to do when he didn't specify it manually.

(Or more specifically, it works with a warning even without this in 5.5/5.6. It results in an error only in 5.7)

mysql-test/t/percona_heap_var.test
dutow: The file is already inconsistent, upper case characters are only used by the tests for bug173483, added recently - and I followed the way the older parts looks like, as I originally added this change at the gca point.

Should I change the first or the latter part of the test?

laurynas-biveinis: Oh, my fault then. I'd prefer uppercase then

mysql-test/t/percona_heap_var.test
dutow: Without my change, the original create statement didn't have a key_block_size. After the first alters, a correctly sized key_block_size (256) appeared, and it increased with the later alters - so it was different each time.

laurynas-biveinis: Right, then please add comments "with the bug present, key_block_size now would be ... " or similar

mysql-test/t/percona_heap_var.test
dutow: Yes. While the ticket only described the first version, the same issue also happened when the user specifically requested some key block size.

dutow:
Updated.

mysql-test/t/percona_heap_var.test
```diff
@@ -4,57 +4,56 @@
 --source include/have_innodb.inc
 
 --disable_warnings
-drop table if exists t1;
+DROP TABLE IF EXISTS t1;
```
laurynas-biveinis, line 7: I didn't mean to suggest fixing existing testcase capitalization, but if you have done it, let's keep it

mysql-test/t/percona_heap_var.test
```diff
@@ -165,3 +164,57 @@ ALTER TABLE t1 ADD PRIMARY KEY (a, c);
...
+
+CREATE TABLE `t1` (
+  `id` INT(11) NOT NULL AUTO_INCREMENT,
+  `json_column` TEXT,
+  PRIMARY KEY (`id`)
+) ENGINE=MEMORY;
+# no key_block_size displayed even before the bugfix
+SHOW CREATE TABLE t1;
+
+ALTER TABLE  t1 ENGINE=INNODB;
```
laurynas-biveinis, line 180: Extra space between TABLE and t1. Only fix if there are other comments

laurynas-biveinis (approved):

dutow:
Jenkins: https://jenkins.percona.com/job/percona-server-5.5-param/1658/

PR #2312, opened 2018-04-12
Bug: PS-3877 User experience improvement for Percona keyring use

robgolebiowski:
When it is not possible to encrypt relay log - due to lack of keyring plugin installed -
an appropriate error message - mentioning relay log - is generated. Also an error messages
were improved and now provide more information.

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -4979,6 +4979,7 @@ bool MYSQL_BIN_LOG::open_binlog(const char *log_name,
   DBUG_ASSERT(need_sid_lock || !need_lock_index);
   DBUG_ENTER("MYSQL_BIN_LOG::open_binlog(const char *, ...)");
   DBUG_PRINT("enter",("base filename: %s", log_name));
+  const char *log_to_encrypt= is_relay_log ? "relay_log" : "binlog";
 
   mysql_mutex_assert_owner(get_log_lock());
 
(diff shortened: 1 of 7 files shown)


robgolebiowski:
JR => https://jenkins.percona.com/job/mysql-5.7-param/1761/

sql/binlog.cc
```diff
@@ -5320,7 +5321,9 @@ bool MYSQL_BIN_LOG::open_binlog(const char *log_name,
     std::string err_msg= "Either disk is full or file system is read only ";
     if (encrypt_binlog)
       err_msg+= "or encryption failed ";
-    err_msg+= "while opening the binlog. Aborting the server.";
+    err_msg+= "while opening the ";
```
RoelVdP, line 5324: does this and the other err_msg+ need to be two spaces more to the right, and does the "if (encrypt_binlog)" need { } ? or is the if (encrypt_binlog) just applicable to err_msg+= "or encryption failed "; ? In which case this comment can be ignored :)

RoelVdP (approved):

sql/binlog.cc
laurynas-biveinis: This is correct.

laurynas-biveinis (approved):

percona-ysorokin (approved):
LGTM

PR #2313, opened 2018-04-12
Native partitioning tokudb draft 5.7

vlad-lesin:
This PR is for preliminary code review and implementation details discussion, please, DON'T merge it.

It's more convenient to review the code commit-by-commit because the first two commits is just a copy of ha_partition.* and renaming the correspondent classes. Otherwise to see the changes in non-native-partitioning handler it is necessary to use "diff".

Also it's supposed that Partition_base will be used for RocksDB too. In the code there are first steps to split common and partition-specific code.


storage/tokudb/ha_tokupart.cc
```diff
@@ -0,0 +1,83 @@
...
+handler *ha_tokupart::clone(const char *name, MEM_ROOT *mem_root)
+{
+  ha_tokupart *new_handler;
+
+  DBUG_ENTER("Partition_base::clone");
+
+  /* If this->table == NULL, then the current handler has been created but not
+  opened. Prohibit cloning such handler. */
+  if (!table)
+    DBUG_RETURN(NULL);
```
george-lorch, line 35: Is this something we would expect to happen at all or is this really some strong error case where an assertion might be more appropriate?

storage/tokudb/ha_tokupart.cc
```diff
@@ -0,0 +1,83 @@
...
+
+  new_handler= new (mem_root) ha_tokupart(ht, table_share, m_part_info,
+                                           this, mem_root);
+  if (!new_handler)
+    DBUG_RETURN(NULL);
```
george-lorch, line 40: Same, this seems like an OOM problem so maybe a hard assertion/abort is in order?

dutow:
Closing this now, as native partitioning is already merged in #2493

PR #2316, opened 2018-04-12
TDB-189 : Fix -Werror=suggest-attribute=format

george-lorch:
- Annotated toku_dir_cmd_set_error with attribute((format...))


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2152/

laurynas-biveinis (approved):

PR #2318, opened 2018-04-13
PS-3936, PS-3940, PS-3943: key_block_size shouldn't be set automatically for the heap engine. (5.6)

dutow:
Key_block_size was set automatically by the improved heap storage engine, which caused:
* warnings (5.5/5.6) or errors (5.7) when changing the engine type to innodb
* constantly growing key_block_size during alter operations

With this change, key_block_size won't be set unless the user specifies a custom value.
* switching to innodb will only result in an error if the user manually specified a key_block_size before, which is expected.
* key_block_size was growing because max_chuck_size is set based on the specified key_block_size, and then key_block_size was set to max_chunk_size plus an overhead value, increasing it every time.
  by leaving out the second step, it remains stable.

Note that key_block_size was set so the chunk size is visible even when calculated automatically.
As it is always set to 256 unless the user specifies something else, this didn't provide too much information.

(cherry picked from commit 3c8964316ad5c931eaabcd9eb2a859c98717791e)

### DIFF
--- a/mysql-test/r/percona_heap_var.result
+++ b/mysql-test/r/percona_heap_var.result
@@ -1,11 +1,11 @@
-drop table if exists t1;
+DROP TABLE IF EXISTS t1;
 set @@session.max_heap_table_size=16*1024*1024;
-create table t1 (a int not null, b varchar(400), c int, primary key (a), key (c)) engine=heap comment="testing heaps" key_block_size=128;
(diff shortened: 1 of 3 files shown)


laurynas-biveinis (approved):

dutow:
Jenkins: https://jenkins.percona.com/job/percona-server-5.6-param/2161/

PR #2322, opened 2018-04-17
TDB-191 : Implement better compression algo testing

george-lorch:
- Implement new test that performs about as full of an end-to-end validation
  that the requested compression algo at the SQL layer is actually applied to
  the FT data files. The only practical way to check this is to use tokuftdump
  and look at the data files header value for compression_method.


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2158/

mysql-test/mysql-test-run.pl
```diff
@@ -2591,6 +2591,17 @@ sub environment_setup {
            "$basedir/storage/rocksdb/sst_dump");
   $ENV{'MYSQL_SST_DUMP'}= native_path($exe_sst_dump);
 
+  # ----------------------------------------------------
```
laurynas-biveinis, line 2594: I vaguely recall but am unable to confirm in the source code that it's possible to have MTR suite-specific Perl files for setup (suite.pm (?) but grep finds nothing). Would that be a better place for this?

mysql-test/suite/tokudb/t/compressions.test
```diff
@@ -0,0 +1,68 @@
...
+
+# LZMA - compression_method=10
+CREATE TABLE t4 (a INT) ENGINE=TokuDB ROW_FORMAT=TOKUDB_LZMA;
+--let $t4_file = query_get_value(SELECT internal_file_name FROM information_schema.tokudb_file_map WHERE dictionary_name = './test/t4-main', internal_file_name, 1)
+
+# ZLIB (without checksum) - compression_method=11
+CREATE TABLE t5 (a INT) ENGINE=TokuDB ROW_FORMAT=TOKUDB_ZLIB;
+--let $t5_file = query_get_value(SELECT internal_file_name FROM information_schema.tokudb_file_map WHERE dictionary_name = './test/t5-main', internal_file_name, 1)
+
+--let $datadir = query_get_value(SELECT @@global.datadir AS datadir, datadir, 1)
```
laurynas-biveinis: ```--let $datadir = `SELECT @@datadir` ```
likewise for TOKUDB_FILE_MAP queries

laurynas-biveinis (changes requested):

mysql-test/mysql-test-run.pl
george-lorch: Possibly, but we kind of already set this precedent by also taking in sst_dump for my rocks. It is worth a try though to see if it works, and if so, remove the sst_dump as well from the primary runner.

mysql-test/suite/tokudb/t/compressions.test
george-lorch: Likewise what? I do not understand how this has anything to do with test setups. This seems to me to be a pretty normal and straightforward way to compile a path to a specific data file and is based on copying ideas from some tests from other suites.

laurynas-biveinis: I am only suggesting to use backticks instead of query_get_value with three args for the same

mysql-test/mysql-test-run.pl
laurynas-biveinis: Yeah, this could have been done for Rocks too

mysql-test/suite/tokudb/t/compressions.test
george-lorch: Ahh, OK, formatting in browser was weird so the example was not clear/visible, I see the comment in the diff, I get it now...

mysql-test/mysql-test-run.pl
george-lorch: So I blindly tried a few combinations of suite .pm .pl, etc to no avail. I then rooted through the mysql-test0run.pl and lib/my_*.pm and found no evidence of any kind of suite specific perl setup modules. I am no perl expert by any stretch but from what I can tell, there is no such construct for this. It does seem like a nice idea. That said, I did find suite specific plugin.defs but IIRC we visited this once and there was something about these not being used/recognized during the initial bootstrap and test qualification and caused tests that required a plugin to run to be skipped.

george-lorch:
Converted the query_get_value to `blah`
Jenkins re-spin https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2159/

laurynas-biveinis (approved):

PR #2326, opened 2018-04-17
PS-3957 : TokuDB tests skipped because they expect non-default binlog…

george-lorch:
… format

- Added specific .opt files to force explicit setting binlog-format for rpl
  tests that are not under a suite with binlog-format in combinations.
- Re-recorded rpl_mixed_replace_into test to pick up Warning that is new to 5.7

### DIFF
--- a/mysql-test/suite/tokudb.bugs/r/rpl_mixed_replace_into.result
+++ b/mysql-test/suite/tokudb.bugs/r/rpl_mixed_replace_into.result
@@ -1,4 +1,7 @@
 include/master-slave.inc
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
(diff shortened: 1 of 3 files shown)


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1779/

laurynas-biveinis (approved):
The fix is correct, but the test looks like it's copy pasted three times with different binlog formats and is indeed a rpl one. It looks like it should go to tokudb.rpl, if the different formats require different .result files, then the copy-paste could be extracted to an include file. Again, it's OK to merge the current PR and revisit this later (or never)

george-lorch:
Yeah, I was thinking the exact same thing. Why is a rpl test with combinations in .bug instead of .rpl where it could get proper combination treatment. I didn't dig further out of pure laziness assuming that there was some real reason.

george-lorch:
Looking at the content of the test and the file history, there really isn't any indication that this is targeted at a specific bug and looks like a basic functional test. So, I am going to go ahead and just close this PR against 5.7 and create a new 5.6 GCA and 5.7 branch that moves the tests into the tokudb.rpl suite where they will run under the binlog_format combinations.

PR #2327, opened 2018-04-17
Fix BLD-1047 (Installing QRT plugins causes server crash) (5.6)

percona-ysorokin:
https://jira.percona.com/browse/BLD-1047

installed
'session_stats' plugin session variable used in this fuction
'QUERY_RESPONSE_TIME'.

The same type of crash potentially could have been caused by accessing
'exec_time_debug' variable from 'query_response_time_audit_notify()'.

Fixed by moving 'session_stats' and 'exec_time_debug' declarations from
'QUERY_RESPONSE_TIME' to 'QUERY_RESPONSE_TIME_AUDIT'.

without terminating 'NullS'.

Side effects:
- 'query_response_time_session_stats' plugin session variable renamed to
  'query_response_time_audit_session_stats'
- 'query_response_time_exec_time_debug' plugin debug-only session variable
   renamed to 'query_response_time_audit_exec_time_debug'

Re-recorded and a number of MTR test cases in the 'query_response_time'
test suite becaue of the renamed variables.


percona-ysorokin:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2160/

laurynas-biveinis (changes requested):
* There is a problem with user interface now: an asymmetry is introduced between the main collection variable query_response_time_stats and its session counterpart which now will have an extra "audit" in it. The right place for the former variable is the "audit" not I_S plugin too, but this change cannot be made in the middle of the series. I'd consider fixing by using a flag whether the main I_S plugin is installed before accessing the THDVAR (if it's not installed, assume THDVAR is false). exec_debug one is OK as-is
* Please move BLD to PS, update the commit message accordingly

percona-ysorokin:
Closed and reopened as PR #2329 to change misleading branch name

PR #2329, opened 2018-04-18
Fix PS-3959 (Installing QRT plugins causes server crash) (5.6)

percona-ysorokin:
https://jira.percona.com/browse/PS-3959

installed
'session_stats' plugin session variable used in this function
'QUERY_RESPONSE_TIME'.

Fixed by introducing a new global flag 'query_response_time_vars_initialized'
indicating whether all the plugin variables from 'QUERY_RESPONSE_TIME' were
registered and initialized and it is safe to access them.

'query_response_time_should_log()' now attempts to access 'session_stats'
variable only if this flag is set. If it is not set, its value considered to
be 'OFF'.

without terminating 'NullS'.

Added new MTR test case 'query_response_time.install_query_response_time'
which simulates the scenario reported in the original bug description.

### DIFF
--- a/plugin/query_response_time/plugin.cc
+++ b/plugin/query_response_time/plugin.cc
@@ -29,6 +29,46 @@ ulong opt_query_response_time_range_base= QRT_DEFAULT_BASE;
 static my_bool opt_query_response_time_stats= FALSE;
 static my_bool opt_query_response_time_flush= FALSE;
 
+class qrt_atomic_flag
+{
+  public:
+    explicit qrt_atomic_flag(bool initial_value):
+      value_(initial_value ? 1 : 0)
+    {
+      my_atomic_rwlock_init(&lock_);
+    }
+    ~qrt_atomic_flag()
+    {
+      my_atomic_rwlock_destroy(&lock_);
+    }
+    void set()
+    {
+      my_atomic_rwlock_wrlock(&lock_);
+      my_atomic_store32(&value_, 1);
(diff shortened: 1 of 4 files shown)


percona-ysorokin:
This is a continuation of PR #2327

percona-ysorokin:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2162/

laurynas-biveinis (commented):
INSTALL PLUGIN for I_S might happen in parallel with audit attempts for audit plugin - should the flag be atomic?

plugin/query_response_time/plugin.cc
```diff
@@ -154,8 +204,9 @@ static struct st_mysql_information_schema query_response_time_info_descriptor=
 
 static bool query_response_time_should_log(MYSQL_THD thd)
 {
-  const enum session_stat session_stat_val
-    = static_cast<session_stat>(THDVAR(thd, session_stats));
+  enum session_stat session_stat_val= session_stat_off;
```
laurynas-biveinis: I'd prefer "const enum = q_v_i.is_set() ? : " but gain is marginal at best, certainly no need to respin the PR

laurynas-biveinis (approved):

percona-ysorokin:
Updated Jenkins build link (after fixing shutdown problem)
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2165/

laurynas-biveinis (approved):

PR #2328, opened 2018-04-17
Fix BLD-1047 (Installing QRT plugins causes server crash) (5.7)

percona-ysorokin:
https://jira.percona.com/browse/BLD-1047

When only the audit part of the QRT plugin ('QUERY_RESPONSE_TIME_AUDIT') was
installed
"INSTALL PLUGIN QUERY_RESPONSE_TIME_AUDIT SONAME 'query_response_time.so'",
server used to crash in 'query_response_time_audit_notify()' as
'session_stats' plugin session variable used in this fuction
was not registered by MySQL plugin loading system in THD. This used to happen
because this variable was declared in another part of the same plugin -
'QUERY_RESPONSE_TIME'.

The same type of crash potentially could have been caused by accessing
'exec_time_debug' variable from 'query_response_time_audit_notify()'.

Fixed by moving 'session_stats' and 'exec_time_debug' declarations from
'QUERY_RESPONSE_TIME' to 'QUERY_RESPONSE_TIME_AUDIT'.

In addition, fixed problem with declaring 'session_stat_names' array
without terminating 'NullS'.

Side effects:
- 'query_response_time_session_stats' plugin session variable renamed to
  'query_response_time_audit_session_stats'
- 'query_response_time_exec_time_debug' plugin debug-only session variable
   renamed to 'query_response_time_audit_exec_time_debug'

Re-recorded and a number of MTR test cases in the 'query_response_time'
test suite becaue of the renamed variables.

PR #2330, opened 2018-04-18
Fix PS-3959 (Installing QRT plugins causes server crash) (5.7)

percona-ysorokin:
https://jira.percona.com/browse/PS-3959

installed
'session_stats' plugin session variable used in this function
'QUERY_RESPONSE_TIME'.

Fixed by introducing a new global flag 'query_response_time_vars_initialized'
indicating whether all the plugin variables from 'QUERY_RESPONSE_TIME' were
registered and initialized and it is safe to access them.

'query_response_time_should_log()' now attempts to access 'session_stats'
variable only if this flag is set. If it is not set, its value considered to
be 'OFF'.

without terminating 'NullS'.

Added new MTR test case 'query_response_time.install_query_response_time'
which simulates the scenario reported in the original bug description.

### DIFF
--- a/plugin/query_response_time/plugin.cc
+++ b/plugin/query_response_time/plugin.cc
@@ -30,6 +30,33 @@ ulong opt_query_response_time_range_base= QRT_DEFAULT_BASE;
 static my_bool opt_query_response_time_stats= FALSE;
 static my_bool opt_query_response_time_flush= FALSE;
 
+class qrt_atomic_flag
+{
+  public:
+    explicit qrt_atomic_flag(bool initial_value):
+      value_(initial_value ? 1 : 0)
+    {}
+    void set()
+    {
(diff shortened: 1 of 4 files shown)


percona-ysorokin:
This is a continuation of PR #2328

percona-ysorokin:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1782/

laurynas-biveinis (commented):
Same question as on the 5.6 version

laurynas-biveinis (approved):

laurynas-biveinis (approved):

percona-ysorokin:
Updated Jenkins build link (after fixing shutdown problem)
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1787/

PR #2331, opened 2018-04-18
PS-3957 : TokuDB tests skipped because they expect non-default binlog…

george-lorch:
… format

- These tests were located in tokudb.bugs suite which does not have
  binlog_format in the suite combinations like tokudb.rpl does. By all
  appearances, these tests are located in the wrong suite, and as such,
  only one of the tests would ever run depending on the value of binlog_format
  of the running server instance. Moved the tests to the tokudb.rpl suite and
  re-recorded as necessary to pick up new server warnings that were never
  recorded because the tests were never run.


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2163/

PR #2334, opened 2018-04-18
TDB-193 : Handlerton. trx->tokudb_lock_count > 0

george-lorch:
- Commit https://github.com/percona/percona-server/commit/530b938570593d661c496978bd4d59c8695a5e83
  for TDB-129 incorrectly removed 5.5 and MariaDB specific compile out guards
  for code that should not exist within Percona Server 5.7. This eventually
  leads to the assertion in the issue report when an the FT layer returns an
  error during an alter table.

### DIFF
--- a/mysql-test/suite/tokudb.bugs/t/TDB-193.test
+++ b/mysql-test/suite/tokudb.bugs/t/TDB-193.test
@@ -1,0 +1,20 @@
+--source include/have_tokudb.inc
+
+CREATE TABLE t(a INT, b INT, c CHAR(1), d CHAR (1), e VARCHAR(1), f VARCHAR(1), g BLOB, h BLOB, id INT, KEY(b), CLUSTERING KEY(e)) ENGINE=TokuDB;
+
+# this will return error -10000 from PerconaFT which is TOKUDB_OUT_OF_LOCKS but
+# mtr does not allow a negative value to be passed to --error, so, disable error
+# handling for now
+--disable_abort_on_error
+--replace_regex /#sql[0-9a-f_-]*/#sql-temporary/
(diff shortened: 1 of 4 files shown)


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1784/

mysql-test/suite/tokudb.bugs/t/TDB-193.test
```diff
@@ -0,0 +1,20 @@
+--source include/have_tokudb.inc
+
+CREATE TABLE t(a INT, b INT, c CHAR(1), d CHAR (1), e VARCHAR(1), f VARCHAR(1), g BLOB, h BLOB, id INT, KEY(b), CLUSTERING KEY(e)) ENGINE=TokuDB;
+
+# this will return error -10000 from PerconaFT which is TOKUDB_OUT_OF_LOCKS but
```
laurynas-biveinis, line 5: Are error codes signed in the first place?

laurynas-biveinis (approved):

mysql-test/suite/tokudb.bugs/t/TDB-193.test
george-lorch: Yes, but as far as I know, none of the MySQL error codes are < 0. This is an error coming out of the ydb layer of FT that is propagated directly out to the client. That may be an issue in itself, but is not related to this particular patch.

`mysql> ALTER TABLE t CHANGE COLUMN a a CHAR(1);
ERROR 1025 (HY000): Error on rename of './test/t' to './test/#sql2-a6e-3' (errno: -100000 - Unknown error -100000)`

PR #2335, opened 2018-04-19
PS-3950 (5.7): gcc-8 compilation warnings

inikep:
1. Add gcc-8 to Travis config file
2. Fix the following gcc-8 warnings:
- cast between incompatible function types [-Werror=cast-function-type]
- clearing an object with no trivial copy-assignment [-Werror=class-memaccess]
- directive output may be truncated [-Werror=format-truncation=]
- type qualifiers ignored on cast result type [-Werror=ignored-qualifiers]
- catching polymorphic type 'class std::exception' by value [-Werror=catch-value=]

(cherry picked from commit 31b49ba6b4123dfab049357fc9a18d422dad741c)

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -3810,9 +3810,10 @@ static int generate_server_uuid()
 
   delete thd;
 
-  strncpy(server_uuid, uuid.c_ptr(), UUID_LENGTH);
+  strncpy(server_uuid, uuid.c_ptr(), sizeof(server_uuid));
   DBUG_EXECUTE_IF("server_uuid_deterministic",
-                  strncpy(server_uuid, "00000000-1111-0000-1111-000000000000", UUID_LENGTH););
+                  strncpy(server_uuid, "00000000-1111-0000-1111-000000000000",
+                          sizeof(server_uuid)););
   server_uuid[UUID_LENGTH]= '\0';
   return 0;
 }
--- a/storage/innobase/buf/buf0dblwr.cc
+++ b/storage/innobase/buf/buf0dblwr.cc
@@ -346,7 +346,7 @@ buf_parallel_dblwr_make_path(void)
 	if (parallel_dblwr_buf.path)
 		return(DB_SUCCESS);
 
-	char path[FN_REFLEN];
+	char path[FN_REFLEN + 1 /* OS_PATH_SEPARATOR */];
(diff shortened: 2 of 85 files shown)


sql/mysqld.cc
```diff
@@ -3810,9 +3810,10 @@ static int generate_server_uuid()
 
   delete thd;
 
-  strncpy(server_uuid, uuid.c_ptr(), UUID_LENGTH);
+  strncpy(server_uuid, uuid.c_ptr(), sizeof(server_uuid));
```
laurynas-biveinis, line 3813: This is asking for ```static_assert(sizeof(server_uuid) >= UUID_LENGTH)``` but I am not sure it's our job to add it if no other comments on the PR.

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -346,7 +346,7 @@ buf_parallel_dblwr_make_path(void)
 	if (parallel_dblwr_buf.path)
 		return(DB_SUCCESS);
 
-	char path[FN_REFLEN];
+	char path[FN_REFLEN + 1 /* OS_PATH_SEPARATOR */];
```
laurynas-biveinis, line 349: Why?

laurynas-biveinis (approved):

storage/innobase/buf/buf0dblwr.cc
inikep: The reason is here:
```
			ut_snprintf(path, sizeof(path), "%s%c%s",
				    dir,
				    OS_PATH_SEPARATOR,
				    srv_parallel_doublewrite_path);
```

laurynas-biveinis: Yes, but I am not sure it makes sense - if FN_REFLEN is the maximum allowed path name, any concatenations still must fit the same FN_REFLEN

inikep: Linux has maximum path of 4096 characters.
`FN_REFLEN` seems to be MySQL limitation.
In this case `FN_REFLEN+1` should work well because after `ut_snprintf(path,...` we have:
```
parallel_dblwr_buf.path = mem_strdup(path);
```

percona-ysorokin (approved):
LGTM

PR #2336, opened 2018-04-19
PS-3325: Prevent changes with gtid_deployment_step set

dutow:
Changes executed after gtid_deployment_step was set resulted in binlog damage, or assertions/crashes of the server.

This commit ensures that checks are based on the correct, lock protected variable, and also enforces super_read_only while gtid_deployment_step is set.

It incorporates PR#1554 by keyurdg, which implements the first part of the above changes.

### DIFF
--- a/mysql-test/t/percona_bug_ps3325-slave.opt
+++ b/mysql-test/t/percona_bug_ps3325-slave.opt
@@ -1,0 +1,3 @@
+--gtid_mode=ON
+--log_slave_updates=ON
+--enforce_gtid_consistency
--- a/mysql-test/t/percona_bug_ps3325.test
+++ b/mysql-test/t/percona_bug_ps3325.test
@@ -1,0 +1,48 @@
+#
+# PS-3325: Queries while/after setting gtid_deployment_step cause binlog corruption and assertions
+#
+--source include/have_innodb.inc
+--source include/master-slave.inc
+
+connect (mm,localhost,root,,);
+
+connection master;
+CREATE TABLE test_table(id INT) ENGINE=innodb;
+--source include/sync_slave_sql_with_master.inc
+
+connection master;
+INSERT INTO test_table VALUES (1);
+--source include/sync_slave_sql_with_master.inc
+
+connection master;
+SET GLOBAL gtid_deployment_step = ON;
+--error ER_GTID_DEPLOYMENT_STEP_ACTIVE
+SET GLOBAL super_read_only=1;
+--error ER_GTID_DEPLOYMENT_STEP_ACTIVE
+SET GLOBAL super_read_only=0;
+
+connection master;
+--error ER_OPTION_PREVENTS_STATEMENT
+INSERT INTO test_table VALUES (2);
+
+SET GLOBAL gtid_deployment_step = OFF;
+
+SET GLOBAL super_read_only = ON;
+SET GLOBAL gtid_deployment_step = ON;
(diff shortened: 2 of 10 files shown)


dutow:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1848/

mysql-test/t/percona_bug_ps3325-slave.opt
```diff
@@ -0,0 +1,3 @@
+--gtid_mode=ON
+--log_slave_updates
```
laurynas-biveinis: Nit: inconsistent with -master.opt

mysql-test/t/percona_bug_ps3325.test
```diff
@@ -0,0 +1,49 @@
...
+# PS-3325: Queries while/after setting gtid_deployment_step cause binlog corruption and assertions
+#
+--source include/have_innodb.inc
+--source include/master-slave.inc
+
+connect (mm,localhost,root,,);
+
+connection master;
+CREATE TABLE test_table(id INT) ENGINE=innodb;
+sync_slave_with_master;
```
laurynas-biveinis: --source include/sync_slave_sql_with_master.inc here and everywhere

mysql-test/t/percona_bug_ps3325.test
```diff
@@ -0,0 +1,49 @@
...
+
+connection master;
+INSERT INTO test_table VALUES (1);
+sync_slave_with_master;
+
+connection master;
+SET GLOBAL gtid_deployment_step = ON;
+--error ER_GTID_DEPLOYMENT_STEP_ACTIVE
+set global super_read_only=1;
```
laurynas-biveinis: Inconsistent SQL case

mysql-test/t/percona_bug_ps3325.test
```diff
@@ -0,0 +1,49 @@
...
+
+
+connection slave;
+SELECT * FROM test_table;
+
+connection master;
+DROP TABLE test_table;
+
+--source include/rpl_end.inc
+--source include/log_cleanup.inc
```
laurynas-biveinis: Why?

sql/sys_vars.cc
```diff
@@ -5227,13 +5230,20 @@ static bool fix_gtid_deployment_step(sys_var *self, THD *thd, enum_var_type type
...
     goto end_with_read_lock;
 
-  /*
-   Change the opt_deployment_step system variable,
-   safe because the lock is held
-  */
   opt_gtid_deployment_step= new_gtid_deployment_step;
   result= false;
 
+  if(opt_gtid_deployment_step)
```
laurynas-biveinis: Space after if

laurynas-biveinis (changes requested):
gtid_deployment_step variables now can be made static to sys_vars.cc?

dutow:
Updated.
Jenkins: https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/6/

laurynas-biveinis (commented):
What about making gtid_deployment_step variable static to sys_vars.cc?

dutow:
That's doable too, I kept it this way because now both variables are initialized in `sql/mysqld.cc` at the same location, and that seems logical. Making it static would mean that at least one of them should be initialized in a different file.

laurynas-biveinis (approved):
Please prepare a null-merge to 5.7 and push both

PR #2340, opened 2018-04-20
Fix for DOC 722 (a warning on mixing secret_mount_points)

fiowro:


laurynas-biveinis:
@fiowro, who has reviewed this PR?

PR #2346, opened 2018-04-23
PS-3842: PS 5.7 cannot be built with -DWITH_LZ4=system on Ubuntu 14.04

inikep:
Use a newer LZ4 build from https://launchpad.net/~jonathonf


percona-ysorokin (approved):
LGTM

PR #2359, opened 2018-04-26
Release 5.7.21 21

EvgeniyPatlan:


laurynas-biveinis:
@EvgeniyPatlan, does the release tag point to the right commit? Is there more than one tag?

EvgeniyPatlan:
@laurynas-biveinis yes tag is set to correct commit. We have only one tag as we have updated package version not PS version.

laurynas-biveinis:
@EvgeniyPatlan I am slightly confused because the tag should include the correct complete source tree, packaging included. So which packaging commit does it point to?

EvgeniyPatlan:
@laurynas-biveinis it is https://github.com/percona/percona-server/pull/2359/commits/2a37e4ea859cd2c2de51388ecc0cbf6fb10d0ed5

There is no package version in it as it is set during package build

laurynas-biveinis:
Thanks, it looks good then

PR #2361, opened 2018-04-27
PS-3968: Tune clang-format diff checkings for existing SQL and InnoDB style

inikep:
Create two `.clang-format` config files: one for InnoDB and one for the rest (SQL).
The files are compatible with clang-format-5.0 or newer.

### DIFF
--- a/storage/innobase/.clang-format
+++ b/storage/innobase/.clang-format
@@ -1,0 +1,111 @@
+# generated with:
+# clang-format-5.0 -style=Google --dump-config
+
+Language:        Cpp
+# BasedOnStyle:  Google
+AccessModifierOffset: -1
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignEscapedNewlines: Left
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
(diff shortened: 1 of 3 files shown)


laurynas-biveinis (commented):
How the styles were verified to match the SQL/InnoDB styles?

inikep:
I tried manually to minimize differences between original files and clang-formatted. There is no config that will 100% match all files as I found different formatting styles even in the same file. This PR will only reduce a number of formatting errors from Travis. Moreover adding `-style=file` causes that a proper submodule `.clang-format` for RocksDB and TokuDB will be used.

laurynas-biveinis (approved):

storage/innobase/.clang-format
```diff
@@ -0,0 +1,111 @@
...
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Auto
+TabWidth:        8
+
+
+# changes for MySQL 5.x
```
percona-ysorokin: `changes for MySQL 5.x (InnoDB)`

percona-ysorokin (approved):
LGTM
minor comment can be ignored.

PR #2367, opened 2018-05-09
PS-3951: Fix decreasing status counters (5.6)

dutow:
Status counters were cleaned in THD::release_resources, before an optional thread cleanup.
The cleanup could possibly perform a rollback, or other counter-increasing operations.
In this case, these additions to the thread local counters were lost during the destruction of the thread, but were visible for a short time, before the thread was removed from the global thread list.

This change avoids this issue by moving the global counter update to the end of the release_resources method, ensuring that possible counter-changing operation happens before it.


sql/sql_class.cc
```diff
@@ -1773,10 +1773,6 @@ void THD::release_resources()
   mysql_mutex_assert_not_owner(&LOCK_thread_count);
   DBUG_ASSERT(m_release_resources_done == false);
 
-  mysql_mutex_lock(&LOCK_status);
-  add_to_status(&global_status_var, &status_var);
-  memset(&status_var, 0, sizeof(status_var));
-  mysql_mutex_unlock(&LOCK_status);
```
laurynas-biveinis: Remove one empty line too

laurynas-biveinis (changes requested):
- Based on the description, it would seem that a DEBUG_SYNC/DBUG_EXECUTE_IF testcase would be possible?
- It is slightly suspicious that a debug build does not reproduce the issue. Is it because the race window becomes much smaller due to some extra work/sync?
- Does it make sense to add DBUG_ASSERT that counters are empty to the destructor?

dutow:
Updated with a test case - so it is possible to crash the debug build with sync points, the issue was that I didn't use an innodb table in my mtr test case - and if there's nothing to roll back, it doesn't happen. 

https://jenkins.percona.com/job/percona-server-5.6-param/2184/

mysql-test/t/bug90351.test
```diff
@@ -0,0 +1,27 @@
+#
+# Debug build tests for INFORMATION_SCHEMA.GLOBAL_TEMPORARY_TABLES
```
laurynas-biveinis: Comment from a different testcase

mysql-test/t/bug90351.test
```diff
@@ -0,0 +1,27 @@
...
+#
+--source include/have_innodb.inc
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+
```
laurynas-biveinis: --source include/count_sessions.inc

mysql-test/t/bug90351.test
```diff
@@ -0,0 +1,27 @@
...
+SET DEBUG_SYNC="thd_cleanup_start WAIT_FOR increased_counter";
+CREATE TABLE t1 (i INTEGER);
+INSERT INTO t1 VALUES (1);
+let $con1_id= `SELECT connection_id()`;
+SET AUTOCOMMIT=0;
+START TRANSACTION;
+DELETE FROM t1;
+
+connection con2;
+SET DEBUG_SYNC="after_ha_rollback_count_increased SIGNAL increased_counter";
```
laurynas-biveinis: How does this work? debug_sync is a session variable, and con2 does not do rollback?

mysql-test/t/bug90351.test
```diff
@@ -0,0 +1,27 @@
...
+--disable_query_log
+eval KILL $con1_id;
+--enable_query_log
+
+connection default;
+DROP TABLE t1;
```
laurynas-biveinis, line 23: Missing con1, con2 disconnects, and then add --source include/wait_until_count_sessions.inc at the very end

laurynas-biveinis (changes requested):

mysql-test/t/bug90351.test
```diff
@@ -0,0 +1,27 @@
+#
+# Debug build tests for INFORMATION_SCHEMA.GLOBAL_TEMPORARY_TABLES
+#
+--source include/have_innodb.inc
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+
+connect (con1,localhost,root,,);
+connect (con2,localhost,root,,);
```
percona-ysorokin: I believe, you do not need 3 connections (`default`, `con1` and `con2`) in this test case - just `default` and `con1` should suffice.

mysql-test/t/bug90351.test
dutow: You are right, that wasn't necessary - the other debug sync is enough.

dutow:
Updated, jenkins: https://jenkins.percona.com/job/percona-server-5.6-param/2186/

mysql-test/t/bug90351.test
```diff
@@ -0,0 +1,29 @@
...
+#
+--source include/have_innodb.inc
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+--source include/count_sessions.inc
+
+connect (con1,localhost,root,,);
+
+connection con1;
+SET DEBUG_SYNC="thd_cleanup_start WAIT_FOR increased_counter";
```
laurynas-biveinis: Now the testcase syncs does not make sense: so this thread, upon reaching thd_cleanup_start, waits for increased_counter. But nothing signals it.

laurynas-biveinis (changes requested):

mysql-test/t/bug90351.test
dutow: I could signal it from a different connection with a dummy statement or something - but I found that its not needed.

* With the unmodified code, the thread waits here, and the assertion is hit
* With the modified code, the test case passes, the thread ends even if I do not send a signal to this sync point. This doesn't happen in my other tests, so I think this could be related to the kill statement.

laurynas-biveinis: I see, but the testcase needs all connections actually closed - so do you rely on debug_sync timeout then?

dutow:
Single testcase jenkins run: https://jenkins.percona.com/job/percona-server-5.6-param/2193/

laurynas-biveinis (approved):

percona-ysorokin (approved):
LGTM

PR #2370, opened 2018-05-10
PS-4469 : variable use out of scope bug in get_last_key test detected…

george-lorch:
… by ASAN in clang 6

PS-4470 : the cachetable-simple-pin-nonblocking-cheap test occasionally fails due to a locking conflict with the cachetable evictor

- Advance PerconaFT git submodule commit pointer to pick up changes.


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2175/

PR #2372, opened 2018-05-11
PS-4471 (5.6): gcc-4.4 error: #pragma GCC diagnostic not allowed inside functions

inikep:
Use separate macros for gcc-4.5 and older to disable "-Wuninitialized" warning.


inikep:
https://jenkins.percona.com/job/percona-server-5.6-param/2176/

percona-ysorokin (approved):
LGTM

inikep:
I wanted to show that `GCC45_DISABLE_WARN_UNINITIALIZED` macro works only for older gcc.
`MY_DISABLE_WARN_MAYBE_UNINITIALIZED_GLOBAL` may suggest that we disable warnings for all gcc versions.

PR #2385, opened 2018-05-18
Fix for DOC-734 changing launchpad links to JIRA, 5.6 version

fiowro:


laurynas-biveinis (changes requested):
Same comments as for the 5.5 version, and the last check date is not updated at all

laurynas-biveinis (changes requested):
Same trailing whitespace comments as for the 5.5 version

fiowro:
Highlighted trailing whitespaces are just visual representation of the github diff. Raw file format shows the whole table is filled with spaces

laurynas-biveinis (approved):

PR #2386, opened 2018-05-18
Fix for DOC-734 changing launchpad links to JIRA, 5.7 version

fiowro:


laurynas-biveinis (changes requested):

laurynas-biveinis (changes requested):
Same trailing whitespace comments as for 5.5 and 5.6

fiowro:
Same answer on highlighted trailing whitespaces (just a diff visual highlighting; the whole table is filled with spaces)

laurynas-biveinis (approved):

PR #2391, opened 2018-05-21
Prevent reusing iterator bounds between different iterators

george-lorch:
Upstream commit ID : fb-mysql-5.6-35/cd29bb9bf8fa73cd4f5f8f4cb3434d20f9109b09
PS-4495 : UPDATE with unique constraint does not work

Summary:
Update statements were returning incorrect results because the iterator scanning the table shared the same upper/lower bounds as the iterator locking secondary keys.

To fix, the code has been restructured so that it is clearer that each iterator needs to maintain and own their upper and lower bounds, and the buffers backing them.

Fixes https://github.com/facebook/mysql-5.6/issues/830
Closes https://github.com/facebook/mysql-5.6/pull/833

Differential Revision: D8037160

Pulled By: lth

fbshipit-source-id: e432e33


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1815/4

laurynas-biveinis (approved):
I guess the patch is identical to the upstream version?

george-lorch:
Not exactly identical, there was some minor cherry-pick conflict and the test case needed a little work.

PR #2392, opened 2018-05-23
PS-3951: Fix decreasing status counters (5.7)

dutow:
I also added asserts for the other counter changes, but I ended up not committing them, to keep the patch smaller - the test case only requires this one.

I still have them as a local change, so I can add them if we need it.


dutow:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1825/

dutow:
Updated based on the 5.6 changes

laurynas-biveinis (approved):

PR #2394, opened 2018-05-23
PS-4506 (5.7): Backport fixes from 8.0 for InnoDB memcached Plugin

inikep:
The original commit is available at:
https://github.com/mysql/mysql-server/commit/fb8d67315d2cb956976d838559ae3bdc88d43d60

Bug#26442367 ALWAYS ENABLE SAFE_MUTEX IN DEBUG BUILDS, MAKE MEMCACHED UBSAN CLEAN

CMakeLists.txt contains this snippet:

IF(NOT WITH_INNODB_MEMCACHED)
  FOREACH(LANG C CXX)
      SET(CMAKE_${LANG}_FLAGS_DEBUG
          "${CMAKE_${LANG}_FLAGS_DEBUG} -DSAFE_MUTEX")
  ENDFOREACH()
ENDIF()

There  should be  no  reason to  disable safe_mutex  if  we build  the
memcached  plugin  (it  actually   hides  real  problems,  since  most
pushbuild trees are built with  the memcached plugin) Enabling it, and
running with  UBSAN shows a  few issues,  with wrong mutex  usage, and
undefined behaviour.

engine.h
src/innodb_engine.c:312:32:
runtime error: index 1 out of bounds for type 'feature_info [1]'

config_parser.c
config_parser.c:127:24:
runtime error: signed integer overflow: 1073741824 * 1024 cannot be represented in type 'int'

innodb_engine.c
the double assignment to innodb_eng->info.info.features[0].feature *must* be wrong
innodb_engine.c:2341:3: runtime error:
null pointer passed as argument 1, which is declared to never be null

sql/sql_plugin.cc
==9557== Process terminating with default action of signal 6 (SIGABRT)
==9557==    at 0x78DD91F: raise (in /usr/lib64/libc-2.24.so)
==9557==    by 0x78DF519: abort (in /usr/lib64/libc-2.24.so)
==9557==    by 0x78D5DA6: __assert_fail_base (in /usr/lib64/libc-2.24.so)
==9557==    by 0x78D5E51: __assert_fail (in /usr/lib64/libc-2.24.so)
==9557==    by 0x23272A2: safe_mutex_assert_owner (thr_mutex.h:151)
==9557==    by 0x2329884: plugin_del(st_plugin_int*) (sql_plugin.cc:1097)
==9557==    by 0x232C1FB: memcached_shutdown() (sql_plugin.cc:1889)
==9557==    by 0x21BC38F: clean_up(bool) (mysqld.cc:1925)
==9557==    by 0x21C7A72: mysqld_main(int, char**) (mysqld.cc:5970)
==9557==    by 0x21B9795: main (main.cc:25)

Change-Id: I88bb6751ae8401d4c461a6f03dd098f016e0adf8


laurynas-biveinis (approved):

percona-ysorokin (approved):
LGTM

PR #2395, opened 2018-05-25
PS-3828 (InnoDB statistics in extended slow query log always zero)

laurynas-biveinis:
Percona Server 5.7 port had a regression where InnoDB slow query log
extensions always returned zero values instead of actual data. This
regression went unnoticed with existing MTR tests, which only checked
for the correct slow query log format, without attempting to do any
actual value validation.

While 5.5 does not have the same bug, it is tested the same, thus
rewrite the MTR tests to expand the testing coverage: that
InnoDB_trx_id is not zero; test specifically that I/O reads are
non-zero when I/O happens and zeros otherwise; likewise for record
lock and InnoDB queue waits.

The InnoDB queue wait was not tested at all, thus add the necessary
debug sync points and a new test innodb.percona_log_slow_innodb_debug
for it.

At the same time do other test cleanups, including moving them to the
right suites.

https://jenkins.percona.com/job/percona-server-5.5-param/1663/


percona-ysorokin (approved):
LGTM

PR #2396, opened 2018-05-25
Ps 3828 5.6

laurynas-biveinis:
https://jenkins.percona.com/job/percona-server-5.6-param/2185/


mysql-test/suite/innodb/t/percona_log_slow_innodb.test
```diff
@@ -0,0 +1,100 @@
+#
+# Test that InnoDB stats for the extended slow query log are collected.
+#
+--source include/have_innodb.inc
+
+CREATE TABLE t1(a INT AUTO_INCREMENT PRIMARY KEY) ENGINE=InnoDB;
+INSERT INTO t1 VALUES ();
```
percona-ysorokin, line 7: Don't we need more than 1 record here as in 5.5?

laurynas-biveinis: 5.5 needs not more than one record, but more than one index page, otherwise, on table opening it will be read for stats, putting the whole table in the buffer pool. 5.6 does not need this due to persistent stats. Or is it a different issue you had in mind?

percona-ysorokin: Yes, exactly the issue you described. Understood.

percona-ysorokin (approved):
LGTM

PR #2397, opened 2018-05-25
Ps 3828 5.7

laurynas-biveinis:
https://jenkins.percona.com/job/mysql-5.7-param/1828/


sql/sql_class.h
```diff
@@ -1566,6 +1567,66 @@ typedef struct
   ulonglong start_utime;
 } QUERY_START_TIME_INFO;
 
+/**
+   A single-hash-function bloom filter for approximate accessed page
+   counter.
+*/
+class Bloom_filter {
```
percona-ysorokin, line 1574: non-copyable

sql/sql_class.h
```diff
@@ -1566,6 +1567,66 @@ typedef struct
...
+     Check whether key is maybe a member of a set
+
+     @param[in, out]    mem_root        MEM_ROOT to allocate the bit set in, if
+      not allocated already
+     @param[in]         key             key whose presence to check
+
+     @return if true, the key might be a member of the set. If false, the key
+     is definitely not a member of the set.
+  */
+  bool test_and_set(MEM_ROOT *mem_root, ulong key)
```
percona-ysorokin, line 1613: Shouldn't `mem_root` be a constructor's parameter?

sql/sql_class.h
```diff
@@ -1566,6 +1567,66 @@ typedef struct
...
+  {
+    if (!bit_set)
```
percona-ysorokin, line 1615: Does `allocate on first use` strategy really save us resources. Or is it better to move it to the constructor?

storage/innobase/include/trx0trx.h
```diff
@@ -1322,6 +1450,100 @@ trx_is_started(
...
+	ulint sec, ms;
+	ut_usectime(&sec, &ms);
+	return static_cast<ib_uint64_t>(sec * 1000000 + ms);
+}
+
+inline ib_uint64_t
+trx_stats::start_io_read(trx_t *trx, ulint bytes)
+{
+	if (UNIV_LIKELY_NULL(trx))
+		return trx->stats.start_io_read(*trx, bytes);
```
percona-ysorokin: Why not just `return start_io_read(*trx, bytes);`?

storage/innobase/include/trx0trx.h
```diff
@@ -1322,6 +1450,100 @@ trx_is_started(
...
+	thd_report_innodb_stat(trx.mysql_thd, trx.id,
+			       MYSQL_TRX_STAT_IO_READ_WAIT_USECS,
+			       finish_time - start_time);
+}
+
+inline void
+trx_stats::end_io_read(trx_t *trx, ib_uint64_t start_time)
+{
+	if (UNIV_UNLIKELY(start_time != 0))
+		trx->stats.end_io_read(*trx, start_time);
```
percona-ysorokin: Just `end_io_read(*trx, start_time);`?

storage/innobase/include/trx0trx.h
```diff
@@ -1322,6 +1450,100 @@ trx_is_started(
...
+{
+	thd_report_innodb_stat(trx.mysql_thd, trx.id,
+			       MYSQL_TRX_STAT_IO_READ_BYTES, n);
+}
+
+inline void
+trx_stats::bump_io_read(trx_t *trx, ulint n)
+{
+	if (UNIV_LIKELY_NULL(trx))
+		trx->stats.bump_io_read(*trx, n);
```
percona-ysorokin: Likewise

storage/innobase/include/trx0trx.h
```diff
@@ -1322,6 +1450,100 @@ trx_is_started(
...
+			       MYSQL_TRX_STAT_LOCK_WAIT_USECS,
+			       now - lock_que_wait_ustarted);
+	ut_d(lock_que_wait_ustarted = 0);
+}
+
+inline void
+trx_stats::inc_page_get(trx_t *trx, ulint page_id_fold)
+{
+	if (UNIV_LIKELY_NULL(trx))
+		trx->stats.inc_page_get(*trx, page_id_fold);
```
percona-ysorokin: Likewise

storage/innobase/include/trx0trx.h
```diff
@@ -902,6 +903,146 @@ struct TrxVersion {
...
+	@param	trx	transaction to account I/O to
+	@param	bytes	number of bytes read
+	*/
+	static void
+	bump_io_read(const trx_t &trx, ulint bytes);
+
+	/**
+	Register, if needed, a start of lock wait */
+	void
+	start_lock_wait() {
```
percona-ysorokin, line 980: Any reason why only this method is defined here in place whereas others are defined externally via `inline` keyword?

storage/innobase/include/trx0trx.h
```diff
@@ -902,6 +903,146 @@ struct TrxVersion {
...
+		ut_usectime(&sec, &ms);
+		lock_que_wait_ustarted =
+		    static_cast<ib_uint64_t>(sec * 1000000 + ms);
+	}
+
+	/**
+	Register, if needed, a finished lock wait.
+
+	@param	trx	transaction to account to, if needed */
+#ifdef UNIV_DEBUG
```
percona-ysorokin: I would declare `stop_lock_wait` non-const regardless of `UNIV_DEBUG` just for clarity/ symmetry reasons unless you are fighting with modern compiler warnings.

storage/innobase/include/trx0trx.h
```diff
@@ -902,6 +903,146 @@ struct TrxVersion {
...
+	Register, if needed, start of an I/O read or wait.
+
+	@param	trx	transaction to account I/O to or NULL
+	@param	bytes	number of bytes to read or 0 if this is a wait for an
+	 already posted read in progress
+	@return value to be passed to end_io_read
+	*/
+	MY_NODISCARD
+	static ib_uint64_t
+	start_io_read(trx_t *trx, ulint bytes);
```
percona-ysorokin, line 928: Also consider an auxiliary RAII class which would do `start_io_read()` in constructor and `end_io_read()` in destructor.

storage/innobase/include/trx0trx.h
percona-ysorokin: Also consider an auxiliary RAII class which would do `start_lock_wait()` in constructor and `stop_lock_wait()` in destructor.

percona-ysorokin (changes requested):

sql/sql_class.h
laurynas-biveinis: Saves a pointer

sql/sql_class.h
laurynas-biveinis: The we'd have 8K-larger THD every time, regardless of slow query log configuration.

storage/innobase/include/trx0trx.h
laurynas-biveinis: Here and everywhere below: the fast-path through extra logging is with InnoDB extensions disabled, and that happens when trx == NULL

storage/innobase/include/trx0trx.h
laurynas-biveinis: I'd prefer everything inplace but forward references happen. Thus it's a mix of the two. Do you want everything defined externally?

laurynas-biveinis: I'd rather do RAII as a follow-up. Too many things going at once in the PR already

storage/innobase/include/trx0trx.h
percona-ysorokin: Still not quiet get it. My main question why do we call `STATIC` methods (like `start_io_read()`) via a pointer to object

laurynas-biveinis: Sorry I misunderstood, I thought you suggested removing the if () statement. Yes, will not call static methods through object pointer

storage/innobase/include/trx0trx.h
percona-ysorokin: Regarding `inline`: just wanted to hear that there is a reason for doing this. I doubt this is worth reworking.

storage/innobase/include/trx0trx.h
laurynas-biveinis: Making non-const

laurynas-biveinis:
Comments addressed, https://jenkins.percona.com/job/mysql-5.7-param/1841/

sql/sql_class.h
```diff
@@ -1566,6 +1567,70 @@ typedef struct
...
+    {
+      void *bit_set_place= alloc_root(mem_root, sizeof(Bit_set));
+      // FIXME: memory allocation failure is eaten silently. Nonexact stats are
+      // the least of the concerns then.
+      if (!bit_set_place)
+        return false;
+      bit_set= new (bit_set_place) Bit_set();
+    }
+    // Duplicating ut_hash_ulint calculation
+    const ulong pos= (key ^ 1653893711) % SIZE;
```
percona-ysorokin, line 1625: `ut_hash_ulint()` assumes that the second parameter (modulus) is a prime number.
```c++
/*******************************************************//**
The following function generates a hash value for a ulint integer
to a hash table of size table_size, which should be a prime or some
random number to work reliably.
@return	hash value */
UNIV_INLINE
ulint
ut_hash_ulint(
/*==========*/
	ulint	 key,		/*!< in: value to be hashed */
	ulint	 table_size);	/*!< in: hash table size */
```
In case of the modulus being equal to a power of 2, xoring with 1653893711 does not make the distribution better and makes it equivalent to `const ulong pos= key % SIZE;` which is just taking the last 13 bits in this particular case (when SIZE == 8192).

Consider using something like `const ulong pos= (key * <BIG_PRIME>) % SIZE;` where `BIG_PRIME < SIZE`.
BIG_PRIME == 8191, for instance.

percona-ysorokin (changes requested):

sql/sql_class.h
laurynas-biveinis: Good catch - since this calculation is the only way to index into the bitset, OK to make SIZE 8191 in order not to introduce another constant?

percona-ysorokin: Yes, making `SIZE=8191` is even better.

laurynas-biveinis:
Updated, https://jenkins.percona.com/job/mysql-5.7-param/1850/

percona-ysorokin (approved):
LGTM

laurynas-biveinis:
RAII logged at https://jira.percona.com/browse/PS-4534

PR #2407, opened 2018-05-30
Null-merge branch 5.6 (commit f5c4f8d) into 5.7

inikep:
This is a null merge after merging upstream 5.6.40 and 5.7.22.


sql/sp_head.cc
```diff
@@ -1973,9 +1973,10 @@ bool sp_head::show_routine_code(THD *thd)
     */
     if (ip != i->get_ip())
     {
-      const char *format= "Instruction at position %u has m_ip=%u";
-      char tmp[64 + 2 * MY_INT32_NUM_DECIMAL_DIGITS];
-      sprintf(tmp, format, ip, i->get_ip());
+      const char format[]= "Instruction at position %u has m_ip=%u";
```
laurynas-biveinis: If this is touched, might as well be made static

laurynas-biveinis (changes requested):
It is slightly suspicious that the warning fixes, which motivated the changes being ported, need re-doing in 5.7?

inikep:
5.5 and 5.6 had a valid warning because it was defined: 
```
const char *format = ...
char tmp[sizeof(format) + 2 * sizeof(uint) + 1];
```

5.7 had no warning because of:
```
const char *format = ...
char tmp[64 + 2 * MY_INT32_NUM_DECIMAL_DIGITS];
```

laurynas-biveinis:
So then it should be a null merge - cleanup patches without a clear objective (e.g. a warning fix) is not something we should maintain

inikep:
Changes within `sp_head.cc` were removed.

laurynas-biveinis:
The question still stands for the remaining changes

dutow:
As I see, the warning fixes are mostly about the register keyword.

* In 5.6, I included the commit which removed them (180b7e4). That commit is also part of the current 5.7 branch.
* However, in 5.7 commits after 180b7e4 again added the register keyword to some locations - so its not a merge issue, they are simply newer than the commit which removes them.

This shouldn't result in any warnings/errors in 5.7 (that's why I didn't notice them when I fixed the 5.7 warnings), as we always build 5.7 in C++03/11 mode (compared to 5.5/5.6, where we use the default language level of the compiler) to avoid some other deprecation warnings in the standard library.

inikep:
After removing changes without any visible payoff the null-merge is empty.

percona-ysorokin (approved):
LGTM

PR #2410, opened 2018-05-31
DOC-715 fix, clarification of STWCS behavior with MyRocks

fiowro:


george-lorch:
This still is not totally correct or clear to a user looking to compare. Allow me to add some color that you can try to translate into clearer comparison.

Facebook MySQL syntax is slightly different from Percona Server. Facebook MySQL syntax requires engine specification (i.e. it is NOT cross engine consistent) "START TRANSACTION WITH CONSISTENT *ROCKSDB* SNAPSHOT" whereas Percona Server implementation is consistent across engines and has simpler syntax of "START TRANSACTION WITH CONSISTENT SNAPSHOT"

Facebook MyRocks _requires_ binlog enabled (and in row format) in order to use "START TRANSACTION WITH CONSISTENT ROCKSDB SNAPSHOT", Percona Server does not. If you attempt to "S T W C R S" in Facebook MySQL, server will emit this error to client:
```
ERROR 1105 (HY000): Cannot start transaction or binlog disabled
```

TRANSACTION ISOLATION LEVEL specific differences:
1) TRANSACTION ISOLATION LEVEL = REPEATABLE READ
  Both Facebook MyRocks and Percona Server MyRocks behave the same.
  MyRocks does not allow updates/writes to snapshots. Any attempt to use any DML that would modify the snapshot will emit the SQL error:
```ERROR 12005 (HY000): Can't execute updates when you started a transaction with START TRANSACTION WITH CONSISTENT [ROCKSDB] SNAPSHOT.```
This behavior is common to both Facebook MyRocks and Percona Server MyRocks except the error message wording may be slightly different across versions..

2) TRANSACTION ISOLATION LEVEL = READ COMMITTED
  Facebook MyRocks and Percona Server MyRocks behave slightly differently:
  Facebook MyRocks in READ COMMITTED will emit the SQL error to the client 
when "START TRANSACTION WITH CONSISTENT ROCKSDB SNAPSHOT" is executed:
```
ERROR 12048 (HY000): Only REPEATABLE READ isolation level is supported for START TRANSACTION WITH CONSISTENT|SHARED|EXISTING SNAPSHOT in RocksDB Storage Engine.
```
  Percona Server MyRocks does not allow updates/writes to snapshots. Any attempt to use any DML that would modify the snapshot will emit the SQL error:
```ERROR 12005 (HY000): Can't execute updates when you started a transaction with START TRANSACTION WITH CONSISTENT [ROCKSDB] SNAPSHOT.```
  So the main difference is that Facebook MySQL emits an error on the "S T W C R S" command, Percona Server emits an error after the "S T W C S" command when any DML is executed that would violate the read only constraint.

**Comment by fiowro** [comment]
Tried to precisely reflect success/fail compatibility details disclosed in the last comment

File: doc/source/myrocks/differences.rst [hunk 1]
```diff
@@ -21,10 +21,17 @@ and includes the following differences from the original implementation:
   | InnoDB           | Success            | Success                         |
   +------------------+--------------------+---------------------------------+
   | Facebook MyRocks | Fail               | Success                         |
+  |                  |                    | (MyRocks engine only; read-only,|
+  |                  |                    | as all MyRocks engine snapshots)|
   +------------------+--------------------+---------------------------------+
-  | Percona MyRocks  | Fail               | Success for read-only snapshots |
+  | Percona MyRocks  | Fail with andy DML | Success                         |
```
george-lorch: s/andy/any

george-lorch (approved):
Approve pending minor spelling correction

PR #2416, opened 2018-06-07
Fix PS-4513 (Inconsistent SELECT...ORDER BY results with euckr char set / euckr_bin collation) (5.6)

percona-ysorokin:
https://jira.percona.com/browse/PS-4513

Added missing reset 'end_range' statement in 'ha_innobase::reset_template()'
in 'ha_innodb.cc' which is present in the fix for Bug #23481444
"OPTIMISER CALL ROW_SEARCH_MVCC() AND READ THE INDEX APPLIED BY UNCOMMITTED ROWS"
(commit mysql/mysql-server@15dbdac4)
in 8.0 but not in 5.6 and 5.7.

Added new 'main.bug91091' MTR test case which simulates the problem with
different results of 'SELECT' statements.


percona-ysorokin:
Jenkins build link
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2194/

mysql-test/include/bug91091.inc
```diff
@@ -0,0 +1,159 @@
+DELETE FROM t WHERE c2 = '*****12889aaTHLIN_ACQS_DT_S8' AND c1 = 'Eid4i*******11372012******haHLIN_PBCR_*****_******00';
```
laurynas-biveinis: Is it possible to reduce the testcase?

mysql-test/t/bug91091.test
```diff
@@ -0,0 +1,60 @@
+--source include/have_innodb.inc
```
laurynas-biveinis: Please move to innodb suite

laurynas-biveinis (changes requested):

mysql-test/t/bug91091.test
percona-ysorokin: Initially I thought that this problem was collation-related and I put the test case to the `main` suite. The problem turned out to be in InnoDB code. Moving to the `innodb` test suite.

percona-ysorokin:
MTR test case moved to the `innodb` test suite. Long `INSERT` statement reformatted.
Updated Jenkins build link
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2195

laurynas-biveinis (approved):
Please also upstream and PR to 8.0 seed branch

PR #2417, opened 2018-06-07
Fix PS-4513 (Inconsistent SELECT...ORDER BY results with euckr char set / euckr_bin collation) (5.7)

percona-ysorokin:


laurynas-biveinis (changes requested):
Same comments as for 5.6

percona-ysorokin:
Updated Jenkins build link
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1849/

laurynas-biveinis (approved):

PR #2418, opened 2018-06-11
PS-3840: With FIPS enabled and performance_schema=off, instance crash…

satya-bodapati:
…es on CREATE VIEW

Description:
-----------
MySQL Server compiled with OpenSSL (not YaSSL) and running the
server under FIPS enabled environment, leads to crash.

This is because MD5 algorithm is not FIPS compliant.

Fix:
----
Disable FIPS mode for MySQL which allows mysqld to be run under
FIPS environment.

### DIFF
--- a/mysql-test/include/have_fips.inc
+++ b/mysql-test/include/have_fips.inc
@@ -1,0 +1,30 @@
+# This inc file detects FIPS only on centOS/RHEL
+
+--source include/not_windows.inc
+
+--let OUTFILE= $MYSQLTEST_VARDIR/tmp/fips.out
+
+perl;
+use strict;
+# This file contents should only be either 0 or 1
+my $fips_file = "/proc/sys/crypto/fips_enabled";
+my $fips = 0;
+if ((-e $fips_file) && (-r $fips_file))
+{
+  local $/=undef;
+  open FILE, $fips_file or die "Couldn't open file: $!";
+  $fips = <FILE>;
+  close FILE;
+}
+
+open(FILE, ">$ENV{OUTFILE}") or die;
+print FILE "let \$fips= $fips;\n";
+close(FILE);
+EOF
+
+--source $OUTFILE
+--remove_file $OUTFILE
+
+if (!$fips) {
+  --skip Test requires FIPS enabled environment like centOS/RHEL
+}
--- a/mysql-test/t/fips.test
+++ b/mysql-test/t/fips.test
@@ -1,0 +1,15 @@
+--source include/have_fips.inc
+--echo # Check for warnings in the error log
+--let SEARCH_FILE= $MYSQLTEST_VARDIR/log/fips.err
+
+--replace_result $SEARCH_FILE SEARCH_FILE
+--let $restart_parameters=restart: --log-error=$SEARCH_FILE --no-console
+--source include/restart_mysqld.inc
+
+--let SEARCH_PATTERN = Percona Server cannot operate under OpenSSL FIPS mode. Disabling FIPS
+--source include/search_pattern_in_file.inc
+SELECT MD5('mysql running under fips');
+
+--let $restart_parameters=
+--source include/restart_mysqld.inc
+--remove_file $SEARCH_FILE
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -147,6 +147,10 @@
 #include "item_strfunc.h"               // Item_func_uuid
 #include "handler.h"
 
+#if defined(HAVE_OPENSSL) && !defined(HAVE_YASSL)
+#include <openssl/crypto.h>
+#endif
+
 #ifndef EMBEDDED_LIBRARY
 #include "srv_session.h"
 #endif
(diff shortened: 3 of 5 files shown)


sql/mysqld.cc
```diff
@@ -3664,6 +3668,16 @@ int warn_self_signed_ca()
 static int init_ssl()
 {
 #ifdef HAVE_OPENSSL
+#ifndef HAVE_YASSL
+  unsigned int fips_mode = FIPS_mode();
+  if (fips_mode != 0)
+  {
+    /* FIPS is enabled, Log warning and Disable it now */
+    sql_print_warning("MySQL cannot operate under OpenSSL FIPS mode."
```
laurynas-biveinis: While the statement is true for MySQL too in general, the product here is Percona Server

sql/mysqld.cc
```diff
@@ -3664,6 +3668,16 @@ int warn_self_signed_ca()
 static int init_ssl()
 {
 #ifdef HAVE_OPENSSL
+#ifndef HAVE_YASSL
+  unsigned int fips_mode = FIPS_mode();
```
laurynas-biveinis: OpenSSL docs suggest plain "int" as FIPS_mode return type

extra/fipstest.cc
```diff
@@ -0,0 +1,19 @@
...
+int main() {
+
+#if defined(HAVE_OPENSSL) && !defined(HAVE_YASSL)
+  unsigned int fips_mode = FIPS_mode();
+  if (fips_mode != 0) {
+    fprintf(stdout, "FIPS enabled\n");
+    return(0);
+  }
+#else
+  /* Other SSLs, we dont care */
```
laurynas-biveinis: don't

mysql-test/include/have_fips.inc
```diff
@@ -0,0 +1,21 @@
+--let FILE= $MYSQLTEST_VARDIR/tmp/fips.out
```
laurynas-biveinis: While the utility app works everywhere, I wonder if a shell-based test would be enough for all of our practical testing use cases - i.e. on CentOS, `cat /proc/sys/crypto/fips_enabled`. We probably won't be testing FIPS on Debian/Ubuntu anyway for time being

mysql-test/t/fips.test
```diff
@@ -0,0 +1,8 @@
+--source include/have_fips.inc
+--echo # Check for warnings in the error log
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
```
laurynas-biveinis: grep mysqld.1.err is an antipattern. It breaks --repeat (the 2nd invocation will just hit the 1st one's output), it will break if any of the previous tests on the same worker got the same warning. Deleting the log file to fix the above is not good neither. The best solution I know so far is to restart the server with custom error log path, grep, restart with default, delete the custom error log.

extra/fipstest.cc
```diff
@@ -0,0 +1,19 @@
...
+#if defined(HAVE_OPENSSL) && !defined(HAVE_YASSL)
+#include <openssl/crypto.h>
+#endif
+
+int main() {
+
+#if defined(HAVE_OPENSSL) && !defined(HAVE_YASSL)
+  unsigned int fips_mode = FIPS_mode();
+  if (fips_mode != 0) {
+    fprintf(stdout, "FIPS enabled\n");
```
laurynas-biveinis: Just printf I guess

extra/fipstest.cc
```diff
@@ -0,0 +1,19 @@
+#include <stdio.h>
```
laurynas-biveinis: (see the have_fips.inc review comments first)

extra/fipstest.cc
```diff
@@ -0,0 +1,19 @@
...
+#if defined(HAVE_OPENSSL) && !defined(HAVE_YASSL)
+  unsigned int fips_mode = FIPS_mode();
+  if (fips_mode != 0) {
+    fprintf(stdout, "FIPS enabled\n");
+    return(0);
+  }
+#else
+  /* Other SSLs, we dont care */
+#endif
+  fprintf(stdout, "FIPS not enabled\n");
```
laurynas-biveinis: printf

laurynas-biveinis (changes requested):

sql/mysqld.cc
satya-bodapati: Done! changed to Percona Server

extra/fipstest.cc
satya-bodapati: removed this file

mysql-test/include/have_fips.inc
satya-bodapati: Ok, my approach was more generic. I have implemented the suggestion. The change looks smaller now :)

mysql-test/t/fips.test
satya-bodapati: Fixed. the test is now named as percona_fips.test

laurynas-biveinis (changes requested):
The diff itself looks good, other comments:
- please squash commits
- please provide Jenkins run link (in fact I prefer requesting PRs for my work when Jenkins completes and I review it, others have other styles)

satya-bodapati:
Jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1842/

laurynas-biveinis:
Jenkins checks out OK. Before merging:
- please squash your commits
- please do not merge from trunk (rebase on the trunk instead but in this case there is no pressing need unless you get merge conflicts otherwise)

PR #2419, opened 2018-06-11
PS-4514: rpl.bug68490 is unstable in 5.7

inikep:
The issues here was related to restart of master replication server while slave was still connected to master.
Under heavy load `stop_slave.inc` can return Last_IO_Errno=2003 that is `error reconnecting to master 'root@127.0.0.1:13001'`.
To solve this `sync_slave_sql_with_master.inc` should be run before `stop_slave.inc`.


inikep:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1839/

laurynas-biveinis (approved):

PR #2423, opened 2018-06-12
DOC-751 (mentioning Ubuntu 18.04) with DOC-752 (percona-release 0.1.6…

fiowro:
…) fix for Ubuntu and Debian docs.


hrvojem:
This should be fixed in lower versions (5.6 and 5.6) as well, and merged from there.

PR #2426, opened 2018-06-14
PS-4521: auto_ptr deprecated in C++11/C++14 (GCC6 default)

inikep:
Upstream uses "-std=gnu++03" for g++-6 and newer to reduce number of warnings.
This commit fixes a regression in Percona Server that "-std=gnu++03"
was not set with CMake older that 3.1 because `cmake-compile-features` were introduced in CMake 3.1:
https://cmake.org/cmake/help/v3.1/manual/cmake-compile-features.7.html#compile-feature-requirements


inikep:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1844/

To test the commit properly we need g++-6 or newer and CMake older that 3.1. None of our Jenkins and Travis test configurations meet these requirements. Tested locally with CentOS 6, gcc-6.3.1, and cmake 2.8.12.2.

laurynas-biveinis (approved):

laurynas-biveinis:
@inikep, please also edit the JIRA issue title to include too-old cmake cause

PR #2429, opened 2018-06-21
PS-3851: Percona Ver 5.6.39-83.1 Failing assertion: sym_node->table !…

satya-bodapati:
…= NULL

Problem:
--------
A DDL on table with FTS index with concurrent FTS sync leads to crash.
When table rebuilding ALTERs happen on table with FTS indexes, a new
copy of table is created, the rows are inserted. This copy of table
is created with a temporary name and renamed to the original table
name at end of ALTER.

FTS sync thread stores the direct reference of intermediate
table before rename (fts_table_t->parent pointing to dict_table_t->name.m_name)
and later tries to acccess this intermediate table after rename.

When the table is renamed, the old name is freed and new name is
allocated. The inter-leaving FTS sync thread still has access to the
freed location.

As part of fts sync, it writes the tokenized words to FTS AUX tables.
It is executed as internal SQL and as part of internal SQL, FTS sync
deferences freed location, builds a wrong AUX INDEX table name.

Hence the table is neither found in cache nor it can load from dictionary.
The assert says table must have been found.

Fix:
----
When renaming a table, wait for a running sync to finish.

Jenkins: https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2197/

### DIFF
--- a/mysql-test/suite/innodb_fts/t/sync.test
+++ b/mysql-test/suite/innodb_fts/t/sync.test
@@ -167,4 +167,30 @@ SELECT * FROM t1 WHERE MATCH(title) AGAINST ('mysql database good');
 
 DROP TABLE t1;
 
+--echo #
+--echo # PS-3851: Percona Ver 5.6.39-83.1 Failing assertion: sym_node->table != NULL
+--echo #
+CREATE TABLE `questions_and_answers` (
+  `id` int(11) unsigned NOT NULL AUTO_INCREMENT COMMENT 'Unique ID',
+  `keywords` text COMMENT 'CSV list of keywords related to this entry',
+  PRIMARY KEY (`id`),
+  FULLTEXT KEY `keywords` (`keywords`)
+) ENGINE=InnoDB AUTO_INCREMENT=3620 DEFAULT CHARSET=utf8 COMMENT='Main table for Q&A module ';
+
+INSERT INTO `questions_and_answers` (`id`,`keywords`) VALUES (1854,'support, services, essay, writing, maths, help, advice, study, advice, advise, careers, employability, graduating, graduate, student_support, academic_support, academic_help');
+INSERT INTO `questions_and_answers` (`id`,`keywords`) VALUES (1855,'accommodation, housing, halls, residence, live, living, bed, beds, acomodation, accomodation, accommidation, acomidation, hall, campus_living, uni_living, uni_halls, university_life, university_accommodation');
+INSERT INTO `questions_and_answers` (`id`,`keywords`) VALUES (1856,'interview, interviews, audition, auditions, portfolio, review, admission, interview_preparation, booking, book, change_interview');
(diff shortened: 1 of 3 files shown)


storage/innobase/row/row0mysql.cc
```diff
@@ -5500,6 +5500,19 @@ row_rename_table_for_mysql(
 		goto funct_exit;
 	}
 
+
+	/* Wait for background fts sync to finish */
+	for (retry = 1; dict_fts_index_syncing(table); ++retry) {
+		DICT_BG_YIELD(trx);
+		if (retry % 100 == 0) {
+			fprintf(stderr,
```
laurynas-biveinis: ib_logf (the upstream fixes are not OK in this regard neither)

storage/innobase/row/row0mysql.cc
```diff
@@ -5500,6 +5500,19 @@ row_rename_table_for_mysql(
 		goto funct_exit;
 	}
 
+
+	/* Wait for background fts sync to finish */
+	for (retry = 1; dict_fts_index_syncing(table); ++retry) {
```
laurynas-biveinis: I don't quite understand how the accesses to dict_index_t::index_fts_syncing itself are synchronized - the writes are protected by fts_cache_t::lock, and reads by nothing? If so, that looks questionable, but not our job to fix

mysql-test/suite/innodb_fts/t/sync.test
```diff
@@ -168,3 +168,1190 @@ SELECT * FROM t1 WHERE MATCH(title) AGAINST ('mysql database good');
 DROP TABLE t1;
 
 --source include/wait_until_count_sessions.inc
+
+--echo #
```
laurynas-biveinis: Is it possible to have a deterministic DEBUG_SYNC/DBUG_EXECUTE_IF testcase?

mysql-test/suite/innodb_fts/t/sync.test
```diff
@@ -168,3 +168,1190 @@ SELECT * FROM t1 WHERE MATCH(title) AGAINST ('mysql database good');
 DROP TABLE t1;
 
 --source include/wait_until_count_sessions.inc
```
laurynas-biveinis: The new testcase - if it stays in this file - should probably go above this line

laurynas-biveinis (changes requested):
Please unlink the upstream bug (and file a new one, or ask OP to file a new one) - the original upstream bug seems to be about a crash in FTS starting with 5.6.10, which cannot be caused by the same bisected commit

mysql-test/suite/innodb_fts/t/sync.test
satya-bodapati: One of the thread involved is background thread. AFAIR, DEBUG_SYNC cannot control background threads.

storage/innobase/row/row0mysql.cc
satya-bodapati: Right, I don't like this variable and sync mechanism. May be a rw_lock or MDL lock on table should be used. The changes could be huge and will differ from upstream syncing logic.

mysql-test/suite/innodb_fts/t/sync.test
laurynas-biveinis: Yep, hence the question also involves DBUG_EXECUTE_IF

satya-bodapati:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2197/

laurynas-biveinis (approved):

PR #2431, opened 2018-06-21
PS-3702: Unnecessary flush list iteration during ALTER TABLE ADD INDEX

satya-bodapati:
Problem:
    --------
    ALTER TABLE ADD INDEX in 5.7 uses bottom-up index build(BulkInsert).
    It skips redo logging and flushes all dirty pages dirtied by ALTER.
    FlushObserver is used which flushes and waits until dirty pages
    are flushed and written to disk.
    
    There are two iterations of flush list as part of ALTER:
    1. For PFS monitoring, an estimate of dirty pages is required.
       (buf_flush_get_dirty_pages_count())
    2. To do the actual flushing of dirty pages of ALTER
    
    When there are lot of dirty pages, this iteration takes significant
    amount of time.
    
    Fix:
    ----
    1. Remove one iteration of flush list of all buffer pool instances.
       i.e the iteration used for estimate.
    
    2. Note the LSN at which Observer is created. When mtr commits
       with an Observer, as part of commit, dirty pages
       are added to flush list. Count the dirty pages that are dirted
       before Observer (newest_modification < observer_lsn) and also
       the new blocks that are dirtied. This count serves
       as a very good estimate
    
    3. Make buf_flush_get_dirty_pages_count() debug only function.
       This is no longer required in release builds.

```
Performance results with 200GB buffer pool
-----------------------------------------------------------
bp     ps        ps            ps
dirty  5.7.21    5.7.21+fix    5.6.40
pct
0      0.03       0.02          0.01
5      0.42       0.29          0.48
10     1.21       0.66          0.05
15     1.70       0.85          0.04
20     2.23       1.16          0.12
25     2.75       1.50          0.04
30     3.17       1.42          0.16
35     4.13       2.07          0.02
40     3.51       1.77          0.04
45     5.36       2.11          0.20
50     4.61       2.23          0.18
55     6.13       2.93          0.29
60     6.19       3.34          0.30
65     6.37       2.91          0.03
70     7.86       4.39          0.05

```
jenkins: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1863/

### DIFF
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -3819,6 +3819,7 @@ buf_flush_validate(
 #endif /* UNIV_DEBUG || UNIV_BUF_DEBUG */
 #endif /* !UNIV_HOTBACKUP */
 
+#ifdef UNIV_DEBUG
 /******************************************************************//**
 Check if there are any dirty pages that belong to a space id in the flush
 list in a particular buffer pool.
@@ -3880,6 +3881,7 @@ buf_flush_get_dirty_pages_count(
 
 	return(count);
 }
+#endif /* UNIV_DEBUG */
 
 /** FlushObserver constructor
 @param[in]	space_id	table space id
@@ -3895,7 +3897,9 @@ FlushObserver::FlushObserver(
 	m_space_id(space_id),
 	m_trx(trx),
 	m_stage(stage),
-	m_interrupted(false)
+	m_interrupted(false),
+	m_estimate(),
+	m_lsn(log_get_lsn())
 {
 	m_flushed = UT_NEW_NOKEY(std::vector<ulint>(srv_buf_pool_instances));
 	m_removed = UT_NEW_NOKEY(std::vector<ulint>(srv_buf_pool_instances));
@@ -3987,9 +3991,7 @@ FlushObserver::flush()
 		buf_remove = BUF_REMOVE_FLUSH_WRITE;
 
 		if (m_stage != NULL) {
-			ulint	pages_to_flush =
-				buf_flush_get_dirty_pages_count(
-					m_space_id, this);
+			ulint	pages_to_flush = get_estimate();
 
 			m_stage->begin_phase_flush(pages_to_flush);
 		}
@@ -4004,5 +4006,16 @@ FlushObserver::flush()
 
 			os_thread_sleep(2000);
 		}
+	}
+}
+
+/** Increase the estimate of dirty pages by this observer
+@param[in]	block		buffer pool block */
+void
+FlushObserver::inc_estimate(const buf_block_t*	block)
+{
+	if (block->page.oldest_modification == 0
+	    || block->page.newest_modification < m_lsn) {
+		os_atomic_increment_ulint(&m_estimate, 1);
 	}
 }
--- a/storage/innobase/include/buf0flu.h
+++ b/storage/innobase/include/buf0flu.h
@@ -314,6 +314,7 @@ buf_flush_ready_for_flush(
 	buf_flush_t	flush_type)/*!< in: type of flush */
 	MY_ATTRIBUTE((warn_unused_result));
 
+#ifdef UNIV_DEBUG
 /******************************************************************//**
 Check if there are any dirty pages that belong to a space id in the flush
 list in a particular buffer pool.
@@ -332,6 +333,7 @@ buf_flush_get_dirty_pages_count(
 /*============================*/
 	ulint		id,		/*!< in: space id to check */
 	FlushObserver*	observer);	/*!< in: flush observer to check */
+#endif /* UNIV_DEBUG */
 
 /*******************************************************************//**
 Signal the page cleaner to flush and wait until it and the LRU manager clean
@@ -400,6 +402,17 @@ public:
 	void notify_remove(
 		buf_pool_t*	buf_pool,
 		buf_page_t*	bpage);
+
+	/** Increase the estimate of dirty pages by this observer
+	@param[in]	block		buffer pool block */
+	void inc_estimate(const buf_block_t*	block);
+
+	/** @return estimate of dirty pages to be flushed */
+	ulint get_estimate() const {
+		os_rmb;
+		return(m_estimate);
+	}
+
 private:
 	/** Table space id */
 	ulint			m_space_id;
@@ -422,6 +435,13 @@ private:
 
 	/* True if the operation was interrupted. */
 	bool			m_interrupted;
+
+	/* Estimate of pages to be flushed */
+	ulint			m_estimate;
+
+	/** LSN at which observer started observing. This is
+	used to find the dirty blocks that are dirtied before Observer */
+	const lsn_t		m_lsn;
 };
 
 #endif /* !UNIV_HOTBACKUP */
--- a/storage/innobase/include/buf0flu.ic
+++ b/storage/innobase/include/buf0flu.ic
@@ -82,6 +82,10 @@ buf_flush_note_modification(
 	mutex_enter(&block->mutex);
 
 	ut_ad(block->page.newest_modification <= end_lsn);
+	if (observer != NULL) {
+		observer->inc_estimate(block);
+	}
+
 	block->page.newest_modification = end_lsn;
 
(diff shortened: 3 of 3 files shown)


storage/innobase/include/buf0flu.h
```diff
@@ -314,6 +314,7 @@ buf_flush_ready_for_flush(
 	buf_flush_t	flush_type)/*!< in: type of flush */
 	MY_ATTRIBUTE((warn_unused_result));
 
+#ifdef UNIV_DEBUG
```
laurynas-biveinis, line 317: Might as well remove the prototype and make the definition static

storage/innobase/include/buf0flu.h
```diff
@@ -422,6 +435,9 @@ class FlushObserver {
 
 	/* True if the operation was interrupted. */
 	bool			m_interrupted;
+
+	/* Estimate of pages to be flushed */
+	ulint			m_estimate;
```
laurynas-biveinis, line 440: Where is it initialized to zero?

storage/innobase/include/buf0flu.h
```diff
@@ -400,6 +402,17 @@ class FlushObserver {
...
 		buf_pool_t*	buf_pool,
 		buf_page_t*	bpage);
+
+	/** Increase the estimate of dirty pages by this observer */
+	void inc_estimate() {
+		os_atomic_increment_ulint(&m_estimate, 1);
+	}
+
+	/** @return esimtate of dirty pages to be flushed */
+	ulint get_estimate() {
```
laurynas-biveinis: const method

storage/innobase/include/buf0flu.h
```diff
@@ -400,6 +402,17 @@ class FlushObserver {
...
+		return(m_estimate);
```
laurynas-biveinis: In the absence of std::atomic<>::load, I think you need os_rmb above it

storage/innobase/include/buf0flu.ic
```diff
@@ -89,6 +89,9 @@ buf_flush_note_modification(
 	ut_ad(block->page.flush_observer == NULL
 	      || block->page.flush_observer == observer);
 	block->page.flush_observer = observer;
+        if (observer != NULL) {
```
laurynas-biveinis: I re-read our Friday discussion, and reviewed FlushObserver usage in the code, but I am still unable to convince myself why is this correct - it seems to depend on the page write pattern with FlushObserver active, that there is roughly one mtr per page dirtied. Is this the case with this and other uses?

laurynas-biveinis (changes requested):

storage/innobase/include/buf0flu.ic
satya-bodapati: Right, PageBulk in btr0bulk.h has one mtr and pageBulk is per page. The reason I was confident that many duplicates wont exist in this pattern is because a new page is not allocated until the last one on each level is full. 

Currently FlushObserver is not used anywhere else except ALTER ADD INDEX.

But I have another idea (this is easier to understand, general purpose and should satisfy the requirements)

1. Create observer_lsn. The LSN at which flushobserver is created. (Assign log_sys->lsn to observer_lsn)
2. For mtrs with observers, add the page to tracking (increment number of pages) if
    if (oldest_modification == 0 /* new blocks that are added to flush list */
        || oldest_modification < observer->m_observe_lsn /* existing blocks that are dirtied before flush observer */
        observer->inc_estimate() 
   }

This becomes very accurate for an estimate ;) but I am OK with implementing this :)

laurynas-biveinis: I like this suggestion

satya-bodapati: Ok, will implement this

storage/innobase/include/buf0flu.h
satya-bodapati: Also used in buf0lru.cc, cannot be made static.

storage/innobase/include/buf0flu.h
satya-bodapati: Fixed now. Added to initializer list in constructor

storage/innobase/include/buf0flu.ic
satya-bodapati: One small correction, observer->lsn() should be compared with newest modification not oldest modification.

storage/innobase/buf/buf0flu.cc
```diff
@@ -3895,8 +3897,10 @@ FlushObserver::FlushObserver(
 	m_space_id(space_id),
 	m_trx(trx),
 	m_stage(stage),
-	m_interrupted(false)
+	m_interrupted(false),
+	m_estimate()
 {
+	m_lsn = log_get_lsn();
```
laurynas-biveinis: Move as m_lsn(log_get_lsn()) to the constructor initializer list

storage/innobase/include/buf0flu.h
```diff
@@ -400,6 +402,23 @@ class FlushObserver {
 	void notify_remove(
 		buf_pool_t*	buf_pool,
 		buf_page_t*	bpage);
+
+	/** Increase the estimate of dirty pages by this observer */
+	void inc_estimate() {
+		os_atomic_increment_ulint(&m_estimate, 1);
+	}
+
+	/** @return esimtate of dirty pages to be flushed */
```
laurynas-biveinis: "estimate"

storage/innobase/include/buf0flu.h
```diff
@@ -400,6 +402,23 @@ class FlushObserver {
...
+	ulint get_estimate() const {
+		os_rmb;
+		return(m_estimate);
+	}
+
+	/** @return the observer LSN */
+	lsn_t lsn() const {
```
laurynas-biveinis: This LSN should be internal observer business, please see a comment at buf_flush_note_modification

storage/innobase/include/buf0flu.h
```diff
@@ -422,6 +441,13 @@ class FlushObserver {
 
 	/* True if the operation was interrupted. */
 	bool			m_interrupted;
+
+	/* Estimate of pages to be flushed */
+	ulint			m_estimate;
+
+	/** LSN at which observer started observing. This is
+	used to find the dirty blocks that are dirtied before Observer */
+	lsn_t			m_lsn;
```
laurynas-biveinis: const lsn_t, it cannot change after the constructor

storage/innobase/include/buf0flu.ic
```diff
@@ -90,6 +91,15 @@ buf_flush_note_modification(
 	      || block->page.flush_observer == observer);
 	block->page.flush_observer = observer;
 
+        if (observer != NULL) {
+		/* Count all new blocks and blocks that are dirtied
+		before Observer started and dirtied again */
+		if (block->page.oldest_modification == 0
+		    || newest_modification < observer->lsn()) {
```
laurynas-biveinis: Caller should not care about observer's LSN, the observer itself should - i.e. what do you think about passing newest_modification to inc_estimate and moving the corresponding part of if statement there?

laurynas-biveinis (changes requested):

storage/innobase/buf/buf0flu.cc
satya-bodapati: Fixed

storage/innobase/include/buf0flu.h
satya-bodapati: Fixed

storage/innobase/include/buf0flu.h
satya-bodapati: Fixed

storage/innobase/include/buf0flu.h
satya-bodapati: Fixed

storage/innobase/include/buf0flu.ic
satya-bodapati: Right, I passed the entire block object. Looks cleaner now

satya-bodapati:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1863/

laurynas-biveinis (approved):

PR #2433, opened 2018-06-25
PS-4543: MTR: Use gnome-terminal with --gdb option instead of xterm

satya-bodapati:
Problem
----------
When using --gdb option with mtr, it spawns xterm as terminal window
by default.

xterm is really outdated.
1. Can't easily copy text from it.
2. Text sometimes appear small by deafult
3. Window buffer isn't resized properly on maximize

Fix:
----
On linux systems, where gnome-termnial (/usr/bin/gnome-terminal) is
available, try to use it.

Introduce --gterm option. This is to be used along with --gdb option.

Ex: ./mtr --mem main.1st --gdb --gterm

### DIFF
--- a/mysql-test/mysql-test-run.pl
+++ b/mysql-test/mysql-test-run.pl
@@ -265,6 +265,7 @@ our $opt_manual_ddd;
 our $opt_manual_debug;
 our $opt_debugger;
 our $opt_client_debugger;
+our $opt_gterm;
 
 my $config; # The currently running config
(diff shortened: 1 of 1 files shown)


laurynas-biveinis (approved):
Feel free to backport to lower versions when convenient

laurynas-biveinis (changes requested):
Sorry, forgot: please file a JIRA ticket and edit the commit message to refer to it

laurynas-biveinis (approved):

PR #2435, opened 2018-06-26
[PS-4546] Set unlimited TimeoutSec for ubuntu/debian

EvgeniyPatlan:


laurynas-biveinis (changes requested):
Packaging bugs must be public too - please move to PS and adjust the commit message accordingly

EvgeniyPatlan:
@laurynas-biveinis done. Also I got confirmation from upstream that TimeoutSec should be disabled

abychko (approved):
looks good to me

tplavcic:
@EvgeniyPatlan How will this work if I specify wrong option in the config file - will it fail immediately or wait unlimited ?

EvgeniyPatlan:
@tplavcic It shouldn't wait timeout and will fail immediately

PR #2436, opened 2018-06-26
PS-4545: protobuf compilation issue with Travis CI

inikep:
Use an explicit version of the protobuf library.


percona-ysorokin (approved):
LGTM

laurynas-biveinis (changes requested):
Please add a comment why - it's impossible to tell that e.g. for 8.0 we can switch to the latest version again just by looking at Travis config

inikep:
I added a comment.

laurynas-biveinis (approved):

PR #2437, opened 2018-06-27
Fix PS-4539 (7.5Gb RAM should be enough to run xtradb_compressed_columns_consistency_antelope_compact test) (5.6)

percona-ysorokin:
https://jira.percona.com/browse/PS-4539

Reorganized
'main.xtradb_compressed_columns_consistency_<file_format>_<row_format>'
series of MTR test cases. Instead of testing the whole set of data types
('VARCHAR', 'NATIONAL VARCHAR', 'VARBINARY', 'TINYBLOB', 'TINYTEXT', 'BLOB',
'TEXT', 'MEDIUMBLOB', 'MEDIUMTEXT', 'LONGBLOB', 'LONGTEXT') in a single test
file, we now have them split into several
'main.xtradb_compressed_columns_consistency_<file_format>_<row_format>_set<n>'
test cases:
* 'set1' - ('VARCHAR', 'NATIONAL VARCHAR', 'VARBINARY');
* 'set2' - ('TINYBLOB', 'TINYTEXT', 'BLOB', 'TEXT');
* 'set3' - ('MEDIUMBLOB', 'MEDIUMTEXT', 'LONGBLOB', 'LONGTEXT').
This helped to significantly decrease the number of nested MTR file includes
('--source xxx.inc') and therefore decrease memory consumption of 'mysqltest'
utility from 1.4 GB to 400 MB ('set1') and 600 MB ('set2' and 'set3').

In addition, only 'set3' tests are now marked with
'--source include/big_test.inc'. 'set1' and 'set2' ones will be run without
specifying '--big-test'.

Additional server optimizations.
* Decreased '--innodb-log-file-size' from 256M to 192M.
* To save some server resources we no longer set '--max-allowed-packet' to
  48M at server restart, instead we set this system variable dynamically
  depending on the '$field_length' (the length of the value being inserted).
* In order to completely deallocate memory resources allocated for
  '@inserted_value' variable
  ("eval SET @inserted_value = REPEAT('z', $current_value_length);")
  we now establish / close a new connection for each
  'CREATE TABLE' - 'INSERT VALUE' - 'SELECT AND ASSERT' - 'DELETE TABLE'
  cycle. Moreover, this connection is established with 'COMPRESS' flag which
  decreases the amount of data transferred via network.


percona-ysorokin:
Jenkins build link (--do-test=xtradb_compressed_columns_consistency --big-test --repeat=4)
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2199/

laurynas-biveinis (approved):

PR #2440, opened 2018-06-29
PS-4524: innodb-optimize-keys fails when variable name is a prefix (5.5)

inikep:
Fixed by checking lengths of compared strings (column names).

https://ps.cd.percona.com/view/5.5/job/percona-server-5.5-param/18/

### DIFF
--- a/client/mysqldump.c
+++ b/client/mysqldump.c
@@ -2544,6 +2544,7 @@ static const char *parse_quoted_identifier(const char *str,
   considered indexed for such key specification.
 */
 static my_bool contains_autoinc_column(const char *autoinc_column,
+                                       ssize_t autoinc_column_len,
                                        const char *keydef,
                                        key_type_t type)
 {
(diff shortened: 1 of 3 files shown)


mysql-test/t/percona_mysqldump_innodb_optimize_keys.test
```diff
@@ -392,6 +392,46 @@ INSERT INTO t1(`id`) VALUES (4);
...
+) ENGINE=InnoDB;
+
+INSERT INTO t2 VALUES (1,10), (2,11), (3,12);
+INSERT INTO t2(`prefix`) VALUES (13);
+
+
+--exec $MYSQL_DUMP --skip-comments --innodb-optimize-keys test t1 t2 >$file
+
+--echo ######################################
+--cat_file $file
```
laurynas-biveinis, line 424: I am not sure we need the dump itself in the output - I'd test after loading the dump SHOW CREATE TABLE instead

client/mysqldump.c
```diff
@@ -2544,6 +2544,7 @@ static const char *parse_quoted_identifier(const char *str,
   considered indexed for such key specification.
 */
 static my_bool contains_autoinc_column(const char *autoinc_column,
+                                       intptr_t autoinc_column_len,
```
laurynas-biveinis: size_t

laurynas-biveinis (changes requested):

inikep:
`ssize_t` used instead of `intptr_t`.

https://ps.cd.percona.com/view/5.5/job/percona-server-5.5-param/19/

laurynas-biveinis (approved):

PR #2443, opened 2018-06-29
PS-4562: PS compilation with GCC 7.3.0 + ASAN failure (5.7)

inikep:


laurynas-biveinis (commented):
Why different from 5.5/5.6?

inikep:
The issue from 5.5 and 5.6 was fixed in 5.7 by upstream.
`m_sort_length` occurs only in 5.7.

laurynas-biveinis (approved):
OK, it looks like 5.5/5.6 and 5.7 issues are completely unrelated and should have been logged separately. In any case LGTM now

PR #2454, opened 2018-07-09
Fixed PS-4581 (stack-use-after-scope in _db_enter_() / mysql_select_db() detected by ASan) (5.5)

percona-ysorokin:
https://jira.percona.com/browse/PS-4581

Function 'dbConnect()' in 'client/mysqlcheck.c' has 'DBUG_ENTER("dbConnect")'
at the beginning but ends with plain 'return' which causes stack corruption
in Debug mode.

Fixed by changing plain 'return's to 'DBUG_RETURN()'s.


laurynas-biveinis (approved):
Excellent, please report upstream and send the patch

PR #2459, opened 2018-07-12
PS-4565 Inconsistent behaviour of innodb_temp_tablespace_encrypt=ON a…

satya-bodapati:
…nd CREATE TABLE with ENCRYPTION option

Problem:
-------
Encryption behaviour of temporary tables is inconsistent. When a
temp tablespace is encrypted, CREATE TEMPORARY TABLE t1() ENCRYPTION='Y'
fails.

Also temporary table do not adhere to the behaviour of innodb_encrypt_tables;

Fix:
----
When innodb_temp_tablespace_encrypt is ON

CREATE TEMPORARY TABLE t1 (a INT) ENCRYPTION='Y' :- table created and encrypted
CREATE TEMPORARY TABLE t1 (a INT) ENCRYPTION='N' :- error
innodb_encrypt_tables=OFF && CREATE TEMPORARY TABLE t1 (a INT) :- error
innodb_encrypt_tables=ON && CREATE TEMPORARY TABLE t1 (a INT) :- table created and encrypted

When innodb_temp_tablespace_encrypt is OFF
CREATE TEMPORARY TABLE t1 (a INT) ENCRYPTION='Y' :- error
CREATE TEMPORARY TABLE t1 (a INT) ENCRYPTION='N' :- table created and unencrypted
innodb_encrypt_tables=OFF && CREATE TEMPORARY TABLE t1 (a INT) :- table created and unencrypted
innodb_encrypt_tables=ON && CREATE TEMPORARY TABLE t1 (a INT) :- error

Instrinsic temp tables go by the temp tablespace property.

### DIFF
--- a/mysql-test/suite/innodb/r/temp_table_encrypt_keyring_file.result
+++ b/mysql-test/suite/innodb/r/temp_table_encrypt_keyring_file.result
@@ -9,19 +9,20 @@ Pattern found.
 DROP TABLE t04;
 SET GLOBAL innodb_temp_tablespace_encrypt = ON;
 CREATE TEMPORARY TABLE t02 (a INT) ENGINE=InnoDB ENCRYPTION='Y';
-ERROR HY000: InnoDB: Unsupported encryption option for temporary tables.
-CREATE TEMPORARY TABLE t03 (a TEXT) ENGINE=InnoDB ENCRYPTION='N';
+CREATE TEMPORARY TABLE t03 (a TEXT) ENGINE=InnoDB ENCRYPTION='Y';
 INSERT INTO t03 VALUES ('Curabitur laoreet, velit non interdum venenatis');
 Pattern not found.
+SET GLOBAL innodb_encrypt_tables=ON;
 CREATE TEMPORARY TABLE t04 (a TEXT) ENGINE=InnoDB ROW_FORMAT=COMPRESSED;
+SET GLOBAL innodb_encrypt_tables=OFF;
 INSERT INTO t04 VALUES ('Praesent tristique eros a tempus fringilla');
 Pattern not found.
 CREATE TEMPORARY TABLE t05 (a INT) ENGINE=InnoDB ROW_FORMAT=COMPRESSED ENCRYPTION='y';
-ERROR HY000: InnoDB: Unsupported encryption option for temporary tables.
+INSERT INTO t05 VALUES (1), (2), (3);
 CREATE TEMPORARY TABLE t06 (a INT) ENGINE=InnoDB ROW_FORMAT=COMPRESSED ENCRYPTION='n';
-INSERT INTO t06 VALUES (1), (2), (3);
+ERROR HY000: InnoDB: Tablespace `innodb_temporary` can contain only an ENCRYPTED tables.
 SET GLOBAL innodb_temp_tablespace_encrypt = OFF;
-CREATE TEMPORARY TABLE t07 (a INT) ENGINE=InnoDB;
+CREATE TEMPORARY TABLE t07 (a INT) ENGINE=InnoDB ENCRYPTION='Y';
 INSERT INTO t07 VALUES (1), (2), (3);
 SET GLOBAL innodb_temp_tablespace_encrypt = ON;
 CREATE TABLE t10 (a INT AUTO_INCREMENT PRIMARY KEY, b INT);
@@ -44,16 +45,16 @@ DROP INDEX t10_b ON t10;
 CREATE INDEX t10_b ON t10 (b) ALGORITHM=COPY;
 DROP INDEX t10_b ON t10 ALGORITHM=COPY;
 # restart:<hidden args>
-CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB;
+CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB ENCRYPTION='Y';
 INSERT INTO t01 VALUES (1), (2), (3);
 # Kill and restart:<hidden args>
-CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB;
+CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB ENCRYPTION='Y';
 INSERT INTO t01 VALUES (1), (2), (3);
 # restart:<hidden args>
 CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB;
 INSERT INTO t01 VALUES (1), (2), (3);
 # restart:<hidden args>
-CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB;
+CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB ENCRYPTION='Y';
 INSERT INTO t01 VALUES (1), (2), (3);
 DROP TABLE t10;
 # restart: --innodb-temp-tablespace-encrypt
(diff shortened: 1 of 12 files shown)


mysql-test/suite/innodb/r/temp_table_encrypt_keyring_file.result
```diff
@@ -9,19 +9,20 @@ Pattern found.
 DROP TABLE t04;
 SET GLOBAL innodb_temp_tablespace_encrypt = ON;
 CREATE TEMPORARY TABLE t02 (a INT) ENGINE=InnoDB ENCRYPTION='Y';
-ERROR HY000: InnoDB: Unsupported encryption option for temporary tables.
```
gl-sergei, line 12: This is not review, just a comment. `temp_table_encrypt_keyring_vault.result` should also be re-recorded I guess.

satya-bodapati: right, missed it

satya-bodapati:
Jenkins result: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1870/

Fixed the two testcase failures(result content mismatch): innodb.table_encrypt_1 and innodb.table_encrypt_3

gl-sergei:
@satya-bodapati @robgolebiowski Is it indeed desired behaviour? Should not `innodb_temp_tablespace_encrypt=ON` by itself force encryption of temporary tables without respect to `innodb_encrypt_tables`?

robgolebiowski:
I think innodb_temp_tablespace_encrypt should not depend on innodb_encrypt_tables

robgolebiowski:
@gl-sergei @satya-bodapati I think Sergei is right and innodb_temp_tablespace_encrypt should force encryption of temp tables by itself. Also I am wondering if innodb_temp_tablespace_encrypt should not have 'FORCE' option? With current proposition ON is behaving like FORCE - there is not option of overriding encryption and creating temp table without encryption when innodb_temp_tablespace_encrypt is ON.

satya-bodapati:
I think it is about how you view the encryption property.

If you view encryption as table property, then the current innodb_encrypt_tables is the right way to affect
table attribute. It doesn't matter which tablespace they go. (current implementation)

If you view encryption as tablespace property, then innodb_encrypt_tables shouldn't have any affect and tables in shared tablespace will inherit tablespace property automatically. (your proposal).

We seem to caught in-between. Sometimes we want to use innodb_encrypt_tables, sometimes we want to automatically inherit the tablespace property. There is lack of consistency.

May be find middle ground and make the behaviour consistent for 'all shared' tablespaces. system, temp, general.

With the new proposal:
innodb_encrypt_tables -> This option is only for file_per_table tables.
For shared tablespace tables (system, temp, general): when explicit encryption property is not specified, they inherit the tablespace property.

Do we agree to this? @gl-sergei @robgolebiowski  @laurynas-biveinis

laurynas-biveinis:
I share @gl-sergei @robgolebiowski ' concerns and IMHO @satya-bodapati proposal is the right one to limit i_e_t to file_per_table and make shared tablespaces tables follow their tablespace setting.

satya-bodapati:
After discussion, we decided to use innodb_encrypt_tables for shared tablespaces as well. The reason is once innodb_encrypt_tables=ON, user expects all tables to created are encrypted, unless overridden by user with explicit encryption='N'

@laurynas-biveinis @gl-sergei hope you can proceed with the review now.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -10697,6 +10697,41 @@ innodb_base_col_setup_for_stored(
 	}
 }
 
+/** If encryption is requested, check for master key availability
+and set the encyrption flag in table flags
```
laurynas-biveinis: "encyrption"

storage/innobase/handler/ha_innodb.cc
```diff
@@ -10697,6 +10697,41 @@ innodb_base_col_setup_for_stored(
...
+	/* Set the encryption flag. */
+	byte*			master_key = NULL;
+	ulint			master_key_id;
+	Encryption::Version	version;
+
+	/* Check if keyring is ready. */
+	Encryption::get_master_key(
+		&master_key_id, &master_key, &version);
+
+	if (master_key == NULL) {
```
laurynas-biveinis: Move err declaration here

storage/innobase/include/fsp0space.h
```diff
@@ -206,6 +206,12 @@ class Tablespace {
 	/** Check if undo tablespace.
 	@return true if undo tablespace */
 	static bool is_undo_tablespace(ulint id);
+
+	/** @return true if tablespace is encrypted */
+	bool is_encrypted() {
```
laurynas-biveinis: const method

laurynas-biveinis (changes requested):

storage/innobase/handler/ha_innodb.cc
satya-bodapati: fixed

storage/innobase/handler/ha_innodb.cc
satya-bodapati: fixed

storage/innobase/include/fsp0space.h
satya-bodapati: fixed

mysql-test/suite/innodb/r/temp_table_encrypt_keyring_file.result
satya-bodapati: this test cannot be run as of now because of some dependency with key_ring vault plugin. Robert promised me to ping me when the test setup is fixed. Will fix that when it is ready

sql/sql_tmp_table.cc
```diff
@@ -2256,6 +2256,9 @@ bool create_innodb_tmp_table(TABLE *table, KEY *keyinfo)
   create_info.row_type= table->s->row_type;
   create_info.options|= HA_LEX_CREATE_TMP_TABLE |
                         HA_LEX_CREATE_INTERNAL_TMP_TABLE;
+
+  table->file->adjust_create_info_for_frm(&create_info);
```
gl-sergei, line 2260: `ha_innobase::adjust_create_info_for_frm` is looking at `srv_encrypt_tables`, but for tmp tables we should look at the `temp_tablespace_encrypt`?

gl-sergei (changes requested):
Test `keyring_vaut` needs to be re-recorded everything else looks good.

sql/sql_tmp_table.cc
gl-sergei: ah. I missed that `adjust_create_info_for_frm` is changed

satya-bodapati: No, the conclusion from yesterday's discussion with Robert & Laurynas is:
1. innodb_temp_tablespace_encrypt is only meant to enable/disable encryption of temp tablespace. Not for decision to encrypt temp table or not
2. temporary, system and general tablespaces should follow same rules/pattern
3. When tables lacks explicit encryption option, innodb_encrypt_tables is the only option to decide whether the table should be encrypted or not.

satya-bodapati:
Regarding the keyring_vaut, there is problem with the test /test setup and it cannot be run. So I cannot record/adjust the test.
Waiting for fix from Robert.

sql/sql_tmp_table.cc
gl-sergei: yes, I remember that. My original comment was about intrinsic tables, but later I noticed that their creation flag is dependant on tablespace encryption flag, which looks right.

satya-bodapati: Right, intrinsic should go by tablespace property.

satya-bodapati:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1875/

laurynas-biveinis (approved):

PR #2460, opened 2018-07-12
PS-3822: InnoDB system tablespace encryption

satya-bodapati:
Encryption will be done using Master Key encryption. The decision to make system
tablespace encrypted can be done only at bootstrap.

A new variable “innodb_sys_tablespace_encrypt” will be introduced to encrypt
system tablespace. Default is OFF. This variables has to be turned ON and
passed as bootstrap parameter to encrypt system tablespace

Variables introduced:
innodb_sys_tablespace_encrypt
innodb_parallel_dblwr_encrypt

A new option innodb_parallel_dblwr_encrypt is introduced to encrypt parallel
doublewrite file. Default is OFF and the option is dynamic.

When turned on, the pages in parallel doublewrite buffer are encrypted using
the respective tablespace key.


satya-bodapati:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1871/
the temp table encrypton failures will be fixed separately by PS-4565

mysql-test/suite/innodb/r/parallel_dblwr_encrypt.result
```diff
@@ -0,0 +1,59 @@
+SET GLOBAL innodb_page_cleaner_disabled_debug=ON;
```
laurynas-biveinis: percona_ prefix for the Percona feature testcases

mysql-test/suite/innodb/t/parallel_dblwr_encrypt.test
```diff
@@ -0,0 +1,92 @@
...
+INSERT INTO t3 (a) VALUES ('Abracadabra is of unknown origin, and its first occurrence is');
+INSERT INTO t3 (a) VALUES ('Abracadabra is of unknown origin, and its first occurrence is');
+INSERT INTO t3 (a) VALUES ('Abracadabra is of unknown origin, and its first occurrence is');
+INSERT INTO t3 (a) VALUES ('Abracadabra is of unknown origin, and its first occurrence is');
+INSERT INTO t3 (a) VALUES ('Abracadabra is of unknown origin, and its first occurrence is');
+INSERT INTO t3 (a) VALUES ('Abracadabra is of unknown origin, and its first occurrence is');
+
+SET GLOBAL innodb_page_cleaner_disabled_debug=OFF;
+SET GLOBAL innodb_buf_flush_list_now=ON;
+SET GLOBAL innodb_fast_shutdown=2;
```
laurynas-biveinis: Why not just kill_mysqld.inc?

mysql-test/suite/innodb/t/parallel_dblwr_encrypt.test
```diff
@@ -0,0 +1,92 @@
+# Test Parallel double write buffer encryption
+
+--source include/have_innodb.inc
+--source include/not_embedded.inc
+--source include/have_debug.inc
```
laurynas-biveinis: This does not have to be a debug testcase: issue workload, wait until buffer pool is clean, kill/shutdown mysqld - the doublewrite will have lots of page images from previous flushes for grepping

mysql-test/suite/innodb/t/parallel_dblwr_encrypt.test
```diff
@@ -0,0 +1,92 @@
...
+
+--echo # Writes to Parallel dblwr are not encrypted, so it should be "Pattern found"
+--let $grep_pattern= first occurrence
+--let $grep_file= $MYSQLD_DATADIR/xb_doublewrite
+--let $grep_output= boolean
+--source include/grep_pattern.inc
+
+--source include/start_mysqld.inc
+DROP TABLE t1, t2, t3;
+CREATE TABLE t1(a TEXT) ENCRYPTION='Y';
```
laurynas-biveinis: What does the 2nd half of the testcase test?

mysql-test/suite/innodb/t/sys_tablespace_encrypt.test
```diff
@@ -0,0 +1,181 @@
...
+let $MYSQLD_BASEDIR= `select @@basedir`;
+--mkdir $MYSQL_TMP_DIR/datadir1
+
+let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/datadir1/data;
+
+--echo # Do tests on un-encrypted system tablespace
+SELECT @@innodb_sys_tablespace_encrypt;
+
+--echo # Check that Variable is not dynamic
+--error ER_INCORRECT_GLOBAL_LOCAL_VAR
```
laurynas-biveinis: Belongs to a sys_vars test

mysql-test/suite/innodb/t/sys_tablespace_encrypt.test
```diff
@@ -0,0 +1,181 @@
...
+--source include/grep_pattern.inc
+
+--echo # Start encrypted system without early-plugin-load and only --plugin-load
+--error 1
+--exec $MYSQLD_CMD --datadir=$MYSQLD_DATADIR1 --innodb-sys-tablespace-encrypt=ON $KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_LOAD $KEYRING_DATA > $ERROR_LOG 2>&1
+--let $grep_pattern= Encryption can't find master key, please check the keyring plugin is loaded
+--let $grep_file= $ERROR_LOG
+--let $grep_output= boolean
+--source include/grep_pattern.inc
+
```
laurynas-biveinis: remove ERROR_LOG after each grep

mysql-test/suite/innodb/t/sys_tablespace_encrypt.test
```diff
@@ -0,0 +1,181 @@
...
+--echo # rebuild to a table unencrypted and in system should be disalowed
+--error ER_ILLEGAL_HA_CREATE_OPTION
+ALTER TABLE t7 ALGORITHM=COPY, ENCRYPTION='N';
+SHOW CREATE TABLE t7;
+
+--error ER_ILLEGAL_HA_CREATE_OPTION
+ALTER TABLE t7 ENCRYPTION='N';
+SHOW CREATE TABLE t7;
+
+--echo #Move encrypted table in system tablespace to file_per_table tablespace
```
laurynas-biveinis: ```# Move```

mysql-test/suite/innodb/t/sys_tablespace_encrypt.test
```diff
@@ -0,0 +1,181 @@
...
+--echo # t7 should be encrypted table. Verify the ibd later
+ALTER TABLE t7 TABLESPACE=`innodb_file_per_table`;
+SHOW CREATE TABLE t7;
+
+--echo # Move encrypted table in sytem tablespace as an unencrypted table
+ALTER TABLE t6 TABLESPACE=`innodb_file_per_table`, ENCRYPTION='N';
+SHOW CREATE TABLE t6;
+
+--echo #Move unencrypted file_per_table to a table in encrypted system tablespace
```
laurynas-biveinis: ```# Move```

mysql-test/suite/innodb/t/sys_tablespace_encrypt_dblwr.test
```diff
@@ -0,0 +1,155 @@
...
+
+--echo # Make the first page dirty for table t1
+set global innodb_saved_page_number_debug = 3;
+set global innodb_fil_make_page_dirty_debug = @space_id;
+
+--source include/kill_mysqld.inc
+
+--echo # Make the 3rd page of the user tablespace
+--echo # corrupted
+perl;
```
laurynas-biveinis: Extract to an include file please (or maybe one exists already?)

mysql-test/suite/sys_vars/r/innodb_sys_tablespace_encrypt-basic.result
```diff
@@ -0,0 +1,88 @@
+Valid values are 'ON' and 'OFF'
```
laurynas-biveinis: _basic not -basic

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -719,10 +719,14 @@ buf_dblwr_init_or_load_pages(
 			return(DB_ERROR);
 		}
 
+		byte zero_page[UNIV_PAGE_SIZE_MAX] = {0};
 		for (page = recovery_buf; page < recovery_buf + size;
 		     page += UNIV_PAGE_SIZE) {
 
-			recv_dblwr.add(page);
+			/* Skipping all zero pages */
+			if (memcmp(page, zero_page, UNIV_PAGE_SIZE) != 0) {
```
laurynas-biveinis: That's a big memcmp. Can you check the checksum field first?

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -1166,6 +1173,42 @@ buf_dblwr_write_block_to_datafile(
...
+				out: encrypted page (if tablespace is
+				encrypted */
+static
+void
+buf_dblwr_encrypt_page(
+	const buf_block_t*	block,
+	page_t*			dblwr_page)
+{
+
+	ulint		space_id = block->page.id.space();
```
laurynas-biveinis: const all locals that can be const'ed

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -1166,6 +1173,42 @@ buf_dblwr_write_block_to_datafile(
...
+	fil_space_t*	space = fil_space_acquire_silent(space_id);
+
+	if (space == NULL) {
+		/* Tablespace dropped */
+		return;
+	}
+
+	byte*		encrypted_buf = static_cast<byte*>(
+		ut_zalloc_nokey(UNIV_PAGE_SIZE));
```
laurynas-biveinis, line 1201: Something about OOM? ut_a(encrypted_buf)? Same effect as just dereferencing attempt, but perhaps with clearer diagnostics

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -1210,6 +1253,8 @@ buf_dblwr_flush_buffered_writes(
 
 	write_buf = dblwr_shard->write_buf;
 
+	bool	encrypt_parallel_dblwr = srv_parallel_dblwr_encrypt;
```
laurynas-biveinis: const

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -1231,7 +1278,18 @@ buf_dblwr_flush_buffered_writes(
...
 		buffer has sane LSN values. */
-		buf_dblwr_check_page_lsn(write_buf + len2);
+		buf_dblwr_check_page_lsn(dblwr_page);
+
+		/* Disable encryption of Page 0 of any tablespace or if it is
+		system tablespace, do not encrypt pages upto TRX_SYS_PAGE_NO
+	        (including). TRX_SYS_PAGE should be not encrypted because dblwr
+		buffer is found from this page */
+		if (encrypt_parallel_dblwr
+		    && block->page.id.page_no() != 0
```
laurynas-biveinis: Extract this logic to a helper function (w/o encrypt_parallel_dblwr) and reuse below too

storage/innobase/include/os0file.h
```diff
@@ -787,12 +791,25 @@ class IORequest {
 		return((m_type & NO_COMPRESSION) == 0);
 	}
 
+	/** @return true if the page write should not be encrypted */
+	bool is_encryption_disabled() const
+		MY_ATTRIBUTE((warn_unused_result))
```
laurynas-biveinis: We have MY_NODISCARD macro for this

storage/innobase/os/os0file.cc
```diff
@@ -9676,6 +9676,100 @@ bool Encryption::check_keyring()
...
+
+	ulint 	bytes = page_size.physical();
+
+	/* After successful encryption, in_page will point
+	to a new memory block which is encrypted and
+	the bytes will have value of length of encrypted data */
+	void*		in_page_before	= in_page;
+	Block* block = os_file_encrypt_page(write_request,
+			in_page_before,
+			&bytes);
```
laurynas-biveinis: Can os_file_encrypt_page return NULL?

storage/innobase/srv/srv0start.cc
```diff
@@ -1491,6 +1491,67 @@ srv_prepare_to_delete_redo_log_files(
...
+	if (create_new_db && srv_sys_tablespace_encrypt) {
+
+		space->flags |= FSP_FLAGS_MASK_ENCRYPTION;
+		srv_sys_space.set_flags(space->flags);
+
+		err = fil_set_encryption(space->id,
+					 Encryption::AES,
+					 NULL,
+					 NULL);
+		ut_ad(err == DB_SUCCESS);
```
laurynas-biveinis, line 1516: Is this required? I.e. wouldn't the release build give an fitting error up in the callstack?

laurynas-biveinis (changes requested):
- I am not sure we need a separate option for the parallel doublewrite buffer encryption. Can we consider it a part of system tablespace for the purposes of encryption?

satya-bodapati:
Regarding this comment "
 I am not sure we need a separate option for the parallel doublewrite buffer encryption. Can we consider it a part of system tablespace for the purposes of encryption?"

we could avoid this variable and encrypt doublewrite buffer only when sys is encrypted. I see two benefits.
1. Since system tablespace can't be encrypted dynamically, and assuming no user tables in system tablespaces (most users would avoid), then encrypting parallel double write buffer independently would secure almost all writes (except for single page flush that goes to system tablespace). Most user
upgrading from earlier versions, will benefit from this dynamic option.

2. The encryption of doublewrite buffer in system tablespace and the parallel doublewrite buffer is different. The pages in system tablespace are encrypted using tablespace key. Whereas the parallel doublewrite buffer pages are using their respective tablespace keys.

laurynas-biveinis:
OK, there are IMHO convincing differences to have a separate option, much as I don't like new option introduction. @robgolebiowski , @gl-sergei ?

mysql-test/suite/innodb/t/parallel_dblwr_encrypt.test
satya-bodapati: because we would like to have all dirty pages flushed and then verify the doublewrite buffer.
A normal shutdown removes parallel doublewrite file. So fast_shutdown=2 is used to retain parallel doublewrite buffer file and verify its contents.

mysql-test/suite/innodb/t/parallel_dblwr_encrypt.test
satya-bodapati: In first half, doublewrite buffer is not encrypted, so the contents are visible in xb_doublewrite file "Pattern found" case.

In second half, doublewrite buffer encryption is turned on, the contents shouldn't be readable in xb_doublewrite file. "Pattern not found" case.

mysql-test/suite/innodb/t/parallel_dblwr_encrypt.test
satya-bodapati: trying with kill_mysqld.inc

satya-bodapati: Done

mysql-test/suite/innodb/t/parallel_dblwr_encrypt.test
satya-bodapati: Done

storage/innobase/buf/buf0dblwr.cc
satya-bodapati: This is more like an optimization. I observed that lot of 'all-zero' are added in doublwrite recovery page list. And this iteration is of fixed size (bufferpools * 127).

Checksum can also be zero. So we cannot use it.
There is existing function buf_page_is_all_zeroes(). Since it is using for loop and byte by byte checking, I decided to use memcp.

laurynas-biveinis: I mean:
check if checksum field is zero
if not zero, page is not all zeroes
if zero, memcmp the whole page

mysql-test/suite/innodb/t/sys_tablespace_encrypt_dblwr.test
satya-bodapati: Doesn't exist. Created one. And also fixed doublewrite.inc to use corrrupt_page.inc

storage/innobase/os/os0file.cc
satya-bodapati: The blocks are allocated from block cache (see os_alloc_block()). There is retry logic to get from cache. After this, it allocates a new block. ut_malloc_nokey() which again has retry logic.
Only in very rare scenario, I see block allocation fails.

Do you want me to add assert ?

storage/innobase/srv/srv0start.cc
satya-bodapati: The assertion is superficial. Only when fil_space_t of tablespace is not found, it would fail. and this cannot happen for hardcoded sys / temp tablespace.

The assert is debug build only. If you wish, I can remove this.

mysql-test/suite/innodb/t/sys_tablespace_encrypt.test
satya-bodapati: Done

storage/innobase/buf/buf0dblwr.cc
satya-bodapati: ah, OK. Will implement this.

storage/innobase/os/os0file.cc
laurynas-biveinis: So in theory it's possible to return NULL, although no-one ever is likely to see one. Since it's possible, I'd prefer a check, but in this case ut_ad would be fine I guess

storage/innobase/srv/srv0start.cc
laurynas-biveinis: Thanks, then it's OK as-is

satya-bodapati: Ok

storage/innobase/os/os0file.cc
satya-bodapati: Done. added debug assert

storage/innobase/include/os0file.h
satya-bodapati: Fixed

storage/innobase/buf/buf0dblwr.cc
satya-bodapati: Fixed

storage/innobase/buf/buf0dblwr.cc
satya-bodapati: Fixed

mysql-test/suite/sys_vars/r/innodb_sys_tablespace_encrypt-basic.result
satya-bodapati: Fixed

mysql-test/suite/innodb/t/sys_tablespace_encrypt.test
satya-bodapati: Fixed

mysql-test/suite/innodb/t/sys_tablespace_encrypt.test
satya-bodapati: Fixed

mysql-test/suite/innodb/t/sys_tablespace_encrypt.test
satya-bodapati: Fixed

mysql-test/suite/innodb/r/parallel_dblwr_encrypt.result
satya-bodapati: Fixed

satya-bodapati:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1874/

innodb_temp_table_encrypt_* failures will be fixed by PS-4565

mysql-test/suite/innodb/t/percona_sys_tablespace_encrypt.test
```diff
@@ -0,0 +1,235 @@
...
+EOF
+
+--echo # start unencrypted system with --innodb-sys-tablespace-encrypt=ON
+--error 1
+--exec $MYSQLD_CMD --innodb-sys-tablespace-encrypt=ON > $ERROR_LOG 2>&1
+--let $grep_pattern= The system tablespace is not encrypted but --innodb_sys_tablespace_encrypt is ON
+--let $grep_file= $ERROR_LOG
+--let $grep_output= boolean
+--source include/grep_pattern.inc
+
```
laurynas-biveinis: --remove_file $ERROR_LOG

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -1166,6 +1177,63 @@ buf_dblwr_write_block_to_datafile(
...
+@return true if encryption should be disabled for the block, else flase */
+static
+bool
+buf_dblwr_disable_encryption(
+	const buf_block_t*	block)
+{
+	if (block->page.id.page_no() == 0
+	    || (block->page.id.space() == TRX_SYS_SPACE
+		&& block->page.id.page_no() <= TRX_SYS_PAGE_NO)) {
+		return(true);
```
laurynas-biveinis: if (x) return true; else return false; -> return x;

storage/innobase/fil/fil0fil.cc
```diff
@@ -5471,6 +5471,22 @@ fil_io_set_encryption(
...
 	fil_space_t*		space)
 {
+
+	/* Explicit request to disable encryption */
+	if (req_type.is_encryption_disabled()) {
+		req_type.clear_encrypted();
+		return;
+	}
+
+	/* Dont encrypt pages of system tablespace upto
```
laurynas-biveinis: Don't

storage/innobase/log/log0recv.cc
```diff
@@ -4544,6 +4544,43 @@ recv_dblwr_t::find_page(ulint space_id, ulint page_no)
...
+	decrypt_request.encryption_algorithm(
+		Encryption::AES);
+
+	Encryption	encryption(
+		decrypt_request.encryption_algorithm());
+
+	for (list::iterator i = sys_pages.begin(); i != sys_pages.end(); ++i) {
+		byte*	page = *i;
+
+		/** System tablespace encryption key will be used to decrypt the
```
laurynas-biveinis: Probably a regular not doxygen comment

laurynas-biveinis (changes requested):
Minor comments

mysql-test/suite/innodb/t/percona_sys_tablespace_encrypt.test
satya-bodapati: Fixed

storage/innobase/buf/buf0dblwr.cc
satya-bodapati: Fixed

storage/innobase/fil/fil0fil.cc
satya-bodapati: Fixed

storage/innobase/log/log0recv.cc
satya-bodapati: Fixed

satya-bodapati:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1877/
perfschema.show_sanity is fixed now.

satya-bodapati:
Updated patch with fix for perfschema.show_sanity

laurynas-biveinis (approved):

robgolebiowski:
Hi @satya-bodapati @laurynas-biveinis 

The system tablespace is encrypted with system's tablespace key. This means we need to encrypt twice - first - with system tablespace key when we add a page into double write buffer and then the second time when we write a page to datafile - with a key of the target tablespace. Have you consider encrypting the page in system tablespace with target tablespace's key - as you are doing for parallel double write buffer ? Have you had a chance to measure how big is impact of this double encryption on performance ?

robgolebiowski:
I am also seeing problem with encrypting pages with system tablespace key in doublewrite buffer. If we decide that we need to have one encryption for both doublewrite buffer and datafile (instead of two currently) we will need to change the key used to decrypt double write buffer from the system's table key to tablespace key - which might break backward compatibility. The reason why single encryption might be quite important for us is that we are trying to get at least the performance that MariaDB has for their encryption and they are encrypting pages on different level (in buffer pools) so they encrypt each page only once. I am very interested in seeing those benchmark results. Anyways - having tables encrypted with tablespace keys will make our live easier if there will be need in the future to have single encryption.

I understand this is only done for single writes

PR #2468, opened 2018-07-21
PS-4561: Read after free at Binlog_crypt_data::load_latest_binlog_key()

robgolebiowski:
Removed access to system_key after it was freed. Also initliazed nonce
to 0s.

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -5214,10 +5214,6 @@ bool MYSQL_BIN_LOG::open_binlog(const char *log_name,
 
   if (encrypt_binlog)
   {
-    uchar nonce[Binlog_crypt_data::BINLOG_NONCE_LENGTH];
-    if (my_rand_buffer(nonce, sizeof(nonce)))
-      goto err;
-
     if (crypto.load_latest_binlog_key())
     {
       sql_print_error("Failed to fetch or create percona_binlog key from/in keyring and thus "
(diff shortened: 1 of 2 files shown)


robgolebiowski:
https://jenkins.percona.com/job/mysql-5.7-valgrind-param/97/
Tests rpl_encryption.encrypted_slave rpl_encryption.encrypted_master are failing due to timeouts.

sql/binlog.cc
```diff
@@ -5228,6 +5224,16 @@ bool MYSQL_BIN_LOG::open_binlog(const char *log_name,
     DBUG_EXECUTE_IF("check_consecutive_binlog_key_versions",
                     { static uint next_key_version = 0;
                       DBUG_ASSERT(crypto.get_key_version() == next_key_version++);});
+
+    uchar nonce[Binlog_crypt_data::BINLOG_NONCE_LENGTH];
+    memset(nonce, 0, Binlog_crypt_data::BINLOG_NONCE_LENGTH);
+    if (my_rand_buffer(nonce, sizeof(nonce)))
+      goto err;
+
+    // Check that after my_rand_buffer nonce does not contain all zeroes
```
laurynas-biveinis: I am not sure this check is useful: it's supposed to be an extremely rare occurrence. If it happens on a production server, debug check will not help, if it happens on a debug build, it's non-actionable. An option would be to generate a random not-zero byte and poke it somewhere, but this changes PRNG properties, and I'm not sure if for the better. What's the textbook standard way of dealing with this?

laurynas-biveinis (changes requested):

sql/binlog.cc
robgolebiowski: I just wanted to be extra sure that after initializing nonce with 0 and then running my_rand_buffer on it - it for sure will be initialized with random bytes ... more like a sanity check. But when I look at it now I think I was bit over paranoid. I am OK with removing this check. Should I ? :)

laurynas-biveinis: Yes, please remove

robgolebiowski:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-valgrind-param/

laurynas-biveinis (approved):

PR #2470, opened 2018-07-25
Fixed PS-4570 (Flush status; adds twice to global values)

percona-ysorokin:
https://jira.percona.com/browse/PS-4570

Problem:
Executing 'FLUSH STATUS' command when 'show_compatibility_56' is set to 'OFF'
could result in accumulating status variables twice.

Consider the following scenario:
1. A new connection 'con1' is opened.
  "SHOW GLOBAL STATUS LIKE 'Slow_queries'" is executed and returns 0.
2. Another connection 'con2' is opened and a very slow query like
  "SELECT SLEEP(10)" is executed. After that both
  "SHOW GLOBAL STATUS LIKE 'Slow_queries'" and
  "SHOW SESSION STATUS LIKE 'Slow_queries'" (for 'con2') return 1.
3. 'con2' is closed.
4. In 'con1' "SHOW GLOBAL STATUS LIKE 'Slow_queries'" is executed and returns
   1 as expected. Moreover "SHOW SESSION STATUS LIKE 'Slow_queries'" also
   returns expected value 0.
5. "FLUSH STATUS" is executed in 'con1' and
   "SHOW GLOBAL STATUS LIKE 'Slow_queries'" now returns 2.

This happens because when a connection is closed 'THD::release_resources()'
method is called which in turn calls
"add_to_status(&global_status_var, &status_var, false);".
The last parameter 'false' indicates that session values must not be reset
after aggregation to global ones. The comment in the code says that the
status values are not reset so that they can still be read by performance
schema.

On the other hand, when "FLUSH STATUS" is executed, 'refresh_status()'
function is called which in turn calls 'reset_pfs_status_stats()'.
This function propagates all PFS stats accumulated on different levels:
PFS_thread -> PFS_account -> PFS_user -> PFS_host -> Global status vars.
However, it does not take into account that some values have already been
added to global status vars when other connections were closed.

Fix:
'PFS_status_stats' class extended with additional set of counters
'm_stats_aggregated_in_global' that will be used similarly to 'm_stats'
to accumulate values that have already been added to global status vars.
'PFS_status_stats::aggregate_from()' method is extended with additional
boolean parameter 'already_aggregated' which determines to which array
('m_stats' or 'm_stats_aggregated_in_global') the passed values must be
accumulated. This method is now called with 'thd->status_var_aggregated'
passed as this boolean parameter.
On the other hand 'PFS_status_stats::aggregate_to()' methods is also
extended with additional boolean parameter 'include_aggregated' indicating
whether stats already aggregated in global status vars should be added to
the output object. This allows to eliminate double aggregation when this
methos is called for the global status vars
"xxx.aggregate_to(&global_status_var, false);". For other local calculations
it is called with 'true'
"xxx.aggregate_to(&yyy, true);".

Added new 'perfschema.bug91541' MTR test case which simulates the scenario
above.

### DIFF
--- a/mysql-test/suite/perfschema/t/bug91541.test
+++ b/mysql-test/suite/perfschema/t/bug91541.test
@@ -1,0 +1,67 @@
+--echo #
+--echo # Test for bug bug91541 ("Flush status" statement adds twice to global values)
+--echo #
+
+--source include/have_perfschema.inc
+
+SET @saved_show_compatibility_56 = @@global.show_compatibility_56;
+SET GLOBAL show_compatibility_56 = OFF;
+
+--echo #
+--echo # default connection
+--echo #
+--let $initial_global_value = query_get_value(SHOW GLOBAL STATUS LIKE 'Slow_queries', Value, 1)
+
+--source include/count_sessions.inc
+
+--echo #
+--echo # first connection
+--echo #
+--connect (con1,localhost,root,,)
+--let $current_global_value = query_get_value(SHOW GLOBAL STATUS LIKE 'Slow_queries', Value, 1)
+--let $assert_cond = $current_global_value - $initial_global_value = 0
+--let $assert_text = Global value requested from the first session must not change
+--source include/assert.inc
+
+SET long_query_time = 2;
+SELECT SLEEP(3);
+SET long_query_time = @@global.long_query_time;
+
+--let $current_global_value = query_get_value(SHOW GLOBAL STATUS LIKE 'Slow_queries', Value, 1)
+--let $first_session_value = query_get_value(SHOW SESSION STATUS LIKE 'Slow_queries', Value, 1)
+--let $assert_cond = $current_global_value - $initial_global_value = 1
+--let $assert_text = Global value requested from the first session after SLEEP() must inccrease by 1
+--source include/assert.inc
+--let $assert_cond = $first_session_value = 1
+--let $assert_text = First session value must increase by 1
+--source include/assert.inc
+
+--connection default
+--disconnect con1
+--source include/wait_until_count_sessions.inc
+
+--echo #
+--echo # default connection
+--echo #
+--connection default
+--let $current_global_value = query_get_value(SHOW GLOBAL STATUS LIKE 'Slow_queries', Value, 1)
+--let $default_session_value = query_get_value(SHOW SESSION STATUS LIKE 'Slow_queries', Value, 1)
+--let $assert_cond = $current_global_value - $initial_global_value = 1
+--let $assert_text = Global value requested from the default session after DISCONNECT must remain the same
+--source include/assert.inc
+--let $assert_cond = $default_session_value = 0
+--let $assert_text = Default session value must remain zero
+--source include/assert.inc
+
+FLUSH STATUS;
+
+--let $current_global_value = query_get_value(SHOW GLOBAL STATUS LIKE 'Slow_queries', Value, 1)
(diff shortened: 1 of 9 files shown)


percona-ysorokin:
Jenkins build link
https://ps.cd.percona.com/job/percona-server-5.7-param/32/

laurynas-biveinis:
I must be missing something very obvious: why THD::release_resources cannot pass true for add_to_status last arg?

percona-ysorokin:
This was my very first attempt to fix this to be honest :)

However, I've got `perfschema.show_aggregate` MTR test case failing immediately.
It looks like this test relies on the fact that `performance_schema.status_by_thread` table keeps session values for individual threads even after corresponding connections are closed.
In my personal opinion, it's very questionable whether this is correct behavior. However, I doubt Oracle will agree to change it because it's been there for years.

Also, from the source code
```c++
  mysql_mutex_lock(&LOCK_status);
  add_to_status(&global_status_var, &status_var, false);
  /*
    Status queries after this point should not aggregate THD::status_var
    since the values has been added to global_status_var.
    The status values are not reset so that they can still be read
    by performance schema.
  */
  status_var_aggregated= true;
  mysql_mutex_unlock(&LOCK_status);
```

**The status values are not reset so that they can still be read by performance schema.**

mysql-test/suite/perfschema/t/bug91541.test
```diff
@@ -0,0 +1,75 @@
...
+
+SET @saved_show_compatibility_56 = @@global.show_compatibility_56;
+SET GLOBAL show_compatibility_56 = OFF;
+
+--echo #
+--echo # default connection
+--echo #
+--let $initial_global_value = query_get_value(SHOW GLOBAL STATUS LIKE 'Slow_queries', Value, 1)
+--let $default_session_value = query_get_value(SHOW SESSION STATUS LIKE 'Slow_queries', Value, 1)
+--let $assert_cond = $default_session_value = 0
```
laurynas-biveinis: Perhaps it's safe to assume that session Slow_queries are always 0 at the start of the testcase for a bit of simplification and better readability?

mysql-test/suite/perfschema/t/bug91541.test
```diff
@@ -0,0 +1,75 @@
...
+--echo #
+--echo # first connection
+--echo #
+--connect (con1,localhost,root,,)
+--let $current_global_value = query_get_value(SHOW GLOBAL STATUS LIKE 'Slow_queries', Value, 1)
+--let $first_session_value = query_get_value(SHOW SESSION STATUS LIKE 'Slow_queries', Value, 1)
+--let $assert_cond = $current_global_value - $initial_global_value = 0
+--let $assert_text = Global value requested from the first session must not change
+--source include/assert.inc
+--let $assert_cond = $first_session_value = 0
```
laurynas-biveinis: A brand new connection will start with zero session counter for sure?

laurynas-biveinis (changes requested):
Yeah, a mess. I don't have a better fix idea, let's go with this one

mysql-test/suite/perfschema/t/bug91541.test
percona-ysorokin: Agree. Assertion removed.

mysql-test/suite/perfschema/t/bug91541.test
percona-ysorokin: Agree. Assertion removed.

percona-ysorokin:
Updated Jenkins build link (--unit-tests-report perfschema.bug91541 perfschema.show_aggregate --repeat=4)
https://ps.cd.percona.com/job/percona-server-5.7-param/34/

PR #2475, opened 2018-08-02
Ps 5.7 4196

george-lorch:


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1884/

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4541,10 +4541,10 @@ static ulonglong rdb_get_int_col_max_value(const Field *field) {
 void ha_rocksdb::load_auto_incr_value() {
   ulonglong auto_incr = 0;
   bool validate_last = false, use_datadic = true;
-#ifndef NDEBUG
+#if !defined(DBUG_OFF)
```
laurynas-biveinis, line 4544: I recall pointing out NDEBUG to upstream and them replying that it's fine.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -12429,7 +12442,7 @@ int mysql_value_to_bool(struct st_mysql_value *value, my_bool *return_value) {
     value->val_int(value, &intbuf);
     if (intbuf > 1)
       return 1;
-    *return_value = intbuf > 0 ? TRUE : FALSE;
+    *return_value = intbuf > 0 ? true : false;
```
laurynas-biveinis: ugh, (condition ? true : false) -> (condition)

storage/rocksdb/rdb_datadic.cc
```diff
@@ -964,6 +964,15 @@ bool Rdb_key_def::covers_lookup(TABLE *const table,
   return bitmap_is_subset(lookup_bitmap, &covered_bitmap);
 }
 
+/* Indicates that all key parts can be unpacked to cover a secondary lookup */
```
laurynas-biveinis, line 967: It looks like it is necessary to call can_cover_lookup() before each covers_lookup() call. Fold the former into the latter, by inlining or by having a private method?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -1252,6 +1262,468 @@ uint Rdb_key_def::pack_hidden_pk(const longlong &hidden_pk_id,
   Function of type rdb_index_field_pack_t
 */
 
+#if !defined(DBL_EXP_DIG)
```
laurynas-biveinis: Add a comment about the origin of this code?

laurynas-biveinis (changes requested):

storage/rocksdb/rdb_datadic.cc
george-lorch: Not necessarily. If you look at the logic w.r.t. m_keyread_only which _forces_ a key read only with no test to see if the read can be satisfied via key.  That is what this fixes.  covers_lookup() is for a different purpose and only returns true if a lookup _can_ be satisfied by a keyread only.  At this point it will always return false because that is how upstream tries to keep this functionality disabled. The logic where this is used changes from:

m_keyread_only || x->covers_lookup(...)

to

(m_keyread_only && x->can_cover_lookup()) || x->covers_lookup(...)

storage/rocksdb/ha_rocksdb.cc
george-lorch: I do seem to recall that conversation as well.  This does now cause issues in 8.0 and I do believe it is wrong to use NDEBUG when attempting any kind of conditional compilation that interacts with my_dbug.h.  All functionality of my_dbg.h is based on !DBUG_OFF, therefore, any use of my_dbug.h and external conditional compilation in support of its use should use the same base indicator (DBUG_OFF) and not rely on an assumption that NDEBUG == DBUG_OFF because they are not the same.

george-lorch:
Touched up a few minor things on review commentary, re-tested and jenkins https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1885/

laurynas-biveinis (approved):

PR #2476, opened 2018-08-02
Merge upstream 5.6.41 into Percona Server

percona-ysorokin:


percona-ysorokin:
Jenkins build link (OpenSSL)
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/8/

Jenkins build link (YaSSL)
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/9/

plugin/innodb_memcached/innodb_memcache/src/innodb_engine.c
```diff
@@ -1135,22 +1135,22 @@ innodb_conn_init(
 			}
 		}
 	} else {
+        bool auto_commit = (engine->read_batch_size == 1 &&
+                !(engine->cfg_status & IB_CFG_DISABLE_ROWLOCK)) ? true : false;
```
laurynas-biveinis, line 1139: When will this "bool foo = condition ? true : false" stop
(does not need fixing here)

laurynas-biveinis (approved):

plugin/innodb_memcached/innodb_memcache/src/innodb_engine.c
percona-ysorokin: Yeah :)
This is not the best example though.
My favorite is
```c++
if(flag == true)
{
  ...
}
```

PR #2477, opened 2018-08-06
PS-4620: Enable encryption of temporary tablespace from foreground th…

satya-bodapati:
…read

Issues:
-------
1. SET GLOBAL innodb_temp_tablespace_encrypt=ON doesn't immediately enable
   encryption of temporary tablespace. Users have to wait certain period

2. SELECT @@innodb_temp_tablespace_encrypt value depends on background thread
   enabling encryption of temporary tablespace.

3. For keyring plugin errors, users are asked to check error log. A warning or
   error is not thrown.

4. innodb_temp_tablespace_encrypt is set to ON when temp tablespace couldn't be
   encrypted at startup. Didn't check for keyring. Wrongly shows temp tablespace
   as encrypted.

5. Race condition when encryption is enabled on temporary tablespace.

Fix:
----
1. Enable encryption of temp tablespace from foreground user threads.
2. Throw warnings when encryption cannot be enabled for temporary tablespace
3. At startup, if innodb_temp_tablespace_encrypt=ON and keyring plugin
   is not loaded,  aborts startup with error message.
4. Fix race condition by using fil_system mutex when reading or setting
   encryption attributes in tablespace object (fil_space_t)
5. Remove unnecessary debug variable "innodb_master_encrypt_debug"
6. Added more test scenarios
7. Reverted PS-3958 code changes. It cannot happen with PS-4620


mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -0,0 +1,85 @@
+--source include/have_innodb.inc
+--source include/count_sessions.inc
```
laurynas-biveinis: Don't count sessions here, you are restarting the server later - move it after the last in-testcase restart

laurynas-biveinis (changes requested):
Please clarify point 3 in Fix section in the commit message

mysql-test/suite/innodb/t/temp_table_encrypt.test
satya-bodapati: Right.. added some testcase scenarios at top and this is now placed at wrong position

satya-bodapati: clarified point 3

satya-bodapati:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1888/ (brewing..)

satya-bodapati:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1889/ (with fixes to two tests)

laurynas-biveinis (commented):
Do perfschema.show_sanity tests reference the removed variable?

satya-bodapati:
I found one reference to the removed variable in that test. The test still passes even without removing it.

Anyways I have removed it now.

 git diff
diff --git a/mysql-test/suite/perfschema/t/show_sanity.test b/mysql-test/suite/perfschema/t/show_sanity.test
index ffdacc37e89..f9b70aa0047 100644
--- a/mysql-test/suite/perfschema/t/show_sanity.test
+++ b/mysql-test/suite/perfschema/t/show_sanity.test
@@ -481,7 +481,6 @@ insert into test.sanity values
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "INNODB_LOG_GROUP_HOME_DIR"),
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "INNODB_LOG_WRITE_AHEAD_SIZE"),
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "INNODB_LRU_SCAN_DEPTH"),
-  ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "INNODB_MASTER_ENCRYPT_DEBUG"),
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "INNODB_TEMP_TABLESPACE_ENCRYPT"),
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "INNODB_MASTER_THREAD_DISABLED_DEBUG"),
   ("JUNK: GLOBAL-ONLY", "I_S.SESSION_VARIABLES", "INNODB_MAX_BITMAP_FILE_SIZE"),

laurynas-biveinis (approved):
OK with perfschema show_sanity (and show_sanity56?) changes.

satya-bodapati:
Thank you @laurynas-biveinis ! show_sanity56 doesn't exist.

PR #2478, opened 2018-08-08
PS-4718 : 8.0.12 changes Field_blob::make_sort_key results.

george-lorch:
- MyRocks relies on server code Field_blob::make_sort_key to generate key data
  that it will persist. The problem is that the results of these methods
  are/were never intended to be persisted and the results of these methods may
  change over time. When they do they break MyRocks ability decode its own key
  data that it has previously stored. This has now happened in 8.0. as of 8.0.12
  the returned format for al BLOB types is different than is for 5.7. This
  breaks upgrades and also breaks MyRocks ability to encode, store, retrieve,
  and decode keys of any BLOB types.

- In order to prevent this from happening again in 5.7 and 8.0, this commit
  copies the current logic from Field_blob::make_sort_key into the MyRocks data
  dictionary code, thus breaking the MyRocks dependency on the
  Field_blob::make_sort_key functionality. Only fixed length CHAR types remain
  using the make_sort_key method.

### DIFF
--- a/storage/rocksdb/rdb_datadic.cc
+++ b/storage/rocksdb/rdb_datadic.cc
@@ -1731,6 +1731,68 @@ void Rdb_key_def::pack_newdate(
   *dst += length;
 }
 
+void Rdb_key_def::pack_blob(
+    Rdb_field_packing *const fpi, Field *const field,
+    uchar *const buf MY_ATTRIBUTE((__unused__)), uchar **dst,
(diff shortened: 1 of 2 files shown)


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1890/

laurynas-biveinis (approved):

PR #2481, opened 2018-08-10
PS-4727 - intrinsic temp table behaviour shouldn't depend on innodb_e…

satya-bodapati:
…ncrypt_tables

Problem:
--------
Optimizer temporary tables (aka Intrinsic temp table) encryption attribute is based on innodb_encrypt_tables. So when innodb_encrypt_tables is ON and innodb_temp_tablespace_encrypt is OFF,  queries that use disk temp storage engine will fail. For example  statements that use GROUP BY, ORDER BY INSERT SELECT,  etc will fail.

Fix:
----
Intrinsic temp tables should derive encryption attribute from
temporary tablespace property and not rely on innodb_encrypt_tables

### DIFF
--- a/mysql-test/suite/innodb/t/temp_table_encrypt.test
+++ b/mysql-test/suite/innodb/t/temp_table_encrypt.test
@@ -4,6 +4,18 @@
 call mtr.add_suppression("\\[Error\\] InnoDB: keyring error: please check that a keyring plugin is loaded");
 call mtr.add_suppression("\\[Error\\] InnoDB: Can't set temporary tablespace to be encrypted because keyring plugin is not available");
 
+--echo #
+--echo # PS-4727: instrinsic temp table behaviour shouldn't depend on innodb_encrypt_tables
+--echo #
+CREATE TABLE t1 (a INT, b BLOB);
+INSERT INTO t1 VALUES (1, 'hello'), (2, 'hi'), (3, 'satya');
+SET GLOBAL innodb_encrypt_tables=ON;
+SET big_tables=ON;
+INSERT INTO t1 SELECT * FROM t1;
+DROP TABLE t1;
+SET big_tables=default;
+SET GLOBAL innodb_encrypt_tables=default;
+
 --echo # without keyring plugin, try to enable encryption of temporary
 --echo # tablespace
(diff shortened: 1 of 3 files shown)


satya-bodapati:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1896/ (brewing)

satya-bodapati:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1896/ is finished. Only unrelated tests(tokudb & rocksdb) tests failing

mysql-test/suite/innodb/t/temp_table_encrypt.test
```diff
@@ -4,6 +4,18 @@
...
+--echo #
+--echo # PS-4727: instrinsic temp table behaviour shouldn't depend on innodb_encrypt_tables
+--echo #
+CREATE TABLE t1 (a INT, b BLOB);
+INSERT INTO t1 VALUES (1, 'hello'), (2, 'hi'), (3, 'satya');
+SET GLOBAL innodb_encrypt_tables=ON;
+SET big_tables=ON;
+INSERT INTO t1 SELECT * FROM t1;
+DROP TABLE t1;
+SET big_tables=default;
```
laurynas-biveinis, line 16: Instead of manipulating big_tables, a shorter version would be SELECT SQL_BIG_RESULT ...

laurynas-biveinis (approved):
Minor comment, up to you to address or not. If you do, it will be sufficient to test only that testcase

mysql-test/suite/innodb/t/temp_table_encrypt.test
satya-bodapati: From documentation, "SQL_BIG_RESULT or SQL_SMALL_RESULT can be used with GROUP BY or DISTINCT to tell the optimizer that the result set has many rows or is small, respectively.'
Since I am using INSERT SELECT (although it may work), I would prefer SET big_tables here.

BTW, this is added as safety measure. With BLOBs, innodb storage engine is used in 5.7. In future (with temptable engine in 8.0 etc), this may not be true. So I added this protection

PR #2482, opened 2018-08-10
PS-4729 : 8.0 eliminates TRUE and FALSE for bool values, only true an…

george-lorch:
…d false are valid.

- s/TRUE/true   s/FALSE/false

### DIFF
--- a/storage/tokudb/hatoku_hton.cc
+++ b/storage/tokudb/hatoku_hton.cc
@@ -575,10 +575,10 @@ static int tokudb_init_func(void *p) {
 
     db_env->set_update(db_env, tokudb_update_fun);
 
-    db_env_set_direct_io(tokudb::sysvars::directio == TRUE);
+    db_env_set_direct_io(tokudb::sysvars::directio);
 
     db_env_set_compress_buffers_before_eviction(
-        tokudb::sysvars::compress_buffers_before_eviction == TRUE);
+        tokudb::sysvars::compress_buffers_before_eviction);
 
     db_env->change_fsync_log_period(db_env, tokudb::sysvars::fsync_log_period);
 


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2201/

storage/tokudb/hatoku_hton.cc
```diff
@@ -575,10 +575,10 @@ static int tokudb_init_func(void *p) {
 
     db_env->set_update(db_env, tokudb_update_fun);
 
-    db_env_set_direct_io(tokudb::sysvars::directio == TRUE);
+    db_env_set_direct_io(tokudb::sysvars::directio == true);
```
laurynas-biveinis: cond == true -> cond, but see the comment below

storage/tokudb/hatoku_hton.cc
```diff
@@ -575,10 +575,10 @@ static int tokudb_init_func(void *p) {
...
 
     db_env_set_compress_buffers_before_eviction(
-        tokudb::sysvars::compress_buffers_before_eviction == TRUE);
+        tokudb::sysvars::compress_buffers_before_eviction == true);
```
laurynas-biveinis: cond == true -> cond, but see the comment below

storage/tokudb/tokudb_sysvars.cc
```diff
@@ -48,41 +48,41 @@ const size_t error_buffer_max_size = 1024;
 ulonglong   cache_size = 0;
 uint        cachetable_pool_threads = 0;
 int         cardinality_scale_percent = 0;
-my_bool     checkpoint_on_flush_logs = FALSE;
+my_bool     checkpoint_on_flush_logs = false;
```
laurynas-biveinis: false/true are not values of my_bool type, which is char in 5.6. IMHO FALSE/TRUE should be kept for my_bool

laurynas-biveinis (changes requested):

george-lorch:
Made changes to revert the changes to my_bool types, but fixed/kept minor conditional changes. New jenkins https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2205/.

Since changes here are trivial and jenkins indicates no issue, merging ahead of further approval.

PR #2483, opened 2018-08-10
Ps 5.7 4729

george-lorch:


george-lorch:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1897/

laurynas-biveinis (changes requested):
Same comments as for the 5.6 version. IMHO the conversion should be done in 8.0 only, if there are no variables that can be converted from my_bool to bool in lower versions.

george-lorch:
Made changes to revert the changes to my_bool types, but fixed/kept minor conditional changes. New jenkins https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1902/.

Since changes here are trivial and jenkins indicates no issue, merging ahead of further approval.

PR #2484, opened 2018-08-11
PS-4733 : Remove unused declaration for tokudb_open_tables

george-lorch:
- Removed unused declaration that eventually causes compilation problems for 8.0


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2203/

PR #2490, opened 2018-08-14
PS-4386 : Fix various places where macros are redefined or used when …

george-lorch:
…undefined

- Fix various places where macros are redefined or used when undefined-

### DIFF
--- a/storage/tokudb/tokudb_sysvars.h
+++ b/storage/tokudb/tokudb_sysvars.h
@@ -93,10 +93,10 @@ extern my_bool      gdb_on_fatal;
 
 extern my_bool         check_jemalloc;
 
-#if TOKUDB_DEBUG
+#if defined(TOKUDB_DEBUG)
 // used to control background job manager
 extern my_bool      debug_pause_background_job_manager;
-#endif // TOKUDB_DEBUG
+#endif // defined(TOKUDB_DEBUG)
 
 // session/thread
 my_bool     alter_print_error(THD* thd);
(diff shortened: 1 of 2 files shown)


george-lorch:
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2206/

george-lorch:
new jenkins https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/10/

storage/tokudb/tokudb_sysvars.h
```diff
@@ -93,10 +93,10 @@ extern my_bool      gdb_on_fatal;
 
 extern my_bool         check_jemalloc;
 
-#if TOKUDB_DEBUG
+#if defined(TOKUDB_DEBUG) && TOKUDB_DEBUG
```
laurynas-biveinis: Now there are two forms of this check: above defined(FOO) and here defined(FOO) && FOO. Should be a single form (IMHO the former one) only

laurynas-biveinis (changes requested):

storage/tokudb/tokudb_sysvars.h
george-lorch: Ahh, yeah, good point.

george-lorch:
Fixed and new jenkins https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/12/

laurynas-biveinis (approved):

PR #2493, opened 2018-08-15
PS-3889: Native Partitioning for TokuDB and RocksDB for 5.7

vlad-lesin:
TokuDB and MyRocks native partitioning is implemented as a copy of partition
storage engine in 5.7 with corresponding changes.

The base class - Partition_base does the most of the work.

The general differences from ha_partition class are the following:

1) all code which works with .par files was removed;
2) the code to read partition information from .frm files was added;
3) the responsibility for creating new handlers for separate partitions is moved
to Partition_base class descendants;
4) the initialization sequence for native and non-native partitioning handlers
is different, that is why create() and open() functions are changed
5) delete and rename table functions are changed both for Partition_base and
it's descendants(ha_tokupart and ha_rockspart) because it's necessary to
determine if the table is partitioned before calling the corresponding
functionality.

The changes in the server code are the following:

1) remove .frm file after a table has been removed to let handler the ability
to read partition info from .frm file during "delete table" execution;
2) remove "static" keyword from some functions signature to use them outside of
their module.

To see the difference between the original ha_partition and Partition_base classes use the following command: diff -u -w ./storage/partition/ha_partition.cc ./sql/partitioning/partition_base.cc

Testing: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1908/


mysql-test/suite/parts/inc/upgrade_parts_from_prev_ver_alter.inc
```diff
@@ -0,0 +1,37 @@
...
+--echo # Unzip the zip file.
+--file_exists $DATA_ARCH_PATH
+--exec unzip -qo $DATA_ARCH_PATH -d $MYSQL_TMP_DIR
+
+--echo # Create a bootstrap file in temp location
+--replace_result $MYSQL_TMP_DIR MYSQL_TMP_DIR
+--exec echo $MYSQL_TMP_DIR/bootstrap.log
+
+--echo # Restart the server
+--exec echo "restart: --loose-skip-log-bin --skip-log-slave-updates --datadir=$MYSQLD_DATADIR1 $ADDITIONAL_OPTS" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
```
laurynas-biveinis: Use restart_mysqld.inc

mysql-test/suite/parts/inc/upgrade_parts_from_prev_ver_mysql_upgrade.inc
```diff
@@ -0,0 +1,30 @@
+--echo #########
+--echo # Test for upgrade partitioning with mysql_upgrade
+--echo ###
+
+--echo # Unzip the zip file.
+--file_exists $DATA_ARCH_PATH
+--exec unzip -qo $DATA_ARCH_PATH -d $MYSQL_TMP_DIR
+
+--echo # Restart the server
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/t/native_part_dev.test
```diff
@@ -0,0 +1,244 @@
+# This test is for debugging
```
george-lorch: I'm not understanding the purpose of this test.  Is it purely for development purposes and should be removed or does it contain useful test cases that cover cases that are not covered in other tests?

george-lorch (commented):
There is a lot to review and understand here.  Fortunately, the great bulk of the working code is literally directly copied from the existing generic partition handler.  The most interesting thing is how most of the existing partitioning tests pass without any modifications (except for the deprecation warning).  Ttat is a very good sign.  I like the reuse/factoring out of the common code to be shared by the two engines and how minimal the actual engine changes are.  I am a _little_ concerned about the changes needed in the server code, but, they are not terribly invasive and seem needed in order to use the .frm files and not the .par files.  Other than the one 'dev' test, nothing really jumps out at me as a potential problem.  I will be looking at this again a few times to see if anything else comes out, but, for now I wouldn't hesitate to approve this pending a review from @laurynas-biveinis

george-lorch:
One thing I realized is missing that we talked about a while ago is the ability to enable/disable the use of native parts in 5.7.  I don't know how difficult this might be.  Ideally we want new options tokudb_enable_native_partition and rocksdb_enable_native_partition with defaults to off.  This is specifically because we are releasing this mid-series 5.7 and this changes the default behavior of partitioning.  We need try to maintain the existing behavior as well as offer a sort of 'emergency exit' just in case a user runs into some critical issue with the native partitioning.  These options do not need to be dynamic, they can be global read-only.  @vlad-lesin do you think this would be difficult or take too much time?

mysql-test/suite/rocksdb/t/native_part_dev.test
laurynas-biveinis: More descriptive comment please,for exampple  "RocksDB native partitioning development tests"

mysql-test/suite/rocksdb/t/native_part_dev.test
```diff
@@ -0,0 +1,244 @@
...
+
+--source include/have_rocksdb.inc
+
+CREATE TABLE t1 (i INT, j INT, k INT, PRIMARY KEY (i)) ENGINE = ROCKSDB PARTITION BY KEY(i) PARTITIONS 4;
+ALTER TABLE t1 CHECK PARTITION p1;
+DROP TABLE t1;
+--exit
```
laurynas-biveinis: Either enable all of the test, either truncate here

mysql-test/suite/rocksdb/t/upgrade_parts_from_prev_ver.test
```diff
@@ -0,0 +1,34 @@
+--source include/have_rocksdb.inc
+--source ../include/have_write_committed.inc
+
+# For data directory created in previous major version branch.
+--source include/not_valgrind.inc
```
laurynas-biveinis: Why?

mysql-test/suite/rocksdb/t/upgrade_parts_from_prev_ver.test
```diff
@@ -0,0 +1,34 @@
...
+--let $MYSQLD_DATADIR1=$MYSQL_TMP_DIR/$DATA_DIR_NAME
+--let $DATA_ARCH_PATH=$MYSQL_TMP_DIR/$DATA_DIR_NAME.zip
+
+--echo # Copy the remote tablespace & DB zip files from suite location to working location.
+--copy_file $MYSQLTEST_VARDIR/std_data/$DATA_DIR_NAME.zip $DATA_ARCH_PATH
+
+--echo # Check that the file exists in the working folder.
+--file_exists $DATA_ARCH_PATH
+
+--echo # Stop DB server which was created by MTR default
```
laurynas-biveinis, line 26: Use shutdown_mysqld.inc

mysql-test/suite/tokudb.parts/t/native_part_dev.test
```diff
@@ -0,0 +1,118 @@
+--echo # This test is for debugging
```
laurynas-biveinis: Likewise

mysql-test/suite/tokudb.parts/t/native_part_dev.test
```diff
@@ -0,0 +1,118 @@
...
+--exit
```
laurynas-biveinis: Likewise

mysql-test/suite/tokudb.parts/t/upgrade_from_prev_ver.test
```diff
@@ -0,0 +1,34 @@
+--source include/have_tokudb.inc
+
+# For data directory created in previous major version branch.
+--source include/not_valgrind.inc
```
laurynas-biveinis: Why?

mysql-test/suite/tokudb.parts/t/upgrade_from_prev_ver.test
```diff
@@ -0,0 +1,34 @@
...
+--let $MYSQLD_DATADIR1=$MYSQL_TMP_DIR/$DATA_DIR_NAME
+--let $DATA_ARCH_PATH=$MYSQL_TMP_DIR/$DATA_DIR_NAME.zip
+
+--echo # Copy the remote tablespace & DB zip files from suite location to working location.
+--copy_file $MYSQLTEST_VARDIR/std_data/$DATA_DIR_NAME.zip $DATA_ARCH_PATH
+
+--echo # Check that the file exists in the working folder.
+--file_exists $DATA_ARCH_PATH
+
+--echo # Stop DB server which was created by MTR default
```
laurynas-biveinis, line 25: shutdown_mysqld.inc

sql/handler.cc
```diff
@@ -8782,16 +8782,15 @@ bool handler::is_using_prohibited_gap_locks(TABLE* table,
...
-          (lock_type != TL_IGNORE && thd->lex->sql_command != SQLCOM_SELECT)))
+  if (!using_full_primary_key && has_transactions() && !has_gap_locks() &&
+      thd_tx_isolation(thd) >= ISO_REPEATABLE_READ && !thd->rli_slave &&
+      (thd->lex->table_count >= 2 || thd->in_multi_stmt_transaction_mode()) &&
+      (lock_type >= TL_WRITE_ALLOW_WRITE ||
+       lock_type == TL_READ_WITH_SHARED_LOCKS ||
+       lock_type == TL_READ_NO_INSERT ||
+       (lock_type != TL_IGNORE && thd->lex->sql_command != SQLCOM_SELECT)) &&
+      thd->lex->sql_command != SQLCOM_ALTER_TABLE &&
+      thd->lex->sql_command != SQLCOM_CHECK)
```
laurynas-biveinis: Why? Why would gap lock checking would happen for ALTER TABLE / CHECK at all?

sql/CMakeLists.txt
```diff
@@ -475,6 +475,16 @@ ELSE()
   SET(MYSQLD_SOURCE main.cc ${DTRACE_PROBES_ALL})
 ENDIF()
 
+CHECK_CXX_COMPILER_FLAG(-std=gnu++11 CXX11_OPTION_SUPPORTED)
```
laurynas-biveinis: Try to reuse cmake/check_stcxx11.cmake - i.e. IF(HAVE_STDCXX11) ...

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+    We keep track if all underlying handlers are MyISAM since MyISAM has a
+    great number of extra flags not needed by other handlers.
+  */
+  bool m_myisam;  // Are all underlying handlers
+                  // MyISAM
+  /*
+    We keep track of InnoDB handlers below since it requires proper setting
+    of query_id in fields at index_init and index_read calls.
+  */
+  bool m_innodb;  // Are all underlying handlers
```
laurynas-biveinis: Remove?

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+  ulong m_low_byte_first;
+  enum enum_handler_status {
+    handler_not_initialized= 0,
+    handler_initialized,
+    handler_opened,
+    handler_closed
+  };
+  enum_handler_status m_handler_status;
+
+  uint m_num_locks;  // For engines like ha_blackhole, which needs no locks
```
laurynas-biveinis: Replace in Partition_base::lock_count with (*file)->lock_count() and remove

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+    "own" the m_part_info structure.
+  */
+  Partition_base *m_is_clone_of;
+  MEM_ROOT *      m_clone_mem_root;
+
+  /*
+    We keep track if all underlying handlers are MyISAM since MyISAM has a
+    great number of extra flags not needed by other handlers.
+  */
+  bool m_myisam;  // Are all underlying handlers
```
laurynas-biveinis: Remove?

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+                  // InnoDB
+  /*
+    When calling extra(HA_EXTRA_CACHE) we do not pass this to the underlying
+    handlers immediately. Instead we cache it and call the underlying
+    immediately before starting the scan on the partition. This is to
+    prevent allocating a READ CACHE for each partition in parallel when
+    performing a full table scan on MyISAM partitioned table.
+    This state is cleared by extra(HA_EXTRA_NO_CACHE).
+  */
+  bool m_extra_cache;
```
laurynas-biveinis: Remove?

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+  uint m_extra_cache_size;
```
laurynas-biveinis: Remove?

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+  /* The same goes for HA_EXTRA_PREPARE_FOR_UPDATE */
+  bool m_extra_prepare_for_update;
```
laurynas-biveinis: Remove?

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+  /* Which partition has active cache */
+  uint m_extra_cache_part_id;
```
laurynas-biveinis: Remove?

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+                            partition_element *p_elem);
+  int  write_row_in_new_part(uint part_id);
+  void close_new_partitions();
+  /*
+    delete_table and rename_table uses very similar logic which
+    is packed into this routine.
+  */
+  int del_ren_table(const char *from, const char *to);
+  /*
+    One method to create the table_name.par file containing the names of the
```
laurynas-biveinis: Outdated comment?

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+  virtual int close(void);
+
+  /*
+    -------------------------------------------------------------------------
+    MODULE start/end statement
+    -------------------------------------------------------------------------
+    This module contains methods that are used to understand start/end of
+    statements, transaction boundaries, and aid for proper concurrency
+    control.
+    The partition handler need not implement abort and commit since this
```
laurynas-biveinis, line 232: Outdated comment paragraph

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+    (i.e. ALTER TABLE t1 TRUNCATE PARTITION p).
+
+    @remark This method is a partitioning-specific hook
+            and thus not a member of the general SE API.
+  */
+  int truncate_partition_low();
+
+  virtual bool is_ignorable_error(int error)
+  {
+    if (handler::is_ignorable_error(error) ||
```
laurynas-biveinis: if (cond) return true; return false; -> return cond;

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+                                               find_flag);
+  }
+  virtual int index_init(uint idx, bool sorted)
+  {
+    return Partition_helper::ph_index_init(idx, sorted);
+  }
+  virtual int index_end() { return Partition_helper::ph_index_end(); }
+
+  /**
+    @breif
```
laurynas-biveinis: brief

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+  {
+    *engine_callback= NULL;
+    *engine_data= 0;
+    return FALSE;
+  }
+
+ private:
+  static const uint NO_CURRENT_PART_ID;
+  int               loop_extra(enum ha_extra_function operation);
+  void              late_extra_cache(uint partition_id);
```
laurynas-biveinis: TODO remove? Seems MyISAM-related

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+}
+
+bool get_part_str_for_path(const char *path, std::string &result)
+{
+  // Check for .frm file existence
+  MY_STAT frm_stat_info;
+  if (!my_stat(path, &frm_stat_info, MYF(0)))
+  {
+    // if .frm file does not exist don't treat it as error
+    if (my_errno() == ENOENT)
```
laurynas-biveinis: if (cond) return true; return false; -> return cond;

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+  if (mysql_file_read(file, head, sizeof(head), MYF(MY_NABP)))
+    return false;
+
+  if (!strncmp((char *)head, "TYPE=VIEW\n", strlen("TYPE=VIEW\n")))
+    return true;
+
+  if (head[0] != (uchar)254 || head[1] != 1 ||
+      !(head[2] == FRM_VER || head[2] == FRM_VER + 1 ||
+        (head[2] >= FRM_VER + 3 && head[2] <= FRM_VER + 4)))
+    // Upon failure, return NULL, but here, we have to close the file first.
```
laurynas-biveinis: Comment differs from code

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+    // We close the file and return success, as we no form info
+    // or extra segment.
+    return false;
+  }
+
+  mysql_file_seek(file, pos, MY_SEEK_SET, MYF(0));
+
+  uchar *extra_segment_buff= static_cast<uchar *>(
+      my_malloc(key_memory_frm_extra_segment_buff, n_length, MYF(MY_WME)));
+  std::unique_ptr<uchar, decltype(&my_free)> extra_segment_buff_uptr(
```
laurynas-biveinis, line 168: Consider backporting map_helpers.h from 8.0, which provide unique_ptr_my_free etc.

laurynas-biveinis (changes requested):
(partial review as discussed)

sql/handler.cc
george-lorch: This is a temporary fix resulting from https://jira.percona.com/browse/PS-4538 and the regression introduced here https://jira.percona.com/browse/PS-3919.  Basically, it seems now that all DDL goes through here and conditions are that it always triggers gap lock detection in REPEATABLE-READ.

laurynas-biveinis: @georgelorchpercona What would be the "permanent" fix for PS-4538? Can it be pushed separately from this PR?

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA
+*/
+
+/*
```
laurynas-biveinis, line 18: Remove

mysql-test/suite/rocksdb/t/native_part_dev.test
vlad-lesin: This test is only for debugging, forgot to remove.

vlad-lesin: This test is only for debugging, forgot to remove.

mysql-test/suite/rocksdb/t/native_part_dev.test
vlad-lesin: This test is only for debugging, forgot to remove.

mysql-test/suite/parts/inc/upgrade_parts_from_prev_ver_alter.inc
vlad-lesin: fixed

mysql-test/suite/parts/inc/upgrade_parts_from_prev_ver_mysql_upgrade.inc
vlad-lesin: fixed

mysql-test/suite/rocksdb/t/upgrade_parts_from_prev_ver.test
vlad-lesin: Just copied from somewhere. Removed.

sql/partitioning/partition_base.h
vlad-lesin: This is the original comment. I was not so bored to fix the original typos :) Fixed.

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+    char   subpart_name[FN_REFLEN];
+    size_t subpart_name_length=
+        tablename_to_filename(elem->partition_name, subpart_name, FN_REFLEN);
+    DBUG_ASSERT(part_name_length + subpart_name_length < sizeof(part_name));
+    strncat(part_name, subpart_name,
+            sizeof(part_name) - part_name_length - 1);
+    part_name_length+= subpart_name_length;
+  }
+
+  create_partition_name(out_buf, path, part_name, NORMAL_PART_NAME, FALSE);
```
laurynas-biveinis: s/FALSE/false/g

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+}
+
+Parts_share_refs::Parts_share_refs() : num_parts(0), ha_shares(NULL) {}
```
laurynas-biveinis: s/NULL/nullptr/g

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+
+
+Parts_share_refs::~Parts_share_refs()
+{
+  uint i;
```
laurynas-biveinis: Declare at use

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1181 @@
...
+ private:
+  uint m_file_tot_parts;  // Debug
+  /*
+    Since the partition handler is a handler on top of other handlers, it
+    is necessary to keep information about what the underlying handler
+    characteristics is. It is not possible to keep any handler instances
+    for this since the MySQL Server sometimes allocating the handler object
+    without freeing them.
+  */
+  ulong m_low_byte_first;
```
laurynas-biveinis: Remove

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+    file at this point if you wish to change the table definition, but there
+    are no methods currently provided for doing that.
+
+    Called from handler.cc by ha_create_table().
+*/
+
+int Partition_base::create(const char *name, TABLE *table_arg,
+                           HA_CREATE_INFO *create_info)
+{
+  int                                   error;
```
laurynas-biveinis: Declare local vars at first use

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+  info(HA_STATUS_AUTO);
+
+  if (!(create_info->used_fields & HA_CREATE_USED_AUTO))
+    create_info->auto_increment_value= stats.auto_increment_value;
+
+  /*
+    DATA DIRECTORY and INDEX DIRECTORY are never applied to the whole
+    partitioned table, only its parts.
+  */
+  my_bool from_alter= (create_info->data_file_name == (const char *)-1);
```
laurynas-biveinis: s/my_bool/bool/g

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+  for (uint i= 0; i < m_part_info->num_parts; ++i)
+  {
+    partition_element *part_elem= part_it++;
+    if (m_is_sub_partitioned)
+    {
+      List_iterator_fast<partition_element> sub_it(part_elem->subpartitions);
+      for (uint j= 0; j < m_part_info->num_subparts; ++j)
+      {
+        partition_element *sub_part_elem= sub_it++;
+        if (!fn(part_elem, sub_part_elem))
```
laurynas-biveinis, line 525: Pass const references, not const pointers, unless they can be legally NULL

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+    }
+    memset(m_file, 0, alloc_len);
+    /*
+      Populate them by cloning the original partitions. This also opens them.
+      Note that file->ref is allocated too.
+    */
+    file= m_is_clone_of->m_file;
+
+    /*
+      The following code can be refactored. The same pattern of iteration
```
laurynas-biveinis: Remove the comment :)

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5823 @@
...
+      /* set file to last handler, so all of them are closed */
+      file= &m_file[m_tot_parts - 1];
+      goto err_handler;
+    }
+  }
+  key_used_on_scan= m_file[0]->key_used_on_scan;
+  implicit_emptied= m_file[0]->implicit_emptied;
+  /*
+    Add 2 bytes for partition id in position ref length.
+    ref_length=max_in_all_par
```
laurynas-biveinis: Remove

sql/table.cc
```diff
@@ -3658,10 +3656,10 @@ void free_blob_buffers_and_reset(TABLE *table, uint32 size)
   @retval The form position.
 */
 
-static ulong get_form_pos(File file, uchar *head)
+ulong get_form_pos(File file, uchar *head)
 {
   uchar *pos, *buf;
-  uint names, length;
+  uint   names, length;
```
laurynas-biveinis: Spurious change

sql/table.h
```diff
@@ -3054,10 +3054,10 @@ inline bool is_user_table(TABLE * table)
 bool is_simple_order(ORDER *order);
 
 void repoint_field_to_record(TABLE *table, uchar *old_rec, uchar *new_rec);
-bool update_generated_write_fields(const MY_BITMAP *bitmap, TABLE *table);
-bool update_generated_read_fields(uchar *buf, TABLE *table,
-                                  uint active_index= MAX_KEY);
-
+bool  update_generated_write_fields(const MY_BITMAP *bitmap, TABLE *table);
```
laurynas-biveinis: Whitespace-only changes above the added declaration

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3917,9 +3921,11 @@ static int rocksdb_init_func(void *const p) {
   rocksdb_hton->flags = HTON_TEMPORARY_NOT_SUPPORTED |
                         HTON_SUPPORTS_EXTENDED_KEYS | HTON_CAN_RECREATE;
 
+  rocksdb_hton->partition_flags = rocksdb_partition_flags;
+
   DBUG_ASSERT(!mysqld_embedded);
 
-  if (rocksdb_db_options->max_open_files > (long) open_files_limit) {
+  if (rocksdb_db_options->max_open_files > (long)open_files_limit) {
```
laurynas-biveinis: Whitespace-only change

storage/rocksdb/ha_rocksdb.h
```diff
@@ -499,7 +499,9 @@ class ha_rocksdb : public my_core::handler {
   /* Array of index descriptors */
   std::shared_ptr<Rdb_key_def> *m_key_descr_arr;
 
-  bool check_keyread_allowed(uint inx, uint part, bool all_parts) const;
+  static bool check_keyread_allowed(bool &pk_can_be_decoded,
+                                    const TABLE_SHARE *table_share, uint inx,
```
laurynas-biveinis, line 503: Is table_share arg required? It seems that it's always the same value - field table_share passed

storage/rocksdb/ha_rocksdb.h
```diff
@@ -835,6 +837,10 @@ class ha_rocksdb : public my_core::handler {
 
   bool init_with_fields() override;
 
+  static ulong index_flags(bool &pk_can_be_decoded,
+                           const TABLE_SHARE *table_share, uint inx, uint part,
```
laurynas-biveinis, line 841: Likewise

storage/rocksdb/ha_rocksdb.h
```diff
@@ -865,6 +871,13 @@ class ha_rocksdb : public my_core::handler {
     return m_store_row_debug_checksums && (rand() % 100 < m_checksums_pct);
   }
 
+  int rename_partitioned_table(const char *const from, const char *const to,
+                               const std::string &partition_string)
+      MY_ATTRIBUTE((__warn_unused_result__));
```
laurynas-biveinis: s/.../MY_NODISCARD/g

storage/tokudb/ha_tokudb.h
```diff
@@ -1000,14 +1000,23 @@ class ha_tokudb : public handler {
     MY_NODISCARD bool check_upsert(THD *thd,
                                    List<Item> &update_fields,
                                    List<Item> &update_values);
-    MY_NODISCARD int send_upsert_message(List<Item> &update_fields,
-                                         List<Item> &update_values,
-                                         DB_TXN *txn);
+    MY_NODISCARD int send_upsert_message(List<Item>& update_fields,
```
laurynas-biveinis: Formatting-only change

laurynas-biveinis (changes requested):
Completed a review pass in individual comments

sql/handler.cc
george-lorch: Reverting PS-3919 and fixing it correctly so that it does not change/create this new behavior that forces the gap lock detection logic to become involved for DDL/ANALYZE/OPTIMIZE ?  Otherwise this is the fix for PS-4538 and should probably be pushed on its own with specific test case(s).

laurynas-biveinis: No opinion on the revert of PS-3919: on one hand, its benefit/regression ratio is extremely bad, on another, it's a 8.0 backport, so prepares us for things there

george-lorch: The thing is, I don't believe we are seeing this same behavior (DDL triggering gap lock detection) in 8.0 (.4, .11, or .12).  So that implies that the backport may be incomplete or is possibly relying on behaviors found in 8.0 but not in 5.7.

sql/partitioning/partition_base.h
vlad-lesin: Yes, removed MyISAM-related stuff.

vlad-lesin: The testing will be latter.

sql/partitioning/partition_base.cc
vlad-lesin: Yes.

sql/partitioning/partition_base.cc
vlad-lesin: Removed the comment. The file will be closed on scope exit.

sql/partitioning/partition_base.cc
vlad-lesin: I would not do this, the pattern works well with standard library. This might make sense in 8.0 to reuse the code, but there is no reason to bring new dependency in 5.7.

sql/partitioning/partition_base.cc
vlad-lesin: Maybe add our (c) instead of removing?

sql/handler.cc
george-lorch: OK, I am correcting myself here, yes, I am now seeing this this behavior in Percona Server 8.0.  Consider this statement `alter table t modify h varchar(31) character set cp1257 collate cp1257_bin not null`.  In 5.7 this did NOT trigger a gap lock detection, in Percona Server 8.0 it does.  So it seems that proper fix here is to take this particular piece of this changeset into 5.7, either here or as a stand along change, and port it forward to 8.0.

sql/partitioning/partition_base.cc
laurynas-biveinis: No, remove the comment block below, not above

sql/partitioning/partition_base.cc
laurynas-biveinis: OK

sql/partitioning/partition_base.cc
vlad-lesin: fixed

sql/partitioning/partition_base.cc
vlad-lesin: fixed

sql/partitioning/partition_base.cc
vlad-lesin: fixed

sql/partitioning/partition_base.h
vlad-lesin: removed

sql/partitioning/partition_base.cc
vlad-lesin: done

sql/partitioning/partition_base.cc
vlad-lesin: removed

sql/partitioning/partition_base.cc
vlad-lesin: done

sql/table.cc
vlad-lesin: Yes, this is autoformatting with clang-format-diff. Removed as discussed.

sql/table.h
vlad-lesin: The same, fixed.

storage/rocksdb/ha_rocksdb.cc
vlad-lesin: The same, fixed.

storage/rocksdb/ha_rocksdb.h
vlad-lesin: Yes, but the function is static after the changes. It's necessary to reuse the code outside the class.

storage/rocksdb/ha_rocksdb.h
vlad-lesin: Likewise

storage/rocksdb/ha_rocksdb.h
vlad-lesin: It is fixed for the changed code.

storage/tokudb/ha_tokudb.h
vlad-lesin: fixed

sql/partitioning/partition_base.cc
vlad-lesin: I agree, but the first parameter can be NULL, the second one can't be NULL. It is not so essentially for me, I can fix if you insist. But I followed the principle of uniformity, it would look strange if two parameters of the same type and meaning were declared differently. What do you think?

sql/partitioning/partition_base.cc
vlad-lesin: removed

sql/partitioning/partition_base.cc
laurynas-biveinis: I agree, OK to leave both as is

mysql-test/suite/rocksdb.sys_vars/r/all_vars.result
```diff
@@ -9,5 +9,7 @@ There should be *no* long test name listed below:
 select variable_name as `There should be *no* variables listed below:` from t2
 left join t1 on variable_name=test_name where test_name is null ORDER BY variable_name;
 There should be *no* variables listed below:
+rocksdb_enable_native_partition
```
laurynas-biveinis: Please add a sys_vars test

mysql-test/suite/rocksdb.rpl/combinations
```diff
@@ -12,3 +12,25 @@ binlog-format=statement
 
 [mix]
 binlog-format=mixed
+
+[row-write-committed-native-partitioning]
+mtr-result-dir = r-native-partitioning
+mtr-tests-list = rpl_rocksdb_extra_col_slave
```
laurynas-biveinis, line 18: Is it possible to have something like "mtr-tests-list = *" here or omit it altogether for the same effect?

laurynas-biveinis (changes requested):
Can you show how the MTR run log runs with your added combinations?
The MTR combination feature looks good, but I am not sure whether its application here is as good - it adds half a million .result file lines, which are impossible to review manually. The way this ideally should be done that the tests run in combinations, but using a single output file, by disabling warnings etc. @georgelorchpercona what do you think?

vlad-lesin:
> Can you show how the MTR run log runs with your added combinations?
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1908
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1918 - the latest run for minor fixes,


> The MTR combination feature looks good, but I am not sure whether its application here is as good - it adds half a million .result file lines, which are impossible to review manually. The way this ideally should be done that the tests run in combinations, but using a single output file, by disabling warnings etc. 

Yes, but what is the alternative? You proposed to use "set parameters, include common test file" pattern which would not only add new *.result files, but also *.test files and necessity to port then on 8.0.

About warnings suppression. This solution requires fixes for each test of several tens of tests, and the fixes must be selective, otherwise we can miss some warning which would lead to test failing. Besides, there are some tests which return warning text in query results, like:

+ALTER TABLE t1 OPTIMIZE PARTITION p0, p1;
 Table  Op      Msg_type        Msg_text
 test.t1        optimize        status  OK
+test.t1        optimize        warning The partition engine, used by table 'test.t1', is deprecated and will be removed in a future release. Please use native partitioning instead.

All the above requires additional time for fixes.

At the other hand, you question is reasonable. The solution adds several tens of 'result' files, and how to check them?

My answer is:

1) All files in "r" directories must be unchangeable when the commit is squashed, because native partitioning is switched off by default.

2) All files in "r-native-partitioning" directories can be checked with 'diff -u r/file.resut r-native-partitioning/file.result'. I could attach the script which creates big diff for all .result files for native partitioning switched on.

george-lorch:
The crux of the problem with the test results is that we can not conditionally 'expect' or 'ignore' a specific warning with the mtr framework.  This prevents us from having a single test with a common result for tests that use partitioning.  I do not like the idea of fully disabling warnings for the operations where the deprecation warning appears, it can easily mask real problems.  To me, the separate result files for combinations seems to me like a logical extension of the combinations functionality, but, the pain is as you point out @laurynas-biveinis that the initial commit implementation has a rather insane and unreviewable number of result file changes/movements.  The interesting thing though is because this duality of native vs. generic only exists in 5.7, the port to 8.0 pretty much completely reverts this work and leaves us back with no combinations and only native tests and results.   In other words :-\ </shrug>

laurynas-biveinis:
I am looking at https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1908/BUILD_TYPE=debug,Host=min-bionic-x64/consoleText and I am not seeing partitions combinations tests - can you point me to them?

@georgelorchpercona @vlad-lesin let's go with the current implementation in 5.7 and back to regular MTR in 8.0. It is to our advantage that we don't have to port it forward - the 8.0 tests will be reasonable, and 5.7 ones should be relatively stable.

vlad-lesin:
To the question how to examine result files. The following commands can be used from source dir:

``
for r in `find mysql-test -path '*/r-native*' -type f -name *.result`; do suite_dir=$(dirname $(dirname $r)); res_file=$(basename $r); diff -u $suite_dir/r/$res_file $r; done | colordiff | less -R
``

vlad-lesin:
> I am looking at https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1908/BUILD_TYPE=debug,Host=min-bionic-x64/consoleText and I am not seeing partitions combinations tests - can you point me to them?

Yes, you are right, 1908 - this is the old build. You can find partition combination tests here:

https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1918/BUILD_TYPE=debug,Host=min-bionic-x64/consoleText

by "native_partitioning" key.

mysql-test/suite/rocksdb.rpl/combinations
vlad-lesin: Yes, if that option is omitted, all tests from the suite will take part in the combination testing. There can be the situation when, for example, for one set of options we need to test the whole suite (like 'write-committed' and 'write-prepared' in MyRocks), but for the other set of options we need to run only the certain tests(which are selected in this option).

laurynas-biveinis:
Thanks, I see the test run now. @vlad-lesin, it also needs a --big-test run (ok to limit to the relevant suites)

mysql-test/suite/rocksdb.sys_vars/r/all_vars.result
vlad-lesin: Added both for MyRocks and TokuDB.

vlad-lesin:
> Thanks, I see the test run now. @vlad-lesin, it also needs a --big-test run (ok to limit to the relevant suites)

https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1919

vlad-lesin:
Ready for the next review round.

mysql-test/suite/rocksdb.sys_vars/r/all_vars.result
laurynas-biveinis: With tests added, this all_vars diff should have disappeared

mysql-test/suite/rocksdb/t/upgrade_parts_from_prev_ver.test
```diff
@@ -0,0 +1,31 @@
+--source include/have_rocksdb.inc
```
laurynas-biveinis, line 2: Any unzip-calling test, directly or through an include file, should have --source include/have_util_unzip.inc

sql/CMakeLists.txt
laurynas-biveinis: Comment not addressed

sql/handler.cc
laurynas-biveinis: @vlad-lesin @georgelorchpercona what's the state of this?

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5581 @@
...
+          return false;
+
+        table_level_options.put_to_info(create_info);
+        ++file;
+        part_name_collection.push_back(name_buff);
+        return true;
+      }))
+    DBUG_RETURN(0);
+
+  auto part_name_i= part_name_collection.begin();
```
laurynas-biveinis: cbegin

sql/partitioning/partition_base.cc
```diff
@@ -0,0 +1,5581 @@
...
+}
+
+
+/*
+  Disabled since it is not possible to prune yet.
+  without pruning, it need to rebind/unbind every partition in every
+  statement which uses a table from the table cache. Will also use
+  as many PSI_tables as there are partitions.
+*/
+#ifdef HAVE_M_PSI_PER_PARTITION
```
laurynas-biveinis: Just remove if this is a never-enabled define

sql/partitioning/partition_base.h
laurynas-biveinis: Comment not addressed

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1137 @@
...
+    Call to hint about semi consistent read
+  */
+  virtual void try_semi_consistent_read(bool);
+
+  /*
+    NOTE: due to performance and resource issues with many partitions,
+    we only use the m_psi on the Partition_base handler, excluding all
+    partitions m_psi.
+  */
+#ifdef HAVE_M_PSI_PER_PARTITION
```
laurynas-biveinis: Remove if undefined

sql/partitioning/partition_base.h
laurynas-biveinis: Comment not addressed

sql/partitioning/partition_base.h
```diff
@@ -0,0 +1,1137 @@
...
+    (MyISAM)
+
+    HA_FAST_KEY_READ:
+    Setting this flag indicates that the handler is equally fast in
+    finding a row by key as by position.
+    This flag is used in a very special situation in conjunction with
+    filesort's. For further explanation see intro to init_read_record.
+    (BDB, HEAP, InnoDB)
+
+    HA_nullptr_IN_KEY:
```
laurynas-biveinis: s/NULL/nullptr should not have been done indiscriminately, please check elsewhere too

sql/sql_table.cc
```diff
@@ -2966,24 +2966,34 @@ bool quick_rm_table(THD *thd, handlerton *base, const char *db,
...
+    (for example, partition info).
+  */
+  size_t path_length= build_table_filename(path, sizeof(path) - 1, db,
+                                           table_name, reg_ext, flags);
+  strncpy(frm_path, path, sizeof(frm_path) - 1);
+  path[path_length - reg_ext_length]= '\0';  // Remove reg_ext
   if (flags & NO_HA_TABLE)
   {
-    handler *file= get_new_handler((TABLE_SHARE*) 0, thd->mem_root, base);
+    handler *file= get_new_handler((TABLE_SHARE *)0, thd->mem_root, base);
```
laurynas-biveinis: Spurious change

sql/sql_table.cc
```diff
@@ -2966,24 +2966,34 @@ bool quick_rm_table(THD *thd, handlerton *base, const char *db,
...
-    (void) file->ha_create_handler_files(path, NULL, CHF_DELETE_FLAG, NULL);
+    {
+      error|= true;
+      goto exit;
+    }
+    (void)file->ha_create_handler_files(path, NULL, CHF_DELETE_FLAG, NULL);
     delete file;
   }
-  if (!(flags & (FRM_ONLY|NO_HA_TABLE)))
+  if (!(flags & (FRM_ONLY | NO_HA_TABLE)))
```
laurynas-biveinis: Spurious change

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4767,6 +4782,18 @@ void Rdb_open_tables_map::release_table_handler(
...
                                        my_core::TABLE_SHARE *const table_arg,
                                        my_core::MEM_ROOT *const mem_root) {
+
+  if (rocksdb_enable_native_partition &&
+      table_arg && table_arg->db_type() == rocksdb_hton &&
+      table_arg->partition_info_str && table_arg->partition_info_str_len) {
+    ha_rockspart *file = new (mem_root) ha_rockspart(hton, table_arg);
+    if (file && file->init_partitioning(mem_root)) {
+      delete file;
+      return (NULL);
```
laurynas-biveinis: nullptr

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7212,13 +7241,16 @@ int ha_rocksdb::read_row_from_secondary_key(uchar *const buf,
...
 */
 
-ulong ha_rocksdb::index_flags(uint inx, uint part, bool all_parts) const {
+ulong ha_rocksdb::index_flags(bool &pk_can_be_decoded,
+                              const TABLE_SHARE *table_share, uint inx,
+                              uint part, bool all_parts) {
   DBUG_ENTER_FUNC();
 
-  ulong base_flags = HA_READ_NEXT | // doesn't seem to be used
+  ulong base_flags = HA_READ_NEXT |  // doesn't seem to be used
```
laurynas-biveinis: Spurious change

storage/rocksdb/ha_rockspart.cc
```diff
@@ -0,0 +1,106 @@
...
+
+handler *ha_rockspart::clone(const char *name, MEM_ROOT *mem_root) {
+  ha_rockspart *new_handler;
+
+  DBUG_ENTER("native_part::Partition_base::clone");
+
+  /* If this->table == NULL, then the current handler has been created but not
+  opened. Prohibit cloning such handler. */
+  if (!table)
+    DBUG_RETURN(NULL);
```
laurynas-biveinis: nullptr

storage/rocksdb/ha_rockspart.cc
```diff
@@ -0,0 +1,106 @@
...
+
+  new_handler =
+      new (mem_root) ha_rockspart(ht, table_share, m_part_info, this, mem_root);
+  if (!new_handler)
+    DBUG_RETURN(NULL);
```
laurynas-biveinis: nullptr

storage/rocksdb/ha_rockspart.cc
```diff
@@ -0,0 +1,106 @@
...
+    goto err;
+
+  new_handler->m_pk_can_be_decoded = m_pk_can_be_decoded;
+  new_handler->set_pk_can_be_decoded_for_each_partition();
+
+  DBUG_RETURN((handler *)new_handler);
+
+err:
+  delete new_handler;
+  DBUG_RETURN(NULL);
```
laurynas-biveinis: nullptr

storage/tokudb/ha_tokupart.cc
```diff
@@ -0,0 +1,83 @@
...
+
+handler *ha_tokupart::clone(const char *name, MEM_ROOT *mem_root) {
+    ha_tokupart *new_handler;
+
+    DBUG_ENTER("native_part::Partition_base::clone");
+
+    /* If this->table == NULL, then the current handler has been created but not
+    opened. Prohibit cloning such handler. */
+    if (!table)
+        DBUG_RETURN(NULL);
```
laurynas-biveinis: nullptr

storage/tokudb/ha_tokupart.cc
```diff
@@ -0,0 +1,83 @@
...
+
+    new_handler = new (mem_root)
+        ha_tokupart(ht, table_share, m_part_info, this, mem_root);
+    if (!new_handler)
+        DBUG_RETURN(NULL);
```
laurynas-biveinis: nullptr

storage/tokudb/ha_tokupart.cc
```diff
@@ -0,0 +1,83 @@
...
+                             name,
+                             table->db_stat,
+                             HA_OPEN_IGNORE_IF_LOCKED | HA_OPEN_NO_PSI_CALL))
+        goto err;
+
+    DBUG_RETURN((handler *)new_handler);
+
+err:
+    delete new_handler;
+    DBUG_RETURN(NULL);
```
laurynas-biveinis: nullptr

storage/tokudb/hatoku_hton.cc
```diff
@@ -655,11 +659,25 @@ static int tokudb_done_func(TOKUDB_UNUSED(void* p)) {
...
+static handler* tokudb_create_handler(handlerton* hton,
+                                      TABLE_SHARE* table,
+                                      MEM_ROOT* mem_root) {
+    if (tokudb::sysvars::enable_native_partition &&
+        table && table->db_type() == tokudb_hton && table->partition_info_str &&
+        table->partition_info_str_len) {
+        ha_tokupart* file = new (mem_root) ha_tokupart(hton, table);
+        if (file && file->init_partitioning(mem_root)) {
+            delete file;
+            return (NULL);
```
laurynas-biveinis: nullptr

laurynas-biveinis (changes requested):

vlad-lesin:
The comment about this https://github.com/percona/percona-server/pull/2493#discussion_r220522575 note.

If we remove the change there will be test fails like the following:
```
--- ./mysql-test/suite/rocksdb/r/partition.result   2018-09-27 15:55:34.739805720 +0300
+++ build/mysql-test/var/44/log/partition.reject      2018-09-27 16:08:47.857325052 +0300
@@ -12,10 +12,14 @@
 test.t1        analyze status  OK
 test.t1        analyze warning The partition engine, used by table 'test.t1', is deprecated and will be removed in a future release. Please use native partitioning instead.
 Table  Op      Msg_type        Msg_text
-test.t1        repair  status  OK
+test.t1        repair  error   Partition p0 returned error
+test.t1        repair  Error   Using Gap Lock without full unique key in multi-table or multi-statement transactions is not allowed. You need to either rewrite queries to use all unique key columns in WHERE equal conditions, or rewrite to single-table, single-statement transaction.  Query: ALTER TABLE t1 REPAIR PARTITION p0,p1
+test.t1        repair  error   Unknown - internal error 149 during operation
 test.t1        repair  warning The partition engine, used by table 'test.t1', is deprecated and will be removed in a future release. Please use native partitioning instead.
 Table  Op      Msg_type        Msg_text
-test.t1        check   status  OK
+test.t1        check   error   Partition p1 returned error
+test.t1        check   Error   Using Gap Lock without full unique key in multi-table or multi-statement transactions is not allowed. You need to either rewrite queries to use all unique key columns in WHERE equal conditions, or rewrite to single-table, single-statement transaction.  Query: ALTER TABLE t1 CHECK PARTITION p1
+test.t1        check   error   Unknown - internal error 149 during operation
 test.t1        check   warning The partition engine, used by table 'test.t1', is deprecated and will be removed in a future release. Please use native partitioning instead.
 Warnings:
```

See also https://jira.percona.com/browse/PS-4762.

mysql-test/suite/rocksdb/t/upgrade_parts_from_prev_ver.test
vlad-lesin: fixed

sql/partitioning/partition_base.cc
vlad-lesin: removed

sql/partitioning/partition_base.h
vlad-lesin: removed

vlad-lesin:
Also comments were added for https://github.com/percona/percona-server/pull/2493#discussion_r220522022 and https://github.com/percona/percona-server/pull/2493#discussion_r220547944, and removed for https://github.com/percona/percona-server/pull/2493#discussion_r220546933.

vlad-lesin:
https://github.com/percona/percona-server/pull/2493#discussion_r220519932 - sys_vars testing: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1925/ .

vlad-lesin:
Ready for the next round.

laurynas-biveinis:
I don't see any added comments for e.g. sql/cmakelists.txt comment

vlad-lesin:
sql/CMakeLists.txt is fixed, see build tests here:  https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1926/ .

vlad-lesin:
Ready for the next round.

sql/CMakeLists.txt
```diff
@@ -475,6 +475,18 @@ ELSE()
   SET(MYSQLD_SOURCE main.cc ${DTRACE_PROBES_ALL})
 ENDIF()
 
+# partition_base.cc contains c++11 code, what is reasonable
```
laurynas-biveinis: s/what/which

sql/partitioning/partition_base.cc
laurynas-biveinis: Not done?

sql/partitioning/partition_base.h
laurynas-biveinis: Not done?

laurynas-biveinis (changes requested):
Please let me know if again some of the comments appear to be replies to hidden/missing comments
Is @georgelorchpercona OK with https://github.com/percona/percona-server/pull/2493/files#r210890922 ?

sql/partitioning/partition_base.h
vlad-lesin: Yes, should be done now.

vlad-lesin:
All the above fixes are done.

sql/handler.cc
george-lorch: PS-3919 was reverted, this change should no longer be necessary.

laurynas-biveinis (approved):
LGTM, please address @georgelorchpercona comment and merge

sql/handler.cc
george-lorch: @vlad-lesin ^^ PS-3919 was reverted, this should no longer be needed and there should be no 'gap lock' related errors, please be sure you are re-based to include that commit https://github.com/Percona/percona-server/commit/4d803cdd69ec88a65e92a0b3fdadd75db4ce59c3

vlad-lesin: Rebased on the newest 5.7, removed the above changes and tested: https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1933/ . The results look good to me.

george-lorch: Yup, looks good.

george-lorch (approved):

PR #2495, opened 2018-08-16
PS-275: Deprecate and obsolete max_binlog_files with binlog_space_limit (5.7)

inikep:
Implementation details:
- binlog_space_limit can be set with --binglog-space-limit (it's 0 that is disabled by default)
- binlog_space_limit is checked during each write to binlog, server start, log rotate, FLUSH LOGS, RESET MASTER

- binlog_space_total is a total size of binary logs (except the active one)
- binlog_space_total is updated during server start, log rotate, FLUSH LOGS, RESET MASTER, PURGE BINARY LOGS [TO/BEFORE]


inikep:
https://ps.cd.percona.com/job/percona-server-5.7-param/44/

mysql-test/r/mysqld--help-notwin.result
```diff
@@ -577,7 +580,8 @@ The following options may be given as the first argument:
  Maximum number of binlog files. Used with
  --max-binlog-size this can be used to limit the total
  amount of disk space used for the binlog. Default is 0,
- don't limit.
+ don't limit.This variable is deprecated and will be
```
laurynas-biveinis: Space after dot

mysql-test/suite/sys_vars/t/binlog_space_limit_basic.test
```diff
@@ -0,0 +1,88 @@
...
+
+#
+# show that it's read-only
+#
+--error ER_INCORRECT_GLOBAL_LOCAL_VAR
+set global binlog_space_limit=1;
+--error ER_INCORRECT_GLOBAL_LOCAL_VAR
+set session binlog_space_limit=1;
+
+FLUSH LOGS;
```
laurynas-biveinis: Please don't mix sys_vars and functional test - keep only the variable properties (only global, read only) in this test, create a new test in binlog suite for the actual functionality

mysql-test/suite/sys_vars/t/binlog_space_limit_basic.test
```diff
@@ -0,0 +1,88 @@
...
+# binary.000001 should be deleted now
+SHOW BINARY LOGS;
+
+CREATE TABLE `t1` (
+  `v1` int(11) DEFAULT NULL,
+  `v2` int(11) DEFAULT NULL,
+  KEY `v1` (`v1`)
+) ENGINE=MyISAM DEFAULT CHARSET=latin1;
```
laurynas-biveinis: This would require include/have_myisam.inc, but any particular reason for a MyISAM table? If possible, just do not specify the engine

sql/binlog.cc
```diff
@@ -6790,6 +6794,170 @@ int MYSQL_BIN_LOG::purge_index_entry(THD *thd, ulonglong *decrease_log_space,
...
+  @retval
+    LOG_INFO_PURGE_NO_ROTATE	Binary file that can't be rotated
+    LOG_INFO_FATAL              if any other than ENOENT error from
+                                mysql_file_stat() or mysql_file_delete()
+*/
+
+int MYSQL_BIN_LOG::count_binlog_space(bool need_lock_index) {
+  int error;
+  LOG_INFO log_info;
+  MY_STAT stat_area;
```
laurynas-biveinis: Declare locals at first use, unless gotos prohibit

sql/binlog.cc
```diff
@@ -6790,6 +6794,170 @@ int MYSQL_BIN_LOG::purge_index_entry(THD *thd, ulonglong *decrease_log_space,
...
+                                mysql_file_stat() or mysql_file_delete()
+*/
+
+int MYSQL_BIN_LOG::purge_logs_by_size(bool need_lock_index) {
+  int error = 0;
+  char to_log[FN_REFLEN];
+  LOG_INFO log_info;
+  MY_STAT stat_area;
+  THD *thd = current_thd;
+  my_off_t binlog_pos;
```
laurynas-biveinis: Declare locals at first use, unless gotos prohibit

sql/binlog.cc
```diff
@@ -6790,6 +6794,170 @@ int MYSQL_BIN_LOG::purge_index_entry(THD *thd, ulonglong *decrease_log_space,
...
+  if (need_lock_index)
+    mysql_mutex_lock(&LOCK_index);
+  else
+    mysql_mutex_assert_owner(&LOCK_index);
+
+  count_binlog_space(false);
+  binlog_pos= my_b_tell(&log_file);
+  if (!binlog_space_total ||
+      binlog_space_total + binlog_pos <= binlog_space_limit)
+    goto err;
```
laurynas-biveinis: This is also a not-error exit, thus "err" is better name something else

sql/binlog.cc
```diff
@@ -6790,6 +6794,170 @@ int MYSQL_BIN_LOG::purge_index_entry(THD *thd, ulonglong *decrease_log_space,
...
+        if (thd) {
+          push_warning_printf(thd, Sql_condition::SL_WARNING,
+                              ER_BINLOG_PURGE_FATAL_ERR,
+                              "a problem with getting info on being purged %s; "
+                              "consider examining correspondence "
+                              "of your binlog index file "
+                              "to the actual binlog files",
+                              log_info.log_file_name);
+        } else {
+          sql_print_information("Failed to delete log file '%s'",
```
laurynas-biveinis: Is the error message correct - we did not fail to delete, we failed to stat?

sql/binlog.cc
```diff
@@ -6790,6 +6794,170 @@ int MYSQL_BIN_LOG::purge_index_entry(THD *thd, ulonglong *decrease_log_space,
...
+                                log_info.log_file_name);
+        }
+        error = LOG_INFO_FATAL;
+        goto err;
+      }
+    }
+    /* check if the binary log file is older than the purge_time
```
laurynas-biveinis: Not "older than the purge_time"

sql/binlog.cc
```diff
@@ -8282,6 +8454,14 @@ bool MYSQL_BIN_LOG::write_cache(THD *thd, binlog_cache_data *cache_data,
     update_thd_next_event_pos(thd);
   }
 
+#ifdef HAVE_REPLICATION
+  if (binlog_space_limit && binlog_space_total &&
+      binlog_space_total + my_b_tell(&log_file) > binlog_space_limit)
+  {
+    purge_logs_by_size(true);
```
laurynas-biveinis: I am not sure we should purge potentially after each write. Purging after close/reopen/rotate etc. operations - i.e. at file granularity - should be OK, even if that means that the space limit will be intermittently violated by a larger value. Stalls in a middle of binlog files would be unexpected to users.

sql/binlog.cc
```diff
@@ -6790,6 +6794,170 @@ int MYSQL_BIN_LOG::purge_index_entry(THD *thd, ulonglong *decrease_log_space,
...
+      strmake(to_log, log_info.log_file_name,
+              sizeof(log_info.log_file_name) - 1);
+    } else
+      break;
+    if (find_next_log(&log_info, false /*need_lock_index=false*/)) break;
+  }
+
+  error = (to_log[0] ? purge_logs(to_log, true, false /*need_lock_index=false*/,
+                                  true /*need_update_threads=true*/,
+                                  (ulonglong *)0, true)
```
laurynas-biveinis: s/(ulonglong *)0/NULL

laurynas-biveinis (changes requested):

sql/binlog.cc
inikep: `max_binlog_size` doesn't wait for close/reopen/rotate. When log is bigger than `max_binlog_size`, server rotates and purges logs. Doc says:
```
If a write to the binary log causes the current log file size to exceed the value of this variable, the server rotates the binary logs
```

inikep: I designed `binlog_space_limit` after `relay_log_space_limit` which is also checked in each `next_event()` call.

laurynas-biveinis: OK then. Just make sure that no filesystem binlog enumeration happens for each binlog write, but only in the event of space limit violation.

inikep: `binlog_space_total` is created to reduce binlog enumeration. It is updated after log rotation or RESET MASTER. The following condition:
```
  if (binlog_space_limit && binlog_space_total &&
      binlog_space_total + my_b_tell(&log_file) > binlog_space_limit)
```
is true only when logs need to be purged. After purging logs the condition will be false.
In other words `purge_logs_by_size(true)` should be called only once per each purged file.

inikep:
All comments are addressed. The current version is purge without log rotation after exceeding the `binlog_space_limit`. Tested at
https://ps.cd.percona.com/job/percona-server-5.7-param/50/

sql/binlog.cc
```diff
@@ -6790,6 +6794,167 @@ int MYSQL_BIN_LOG::purge_index_entry(THD *thd, ulonglong *decrease_log_space,
...
+  Count a total size of binary logs (except the active one) to the variable
+  binlog_space_total.
+
+  @param need_lock_index  If true, this function acquires LOCK_index;
+                          otherwise the caller should already have acquired it.
+
+  @retval
+    0				ok
+  @retval
+    LOG_INFO_PURGE_NO_ROTATE	Binary file that can't be rotated
```
laurynas-biveinis: LOG_INFO_PURGE_NO_ROTATE does not look like a value that could be returned by this function. Looking at the code, it's either 0, LOG_INFO_FATAL, or one of MYSQL_BIN_LOG::find_log_pos return values (LOG_INFO_IO, LOG_INFO_EOF)

sql/binlog.cc
```diff
@@ -9982,6 +10151,15 @@ int MYSQL_BIN_LOG::ordered_commit(THD *thd, bool all, bool skip_commit)
     else if (check_purge)
       purge();
   }
+#ifdef HAVE_REPLICATION
+  else
```
laurynas-biveinis: Why is this in "else"? Shouldn't purging by size happen regardless of the if condition?

sql/sys_vars.cc
```diff
@@ -1678,13 +1678,22 @@ static Sys_var_ulong Sys_expire_logs_days(
        GLOBAL_VAR(expire_logs_days),
        CMD_LINE(REQUIRED_ARG), VALID_RANGE(0, 99), DEFAULT(0), BLOCK_SIZE(1));
 
+static Sys_var_ulonglong Sys_binlog_space_limit(
+       "binlog_space_limit", "Maximum space to use for all binary logs. "
+       "Default is 0, this feature is disabled.",
+       READ_ONLY GLOBAL_VAR(binlog_space_limit), CMD_LINE(REQUIRED_ARG),
+       VALID_RANGE(0, ULONG_MAX), DEFAULT(0), BLOCK_SIZE(1));
```
laurynas-biveinis, line 1685: ulonglong var - ULLONG_MAX for the maximum allowed value?

laurynas-biveinis (changes requested):

sql/binlog.cc
inikep: I used `else` because in case of `true` there is `purge()` which also unconditionally calls `purge_logs_by_size(true)`. But you are right that without `else` it should be simpler and more readable.

sql/sys_vars.cc
inikep: This is copied from `relay_log_space_limit`. `ULONG_MAX` depends on CPU architecture:
ULONG_MAX Value (64-bit platforms): 18446744073709551615
ULONG_MAX Value (32-bit platforms): 4294967295

inikep:
The last commit fixes comments and `else`. Tested at:
https://ps.cd.percona.com/job/percona-server-5.7-param/55/

laurynas-biveinis (approved):

PR #2511, opened 2018-08-20
Fixed ps-4752 by removing the sentence that refers to the data at res…

borys-belinsky:
…t encryption feature as beta


dbadapt (approved):
LGTM

PR #2520, opened 2018-08-27
PS-269: Fix percona_userstat MTR test

inikep:
Fix percona_userstat MTR test
Fix PS-4781: sql_yacc.yy uses SQLCOM_SELECT instead of SQLCOM_SHOW_XXXX_STATS

### DIFF
--- a/mysql-test/t/percona_userstat.test
+++ b/mysql-test/t/percona_userstat.test
@@ -139,7 +139,7 @@ FROM
 --echo # Bug lp:1659992 "Function over userstat tables cause mysql to crash"
 
 delimiter |;
-CREATE FUNCTION utility_get_global_variable(in_name VARCHAR(64)) RETURNS VARCHAR(1024) CHARSET utf8 DETERMINISTIC SQL SECURITY INVOKER
+CREATE FUNCTION utility_get_global_variable(in_name VARCHAR(64)) RETURNS VARCHAR(1024) CHARSET UTF8MB4 DETERMINISTIC SQL SECURITY INVOKER
 BEGIN
   DECLARE var_value VARCHAR(1024);
   SELECT v.variable_value FROM performance_schema.global_variables v WHERE v.variable_name = in_name INTO var_value;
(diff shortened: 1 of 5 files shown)


mysql-test/t/percona_userstat.test
```diff
@@ -139,7 +139,7 @@ FROM
 --echo # Bug lp:1659992 "Function over userstat tables cause mysql to crash"
 
 delimiter |;
-CREATE FUNCTION utility_get_global_variable(in_name VARCHAR(64)) RETURNS VARCHAR(1024) CHARSET utf8 DETERMINISTIC SQL SECURITY INVOKER
+CREATE FUNCTION utility_get_global_variable(in_name VARCHAR(64)) RETURNS VARCHAR(1024) CHARSET utf8mb3 DETERMINISTIC SQL SECURITY INVOKER
```
laurynas-biveinis: Let's do actual UTF-8 (utf8mb4)

sql/sql_yacc.yy
```diff
@@ -12619,35 +12619,35 @@ show_param:
         | CLIENT_STATS_SYM opt_wild_or_where
           {
            LEX *lex= Lex;
-           Lex->sql_command= SQLCOM_SELECT;
+           Lex->sql_command= SQLCOM_SHOW_PLUGINS;
```
laurynas-biveinis: Why? (Any discussion previously on Slack needs to be here and in the commit message later)

laurynas-biveinis (changes requested):

sql/sql_yacc.yy
inikep: A description added above and in the commit message.

laurynas-biveinis: "SQLCOM_SELECT implementation in sql_parse.cc is rewritten in 8.0." - why the rewritten implementation no longer works for userstat? And why SQLCOM_SHOW_PLUGINS` is a good replacement?

laurynas-biveinis (approved):

PR #2529, opened 2018-08-31
PS-3945: Convert user statistics time columns to floating point

inikep:


sql/sql_show.cc
```diff
@@ -5502,11 +5500,11 @@ static ST_FIELD_INFO user_stats_fields_info[] = {
      MY_I_S_UNSIGNED, "Total_connections", SKIP_OPEN_TABLE},
     {"CONCURRENT_CONNECTIONS", MY_INT64_NUM_DECIMAL_DIGITS, MYSQL_TYPE_LONGLONG,
      0, MY_I_S_UNSIGNED, "Concurrent_connections", SKIP_OPEN_TABLE},
-    {"CONNECTED_TIME", MY_INT64_NUM_DECIMAL_DIGITS, MYSQL_TYPE_LONGLONG, 0,
+    {"CONNECTED_TIME", MY_INT64_NUM_DECIMAL_DIGITS, MYSQL_TYPE_DOUBLE, 0,
```
laurynas-biveinis, line 5503: There are testcases that check I_S schema - please grep the testsuite for the changed columns

laurynas-biveinis (changes requested):

inikep:
Re-recorded `mysql-test/suite/funcs_1/r/is_columns_is.result`

sql/sql_show.cc
laurynas-biveinis: MY_INT64_NUM_DECIMAL_DIGITS might work as display length for double too but even then it needs a better mnemonic

laurynas-biveinis (changes requested):

sql/sql_show.cc
inikep: I kept `MY_INT64_NUM_DECIMAL_DIGITS` because MariaDB did so.

PR #2537, opened 2018-09-04
PS-281: Convert LP 1354988 fix to new "SHOW EFFECTIVE GRANTS"

inikep:


sql/sql_yacc.yy
```diff
@@ -1244,6 +1244,7 @@ void warn_about_deprecated_national(THD *thd)
 %token<keyword> TOKU_SMALL_SYM
 %token<keyword> TOKU_DEFAULT_SYM
 %token<keyword> USER_STATS_SYM
+%token<keyword> EFFECTIVE_SYM
```
laurynas-biveinis: Add a header comment "Tokens from Percona Server 8.0" for this one

sql/auth/sql_authorization.cc
```diff
@@ -1184,8 +1185,12 @@ void get_sp_access_map(
     */
 
     if (!strcmp(acl_user->user, user) &&
-        grant_proc->host.compare_hostname(acl_user->host.get_host(),
-                                          acl_user->host.get_host())) {
+        ((!effective_grants &&
```
laurynas-biveinis: The condition (!effective_grants && !my_strcasecmp()) || (effective_grants && compare_hostname()) is repeated several times - please extract (and remove one of the "effective_grants" in effective_grants || !effective_grants)

laurynas-biveinis (changes requested):

sql/auth/sql_authorization.cc
```diff
@@ -1164,10 +1164,19 @@ void make_table_privilege_statement(THD *thd, ACL_USER *role,
   }
 }
 
+static bool hosts_match_for_grants(ACL_HOST_AND_IP &acl_db_host,
```
laurynas-biveinis, line 1167: Can this arg be const ref?

laurynas-biveinis (approved):

sql/auth/sql_authorization.cc
inikep: no

PR #2540, opened 2018-09-06
PS-4811 : 5.7 Merge and fixup for old DB-937 introduces possible regr…

george-lorch:
…ession

- Reverted refactoring and most of DB-937 as well as entire functionality of
  former feature pk_insert_mode.  DB-937 rendered the optimization ineffective
  by forcing the FT layer to do a unique check anyway when setting put flags to
  DB_NOOVERWRITE.  So, simply removing the feature entirely makes more sense to
  ensure index consistency, keep the code simpler and cleaner, and to allow the
  engine to adhere to the server requirement to properly return
  HA_ERR_FOUND_DUPP_KEY on REPLACE INTO/INSERT IGNORE when a record already
  exists.  This is mandatory for 8.0 and replication so to keep the trees
  aligned and maintain some sanity and consistency, we went this way rather than
  trying to keep the feature limping along, particularly now that we have the
  'fast insert/upsert' feature where a user can now use this explicitly.

### DIFF
--- a/mysql-test/suite/tokudb/r/i_s_tokudb_lock_waits_released.result
+++ b/mysql-test/suite/tokudb/r/i_s_tokudb_lock_waits_released.result
@@ -1,7 +1,5 @@
-set default_storage_engine='tokudb';
 set tokudb_prelock_empty=false;
-drop table if exists t;
-create table t (id int primary key);
+create table t (id int primary key) engine=tokudb;
 t should be empty
 select trx_id,trx_mysql_thread_id from information_schema.tokudb_trx;
 trx_id	trx_mysql_thread_id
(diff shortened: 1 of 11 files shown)


george-lorch:
https://ps.cd.percona.com/job/percona-server-5.7-param/60/

mysql-test/suite/tokudb/r/i_s_tokudb_lock_waits_released.result
```diff
@@ -68,9 +70,12 @@ TRX_ID	MYSQL_ID
 TRX_ID	MYSQL_ID
 commit;
 verify that the lock on the 1st transaction is released and replaced by the lock for the 2nd transaction
+Timeout in wait_condition.inc for select count(*)=1 from information_schema.tokudb_locks where locks_dname='./test/t-main'
```
laurynas-biveinis: ?

storage/tokudb/ha_tokudb.cc
```diff
@@ -3836,41 +3781,16 @@ void ha_tokudb::test_row_packing(uchar* record, DBT* pk_key, DBT* pk_val) {
...
-    // is the main table (or all indexes are a subset of the pk), 
-    // then we can simply insert the element
-    // with DB_YESOVERWRITE. If the element does not exist,
-    // it will act as a normal insert, and if it does exist, it 
-    // will act as a replace, which is exactly what REPLACE INTO is supposed
-    // to do. We cannot do this if otherwise, because then we lose
-    // consistency between indexes
-    //
+
     if (hidden_primary_key) {
```
laurynas-biveinis: Merge if (cond) { statement; } else if (another cond) { the same statement; } else if ... to if (cond || another cond) { the same statement} else if ...

laurynas-biveinis (changes requested):

mysql-test/suite/tokudb/r/i_s_tokudb_lock_waits_released.result
george-lorch: This is actually a very good catch that I missed in my self-review.  This patch impacts locking and what locks are taken/needed for sure, but, in diagnosing this I discovered that I reverted too much and removed a needed sub/child FT transaction that needs to be used on a REPLACE INTO/IGNORE.  Pushed fixed code and test case with a new jenkins here https://ps.cd.percona.com/job/percona-server-5.7-param/62/

laurynas-biveinis (approved):

PR #2542, opened 2018-09-07
PS-4439: Port clustering secondary index patch to 8.0

satya-bodapati:
Summary:
--------
Support CLUSTERING KEY type of TokuDB with new Data Dictionary

A new options property "clustering_key" is used to determine the
CLUSTERING KEY type. This is stored as part of dd::Index::options

KEY* flag HA_CLUSTERING is set based on clustering key property of
dd::Index::options (fill_index_from_dd())

Testing which are failing due to collation changes, "row" count
mismatch of explain output (wrong column masked) are rerecorded.


sql/dd_table_share.cc
```diff
@@ -1187,6 +1187,16 @@ static bool fill_index_from_dd(THD *thd, TABLE_SHARE *share,
       break;
   }
 
+  /* Check if this index is of clustering key type. Used by TokuDB */
+  const dd::Properties *index_options = &idx_obj->options();
```
laurynas-biveinis: nit: why not a reference instead of pointer?

laurynas-biveinis (approved):

sql/dd_table_share.cc
satya-bodapati: can be written that way too. Updated the patch

mysql-test/t/tokudb_clustering_key_grammar.test
```diff
@@ -51,7 +51,8 @@ ALTER TABLE t1 ADD CONSTRAINT c1 UNIQUE CLUSTERING INDEX b (b);
 ALTER TABLE t1 ADD CONSTRAINT c1 CLUSTERING UNIQUE INDEX b (b);
 --error ER_PARSE_ERROR
 ALTER TABLE t1 ADD CONSTRAINT c1 UNIQUE UNIQUE INDEX b (b);
---error ER_PARSE_ERROR
+#--error ER_PARSE_ERROR
```
george-lorch: Just remove, no need to keep commented error expectation

george-lorch (approved):
Loos good, all test result changes are perfectly in line with what is known to have changed in 8.0.

PR #2543, opened 2018-09-10
PS-269: Fix innodb.percona_force_encryption and innodb.general_ts_encrypt (8.0)

inikep:


laurynas-biveinis (approved):

sql/dd/dd_table.cc
```diff
@@ -2190,6 +2190,7 @@ static std::unique_ptr<dd::Table> create_dd_system_table(
                           ? dd::Abstract_table::HT_HIDDEN_SYSTEM
                           : dd::Abstract_table::HT_VISIBLE);
 
+  file->adjust_create_info_for_dd(create_info);
```
satya-bodapati: I think, with global DD, this is not the right approach. dd::Table can be adjusted by InnoDB at create_table API.

And remove this API.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -11495,15 +11495,28 @@ const char *create_table_info_t::create_options_are_invalid() {
...
+@param[in,out]  table_def dd::Table object to be modified.*/
+void ha_innobase::adjust_encryption_options(HA_CREATE_INFO *create_info,
+                                            dd::Table *table_def) noexcept {
   const bool is_intrinsic =
       (create_info->options & HA_LEX_CREATE_INTERNAL_TMP_TABLE) != 0;
   if (create_info->encrypt_type.length == 0 &&
       create_info->encrypt_type.str == nullptr &&
       (srv_encrypt_tables != SRV_ENCRYPT_TABLES_OFF ||
-       (srv_tmp_space.is_encrypted() && is_intrinsic)))
+       (srv_tmp_space.is_encrypted() && is_intrinsic))) {
```
satya-bodapati: 5.7, I see a bit different implementation. I would prefer to bring the exact implementation to 8.0 but..
if you think it would create merge mess (and makes merging of mising 5.7 commits), then it is OK to leave it as is

satya-bodapati (changes requested):

satya-bodapati (approved):
Looks good!

PR #2546, opened 2018-09-11
Fix PS-4723 (PURGE CHANGED_PAGE_BITMAPS doesn't work when innodb_data_home_dir used) (5.6)

percona-ysorokin:


percona-ysorokin:
Jenkins build link
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/22/

mysql-test/suite/innodb/t/percona_changed_page_bmp_requests_debug.test
```diff
@@ -25,3 +26,133 @@ SET @@GLOBAL.innodb_log_checkpoint_now=TRUE;
...
+--echo
+--echo # Restarting with redo log tracking disabled.
+--let $restart_parameters = restart:--innodb-track-changed-pages=0 --innodb-max-bitmap-file-size=4096 --innodb-data-home-dir=$CUSTOM_INNODB_DATA_HOME_DIR --innodb-log-group-home-dir=$CUSTOM_INNODB_DATA_HOME_DIR --innodb-file-per-table=0
+--source include/restart_mysqld.inc
+
+--let $assert_text = InnoDB redo log tracking must be disabled
+--let $assert_cond = @@innodb_track_changed_pages = 0
+--source include/assert.inc
+
+--let $assert_text = InnoDB file per table must be disabled
```
laurynas-biveinis: Why does file per table matter with tracking disabled?

percona-ysorokin: I added `--innodb-file-per-table=0` restart parameter to 5.6 and 5.7 (both to the first restart with tracking enabled and to the second one with tracking disabled) just to have the same settings as in 5.5 (in which it is OFF by default).

My reasoning for that is that `$select_lsn_query`, which determines the LSN of the last `.xdb` file, relies on the `space_id` being `0`. And I thought that with `--innodb-file-per-table=1` this would no longer be true.

laurynas-biveinis: I understand why it's needed when tracking is enabled, but why it's needed for this one restart where tracking is disabled?

percona-ysorokin: In the previous restart we had
```
--innodb-track-changed-pages=1
--innodb-max-bitmap-file-size=4096
--innodb-data-home-dir=$CUSTOM_INNODB_DATA_HOME_DIR
--innodb-log-group-home-dir=$CUSTOM_INNODB_DATA_HOME_DIR
--innodb-file-per-table=0
```
I just wanted to keep the changes in restart parameters minimal and leave everything else as it was in the previous restart - so changed only `--innodb-track-changed-pages=1` to `--innodb-track-changed-pages=0` here.

If you think this is misleading, I can change it.

laurynas-biveinis: Assert for it might be misleading, I'd just remove the option from this restart and the assert.

percona-ysorokin:
Removed `--innodb-file-per-table=0` option along with assertion from the second (non-tracking) restart.
Jenkins build link
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/24/ (single test run `innodb.percona_changed_page_bmp_requests_debug --repeat=2`)

laurynas-biveinis (approved):

PR #2550, opened 2018-09-12
Release 5.7.23 23

EvgeniyPatlan:


mysql-test/r/percona_log_slow_innodb.result
```diff
@@ -0,0 +1,50 @@
+DROP TABLE IF EXISTS t1, t2;
```
laurynas-biveinis, line 1: Please remove this file, it has been moved to innodb suite, should not be re-added here

mysql-test/t/percona_log_slow_innodb.test
```diff
@@ -0,0 +1,94 @@
+#
```
laurynas-biveinis, line 1: Likewise

mysql-test/r/percona_log_slow_innodb.result
laurynas-biveinis: @EvgeniyPatlan ping

PR #2552, opened 2018-09-14
Fix PS-3906 (handle_fatal_signal (sig=11) in dict_index_is_auto_gen_clust | sql/signal_handler.cc:223)

percona-ysorokin:
https://jira.percona.com/browse/PS-3906

Problem:
Inside 'create_table_info_t::create_table_update_dict()' which is called from
"CREATE TABLE" / "ALTER TABLE" handlers we invoke
'innobase_parse_hint_from_comment()', which is supposed to parse the hint for
the table and its indexes and update the information in dictionary, after the
table is closed by 'dict_table_close()'.
This can lead to a race condition in which the first thread calls
'dict_table_close()' and therefore the table is no longer locked.
Meanwhile the second thread ('srv_master_thread') triggers
'srv_master_evict_from_table_cache()' for this table and frees its resources.
After that, the first thread (the one which executes "CREATE TABLE" /
"ALTER TABLE") is activated again and calls
'innobase_parse_hint_from_comment()' which in turn tries to access freed
memory.

Fix:
Making sure that 'innobase_parse_hint_from_comment()' is called before
'dict_table_close()'.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -13100,9 +13100,9 @@ create_table_info_t::create_table_update_dict()
 		dict_table_autoinc_unlock(innobase_table);
 	}
 
-	dict_table_close(innobase_table, FALSE, FALSE);
-
 	innobase_parse_hint_from_comment(m_thd, innobase_table, m_form->s);
+
+	dict_table_close(innobase_table, FALSE, FALSE);
 
 	DBUG_RETURN(0);
 }


percona-ysorokin:
Jenkins build link
https://ps.cd.percona.com/job/percona-server-5.7-param/69/

laurynas-biveinis (commented):
Are lower versions affected? Possible to add a testcase?

percona-ysorokin:
@laurynas-biveinis 
This `innobase_parse_hint_from_comment()` function was introduced in
WL#6747: InnoDB: make fill factor settable
(commit 3f4abb52fbfe9118fc94a4b60876c16e5aa2e0e0)
only starting from MySQL Server 5.7.6
to support `MERGE_THRESHOLD=NN` comment extension
(https://dev.mysql.com/doc/refman/5.6/en/index-page-merge-threshold.html).
So, 5.5 and 5.6 branches are not affected by this bug.

As for the test case.
Originally, I thought that to simulate this problem we can add a DEBUG_SYNC point between
```
dict_table_close(innobase_table, FALSE, FALSE);
```
and
```
innobase_parse_hint_from_comment(m_thd, innobase_table, m_form->s);
```
And make other thread evict the table from the table cache. Then, switching back to the first thread will cause server crash. This will definitely help to reproduce the problem before code changes.

However, after the fix, when we have these two statements in the right order
```
innobase_parse_hint_from_comment(m_thd, innobase_table, m_form->s);
dict_table_close(innobase_table, FALSE, FALSE);
```
DEBUG_SYNC will just be useless here.
For instance, if we set a synchronization point between these two statements, the second thread won't be able to evict the table from the cache as it is still open and we will get a deadlock / infinite loop.

Making the order of these two statements conditional depending on some DBUG_EXECUTE_IF also has the same problems - we will be able to test the crash, but won't be able to test the deadlock.

So, the best "testing" I've got so far is just properly describing the problem and the fix in the commit message.

laurynas-biveinis (approved):

PR #2554, opened 2018-09-17
PS-4845 : MyRocks tests leave files behind in $MYSQL_TMP_DIR

george-lorch:
- Fixed up tests to use $MYSQLTEST_VARDIR/tmp and cleanup after themselves.

### DIFF
--- a/mysql-test/suite/rocksdb/include/bloomfilter_load_select.inc
+++ b/mysql-test/suite/rocksdb/include/bloomfilter_load_select.inc
@@ -1,9 +1,11 @@
 # loading some data (larger than write buf size) to cause compaction
---exec perl suite/rocksdb/t/gen_insert.pl t1 >  $MYSQL_TMP_DIR/insert_t1.sql
---exec perl suite/rocksdb/t/gen_insert.pl t2 >  $MYSQL_TMP_DIR/insert_t2.sql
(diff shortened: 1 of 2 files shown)


george-lorch:
https://ps.cd.percona.com/job/percona-server-5.7-param/70/

mysql-test/suite/rocksdb/include/bloomfilter_load_select.inc
```diff
@@ -1,9 +1,11 @@
 # loading some data (larger than write buf size) to cause compaction
---exec perl suite/rocksdb/t/gen_insert.pl t1 >  $MYSQL_TMP_DIR/insert_t1.sql
```
laurynas-biveinis, line 2: Why cannot leave $MYSQL_TMP_DIR?

george-lorch: I suppose we could.  I thought the standard though for temp files was to use VARDIR/tmp which is partially why these were never detected.

laurynas-biveinis: I see, no problem in doing this then

laurynas-biveinis (approved):
please upstream

PR #2562, opened 2018-09-20
PS-4834 - encrypted system tablespace has empty uuid

satya-bodapati:
Problem:
    --------
    This is happening because when master key is created at innodb
    initialization time, it gets server uuid. But server generates
    server_uuid() only after innodb intialization is done.
    So an empty uuid is given to InnoDB by server.
    
    Since master key is created with empty uuid and all tables will
    continue this.
    
    Whats the fix then?:
    
    Can't we not make server generate UUID before innodb is intialized.
    Well, not that simple. the UUID generation code needs THD, which
    inturn needs all plugins to be initialized. So it is kind of
    chicken and egg problem. (InnoDB needs server_uuid() to be ready
    and server_uuid() needs InnoDB to ready)
    
    Ideally server uuid generation shouldn't have any dependencies with InnoDB
    (the best fix but hard)
    
    How upstream avoids empty UUID with tablespaces like redo(in 8.0):
    
    1. Like system, redo is also created with empty uuid, default masterkey
    2. Background threads keep checking for the default master key id and then
       tries to rotate the encryption info of a tablespace
    
    Fix:
    ----
    Unlike upstream, we will use a new handlerton method. This is a better
    way because
    
    1. It doesn't rely on background threads to fix empty UUID
    2. Allows redo and undo encryption to be enabled from foreground threads.
        As part of variable update function
    3. Fixes empty UUID for tablespaces not only at bootstrap but also for
        tablespaces that are encrypted at startup.
    4. Fixes empty UUID not only innodb global tablespaces(system, temp etc),
        but also for all mysql.* ibds. They can be chosen to encrypted if
        --innodb-encrypt-tables is ON during bootstrap
    
    Added debug asserts that crash when empty UUID is written to
    tablespace header and also crash if empty UUID is read from tablespace
    header.

### DIFF
--- a/sql/handler.h
+++ b/sql/handler.h
@@ -1061,6 +1061,17 @@ struct handlerton
   */
   bool (*rotate_encryption_master_key)(void);
 
+ /**
+    @brief
+    Fix empty UUID of tablespaces of an engine. This is used when engine encrypts
+    tablespaces as part of initialization. These tablespaces will have empty UUID
+    because UUID is generated after all plugins are initialized. This API will be
+    called by server only after UUID is available.
+    @returns false on success,
+             true on failure
+  */
+  bool (*fix_tablespaces_empty_uuid)(void);
+
   /**
     Creates a new compression dictionary with the specified data for this SE.
 
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -5144,6 +5144,20 @@ int mysqld_main(int argc, char **argv)
     unireg_abort(MYSQLD_ABORT_EXIT);
   }
 
+  /* Server generates uuid after innodb is initialized. But during
+  initialization, if tablespaces like system, redo, temporary are encrypted,
+  they are initialized with "empty" UUID. Now UUID is available, fix the
+  empty UUID of such tablespaces now */
+  if (innodb_hton != NULL && innodb_hton->fix_tablespaces_empty_uuid)
+  {
+    if (innodb_hton->fix_tablespaces_empty_uuid())
+    {
+      sql_print_error("Fixing empty UUID with InnoDB Engine failed. Please"
+		      " check if keyring plugin is loaded and execute"
+		      " \"ALTER INSTANCE ROTATE INNODB MASTER KEY\"");
+    }
+  }
+
   /*
     Add server_uuid to the sid_map.  This must be done after
     server_uuid has been initialized in init_server_auto_options and
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -7469,14 +7469,45 @@ fil_temp_update_encryption(
 /** Default master key id for bootstrap */
 static const ulint ENCRYPTION_DEFAULT_MASTER_KEY_ID = 0;
 
+/** Rotate the tablespace key by new master key.
+@param[in]	space	tablespace object
+@return true if the re-encrypt suceeds */
+static
+bool
+fil_encryption_rotate_low(const fil_space_t* space)
+{
+	bool success = true;
+	if (space->encryption_type != Encryption::NONE) {
+		mtr_t mtr;
+		mtr_start(&mtr);
+
+		if (fsp_is_system_temporary(space->id)) {
+			mtr_set_log_mode(&mtr, MTR_LOG_NO_REDO);
+		}
+
+		mtr.set_named_space(space->id);
+
+		fil_space_t* space_locked = mtr_x_lock_space(space->id, &mtr);
+
+		byte	encrypt_info[ENCRYPTION_INFO_SIZE_V2];
+		memset(encrypt_info, 0, ENCRYPTION_INFO_SIZE_V2);
+
+		if (!fsp_header_rotate_encryption(space_locked,
+						  encrypt_info,
+						  &mtr)) {
+			success = false;
+		}
+		mtr_commit(&mtr);
+	}
+	return(success);
+}
+
 /** Rotate the tablespace keys by new master key.
 @return true if the re-encrypt suceeds */
 bool
 fil_encryption_rotate()
 {
 	fil_space_t*	space;
-	mtr_t		mtr;
-	byte		encrypt_info[ENCRYPTION_INFO_SIZE_V2];
 
 	for (space = UT_LIST_GET_FIRST(fil_system->space_list);
 	     space != NULL; ) {
@@ -7498,27 +7529,9 @@ fil_encryption_rotate()
 			continue;
 		}
 
-		if (space->encryption_type != Encryption::NONE) {
-			mtr_start(&mtr);
-
-			if (fsp_is_system_temporary(space->id)) {
-				mtr_set_log_mode(&mtr, MTR_LOG_NO_REDO);
-			}
-
-			mtr.set_named_space(space->id);
-
-			space = mtr_x_lock_space(space->id, &mtr);
-
-			memset(encrypt_info, 0, ENCRYPTION_INFO_SIZE_V2);
-
-			if (!fsp_header_rotate_encryption(space,
-							  encrypt_info,
-							  &mtr)) {
-				mtr_commit(&mtr);
-				return(false);
-			}
-
-			mtr_commit(&mtr);
+		bool success = fil_encryption_rotate_low(space);
+		if (!success) {
+			return(false);
 		}
 
 		space = UT_LIST_GET_NEXT(space_list, space);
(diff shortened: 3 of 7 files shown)


laurynas-biveinis (approved):

satya-bodapati:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1920/ (in-progress)

satya-bodapati:
Updated the patch with new approach because when reusing existing handlerton API, it always assumes Keyring is ready and throws error for normal initialize/intialize-insecure

New approach:
1. use new handlerton API (rotate only system, temp from this API)
2. fixes empty UUID for tablespaces not only at bootstrap but for also tablespaces encrypted at startup (temporary, redo, undo)
3. added debug assert to catch empty UUID at write time, previously it was only at read.
4. Paves way for enabling encryption of redo and undo from foreground threads. (future - 8.0)

satya-bodapati:
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1922/

sql/handler.h
```diff
@@ -1061,6 +1061,18 @@ struct handlerton
   */
   bool (*rotate_encryption_master_key)(void);
 
+ /**
+    @brief
+    Initiate key rotation of specific tablespaces of an engine
```
laurynas-biveinis: IMHO not perfectly clear - what specific tablespaces? I'd try calling the handlerton fixup_tablespace_keys or similar - to show that the rotation happens from "incorrect" to "correct"

sql/handler.h
```diff
@@ -1061,6 +1061,18 @@ struct handlerton
...
+    This is used when engine encrypts tablespace as part of initialization.
+    These tablespace will have empty UUID because UUID is generated after
+    all plugins are initialized.  This API will be called by server only after
+    UUID is availabe.
```
laurynas-biveinis: "available"

sql/mysqld.cc
```diff
@@ -5144,6 +5144,22 @@ int mysqld_main(int argc, char **argv)
     unireg_abort(MYSQLD_ABORT_EXIT);
   }
 
+  /* Server generates uuid after innodb is initialized. But during initialize,
```
laurynas-biveinis: s/initialize/initialization ?

sql/mysqld.cc
```diff
@@ -5144,6 +5144,22 @@ int mysqld_main(int argc, char **argv)
...
+  if tablespace like system, redo, temporary are encrypted, they are initialized
+  with "empty" uuid. Do the key rotation now because a proper uuid exists
+  At this point of time, only the above tablespace exist and available for rotation */
```
laurynas-biveinis: what is "the above tablespace" here?

sql/mysqld.cc
```diff
@@ -5144,6 +5144,22 @@ int mysqld_main(int argc, char **argv)
...
+  if (innodb_hton != NULL)
```
laurynas-biveinis: I'd merge the two outermost ifs

storage/innobase/fil/fil0fil.cc
```diff
@@ -7469,14 +7469,45 @@ fil_temp_update_encryption(
 /** Default master key id for bootstrap */
 static const ulint ENCRYPTION_DEFAULT_MASTER_KEY_ID = 0;
 
+/** Rotate the tablespace key by new master key.
+@param[in]	space	tablespace object
+@return true if the re-encrypt suceeds */
+static
+bool
+fil_encryption_rotate_low(const fil_space_t* space)
```
laurynas-biveinis, line 7477: Pass space by const reference?

storage/innobase/fil/fil0fil.cc
```diff
@@ -7498,34 +7529,39 @@ fil_encryption_rotate()
...
+		bool success = fil_encryption_rotate_low(space);
+		if (!success) {
+			return(false);
+		}
 
-			if (fsp_is_system_temporary(space->id)) {
-				mtr_set_log_mode(&mtr, MTR_LOG_NO_REDO);
-			}
+		space = UT_LIST_GET_NEXT(space_list, space);
+		DBUG_EXECUTE_IF("ib_crash_during_rotation_for_encryption",
```
laurynas-biveinis, line 7538: No testcase is added using this - is this an omission (then please add it) or is it a backport of Oracle patch with presumably non-public testcase - in which case don't add this code?

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -53,6 +53,8 @@ Created 11/29/1995 Heikki Tuuri
 
 #include <my_aes.h>
 
+extern bool innodb_inited;
```
laurynas-biveinis: Maybe ha_prototypes.h ?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -127,7 +127,7 @@ static mysql_mutex_t innobase_share_mutex;
 static ulong commit_threads = 0;
 static mysql_cond_t commit_cond;
 static mysql_mutex_t commit_cond_m;
-static bool innodb_inited = 0;
+bool innodb_inited = 0;
```
laurynas-biveinis: false

laurynas-biveinis (changes requested):

sql/handler.h
satya-bodapati: changed API name to fix_tablespaces_empty_uuid()

sql/handler.h
satya-bodapati: Fixed

sql/mysqld.cc
satya-bodapati: Fixed

sql/mysqld.cc
satya-bodapati: Rewrote the comment

sql/mysqld.cc
satya-bodapati: Done

storage/innobase/fil/fil0fil.cc
satya-bodapati: I assume  is nit-pick :)  InnoDB code in general prefers const*. Left this one as is

storage/innobase/fil/fil0fil.cc
satya-bodapati: I didn't add this. The only change here is refactoring into fil_encryption_rotate_low() and yes this looks like a dbug crash point used by Oracle's non-public testcase

storage/innobase/fsp/fsp0fsp.cc
satya-bodapati: Done

storage/innobase/handler/ha_innodb.cc
satya-bodapati: Done

satya-bodapati:
One more change from last patch is the failure of the handlerton API is not hard error. We print error message asking user to load keyring plugin and do ALTER INSTANCE ROTATE INNODB MASTER KEY.

This looks more reasonable.

storage/innobase/fil/fil0fil.cc
laurynas-biveinis: Yes, a nit pick, but it is a C legacy and InnoDB is incorrect here to carry over it to C++. I believe their new code uses const references too as it should, but not uniformly. OK to leave as is

storage/innobase/fil/fil0fil.cc
laurynas-biveinis: So it needs removing then?

laurynas-biveinis (commented):
Minor comments

storage/innobase/fil/fil0fil.cc
satya-bodapati: No, previously (before my change) it is ..

```
fil_encryption_rotate() {
  for () {
     ..
    if(space->encryption != Encryption::None) {
      ..
     }
     ..
  }
  DBUG_EXECUTE_IF("ib_crash_during_rotation_for_encryption",
                                  DBUG_SUICIDE(););
}

Now:
fil_encryption_rotate() {
  for () {
     ..
    fil_encryption_rotate_low(space);
     ..
  }
  DBUG_EXECUTE_IF("ib_crash_during_rotation_for_encryption",
                                  DBUG_SUICIDE(););
}

```
I dont see the need to remove this crash point. Only the code above was refactored into separate function.

laurynas-biveinis: Indeed, sorry, OK to leave as is

laurynas-biveinis (approved):

PR #2570, opened 2018-09-25
Fixed PS-1107 (LP #1703346: Binlog corruption when tmpdir gets full) (5.7)

percona-ysorokin:
https://jira.percona.com/browse/PS-1107

Cherry-picked from ps-5.6-bugps1107-tmpdir_full branch
(commit c7b6226).

Fixed problem with wrapping 'reinit_io_cache()' inside
'MYSQL_BIN_LOG::do_write_cache()' with 'simulate_file_write_error' DBUG
directive. 'DBUG_SET("-d,simulate_file_write_error")' should be called
not only when 'reinit_io_cache()' fails but when it succeeds as welll.
Whithout this patch, any subsequent call to 'my_write()' ends up with
an infinite loop.

Reorganized 'rpl.rpl_bug72457' MTR test case. Instead, we now have two
'rpl.rpl_bug72457_myisam' and 'rpl.rpl_bug72457_innodb' which test the
same (original) scenario with different values of
'--internal-tmp-disk-storage-engine' and '--default-tmp-storage-engine'
variables set to 'MyISAM' and 'InnoDB' correspondingly.

### DIFF
--- a/mysql-test/suite/rpl/include/rpl_bug72457.inc
+++ b/mysql-test/suite/rpl/include/rpl_bug72457.inc
@@ -1,0 +1,41 @@
+--source include/master-slave.inc
+
+call mtr.add_suppression("Slave SQL for channel '': The incident LOST_EVENTS occured on the master\\. Message: The content of the statement cache is corrupted while writing a rollback record of the transaction to the binary log\\. Error_code:");
+call mtr.add_suppression("The content of the statement cache is corrupted while writing a rollback record of the transaction to the binary log\\. An incident event has been written to the binary log which will stop the slaves\\.");
+
+SET @saved_binlog_error_action = @@global.binlog_error_action;
+SET GLOBAL binlog_error_action = IGNORE_ERROR;
+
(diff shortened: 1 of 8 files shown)


percona-ysorokin:
Not following GCA here as proper 'out of disk space' simulation requires 'simulate_disk_full_at_flush_pending' DBUG directive from

Bug #27399620: BINLOG AND ENGINE BECOME INCONSISTENT WHEN BINLOG CACHE FILE GETS OUT OF SPACE
(commit 9395f0a)
which is available in MySQL 5.7.23 only.

Jenkins build link
https://ps.cd.percona.com/job/percona-server-5.7-param/72/

laurynas-biveinis (changes requested):
Same comments as for the 5.5 and 5.6
Please also prepare an 8.0 version if needed
Please also upstream

percona-ysorokin:
Updated Jenkins build link
https://ps.cd.percona.com/job/percona-server-5.7-param/73/

laurynas-biveinis (approved):

PR #2571, opened 2018-09-26
PS-4834 - encrypted system tablespace has empty uuid

satya-bodapati:
Problem:
-------

This is happening because when master key is created at innodb
initialization time, it gets server uuid. But server generates
server_uuid() only after innodb intialization is done.
So an empty uuid is given to InnoDB by server.

Since master key is created with empty uuid and all tables will
continue this.

Whats the fix then?:

Can't we not make server generate UUID before innodb is intialized.
Well, not that simple. the UUID generation code needs THD, which
inturn needs all plugins to be initialized. So it is kind of
chicken and egg problem. (InnoDB needs server_uuid() to be ready
and server_uuid() needs InnoDB to ready)

Ideally server uuid generation shouldn't have any dependencies with InnoDB
(the best fix but hard)

How upstream avoids empty UUID with tablespaces like redo(in 8.0):

1. Like system, redo is also created with empty uuid, default masterkey
2. Background threads keep checking for the default master key id and then
   tries to rotate the encryption info of a tablespace

Fix:
----
Unlike upstream, we will use a new handlerton method. This is a better
way because

1. It doesn't rely on background threads to fix empty UUID
2. Allows redo and undo encryption to be enabled from foreground threads.
   As part of variable update function
3. Fixes empty UUID for tablespaces not only at bootstrap but also for
   tablespaces that are encrypted at startup.
4. Fixes empty UUID not only innodb global tablespaces(system, temp etc),
   but also for all mysql.* ibds. They can be chosen to encrypted if
   --innodb-encrypt-tables is ON during bootstrap

Added debug asserts that crash when empty UUID is written to
tablespace header and also crash if empty UUID is read from tablespace
header.
(cherry picked from commit 30330d0f831f65d3918486d492f388aaeeb63598)

### DIFF
--- a/mysql-test/suite/innodb/t/percona_sys_tablespace_encrypt.test
+++ b/mysql-test/suite/innodb/t/percona_sys_tablespace_encrypt.test
@@ -226,7 +226,5 @@ ALTER TABLE t9 TABLESPACE=`innodb_system`;
 --source include/start_mysqld.inc
 
 --remove_file $BOOTSTRAP_SQL
---let CLEANUP_FOLDER=$MYSQL_TMP_DIR/datadir1
---source include/cleanup_folder.inc
---remove_file $MYSQL_TMP_DIR/mysecret_keyring
+--force-rmdir $MYSQL_TMP_DIR/datadir1
 --remove_file $MYSQLTEST_VARDIR/tmp/boot.log
--- a/sql/handler.h
+++ b/sql/handler.h
@@ -1737,6 +1737,17 @@ typedef bool (*rotate_encryption_master_key_t)(void);
 
 /**
   @brief
+  Fix empty UUID of tablespaces of an engine. This is used when engine encrypts
+  tablespaces as part of initialization. These tablespaces will have empty UUID
+  because UUID is generated after all plugins are initialized. This API will be
+  called by server only after UUID is available.
+  @returns false on success,
+           true on failure
+*/
+using fix_tablespaces_empty_uuid_t = bool (*)(void);
+
+/**
+  @brief
   Retrieve ha_statistics from SE.
 
   @param db_name                  Name of schema
@@ -2015,6 +2026,7 @@ struct handlerton {
   notify_exclusive_mdl_t notify_exclusive_mdl;
   notify_alter_table_t notify_alter_table;
   rotate_encryption_master_key_t rotate_encryption_master_key;
+  fix_tablespaces_empty_uuid_t fix_tablespaces_empty_uuid;
 
   get_table_statistics_t get_table_statistics;
   get_index_column_cardinality_t get_index_column_cardinality;
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -6194,6 +6194,20 @@ int mysqld_main(int argc, char **argv)
   if (!server_id_supplied)
     LogErr(INFORMATION_LEVEL, ER_WARN_NO_SERVERID_SPECIFIED);
 
+  /* Server generates uuid after innodb is initialized. But during
+  initialization, if tablespaces like system, redo, temporary are encrypted,
+  they are initialized with "empty" UUID. Now UUID is available, fix the
+  empty UUID of such tablespaces now */
+  if (innodb_hton != nullptr &&
+      innodb_hton->fix_tablespaces_empty_uuid != nullptr) {
+    if (innodb_hton->fix_tablespaces_empty_uuid()) {
+      sql_print_error(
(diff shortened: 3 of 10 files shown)


mysql-test/suite/innodb/t/percona_sys_tablespace_encrypt.test
```diff
@@ -226,7 +226,6 @@ ALTER TABLE t9 TABLESPACE=`innodb_system`;
 --source include/start_mysqld.inc
 
 --remove_file $BOOTSTRAP_SQL
---let CLEANUP_FOLDER=$MYSQL_TMP_DIR/datadir1
---source include/cleanup_folder.inc
---remove_file $MYSQL_TMP_DIR/mysecret_keyring
+--force-rmdir $MYSQL_TMP_DIR/datadir1
+#--remove_file $MYSQL_TMP_DIR/mysecret_keyring
```
laurynas-biveinis: Remove commented out code - but is it indeed correct, does the testcase pass the testcase check?

sql/handler.h
```diff
@@ -1735,6 +1735,17 @@ typedef bool (*notify_alter_table_t)(THD *thd, const MDL_key *mdl_key,
...
+/**
+  @brief
+  Fix empty UUID of tablespaces of an engine. This is used when engine encrypts
+  tablespaces as part of initialization. These tablespaces will have empty UUID
+  because UUID is generated after all plugins are initialized. This API will be
+  called by server only after UUID is available.
+  @returns false on success,
+           true on failure
+*/
+typedef bool (*fix_tablespaces_empty_uuid_t)(void);
```
laurynas-biveinis: nit: C++11: using fix_tablespaces_empty_uuid_t = bool (*)(void);

sql/mysqld.cc
```diff
@@ -6194,6 +6194,19 @@ int mysqld_main(int argc, char **argv)
   if (!server_id_supplied)
     LogErr(INFORMATION_LEVEL, ER_WARN_NO_SERVERID_SPECIFIED);
 
+  /* Server generates uuid after innodb is initialized. But during
+  initialization, if tablespaces like system, redo, temporary are encrypted,
+  they are initialized with "empty" UUID. Now UUID is available, fix the
+  empty UUID of such tablespaces now */
+  if (innodb_hton != NULL && innodb_hton->fix_tablespaces_empty_uuid) {
```
laurynas-biveinis: s/NULL/nullptr. Also either both predicates should be != nullptr either if (innodb_hton && innodb_hton->fix_tablespaces_empty_uuid)

storage/innobase/fil/fil0fil.cc
```diff
@@ -8416,13 +8416,37 @@ dberr_t fil_temp_update_encryption(fil_space_t *space) {
   return (err);
 }
 
+/** Rotate the tablespace key by new master key.
+@param[in]	space	tablespace object
+@return true if the re-encrypt suceeds */
+static bool encryption_rotate_low(fil_space_t *space) {
```
laurynas-biveinis, line 8422: nit: const ref, but OK to leave as is

storage/innobase/handler/ha_innodb.cc
```diff
@@ -3813,6 +3813,76 @@ bool innobase_encryption_key_rotation() {
...
+  }
+
+  /* We only need to handle the case when an encrypted tablespace
+  is created at startup. If it is 0, there is no encrypted tablespace,
+  If it is > 1, it means we already have fixed the UUID */
+  if (Encryption::s_master_key_id != 1) {
+    return (false);
+  }
+
+  byte *master_key = NULL;
```
laurynas-biveinis: s/NULL/nullptr/g

storage/innobase/include/fil0fil.h
```diff
@@ -1771,4 +1771,11 @@ void fil_space_update_name(fil_space_t *space, const char *name);
 @param space_id	space id */
 void fil_space_set_corrupt(space_id_t space_id);
 
+using space_id_vec = std::vector<ulint>;
```
laurynas-biveinis: s/ulint/space_id_t

laurynas-biveinis (changes requested):

mysql-test/suite/innodb/t/percona_sys_tablespace_encrypt.test
satya-bodapati: I think it is bug in MTRs orphan file check.  Will remove the commented out line

satya-bodapati:
Patch is ready for review

laurynas-biveinis (approved):

PR #2576, opened 2018-09-28
Fix after release 5.7.23

EvgeniyPatlan:


mysql-test/t/percona_log_slow_innodb.test
```diff
@@ -1,4 +1,3 @@
-#
```
laurynas-biveinis, line 1: Please remove the whole file not just the first line

laurynas-biveinis (changes requested):
Please refer to which release in the commit message

mysql-test/t/percona_log_slow_innodb.test
EvgeniyPatlan: @laurynas-biveinis done

EvgeniyPatlan:
commit message has been updated

laurynas-biveinis (approved):

PR #2581, opened 2018-10-01
PS-4885 : [PS8QA] Using ALTER ... ROW_FORMAT=TOKUDB_QUICKLZ leads to …

george-lorch:
…InnoDB: Assertion failure: ha_innodb.cc:12198:m_form->s->row_type == m_create_info->row_type

- Completed mapping of SQL parse row_format types to and from data dictionary
  types.  Re-recorded impacted tests.

### DIFF
--- a/mysql-test/suite/tokudb.alter_table/r/row_format_alter.result
+++ b/mysql-test/suite/tokudb.alter_table/r/row_format_alter.result
@@ -16,7 +16,7 @@ ALTER TABLE foo row_format=tokudb_lzma, add column c2 int not null;
 SELECT CREATE_OPTIONS
 FROM INFORMATION_SCHEMA.TABLES  WHERE (table_name LIKE 'foo');
 CREATE_OPTIONS
-row_format=COMPRESSED
+row_format=TOKUDB_LZMA
 SHOW CREATE TABLE foo;
 Table	Create Table
 foo	CREATE TABLE `foo` (
@@ -28,7 +28,7 @@ ALTER TABLE foo row_format=TOKUDB_QUICKLZ, add column c3 int not null;
 SELECT CREATE_OPTIONS
 FROM INFORMATION_SCHEMA.TABLES  WHERE (table_name LIKE 'foo');
 CREATE_OPTIONS
-row_format=COMPRESSED
+row_format=TOKUDB_QUICKLZ
 SHOW CREATE TABLE foo;
 Table	Create Table
 foo	CREATE TABLE `foo` (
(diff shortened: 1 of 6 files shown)


george-lorch:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/55/  jobs timed out, probably should have used stronger instances.  Everything expected passes locally.

laurynas-biveinis (approved):

satya-bodapati:
how is this stored on disk?  (@laurynas-biveinis @georgelorch ) 

sql/dd/impl/tables/tables.cc

Tables::Tables() {
  m_target_def.set_table_name("tables");

  m_target_def.add_field(FIELD_ID, "FIELD_ID",
                         "id BIGINT UNSIGNED NOT NULL AUTO_INCREMENT");
  m_target_def.add_field(FIELD_SCHEMA_ID, "FIELD_SCHEMA_ID",
 YSQL_VERSION_ID",
                         "mysql_version_id INT UNSIGNED NOT NULL");
  **m_target_def.add_field(FIELD_ROW_FORMAT, "FIELD_ROW_FORMAT",
                         "row_format ENUM('Fixed', 'Dynamic', 'Compressed',"
                         "                'Redundant','Compact','Paged')");**

How can the new types be stored on disk, without adding the types here? After create table, can you do restart and see if the 'row_format' still shows correct?

george-lorch:
That's a great question and one that makes me very happy to have you on-board.  I suspected this was going to be an upgrade problem, but did not consider this condition.  Realistically, this should not ever have been be a row format, but, in Tokutek attempts to appear mainstream, they resorted to changes like this.  IMHO, even InnoDB row formats should not be a part of the SQL grammar if it is not 100% extendable to any/all engines.

laurynas-biveinis:
It might have been exactly why I didn't do this originally and then forgot about it :( @georgelorchpercona , please revert and let's discuss offline how to go forward. We cannot change DD schema, hence we might have to re-do this as table/column comments ASAP

satya-bodapati:
@georgelorch I agree, it is not extensible. I faced the same issue when I wanted to add a new index type. Even that is ENUM on disk and doesn't allow to add new types (unless the on-disk definition is changed).

I am seeing some shortcomings of DD from flexibility point of view
1. Not flexible enough to add new types (index, row_type)
2. Cannot easily add a new DD table (too much of hardcoding) and compatibility
3. Cannot create a new DD table on startups (like InnoDB find_or_create. See SYS_VIRTUAL, SYS_ZIP_DICT, etc)
4. Cannot use DD interfaces (for tables created in mysql.*) (kind of workaround for 2, if 2 is possible, this is not necessary)

PR #2582, opened 2018-10-03
PS-269: Fix MTR tests from main (8.0)

inikep:
Fixed:
main.parser-big-64bit (Fix output when using UBSAN)
main.view (re-record)
main.with_recursive_wl9248 (increase max_heap_table_size, fix UBSan warning)

### DIFF
--- a/sql/table.h
+++ b/sql/table.h
@@ -912,7 +912,7 @@ struct TABLE_SHARE {
     without explicit pk and corresponding warning was issued to disable
     repeated warning.
   */
-  bool rfr_lookup_warning;
+  bool rfr_lookup_warning{false};
 
   /// For materialized derived tables; @see add_derived_key().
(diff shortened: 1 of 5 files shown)


sql/table.h
```diff
@@ -912,7 +912,7 @@ struct TABLE_SHARE {
     without explicit pk and corresponding warning was issued to disable
     repeated warning.
   */
-  bool rfr_lookup_warning;
+  bool rfr_lookup_warning{false};
```
laurynas-biveinis, line 915: Does this point to a bug in lower versions?

laurynas-biveinis (approved):

sql/table.h
inikep: There is no issue with 5.7 because of `memset` in the constructor:
``` 
TABLE_SHARE() { memset(this, 0, sizeof(*this)); }
```

PR #2583, opened 2018-10-03
Ps 8.0 4885

george-lorch:
Revert fix(es) for PS-4750 and PS-4885.


laurynas-biveinis (approved):

laurynas-biveinis:
Merging myself due to clang-format error introduced in the reverted commit

PR #2587, opened 2018-10-08
PS-269: Fix search_pattern_multiline.inc (8.0)

inikep:
This patch removes code starting with `elsif ( $found == 0 )` because
1. `$found` is never changed
2. Search errors are detected and throw in `foreach` when first not matched file is found.

It fixes files that include `search_pattern_multiline.inc`:
- auth_sec.plugin_auth_caching_sha2_password_default_qa
- innodb.innodb_page_size_func
- innodb.innodb-status-output


laurynas-biveinis (changes requested):
Please explain in the commit message why this is the right thing to do?

inikep:

laurynas-biveinis (approved):

PR #2589, opened 2018-10-09
si->thd use after free/delete segfault

keyurdg:
When a slave disconnects after sending COM_REGISTER_SLAVE without sending COM_BINLOG_DUMP(_GTID) (say due to being in a bad network environment), the master can crash if "show slave hosts" is executed. This is because si keeps a copy of the thd to print the slave's UUID, and this thd can be destructed/free'd outside of the update to the slave_info hashtable. Attached is a patch that moves the copying of si->thd into the BINLOG_DUMP methods, which will call the unregister method in case of any errors.

_The change to the mysqlbinlog client in the PR is only for test case purposes and can be ignored._


dutow:
Hello!

PR #2593, opened 2018-10-12
PS-4434 : Port TokuDB compression grammar to 8.0

george-lorch:
- It was decided that we will NOT port this to 8.0 due to the invasive and
  incompatible changes with the new 8.0 data dictionary.  This functionality is
  mostly syntactical sugar that drags engine speficific SQL grammar and keywords
  up into the SQL/server layer.  Specifying TokuDB compression format for a
  table and each index can be done by setting the tokudb_row_format session
  variable to the desired algorithm just prior to creating a table or index.

- This commit removes all traces of the custom row_format keywords,
  enumerations, conversions, and test case uses.


george-lorch:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/66/

mysql-test/suite/tokudb.bugs/r/alter_external_lock_assert.result
```diff
@@ -18,4 +16,6 @@ ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
 UNLOCK TABLES;
 COMMIT;
 DROP TABLE IF EXISTS i,p,t1_tokudb_tokudb_smallvc_250;
+Warnings:
```
laurynas-biveinis: Don't drop "p"

mysql-test/suite/tokudb.bugs/t/alter_external_lock_assert.test
```diff
@@ -20,7 +17,5 @@ ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
 ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
 UNLOCK TABLES;
 COMMIT;
-disable_warnings;
 DROP TABLE IF EXISTS i,p,t1_tokudb_tokudb_smallvc_250;
```
laurynas-biveinis: See the comment at the .result file

mysql-test/suite/tokudb.bugs/t/alter_external_lock_assert.test
```diff
@@ -20,7 +17,5 @@ ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
...
-enable_warnings;
 DROP TABLE t1_tokudb_tokudb_smallvc_250;
```
laurynas-biveinis, line 21: It's dropped right above?

mysql-test/suite/tokudb.bugs/t/alter_table_comment_rebuild_data.test
```diff
@@ -168,14 +168,6 @@ show create table t1;
 --let $assert_cond= "$ori_file" = "$new_file"
 --source include/assert.inc
 
-# Alter table with compression method
```
laurynas-biveinis, line 171: I assume rebuilding ALTER TABLE (as OPTIMIZE TABLE perhaps) with different compression methods set through system vars are tested elsewhere?

laurynas-biveinis (changes requested):
Good riddance

mysql-test/suite/tokudb.bugs/t/alter_table_comment_rebuild_data.test
george-lorch: Good thought, I don't believe there is an explicit test that an ALTER - COPY with a different tokudb_row_format checks the ft header for the new compression type.  I think this can easily be added to the compressions.test

mysql-test/suite/tokudb.bugs/r/alter_external_lock_assert.result
george-lorch: Ahh, actually, that seems to be the cause of the tests need for the duplicate DROP.

george-lorch:
Pushed minor changes and new jenkins on impacted tests https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/68/

mysql-test/suite/tokudb.bugs/t/alter_external_lock_assert.test
```diff
@@ -20,7 +17,5 @@ ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
 ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
 UNLOCK TABLES;
 COMMIT;
-disable_warnings;
-DROP TABLE IF EXISTS i,p,t1_tokudb_tokudb_smallvc_250;
-enable_warnings;
+DROP TABLE IF EXISTS i,t1_tokudb_tokudb_smallvc_250;
```
laurynas-biveinis: I still don't understand why there is IF EXISTS here and the 2nd DROP TABLE statement below

laurynas-biveinis (changes requested):

mysql-test/suite/tokudb.bugs/t/alter_external_lock_assert.test
george-lorch: This all comes from the original test created by Tokutek, so I am unaware of the reason for the test or the history.

Without the two drops of t1_tokudb_tokudb_smallvc_250 we end up with the mtr check_test error:
> --- /home/glorch/dev/PS-8.0/percona-server-install-8.0/mysql-test/var/tmp/check-mysqld_1.result 2018-10-16 18:27:33.836619225 +0300
> +++ /home/glorch/dev/PS-8.0/percona-server-install-8.0/mysql-test/var/tmp/check-mysqld_1.reject 2018-10-16 18:27:34.192619509 +0300
> @@ -954,6 +954,7 @@
>  ./undo_001     UNDO LOG        innodb_undo_001 InnoDB
>  ./mysql.ibd    TABLESPACE      mysql   InnoDB
>  tables_in_test
> +t1_tokudb_tokudb_smallvc_250
>  tables_in_mysql
>  mysql.columns_priv
>  mysql.component

Digging into it, it it seems that there are two tables named t1_tokudb_tokudb_smallvc_250, one persistent, one temporary.  I don't know how or why that's possible, but that is what seems so.  I changed the test to use InnoDB to rule out any TokuDB specific issues and added an extra show tables, so let's look at the test:
>   5 BEGIN;
>   6 CREATE TABLE t1_tokudb_tokudb_smallvc_250(
>   7   c1 BLOB,
>   8   c2 DATE,
>   9   unique(c2)) ENGINE=innodb;
>  10 CREATE TEMPORARY TABLE p LIKE t1_tokudb_tokudb_smallvc_250;
>  11 ALTER TABLE p RENAME t1_tokudb_tokudb_smallvc_250;
>  12 CREATE TEMPORARY TABLE i SELECT 1 t1_tokudb_tokudb_smallvc_250;
>  13 show tables;
>  14 LOCK TABLE i READ;
>  15 ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
>  16 ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
>  17 ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
>  18 UNLOCK TABLES;
>  19 COMMIT;
>  20 DROP TABLE t1_tokudb_tokudb_smallvc_250;
>  21 show tables;
>  22 DROP TABLE t1_tokudb_tokudb_smallvc_250;
>  23 show tables;

And now let's look at the result:
>   1 BEGIN;
>   2 CREATE TABLE t1_tokudb_tokudb_smallvc_250(
>   3 c1 BLOB,
>   4 c2 DATE,
>   5 unique(c2)) ENGINE=innodb;
>   6 CREATE TEMPORARY TABLE p LIKE t1_tokudb_tokudb_smallvc_250;
>   7 ALTER TABLE p RENAME t1_tokudb_tokudb_smallvc_250;
>   8 CREATE TEMPORARY TABLE i SELECT 1 t1_tokudb_tokudb_smallvc_250;
>   9 show tables;
>  10 Tables_in_test
>  11 t1_tokudb_tokudb_smallvc_250
>  12 LOCK TABLE i READ;
>  13 ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
>  14 ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
>  15 ALTER TABLE t1_tokudb_tokudb_smallvc_250 MODIFY c1 DATE;
>  16 UNLOCK TABLES;
>  17 COMMIT;
>  18 DROP TABLE t1_tokudb_tokudb_smallvc_250;
>  19 show tables;
>  20 Tables_in_test
>  21 t1_tokudb_tokudb_smallvc_250
>  22 DROP TABLE t1_tokudb_tokudb_smallvc_250;
>  23 show tables;
>  24 Tables_in_test

We can see that the rename on line 11 of the test succeeds, which means that the temporary table p was successfully renamed to t1_tokudb_tokudb_smallvc_250, yet the persistent table t1_tokudb_tokudb_smallvc_250 of the same name already exists.

In the result, we can see that after the DROP on line 18, the t1_tokudb_tokudb_smallvc_250 still exists, but after the second DROP on line 22, it now doesn't.

My guess is that somehow, there are two table instances with the same name (a server bug?), one persistent, one temporary.  The first DROP drops the temporary, the second drops the persistent.

laurynas-biveinis: The test looks like it could be coming from RQG, and, the actual test is highly likely to be not in DROP TABLE. And indeed temp and regular tables do not share the namespace. I'd suggest
```
DROP TEMPORARY TABLE t1_tokudb_tokudb_smallvc_250;
DROP TABLE t1_tokudb_tokudb_smallvc_250, etc;
```

george-lorch: That makes sense and your suggestion should work.  I was not aware that temp tables go into their own namespace.  Seems like a bad idea all around to me, but than I am not that familiar with temp tables and their use cases.

george-lorch: Yes, that did it.  Pushed the change, test passes locally.  I don't think it is necessary to spawn a new jenkins job for this one minor logical change.

laurynas-biveinis (approved):

PR #2594, opened 2018-10-16
Make `proxy_protocol_networks` variable read-write.

SteamedFish:
Make `proxy_protocol_networks` variable read-write, so that we can dynamically increase the number of haproxy instances when high traffic arrives, without needing to restart percona server.


dutow:
Hello!

The number of proxy instances can already be increased dynamically, as the variable accepts CIDR ranges. The variable is also read-only for security considerations.

I'm closing this pull request for now, if you still have an issue with it, please describe it as a comment.

PR #2595, opened 2018-10-16
PS-269: Fix a crash on querying GLOBAL_TEMPORARY_TABLES along with partitioned InnoDB tables (8.0)

inikep:
`dd_table` can be `nullptr` e.g. in case when it's a cloned handler for I_S.GLOBAL_TEMPORARY_TABLES query.
This commit fixes the issue by detection of `dd_table == nullptr` in `Ha_innopart_share::open_table_parts`.
It also fixes `main.percona_bug1193264` MTR tests.

### DIFF
--- a/mysql-test/r/percona_bug1193264.result
+++ b/mysql-test/r/percona_bug1193264.result
@@ -1,4 +1,3 @@
-DROP TABLE IF EXISTS t1;
 CREATE TABLE t1(a INT PRIMARY KEY, b INT, c INT) ENGINE=InnoDB PARTITION BY HASH(a) PARTITIONS 2;
 INSERT INTO t1 VALUES (1, 1, 1);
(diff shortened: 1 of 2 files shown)


laurynas-biveinis (changes requested):
All these changes are not related to each other - please PR separately

inikep:
PR was divided into 3 separate PRs.

laurynas-biveinis (changes requested):
Please elaborate in the commit message - why is this a correct change? Why is it legal for dd_table to be nullptr here? Is it because it's a cloned handler for for I_S.GLOBAL_TEMP_TABLES query?
Please also change the commit message header accordingly: it's PS-269, but not merely fixing one test, more like "fix a crash on querying GLOBAL_TEMP_TABLES in the presence of partitioned InnoDB tables"

inikep:

laurynas-biveinis (approved):

PR #2597, opened 2018-10-17
PS-269: Fix SEGFAULT in subselect_hash_sj_engine::cleanup() (8.0)

inikep:
The issue was introduced with a wrong merge of PS-4620 from 5.7 to 8.0 regarding changes in `sql/item_subselect.cc`:
https://github.com/percona/percona-server/commit/a5e269a7036c97a02f23acffa78c062d17d6e3d4

This commit reverts `sql/item_subselect.cc` to match 8.0 upstream what fixes `main.subselect_debug` MTR test.


laurynas-biveinis (changes requested):
Commit message needs mention what crash was fixed, and the testcase can be mentioned in the body, as secondary thing.

Have you git blame'd, what introduced the difference in the first place? Does this appear like a merge error?

inikep:

laurynas-biveinis (approved):

PR #2599, opened 2018-10-19
Ps 5.7 4397 4946

george-lorch:


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1518,6 +1520,12 @@ static MYSQL_SYSVAR_BOOL(error_on_suboptimal_collation,
                          "collation is used",
                          nullptr, nullptr, false);
 
+static MYSQL_SYSVAR_BOOL(
+    no_create_column_family, rocksdb_no_create_column_family,
+    PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_READONLY,
+    "Do not allow creation of new Column Families from client interface.", nullptr,
```
laurynas-biveinis: "through index comments"? Are there other affected client interfaces?

storage/rocksdb/rdb_cf_options.h
```diff
@@ -46,7 +46,8 @@ class Rdb_cf_options {
   Rdb_cf_options &operator=(const Rdb_cf_options &) = delete;
   Rdb_cf_options() = default;
 
-  void get(const std::string &cf_name,
+  /* bool true return indicates cf_name was found */
+  bool get(const std::string &cf_name,
```
laurynas-biveinis: Consider MY_NODISCARD annotation to catch unintended return value non-checks easier

storage/rocksdb/rdb_cf_options.h
```diff
@@ -67,7 +68,8 @@ class Rdb_cf_options {
   std::shared_ptr<rocksdb::MergeOperator>
   get_cf_merge_operator(const std::string &cf_name);
 
-  void get_cf_options(const std::string &cf_name,
+  /* bool true return indicates cf_name was found, otherwise default */
+  bool get_cf_options(const std::string &cf_name,
```
laurynas-biveinis, line 72: Likewise

laurynas-biveinis (changes requested):

storage/rocksdb/ha_rocksdb.cc
george-lorch: Currently, no.  This CREATE|ALTER index comment is the only path through here where the create may be prohibited.  I considered a more specific message but thought that the more generic message may be more appropriate since we may (probably will) create or invent new ways for explicit Column Family creation and management in the future and will probably forget to re-word the message.

Re-reading this comment I realize that's a lot of  'probably' and 'ifs'.  So yeah, maybe your right.  A specific message for now and worry about rewording later on if/when we do revisit CF management.

storage/rocksdb/rdb_cf_options.h
george-lorch: That's a good consideration, but I think not all calls to this will care if the options returned were from a specific column family override or from the 'default' set.  I traced the calls back and also went ahead and tried it and it seems that all current uses do check the result so it can't hurt to make it more strict and then loosen it up again later on if need be.

george-lorch:
OK, implemented the minor suggestions and ran mtr locally.  I don't believe a full jenkins run is necessary.

storage/rocksdb/rdb_cf_options.h
george-lorch: Actually, there is one location that calls get_cf_options that does not care about the return code, that is from the plugin init func at ha_rocksdb.cc:4150, so that can't work on this func.

laurynas-biveinis (approved):

PR #2601, opened 2018-10-19
PS-269: Fix `backup_locks_binlog` MTR tests (8.0)

inikep:
1. Add missing `--connection` that were removed along with `LOCK BINLOG FOR BACKUP`.
2. Re-record `main.backup_locks_binlog` MTR tests.


mysql-test/include/backup_locks_binlog.inc
```diff
@@ -258,8 +258,14 @@ INSERT INTO tt_csv VALUES(5);
...
 
+--connection default
+--echo # connection default
+
 UNLOCK TABLES;
 
+--connection con1
+--echo # connection con1
+
 COMMIT;
```
laurynas-biveinis: Does this COMMIT make sense in con1? I don't see an open transaction there

laurynas-biveinis: Does this COMMIT make sense in con1? I don't see an open transaction there

laurynas-biveinis: Is this COMMIT still required in con1? There does not seem to be an open transaction there

laurynas-biveinis: Is this COMMIT still required in con1? There does not seem to be an open transaction there

inikep:
PR was updated

laurynas-biveinis (approved):

PR #2605, opened 2018-10-22
PS-269: fix changed page tracking for 8.0

laurynas-biveinis:
- Fix a merge error in mysql_execute_command that allowed PURGE
  CHANGED_PAGE_BITMAPS to be issued for users with BINLOG_ADMIN
  privilege.
- Fix a merge error in the parser grammar that resulted in
  CHANGED_PAGE_BITMAPS not being allowed as an identifier.
- Do not try to start server on an empty InnoDB instance (which no
  longer works in 8.0 with a different --innodb-data-home-dir setting
  in innodb.percona_changed_page_bmp, but rather initialize it
  properly. The test innodb.percona_changed_page_bmp_1364315 also
  attempts the same, fold it into the former test instead.
- Support read-only parsing of metadata-related MTR types: extend the
  MetatadaRecover class with a bool read_only field and its
  constructor with a corresponding arg, create a new read-only
  instance log_online_metadata_recover, use it for changed page
  tracking instead of the regular crash recovery instance. Add an
  "apply" flag argument to MetadataRecover::parseMetadataLog.
- Refactor log_bmp_sys read and parse buffers to separate classes
  log_read_buffer and log_parse_buffer, with a common log_buffer
  ancestor. Use std::array instead of C-style array. New classes
  facilitate removing parse state from log_bmp_sys, which also got
  more complex in 8.0: a checkpoint may fall in a middle of a log
  record. Partially convert older procedural style code to the new
  classes: split out log_read_buffer::read and
  log_read_buffer::advance from log_online_follow_log_seg, refactor
  log_online_is_valid_log_seg to
  log_read_buffer::is_current_block_valid, split out
  log_parse_buffer::can_parse_current_data,
  log_parse_buffer::parse_next_record, log_parse_buffer::advance, and
  log_parse_buffer::move_unprocessed_to_front from
  log_online_parse_redo_log, refactor log_online_add_to_parse_buf to
  log_parse_buffer::add, refactor log_bitmap_struct::next_parse_lsn to
  log_parse_buffer::current_lsn, move log_bitmap_struct::parse_buf_end
  to log_parse_buffer::end_ptr.
- Add numerous new debug asserts.
- Add some simple methods to struct log_bitmap_struct, but not enough
  to make it a class yet.
- Clarify that MODIFIED_PAGE_END_LSN points to one past the last
  parsed log byte.
- Tweak bitmap_page_write_error debug injection point to fire only
  after server startup, to accomodate its use case and the server
  writing (and tracking) much more log on startup in 8.0. Similarly
  tweak bitmap_page_2_write_error to fire only on file-per-table
  spaces.
- Fix bitmap file sequence enumeration in
  log_online_setp_bitmap_file_range.
- Fix merge errors and remove redundant code in srv_start.
- Use space_id_t and page_no_t instead of ulint where applicable.
- Downgrade server diagnostics on re-tracking redo log on startup from
  warning to information level.
- Merge log_online_rec_has_page and log_online_rec_page_means_page to
  log_online_rec_has_page, update it for the 8.0 MTR types, mark as
  constexpr.
- Bump log file size to 25M for innodb.percona_changed_page_bmp to
  reflect higher InnoDB redo log write volume in 8.0.
- Re-record main.information_schema_keywords,
  innodb.percona_changed_page_bmp_requests. Use force-rmdir mysqltest
  command where applicable. Drop --innodb-file-per-table from
  innodb.percona_changed_page_bmp_debug option file as it matches the
  default. Cleanup trailing whitespace on some lines in
  innodb.percona_changed_pages and
  innodb.percona_changed_page_bmp_debug.
- Replace some selects with asserts in innodb.percona_changed_pages.
- Tweak innodb.percona_changed_page_bmp_crash to list bitmap files
  after the fast shutdown.
- Cleanup innodb.percona_changed_pages_log_tracking_off not to have a
  server option file, remove any stray bitmaps at the start of the
  test

https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/77/, there are unstable testcases to be fixed in a follow-up


storage/innobase/include/log0recv.h
```diff
@@ -426,8 +426,10 @@ class MetadataRecover {
       ut_allocator<std::pair<const table_id_t, PersistentTableMetadata *>>>;
 
  public:
-  /** Default constructor */
-  MetadataRecover() UNIV_NOTHROW {}
+  /** Default constructor
+  @param[in]    read_only_      if set, the instance will only parse the log
+                                without applying any changes */
+  MetadataRecover(bool read_only_) UNIV_NOTHROW : read_only(read_only_) {}
```
percona-ysorokin: explicit

storage/innobase/log/log0online.cc
```diff
@@ -44,37 +45,392 @@ Online database log parsing for changed page tracking */
...
+
+static const constexpr auto LOG_BLOCK_SIZE_NO_TRL =
+    OS_FILE_LOG_BLOCK_SIZE - LOG_BLOCK_TRL_SIZE;
+
+static const constexpr auto LOG_BLOCK_BOUNDARY_LSN_PAD =
+    LOG_BLOCK_HDR_SIZE + LOG_BLOCK_TRL_SIZE;
+
+/** The buffer for reading log data, filled in by recv_read_log_recs and moved
+in chunks to the parse buffer while skipping log block headers and trailers. */
+class log_read_buffer final : public log_buffer<FOLLOW_SCAN_SIZE> {
```
percona-ysorokin, line 130: Private inheritance as `log_buffer<FOLLOW_SCAN_SIZE>` is an implementation detail
```c++
private:
  using base_type = log_buffer<FOLLOW_SCAN_SIZE>;
public:
  using base_type::ccurrent;
  using base_type::set_limit;
  using base_type::get_current_lsn;
```

Using `log_buffer<FOLLOW_SCAN_SIZE>* ptr = new log_read_buffer()` should not be allowed (there is no virtual destructor and it is not needed).

storage/innobase/log/log0online.cc
```diff
@@ -44,37 +45,392 @@ Online database log parsing for changed page tracking */
...
+  @return true if the log block checksum is OK, false otherwise.  */
+  MY_NODISCARD bool is_current_block_valid() const noexcept;
+
+  MY_NODISCARD uint32_t get_current_block_data_len() const noexcept {
+    ut_ad(is_current_block_valid());
+    return log_block_get_data_len(ccurrent());
+  }
+
+  MY_NODISCARD const_iterator cend() const noexcept {
+    return &buffer[current_size];
```
percona-ysorokin, line 153: `buffer.cend()` ?

storage/innobase/log/log0online.cc
```diff
@@ -44,37 +45,392 @@ Online database log parsing for changed page tracking */
...
+    ut_ad(current_lsn - start_lsn <= current_size);
+    current_ptr += OS_FILE_LOG_BLOCK_SIZE;
+  }
+
+#ifdef UNIV_DEBUG
+ private:
+  MY_NODISCARD bool invariants() const noexcept {
+    ut_ad(log_buffer::invariants());
+    // Until C++17 overaligned allocation support verify the alignment manually
+    ut_ad(reinterpret_cast<uintptr_t>(&buffer[0]) %
```
percona-ysorokin, line 169: `(&buffer[0] - nullptr)` gives `std::ptrdiff_t`
So, may be, `(&buffer[0] - nullptr) % INNODB_LOG_WRITE_AHEAD_SIZE_MAX == 0`

storage/innobase/log/log0online.cc
```diff
@@ -44,37 +45,392 @@ Online database log parsing for changed page tracking */
...
+                << ", "
+                << "stored log block n:o " << no << ", "
+                << "expected log block n:o " << expected_no;
+  }
+
+  return checksum_is_ok;
+}
+
+/** The redo log parse buffer */
+class log_parse_buffer final : public log_buffer<RECV_PARSING_BUF_SIZE> {
```
percona-ysorokin, line 219: The same comment about private inheritance here.

storage/innobase/log/log0online.cc
```diff
@@ -378,22 +751,19 @@ static void log_online_track_missing_on_startup(
   ut_ad(last_tracked_lsn != tracking_start_lsn);
   ut_ad(srv_track_changed_pages);
 
-  ib::warn() << "Last tracked LSN in \'" << log_bmp_sys->out.name << "\' is "
-             << last_tracked_lsn << ", but the last checkpoint LSN is "
-             << tracking_start_lsn
-             << ".  This might be due to a server "
-                "crash or a very fast shutdown.";
+  ib::info() << "Last tracked LSN in \'" << log_bmp_sys->out.name << "\' is "
```
percona-ysorokin, line 754: Is changing `warn` to `info` here intentional?

percona-ysorokin (approved):
LGTM

Only minor stylistic comments.

storage/innobase/log/log0online.cc
laurynas-biveinis: might point differently: buffer.cend() is always at FOLLOW_SCAN_SIZE, current_size is set by ::read()

storage/innobase/log/log0online.cc
laurynas-biveinis: Is this idiomatic? Googling gives mixed answers, including UB. IMHO cast to uinptr_t has clearer intent

storage/innobase/log/log0online.cc
laurynas-biveinis: Yes: one of things changed in 8.0 that checkpoint LSN can be legally more often not equal to last tracked LSN, making this diagnostic not very useful as a warning

storage/innobase/log/log0online.cc
laurynas-biveinis: Actually no, both child class objects get their parent public methods called: ccurrent(), set_limit(), get_current_lsn()

storage/innobase/log/log0online.cc
percona-ysorokin: Oh, my bad. Missed the `current_size` part.

storage/innobase/log/log0online.cc
percona-ysorokin: For some reason, I've been scared recently with using `reinterpret_cast<>`s on pointers because modern compilers started consider this UB in a lot of cases and optimize this out. In this particular case, though. I believe it OK. So, I am fine with leaving this as is.

storage/innobase/log/log0online.cc
percona-ysorokin: My main concern is still the same but theoretical. In the following usage scenario
```c++
log_buffer<FOLLOW_SCAN_SIZE>* ptr = new log_read_buffer();
delete ptr;
```
only the desctuctor of the base class will be called and therefore we will get memory leaks.
On the other hand, it has nothing in common with how these classes are used in your code. So, up to you.

percona-ysorokin:
No more comments, OK to merge.

laurynas-biveinis:
Only difference is added "explicit". Jenkins is down, tested in Travis and locally only

PR #2608, opened 2018-10-22
PS-4953 : rocksdb.truncate_table3 unstable

george-lorch:
- Test is missing error log suppressions that will occur if test is run on a
  clean data-dir.  Added missing suppressions.


george-lorch:
https://ps.cd.percona.com/job/percona-server-5.7-param/94/

laurynas-biveinis (approved):

PR #2609, opened 2018-10-23
Ps 8.0 4953 4947

george-lorch:


george-lorch:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/78/

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8381,12 +8381,23 @@ ha_rocksdb::generate_cf_name(const uint index, const TABLE *const table_arg,
...
+  // Strip leading and trailing whitespace as that is not part of a valid name
+  if (!cf_name.empty()) {
+    std::string::size_type start = cf_name.find_first_not_of(" ");
+    std::string::size_type end = cf_name.find_last_not_of(" ");
+
+    if (start == std::string::npos && end == std::string::npos) {
+      cf_name.erase();
+    } else {
+      if (end != std::string::npos)
+        cf_name.erase(end + 1);
```
laurynas-biveinis: This cannot be right, it only erases one character

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8381,12 +8381,23 @@ ha_rocksdb::generate_cf_name(const uint index, const TABLE *const table_arg,
...
-  // use any qualifiers at the moment. (aka its a special case)
-  if (cf_name.empty() && !key_comment.empty()) {
-    return key_comment;
+  // If we didn't find any partitioned/non-partitioned qualifiers, DO NOT return
+  // the  comment itself. FB Behavior is exactly the opposite which then causes
+  // any index comment to create a new column family.
+
+  // Strip leading and trailing whitespace as that is not part of a valid name
+  if (!cf_name.empty()) {
+    std::string::size_type start = cf_name.find_first_not_of(" ");
```
laurynas-biveinis: (since you have to repush because of a comment below, if not, ignore)
const auto
const auto

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8381,12 +8381,23 @@ ha_rocksdb::generate_cf_name(const uint index, const TABLE *const table_arg,
...
   }
 
-  // If we didn't find any partitioned/non-partitioned qualifiers, return the
-  // comment itself.  NOTE: this currently handles returning the cf name
-  // specified in the index comment in the case of no partitions, which doesn't
-  // use any qualifiers at the moment. (aka its a special case)
-  if (cf_name.empty() && !key_comment.empty()) {
-    return key_comment;
+  // If we didn't find any partitioned/non-partitioned qualifiers, DO NOT return
+  // the  comment itself. FB Behavior is exactly the opposite which then causes
```
laurynas-biveinis: (since you have to repush because of a comment below, if not, ignore)
- double space between "the" and "comment"
- please narrow FB version to 5.6, they could change this in 8.0

storage/rocksdb/rdb_utils.cc
```diff
@@ -224,9 +224,15 @@ const std::vector<std::string> parse_into_tokens(const std::string &s,
   std::vector<std::string> tokens;
   std::string t;
   std::stringstream ss(s);
-
   while (getline(ss, t, delim)) {
-    tokens.push_back(t);
+    std::string::size_type start = t.find_first_not_of(" ");
+    std::string::size_type end = t.find_last_not_of(" ");
+    if (end != std::string::npos)
+      t.erase(end + 1);
```
laurynas-biveinis: This cannot be right, it only erases one character

storage/rocksdb/rdb_utils.cc
```diff
@@ -224,9 +224,15 @@ const std::vector<std::string> parse_into_tokens(const std::string &s,
   std::vector<std::string> tokens;
   std::string t;
   std::stringstream ss(s);
-
   while (getline(ss, t, delim)) {
-    tokens.push_back(t);
+    std::string::size_type start = t.find_first_not_of(" ");
```
laurynas-biveinis: Duplicated - pull up a trim function

laurynas-biveinis (changes requested):

storage/rocksdb/ha_rocksdb.cc
george-lorch: > sequence (1)  string& erase (size_t pos = 0, size_t len = npos)'
> ...
> (1) sequence Erases the portion of the string value that begins at the character position pos and spans len characters (or until the end of the string, if either the content is too short or if len is string::npos. Notice that the default argument erases all characters in the string (like member function clear).

This does raise a good point though, I did not add a specific test case for these parser changes and there should be one.

laurynas-biveinis: You are right, I confused this with a single arg iterator overload

george-lorch:
Made some minor changes, added new test specifically for column family naming and use within index hints.  New jenkins https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/79/

mysql-test/suite/rocksdb/t/rocksdb_cf_names.test
```diff
@@ -0,0 +1,92 @@
+--source include/have_rocksdb.inc
+
+# Column Family names and name specifications have some pretty strict rules:
```
laurynas-biveinis, line 3: Looks like you wrote the user doc page for this too, make sure Doc engineer sees this

laurynas-biveinis (approved):

mysql-test/suite/rocksdb/t/rocksdb_cf_names.test
george-lorch: That was the intention, there is actually a fair amount more when discussing partitions.

PR #2614, opened 2018-10-25
PS-4972 and PS-269: Fix "mysqldump --innodb-optimize-keys" (8.0)

inikep:
1. Fix SEGFAULT related to using `static std::list<std::string>* skipped_keys_list` as nullptr
2. Don't optimize tables with `FOREIGN KEYS` with `REFERENCES` to another table
3. Check optimization with FOREIGN KEYs with different table names in `main.percona_mysqldump_innodb_optimize_keys`
4. Re-record `main.percona_mysqldump_innodb_optimize_keys`

### DIFF
--- a/client/mysqldump.cc
+++ b/client/mysqldump.cc
@@ -232,8 +232,8 @@ collation_unordered_set<string> *ignore_table;
 
 static collation_unordered_set<std::string> *processed_compression_dictionaries;
 
-static std::list<std::string> *skipped_keys_list = nullptr;
-static std::list<std::string> *alter_constraints_list = nullptr;
+static std::list<std::string> skipped_keys_list;
+static std::list<std::string> alter_constraints_list;
 
 static struct my_option my_long_options[] = {
     {"all-databases", 'A',
@@ -2693,6 +2693,7 @@ static bool contains_autoinc_column(const char *autoinc_column,
 
   SYNOPSIS
   skip_secondary_keys()
+  table                     table name
   create_str                SHOW CREATE TABLE output
   has_pk                    TRUE, if the table has PRIMARY KEY
   (or UNIQUE key on non-nullable columns)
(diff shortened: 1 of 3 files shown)


client/mysqldump.cc
```diff
@@ -2706,16 +2707,27 @@ static bool contains_autoinc_column(const char *autoinc_column,
   alter_constraints_list and removes them from the input string.
 */
 
-static void skip_secondary_keys(char *create_str, bool has_pk) noexcept {
+static void skip_secondary_keys(char *table, char *create_str,
```
laurynas-biveinis: const char *table? (even const std::string &table if not too work)

inikep:
PR was updated

client/mysqldump.cc
```diff
@@ -2759,9 +2774,9 @@ static void skip_secondary_keys(char *create_str, bool has_pk) noexcept {
           my_strndup(PSI_NOT_INSTRUMENTED, ptr, end - ptr + 1, MYF(MY_FAE));
 
       if (type == key_type_t::CONSTRAINT)
-        alter_constraints_list->emplace_front(data);
+        alter_constraints_list.emplace_back(data);
```
laurynas-biveinis, line 2777: Why s/front/back?

mysql-test/r/percona_mysqldump_innodb_optimize_keys.result
```diff
@@ -7,7 +7,7 @@ CREATE TABLE t1 (a INT NOT NULL PRIMARY KEY, b INT, KEY(b)) ENGINE=MyISAM;
 /*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
 /*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
 /*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;
-/*!40101 SET NAMES utf8 */;
+ SET NAMES utf8mb4 ;
```
laurynas-biveinis, line 10: Does extra leading space and space before the semicolon come from our patches or upstream?

client/mysqldump.cc
inikep: You removed 
```
    alter_constraints_list= list_reverse(alter_constraints_list);
```
in `dump_skipped_keys`.

mysql-test/r/percona_mysqldump_innodb_optimize_keys.result
inikep: It's upstream at https://github.com/percona/percona-server/commit/4491eaf786696c5423560747eebc75b4132d082f

client/mysqldump.cc
laurynas-biveinis: That's right, but haven't I also changed which end of the list gets inserted to at the same time?

inikep: `list_cons()` in 5.7 adds elements to the front of a list so it should be replaced with `emplace_front` + list_reverse or `emplace_back` without reversing.

laurynas-biveinis: OK, I must have made a mistake in my initial port here

laurynas-biveinis (approved):

PR #2617, opened 2018-10-25

laurynas-biveinis:
Cleanup TokuDB testsuite to use the new mysqltest primitives in 8.0,
such as force-rmdir, to reduce exec/system invocation of shell
commands. Remove couple of broken tests.

https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/83/
(The rpl_tokudb_commit_sync failure happens after the testcase change point)


george-lorch (approved):
Excellent...

PR #2618, opened 2018-10-26

laurynas-biveinis:
Partially clean up MyRocks testcases not to use shell commands when a
mysqltest alternative is available, mainly because of force-rmdir
mysqltest command, introduced in 8.0.

Testing [1] shows some failures, but no diff against baseline [2]
[1] https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/85/#showFailuresLink
[2] https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/84/#showFailuresLink

### DIFF
--- a/mysql-test/suite/rocksdb/t/compact_deletes.test
+++ b/mysql-test/suite/rocksdb/t/compact_deletes.test
@@ -28,8 +28,8 @@ while ($i<1000)
 set global rocksdb_force_flush_memtable_now=1;
 optimize table r1;
 
---exec echo Test 1: Do a bunch of updates without setting the compaction sysvar
---exec echo Expect: no compaction
+--echo Test 1: Do a bunch of updates without setting the compaction sysvar
+--echo Expect: no compaction
 let $window = 0;
 let $deletes = 0;
 let $file_size = 0;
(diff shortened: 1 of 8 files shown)


mysql-test/suite/rocksdb/t/compact_deletes.test
```diff
@@ -58,8 +58,8 @@ let $primary = 1;
 let $no_more_deletes = 0;
 --source suite/rocksdb/include/compact_deletes_test.inc
 
---exec echo Test 4: Do a bunch of secondary key updates and set the compaction sysvar
---exec echo Expect: compaction
+--echo Test 4: Do a bunch of secondary key updates and set the compaction sysvar
+--echo Expect: compaction
```
george-lorch, line 62: This can probably be upstreamed.  I'll make a note to do it in the next test fixup push.

george-lorch (approved):

mysql-test/suite/rocksdb/t/compact_deletes.test
laurynas-biveinis: Everything except force-rmdir can be upstreamed to 5.6

PR #2627, opened 2018-10-30
PS-4805: Re-implement compression dictionaries in 8.0

satya-bodapati:
1. Re-implement CREATE COMPRESSION_DICTIONARY, DROP COMPRESSION_DICTIONARY
2. Allow CREATE TABLE to use 'WITH COMPRESSION_DICTIONARY'
3. DDLs on tables with compression dictionary
4. Prevent dropping of COMPRESSION DICTIONARY with existing references
5. Allow compression_dictionary on tables with partitions
6. Fix mysqldump to show compression dictionary
7. New I_S views on compression dictionary tables
8. Fix tests to use new I_S views
9. Throw proper errors (ER_COMPRESSION_*)
10. Error handling in innodb_read_only mode
11. Handle upgrade from 5.7
12. Handle creation of dictionary tables in 8.0 at startup (mysql-8.0 -> PS8.0 )
13. Add upgrade test
14. Prevent direct access to mysql.compression_dictionary and mysql.compression_dictionary_cols table


satya-bodapati:
All pending tasks completed. Ready for review :)

mysql-test/suite/innodb/r/xtradb_compressed_columns_alter_table.result
```diff
@@ -230,7 +230,7 @@ long_value_match
 short_value_match
 1
 original_dictionary_reference_cleared
-1
+0
```
laurynas-biveinis, line 233: Why has this changed?

mysql-test/suite/innodb/r/xtradb_compressed_columns_consistency_antelope_compact_set1.result
```diff
@@ -3,10 +3,53 @@ call mtr.add_suppression("InnoDB: Starting to delete and rewrite log files");
 call mtr.add_suppression("InnoDB: New log files created");
```
laurynas-biveinis, line 3: "antelope" is now confusing in the name of this and other test file names. And the tests themselves now duplicate the barracuda tests now that there is only one format, thus can be safely removed?

mysql-test/suite/innodb/r/xtradb_compressed_columns_consistency_barracuda_compact_set1.result
```diff
@@ -3,10 +3,53 @@ call mtr.add_suppression("InnoDB: Starting to delete and rewrite log files");
 call mtr.add_suppression("InnoDB: New log files created");
```
laurynas-biveinis: For all tests that have it, rename to remove "barracuda" from the file name

mysql-test/suite/innodb/t/xtradb_compressed_columns_with_dictionaries.test
```diff
@@ -66,26 +66,24 @@ SET @long_dictionary_name = REPEAT('d', 64);
 
 # max dictionary data size is 32506 bytes - check if an error
 # is reported if this number is exceeded
+# From 8.0, it is BLOB and size is unlimited. No error for long_dictionary_data
```
percona-ysorokin: `32506` is the max size of the dictionary dictated by zlib itself, not by the max blob size.
So, I believe, creating a dictionary longer than 32506 should still be an error.

sql/dd/impl/bootstrapper.cc
```diff
@@ -2030,6 +2032,68 @@ bool repopulate_charsets_and_collations(THD *thd) {
   return error;
 }
 
+/** On startup from mysql datadir to Percona Server, compression dictionary
+tables and I_S views on them will be missing. We check if they are missing
+and create the ompression dictionary tables mysql.compression_dictionary
```
percona-ysorokin: s/ompression/compression/

sql/dd/impl/bootstrapper.cc
```diff
@@ -2214,6 +2278,24 @@ bool initialize_dictionary(THD *thd, bool is_dd_upgrade_57,
     return true;
   }
 
+  // Create the mysql.compression_dictionary_table
+  if (execute_query(thd, dd::String_type(compression_dict::dict_table_str))) {
+    return true;
+  }
+
+  if (execute_query(
```
percona-ysorokin: Should this be 
```c++
if (execute_query(thd,
                    dd::String_type(compression_dict::dict_cols_table_str))) {
    return true;
  }
```
?

sql/dd/impl/system_views/compression_dictionary.cc
```diff
@@ -0,0 +1,40 @@
...
+
+*****************************************************************************/
+
+#include "sql/dd/impl/system_views/compression_dictionary.h"
+
+namespace dd {
+namespace system_views {
+
+const Compression_dictionary &Compression_dictionary::instance() {
+  static Compression_dictionary *s_instance = new Compression_dictionary();
```
percona-ysorokin, line 25: Why `new` here? Memory checkers will complain about the leak here.
May be just
```c++
static Compression_dictionary s_instance;
return s_instance;
```
?
Please also notice that this is thread safe only in c++11 mode, where we have magic statics.

sql/dd/impl/system_views/compression_dictionary.h
```diff
@@ -0,0 +1,54 @@
...
+class Compression_dictionary
+    : public System_view_impl<System_view_select_definition_impl> {
+ public:
+  enum enum_fields { FIELD_DICT_VERSION, FIELD_DICT_NAME, FIELD_DICT_DATA };
+
+  Compression_dictionary();
+
+  static const Compression_dictionary &instance();
+
+  static const String_type &view_name() {
```
percona-ysorokin: `noexcept`

sql/dd/impl/system_views/compression_dictionary.h
```diff
@@ -0,0 +1,54 @@
...
+    static String_type s_view_name("COMPRESSION_DICTIONARY");
+    return s_view_name;
+  }
+
+  virtual const String_type &name() const {
```
percona-ysorokin: `noexcept override`

sql/dd/impl/system_views/compression_dictionary_tables.cc
```diff
@@ -0,0 +1,52 @@
...
+
+*****************************************************************************/
+
+#include "sql/dd/impl/system_views/compression_dictionary_tables.h"
+
+namespace dd {
+namespace system_views {
+
+const Compression_dictionary_tables &Compression_dictionary_tables::instance() {
+  static Compression_dictionary_tables *s_instance =
```
percona-ysorokin, line 26: Likewise,
```c++
static Compression_dictionary_tables s_instance;
return s_instance;
```

sql/dd/impl/system_views/compression_dictionary.h
```diff
@@ -0,0 +1,54 @@
...
+  view definition
+*/
+class Compression_dictionary
+    : public System_view_impl<System_view_select_definition_impl> {
+ public:
+  enum enum_fields { FIELD_DICT_VERSION, FIELD_DICT_NAME, FIELD_DICT_DATA };
+
+  Compression_dictionary();
+
+  static const Compression_dictionary &instance();
```
percona-ysorokin: `noexcept`

sql/dd/impl/system_views/compression_dictionary_tables.h
```diff
@@ -0,0 +1,59 @@
...
+  enum enum_fields {
+    FIELD_TABLE_SCHEMA,
+    FIELD_TABLE_NAME,
+    FIELD_COLUMN_NAME,
+    FIELD_DICT_NAME
+  };
+
+  Compression_dictionary_tables();
+
+  static const Compression_dictionary_tables &instance();
```
percona-ysorokin: `noexcept`

sql/dd/impl/system_views/compression_dictionary_tables.h
```diff
@@ -0,0 +1,59 @@
...
+
+  static const String_type &view_name() {
```
percona-ysorokin: `noexcept`

sql/dd/impl/system_views/compression_dictionary_tables.h
```diff
@@ -0,0 +1,59 @@
...
+    static String_type s_view_name("COMPRESSION_DICTIONARY_TABLES");
+    return s_view_name;
+  }
+
+  virtual const String_type &name() const {
```
percona-ysorokin: `noexcept override`

sql/handler.h
```diff
@@ -1738,6 +1671,12 @@ typedef bool (*rotate_encryption_master_key_t)(void);
 */
 using fix_tablespaces_empty_uuid_t = bool (*)(void);
 
+using compression_dict_data_vec_t =
+    std::vector<std::pair<std::string, std::string>>;
```
percona-ysorokin, line 1675: mysql allocator?

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,226 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
 #include "sql/sql_class.h"  // THD
 #include "sql/sql_table.h"  // write_bin_log
+#include "sql/table.h"
+#include "sql/thd_raii.h"
+#include "sql/transaction.h"
+
+namespace compression_dict {
+
+/** Fields of mysql.compression_dictionary table */
+static const uint32_t COMPRESSION_DICTIONARY_FIELD_ID = 0;
```
percona-ysorokin: Here and below
`static constexpr`?

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,226 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+
+/** Index of compression_dictionary_cols table */
+static const uint32_t COMPRESSION_DICTIONARY_COLS_TABLE_INDEX = 0;
+
+static const char *COMPRESSION_DICTIONARY_COLS_DB = "mysql";
+static const char *COMPRESSION_DICTIONARY_COLS_TABLE =
+    "compression_dictionary_cols";
+
+bool upgrade_transfer_compression_dict_data(THD *thd) {
+  std::vector<std::pair<std::string, std::string>> zip_dict_vec;
```
percona-ysorokin: `typedef/using` for complex type.
Consider putting mysql allocator here as well.

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,226 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+    std::string &data = elem.second;
+
+    int ret = create_zip_dict(thd, name.c_str(), name.length(), data.c_str(),
+                              data.length(), false, true);
+
+    if (ret != 0) {
+      return (true);
+    }
+
+    std::cout << "Name is: " << name << " Data is: " << data << std::endl;
```
percona-ysorokin: Should this be some king of XXX_LOG instead of direct `cout`?

mysql-test/suite/innodb/t/xtradb_compressed_columns_with_dictionaries.test
satya-bodapati: I see. Let me debug and find out why it is not giving error in 8.0

sql/sql_zip_dict.cc
satya-bodapati: I will remove this

mysql-test/suite/innodb/t/xtradb_compressed_columns_with_dictionaries.test
satya-bodapati: atleast while creating compression dictionary, it is just about storing data into table. The size can be unlimited
"
The implementation may silently use a subset of the provided dictionary if the dictionary cannot fit in the current window associated with stream (see deflateInit2_()). 
"
So the window size for zlib is maxed at 32K ?

satya-bodapati:
```
int deflateSetDictionary(z_streamp stream, const Bytef * dictionary, uInt dictlen);

Description

The deflateSetDictionary() function shall initialize the compression dictionary associated with stream using the dictlen bytes referenced by dictionary.

The implementation may silently use a subset of the provided dictionary if the dictionary cannot fit in the current window associated with stream (see deflateInit2_()). The application should ensure that the dictionary is sorted such that the most commonly used strings occur at the end of the dictionary.

```

sql/dd/impl/bootstrapper.cc
satya-bodapati: oh yes, I missed it

sql/dd/impl/system_views/compression_dictionary.cc
satya-bodapati: I followed the other view creations. For example:

For example:
```
const Schemata &Schemata::instance() {
  static Schemata *s_instance = new Schemata();
  return *s_instance;
}
```
May be the memory is freed on shutdown

satya-bodapati: sql/dd/impl/system_views/schemata.cc

sql/sql_zip_dict.cc
satya-bodapati: const expr? there is no evalution here

sql/sql_zip_dict.cc
satya-bodapati: made it DBUG_LOG

sql/sql_zip_dict.cc
satya-bodapati: Any example? std::vector with mysql allocator?

sql/sql_zip_dict.cc
laurynas-biveinis: static const constexpr, there is evaluation, COMPRESSION_DICTIONARY_FIELD_ID variable evaluates to constant zero at compilation time, whereas static const is still a variable with memory location assigned, which can be evaluated at runtime (or at compilation time as optimization)

sql/sql_zip_dict.cc
laurynas-biveinis: See sql/malloc_allocator.h: 
  Example usage:
  vector<int, Malloc_allocator<int>>
    v((Malloc_allocator<int>(PSI_NOT_INSTRUMENTED)));

mysql-test/suite/innodb/r/xtradb_compressed_columns_consistency_antelope_compact_set1.result
```diff
@@ -3,10 +3,53 @@ call mtr.add_suppression("InnoDB: Starting to delete and rewrite log files");
...
 call mtr.add_suppression("InnoDB: The log sequence numbers [0-9]+ and [0-9]+ in ibdata files do not match the log sequence number [0-9]+ in the ib_logfiles");
 # restart:--innodb-log-file-size=192M
-SET @saved_innodb_file_format = @@global.innodb_file_format;
 SET @saved_innodb_compressed_columns_zip_level = @@global.innodb_compressed_columns_zip_level;
-SET GLOBAL innodb_file_format = Antelope;
 SET GLOBAL innodb_compressed_columns_zip_level = 0;
-SET GLOBAL innodb_file_format = @saved_innodb_file_format;
+Warnings:
```
laurynas-biveinis: Please do the required "CHARACTER SET UTF8MB4" here

mysql-test/suite/innodb/t/xtradb_compressed_columns.test
```diff
@@ -201,6 +201,7 @@ SET GLOBAL innodb_compressed_columns_zip_level = @saved_innodb_compressed_column
 #
 call mtr.add_suppression("InnoDB: Column .* in table .* has unexpected COMPRESSED attribute");
 
+--error ER_NOT_ALLOWED_COMMAND
 ALTER TABLE mysql.innodb_index_stats
   MODIFY stat_description VARCHAR(1024) COLLATE utf8_bin NOT NULL COLUMN_FORMAT COMPRESSED;
 CREATE TABLE t1(i INT);
```
laurynas-biveinis: Either conservatively drop the restoring ALTER TABLE at lines 212-213 - since mysql.innodb_index_chats definition was not changed - or consider that the 1st ALTER TABLE was blocked, hence that could be completely enough for this test

mysql-test/suite/innodb/t/xtradb_compressed_columns_partitioning.test
```diff
@@ -124,21 +124,21 @@ INSERT INTO expected_alter_table_errors VALUES
 
 # iterating over all possible combinations of original table storage engine /
 # storage storage specified in 'ALTER TABLE' statement
-# 2 * 3 = 6
+# 1 * 3 = 3
```
laurynas-biveinis, line 127: Not enough simplification, we are not iterating over "1x3" options, we are iterating over 3 options only - please fix the comment above, remove this comment, pull up let $original_engine = InnoDB above the loop, remove if($original_engine) etc...

mysql-test/suite/innodb/t/xtradb_compressed_columns_partitioning.test
```diff
@@ -124,21 +124,21 @@ INSERT INTO expected_alter_table_errors VALUES
...
   eval SET @partial_index = $combination_index;
   --let $alter_engine = `SELECT CASE @partial_index % 3 WHEN 0 THEN '' WHEN 1 THEN 'MyISAM' ELSE 'InnoDB' END`
   eval SET @partial_index = @partial_index DIV 3;
-  --let $original_engine = `SELECT IF(@partial_index % 2 = 0, 'MyISAM', 'InnoDB')`
+  --let $original_engine = InnoDB
 
   # extracting $expected_error from the 'expected_alter_table_errors' for the given
   # ($original_engine, $alter_engine)
-  --let $expected_error = `SELECT error_code FROM expected_alter_table_errors WHERE original_se = '$original_engine' AND alter_se = '$alter_engine'`
+  --let $expected_error = `SELECT error_code FROM expected_alter_table_errors WHERE original_se = 'InnoDB' AND alter_se = '$alter_engine'`
```
laurynas-biveinis: either hardcode InnoDB everywhere and remove original_engine completely, either $original_engine here

mysql-test/t/dd_upgrade_compression_dict-master.opt
```diff
@@ -0,0 +1 @@
+--bootstrap --innodb_page_size=16k
```
laurynas-biveinis, line 1: If the 16K check is needed (in theory it is but we are far from testing MTR combinations --mysqld=--innodb_page_size=not16K), then include have_innodb_16k.inc too?

mysql-test/t/dd_upgrade_compression_dict.test
```diff
@@ -0,0 +1,171 @@
...
+call mtr.add_suppression("Resolving dependency for the view");
+call mtr.add_suppression("references invalid");
+call mtr.add_suppression("doesn't exist");
+call mtr.add_suppression("information_schema");
+call mtr.add_suppression("Storage engine '.*' does not support system tables. \\[mysql.*\\]");
+call mtr.add_suppression("Table 'mysql.component' doesn't exist");
+call mtr.add_suppression("is expected to be transactional");
+call mtr.add_suppression("table is missing or has an incorrect definition");
+call mtr.add_suppression("ACL DDLs will not work unless mysql_upgrade is executed");
+call mtr.add_suppression(".* Native table .* has the wrong structure");
```
laurynas-biveinis: Here and below, drop any preceding and trailing .*, suppressions are already a substring match

mysql-test/t/dd_upgrade_compression_dict.test
```diff
@@ -0,0 +1,171 @@
...
+let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/data57_zip_dict;
+
+--echo # Copy the remote tablespace & DB zip files from suite location to working location.
+--copy_file $MYSQLTEST_VARDIR/std_data/data57_zip_dict.zip $MYSQL_TMP_DIR/data57_zip_dict.zip
+
+--echo # Check that the file exists in the working folder.
+--file_exists $MYSQL_TMP_DIR/data57_zip_dict.zip
+
+--echo # Unzip the zip file.
+--exec unzip -qo $MYSQL_TMP_DIR/data57_zip_dict.zip -d $MYSQL_TMP_DIR
```
laurynas-biveinis, line 61: Add have_util_unzip.inc at the top

mysql-test/t/dd_upgrade_compression_dict.test
```diff
@@ -0,0 +1,171 @@
...
+
+SET @long_dictionary_data = REPEAT('ab', 32506 / 2);
+
+SELECT dict_data = @long_dictionary_data AS long_dictionary_data_matches FROM information_schema.compression_dictionary WHERE dict_name = 'd1';
+
+SET @dict_data8 = 'aaaaaaaabbbbbbbbccccccccdddddddd';
+SELECT dict_data = @dict_data8 AS short_dictionary_data_matches FROM information_schema.compression_dictionary WHERE dict_name = 'd2';
+
+--echo # Retrieve all compression dictionaries
+--echo # Long dictionaryd data is already retrieved and verifed above
```
laurynas-biveinis: "dictionaryd"

mysql-test/t/dd_upgrade_compression_dict.test
```diff
@@ -0,0 +1,171 @@
...
```
laurynas-biveinis: "dictionaryd"

mysql-test/t/dd_upgrade_test.test
```diff
@@ -674,7 +674,7 @@ if ($MTR_RECORD == 0) {
 
 # If --record is used, save the json files created in the $DEST_DIR
 # back to the $SRC_DIR.
-if ($RECORD == 1) {
+if ($MTR_RECORD == 1) {
```
laurynas-biveinis, line 677: If it is what it looks like, please also report upstream (and attach the fix in OCA)

sql/dd/impl/bootstrapper.cc
laurynas-biveinis: "ompression"

sql/dd/impl/bootstrapper.cc
```diff
@@ -2030,6 +2032,68 @@ bool repopulate_charsets_and_collations(THD *thd) {
...
+tables and I_S views on them will be missing. We check if they are missing
+and create the ompression dictionary tables mysql.compression_dictionary
+and mysql.compression_dictionary_cols
+@param[in,out]  thd  Session context
+@return false on success, true on failure */
+static bool check_and_create_compression_dict_tables(THD *thd) {
+  const dd::Table *new_table_def = nullptr;
+  if (thd->dd_client()->acquire("mysql", "compression_dictionary",
+                                &new_table_def)) {
+    // table doesn't exist
```
laurynas-biveinis: What is the purpose of this if statement?

sql/dd/impl/bootstrapper.cc
```diff
@@ -2030,6 +2032,68 @@ bool repopulate_charsets_and_collations(THD *thd) {
...
+    return true;
+  }
+
+  if (dd::info_schema::create_non_dd_views(thd, false)) {
+    return true;
+  }
+
+  /*
+    We must commit the transaction before executing a new query, which
+    expects the transaction to be empty, and finally, turn FK checks back on.
```
laurynas-biveinis: ", and finally, turn FK checks back on" - where does that happen? In dd::end_transaction?

sql/dd/impl/bootstrapper.cc
```diff
@@ -2214,6 +2278,24 @@ bool initialize_dictionary(THD *thd, bool is_dd_upgrade_57,
...
+
+  if (execute_query(
+          thd, dd::String_type(
+                   "CREATE TABLE IF NOT EXISTS "
+                   "mysql.compression_dictionary_cols(table_id BIGINT UNSIGNED "
+                   "NOT NULL, column_id BIGINT UNSIGNED NOT NULL, dict_id "
+                   "BIGINT UNSIGNED NOT NULL, FOREIGN KEY (dict_id) REFERENCES "
+                   "compression_dictionary(id) ON DELETE RESTRICT ON UPDATE "
+                   "RESTRICT, UNIQUE KEY table_idx(table_id, column_id)) "
+                   "ENGINE=InnoDB CHARACTER SET=utf8 COLLATE=utf8_general_ci "
```
laurynas-biveinis: utf8mb4 / utf8mb4_general_ci ?

sql/dd/impl/dictionary_impl.cc
```diff
@@ -341,7 +341,10 @@ bool Dictionary_impl::is_dd_table_access_allowed(bool is_dd_internal_thread,
   return (table_type == nullptr ||
           (*table_type == System_tables::Types::DDSE_PROTECTED &&
            !is_ddl_statement) ||
-          *table_type == System_tables::Types::SYSTEM);
+          (*table_type == System_tables::Types::SYSTEM &&
```
laurynas-biveinis, line 347: Why is this needed? Probably will require a comment

sql/dd/info_schema/metadata.cc
```diff
@@ -567,10 +568,60 @@ bool create_system_views(THD *thd) {
...
+/** Create INFORMATION_SCHEMA views on non-DD tables like
+mysql.compression_dictionary and mysql.compression_dictionary_cols tables
+@param[in,out]    thd                   Session context
+@param[in]        store_i_s_version     when true, stores the I_S version
+                                        false doesn't store, used when I_S
+                                        tables are created on startup
+                                        mysql-8.0 to PS-8.0
+@return false on success, true on failure */
+bool create_non_dd_views(THD *thd, bool store_i_s_version) {
+  // Force use of utf8 charset.
```
laurynas-biveinis: utf8mb4? Since the server is in transition where utf8 already means / is going to mean shortly utf8mb4, not utf8mb3, better to spell out to avoid confusion

sql/dd/info_schema/metadata.cc
```diff
@@ -567,10 +568,60 @@ bool create_system_views(THD *thd) {
...
+  const CHARSET_INFO *client_cs = thd->variables.character_set_client;
+  const CHARSET_INFO *cs = thd->variables.collation_connection;
+  const CHARSET_INFO *m_client_cs, *m_connection_cl;
+
+  resolve_charset("utf8", system_charset_info, &m_client_cs);
```
laurynas-biveinis: utf8mb4?

sql/dd/upgrade/table.cc
```diff
@@ -1647,6 +1648,29 @@ static bool migrate_table_to_dd(THD *thd, const String_type &schema_name,
     return true;
   }
 
+  const dd::Table *new_table_def = nullptr;
+  if (thd->dd_client()->acquire(schema_name.c_str(), table_name.c_str(),
+                                &new_table_def)) {
+    DBUG_LOG("zip_dict",
+             "Upgrade: Acquiring dictionary table object failed "
+             " table_db: "
+                 << schema_name.c_str()
```
laurynas-biveinis: Are c_str() here required? no operator<< overload?

sql/field.h
```diff
@@ -106,6 +106,8 @@ struct TABLE_SHARE;
 struct TYPELIB;
 struct timeval;
 
+extern const LEX_CSTRING null_lex_cstr;
```
laurynas-biveinis, line 109: Include the header instead?

sql/field.h
```diff
@@ -4315,6 +4317,13 @@ class Create_field {
 
   Nullable<gis::srid_t> m_srid;
 
+  /*
+    Store dict_id after verifying zip_dict_name exists. The stored id
+    is filled in dd::Column::options and later used to fill TABLE_SHARE*
+    zip_dict_name and zip_dict_data
+  */
+  uint64 zip_dict_id;
```
laurynas-biveinis, line 4325: uint64_t ?

sql/handler.h
```diff
@@ -4169,7 +4107,7 @@ class handler {
 
   virtual bool is_ignorable_error(int error);
   MY_NODISCARD virtual bool continue_partition_copying_on_error(
-    int error MY_ATTRIBUTE((unused))) {
+      int error MY_ATTRIBUTE((unused))) {
```
laurynas-biveinis: Please check with @inikep , it looks like you are oscillating formatting here?

sql/item_func.cc
```diff
@@ -7504,14 +7504,21 @@ static bool check_table_and_trigger_access(Item **args, bool check_trigger_acl,
   }
 
   // Make sure we have safe string to access.
```
laurynas-biveinis: Outdated comment now

sql/item_func.cc
```diff
@@ -7504,14 +7504,21 @@ static bool check_table_and_trigger_access(Item **args, bool check_trigger_acl,
...
-  table_name_ptr->c_ptr_safe();
+  const char *sch_name = schema_name_ptr->c_ptr_safe();
+  const char *tbl_name = table_name_ptr->c_ptr_safe();
 
   // Check if table is hidden.
   THD *thd = current_thd;
   if (is_hidden_by_ndb(thd, schema_name_ptr, table_name_ptr))
     DBUG_RETURN(false);
 
+  if ((strcmp(sch_name, "mysql") == 0 &&
```
laurynas-biveinis: Why this needs special treatment?

sql/parse_tree_column_attrs.h
```diff
@@ -302,10 +302,6 @@ class PT_column_format_column_attr : public PT_column_attr_base {
     *type_flags |= format << FIELD_FLAGS_COLUMN_FORMAT;
   }
   virtual bool contextualize(Column_parse_context *pc) {
```
laurynas-biveinis, line 304: Why this change?

sql/sql_partition.cc
```diff
@@ -4094,31 +4095,31 @@ bool get_first_partition_name(THD *thd, Partition_handler *part_handler,
   sql_digest_state *parent_digest = thd->m_digest;
   PSI_statement_locker *parent_locker = thd->m_statement_psi;
 
-  Parser_state parser_state;
+  Partition_expr_parser_state parser_state;
```
laurynas-biveinis, line 4098: Why the changes in this file?

sql/sql_table.cc
```diff
@@ -6287,15 +6293,40 @@ bool mysql_prepare_create_table(
...
+    */
+
+    if (sql_field->column_format() == COLUMN_FORMAT_TYPE_COMPRESSED &&
+        sql_field->zip_dict_name.str != nullptr &&
+        sql_field->zip_dict_name.length != 0) {
+      if (compression_dict::acquire_dict_mdl(thd, MDL_SHARED_READ)) {
+        DBUG_RETURN(true);
+      }
+
+      uint64 zip_dict_id =
```
laurynas-biveinis, line 6319: const auto (and if not, then const uint64_t: with C++11 I don't think there is any need for custom types that merely rename standard types)

sql/sql_yacc.yy
```diff
@@ -2800,7 +2800,7 @@ create:
             Lex->create_info= YYTHD->alloc_typed<HA_CREATE_INFO>();
             if (Lex->create_info == nullptr)
               MYSQL_YYABORT; // OOM
-            Lex->create_info->options= $3;
+            Lex->create_info->options= $3 ? HA_LEX_CREATE_IF_NOT_EXISTS : 0;
```
laurynas-biveinis, line 2803: Why this change?

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,226 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+delete)
+@param[in,out] thd  Session object
+@return TABLE* on success else nullptr */
+static TABLE *open_dictionary_table_write(THD *thd) {
+  TABLE_LIST tablelst;
+  tablelst.init_one_table(C_STRING_WITH_LEN(COMPRESSION_DICTIONARY_DB),
+                          C_STRING_WITH_LEN(COMPRESSION_DICTIONARY_TABLE),
+                          COMPRESSION_DICTIONARY_TABLE, TL_WRITE,
+                          MDL_SHARED_NO_READ_WRITE);
+  tablelst.next_local = tablelst.next_global = 0;
```
laurynas-biveinis: nullptr ?

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,226 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+@return TABLE* on success else nullptr */
+static TABLE *open_dictionary_table_read(THD *thd) {
+  DBUG_ASSERT(!thd->is_attachable_ro_transaction_active());
+  thd->begin_attachable_ro_transaction();
+
+  TABLE_LIST tablelst;
+  tablelst.init_one_table(C_STRING_WITH_LEN(COMPRESSION_DICTIONARY_DB),
+                          C_STRING_WITH_LEN(COMPRESSION_DICTIONARY_TABLE),
+                          COMPRESSION_DICTIONARY_TABLE, TL_READ);
+  tablelst.next_local = tablelst.next_global = 0;
```
laurynas-biveinis: nullptr?

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,226 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+@param[in,out] thd Session object
+@return on success, a TABLE* object else nullptr on failure */
+
+static TABLE *open_dictionary_cols_table_write(THD *thd) {
+  TABLE_LIST tablelst;
+  tablelst.init_one_table(C_STRING_WITH_LEN(COMPRESSION_DICTIONARY_COLS_DB),
+                          C_STRING_WITH_LEN(COMPRESSION_DICTIONARY_COLS_TABLE),
+                          COMPRESSION_DICTIONARY_COLS_TABLE,
+                          TL_WRITE_CONCURRENT_DEFAULT, MDL_SHARED_WRITE);
+  tablelst.next_local = tablelst.next_global = 0;
```
laurynas-biveinis: nullptr?

sql/sql_zip_dict.cc
```diff
@@ -50,58 +249,121 @@ Creates a new compression dictionary with the specified data.
...
+    DBUG_RETURN(error);
+  }
+
+  table->use_all_columns();
+
+  sql_mode_t saved_mode = thd->variables.sql_mode;
+  thd->variables.sql_mode = 0;
+
+  TABLE_SHARE *ts = table->s;
+  if (table->next_number_field == NULL && ts->found_next_number_field) {
```
laurynas-biveinis: nullptr

sql/sql_zip_dict.cc
```diff
@@ -140,64 +425,372 @@ Deletes a compression dictionary.
...
+  thd->mdl_context.release_transactional_locks();
   DBUG_RETURN(error);
 }
+
+/* Get dict_id for a given dict_name from mysql.compression_dictionary table
+An attachable RO trx is used to do the read
+@param[in,out]  thd             Session context
+@param[in]      zip_dict_name   dictionary entry name
+@return 0 for failure, else dictionary id */
+uint64 get_id_for_name(THD *thd, const LEX_CSTRING *zip_dict_name) {
```
laurynas-biveinis: uint64_t, const LEX_CSTRING &

sql/sql_zip_dict.cc
```diff
@@ -140,64 +425,372 @@ Deletes a compression dictionary.
...
+
+  uchar user_key[MAX_KEY_LENGTH];
+  KEY *name_idx_key = table->key_info + COMPRESSION_DICTIONARY_NAME_INDEX;
+  key_copy(user_key, table->record[0], name_idx_key, name_idx_key->key_length);
+
+  int ret = table->file->ha_index_read_idx_map(
+      table->record[0], COMPRESSION_DICTIONARY_NAME_INDEX, user_key,
+      HA_WHOLE_KEY, HA_READ_KEY_EXACT);
+
+  uint64 zip_dict_id = 0;
```
laurynas-biveinis: const auto zip_dict_id = (ret == 0) ? table->field... : 0;

sql/dd/impl/bootstrapper.cc
```diff
@@ -2214,6 +2278,24 @@ bool initialize_dictionary(THD *thd, bool is_dd_upgrade_57,
...
   }
 
+  // Create the mysql.compression_dictionary_table
+  if (execute_query(thd, dd::String_type(compression_dict::dict_table_str))) {
+    return true;
+  }
+
+  if (execute_query(
+          thd, dd::String_type(
+                   "CREATE TABLE IF NOT EXISTS "
```
laurynas-biveinis: Other SQL is in named variables?

sql/sql_zip_dict.h
```diff
@@ -19,9 +19,35 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
 #include "my_inttypes.h"
+#include "sql/dd/types/table.h"  // dd::enum_column_types
+#include "sql/mdl.h"
 
 class THD;
+struct TABLE_SHARE;
+
+namespace compression_dict {
+
+const char *const dict_table_str =
```
laurynas-biveinis: Defining (not declaring) in header directly, doesn't this result in multiple definitions per including source file - put extern declarations here, and definitions in .cc?

storage/innobase/dict/dict0dict.cc
```diff
@@ -7408,38 +7384,42 @@ dict_create_zip_dict(const char *name, ulint name_len, const char *data,
...
-		rw_lock_s_lock(dict_operation_lock);
-	}
-	if (!dict_locked) {
-		dict_mutex_enter_for_mysql();
-	}
+  if (!dict_operation_locked) {
+    rw_lock_s_lock(dict_operation_lock);
+  }
+  if (!dict_locked) {
+    //	dict_mutex_enter_for_mysql();
```
laurynas-biveinis: Debugging leftover?

storage/innobase/dict/dict0dict.cc
```diff
@@ -7408,38 +7384,42 @@ dict_create_zip_dict(const char *name, ulint name_len, const char *data,
...
+  trx_free_for_background(trx);
 
-	if (!dict_locked) {
-		dict_mutex_exit_for_mysql();
-	}
-	if (!dict_operation_locked) {
-		rw_lock_s_unlock(dict_operation_lock);
-	}
+  if (!dict_locked) {
+    //	dict_mutex_exit_for_mysql();
```
laurynas-biveinis: Leftover?

storage/innobase/dict/dict0dict.cc
```diff
@@ -7453,84 +7433,42 @@ Must be freed with mem_free().
...
+                  dict_operation_locked = true;);
 
-	const trx_t *trx = trx_allocate_for_background();
-	trx->op_info = "delete zip_dict";
-	trx->dict_operation_lock_mode = RW_X_LATCH;
-	trx_start_if_not_started(trx, true);
+  if (!dict_operation_locked) {
+    rw_lock_s_lock(dict_operation_lock);
+  }
+  if (!dict_locked) {
```
laurynas-biveinis: Debugging leftover?

storage/innobase/dict/dict0dict.cc
```diff
@@ -7453,84 +7433,42 @@ Must be freed with mem_free().
...
-	trx->op_info = "";
-	trx->dict_operation_lock_mode = 0;
-	trx_free_for_background(trx);
+  trx_commit_for_mysql(trx);
+  trx->dict_operation_lock_mode = 0;
+  trx_free_for_background(trx);
 
-	dict_mutex_exit_for_mysql();
-	rw_lock_x_unlock(dict_operation_lock);
+  if (!dict_locked) {
```
laurynas-biveinis: Debugging leftover?

storage/innobase/dict/dict0upgrade.cc
```diff
@@ -287,9 +287,13 @@ static bool dd_upgrade_match_single_col(const Field *field,
 
   ulint is_virtual = (innobase_is_v_fld(field)) ? DATA_VIRTUAL : 0;
 
+  ulint compressed = field->column_format() == COLUMN_FORMAT_TYPE_COMPRESSED
```
laurynas-biveinis: const ulint

storage/innobase/dict/dict0upgrade.cc
```diff
@@ -1362,3 +1371,64 @@ int dd_upgrade_finish(THD *thd, bool failed_upgrade) {
 
   DBUG_RETURN(0);
 }
+
+bool dd_upgrade_get_compression_dict_data(
+    THD *thd, compression_dict_data_vec_t &data_vector) {
+  ut_ad(srv_is_upgrade_mode);
+
+  btr_pcur_t pcur;
```
laurynas-biveinis: Declare variables at first use

laurynas-biveinis (changes requested):

mysql-test/suite/innodb/r/xtradb_compressed_columns_alter_table.result
satya-bodapati: With old I_S views, we checked for table_id which changes with alter. So after alter, the reference to old table_id doesn't exist.

With new I_S views, the table_id cannot be exposed. So we just check for name and so reference exists even after ALTER is completed.

mysql-test/suite/innodb/r/xtradb_compressed_columns_consistency_antelope_compact_set1.result
satya-bodapati: Right, the barracuda test can be removed and the reference to "antelope" can be removed as well.

satya-bodapati: Just realized that they are different because of row_formats. So I ll keep the tests and remove the names antelope and barracuda

satya-bodapati: compact and redundant, there are duplicates tests. these can be removed.

mysql-test/suite/innodb/t/xtradb_compressed_columns_with_dictionaries.test
percona-ysorokin: @satya-bodapati The main concern here is the user experience.
Even though, according to the documentation, it is OK to call `deflateSetDictionary()` with large (longer than windowSize) data passed as a dictionary, just silently ignoring the part of  this dictionary may not be very user-friendly and may rise a lot of questions. What is even more confusing is that in case of exceeding this 32506 limit, not the bytes at the end but the ones at the beginning will be ignored.

So, I suggest, as of right now we only have single compression algorithm (zlib) available for compression columns, we leave this 32506 compression dictionary size limit as is. In other words, calling `CREATE COMPRESSION_DICTIONARY()` with data longer than 32506 bytes should return en error.

However, I admit that in future we may have additional compression algorithms (like lz4, zstd, etc.) for which this limitation can be different. So, I suggest, we review/remove this limitation then.

sql/dd/impl/system_views/compression_dictionary.cc
percona-ysorokin: This only makes difference if `Schemata` has overloaded operator `new`. Otherwise, statically allocated object is better from the memory leaks detection point of view.

mysql-test/suite/innodb/r/xtradb_compressed_columns_consistency_antelope_compact_set1.result
satya-bodapati: It doesn't work. It is caused by the use of "NATIONAL" keyword for VARCHARs. The warning suggests to avoid NATIONAL.

mysql-test/suite/innodb/t/xtradb_compressed_columns_partitioning.test
satya-bodapati: I didn't get what you mean. Since the MyISAM engine is removed, 2 became 1 and we have now 3 iterations instead of 6. I could have removed the loop and just do the 3 iterations manually but doing this was simpler

mysql-test/t/dd_upgrade_compression_dict.test
satya-bodapati: these are from existing upgrade test :) fixed it

mysql-test/t/dd_upgrade_compression_dict-master.opt
satya-bodapati: No,  have_innodb_16k/4k/8k etc are removed. If not, they should be removed in our branch too.

--bootstrap initializes with the required page size.

The requirement comes from 5.7 datadir which is 16k. So we need to have running server which is 16k (the page size is set in mtr restart options somewhere).

Only 16k can be tested for upgrade now. Same with other upgrade tests

mysql-test/t/dd_upgrade_test.test
satya-bodapati: It has been fixed in 8.0.13

satya-bodapati: in upstream

sql/dd/impl/bootstrapper.cc
satya-bodapati: Copy paste error from repopulate_charsets_and_collations() :) fixed. It is not relevant for the new code.

sql/dd/impl/dictionary_impl.cc
satya-bodapati: Added comment

sql/field.h
satya-bodapati: Generally I prefer forward declarations to avoid cyclic includes.

sql/field.h
satya-bodapati: When the id is retrieved from DD, it is dd::Object_id which is 
typedef unsigned long long Object_id;

 There might be compilation warnings because of this change.

sql/item_func.cc
satya-bodapati: To hide compression dictionary tables from "SHOW TABLES". This is needed because mysqlcheck tries to access tables listed by "SHOW TABLES" and when it tries to do CHECK TABLE mysql.compression_dictionary table, it gets "access denied error".

So in short, it should be treated as hardcoded dictionary tables and hide them from SHOW TABLES.

sql/parse_tree_column_attrs.h
satya-bodapati: WIthout this change, COMPRESSED and STORED is not allowed.

sql/sql_partition.cc
satya-bodapati: The string to be parsed is " PARTITION BY .. " 

With Parser_state, parsing fails because it is not valid start token.

With Partition_expr_parser, the above string is parsed correctly and partition info is extracted.

Partition_expr_parser_state is used elsewhere too. See mysql_unpack_partition()

sql/sql_table.cc
satya-bodapati: uint64 is used because of dd::Object_id. Leaving this open, will check if I can use dd::Object_id.

sql/sql_yacc.yy
satya-bodapati: Without this change, COMPRESSION DICTIONARY  create if not exists, doesn't work.

sql/sql_zip_dict.h
satya-bodapati: It doesn't result in multiple definition error because of const * ptr const has 'internal' linkage.

```
    [C++11: 3.5/3]: A name having namespace scope (3.3.6) has internal linkage if it is the name of

        a variable, function or function template that is explicitly declared static; or,
        a variable that is explicitly declared const or constexpr and neither explicitly declared extern nor previously declared to have external linkage; or
        a data member of an anonymous union.
```
I generally do what you have said. For a change, I tried different thing which is also correct.

if you prefer deinition in .cc file, let me know.

mysql-test/suite/innodb/t/xtradb_compressed_columns_partitioning.test
laurynas-biveinis: Please see if you can address the 2nd half of my comment? What I tried to say is that after removing one of the iteration variables, the cleanup is not complete, there is further "constant propagation" to do

mysql-test/t/dd_upgrade_compression_dict-master.opt
laurynas-biveinis: https://github.com/mysql/mysql-server/blob/8.0/mysql-test/include/have_innodb_16k.inc present in upstream, since you know the circumstances better, can you please file a bug report with them?

sql/field.h
laurynas-biveinis: I see, but will this it get diagnosed if the declarations start mismatching subtly?

sql/item_func.cc
laurynas-biveinis: Please add a comment that these need to be hidden explicitly like DD tables while they are not technically DD tables

sql/sql_zip_dict.h
laurynas-biveinis: If I understand correctly, internal linkage here will result in one copy per including source file?

sql/item_func.cc
satya-bodapati: I didn't write this but I can remove

sql/dd/impl/bootstrapper.cc
satya-bodapati: Fixed. comment is wrong. We should return error here

mysql-test/suite/innodb/t/xtradb_compressed_columns_with_dictionaries.test
satya-bodapati: Ok @percona-ysorokin I will limit the dictionary size

sql/sql_zip_dict.cc
satya-bodapati: ok, and why do we prefer the runtime evaluated type?  for debugging?

satya-bodapati:
Addressed most review comments, work in progress for some

sql/dd/impl/system_views/compression_dictionary.cc
satya-bodapati: Ok, Let me check how the object is freed in current way. I'm fine with either way.

sql/sql_zip_dict.cc
satya-bodapati: I think there is little value with allocator in this case because server is in upgrade mode. connections are not accepted at this stage. We cannot see the PFS data either.

mysql-test/suite/innodb/t/xtradb_compressed_columns_partitioning.test
satya-bodapati: Ok, fixing the other part

mysql-test/t/dd_upgrade_compression_dict-master.opt
satya-bodapati: Ok, will check the tests using this .inc file and file a bug

satya-bodapati: Found the regression commit.

commit 66bba1d591b847128d7e176e89fcd880b10ce6c2
Author: Bernt M. Johnsen <bernt.johnsen@oracle.com>
Date:   Mon Mar 6 14:17:58 2017 +0100

    WL#7554 Post push fix. Run innodb_zip.bug36169 with 16k innodb page size only.

The right fix is to just bootstrap with 16k.

cd mysql-test/include
ls have_innodb_*
have_innodb_16k.inc  have_innodb_max_16k.inc  have_innodb_min_8k.inc

The  max  .inc files are fine.

sql/handler.h
satya-bodapati: Looks like the clang-format is doing this for me. Have to check why this isn't the original format

mysql-test/suite/innodb/t/xtradb_compressed_columns_with_dictionaries.test
satya-bodapati: Fixed now

sql/dd/impl/system_views/compression_dictionary.cc
satya-bodapati: There is no memory leak and the free is done by destructor

#0  __GI___libc_free (mem=0x55555bc27420) at malloc.c:3086
#1  0x00005555591036fc in dd::system_views::Schemata::~Schemata (this=0x55555bc27420, __in_chrg=<optimized out>, __vtt_parm=<optimized out>)
    at /home/satya/WORK/PS-8.0-4805/sql/dd/impl/system_views/schemata.h:37
#2  0x0000555558fa3122 in dd::Entity_element<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const, std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const>, dd::system_views::System_view const, dd::System_views::Types, &dd::System_views::type_name, true>::~Entity_element (this=0x55555bc279b0, 
    __in_chrg=<optimized out>) at /home/satya/WORK/PS-8.0-4805/sql/dd/impl/system_registry.h:76
#3  0x0000555558fa1c8a in dd::Entity_registry<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const, std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const>, dd::system_views::System_view const, dd::System_views::Types, &dd::System_views::type_name, true>::~Entity_registry (
    this=0x55555b9a0da0 <dd::System_views::instance()::s_instance>, __in_chrg=<optimized out>)
    at /home/satya/WORK/PS-8.0-4805/sql/dd/impl/system_registry.h:147
#4  0x0000555558fa0d24 in dd::System_views::~System_views (this=0x55555b9a0da0 <dd::System_views::instance()::s_instance>, __in_chrg=<optimized out>)
    at /home/satya/WORK/PS-8.0-4805/sql/dd/impl/system_registry.h:457
#5  0x00007ffff5db5041 in __run_exit_handlers (status=0, listp=0x7ffff615d718 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true, 
    run_dtors=run_dtors@entry=true) at exit.c:108
#6  0x00007ffff5db513a in __GI_exit (status=<optimized out>) at exit.c:139
#7  0x0000555557e45611 in mysqld_exit (exit_code=0) at /home/satya/WORK/PS-8.0-4805/sql/mysqld.cc:1919
#8  0x0000555557e55b0e in mysqld_main (argc=66, argv=0x55555ba6fb88) at /home/satya/WORK/PS-8.0-4805/sql/mysqld.cc:6721
#9  0x0000555557c776ca in main (argc=8, argv=0x7fffffff5d78) at /home/satya/WORK/PS-8.0-4805/sql/main.cc:30

satya-bodapati: Since there is no correctness issue, I will keep upstream style

sql/handler.h
satya-bodapati: I think there is little value with allocator in this case because server is in upgrade mode. connections are not accepted at this stage. We cannot see the PFS data either.

mysql-test/t/dd_upgrade_compression_dict-master.opt
satya-bodapati: Reported upstream bug https://bugs.mysql.com/bug.php?id=93304

sql/sql_zip_dict.cc
laurynas-biveinis: If you need to take its address

satya-bodapati: Ok, changing to constexpr

sql/sql_zip_dict.h
satya-bodapati: That is right.
I've made the requested changes though (to be consistent with existing style :))

sql/field.h
satya-bodapati: it should. like any other extern usage. and I tried to incude sql_lex.h in field.h I get these compilation errors. Looks like there is header dependency.

[ 36%] Building CXX object storage/perfschema/CMakeFiles/perfschema.dir/pfs.cc.o
In file included from /home/satya/WORK/PS-8.0-4805/sql/sql_lex.h:66:0,
                 from /home/satya/WORK/PS-8.0-4805/sql/field.h:59,
                 from /home/satya/WORK/PS-8.0-4805/storage/csv/ha_tina.cc:63:
/home/satya/WORK/PS-8.0-4805/sql/item.h:162:3: error: ‘Derivation’ does not name a type; did you mean ‘DTCollation’?
   Derivation derivation{DERIVATION_NONE};
   ^~~~~~~~~~
   DTCollation

sql/handler.h
satya-bodapati: Fixed by updating to latest 8.0 branch

satya-bodapati:
Ready for review again. (only pending comment is about uint64 to uint64_t change - wip)

satya-bodapati:
Added one new test dd_upgrade_compression_dict_80.test which simulates upgrade from mysql datadir.
Code changes done to achive the above test in debug builds.

sql/dd/impl/dictionary_impl.cc
```diff
@@ -336,12 +337,15 @@ bool Dictionary_impl::is_dd_table_access_allowed(bool is_dd_internal_thread,
 
   /*
     Access allowed for external DD tables, for DML on protected DDSE tables,
-    and for any operation on SYSTEM tables.
+    and for any operation on SYSTEM tables. Compression dictionary tables
+    are created as SYSTEM type but we don't allow direct access to them.
+    User should use I_S views on compression dictinoary tables
```
laurynas-biveinis: "dictionary"

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,283 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
 #include "sql/sql_class.h"  // THD
 #include "sql/sql_table.h"  // write_bin_log
+#include "sql/table.h"
+#include "sql/thd_raii.h"
+#include "sql/transaction.h"
+
+namespace compression_dict {
+
+/** Fields of mysql.compression_dictionary table */
+static constexpr uint32_t COMPRESSION_DICTIONARY_FIELD_ID = 0;
```
laurynas-biveinis:
```suggestion
static const constexpr uint32_t COMPRESSION_DICTIONARY_FIELD_ID = 0;
```

laurynas-biveinis: etc

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,283 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+static constexpr uint32_t COMPRESSION_DICTIONARY_FIELD_VERSION = 1;
+static constexpr uint32_t COMPRESSION_DICTIONARY_FIELD_NAME = 2;
+static constexpr uint32_t COMPRESSION_DICTIONARY_FIELD_DATA = 3;
+
+/** Index of compression_dictionary table */
+static constexpr uint32_t COMPRESSION_DICTIONARY_NAME_INDEX = 1;
+static constexpr uint32_t COMPRESSION_DICTIONARY_VERSION = 1;
+
+static const char *COMPRESSION_DICTIONARY_DB = "mysql";
```
laurynas-biveinis:
```suggestion
static const constexpr char *COMPRESSION_DICTIONARY_DB = "mysql";
```
etc

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,283 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+
+static const char *COMPRESSION_DICTIONARY_COLS_DB = "mysql";
+static const char *COMPRESSION_DICTIONARY_COLS_TABLE =
+    "compression_dictionary_cols";
+
+/** Max window size (2^15) minus 262 */
+static constexpr uint32_t MAX_DATA_LENGTH = 32506;
+
+/** Schema definition of mysql.compression_dictionary */
+const dd::String_type dict_table_str =
```
laurynas-biveinis: These do not seem to be exported?
```suggestion
static const dd::String_type dict_table_str =
```
etc

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,283 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+/** During upgrade from 5.7 to 8.0, transfer compression dicitonary
+data from 5.7 SYS_ZIP_DICT to 8.0 mysql.compression_dictionary table
+@param[in]   thd   Session context
+@return false on success, true on failure */
+bool upgrade_transfer_compression_dict_data(THD *thd) {
+  compression_dict_data_vec_t zip_dict_vec;
+  handlerton *hton = ha_resolve_by_legacy_type(thd, DB_TYPE_INNODB);
+  hton->upgrade_get_compression_dict_data(thd, zip_dict_vec);
+
+  for (auto &elem : zip_dict_vec) {
```
laurynas-biveinis:
```suggestion
  for (const auto &elem : zip_dict_vec) {
```

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,283 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+    std::string &name = elem.first;
```
laurynas-biveinis:
```suggestion
    const auto &name = elem.first;
```

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,283 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+    std::string &data = elem.second;
```
laurynas-biveinis:
```suggestion
    const auto &data = elem.second;
```

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,283 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+
+    int ret = create_zip_dict(thd, name.c_str(), name.length(), data.c_str(),
```
laurynas-biveinis, line 131: Depending on the other callers, maybe create_zip_dict can take const std::string &-s now?

sql/sql_zip_dict.cc
```diff
@@ -20,27 +20,283 @@ this program; if not, write to the Free Software Foundation, Inc.,
...
+
+  return (false);
+}
+
+/** Open mysql.compression_dictionary table for modification (insert, update,
+delete)
+@param[in,out] thd  Session object
+@return TABLE* on success else nullptr */
+static TABLE *open_dictionary_table_write(THD *thd) {
+  TABLE_LIST tablelst;
```
laurynas-biveinis: why missing "i" :)

laurynas-biveinis (changes requested):
Minor comments only, lighter testing should be enough even if addressed in full amount

sql/sql_zip_dict.cc
satya-bodapati: one of the caller is from parser, and the pointers are from LEX_CSTR. So cannot change it.

sql/sql_zip_dict.cc
satya-bodapati: Table opening code is taken from other place

satya-bodapati: Added "i" ;)

satya-bodapati:
https://ps.cd.percona.com/job/percona-server-8.0-param/26/
(Not sure when it will complete.. ;) )

laurynas-biveinis (approved):

satya-bodapati:
Fix 
1. dd_upgrade_compression_dict.test (mask DATADIR in result file)
2. record funcs suite tests result content mismatches due to the new I_S views on compression dictionary tables.
3. remove the non native partitioning test that tests upgrade from 5.6 to 5.7. Not applicable anymore
    (xtradb_compressed_columns_upgrade & xtradb_compressed_columns_upgrade_partitioning)
    also removed .ibd & frm files in std_data. Used only by these tests
4. Fix I_S tests in main suite
5. Record innodb-index-online-fk test. SELECT * FROM INFORMATION_SCHEMA.INNODB_TABLES shows compression dictionary tables. It should have been hidden though. Will be handled as separate JIRA issue

satya-bodapati:
Reported https://jira.percona.com/browse/PS-5074 for the point 5. This will be handled post merge to trunk.

percona-ysorokin (approved):
LGTM

PR #2629, opened 2018-10-30

laurynas-biveinis:
Convert expected errors in the TokuDB MTR tests from numbers to
mnemonics. While the main motivation for this was client-side error
support introduced in 8.0, the vast majority of converted errors are
in server ones.

https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/89/ - there are some errors which seem to be unrelated


george-lorch (approved):
This is awesome, thanks for fixing type_temporal

PR #2635, opened 2018-10-31
PS-4976: use correct log level when logging undo encryption startup

dutow:


laurynas-biveinis (approved):
Borderline obvious PR, which can be self-approved

PR #2640, opened 2018-10-31
PS-4918 : debug assertion on 8.0 upgrade of TokuDB tables created wit…

george-lorch:
…h explicit ROW_FORMAT

- On table upgrade and conversion from .frm to DD, detect use of TokuDB
  ROW_FORMAT specifiers and reset table row_format to default.

- Refactored mtr upgrade testing from native partitioning to be a bit more
  generic and usable for other type of simple upgrade scenario testing.

- Fixed up TokuDB and RocksDB partition testing to use newly refactored/generic
  upgrade test components, re-recorded for minor differences in result.

- Implemented TokuDB row format upgrade test and added upgrade seed datadir.added upgrade


george-lorch:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/93/

laurynas-biveinis (approved):

PR #2641, opened 2018-10-31
PS-4951: Many libc-related Valgrind errors on CentOS7 (8.0)

inikep:
The issue concerns GNU libc version 2.17 when linux hosts doesn't have IPv6 configured.
This is because libc points the cache (that is freed on exit) to a static variable if there is no ipv6 address, which later results in freeing an invalid pointer.
This patch adds an entry in `valgrind.supp` to solve this issue.

### DIFF
--- a/mysql-test/valgrind.supp
+++ b/mysql-test/valgrind.supp
@@ -2108,3 +2108,18 @@
    fun:ldap_int_tls_start
    fun:ldap_start_tls_s
 }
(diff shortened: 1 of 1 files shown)


inikep:
Tested (only `group_replication.gr_group_name_validation_on_boot`) at:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-valgrind-param/15/

laurynas-biveinis (approved):

PR #2643, opened 2018-11-01
PS-4988: Fix uninitialized btr stat mutex

dutow:


robgolebiowski (approved):

mysql-test/t/percona_bug_ps4988.test
```diff
@@ -0,0 +1,5 @@
+# PS-4988: Uninitailized btr stat mutex caused assertion failure
```
laurynas-biveinis: "Uninitialized"

laurynas-biveinis: Consider naming the test "percona_sync_debug", so that it can receive further similar tests in the future

laurynas-biveinis (changes requested):

dutow:
Updated: https://ps.cd.percona.com/job/percona-server-5.7-param/15/

PR #2644, opened 2018-11-01
PS-4989: Fixing innodb_track_changed_pages debug validation (5.7)

dutow:
In debug builds, this setting is allowed to be turned off temporarily after it was
turned on during startup. Howewer memory garbage also caused it to be accidentally
turned on when it was disabled at startup.


mysql-test/suite/innodb/t/percona_changed_page_bmp_1368530.test
```diff
@@ -6,6 +6,15 @@
```
laurynas-biveinis, line 6: The testcase no longer is "1368530", I'd rename to percona_changed_page_bmp_debug_tracking_off ?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -20636,14 +20636,17 @@ innodb_track_changed_pages_validate(
 		return 1;
 	}
 
+
```
laurynas-biveinis: Spurious

laurynas-biveinis (changes requested):
Are lower versions affected?

dutow:
Testcase renamed, submitted 5.5 and 5.6 PRs

laurynas-biveinis (changes requested):
Same comment as for the 5.5 version

laurynas-biveinis:
@dutow, please submit an 8.0 version too

PR #2654, opened 2018-11-02
PS-4790: Improve user statistics accuracy (5.5)

inikep:
Currently userstat measures times for mysql_parse, mysqld_stmt_reset, mysqld_stmt_fetch, mysqld_stmt_prepare, mysqld_stmt_execute.
MariaDB measures times only for `mysql_parse` and `dispatch_command` which call above functions.
Moreover MariaDB adds measurements for COM_INIT_DB, COM_REGISTER_SLAVE, COM_RESET_CONNECTION, and other in `dispatch_command`.
It makes measurements of BUSY_TIME and CPU_TIME more accurate.
This patch implements a similar solution that measures times only for `dispatch_command` and `mysql_parse` using newly created `sql/userstat.cc` and `sql/userstat.h`.


inikep:
Testing started at https://ps.cd.percona.com/view/5.5/job/percona-server-5.5-param/35/

laurynas-biveinis (approved):

PR #2658, opened 2018-11-05
PS-4943 : Using XA transactions with temporary TokuDB table could lea…

george-lorch:
…d to a server crash

- This actually has nothing to do with either XA or temp tables.  It is due to
  the memory management around the string sysvar tokudb_last_lock_timeout which
  is not an 'input' variable but is used to relay textual information about the
  last lock wait timeout to the user.  Unfortunately, this is a mis-use/abuse of
  the MySQL SYAVAR functionality and is incompatible with the expected plugin
  memory behavior.  Changed this option to remove the command line option and
  made it read-only to prevent further problems.

### DIFF
--- a/storage/tokudb/tokudb_sysvars.cc
+++ b/storage/tokudb/tokudb_sysvars.cc
@@ -662,13 +662,13 @@ static MYSQL_THDVAR_ULONGLONG(
     ~0ULL,
     1);
 
-static MYSQL_THDVAR_STR(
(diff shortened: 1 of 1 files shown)


george-lorch:
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/30/

laurynas-biveinis (approved):
Please update the JIRA issue title/summary to reflect the actual issue

PR #2662, opened 2018-11-06
PS-4971 : Setting utf8mb4 as a key in MyRocks table could lead to a s…

george-lorch:
…erver crash

- 8.0 changes the behavior of the character set strnxfrm methods in a way that
  alters what was a once 'defined' behavior in 5.7 to a now 'undefined'
  behavior.  Quoted right from m_ctype.h "@param num_codepoints Treat the string
  as if it were of type CHAR(num_codepoints)." ... "If the string has more than
  "num_codepoints" codepoints, behavior is undefined; may truncate, may crash,
  or do something else entirely."  The 5.7 behavior was to truncate the result
  to the num_codepoints length and is what the MyRocks packing code expected.
  This patch alters MyRocks to now tell strnxfrm what the srclen is based on the
  lesser of Field_varstring codepoint size or trimmed value length rather than
  always passing it the trimmed value length of the incoming value.

### DIFF
--- a/mysql-test/suite/rocksdb/t/type_varchar.test
+++ b/mysql-test/suite/rocksdb/t/type_varchar.test
@@ -80,3 +80,24 @@ check table t;
 alter table t modify h varchar(31) character set cp1257 collate cp1257_bin not null;
 check table t;
 drop table t;
+
+--echo # PS-4971 : Setting utf8mb4 as a key in MyRocks table could lead to a server crash
+# This sequence would overwrite memory, eventually assert at the end of
+# Rdb_key_def::pack_record and, if successful, would leave a badly malformed and
+# stored secondary key entry.
+create table t1 (
+  str varchar(255) character set utf8mb4,
+  key str (str(2))) default character set utf8mb4 engine=rocksdb;
+insert INTO t1 values("1111111111111111111111111111111111111111111111111111");
+drop table t1;
+
+# With a partial fix, this sequence would find the row in the secondary key but
+# would fail to unpack the t_vers value from the secondary key with a data
+# coruption error.
+create table t1(
+  t_vers varchar(4) not null,
+  key ix_5 (t_vers)
+) charset latin1 engine=rocksdb;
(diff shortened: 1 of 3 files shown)


george-lorch:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/95/

mysql-test/suite/rocksdb/t/type_varchar.test
```diff
@@ -80,3 +80,24 @@ check table t;
 alter table t modify h varchar(31) character set cp1257 collate cp1257_bin not null;
 check table t;
 drop table t;
+
+--echo # PS-4971 : Setting utf8mb4 as a key in MyRocks table could lead to a server crash
+# This sequence would overwrite memory, eventually assert at the end of
+# Rdb_key_def::pack_record and, if successful, would leave a badly malformed and
+# stored secondary key entry.
+eval create table t1 (
```
laurynas-biveinis: Redundant eval

storage/rocksdb/rdb_datadic.cc
```diff
@@ -2764,17 +2764,24 @@ void Rdb_key_def::pack_with_varchar_space_pad(
   const CHARSET_INFO *const charset = field->charset();
   const auto field_var = static_cast<Field_varstring *>(field);
 
+  const uchar *src =
```
laurynas-biveinis: It looks like src is used as const uchar * in one location, and as const char * in four, probably makes sense to swap?

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb/t/type_varchar.test
george-lorch: Oops, yup, cut-n-paste fat finger

george-lorch:
Changes made and locally tested OK, not firing jenkins yet because it still may be shaky, but really probably isn't totally necessary based on the changes.

laurynas-biveinis (approved):

PR #2664, opened 2018-11-07
PS-3829: Post push fix. Fixed zipped page compression encryption by r…

robgolebiowski:
…emoving assigning key_version to pages in buffers. Key version should not be updated in buffers as there are no guarantees that the page's header being flushed is the only copy of the header in the buffer. This true for instance for zipped pages. Also with parallel double write buffer pages are copied so the os0file layer will update only the key version in the copy of the header.


robgolebiowski:
self approved

PR #2665, opened 2018-11-07
PS-4796 : 8.0 DD and atomic DDL breaks DROP DATABASE for engines that…

george-lorch:
… store files in database directory

- During a DROP DATABASE, sql/sql_db.cc::find_unknown_and_remove_deletable_files
  calls ha_known_exts which ends up enumerating all handlertons and accessing
  their hton->file_extensions.  This should contain all file extensions that
  each the engine creates.  Because TokuDB doesn't currently provide/populate
  file_extensions in its handlerton, the DROP DATABASE fails before it even
  starts because it doesn't pass this validation step.

- This patch implements/adds the ".tokudb" file extension to the TokuDB
  handlerton which allows the DROP DATABASE pre-check to recognize that the
  TokuDB files will be properly removed when the tables within the database
  are dropped.

- 8.0 apparantly also fixed an issue that is present in 5.6 and 5.7 and that is
  that a user may rename a table into a database that does not actually exist.
  Since this is no longer permitted in 8.0, the TokuDB test for proper behavior
  of dir_per_db in this situation is no longer needed and has been removed.

- Fixing this issue, allowed several tests to progress and now fail due to
  various 8.0 change and as such needed re-recording for known differences
  (different default character sets, changes to common replication tests,
  etc...)


george-lorch:
No jenkins yet ad it seems misbehaving.  Local test passes with only expected failures:

> ./mtr --force --retry=0 --retry-failure=0 --max-test-fail=0 --testcase-timeout=120 --mysqld=--loose-tokudb-cache-size=512M --parallel=6 --suite=tokudb,tokudb.add_index,tokudb.alter_table,tokudb.bugs,tokudb.parts,tokudb.perfschema,tokudb.rpl,tokudb.sys_vars
> ...
> The servers were restarted 72 times
> Spent 21364.343 of 5784 seconds executing testcases
> 
> Completed: Failed 4/598 tests, 99.33% were successful.
> 
> Failing test(s): tokudb.bugs.xa-6 tokudb.parts.partition_mgm_lc1_tokudb tokudb.rpl.rpl_xa_interleave

storage/tokudb/hatoku_hton.cc
```diff
@@ -159,6 +159,9 @@ static char *tokudb_home;
 // static ulong tokudb_cache_parts = 1;
 const char *tokudb_hton_name = "TokuDB";
 
+// All TokuDB and PerconaFT exts that might appear in a database dir
+static const char *ha_tokudb_exts[]{".tokudb", NullS};
```
laurynas-biveinis: s/NullS/nullptr
(NullS expands to (char *)0, its presence in the current version looks redundant in the first place)

laurynas-biveinis (changes requested):

storage/tokudb/hatoku_hton.cc
george-lorch: Perhaps, but copied from InnoDB \</shrug\>

george-lorch:
Made minor change suggested, completed local mtr run without issue.  I'll go ahead and merge once Travis completes without issue.

laurynas-biveinis (approved):

PR #2667, opened 2018-11-08
PS-4979 : Dropping TokuDB table with non-alphanumeric characters coul…

george-lorch:
…d lead to a crash

- There seems to be cases where the server doesn't detect an invalid table name
  and somehow ends up passing in an empty table name during the create call. The
  engine expects a table name in the form of a relative path like
  "./databasename/tablename".  InnoDB detects this in parsing the table name
  during the creation and returns an error.  TokuDB does not.

- This commit adds an up-front parsing of the database and table names during
  creation and fails in the same way that InnoDB does if anything empty or
  unexpected is discovered.  Also added a new simple test that illustrates the
  detection of the empty table name and error.

### DIFF
--- a/mysql-test/suite/tokudb.bugs/r/PS-4979.result
+++ b/mysql-test/suite/tokudb.bugs/r/PS-4979.result
@@ -1,0 +1,2 @@
+CREATE TABLE `#mysql50#q.q`(f1 INT KEY) ENGINE=TOKUDB;
+ERROR HY000: Got error 1632 from storage engine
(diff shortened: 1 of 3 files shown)


george-lorch:
No jenkins yet because jenkins is still down, passes 100% of local tokudb mtr.

laurynas-biveinis (approved):
OK but looks like we paper over a server-side bug. Is there are a bug to report at least?

george-lorch:
I completely agree, but, this goes all the way back to 5.6 (probably even earlier), and, InnoDB has code to specifically check and handle this situation, thus the duplication of functionality (same MyRocks fix coming soon). My guess is someone on the server side has decided at some point that this is a feature.  There is no bug report that I could find.

laurynas-biveinis:
You are right on all counts, yet I'd still report it. It's easy for them to forget about such "features" if InnoDB had workaround for years

george-lorch:
It seems that this is fixed in 8.0.  The server side parsing error that lead to an empty table name being passed into the handler::create has been corrected.  Now the table name comes in exactly as specified.  So, there is no need for the fix to get ported forward from 5.7.

PR #2676, opened 2018-11-12
PS-4998: Valgrind: compilation fails with writing with no trivial copy-assignment (5.7)

inikep:
Fix compilation failure with Valgrind using `buf->stamp.bytes` instead of `buf` in buf_buddy_alloc_zip().

Upstream Bug URL:
https://bugs.mysql.com/bug.php?id=93105


laurynas-biveinis (changes requested):
Please reference the upstream bug in the commit message too
(and the upstream bug is being handled as 8.0-only, please let them know it's a 5.7 issue too)

inikep:
The commit message was updated and upstream was informed.

laurynas-biveinis (approved):

PR #2686, opened 2018-11-14
Merge remote-tracking branch 'local-5.6/ps-5.6-4529' into ps-5.7-4529

george-lorch:
(cherry picked from commit f4472e3cd8787cbc7ea7e62067fc114f6ab3e2c0)

### DIFF
--- a/mysql-test/suite/rocksdb/r/index_merge_rocksdb2.result
+++ b/mysql-test/suite/rocksdb/r/index_merge_rocksdb2.result
@@ -424,6 +424,7 @@ alter table t3 add keyC int not null, add index iC(keyC);
 update t3 set key9=key1,keyA=key1,keyB=key1,keyC=key1;
 SET @@GLOBAL.innodb_fast_shutdown = 0;
 # restart
+SET SESSION DEFAULT_STORAGE_ENGINE = RocksDB;
 explain select * from t3 where
 key1=1 or key2=2 or key3=3 or key4=4 or
 key5=5 or key6=6 or key7=7 or key8=8 or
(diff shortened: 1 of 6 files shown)


george-lorch:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/105/

mysql-test/suite/rocksdb/r/index_merge_rocksdb2.result
```diff
@@ -535,11 +536,10 @@ key(cola), key(colb)
 INSERT INTO t1 VALUES ('foo','bar', 'ZZ'),('fuz','baz', 'ZZ');
 OPTIMIZE TABLE t1;
 Table	Op	Msg_type	Msg_text
-test.t1	optimize	note	Table does not support optimize, doing recreate + analyze instead
 test.t1	optimize	status	OK
 select count(*) from t1;
 count(*)
-8704
+526
```
laurynas-biveinis, line 542: Say what?

george-lorch: Actually, neither 526 nor 8704 make any sense:

> # BUG#16166
> CREATE TABLE t1 (
> cola char(3) not null, colb char(3) not null,  filler char(200),
> key(cola), key(colb)
> );
> INSERT INTO t1 VALUES ('foo','bar', 'ZZ'),('fuz','baz', 'ZZ');
> OPTIMIZE TABLE t1;
> Table	Op	Msg_type	Msg_text
> test.t1	optimize	status	OK
> select count(\*) from t1;
> count(\*)
> 526

To me it looks like the result should be "2" all day long from any engine, but, both InnoDB and MyISAM return "8704" and MyRocks is returning "526" :-\

george-lorch: Nevermind, this is just from the result, the test itself disables the query log and inserts a bunch of rows.  So yeah, MyRocks is doing something wrong here, but for some reason only in 8.0 (or it's possible this is a new test bit in 8.0).

george-lorch: This is actually another utf8mb4 secondary key bug https://jira.percona.com/browse/PS-5024

laurynas-biveinis (approved):

PR #2689, opened 2018-11-16
PS-269: Fix `main.backup_locks` MTR test (8.0)

inikep:
1. Remove `CREATE PROCEDURE` and `CREATE FUNCTION` as they are transactional in 8.0.
2. Add missing `--connection`.

### DIFF
--- a/mysql-test/t/backup_locks.test
+++ b/mysql-test/t/backup_locks.test
@@ -231,21 +231,6 @@ DROP DATABASE test;
 --error ER_CANT_EXECUTE_WITH_BACKUP_LOCK
 CREATE DATABASE test1;
 
-delimiter |;
---error ER_CANT_EXECUTE_WITH_BACKUP_LOCK
-CREATE PROCEDURE p1()
-BEGIN
-  SELECT 1;
-END|
-
---error ER_CANT_EXECUTE_WITH_BACKUP_LOCK
-CREATE FUNCTION f1() RETURNS INT DETERMINISTIC
-BEGIN
-  RETURN 1;
-END|
-
-delimiter ;|
-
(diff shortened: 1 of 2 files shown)


mysql-test/t/backup_locks.test
```diff
@@ -554,48 +542,21 @@ CREATE DATABASE test1;
 
 --let $id=`SELECT CONNECTION_ID()`
 
---send CREATE PROCEDURE p1() SELECT 1
+--send INSERT INTO t_myisam VALUES(0)
```
laurynas-biveinis, line 541: This is suspicious - it means that LOCK TABLES FOR BACKUP does not block CREATE PROCEDURE, which is DDL, anymore?

laurynas-biveinis: @gl-sergei FYI ^

inikep: `--reap` returned without an error with `send CREATE PROCEDURE p1() SELECT 1`.

laurynas-biveinis: Hence my question :)

inikep: `LOCK TABLES FOR BACKUP` doesn't work for `CREATE PROCEDURE` and `CREATE FUNCTION` anymore as `mysql.proc` (MyISAM) was removed in 8.0 and replaced with `information_schema.routines` (InnoDB).

laurynas-biveinis: OK, We have discussed that's it's OK to allow such kinds of DDL under LTFB

laurynas-biveinis (approved):

PR #2692, opened 2018-11-19

laurynas-biveinis:
TokuDB: remove CMake and GCC version checks because MySQL allowed
minimum versions are higher in 8.0. At the same time drop any support
for building in the old ft-index directory.

https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/10/


george-lorch (approved):
Looks perfectly reasonable to me

PR #2693, opened 2018-11-19

laurynas-biveinis:
MyRocks: remove GCC version check because MySQL allowed minimum
version is higher in 8.0.

https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/11/


george-lorch (approved):
LGTM

PR #2696, opened 2018-11-20
PS-5018: Test fix after PS-3839 merge

dutow:
Previously a debug variable was only declared for debug builds, which resulted in failures in the perfschema.show_sanity test.
As this variable can be useful for testing release builds too, the debug ifdefs were removed.


laurynas-biveinis:
Are you sure? Then we have to document the variable and test its effect on release build

dutow:
I checked how the similar "debug" variables work - and there isn't anything else restricted to debug builds. There are several other debugging variables, and all of them is available in release builds. Some of them simply don't do anything in release mode, but they still exist there. The reason behind this is probably that there is no good way to define a variable as "only exists in debug builds" in perfscheme.show_sanity - so if a variable only exists in debug mode, it causes the test to fail in release mode.

That left me with two options:

* make the variable available in release builds, but make it a no-op or read only.
* or make it fully available in release builds too

The second seemed easier, and I didn't see any downsides with it - the name clearly states that it is a debug variable for testing, not something for production. And with this, we should be able to run some mtr tests with release builds too.

laurynas-biveinis:
That's an acceptable option and having more of MTR to run on release build is very good, but you have to review the variable as if it were a 1st-class release build variable too. QA will find release build testcases with it, some adventurous users might come to depend on it in some nasty use cases etc. Please check if it can be made 1st class release build var, if not, think what can go wrong with using it in release build and try to address it, if anything. The name stating that it's "debug" is not complete defence.

dutow:
Updated: I simply removed the variable, as it's not really needed for the tests.

laurynas-biveinis (approved):

dutow:
One more bugfix based on jenkins: I've added another warning suppression to the innodb_redo_log_encrypt_basic sys_vars test - that was caused by an earlier redo log encrytion change, and since this commit is about fixing another mtr warning, I added it here.

laurynas-biveinis (approved):

dutow:
https://ps2.cd.percona.com/job/percona-server-5.7-param/4

PR #2697, opened 2018-11-20
PS-4985: Two log encryption tests crashed with eatmydata.

dutow:
The issue wasn't realeted directly to libeatmydata, it was only a side effect.
Eatmydata made the boostrap too quick: the master thread loop wasn't executed fully even once during bootstrap.
As undo/redo encryption headers were only written out in this loop, the log files contained encrypted data, but didn't contain any information about the keys.

This change writes encryption info out immediately after the server UUID is available, fixing this issue.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -3909,6 +3909,8 @@ innobase_fix_tablespaces_empty_uuid()
 		return(false);
 	}
 
+	log_enable_encryption_if_set();
+
 	/* We only need to handle the case when an encrypted tablespace
 	is created at startup. If it is 0, there is no encrypted tablespace,
 	If it is > 1, it means we already have fixed the UUID */
@@ -3947,7 +3949,7 @@ innobase_fix_tablespaces_empty_uuid()
 	space_ids.push_back(srv_sys_space.space_id());
 	space_ids.push_back(srv_tmp_space.space_id());
 
-	bool	ret = !fil_encryption_rotate_global(space_ids);
+	bool	ret = !fil_encryption_rotate_global(space_ids) || !log_rotate_encryption();
 
 	my_free(master_key);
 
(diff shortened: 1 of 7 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -3909,6 +3909,8 @@ innobase_fix_tablespaces_empty_uuid()
 		return(false);
 	}
 
+	log_enable_encryption_if_set();
```
satya-bodapati, line 3912: Since encryption of redo log is enabled from here, we can remove  log_eneable_encryption_if_set() from srv_master_thread tasks.

satya-bodapati (changes requested):

storage/innobase/handler/ha_innodb.cc
dutow: Fixed - this also uncovered a missing rotation call, which caused crashes.

storage/innobase/log/log0log.cc
```diff
@@ -62,6 +62,8 @@ Created 12/9/1995 Heikki Tuuri
 
 #include "system_key.h"
 
+static redo_log_encrypt_enum found_log_encryption_mode = REDO_LOG_ENCRYPT_OFF;
```
satya-bodapati, line 65: Can we reuse srv_redo_log_encrypt ? By doing so, we show the correct log encryption typed used? i.e with SELECT @@innodb_redo_log_encrypt?

Some question related:
Looks like the variable is dynamic. Can we just change this variable from RK to MK and MK to RK ?

satya-bodapati (changes requested):

storage/innobase/log/log0log.cc
dutow: We can't: this variable stores what encryption type we found in the existing redo log, even if it is currently turned off.
This is because if it used master_key in the past (has encrypted data / encryption key in it), we have to rotate it during rotate innodb master key, even if the setting is currently off.
Without our redo encryption changes, this was trivial (if it was encrypted previously, it was always master key encrypted), but now we added the keyring_key option, which means we have to store the encryption mode somewhere.

satya-bodapati (approved):

laurynas-biveinis:
@dutow what is the 8.0 status of this?

PR #2701, opened 2018-11-21
PS-4744 : crash after alter table drop index is executed on slave

george-lorch:
- Under a specific situation on a slave, when binlog is off and
  log_slave_updates=0, rocksdb_prepare gets called after an
  external_lock(F_UNLCK) is called.
  During the F_UNLCK processing, the current Rdb_transaction is committed and
  cleared, leaving nothing to do or work with in the rocksdb_prepare call.

- This change solves the issue by implementing the same behavior that TokuDB has.
  That is to check if there is an active transaction at the top of the prepare
  call and exit immediately with HA_EXIT_SUCCESS if there nothing to prepare.
  This prevents falling into code that assumes that there is an active
  transaction and tries to interact with the (missing) transaction instance.

- Added new test that would segfault the slave when configured correctly and
  the slave applier exectues a specific ALTER TABLE.


george-lorch:
https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/25/

laurynas-biveinis (approved):

PR #2703, opened 2018-11-21
PS-5059: sys_vars: NON_PERSIST flag doesn't work without READ_ONLY (8.0)

inikep:
1. Add detection of the NON_PERSIST flag without the READ_ONLY flag
2. Re-record `main.read_only_persisted_variables` MTR test


laurynas-biveinis (approved):
Please also upstream

PR #2711, opened 2018-11-21
PS-4532 : Replace obsolete HAVE_purify with HAVE_VALGRIND in ha_rocks…

george-lorch:
…db.cc

- Replaced HAVE_purify with HAVE_VALGRIND as suggested.


george-lorch:
https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/26/

george-lorch:
Self approving, trivial change.

PR #2715, opened 2018-11-23
PS-5067: Immediate data scrubbing in 8.0

dutow:


dutow:
Updated: fixed formatting issues (accidentally reformatted the entire ha_innodb). Now travis check fails, because it also suggests reformatting of the entire file.

Jenkins: https://ps.cd.percona.com/job/percona-server-8.0-param/24/ (no idea why it shows a red, trying to figure it out)

mysql-test/include/search_pattern_in_file.inc
```diff
@@ -0,0 +1,98 @@
+# Purpose:
```
laurynas-biveinis, line 1: Can you use existing search_pattern.inc? (I recall something being discussed in 5.7 PRs that some include file was backported from 8.0. If that's right, nothing should be forward-ported then?)

storage/innobase/btr/btr0btr.cc
```diff
@@ -568,6 +570,17 @@ void btr_page_free_low(
 
   buf_block_modify_clock_inc(block);
 
+  const bool scrub = srv_immediate_scrub_data_uncompressed;
+
+  if (scrub) {
+    /* MariaDB code says that we can only use this code for blobs,
```
laurynas-biveinis, line 576: Our comments do not use * column for formatting

storage/innobase/btr/btr0scrub.cc
```diff
@@ -0,0 +1,21 @@
...
+#include "btr0cur.h"
+#include "dict0dict.h"
+#include "fsp0fsp.h"
+#include "ibuf0ibuf.h"
+#include "mtr0mtr.h"
+
+/**
+ * scrub data at delete time (e.g purge thread)
+ */
+bool srv_immediate_scrub_data_uncompressed = false;
```
laurynas-biveinis, line 21: Looks short-ish. Will there be something else later?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21056,6 +21058,10 @@ static MYSQL_SYSVAR_ULONG(undo_tablespaces, srv_undo_tablespaces,
                           FSP_MIN_UNDO_TABLESPACES,     /* Minimum value */
                           FSP_MAX_UNDO_TABLESPACES, 0); /* Maximum value */
 
+static MYSQL_SYSVAR_BOOL(immediate_scrub_data_uncompressed,
+    srv_immediate_scrub_data_uncompressed, 0,
+    "Enable scrubbing of data", NULL, NULL, FALSE);
```
laurynas-biveinis, line 21063: s/FALSE/false

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21653,6 +21659,8 @@ static SYS_VAR *innobase_system_variables[] = {
     MYSQL_SYSVAR(ft_ignore_stopwords),
     MYSQL_SYSVAR(encrypt_online_alter_logs),
     MYSQL_SYSVAR(encrypt_tables),
+    /* Scrubing feature */
```
laurynas-biveinis, line 21662: Redundant comment

storage/innobase/include/btr0scrub.h
```diff
@@ -0,0 +1,14 @@
...
+
+#include "univ.i"
+
+#include "btr0types.h"
+#include "data0data.h"
+#include "dict0dict.h"
+#include "mtr0mtr.h"
+#include "page0cur.h"
+
+#endif
```
laurynas-biveinis, line 14: This is essentially an empty file, not even
```
extern bool srv_immediate_scrub_data_uncompressed;
```
Why?

storage/innobase/include/page0page.h
```diff
@@ -57,6 +57,8 @@ this program; if not, write to the Free Software Foundation, Inc.,
                         ==============
 */
 
+extern bool srv_immediate_scrub_data_uncompressed;
```
laurynas-biveinis, line 60: ...and it's here, for some reason?

laurynas-biveinis (approved):
I tried to review it for "does it break existing stuff?" only, which it does not, the comments can be addressed later

mysql-test/include/search_pattern_in_file.inc
dutow: This file was ported from mariadb (and refactored by robert). The testcase using it is also from mariadb.
The difference compared to search_pattern.inc is that versions treats the file as a text file, reading it line by line (this version treats it as binary), and this in_file version also aborts the test immediately - the search_file in 8.0 only prints out a message, and the test fails at the end because of a diff in the output.

storage/innobase/btr/btr0scrub.cc
dutow: Yes, this file contains the implementation of log/background scrubbing. For immediate data scrubbing, it only defines this variable.

storage/innobase/include/btr0scrub.h
dutow: Same reason as for the source file related to it: it will contain more features with the other scrubbing parts, I just split it up into smaller commits.

storage/innobase/include/page0page.h
dutow: I kept it here, because it was declared here in mariadb/5.7. I can move it, but then I'll have to do that in 5.7 too - which will receive several more modificatiots, so I added this to the todo list too.

dutow:
Two jenkins runs: https://ps.cd.percona.com/job/percona-server-8.0-param/24/ https://ps.cd.percona.com/job/percona-server-8.0-param/22/ - seems like some random jenkins failures, looks similar to the current 8.0 head

mysql-test/include/search_pattern_in_file.inc
laurynas-biveinis: Can we add a binary or text option to search_pattern.inc, and then re-record for the abort-vs-result difference?

PR #2717, opened 2018-11-24
PS-4950: Invalid audit log file size when audit_log_rotations is chan…

gl-sergei:
…ged during runtime

audit_log_rotations and audit_log_rotate_on_size were messed up when set
at runtime.

### DIFF
--- a/mysql-test/t/audit_log_rotate.test
+++ b/mysql-test/t/audit_log_rotate.test
@@ -3,6 +3,10 @@
 let $MYSQLD_DATADIR= `select @@datadir`;
 let MYSQLD_DATADIR= $MYSQLD_DATADIR;
 
+--echo #
+--echo # Rotate with "rotations" and "size" given as startup options
+--echo #
+
 --disable_result_log
 --disable_query_log
 --source include/audit_log_events.inc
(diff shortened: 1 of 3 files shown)


gl-sergei:
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/4/

mysql-test/t/audit_log_rotate.test
```diff
@@ -24,6 +24,37 @@ perl;
...
+  }
+  die "Too many rotations!" unless scalar(@files) <= 3;
+EOF
+
+SET GLOBAL audit_log_rotations = @audit_log_rotations_orig;
+SET GLOBAL audit_log_rotate_on_size = @audit_log_rotate_on_size_orig;
+
 --remove_files_wildcard $MYSQLD_DATADIR test_audit.log*
 
 --echo success
```
laurynas-biveinis: This "echo success" is reporting the state of the 1st test above, should be moved there?

mysql-test/t/audit_log_rotate.test
```diff
@@ -24,6 +24,37 @@ perl;
...
+    print $_;
+    die "Files are too small!" unless $size >= 4096;
+  }
+  die "Too many rotations!" unless scalar(@files) <= 3;
+EOF
+
+SET GLOBAL audit_log_rotations = @audit_log_rotations_orig;
+SET GLOBAL audit_log_rotate_on_size = @audit_log_rotate_on_size_orig;
+
 --remove_files_wildcard $MYSQLD_DATADIR test_audit.log*
```
laurynas-biveinis: How about --remove_files_wildcard before this snippet too, so that it's not dependant on the previous snippet?

mysql-test/t/audit_log_rotate.test
laurynas-biveinis, line 69: This "echo success" is reporting the state of the 1st test above, should be moved there?

mysql-test/t/audit_log_rotate.test
```diff
@@ -24,6 +24,37 @@ perl;
   }
   die "Rotation doesn't work!" unless scalar(@files) > 1
 EOF
+
+# PS-4950: Invalid audit log file size when audit_log_rotations is changed during runtime
```
laurynas-biveinis: Please use
```
--echo #
--echo # PS-4950 ...
--echo #
```
so that .result file has it too

laurynas-biveinis (changes requested):
OK with comments addressed, no need for further reviews

gl-sergei:
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/8/

PR #2719, opened 2018-11-24
Fix PS-4950 in 8.0

gl-sergei:


gl-sergei:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/23/

laurynas-biveinis (changes requested):

gl-sergei:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/27/

PR #2720, opened 2018-11-26
PS-4989: Fixing innodb_track_changed_pages debug validation (5.5)

dutow:
In debug builds, this setting is allowed to be turned off temporarily after it was
turned on during startup. Howewer memory garbage also caused it to be accidentally
turned on when it was disabled at startup.


mysql-test/suite/innodb/t/percona_changed_page_bmp_debug_tracking_off.test
```diff
@@ -6,6 +6,15 @@
 
 --Error ER_WRONG_VALUE_FOR_VAR
 SET @@GLOBAL.innodb_track_changed_pages=TRUE;
+
+# PS-4989: changing innodb_track_changed_pages to false modifies it to true
```
laurynas-biveinis:
```
--echo #
--echo # PS-4989 ...
--echo #
```
so that .result file has testcase info too

laurynas-biveinis (changes requested):
OK to push after addressing the change and re-running the testcase locally, no need for further reviews

PR #2722, opened 2018-11-26
PS-269: Fix handling of mysqld's return values in mysqld_safe.sh (8.0)

inikep:
MySQL 8.0 has added handling of MYSQLD_RESTART_EXIT (16) which conflicts with handling of mysqld's return values ported from PS-5.7.
This patch fixes `main.restart_server` MTR test.


laurynas-biveinis (approved):
LGTM
FYI @EvgeniyPatlan

PR #2725, opened 2018-11-27
PS-4951: Post push fix. Many libc-related Valgrind errors on CentOS7 (5.7)

inikep:
Add support for a different valgrind or glibc version that reports:
`(below main) (libc-start.c:287)`
instead of
`main (main.cc:30)`.
This patch fixes issues with debian:jessie.


inikep:
Verified at https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/35/

laurynas-biveinis (approved):

PR #2732, opened 2018-11-28
Null-merge branch 5.5 (commit dd5c4f9) into 5.6

percona-ysorokin:
This is a null merge after merging upstream 5.5.62 and 5.6.42.


percona-ysorokin:
This is an "absolutely null" null merge.

laurynas-biveinis (approved):
As such, it does not need a review, self-approve

PR #2733, opened 2018-11-29
PS-5056 : handle_fatal_signal (sig=11) in ha_tokudb::write_row

george-lorch:
- Regression introduced by PS-4811 at commit 23c0a7280150636dc3f4d12e01748f31b2be06a2
  Fixed minor regression and introduced test case that exposed it.

- Will be cherry-picked to 8.0, GCA point is earlier than commit introducing
  regression.


george-lorch:
Local test looks good, jenkins running https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/38/

laurynas-biveinis (approved):

PR #2734, opened 2018-11-29
PS-269: Fix main.percona_ssl_connections_count MTR test (8.0)

inikep:
1. Fixed by changing a parameter of a call to `increment_connection_count()`
from a pointer to a reference (in sql_connect.cc)
2. Declare `THD::THD(bool enable_plugins = true)` as `explicit` to detect above-described issues

### DIFF
--- a/sql/sql_connect.cc
+++ b/sql/sql_connect.cc
@@ -375,7 +375,7 @@ void update_global_user_stats(THD *thd, bool create_user, ulonglong now) {
     if (thread_it != global_thread_stats->cend())
       update_global_thread_stats_with_thread(*thd, &thread_it->second, now);
     else if (create_user)
-      increment_count_by_id(thread_id, global_thread_stats, thd);
+      increment_count_by_id(thread_id, global_thread_stats, *thd);
   }
 
   thd->last_global_update_time = now;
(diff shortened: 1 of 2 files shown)


sql/sql_connect.cc
```diff
@@ -947,7 +947,7 @@ static bool login_connection(THD *thd, bool extra_port_connection) {
     thd->reset_stats();
 
     // Updates global user connection stats.
-    increment_connection_count(thd, true);
+    increment_connection_count(*thd, true);
```
laurynas-biveinis, line 950: Wow, so how did it even compile before? I see no overloads, nothing

inikep: It compiles without a single warning and calls
```
  THD::THD(bool enable_plugins = true);
```
This is the second similar issue in this code. At https://github.com/percona/percona-server/commit/13f1823397556b3f5014aca75b5630e5569a4537 I fixed:
```
-      update_global_thread_stats_with_thread(thd, &thread_it->second, now);
+      update_global_thread_stats_with_thread(*thd, &thread_it->second, now);
```

laurynas-biveinis: Can we mark that constructor explicit?

inikep: Done and one more issue was found.

laurynas-biveinis (approved):
OK if testing OK

PR #2735, opened 2018-11-29
PS-5084: innodb_buffer_pool_size is an uninitialized variable (5.7)

inikep:
Fix issue in `innodb_buffer_pool_size_validate` when
```srv_buf_pool_size == static_cast<ulint>(intbuf)```
leads to
```*static_cast<longlong *>(save) = requested_buf_pool_size;```
is not called at all.
Then there is an uninitialized read of var->save_result in innodb_buffer_pool_size_update.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -23751,15 +23751,15 @@ innodb_buffer_pool_size_validate(
 		return(1);
 	}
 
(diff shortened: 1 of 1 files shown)


inikep:
Tested at https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/39/

laurynas-biveinis (approved):

PR #2744, opened 2018-12-04
PS-5027 : [PS8QA] handle_fatal_signal (sig=11) in THD::mark_innodb_used

george-lorch:
- Added test in thd_report_innodb_stat to validate that thd is valid and not a
  background thread.

- Introduced new mtr helper wait_for_search_pattern.inc that combines the
  functionality of search_pattern.inc with a retry/wait loop similar to
  wait_condition to be able to wait for a pattern to appear in a file.  this
  is needed/useful to wait for a background thread to report an error to the
  mysqld log file.

- Added new test case that validates issue is fixed.

### DIFF
--- a/mysql-test/include/wait_for_search_pattern.inc
+++ b/mysql-test/include/wait_for_search_pattern.inc
@@ -1,0 +1,137 @@
+
+# Purpose:
+#    Simple search with Perl for a pattern in some file that repeats the
+#    search until a timeout occurs.  Bulk of search logic copied from
+#    include/search_pattern.inc
+#
+#    The environment variables SEARCH_FILE and SEARCH_PATTERN must be set
+#    before sourcing this routine.  The environment variable SEARCH_WAIT
+#    may be set to specify the time to keep waiting for the desired
+#    pattern to appear.  The default timeout is 60 seconds.
+#
+#    In case of
+#    - SEARCH_FILE and/or SEARCH_PATTERN is not set
+#    - SEARCH_FILE cannot be opened
+#    the test will abort immediate.
+
+perl;
+    use strict;
+    my @search_files =         glob($ENV{'SEARCH_FILE'})     or die "SEARCH_FILE not set";
+    my $search_pattern=        $ENV{'SEARCH_PATTERN'}        or die "SEARCH_PATTERN not set";
+    my $search_wait=           $ENV{'SEARCH_WAIT'}     || 60;
+    my $search_echo=           $ENV{'SEARCH_ECHO'}     || "NONE";
+    my $ignore_pattern=        $ENV{'IGNORE_PATTERN'}  || "";
+    my $search_echo_ctx=       $ENV{'SEARCH_ECHO_CTX'} || 0;
+
+    my $flag= 0;
+    my $counter= $search_wait;
+
+    # We include the current line and 'n' previous lines in its context
+    # as a FIFO queue i.e., oldest entries in the queue are removed when
+    # new lines are read from the file.
+    my @context= (undef) x ($search_echo_ctx + 1);
+
+    my $line;
+
+    while ($counter)
+    {
+        foreach my $search_file (@search_files) {
+            open(FILE, "< $search_file") or die("Unable to open '$search_file': $!\n");
+
+            # Read the file line by line
+            while ($line= <FILE>)
+            {
+                my $cur_lineno= $.;
+                my $file_pos= tell(FILE);
+
+                # Context stored as a sliding window of size 'n' lines
+                shift @context;
+                push  @context, $line;
+
+                if ($line =~ /$search_pattern/)
+                {
+                    if (!(length $ignore_pattern) or !($line =~ /$ignore_pattern/))
+                    {
+                        # Line matches the specified pattern
+                        $flag= 1;
+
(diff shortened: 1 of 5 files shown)


george-lorch:
Here is the jenkins job https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/36/
I tried reconciling the failures against an 8.0 trunk (https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/37/) and it seems there are _fewer_ failures in my run, but, none seem to have any relation to the bug at hand.

mysql-test/include/wait_for_search_pattern.inc
```diff
@@ -0,0 +1,142 @@
+
+# Purpose:
+#    Simple search with Perl for a pattern in some file that repeats the
+#    search until a timeout occurs.
+#
+#    The advantages compared to thinkable auxiliary constructs using the
```
laurynas-biveinis: This paragraph does not apply in this include file

mysql-test/suite/innodb/t/percona_bug5027.test
```diff
@@ -0,0 +1,67 @@
+# Test for PS-5027 : handle_fatal_signal (sig=11) in THD::mark_innodb_used
```
laurynas-biveinis: Please merge with innodb.percona_log_slow_innodb - new section with
```
--echo #
--echo # PS-5027: blah blah
--echo #
```

sql/sql_class.cc
```diff
@@ -1105,6 +1105,9 @@ THD::~THD() {
 extern "C" void thd_report_innodb_stat(THD *thd, unsigned long long trx_id,
                                        enum mysql_trx_stat_type type,
                                        unsigned long long value) {
+  if (!thd || thd_is_background_thread(thd))
```
laurynas-biveinis: One thing I don't understand why catching a background thread or nullptr thd here works when it should have worked at an earlier point. In the scenario of this bug:
1. the crash comes from trx_stats::stop_lock_wait calling thd_report_innodb_stat with trx.mysql_thd == nullptr
2. But stop_lock_wait has an early return: if (UNIV_LIKELY(!take_stats)) return;
3. take_stats gets initialized at innobase_trx_init: trx->stats.set(innobase_slow_log_verbose(thd));
4. And innobase_slow_log_verbose already checks for nullptr and background thread thd: return thd && ... && !thd_is_background_thread(thd);

Why this does not work?

george-lorch: This is/was also my question that I could not yet answer and is why I am unconvinced that this is the correct fix.  The thread 2/background thread seems to be not flagged as a background thread (thd_is_background_thread) and is determining that it should take_stats when the transaction is being created when it really should not.  I am still comparing to see where and why the difference between 5.7 as most of the rest of this appears to be unchanged in 8.0.

laurynas-biveinis: Comparing with 5.7 wouldn't be my top choice how to proceed, chances are, the bug is the same here, but background thread difference makes it benign. I'd try dumping innodb_slow_log_verbose for thds in question at each step

george-lorch: OK, I have a partial answer.  The trx used by thread 2 is not created through innobase_trx_init(...).  It is created directly from trx_allocate_for_background(...) -> trx_create_low(...) -> trx_pools->get(...) within dict0stats.cc::dict_stats_exec_sql(...) and is coming out with take_stats=true.

My theory now is that this may also be a bug in 5.7 and only through a specific sequence are we able to hit it in 8.0.  Looking at the way the pools work, there is no memory clearing going on in either get(...) or mem_free(...) processes the block (txn).  So I believe that in this very specific case, we are getting a previously used txn block out that had the take_stats=true and it is not being cleared at any point in the transaction 'allocation' and initialization process.  I did some brute force dumping of transaction, their real memory locations, some contents, etc... and I do see that the transaction that dict_stats_exec_sql got was just previously used by a client thread and it did in fact have take_stats=true.

Following  that thought process, I removed the current fix within thd_report_innodb_stat(...), re-verified the crash, then added  trx->stats.set(false); to trx_free(...) and this seems to have prevented the crash.

Will do more testing with this in place and possibly push it for more discussion.

george-lorch:
OK, moved trx->state.set(false) to trx_allocate_for_background as that is called for transaction allocations and seems to be the place to do this kind of transaction state setup before use.  Seems to work.   Jenkins here https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/45/

laurynas-biveinis (commented):
LGTM, but consider adding DBUG_ASSERT(thd && !thd_is_background_thread(thd)) at the location of the previous fix, should any trxs initialized yet elsewhere pop up in the future

george-lorch:
That's not a bad idea @laurynas-biveinis I am also wondering if we shouldn't back port this to 5.7 or even 5.6.  From what I can see by casual browsing, it _looks_ like the same issue exists there but for whatever reason is just not being hit.

laurynas-biveinis:
We should - even if only as a code cleanup, but probably later. Only to 5.7, 5.6 is different

RoelVdP:
Thank you @georgelorch & @laurynas-biveinis for the great analysis & work.

PR #2745, opened 2018-12-04
Null-merge branch 5.6 (commit 076fb0a) into 5.7

percona-ysorokin:
This is a null merge after merging upstream 5.6.42 and 5.7.24.


percona-ysorokin:
This is an "absolutely null" null merge.
Self-approving.

PR #2750, opened 2018-12-05
PS-5100: percona_bug1289599 fails if USER is not defined (5.5)

inikep:


inikep:
Tested at https://ps.cd.percona.com/view/5.5/job/percona-server-5.5-param/8/

laurynas-biveinis (approved):

laurynas-biveinis:
Please merge even to EOL (since 5.6+ fixes are upmerges and the dev work here is already completed)

PR #2751, opened 2018-12-05
PS-5100: percona_bug1289599 fails if USER is not defined (5.6)

inikep:
3. Replace `error 2006,2013` with `error ER_ACCESS_DENIED_ERROR` (backported from 5.7)
4. Add missing `connection con1;`


inikep:
Tested at https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/10/

laurynas-biveinis (approved):

PR #2752, opened 2018-12-05
PS-5100: percona_bug1289599 fails if USER is not defined (5.7)

inikep:
3. Add missing `connection con1;`


inikep:
Tested at https://ps.cd.percona.com/job/percona-server-5.7-param/42/

laurynas-biveinis (approved):

PR #2753, opened 2018-12-05
PS-5100: percona_bug1289599 fails if USER is not defined (8.0)

inikep:
3. Add missing `connection con1;`


inikep:
Tested at https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/43/

laurynas-biveinis (approved):

PR #2761, opened 2018-12-07
PS-269: Fix a bug with SELECT with partitioned InnoDB tables (8.0)

inikep:
The issue was with `SELECT a,b,c FROM t1 WHERE b = 2 AND a = 2 AND  c > 0 AND c < 100;`
that `ha_innopart::clone` called `handler::clone` which then called `handler::ha_open(table_def=nullptr))`.

This commit fixes the issue by fetching (only for non-temp tables) a dd_table if it's NULL.
It also fixes `main.partition_index_innodb` test.


inikep:
Tested locally with `--do-test=partition`:
```
Completed: Failed 2/109 tests, 98.17% were successful.

Failing test(s): innodb.partition main.partition_myisam
```

laurynas-biveinis (approved):

PR #2771, opened 2018-12-10
PS-5118: Travis CI: Failed jobs should be red (5.6)

inikep:
Add "else false" to detect failed Travis jobs.


inikep:
Tested at https://travis-ci.org/inikep/percona-server/builds/465954344

laurynas-biveinis (approved):

PR #2777, opened 2018-12-10
PS-5043 : Merge MySQL 8.0.13

george-lorch:
- Fixed compilation of TokuDB and MyRocks due to interface change of handler
  xa_recover entrypoint.  Commit cfd4f5c0229d5b982cd870a2fda58de4bf344bb1
  introduced the change.  This is an incomplete fix as the engines do not
  provide table names with recovered XIDs therefore the server can not reapply
  MDL locks to tables with PREPAREd XA transactions after server restart.

- Fixed engine test cases that used --[disable|enable]_parsing.  Upstream
  commit 2b61e71390847626ab81c89ab0abb541dc4a60f3 changed behavior and renamed
  [disable|enable]_parsing to [diable|enable]_testcase and now requires a bug
  number argument.  Used a generic 'BUG#0' argument as this is pretty
  specific to upstream bug tracking.

- Re-recorded tests due to new column 'Expression' in SHOW KEY|INDEX results.

- Re-recorded test results due to new warnings:
    Warning 1287 'utf8mb3' is deprecated and will be removed in a future release. Please use utf8mb4 instead
    Warning 1287 Setting user variables within expressions is deprecated and will be removed in a future release. Please set variables in separate statements instead.
    Warning 3778 '<collation ' is a collation of the deprecated character set UTF8MB3. Please consider using UTF8MB4 with an appropriate collation instead.

- Re-recorded test results due to removed warnings:
    Warning 3719 'utf8' is currently an alias for the character set UTF8MB3, but will be an alias for UTF8MB4 in a future release.

- Re-recorded test results due to new rows in
  INFORMATION_SCHEMA.KEY_COiLUMN_USAGE table:
    compression_dictionary         id
    compression_dictionary         name
    compression_dictionary_cols    column_id
    compression_dictionary_cols    dict_id
    compression_dictionary_cols    table_id

- Re-recorded tests with expected out-of-range value errors and warning that have
  re-ordered due to non-deterministic result ordering.

- Re-recorded tests with non-deterministic ordering results.

- Re-recorded tests where engine error codes shifted:
    207 'Specifying DATA DIRECTORY for an individual table is not supported.' from ROCKSDB is now 208
    208 'Specifying INDEX DIRECTORY for an individual table is not supported.' from ROCKSDB is now 209
    211 'Found data corruption.' from ROCKSDB is now 212

- Fixed and re-recorded tests that suppressed specific warnings where the codes
  shifted.

- Re-recorded tests that SELECT @@optimizer_switch as new default adds skip_scan=on

- Re-recorded tests that use EXPLAIN ... where EXPLAIN output for possible_keys
  now includes all possible keys and selected key remains the same.

- Re-recorded tests that use SHOW COLUMNS with DEFAULT CURRENT_TIMESTAMP()
  columns that now show as DEFAULT_GENERATED and variants.

- Re-recorded tests that expose binlog position as there has been some change
  to binlog events.


george-lorch:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/59/

laurynas-biveinis (approved):
I think the disable_testcase pieces should not be here in the first place, consider deleting them

george-lorch:
I was considering that as well since none of them have any useful commentary as to why the sections are disabled.  If something were specifically temporary due to an in-flight issue, yeah, keep it, otherwise, why permanently disable and keep around forever...

george-lorch:
https://jira.percona.com/browse/PS-5128

PR #2781, opened 2018-12-11
PS-269: Fix position scan state in heap_rrnd() (8.0)

inikep:
Restore `info->current_record = pos->record_no;` from upstream that was incorrectly removed with a merge.

This commit fixes `main.window_functions_big`.main.window_functions_big

### DIFF
--- a/storage/heap/hp_rrnd.cc
+++ b/storage/heap/hp_rrnd.cc
@@ -55,6 +55,9 @@ int heap_rrnd(HP_INFO *info, uchar *record, HP_HEAP_POSITION *pos) {
   if (hp_extract_record(info, record, info->current_ptr))
     DBUG_RETURN(my_errno());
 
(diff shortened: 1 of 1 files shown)


inikep:
Tested locally with `--do-test=percona_heap` and `--do-test=heap`.

laurynas-biveinis (changes requested):
But there is no "info->current_record = pos->record_no;" in the 5.7 branch?

laurynas-biveinis (commented):
But there is no "info->current_record = pos->record_no;" in the 5.7 branch?

inikep:
It's upstream at https://github.com/mysql/mysql-server/blob/mysql-8.0.13/storage/heap/hp_rrnd.cc#L56

laurynas-biveinis (approved):

PR #2787, opened 2018-12-12
PS-4977 : [PS8QA] Flushing the table with read lock could lead to a s…

george-lorch:
…erver crash if PS_metadata locks is queried

- Segfault was happening on a memcpy of a nullptr.  There is conditional logic
  within PFS_column_row::make_row after a large switch statement that determines
  whether or not to copy a column name from an MDL_key into internal storage.
  For all cases and types of mdl_namespace(), the m_object_type,
  m_schema_name_length, m_object_name_length, and m_column_name_length were
  assigned based upon the type...except for the type of BACKUP_TABLES, which did
  not set the m_column_name_length, leaving it 'float' to whatever previous
  value it might have had.  BACKUP_TABLES will never have a schema, object, or
  column, but, in the case of this issue, the m_column_name_length was floating
  and allowed the logic at the tail of the method to believe that there was a
  column_name that needed copying, when there in fact was none.

- This commit adds a simple one line to assign m_column_name_length=0, and adds
  a common test include as well as tests for all Percona storage engines.


george-lorch:
There may be a more preferred way of dealing with the test case, so please suggest any alternative ideas.  Jenkins (assuming it completes) of suites innodb, tokudb, and rocksdb here https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/67/

mysql-test/include/percona_bug4977.inc
```diff
@@ -0,0 +1,15 @@
+# PS-4977 : [PS8QA] Flushing the table with read lock could lead to a server
```
laurynas-biveinis, line 1: I don't think the new testcases are required, because this fixes an existing failure: main.mdl_tablespace, please confirm

george-lorch: main.mdl_tablespace as-is currently does not reproduce the issue on 8.0 HEAD (even with large parallel and repeat), but it definitely does exercise this code path.  The root cause is non-deterministic as it requires that m_column_name_length be 'dirty' or !0 from a previous row/iteration, which is where TokuDB/MyRocks seem to come into play.  They have nothing to do really with the root cause other than perhaps altering the sequence of building out the i_s.metadata_locks result.

I can drop the tests if you prefer.  It may also be possible to contrive and force the issue with some DBUG_SYNC footwork, but I'm not sure it is worth it since this issue and fix is pretty obvious when you look at the code.  Whatever your preference is...

laurynas-biveinis: I see, let's have the testcases then

laurynas-biveinis (approved):

PR #2788, opened 2018-12-12
PS-5108 (Assertion `info->read_pos == info->read_end' failed | size_t…

laurynas-biveinis:
… my_b_fill(IO_CACHE*))

Fix a regression from 8.0.13 merge where newly-introduced
IO_CACHE_binlog_cache_storage (WL#10956) always calls my_b_fill for
reinited cache instead of consuming the in-memory buffer first,
whereas the encrypted cache expects the in-memory buffer to be
consumed before any read.

Fix by returning the memory buffer on the first
IO_CACHE_binlog_cache_storage read, at the begin method. The
alternative was to change the IOCACHE code to support reads without
used up memory buffer, but this would be less efficient too.

At the same time do minor C++'ification for the encrypted IOCACHE
code.

https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/66/#showFailuresLink - no main.encrypt_tmp_files failures nor binlog suite regressions.

@percona-ysorokin (author of the feature) @robgolebiowski (working with WL#10956) adding both you as reviewers, but IMHO one review will be enough

### DIFF
--- a/sql/binlog_ostream.cc
+++ b/sql/binlog_ostream.cc
@@ -110,7 +110,7 @@ bool IO_CACHE_binlog_cache_storage::begin(unsigned char **buffer,
 
 bool IO_CACHE_binlog_cache_storage::next(unsigned char **buffer,
                                          my_off_t *length) {
-  my_b_fill(&m_io_cache);
+  if (my_b_bytes_in_cache(&m_io_cache) == 0) my_b_fill(&m_io_cache);
 
   *buffer = m_io_cache.read_pos;
   *length = my_b_bytes_in_cache(&m_io_cache);
--- a/mysql-test/r/read_only_persisted_plugin_variables.result
+++ b/mysql-test/r/read_only_persisted_plugin_variables.result
@@ -117,7 +117,7 @@ set @@persist_only.extra_port=1111;
 # Must return 100 rows.
 SELECT count(*) from performance_schema.persisted_variables;
 count(*)
-102
+100
 # Restart server
 CALL mtr.add_suppression("Plugin audit_log reported *");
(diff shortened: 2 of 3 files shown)


sql/binlog_ostream.cc
```diff
@@ -105,12 +105,21 @@ bool IO_CACHE_binlog_cache_storage::begin(unsigned char **buffer,
...
+    return false;
+  }
+
   return next(buffer, length);
 }
 
 bool IO_CACHE_binlog_cache_storage::next(unsigned char **buffer,
                                          my_off_t *length) {
-  my_b_fill(&m_io_cache);
+  if (my_b_bytes_in_cache(&m_io_cache) == 0) my_b_fill(&m_io_cache);
```
percona-ysorokin, line 113: In the description of the commit b2a8ee5
I found the following comment
```
'Transaction_Message::append_cache()' method could call 'my_b_fill()' in case
when 'info->read_pos != info->read_end' which is not supported in the
encrypted mode.
```
So, I believe `my_b_bytes_in_cache(&m_io_cache) == 0`  (which is `*m_io_cache->current_end == *m_io_cache->current_pos`) should work the same way as `m_io_cache->read_pos == m_io_cache->read_end` here.

My only concern is why do we need modifications in the `IO_CACHE_binlog_cache_storage::begin()`. To me, just calling the modified `next()` from there should be enough.

percona-ysorokin (approved):
LGTM if MTR testing is OK

laurynas-biveinis:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/71/

PR #2803, opened 2018-12-14
PS-5025: Setting compression flags for fields on DD load

dutow:


mysql-test/t/percona_bug_ps5025.test
```diff
@@ -0,0 +1,5 @@
+# PS-5025: field flag's weren't correctly set for compressed columns, causing a crash on an insert after the column was altered
```
laurynas-biveinis: Please merge with one of the existing compressed column testcases

storage/innobase/dict/dict0dd.cc
```diff
@@ -2918,18 +2918,23 @@ static inline dict_table_t *dd_fill_dict_table(const Table *dd_tab,
 
     bool is_stored = innobase_is_s_fld(field);
 
+    ulint is_compressed =
```
laurynas-biveinis: const

laurynas-biveinis (changes requested):
Please use GCA starting at 5.6 to add the testcase to there and to 5.7 (please address the testcase merging comment too)

dutow:
Updated.

laurynas-biveinis (approved):

PR #2804, opened 2018-12-14
Fix build on bionic

EvgeniyPatlan:


build-ps/percona-server-8.0_builder.sh
```diff
@@ -670,6 +670,10 @@ build_deb(){
     fi
 
     if [ ${DEBIAN_VERSION} = "artful" -o ${DEBIAN_VERSION} = "bionic" -o ${DEBIAN_VERSION} = "cosmic" ]; then
```
george-lorch: Shouldn't you remove "cosmic" from here ?

george-lorch (changes requested):

build-ps/percona-server-8.0_builder.sh
EvgeniyPatlan: fixed

george-lorch (approved):
Approved, but Ideally you should have a jenkins proving the fix, but I understand that you do have a code blocker that prevents compilation from completing.

PR #2805, opened 2018-12-15
Implement PS-2477 (Basic test cases should be created for PAM plugin)

laurynas-biveinis:
Although PAM authentication is impossible to test in MTR without
system cooperation, it is possible to do minimal plugin testing
without testing authentication itself: that INSTALL/UNINSTALL PLUGIN
works and that it is possible to CREATE USER with this authentication
method. Add such tests to the plugin-specific test directory, add this
suite to the default suite list, and install the test files.

https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/15/ (tests actually pass, red dots is a Jenkins issue)


mysql-test/include/have_auth_pam_compat.inc
```diff
@@ -0,0 +1,21 @@
+#
+# Check if server has support for loading plugins
+#
+if (`SELECT @@have_dynamic_loading != 'YES'`) {
+  --skip Example plugin requires dynamic loading
```
dutow: "Example plugin"?

mysql-test/include/have_auth_pam.inc
```diff
@@ -0,0 +1,21 @@
+#
+# Check if server has support for loading plugins
+#
+if (`SELECT @@have_dynamic_loading != 'YES'`) {
+  --skip Example plugin requires dynamic loading
```
dutow: "Example plugin?"

percona-ysorokin (approved):
LGTM

plugin/percona-pam-for-mysql/tests/mtr/auth_pam_compat-master.opt
```diff
@@ -0,0 +1 @@
+$AUTH_PAM_OPT
```
dutow: AUTH_PAM_COMPAT_OPT?

percona-ysorokin:
It looks like I missed everything :)

laurynas-biveinis:
Fixed

PR #2808, opened 2018-12-15
PS-5117 : Cannot upgrade from PS 5.7 to 8.0 if datadir is bootstrappe…

satya-bodapati:
…d with --innodb-encrypt-tables

Problem:
-------

Upgrade from 5.7 to 8.0 doesn't work in the following scenarios.

1. 5.7 datadir has encrypted tables like mysql.plugin/servers (due to
   --innodb-encrypted-tables=ON during bootstrap)

2. presence of --innodb-encrypt-tables as mysqld option during upgrade
   (ie when PS-8.0 mysqld is started on 5.7 datadir)

Symptoms:
For 1) mysql_upgrade fails at "ALTER TABLE mysql.plugin TABLESPACE=mysql" with
error saying "source is encrypted and Destination tablespace (mysql) is unencrypted

For 2) In Debug builds, there is assertion failure when dd_properties table creation
fails. Error message "mysql tablespace cannot contain encrypted table".

Fix:
----
For 1) Detect during upgrade if table in mysql database in encrypted, and create
mysqld.ibd as encrypted.

For 2) If --innodb-encrypt-tables is present during upgrade or normal bootstrap
encrypt mysql.ibd

Once mysql.ibd is encrypted, next startup requires innodb_encrypt_tables=ON.
Note that it can be turned off later.

### DIFF
--- a/mysql-test/t/dd_upgrade_encrypted.test
+++ b/mysql-test/t/dd_upgrade_encrypted.test
@@ -1,0 +1,93 @@
+# This test contains keyring file created on 64bit which is not
+# portable
+--source include/have_64bit.inc
+--source include/no_valgrind_without_big.inc
+--source include/have_util_unzip.inc
+
+--disable_query_log
+call mtr.add_suppression("Resizing redo log from");
+call mtr.add_suppression("Upgrading redo log");
+call mtr.add_suppression("Starting to delete and rewrite log files");
+call mtr.add_suppression("New log files created");
+call mtr.add_suppression("Unknown system variable 'show_compatibility_56'");
+call mtr.add_suppression("You need to use --log-bin to make --binlog-format work");
+call mtr.add_suppression("Creating routine without parsing routine body");
+call mtr.add_suppression("Resolving dependency for the view");
+call mtr.add_suppression("references invalid");
+call mtr.add_suppression("doesn't exist");
+call mtr.add_suppression("information_schema");
+call mtr.add_suppression("Storage engine '.*' does not support system tables. \\[mysql.*\\]");
+call mtr.add_suppression("Table 'mysql.component' doesn't exist");
+call mtr.add_suppression("is expected to be transactional");
+call mtr.add_suppression("table is missing or has an incorrect definition");
+call mtr.add_suppression("ACL DDLs will not work unless mysql_upgrade is executed");
+call mtr.add_suppression("Native table .* has the wrong structure");
+call mtr.add_suppression("Column count of mysql.* is wrong");
+call mtr.add_suppression("Cannot open table mysql/version from the internal data dictionary of InnoDB though the .frm file for the table exists");
+call mtr.add_suppression("Column count of performance_schema.events_statements_summary_by_digest is wrong");
+call mtr.add_suppression("The privilege system failed to initialize correctly");
+call mtr.add_suppression("Missing system table mysql.global_grants");
+# InnoDB reports "Lock wait timeout" warnings when it tries to drop persistent
+# statistics while persistent statistics table is altered during upgrade.
+# This issue doesn't seem to cause any further trouble (as there is no persistent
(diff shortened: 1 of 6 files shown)


mysql-test/t/dd_upgrade_encrypted.test
```diff
@@ -0,0 +1,95 @@
+# This test contains zipped 5.7 data directory with innodb partitioned tables.
```
laurynas-biveinis: Probably not "innodb partitioned tables"?

laurynas-biveinis (commented):
Perhaps this scenario is better handled by creating mysql tablespace encrypted?

mysql-test/t/dd_upgrade_encrypted.test
satya-bodapati: Ok, this is copy-paste from existing upgrade tests.

And regarding encrypting mysql tablespace, yes that is an option. The error could still happen with --innodb_encrypt_tables=OFF too.

So how do we decide to encrypt mysql tablespace? Also, since this is the main hardcoded DD tabespace, I was thinking there might be issues with encrypting mysql.ibd. (May be leave some pages unencrypted like in innodb_system). Just thinking aloud.

laurynas-biveinis: A use case is that we have a 5.7 everything-is-encrypted installation and we want to upgrade to 8.0 and maintain the everything-is-encrypted property.

satya-bodapati: Ok, let me think more on this

satya-bodapati: solved the issue of innodb_encrypt_tables=OFF during upgrade by detecting if one of tables in mysql databse is encrypted or not.

And implemented encryption of mysql.ibd

mysql-test/t/dd_upgrade_encrypted.test
```diff
@@ -0,0 +1,94 @@
...
+call mtr.add_suppression("Failed to create or recover replication info repositories.");
+call mtr.add_suppression("db.opt file not found for test database. Using default Character set");
+call mtr.add_suppression("Skip re-populating collations and character sets tables in InnoDB read-only mode");
+call mtr.add_suppression("Skipped updating resource group metadata in InnoDB read only mode");
+--enable_query_log
+
+--echo # Set different paths for --datadir
+let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/data57_encrypted/data;
+
+--echo # Copy the remote tablespace & DB zip files from suite location to working location.
```
laurynas-biveinis: Not "remote tablespace", the comment looks redundant anyway

storage/innobase/handler/ha_innodb.cc
```diff
@@ -4955,6 +4959,9 @@ static bool dd_open_hardcoded(space_id_t space_id, const char *filename) {
   }
 
   if (fail) {
+    ib::error(ER_XB_MSG_2)
+        << "If mysql.ibd opening failed with flags mismatch, try startup with"
+        << " innodb_encrypt_tables = ON/OFF";
```
laurynas-biveinis: Can you rephrase, when to try ON and when to try OFF?

laurynas-biveinis (changes requested):

mysql-test/t/dd_upgrade_encrypted.test
satya-bodapati: came from existring upgrade test, I didn't bother

storage/innobase/handler/ha_innodb.cc
satya-bodapati: Done

satya-bodapati:
Addressed review comments

laurynas-biveinis (approved):

PR #2810, opened 2018-12-16
PS-5151 : Handle upgrade of encrypted general tablespaces from 5.7 to…

satya-bodapati:
… 8.0.13

Problem:
--------
Since 8.0.13, encryption is tablespace attribute and not table attribute.
Upgrade from Percona Server 5.7 with encrypted tablespace doesn't follow the above rules
This affects the ALTER behaviour and other problems when moving tables around. Also
SHOW CREATE table shows encryption attribute for tables in general tablespace.

Fix:
----
During upgrade of encrypted general tablespaces and tables, the following is done.

1. Do not register encryption attribute in dd::Table for tables in general tablespace.
2. Register encryption attribute for dd::Tablespace of general tablespaces.

### DIFF
--- a/storage/innobase/dict/dict0upgrade.cc
+++ b/storage/innobase/dict/dict0upgrade.cc
@@ -937,6 +937,19 @@ bool dd_upgrade_table(THD *thd, const char *db_name, const char *table_name,
 
   dd_set_table_options(dd_table, ib_table);
 
+  /* Tables in encrypted general tablespace from PS-5.7 have
+  encryption attribute. Remove this attribute as tables in
+  general tablespace shouldn't have it. General tablespace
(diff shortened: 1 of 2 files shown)


storage/innobase/dict/dict0upgrade.cc
```diff
@@ -937,6 +937,19 @@ bool dd_upgrade_table(THD *thd, const char *db_name, const char *table_name,
 
   dd_set_table_options(dd_table, ib_table);
 
+  /* Tables in encrypted general tablespace from PS-5.7 has
```
laurynas-biveinis: s/has/have :)

laurynas-biveinis (approved):

PR #2814, opened 2018-12-17
PS-4904 : UBSAN sql/log_event.cc:8652:22: runtime error: load of valu…

george-lorch:
…e 190, which is not a valid value for type 'bool'

- Fixed minor logic issue which tripped UBSAN.

### DIFF
--- a/sql/log_event.cc
+++ b/sql/log_event.cc
@@ -10148,7 +10148,7 @@ Rows_log_event::decide_row_lookup_algorithm_and_key()
     if (delete_update_lookup_condition &&
         table->s->primary_key == MAX_KEY)
     {
-        if (!table->s->rfr_lookup_warning)
+        if (table->s->rfr_lookup_warning != 0)
         {
           sql_print_warning("Slave: read free replication is disabled "
                             "for tokudb table `%s.%s` "


george-lorch:
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/17/

sql/log_event.cc
```diff
@@ -10148,7 +10148,7 @@ Rows_log_event::decide_row_lookup_algorithm_and_key()
     if (delete_update_lookup_condition &&
         table->s->primary_key == MAX_KEY)
     {
-        if (!table->s->rfr_lookup_warning)
+        if (table->s->rfr_lookup_warning != 0)
```
laurynas-biveinis, line 10151: That looks suspicious: is the logic invert in this patch actually correct?
Moreover, greping for rfr_lookup_warning does not show it ever being initialized to false, but that could be done by some memset on POD type TABLE_SHARE I guess

george-lorch: You are correct, I inverted the logic.  I did not look at how it was used.  I assumed that this was a warning number and not a simple bool flag.  I see only one place where it is assigned true, so it seems to me there must be some path where it is not initialized properly.

PR #2817, opened 2018-12-17
PS-5134 : handle_fatal_signal (sig=11) in thd_get_thread_id

george-lorch:
- Fixed SEGFAULT in innodb print_lock_wait_timeout when the thread that timed
  out is a background thread and has no trx.mysql_thd.  Now when this happens,
  the lock wait info will contain "<background>" for the thread id and
  "<unknown>" for the SQL.  Added test scenario.

### DIFF
--- a/storage/innobase/lock/lock0wait.cc
+++ b/storage/innobase/lock/lock0wait.cc
@@ -196,10 +196,16 @@ static void print_lock_wait_timeout(const trx_t &trx,
   size_t unused;
 
   outs << "Lock wait timeout info:\n";
-  outs << "Requested thread id: " << thd_get_thread_id(trx.mysql_thd) << "\n";
+  if (trx.mysql_thd)
+    outs << "Requested thread id: " << thd_get_thread_id(trx.mysql_thd) << "\n";
+  else
+    outs << "Requested thread id: " << "<background>" << "\n";
   outs << "Requested trx id: " << trx_get_id_for_print(&trx) << "\n";
-  outs << "Requested query: "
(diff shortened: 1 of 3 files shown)


george-lorch:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/100/

storage/innobase/lock/lock0wait.cc
```diff
@@ -196,10 +196,16 @@ static void print_lock_wait_timeout(const trx_t &trx,
   size_t unused;
 
   outs << "Lock wait timeout info:\n";
-  outs << "Requested thread id: " << thd_get_thread_id(trx.mysql_thd) << "\n";
+  if (trx.mysql_thd)
+    outs << "Requested thread id: " << thd_get_thread_id(trx.mysql_thd) << "\n";
```
laurynas-biveinis, line 200: IMHO outs << "foo :" << (trx.mysql_thd ? thd_get_thread_id() : "<background>" << "\n" would be very slightly better. No need to make this change unless you feel like it. Same comment below

laurynas-biveinis (approved):

storage/innobase/lock/lock0wait.cc
george-lorch: Yeah, it's not quite as simple as that, placing conditional within a stream...it is a runtime decision that changes types to operator<<(...) which is needed to be known at compile time.  i.e.  thd_get_thread_id() returns some kind of integer and would call operator<<(int) or something whereas "" would call operator<<(const char*).

PR #2819, opened 2018-12-17
PS-5025: Testcase ensuring that compressed columns work after altering (5.6)

dutow:


laurynas-biveinis (changes requested):
Please use --echo # instead of # for the testcase header

dutow:
updated.

laurynas-biveinis (approved):

PR #2823, opened 2018-12-17
PS-5158 : *** buffer overflow detected *** in mysqld on INSERT query

george-lorch:
- For table/index names expanded into fscs, bulk load status buffer was too
  small to contain entire name and message.  Also, code was using unsafe
  sprintf instead of snprintf.  Made buffer large enought to always contain
  full path and fscs table name as well as converted to using snprintf.
- Created new test that would trigger the overrun/segfault without the fix.
    
- This must be cherry-picked forward from 5.6 as test passage requires the fix
  from commit 53371b029266587b8624d544d33b753d93f2d9ff to be present which is
  not yet behind the 5.6->5.7 GCA point.

### DIFF
--- a/mysql-test/suite/tokudb.bugs/r/PS-5158.result
+++ b/mysql-test/suite/tokudb.bugs/r/PS-5158.result
@@ -1,0 +1,7 @@
+include/assert.inc [Requires tokudb_dir_per_db=1]
+CREATE DATABASE `new..............................................end`;
+USE `new..............................................end`;
+CREATE TABLE t1(a INT KEY,b INT)ENGINE=TokuDB;
+INSERT INTO t1 VALUES(1,11),(2,12),(3,13),(4,14),(5,15);
+USE test;
+DROP DATABASE `new..............................................end`;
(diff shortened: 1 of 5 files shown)


george-lorch:
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/18/

laurynas-biveinis:
Testcase? If the bug is benign on 5.6, still OK to include that 8.0 testcase

george-lorch:
Crap, I actually had one starting on 5.6 but seem to have lost it while creating the GCA branch and initial commit.

george-lorch:
OK, re-added test, cherry-picked new upper version branches and new jenkins https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/24/

laurynas-biveinis (approved):
Not a GCA?

george-lorch:
Nope, see the comment, in order for the test to pass, it requires a fix in 5.7 that is not yet behind the GCA point.  This is how/why I lost the test in the first place.

> This must be cherry-picked forward from 5.6 as test passage requires the fix from commit 53371b0 to be present which is not yet behind the 5.6->5.7 GCA point.

george-lorch:
Actually, 5.6->5.7 can not GCA, but, it is likely possible to GCA from 5.7->8.0, not sure.

PR #2826, opened 2018-12-17
PS-5163 : [PS8QA] handle_fatal_signal (sig=11) in DsMrr_impl::dsmrr_init

george-lorch:
- TokuDB was not properly initializing its DsMrr_impl in all locations needed.
  This was discovered in 8.0 as there has been some added validations/assertions
  there that do not exist in 5.6 or 5.7, but the change still applies to the
  lesser versions.  There isn't a known test scenario in the lesser versions
  that would expose this, but created a test that triggers the assertion in 8.0.


george-lorch:
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/22/

george-lorch:
Approved by Laurynas on slack discussion.

PR #2830, opened 2018-12-18
PS-4758: A sequence of LOCK TABLES FOR BACKUP and STOP SLAVE SQL_THREAD

gl-sergei:
can cause replication to be blocked and cannot be restarted normally

When after LOCK TABLES FOR BACKUP / LOCK BINLOG FOR BACKUP slave SQL
thread is waiting for backup / binlog lock subsequent STOP SLAVE
SQL_THREAD causing deadlock. SQL thread is waiting for backup lock, STOP
SLAVE is waiting for SQL thread to complete current statement.

Fix is to throw an error when the thread which acquired the backup lock
is trying to stop slave.


gl-sergei:
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/25/

laurynas-biveinis (approved):

PR #2832, opened 2018-12-18
Fix PS-4758 in 8.0

gl-sergei:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/110/

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -3655,7 +3655,10 @@ end_with_restore_list:
     client thread has locked tables
   */
   if (thd->locked_tables_mode ||
-      thd->in_active_multi_stmt_transaction() || thd->global_read_lock.is_acquired())
+      thd->in_active_multi_stmt_transaction() ||
+      thd->global_read_lock.is_acquired() ||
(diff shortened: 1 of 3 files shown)


sql/sql_parse.cc
```diff
@@ -3336,7 +3336,20 @@ int mysql_execute_command(THD *thd, bool first_level) {
...
       */
       if (thd->locked_tables_mode || thd->in_active_multi_stmt_transaction() ||
-          thd->global_read_lock.is_acquired()) {
+          thd->global_read_lock.is_acquired() ||
+          thd->backup_tables_lock.is_acquired()) {
+        my_error(ER_LOCK_OR_ACTIVE_TRANSACTION, MYF(0));
+        goto error;
+      }
+
+      Is_instance_backup_locked_result is_instance_locked =
```
laurynas-biveinis, line 3345: const auto, but next time :)

laurynas-biveinis (approved):

sql/sql_parse.cc
gl-sergei: got it!

PR #2834, opened 2018-12-18
PS-3672 / PS-3673 / PS-3568 / PS-3535 in preparation for PS-5041 / PS…

laurynas-biveinis:
…-4999

Improve, cleanup and attempt to stabilize extended slow query log
testcases:
- create new include file log_prepare.inc to be included from each
  testcase that uses log_start.inc/log_stop.inc/log_cleanup.inc, and
  move saving the old slow_query_log, slow_query_log_file, and
  log_output values, and the setting of log_output to FILE from
  log_start.inc. These actions are redundant (and unstable in 8.0) for
  each log_start.inc invocation. Consequently move the restoring of
  these variables from log_stop.inc to log_cleanup.inc;
- adjust the testcases to restore global variables to initial values,
  not to defaults;
- drop any DROP thing IF EXISTS foo; idioms;
- drop any restoring of session variables to default values;
- use SET SESSION instead of plain SET consistently;
- delete main.percona_bug1016991-master.opt and set
  slow_query_log_use_global_control to all in the testcase
  dynamically;
- deleted stray
  innodb.percona_log_slow_slave_statements-innodb.result.

https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/26/


mysql-test/r/percona_log_slow_query_plan.result
```diff
@@ -98,7 +98,7 @@ Sort_merge_passes	6
 SET SESSION sort_buffer_size=default;
 SHOW SESSION STATUS LIKE 'Select_scan';
 Variable_name	Value
-Select_scan	37
+Select_scan	19
```
satya-bodapati, line 101: Why do these values change?

laurynas-biveinis: log_start.inc no longer queries INFORMATION_SCHEMA.GLOBAL_VARIABLES, which bumps this one

satya-bodapati:
LGTM

satya-bodapati (approved):

PR #2846, opened 2018-12-18
PS-4758: post merge fix

gl-sergei:
8.0 version of the fix incorrectly determined whether the thread has
acquired LOCK INSTANCE FOR BACKUP lock. Instead it determined if the
lock has been acquired by any thread.

Fix is to use MDL_context::owns_equal_or_stronger_lock to check if
current thread owns the lock.

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -3342,14 +3342,11 @@ int mysql_execute_command(THD *thd, bool first_level) {
         goto error;
       }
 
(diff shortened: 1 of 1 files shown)


gl-sergei:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/113/

laurynas-biveinis (approved):

george-lorch (approved):
Laurynas beat me to it +1

PR #2848, opened 2018-12-18
PS-5105: Fix 8.0.13 binlog encryption

robgolebiowski:
Removed basic::ostream position() function, which was also added to all
inheriting classes. Binlog encryption/decryption works only on binlog
files. All binlog files interfaces already contain position function.
Thus the addition to basic::ostream was not needed.

### DIFF
--- a/plugin/group_replication/include/observer_trans.h
+++ b/plugin/group_replication/include/observer_trans.h
@@ -84,11 +84,6 @@ class Transaction_message : public Plugin_gcs_message, public Basic_ostream {
   */
   my_off_t length();
 
-  my_off_t position() const noexcept {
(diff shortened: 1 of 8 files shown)


robgolebiowski:
JR => https://ps.cd.percona.com/job/percona-server-8.0-pipeline/1141 quite a number of failures due to Sergei's push. I can re-run if needed.

laurynas-biveinis (changes requested):
Please remove GR Transaction_message::position too

robgolebiowski:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/121/

laurynas-biveinis (approved):

PR #2850, opened 2018-12-19
PS-5153: Post-fix: Fix clang warning caused by unhandled return

satya-bodapati:
type of fil_reset_encryption()


laurynas-biveinis (approved):
OK with commit message typo fixed

PR #2856, opened 2018-12-19
PS-5189: Fixing a possible crash during redo log encryption init (8.0)

dutow:
(cherry picked from commit 07accd9481e9680b3bb380cc8a7e8e49d417b94c)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -3995,12 +3995,13 @@ bool innobase_fix_tablespaces_empty_uuid() {
     return (false);
   }
 
-  log_enable_encryption_if_set();
-
   /* We only need to handle the case when an encrypted tablespace
   is created at startup. If it is 0, there is no encrypted tablespace,
   If it is > 1, it means we already have fixed the UUID */
   if (Encryption::s_master_key_id != 1) {
+    /* keyring_key based redo log encryption has to be  initialized even in
```
satya-bodapati: This comment is not helpful. The question "why" remains.

satya-bodapati (changes requested):

storage/innobase/handler/ha_innodb.cc
dutow: Okay, I'll expand that comment. Basically this is the only place where we initialize the redo log encryption now, and keyring_key has to work even when there isn't anything encrypted with master_key - this was the reason why the call originally was before this condition.

dutow: Updated the comment in the 5.7 PR - if it's good, I'll update this version too.

satya-bodapati: I think there is problem with how(or where) we are enabling the encryption of redo. I think it should be done at startup code (before innodb_inited = true is set).
I missed this point in review of PS-4985.
Fix empty uuid is doing more than it is supposed to do and also becoming hackish.

So here is my suggestion:
1. Enable redo log encryption at startup code (before innodb_inited is set to true). Similar to how temp is encrypted at startup.
2. Remove log_enable_encryption_if_set() from innobase_fix_tablespaces_empty_uuid().
3. Keep the log_rotate_encryption() rotate calls in  innobase_fix_tablespaces_empty_uuid().
4. call log_enable_encryption_if_set() from the SET variable update functions. (For example, see innodb_temp_tablespace_encryption_update())

dutow: This wouldn't work because of keyring_key.
* we already set some encryption info quite early
* but we can't communicate with the keyring yet, as that requires more initialization (for example, a uuid)
* so we generate a key/iv, and mark that it is a version not yet stored in the keyring
* and then log_enable_encryption checks this, and uploads the first version of the key to the keyring, as it is called when the keyring is already working

Master key would work the way you described because there, key rotation only rotates the log header and this can be done quickly. With keyring key, this type of rotation isn't possible, the logs only change key versions when they are closed/opened.

So its doing exactly what the name of the method suggests: it fixes the encryption information immediately after the uuid is set up.

Edit: but (4) is a good idea, I'm not sure why that didn't cause issues so far in the tests

dutow: Updated with (4), and backported some testcase fixes already done in 8.0

satya-bodapati: Well, before this patch too, log_enable_encryption_if_set() is called when uuid is empty ?

Similar problems exists with MK too. before keyring is intialized, there is need for key.

So MK uses  "default key" during init which is fixed by fix_empty_..(). Cant you use something similar?

dutow: I am using something similar, yes. But that doesn't change the base problem: log_enable_encryption has to be called after the uuid exists and innodb is inited, without that, it won't perform a full initialization for RK.

I could move some of the initialization logic to the rotation function, but that would be also a hack, as it's initialization, not rotation.

Also, this function currently exits early if there's no master key, which means it still wouldn't call rotate for RK redo logs, if nothing else is encrypted - a "hack" calling rotate in this condition still would be required.

dutow: "Well, before this patch too, log_enable_encryption_if_set() is called when uuid is empty " - before the original patch (as in the original update code) it is called periodically every time the master loop is executed... :)

satya-bodapati: Ok, if log_enable_encyrption() has to be called even after using something like default master key, then there is no good way to solve this.

Do you need the extra call only when master_key_id == 0? Otherwise we are calling log_enable_encryption() two times on every startup?

dutow: With RK? That's the only place where we call it.
With MK? Not really, but it also won't hurt (as earlier this was executed many many times)

I can add an additional condition around the first call, to only call it for RK?

satya-bodapati: For both RK and MK, I want log_enable_encryption() to be called only once (code can be present at multiple places), even though it wouldn't hurt performance. Is it possible to achieve without making it too ugly?

satya-bodapati: Another question ( I think you added this). WHy is  log_rotate_encryption() present twice?

/* Rotate log tablespace */
  bool failure1 = !log_rotate_encryption();

  bool failure2 =
      !fil_encryption_rotate_global(space_ids) || !log_rotate_encryption();

dutow: Uh, that's a merge mistake in 8.0, I'll also fix that.

dutow: 5.7 PR updated with an improved condition.

satya-bodapati: will check, thank you

satya-bodapati: looks good! and double log_rotation() problem exists in 5.7 too

satya-bodapati (approved):
Looks good

dutow:
https://ps.cd.percona.com/job/percona-server-8.0-param/126/
log_encrypt_kill rerecorded, as in 5.7

PR #2858, opened 2018-12-19
PS-4676: Add section about column families under MyRocks

borys-belinsky:


doc/source/myrocks/column-family.rst
```diff
@@ -0,0 +1,108 @@
...
+
+================================================================================
+|myrocks| Column Families
+================================================================================
+
+
+|MyRocks| stores all data in a single database instance as a collection of
+key-value pairs. This is a flat data structure that requires that keys be unique
+throughout the database. |MyRocks| incorporates table IDs and index IDs into
+the keys.
```
george-lorch: This is misleading.  A reader can easily conflate a MyRocks 'database' with a MySQL 'database'. A more correct wording is something like "MyRocks stores all of it data in a single server instance within a data structure known as a Log Structured Merge (or LSM) tree as a collection of key-value pairs.  This is a flat data structure that requires that keys be unique throughout the data structure." or something like that.  Otherwise a casual reader may take away that each individual MySQL 'database' is an individual MyRocks 'thing'.

doc/source/myrocks/column-family.rst
```diff
@@ -0,0 +1,108 @@
...
+   
+Each column family has its own MemTable. It is an in-memory data structure where
+data are written to before they are flushed to SST files. The queries also use
+MemTables first. To reduce the overall memory consumption, the number of active
+column families should stay low.
+
+The ``COMMENT`` clause can result in implicitly created column families. Although
+the following example does not specify a column family using the
+|cfname| attribute, column families ``my primary key`` and ``key ab`` are
+created nevertheless.
```
george-lorch: This only applies to 5.7.  8.0 requires the cfname=<name> syntax

george-lorch (changes requested):
Other than minor changes, looks very nice.  Thank you!

george-lorch (approved):
LGTM

PR #2860, opened 2018-12-19
PS-5196: Use CircleCI to verify clang-formatting of commits (8.0)

inikep:
Port clang-format testing from TravisCI to Circle CI.
Jobs are described in `.circleci/config.yml`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,23 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/buildpack-deps:xenial
+    steps:
+      - checkout
+      - run:
+          name: Clang-format test
+          command: |
+            curl -sSL "http://apt.llvm.org/llvm-snapshot.gpg.key" | sudo -E apt-key add -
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,23 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/buildpack-deps:xenial
```
dutow, line 5: Future idea: since this is docker based anyway, we could create a docker image based on this image + the added curl/apt-get command, and use our docker image instead. Pretty easy to do with docker hub, as this is opensource anyway.

This way:
* changing lvm scripts/apt packages won't result in breaking things, as we decide when we upgrade and what
* we won't depend on ubuntu/llvm infrastructure
* speed up the test time, as only one image has to be downloaded, which will be cached anyway on circleci

inikep: It's good idea

laurynas-biveinis (approved):
Looks good, docker can happen in a follow-up

PR #2861, opened 2018-12-19
Fix PS-5195 (Possible to enable expand_fast_index_creation variable w…

laurynas-biveinis:
…hen functionality not implemented)

Prevent enabling expand_fast_index_creation:
- by command line option: reset it to false after command line options
  have been processed;
- by SET PERSIST: mark it as non-persistable.


inikep:
LGTM

george-lorch (approved):

PR #2862, opened 2018-12-19
PS-5095: Update values of variables to new defaults

borys-belinsky:


doc/source/myrocks/variables.rst
```diff
@@ -1654,6 +1666,24 @@ Specifies whether to disable the block cache for column families.
...
+  :cli: ``--rocksdb-no-create-column-family``
+  :dyn: No
+  :scope: Global
+  :vartype: Boolean
+  :default: ``ON``
+
+Specifies whether to extract the column family name from the |sql.comment|
+clause.  When this option is set to `ON`, the |sql.comment| clause will not
+treat |cfname| as a special token; it will not be possible to create column
+families using the |sql.comment| clause.
```
george-lorch: Close, but not quite.  The cf_name= is always parsed from comments.  This variable controls what to do if the column family specified within the cf_name does not exist.

If the column family named in the cf_name exists, the index will be placed with that specified column family and this option does not come into play.

If the column family named in the cf_name does not exist, then this option is evaluated to decide what to do.  If this option is false, a new column family is automatically created and given the name specified in the cf_name and the index is placed within.  A note will be written to the mysql system log "Creating a column family <the cf_name> ...".  If this option is true, no new column family will be created and instead the index will be placed within the 'default' column family.  A warning will be returned to the client from either the CREATE or ALTER statement "CREATE | ALTER - can not find column family for storing index data and creation is not allowed."

The purpose of this option is to allow an administrator to allow or deny a user from arbitrarily creating new column families via DDL.

george-lorch (changes requested):

george-lorch (approved):
Nice work!

PR #2866, opened 2018-12-20
PS-5043: remove duplicate innodb.general_ts_encryption tests

laurynas-biveinis:
Upstream 8.0.13 introduced general tablespace encryption and added own
tests, making our testcase mostly a duplicate. Remove the duplicates
and rename the testcase to have percona_ prefix.

https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/127/


satya-bodapati (approved):
LGTM

PR #2874, opened 2018-12-21
PS-5222 : PS crash "Assertion failure: dict0dict.cc:7451:space_id !=…

satya-bodapati:
… SPACE_UNKNOWN"

           during upgrade from PS5.7.23 to PS8.0.13-3 with --innodb_file_per_table=OFF

Problem:
--------

On debug builds, when 5.7 is bootstrapped with innodb_file_per_table=OFF and then upgraded
to 8.0, upgrade fails

This is because, in 8.0, we detect whether mysql database is encrypted or not based on mysql.plugin
table. We look for tablespace flags of "mysql.plugin" table.

With innodb_file_per_table=OFF, mysql.plugin doesn't have tablespace and we assert space_id must be valid.

Fix:
----
Get space_id of "mysql.plugin" from table object (dict_table_t) and use that to determine if database is
encrypted or not for upgrade


mysql-test/t/percona_dd_upgrade_file_per_table_off.test
```diff
@@ -0,0 +1,89 @@
+--source include/not_windows.inc
+--source include/no_valgrind_without_big.inc
+--source include/have_util_unzip.inc
+
+#--disable_query_log
```
laurynas-biveinis: Probably want to uncomment this?

storage/innobase/include/dict0dict.h
```diff
@@ -131,13 +131,11 @@ void dict_table_close_and_drop(
...
 void dict_init(void);
 
-dict_table_t*
-dict_table_open_on_index_id(
-/*==================*/
-        space_index_t      table_id,       /*!< in: table id */
-        bool            dict_locked)    /*!< in: TRUE=data dictionary locked */
-        __attribute__((warn_unused_result));
-
+dict_table_t *dict_table_open_on_index_id(
```
laurynas-biveinis, line 134: Spurious or it has to make CircleCI happy?

laurynas-biveinis (changes requested):

mysql-test/t/percona_dd_upgrade_file_per_table_off.test
satya-bodapati: ah, yes, uncommented for some testing.

storage/innobase/include/dict0dict.h
satya-bodapati: made by git clean filter

laurynas-biveinis: Ok, then proceed by what CircleCI says

satya-bodapati:
addressed review comments

laurynas-biveinis (approved):

PR #2875, opened 2018-12-21
Release notes for PS 8.0.13-3

borys-belinsky:


doc/source/.res/replace.txt
```diff
@@ -19,4 +19,6 @@
 .. |centos| replace:: CentOS
 .. |rhel| replace:: RHEL
 
-		      
+.. |version.next| replace:: 9.0
```
tduzan-te, line 22: We don't actually know the next major version will be called 9.0, however that is our suspicion based on current conversations.

borys-belinsky: version.next is a variable. if it is not 9.0 it can be 42. The point is that it will be easy to update if needed just in this file.

PR #2883, opened 2018-12-24
PS-4797: rpl.rpl_master_errors failing, likely due to binlog encryption

robgolebiowski:
peek_event_header calls my_b_read. my_b_read will return error on
failure, but also on reading outside the file. We are using
peek_event_header to check if next event is START_ENCRYPTION_EVENT.
The fix is to seek back to position before peek_event_header was called,
also in case when peek_event_header returns error.


robgolebiowski:
JR 5.7 => https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/61/

laurynas-biveinis (approved):

PR #2887, opened 2018-12-24
PS-5165: Use Xenial instead of Trusty in Travis-CI (8.0)

inikep:
1. Use Xenial image and modify addresses of repos to Xenial
2. Force to check a commit if `last_commit.txt` is not found for the trunk (solves issues when `Auto cancel branch builds` is turned on for Travis CI)
3. Remove less frequently updated `ppa:jonathonf/gcc`
4. Use system libs for faster compilation
5. Increase time limit to 180 minutes for main repo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
-# Ubuntu 14.04
-dist: trusty
+# Ubuntu 16.04
+dist: xenial
 sudo: required
 language: cpp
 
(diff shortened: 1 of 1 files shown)


inikep:
Unfortunately Xenial compilation seems to be 10% slower even with `-DWITH_ICU=system`.

inikep:
Verified at https://travis-ci.org/percona/percona-server/builds/482615056

inikep:
This PR is ready for review

laurynas-biveinis (approved):

PR #2889, opened 2018-12-27
fix incorrect os_n_pending_reads count

anysql:
os_n_pending_reads is decreased after os_file_io call, but did not increased before os_file_io call


laurynas-biveinis:
Indeed, this is a 8.0 port regression. Thanks!

PR #2891, opened 2019-01-01
PS-5190, PS-5166: Fixing encryption related mtr failures

dutow:


mysql-test/t/all_persisted_variables.test
```diff
@@ -39,7 +39,7 @@
 --source include/have_binlog_format_row.inc
 
 let $total_global_vars=`SELECT COUNT(*) FROM performance_schema.global_variables where variable_name NOT LIKE 'ndb_%'`;
-let $total_persistent_vars=359;
+let $total_persistent_vars=413;
```
laurynas-biveinis, line 42: Do we have 413-359 encryption related variables?

mysql-test/t/all_persisted_variables.test
```diff
@@ -120,14 +123,15 @@ WHERE variable_name NOT IN
...
 'rbr_exec_mode')
 AND variable_name NOT LIKE 'ndb_%'
 ORDER BY variable_name;
 
 --let $count_vars= `SELECT COUNT(*) FROM all_vars;`
 --echo
---let $assert_text= 'Expect 554 variables in the table. Due to open Bugs, we are checking for 547'
---let $assert_cond= [SELECT COUNT(*) as count FROM all_vars, count, 1] = $total_global_vars-7
+--let $assert_text= 'Expect 625 variables in the table. Due to open Bugs, we are checking for 617'
+--let $assert_cond= [SELECT COUNT(*) as count FROM all_vars, count, 1] = $total_global_vars-8
```
laurynas-biveinis, line 134: Why 8?

dutow: I disabled testing one encryption variable, as it was causing errors. 7 was disabled originally by upstream.

mysql-test/t/all_persisted_variables.test
dutow: No, but we never updated this testcase for percona server - so counting all of our variables, that number looked realistic.

mysql-test/t/all_persisted_variables.test
laurynas-biveinis: Is it tracked somewhere?

dutow: I don't think this is a bug, like in the case of several upstream disabled variables - maybe the assertion message should be changed, but I didn't want to edit it more than this. The issue with this variable (and some others) is that when you set them to some invalid value, they produce different diagnostics than what the test expects.

In our case, when you set the encryption_key_id to a value that has no associated key in the keyring, it'll generate an error about it in the error log explaining this - so that's not a real issue.

laurynas-biveinis (approved):

PR #2892, opened 2019-01-02
PS-5263: no tmpdir space can lead to assert in THD::send_statement_status (8.0)

inikep:
1. Update `handle_binlog_flush_or_sync_error()` to set `my_ok(thd)` after `thd->clear_error()`.
2. Update `rpl_bug72457.inc` to support errorless statements that disable binary logging.
3. Re-record `rpl_bug72457_innodb` and `rpl_bug72457_myisam`.

### DIFF
--- a/mysql-test/suite/rpl/include/rpl_bug72457.inc
+++ b/mysql-test/suite/rpl/include/rpl_bug72457.inc
@@ -1,9 +1,19 @@
 --source include/master-slave.inc
 
-call mtr.add_suppression("Slave SQL for channel '': The incident LOST_EVENTS occurred on the master\\. Message: The content of the statement cache is corrupted while writing a rollback record of the transaction to the binary log\\. Error_code:");
-call mtr.add_suppression("The content of the statement cache is corrupted while writing a rollback record of the transaction to the binary log\\. An incident event has been written to the binary log which will stop the slaves\\.");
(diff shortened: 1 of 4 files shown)


inikep:
Tests started at https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/137/

mysql-test/suite/rpl/t/rpl_bug72457_innodb-master.opt
```diff
@@ -1,2 +1,3 @@
 --internal-tmp-disk-storage-engine=InnoDB
 --default-tmp-storage-engine=InnoDB
+--log-error=$MYSQLTEST_VARDIR/tmp/rpl_bug72457.err
```
laurynas-biveinis: This is still not good enough in the case of MTR --repeat: the testcase should restart the server in the .test file itself with the custom path, then restart in the end with default, remove the custom error log

laurynas-biveinis (changes requested):

inikep:
Tested at https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/181/

laurynas-biveinis (approved):

inikep:
Tested at https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/190/

PR #2904, opened 2019-01-04
PS-4949: Bad select+order by+limit performance in 5.7

gl-sergei:
Optimizer choosing ref-access when range access is more optimal. This is
related to upstream bug #26727773 (commit ddd9d614a32). This bug was
fixed differently in 5.7 and 8.0. 5.7 fix doesn't work for this query
and 8.0 looks more robust. It is choosing range access over ref access
if ref access using more keyparts.

Fixed by taking fix and test cases from 8.0.

Take 8.0 version of fix for #26727773

fix/re-record some test cases

### DIFF
--- a/mysql-test/include/range.inc
+++ b/mysql-test/include/range.inc
@@ -1048,7 +1048,8 @@ alter table t2 add index (a,b);
 
 # 500 rows, 1 row
 
-select 'In following EXPLAIN the access method should be ref or range, #rows~=500 (and not 2)' Z;
+--echo Access method can be range/ALL with #rows >= 500.
+--echo Or it can be ref with #rows = 2, only when there is memory limit.
 explain select * from t2 where a=1000 and b<11;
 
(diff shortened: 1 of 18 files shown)


gl-sergei:
https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/67/ it seems no optimizer-related failures left.
8.0 version will be null-merge, I'll create PR if we agree on approach.

laurynas-biveinis (changes requested):
The approach LGTM.

Digging git history suggests to backport 5c93ff30c3af40eced199cbbe3653621c1c9a0ed too, please check

gl-sergei:
Indeed, I missed that bit. Fixed and pushed

gl-sergei:
https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/75/

laurynas-biveinis (approved):

PR #2911, opened 2019-01-08
PS-5309 : rpl.rpl_corruption always failing

robgolebiowski:
Decryption error was reported for corruption failures instead of
corruption error. Also corruption error was generated instead of
decryption error. The fix was to swap those two errors.


robgolebiowski:
PR => https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/147/
straight forward fix, self approved

PR #2912, opened 2019-01-08
PS-5245 : Cannot Upgrade from MySQL 5.7.24-26 to 8.0.13-3

satya-bodapati:
Problem:
    --------
    Upgrade aborts when 5.7 datadir has more than 1K tables.
    
    This happens because the compression dictionary tables are created
    with wrong table_ids which conflicts with user table_id.
    
    Fix:
    ----
    During upgrade or bootstrap, compression dictionary tables should use
    the first available table_id after the reserved range.(ie. the table_id
    of view_table_usage which is 33). compression_dictionary should use 34 and
    compression_dictionary_cols should use 35 as table_id
    
    For the 5.7 datadirs that are successfully upgraded, the
    compression_dictionary tables have wrong table_ids (although no harm).
    Their table_ids are in the reserved range of 1024. The table_id might be
    varying depending on the number of user tables.
    
    Hence, during startup, we fix the table_ids of compression_dictionary
    tables, if they don't have table_ids 34 and 35.


mysql-test/t/percona_dd_upgrade_compression_dict_debug.test
```diff
@@ -0,0 +1,105 @@
...
+call mtr.add_suppression("Failed to create or recover replication info repositories.");
+call mtr.add_suppression("db.opt file not found for test database. Using default Character set");
+call mtr.add_suppression("Skip re-populating collations and character sets tables in InnoDB read-only mode");
+call mtr.add_suppression("Skipped updating resource group metadata in InnoDB read only mode");
+--enable_query_log
+
+--echo # Set different paths for --datadir
+let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/data57_zip_dict;
+
+--echo # Copy the remote tablespace & DB zip files from suite location to working location.
```
laurynas-biveinis: Not "remote" (I believe we already fixed this in other zip dict upgrade tests, please copy-paste from them, not from some upstream test), and consider removing obvious comments (e.g. "unzip" for "unzip")

mysql-test/t/percona_dd_upgrade_compression_dict_debug.test
```diff
@@ -0,0 +1,105 @@
...
+
+--echo # Execute mysql_upgrade
+
+--source include/mysql_upgrade_preparation.inc
+--exec $MYSQL_UPGRADE --skip-verbose --force 2>&1
+--source include/mysql_upgrade_cleanup.inc
+
+SET SESSION DEBUG = '+d,skip_dd_table_access_check';
+
+--let $view_table_usage_table_id = `select  se_private_id from mysql.tables where name = 'view_table_usage'`
```
laurynas-biveinis: extra space after select in "select  se_private_id" here and below

laurynas-biveinis (changes requested):

satya-bodapati:
Addressed review comments, added more tests :)

laurynas-biveinis (approved):

PR #2916, opened 2019-01-08
Ps 5270 5.7

laurynas-biveinis:
https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/78/


percona-ysorokin (approved):
LGTM

percona-ysorokin:
Also consider marking this test case with `--source include/big_test.inc`.

laurynas-biveinis:
The tests became "small" in my local testing, we'll see how they look in Jenkins in parallel to other tests and then will mark as big tests as needed

PR #2918, opened 2019-01-09
Ps 5.7 5217

george-lorch:


george-lorch:
Here is a jenkins run with one passing job, the rest fail due to JEN-936.  I'd rather not merge until the full set passes but who knows how long it will be until jenkins works again.
https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/81/

storage/rocksdb/ha_rocksdb.h
```diff
@@ -149,6 +150,11 @@ const char *const BG_THREAD_NAME = "myrocks-bg";
 */
 const char *const INDEX_THREAD_NAME = "myrocks-index";
 
+/*
+  Name for the manual compaction thread.
+*/
+const char *const MANUAL_COMPACTION_THREAD_NAME = "myrocks-mc";
```
laurynas-biveinis: Ideally
static const char constexpr MANUAL_COMPACTION_THREAD_NAME[] = "myrocks-mc";

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -302,29 +286,46 @@ static int rocksdb_force_flush_memtable_and_lzero_now(
   rocksdb::ColumnFamilyMetaData metadata;
   rocksdb::ColumnFamilyDescriptor cf_descr;
 
+  int i, max_attempts = 3, num_errors = 0;
```
laurynas-biveinis: max_attempts is static const constexpr?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -302,29 +286,46 @@ static int rocksdb_force_flush_memtable_and_lzero_now(
...
-    cf_handle->GetDescriptor(&cf_descr);
-    c_options.output_file_size_limit = cf_descr.options.target_file_size_base;
+    for (i = 0; i < max_attempts; i++) {
+      rdb->GetColumnFamilyMetaData(cf_handle, &metadata);
+      cf_handle->GetDescriptor(&cf_descr);
+      c_options.output_file_size_limit = cf_descr.options.target_file_size_base;
+
+      DBUG_ASSERT(metadata.levels[0].level == 0);
+      std::vector<std::string> file_names;
+      for (auto &file : metadata.levels[0].files) {
```
laurynas-biveinis: const auto &

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -12179,6 +12344,156 @@ void Rdb_background_thread::run() {
...
+      // gets a shutdown signal and this code depended on that behavior.
+      // In 5.7, for whatever reason, this is not the case.  my_sleep will
+      // continue to sleep until the sleep time has elapsed.  For the purpose
+      // of this variable and the accompanying test case, we need to break this
+      // down into a loop that sleeps and checks to see if the thread was
+      // signalled with the stop flag.  It is ugly, but without having DBUG_SYNC
+      // available in background threads, it is good enough for the test.
+      for (uint32_t sleeps = 0; sleeps < rocksdb_debug_manual_compaction_delay;
+           sleeps++) {
+        if (m_stop)
```
laurynas-biveinis: Needs to be protected by m_signal_mutex?

laurynas-biveinis (changes requested):

storage/rocksdb/ha_rocksdb.h
george-lorch: True, inherited from upstream but missed while fixing in port.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Same

storage/rocksdb/ha_rocksdb.cc
george-lorch: Same.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Possibly, but for this specific purpose it is only an unprotected/dirty read which shouldn't cause any real problems.  m_stop is just the indicator that the thread has been requested to terminate.  Since I am going to re-push for the other suggestions, I can go ahead and add in a lock/read/unlock around m_stop.

george-lorch:
OK, fixed up and actually got a decent jenkins run out of it https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/82/ 
The rocksdb.show_table_status failure is known and has already been fixed in 5.7 but is after the GCA point.
The rocksdb.issue255 instability is also known and logged and unrelated to this work.
The rocksdb.rocksdb_parts is also a known issue where we are getting a positive ulonglong max error returned instead of a -1 only on CentOS 6.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: I am not sure whether this is safe even accepting that it's a dirty read: what prevents compiler lifting its load out of the loop?

laurynas-biveinis (approved):

PR #2919, opened 2019-01-10
Fixes PS-2455 - innodb does not report filename in I/O error message

altmannmarcelo:
This commit implements a translation of file descritor to filename
and report the corresponding filename that InnoDB was trying to read
when the io operation failed.

### DIFF
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -6206,12 +6206,18 @@ os_file_read_page(
 		char fdname[FN_REFLEN];
 		snprintf(fdname, sizeof fdname, "/proc/%d/fd/%d", getpid(), file);
 		char filename[FN_REFLEN];
-		my_readlink(filename, fdname, MYF(0));
-		ib::error() << "Tried to read " << n
-			<< " bytes at offset " << offset
-			<< ", but was only able to read " << n_bytes
-			<< " of FD " << file
-			<< ", filename " << std::string(filename);
+		int err_filename = my_readlink(filename, fdname, MYF(0));
+		if (err_filename != -1) {
+			ib::error() << "Tried to read " << n
+				<< " bytes at offset " << offset
+				<< ", but was only able to read " << n_bytes
+				<< " of FD " << file
+				<< ", filename " << std::string(filename);
+		} else {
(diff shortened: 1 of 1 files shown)


storage/innobase/os/os0file.cc
```diff
@@ -6203,9 +6203,15 @@ os_file_read_page(
 			}
 		}
 
+		char fdname[FN_REFLEN];
+		snprintf(fdname, sizeof fdname, "/proc/%d/fd/%d", getpid(), file);
+		char filename[FN_REFLEN];
+		my_readlink(filename, fdname, MYF(0));
```
laurynas-biveinis: Any errors from my_readlink should be handled by gracefully degrading to the old output format

storage/innobase/os/os0file.cc
```diff
@@ -6203,9 +6203,15 @@ os_file_read_page(
...
 		ib::error() << "Tried to read " << n
 			<< " bytes at offset " << offset
-			<< ", but was only able to read " << n_bytes;
+			<< ", but was only able to read " << n_bytes
```
laurynas-biveinis: The MTR testsuite needs to be checked for any testcases expecting and depending on this output format (try grepping for "Tried to read " in mysql-test/*).

laurynas-biveinis (changes requested):
This PR also needs to have a 8.0 version.

Will you be able to address review comments?

storage/innobase/os/os0file.cc
altmannmarcelo: Done

storage/innobase/os/os0file.cc
altmannmarcelo:
```
[root@marcelo-altmann-BUILD-standalone-1 percona-server]# grep -Ri 'Tried to read' mysql-test/ | wc -l
0
```

altmannmarcelo:
@laurynas-biveinis implemented the requested changes.

8.0 PR - https://github.com/percona/percona-server/pull/2925

laurynas-biveinis:
Both 5.7 and 8.0 PRs were converted to https://github.com/percona/percona-server/pull/2938 and https://github.com/percona/percona-server/pull/2939. Thanks!

PR #2920, opened 2019-01-10
PS-4640: FTS query with limit and mecab in file fts0que.cc line 3560

gl-sergei:
Upstream commit 2cd0ebf97e1 introduced the fix for FTS query result
cache overflow issue.

It has couple of issues, for example, it only counts docs added from
fts_query_filter_doc_ids, while docs can also be added from
fts_query_search_phrase.

This problem was partly mitigated by commit 48e07ec3935, but not fully
resolved.

This patch reverts 48e07ec3935 and applies the patch attached to
upstream bug 87518.

### DIFF
--- a/storage/innobase/fts/fts0que.cc
+++ b/storage/innobase/fts/fts0que.cc
@@ -746,6 +746,10 @@ fts_query_union_doc_id(
 
 		query->total_size += SIZEOF_RBT_NODE_ADD
 			+ sizeof(fts_ranking_t) + RANKING_WORDS_INIT_LEN;
+
+		if (query->limit != ULONG_UNDEFINED) {
+			query->n_docs++;
+		}
 	}
 }
 
@@ -3345,7 +3349,7 @@ fts_query_filter_doc_ids(
 		}
 
 		if (query->limit != ULONG_UNDEFINED
-		    && query->limit <= ++query->n_docs) {
+		    && query->n_docs >= query->limit) {
 			goto func_exit;
 		}
 	}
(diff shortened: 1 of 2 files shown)


gl-sergei:
https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/79/

laurynas-biveinis (commented):
No testcase, neither ours nor from https://bugs.mysql.com/bug.php?id=87518 patch?

gl-sergei:
The test case from 87518 patch is essentially the `innodb_fts_limit_check.test`. I could not construct simple test case which would crash with the 48e07ec present. Crash depends on the contents of FTS index. I have a table and a query which leads to crash on that table. After rebuilding the table MySQL stop crashing.

gl-sergei:
I came to the idea that `query->n_docs` is not counted accurately when I debugged different queries and saw that doc_ids are added to the RBT from different places while we count them only in the `fts_query_filter_doc_ids`, then I found the 2cd0ebf97e1, moved `query->n_docs++` closer to the `rbt_add_node` and it helped. Then I tried to compose a simple test case without much luck. Then I found https://bugs.mysql.com/bug.php?id=87518 and it looked more complete than my own fix.

storage/innobase/fts/fts0que.cc
```diff
@@ -3955,6 +3958,10 @@ fts_query_can_optimize(
 		return;
 	}
 
+	if (query->limit != ULONG_UNDEFINED) {
```
laurynas-biveinis: Could you please explain this bit?

laurynas-biveinis, line 3961: Could you please explain this bit?

gl-sergei: `fts_query_can_optimize` enables one more memory optimization, the one implemented in 9f79ead0ac7. It solves the same issue of excess memory consumption by FTS query result as 2cd0ebf does. So, there is no point to enable both of them. This bit is unnecessary to fix the crash and can be removed of course.

laurynas-biveinis (approved):

PR #2922, opened 2019-01-10
Fix PS-5311 (create_zip_dict: debug output may access past string end)

laurynas-biveinis:
In create_zip_dict and drop_zip_dict, name and data args are pointers
to char arrays whose length is determined by other args rathen by
terminating NULs. Thus, passing them as-is to operator<< in DBUG_LOG
results in reads past array end. Fix by creating temp std::string
objects using the passed lengths, and passing that to operator<<.

https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/151/


satya-bodapati (approved):
LGTM! Good that you avoided DBUG_EXECUTE_IF :+1:

PR #2925, opened 2019-01-10
Fixes PS-2455 - innodb does not report filename in I/O error message

altmannmarcelo:
This commit implements a translation of file descriptor to filename
and report the corresponding filename that InnoDB was trying to read
when the io operation failed.

No MTR tests affected:
```
marceloaltmann@Marcelos-MacBook-Pro.local:~/git/percona-server$ grep -Ri 'Tried to read' mysql-test/ | wc -l
       0
```

### DIFF
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -5465,10 +5465,21 @@ static MY_ATTRIBUTE((warn_unused_result)) dberr_t
         return (err);
       }
     }
-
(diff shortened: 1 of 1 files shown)


laurynas-biveinis:
Both 5.7 and 8.0 PRs were converted to https://github.com/percona/percona-server/pull/2938 and https://github.com/percona/percona-server/pull/2939. Thanks!

PR #2937, opened 2019-01-15
PS-5333: Remove need for innodb_encrypt_tables to ON/FORCE when

robgolebiowski:
mysql.ibd is encrypted

PS-5117 introduced requirement that innodb_encrypt_tables need to be set
to ON/FORCE on server startup when mysql.ibd is encrypted. This
requirement was added because there was a need for mysql.ibd space flags
validation against hardcoded, predefined flags. In PS mysql.ibd can be
encrypted and have encrypted flag set - thus generating a mismatch with
predefined flags. In order to check whether mysql.ibd is encrypted we
will check flags from the first page of the tablespace instead of
relaying on innodb_encrypt_tables. Apart from improvement from user
point of view it will also ease keyring encryption threads development
in 8.0 as encryption threads can also change the mysql.ibd's encryption
flag.

### DIFF
--- a/mysql-test/t/percona_dd_upgrade_encrypted.test
+++ b/mysql-test/t/percona_dd_upgrade_encrypted.test
@@ -64,7 +64,8 @@ let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/data57_encrypted/data;
 --source include/shutdown_mysqld.inc
 
 --echo # Start the 8.0 server on 5.7 datadir
---let $restart_parameters = "restart: --loose-skip-log-bin --skip-log-slave-updates --datadir=$MYSQLD_DATADIR1 --innodb-encrypt-tables=ON --keyring_file_data=$MYSQL_TMP_DIR/data57_encrypted/mysecret_keyring --innodb_sys_tablespace_encrypt=ON $KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD"
+--echo # PS-5333: Remove need for innodb_encrypt_tables to ON/FORCE when mysql.ibd is encrypted
+--let $restart_parameters = "restart: --loose-skip-log-bin --skip-log-slave-updates --datadir=$MYSQLD_DATADIR1 --keyring_file_data=$MYSQL_TMP_DIR/data57_encrypted/mysecret_keyring --innodb_sys_tablespace_encrypt=ON $KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD"
 --replace_result $MYSQLD_DATADIR1 MYSQLD_DATADIR1 $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD KEYRING_PLUGIN_EARLY_LOAD
 --source include/start_mysqld.inc
 
@@ -91,3 +92,7 @@ SHOW CREATE TABLE test.t6;
 --echo # Restart the server with default options.
 --let $restart_parameters=
 --source include/start_mysqld.inc
+
+--echo # PS-5333: Remove need for innodb_encrypt_tables to ON/FORCE when mysql.ibd is encrypted
+--let $restart_parameters="restart:--innodb-encrypt-tables=OFF"
+--source include/restart_mysqld.inc
--- a/sql/dd/types/init_mode.h
+++ b/sql/dd/types/init_mode.h
@@ -1,0 +1,37 @@
+/*
+   Copyright (c) 2000, 2018, Oracle and/or its affiliates. All rights reserved.
+   Copyright (c) 2019 Percona LLC and/or its affiliates. All rights reserved.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License, version 2.0,
+   as published by the Free Software Foundation.
+
+   This program is also distributed with certain software (including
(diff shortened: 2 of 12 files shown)


robgolebiowski:
JR => https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/173.
Failing tests are also failing on trunk:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/177/ (suite=innodb)
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/178/ (suite=main)

mysql-test/t/percona_dd_upgrade_encrypted.test
```diff
@@ -91,3 +91,7 @@ SHOW CREATE TABLE test.t6;
 --echo # Restart the server with default options.
 --let $restart_parameters=
 --source include/start_mysqld.inc
+
+--echo #PS-5333: Remove need for innodb_encrypt_tables to ON/FORCE when mysql.ibd is encrypted
```
laurynas-biveinis:
```suggestion
--echo # PS-5333: Remove need for innodb_encrypt_tables to ON/FORCE when mysql.ibd is encrypted 
```

sql/dd/types/init_mode.h
```diff
@@ -0,0 +1,31 @@
+/* Copyright (c) 2018 Percona LLC and/or its affiliates. All rights reserved.
```
laurynas-biveinis: Oracle copyright too, because the enum is cut&pasted from their code

storage/innobase/dict/dict0dict.cc
```diff
@@ -7444,26 +7444,83 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
-      return (false);
-    }
-    encrypt_mysql = FSP_FLAGS_GET_ENCRYPTION(space->flags);
+  if (space == nullptr) {
+    return false;
   }
+  return FSP_FLAGS_GET_ENCRYPTION(space->flags);
+}
+
+static bool dict_should_mysql_ibd_be_create_encrypted() {
```
laurynas-biveinis:
```suggestion
static bool dict_should_mysql_ibd_be_created_encrypted() {
```

storage/innobase/dict/dict0dict.cc
```diff
@@ -7444,26 +7444,83 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
+  return srv_encrypt_tables == SRV_ENCRYPT_TABLES_ON ||
+         srv_encrypt_tables == SRV_ENCRYPT_TABLES_FORCE;
+}
+
+static bool dict_should_mysql_ibd_be_opened_as_encrypted(const char *filepath,
```
laurynas-biveinis, line 7478: Not very intuitive contract: all other bool dict_should_... functions return the "should?" answer, and this one returns success or failure instead, with "should?" in the out arg. But I have no idea of a good fix

storage/innobase/dict/dict0dict.cc
```diff
@@ -7444,26 +7444,83 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
+  pfs_os_file_t file = os_file_create_simple_no_error_handling(
+      innodb_data_file_key, filepath, OS_FILE_OPEN, OS_FILE_READ_ONLY,
+      srv_read_only_mode, &success);
+
+  if (success == false) {
+    return (false);
+  }
+
+  /* Read the first page of the tablespace */
+  byte *buf = static_cast<byte *>(ut_malloc_nokey(2 * UNIV_PAGE_SIZE));
```
laurynas-biveinis: std::unique_ptr ?

storage/innobase/dict/dict0dict.cc
```diff
@@ -7444,26 +7444,83 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
+
+static bool dict_should_mysql_ibd_be_opened_as_encrypted(const char *filepath,
+                                                         bool &encrypt_mysql) {
+  bool success = false;
+
+  pfs_os_file_t file = os_file_create_simple_no_error_handling(
+      innodb_data_file_key, filepath, OS_FILE_OPEN, OS_FILE_READ_ONLY,
+      srv_read_only_mode, &success);
+
+  if (success == false) {
```
laurynas-biveinis:
```suggestion
  if (!success) {
```

storage/innobase/include/dict0dict.h
```diff
@@ -49,6 +49,7 @@ this program; if not, write to the Free Software Foundation, Inc.,
 #include "rem0types.h"
 #include "row0types.h"
 #include "sql/dd/object_id.h"
+#include "sql/dd/types/init_mode.h"  //dict_init_mode_t
```
laurynas-biveinis:
```suggestion
#include "sql/dd/types/init_mode.h"  // dict_init_mode_t
```

laurynas-biveinis (changes requested):
Minor comments, also consider using "[const] auto" more for local var declarations in C++11

storage/innobase/dict/dict0dict.cc
robgolebiowski: I have changed that to return error by parameter and should be encrypted by return

robgolebiowski:
https://ps.cd.percona.com/job/percona-server-8.0-param/200

laurynas-biveinis (approved):

robgolebiowski:
A new PR for this issue will be opened.

PR #2938, opened 2019-01-15
Fix PS-2455 (nnodb does not report filename in I/O error message)

laurynas-biveinis:
Apply a patch by Marcelo Altmann at
https://github.com/percona/percona-server/pull/2919 which attempts to
get the file path from file descriptor through /proc/<pid>/fd/<fd>
symlink, present on Linux.

The original patch was changed to extract the /proc-reading code to a
separate function, which is used to report not only read but also
fsync and fallocate errors.

https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/83/

### DIFF
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -2462,6 +2462,18 @@ SyncFileIO::execute(const IORequest& request)
 	return(n_bytes);
 }
 
+MY_ATTRIBUTE((warn_unused_result))
+static std::string
+os_file_find_path_for_fd(
+	os_file_t fd)
+{
+	char fdname[FN_REFLEN];
+	snprintf(fdname, sizeof fdname, "/proc/%d/fd/%d", getpid(), fd);
+	char filename[FN_REFLEN];
+	const int err_filename = my_readlink(filename, fdname, MYF(0));
+	return std::string((err_filename != -1) ? filename : "");
+}
+
 /** Free storage space associated with a section of the file.
 @param[in]	fh		Open file handle
 @param[in]	off		Starting offset (SEEK_SET)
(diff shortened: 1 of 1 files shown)


storage/innobase/os/os0file.cc
```diff
@@ -2462,6 +2462,18 @@ SyncFileIO::execute(const IORequest& request)
 	return(n_bytes);
 }
 
+MY_ATTRIBUTE((unused))
```
percona-ysorokin: `warn_unused_result` instead of `unused`?

storage/innobase/os/os0file.cc
```diff
@@ -2489,11 +2501,20 @@ os_file_punch_hole_posix(
 		return(DB_IO_NO_PUNCH_HOLE);
 	}
 
-	ib::warn()
-		<< "fallocate(" << fh
-		<<", FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, "
-		<< off << ", " << len << ") returned errno: "
-		<<  errno;
+	const std::string fd_path = os_file_find_path_for_fd(fh);
+	if (fd_path != "") {
```
percona-ysorokin: `if (!fd_path.empty())`

percona-ysorokin (approved):
LGTM with `warn_unused_result` fixed

storage/innobase/os/os0file.cc
```diff
@@ -3490,11 +3511,17 @@ os_file_fsync_posix(
...
 			break;
 
-		case EIO:
+		case EIO: {
 
-                        ib::fatal()
-				<< "fsync() returned EIO, aborting.";
+			const std::string fd_path
+				= os_file_find_path_for_fd(file);
+			if (fd_path != "")
```
percona-ysorokin: likewise

storage/innobase/os/os0file.cc
```diff
@@ -6203,9 +6230,18 @@ os_file_read_page(
 			}
 		}
 
-		ib::error() << "Tried to read " << n
-			<< " bytes at offset " << offset
-			<< ", but was only able to read " << n_bytes;
+		const std::string fd_path = os_file_find_path_for_fd(file);
+		if (fd_path != "") {
```
percona-ysorokin: likewise

laurynas-biveinis:
All comments updated, https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/86/

PR #2940, opened 2019-01-16
Fix PS-5328 (fil0crypt.h:123:7: error: 'memset_s' was not declared in…

laurynas-biveinis:
… this scope on illumos)

memset_s in C11 is only guaranteed to be available if
__STDC_LIB_EXT1__ is defined by the implementation and if user defines
__STDC_WANT_LIB_EXT1__, which we didn't do. Thus add the latter define
to CMake default defines, before the feature checks.

https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/85/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -520,6 +520,9 @@ ELSE()
   ENDIF()
 ENDIF()
 
+# Portably get access to memset_s
+ADD_DEFINITIONS(-D__STDC_WANT_LIB_EXT1__=1)
+
 # Run platform tests
 INCLUDE(configure.cmake)
 


CMakeLists.txt
```diff
@@ -520,6 +520,9 @@ ELSE()
   ENDIF()
 ENDIF()
 
+# Portably get access to memset_s
+ADD_DEFINITIONS(-D__STDC_WANT_LIB_EXT1__=1)
```
percona-ysorokin, line 524: Modern Cmake best practices (https://cmake.org/cmake/help/latest/command/add_definitions.html) suggest using
more specific `ADD_COMPILE_DEFINITIONS(__STDC_WANT_LIB_EXT1__=1)`(notice that here we don't have `-D`) instead of `ADD_DEFINITIONS()`

percona-ysorokin (approved):
LGTM with optional stylistic comment

CMakeLists.txt
percona-ysorokin: Sorry, available only in very modern CMAKE (starting from 3.12). Please, ignore the comment.

PR #2953, opened 2019-01-23
Implementation of PS-3969 - 8.0

altmannmarcelo:
This commit adds a variable to track when an audit log entry was either
 dropped or written directly to the file due to its size being bigger
than audit_log_buffer_size


laurynas-biveinis (changes requested):
Same comments as for 5.6/5.7

altmannmarcelo:
ready for the next review - https://ps.cd.percona.com/job/percona-server-8.0-param/221/

plugin/audit_log/audit_log.cc
```diff
@@ -78,6 +78,8 @@ static char *audit_log_include_databases = nullptr;
 static char *audit_log_exclude_commands = nullptr;
 static char *audit_log_include_commands = nullptr;
 
+std::atomic<ulong> audit_log_buffer_size_overflow(0);
```
laurynas-biveinis: uint64_t ?

plugin/audit_log/audit_log.cc
```diff
@@ -1711,7 +1713,11 @@ static st_mysql_audit audit_log_descriptor = {
 */
 
 static SHOW_VAR audit_log_status_variables[] = {
-    {NullS, NullS, SHOW_LONG, SHOW_SCOPE_GLOBAL}};
+    {"Audit_log_buffer_size_overflow",
+     (char*) &audit_log_buffer_size_overflow,
```
laurynas-biveinis: This is a bit complicated to get absolutely right as SHOW_VAR expects a regular not atomic var. Strictly speaking we should have a function, which would do atomic load, but maybe we can get aways with static_assert(sizeof(type) == sizeof(std::atomic<type>)); next to this instead

plugin/audit_log/buffer.cc
```diff
@@ -163,6 +163,7 @@ int audit_log_buffer_write(audit_log_buffer_t *log, const char *buf,
                       log_record_state_t::COMPLETE);
       audit_log_buffer_resume(log);
     }
+    audit_log_buffer_size_overflow++;
```
laurynas-biveinis: Relax memory ordering from seq-cst to relaxed:
```suggestion
    audit_log_buffer_size_overflow.fetch_add(1, std::memory_order_relaxed)
```

laurynas-biveinis (commented):
Same comments as in 5.6/5.7 + 8.0-specific ones

plugin/audit_log/audit_log.cc
altmannmarcelo: @laurynas-biveinis `SHOW_LONG` is 4 bytes and `std::atomic<uint64_t>` is 8 bytes.
Shall we use  `std::atomic<uint32>` ?

altmannmarcelo: Never mind, I did the function with atomic load path.

altmannmarcelo:
Ready for next review - https://ps.cd.percona.com/job/percona-server-8.0-param/239/ (Jenkins currently running)

plugin/audit_log/audit_log.cc
```diff
@@ -1710,8 +1712,21 @@ static st_mysql_audit audit_log_descriptor = {
   Plugin status variables for SHOW STATUS
 */
 
+static int show_audit_log_buffer_size_overflow(THD *, SHOW_VAR *var,
+                                               char *buff) {
+    var->type = SHOW_LONG;
+    var->value = buff;
+    uint64_t *value = reinterpret_cast<uint64_t *>(buff);
+    *value = static_cast<uint64_t>(audit_log_buffer_size_overflow.load());
```
laurynas-biveinis:
```suggestion
    *value = static_cast<uint64_t>(audit_log_buffer_size_overflow.load(std::memory_order_relaxed));
```

laurynas-biveinis (changes requested):

altmannmarcelo:
Ready for next Review

laurynas-biveinis (approved):

laurynas-biveinis:
@altmannmarcelo , the branch contents are OK, but the subsequent fixes broke the GCA: the final branches are not merged to each other. Can you please fix this, which will also include flattening the fix commits with the base ones?

altmannmarcelo:
replaced by #3058

laurynas-biveinis:
@altmannmarcelo no need to close PRs and open new ones for updates. Force-push into your existing branch and the old PR will update

PR #2957, opened 2019-01-25
PS-4686: update_guide - restructure

borys-belinsky:


hrvojem (changes requested):
I think upgrading from upstream should be a specific paragraph/section, because it's really hard to pick differences from the PS upgrade.

It also should mention that users need to be careful when replacing upstream version with lower PS version as it could contain incompatible changes (for example replacing MySQL 8.0.15 with PS 8.0.14).

borys-belinsky:
Ready for review

hrvojem:
Leaving it as-is now doesn't mention that upstream packages need to be removed. For example: 
```UPGRADING FROM MYSQL 5.6 OR MYSQL 5.7 FOR DEB BASED LINUX DISTRIBUTIONS```
section shows only that PS packages should be installed. There is a link to ```Preliminary steps common to all distributions``` which shows that service only needs to be stopped. Also all this jumping up and down trough the text (so we can re-use the content) might not be the most user friendly option.

File: doc/source/upgrading_guide_upstream.rst [hunk 1]
```diff
@@ -0,0 +1,408 @@
...
+#. Remove |MySQL| server leaving data files intact. In the following
+   example, **MYSQL-PACKAGE-NAME** is a placeholder which represents
+   an installed |MySQL| package that you need to remove. For example,
+   if you have |MySQL| Server Community Edition you need to uninstall
+   `mysql-server` at least.
+
+   .. code-block:: bash
+
+      $ apt remove MYSQL-PACKAGE-NAME
+
```
hrvojem, line 88: it would be more user-friendly if we would supply the line with the package names (instead of a placeholder) that can be copy-pasted to CLI

doc/source/upgrading_guide_upstream.rst
```diff
@@ -0,0 +1,408 @@
...
+found in the section :ref:`installing_from_binaries`.
+
+The in-place upgrade using |Percona| repositories requires the following steps:
+
+.. contents::
+   :local:
+   :depth: 1
+
+.. _percona-server.in-place-upgrade.preliminary.upstream:
+
```
hrvojem, line 45: One of the preliminary steps is to disable the upstream repo (oracle's mysql community) as it may cause conflicts when installing our server packages.

doc/source/upgrading_guide_upstream.rst
```diff
@@ -0,0 +1,408 @@
...
+   .. code-block:: bash
+
+      $ rpm -qa | grep '^mysql-' | xargs rpm -e --nodeps
+   
+   .. include:: .res/text/important.package.removing-without-dependency.txt
+
+.. rubric:: Installing new packages
+
+#. Install the ``percona-server-server-57`` package
+
```
hrvojem, line 130: package name is `Percona-Server-server-57` as mentioned below in the CLI example

doc/source/upgrading_guide_upstream.rst
```diff
@@ -0,0 +1,408 @@
...
+      $ yum install Percona-Server-server-57 
+
+   If you intend to use |Percona Server| 5.7 with |TokuDB|, run :program:`yum
+   install` specifying the |TokuDB| package along with
+   ``percona-server-server-57``.
+
+   .. code-block:: bash
+
+      $ yum install percona-server-server-57 percona-server-tokudb-57
+
```
hrvojem, line 142: rpm package names for 5.6 and 5.7 are upper-case (at least the Percona-Server part)

doc/source/upgrading_guide_upstream.rst
```diff
@@ -0,0 +1,408 @@
...
+5.7.
+
+.. important::
+
+   If you are using the |TokuDB| storage engine, re-enable the storage engine
+   plugin before running :program:`mysql_upgrade`. Otherwise, you'll get errors.
+
+   .. code-block:: bash
+
+      $ ps_tokudb_admin --enable
```
hrvojem, line 220: use `ps-admin` because `ps_tokudb_admin` has been deprecated

doc/source/upgrading_guide_upstream.rst
```diff
@@ -0,0 +1,408 @@
...
+
+#. Install the new server
+
+   .. code-block:: bash
+
+      $ apt install percona-server-server-5.7
+
+   If you intend to use the |TokuDB| storage engine, specify the |TokuDB|
+   package as well:
+
```
hrvojem, line 97: Because 5.7 has MyRocks storage engine, we should mention it as well (especially as TokuDB has been deprecated in 8.0)

doc/source/upgrading_guide_upstream.rst
```diff
@@ -0,0 +1,408 @@
...
+:file:`my.cnf`. After the upgrading or installation process completes, you can
+move the old configuration file back with all unsupported system removed or
+commented out.
+
+.. important::
+
+   If you use the |TokuDB| storage engine, comment out all variables specific to
+   |TokuDB| in your configuration files before starting the server. Otherwise,
+   the server will not be able to start.
+
```
hrvojem, line 167: I don't think there is an upstream version with TokuDB.

doc/source/upgrading_guide_upstream.rst
```diff
@@ -0,0 +1,408 @@
...
+      $ service mysql restart
+
+After the service successfully restarts, you can use the new |Percona Server|
+5.7.
+
+.. important::
+
+   If you are using the |TokuDB| storage engine, re-enable the storage engine
+   plugin before running :program:`mysql_upgrade`. Otherwise, you'll get errors.
+
```
hrvojem, line 217: You can't re-enable the TokuDB engine if it hasn't been installed with the upstream packages.

hrvojem (changes requested):

PR #2959, opened 2019-01-28
PS-5221 : Page Cleaner will sleep for long time if clock changes

george-lorch:
- Added monotonic clock source for transient comparisons.

- Changed InnoDB page flusher, lru, and bp load throttle to use new monotonic
  clock source.

### DIFF
--- a/storage/innobase/buf/buf0dump.cc
+++ b/storage/innobase/buf/buf0dump.cc
@@ -422,21 +422,20 @@ too much normal client queries. */
 UNIV_INLINE
 void
 buf_load_throttle_if_needed(
-/*========================*/
-	ulint*	last_check_time,	/*!< in/out: milliseconds since epoch
+    ut_monotonic_time *last_check_time, /*!< in/out: milliseconds since epoch
 					of the last time we did check if
 					throttling is needed, we do the check
 					every srv_io_capacity IO ops. */
-	ulint*	last_activity_count,
-	ulint	n_io)			/*!< in: number of IO ops done since
+    ulint *	    last_activity_count,
+    ulint	      n_io) /*!< in: number of IO ops done since
 					buffer pool load has started */
 {
 	if (n_io % srv_io_capacity < srv_io_capacity - 1) {
 		return;
 	}
 
-	if (*last_check_time == 0 || *last_activity_count == 0) {
-		*last_check_time = ut_time_ms();
+	if (last_check_time->ms == 0 || *last_activity_count == 0) {
+		*last_check_time = ut_monotonic_time_ms();
 		*last_activity_count = srv_get_activity_count();
 		return;
 	}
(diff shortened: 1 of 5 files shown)


george-lorch:
https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/97/

laurynas-biveinis:
One thing that I missed in the original discussion is that we should fix 5.6 too, right?

storage/innobase/buf/buf0dump.cc
```diff
@@ -451,8 +451,9 @@ buf_load_throttle_if_needed(
 
 	/* There has been other activity, throttle. */
 
-	ulint	now = ut_time_ms();
-	ulint	elapsed_time = now - *last_check_time;
+	const ut_monotonic_time	now = ut_monotonic_time_ms();
+	ut_monotonic_time	elapsed_time;
```
laurynas-biveinis: elapsed_time staying as ulint would result in a smaller diff?

storage/innobase/buf/buf0flu.cc
```diff
@@ -3290,14 +3291,14 @@ DECLARE_THREAD(buf_flush_page_cleaner_coordinator)(
 			/* Request flushing for threads */
 			pc_request(ULINT_MAX, lsn_limit);
 
-			ulint tm = ut_time_ms();
+			ut_monotonic_time tm = ut_monotonic_time_ms();
```
laurynas-biveinis:
```suggestion
			const ut_monotonic_time tm = ut_monotonic_time_ms();
```

storage/innobase/buf/buf0flu.cc
```diff
@@ -3333,14 +3334,14 @@ DECLARE_THREAD(buf_flush_page_cleaner_coordinator)(
 			/* Request flushing for threads */
 			pc_request(n_to_flush, lsn_limit);
 
-			ulint tm = ut_time_ms();
+			ut_monotonic_time tm = ut_monotonic_time_ms();
```
laurynas-biveinis:
```suggestion
			const ut_monotonic_time tm = ut_monotonic_time_ms();
```

storage/innobase/include/ut0ut.h
```diff
@@ -261,6 +261,22 @@ ulint
...
+code and reveal any changes that might later be introduced by upstream.
+*/
+struct ut_monotonic_time
+{
+	ulint ms;
+};
+
+ut_monotonic_time
+ut_monotonic_time_ms(void);
+/*============*/
```
laurynas-biveinis:
```suggestion
```

laurynas-biveinis (changes requested):

george-lorch:
Hmm 5.6?  That is entirely your call to make @laurynas-biveinis  I can't imagine it being difficult to back port this onto a GCA point.

storage/innobase/buf/buf0dump.cc
george-lorch: True, it doesn't look like it currently would be a threat from an upstream change that needs triggering on a merge.

george-lorch:
Re-pushed due to 5.6 and GCA merge forward.  New jenkins with jobs that appear to have been aborted due to spot instance demand https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/111/.  Going to go ahead and merge as requested changes have been made.

PR #2960, opened 2019-01-28
Ps 8.0 5221

george-lorch:


george-lorch:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/193/

storage/innobase/buf/buf0dump.cc
```diff
@@ -399,8 +401,10 @@ void buf_load_throttle_if_needed(ulint *last_check_time,
 
   /* There has been other activity, throttle. */
 
-  ulint now = ut_time_ms();
-  ulint elapsed_time = now - *last_check_time;
+  static const constexpr std::chrono::milliseconds ms1000(1000);
+  const auto now_tm = std::chrono::steady_clock::now();
+  auto elapsed_tm = std::chrono::duration_cast<std::chrono::milliseconds>(
```
laurynas-biveinis:
```suggestion
  const auto elapsed_tm = std::chrono::duration_cast<std::chrono::milliseconds>(
```

storage/innobase/buf/buf0dump.cc
```diff
@@ -421,11 +425,11 @@ void buf_load_throttle_if_needed(ulint *last_check_time,
   "cur_activity_count == *last_activity_count" check and calling
   ut_time_ms() that often may turn out to be too expensive. */
 
-  if (elapsed_time < 1000 /* 1 sec (1000 milli secs) */) {
-    os_thread_sleep((1000 - elapsed_time) * 1000 /* micro secs */);
+  if (elapsed_tm < ms1000 /* 1 sec (1000 milli secs) */) {
```
laurynas-biveinis: Code becomes self-descriptive now:
```suggestion
  if (elapsed_tm < ms1000) {
```

storage/innobase/buf/buf0dump.cc
```diff
@@ -399,8 +401,10 @@ void buf_load_throttle_if_needed(ulint *last_check_time,
 
   /* There has been other activity, throttle. */
 
-  ulint now = ut_time_ms();
-  ulint elapsed_time = now - *last_check_time;
+  static const constexpr std::chrono::milliseconds ms1000(1000);
```
laurynas-biveinis, line 403: Once you prepare the upstream version, make a note (outside the patch, write in the bug) for them to use C++14 chrono literals directly, here and in buf_flush_page_coordinator_thread

storage/innobase/buf/buf0flu.cc
```diff
@@ -2565,16 +2565,19 @@ static ulint page_cleaner_flush_pages_recommendation(lsn_t *lsn_limit,
  @param next_loop_time	time when next loop iteration should start
```
laurynas-biveinis:
```suggestion
  @param next_loop_tm	time when next loop iteration should start
```

storage/innobase/buf/buf0flu.cc
```diff
@@ -2565,16 +2565,19 @@ static ulint page_cleaner_flush_pages_recommendation(lsn_t *lsn_limit,
...
+static ulint pc_sleep_if_needed(
+    const std::chrono::steady_clock::time_point &next_loop_tm,
+    int64_t sig_count) {
 
-  if (next_loop_time > cur_time) {
+  const auto cur_tm = std::chrono::steady_clock::now();
+
+  if (next_loop_tm > cur_tm) {
     /* Get sleep interval in micro seconds. We use
     ut_min() to avoid long sleep in case of wrap around. */
```
laurynas-biveinis, line 2575: Don't see how that wraparound can happen now or even in the old code, but it's outside the work scope

storage/innobase/buf/buf0flu.cc
```diff
@@ -2565,16 +2565,19 @@ static ulint page_cleaner_flush_pages_recommendation(lsn_t *lsn_limit,
...
-    ulint sleep_us;
-
-    sleep_us =
-        ut_min(static_cast<ulint>(1000000), (next_loop_time - cur_time) * 1000);
+    const auto sleep_tm = std::chrono::duration_cast<std::chrono::microseconds>(
+        next_loop_tm - cur_tm);
+    ulint sleep_us = ut_min(static_cast<ulint>(1000000),
```
laurynas-biveinis:
```suggestion
    ulint sleep_us = std::min(static_cast<ulint>(1000000),
```

storage/innobase/buf/buf0flu.cc
```diff
@@ -3009,13 +3015,13 @@ static void buf_flush_page_coordinator_thread(size_t n_page_cleaners) {
...
-      ulint curr_time = ut_time_ms();
-
-      if (curr_time > next_loop_time + 3000) {
+      const auto curr_tm = std::chrono::steady_clock::now();
+      if (curr_tm - next_loop_tm >= ms3000) {
         if (warn_count == 0) {
-          ulint us;
-
-          us = 1000 + curr_time - next_loop_time;
+          ulint us =
```
laurynas-biveinis, line 3020: Outside the scope, but "us" usually means microseconds, and both us/ms short names would clash with C++14 chrono literals if they will be imported with using.

storage/innobase/buf/buf0flu.cc
```diff
@@ -3305,18 +3311,22 @@ void buf_flush_sync_all_buf_pools(void) {
...
-  const auto cur_time = ut_time_ms();
+  const auto cur_tm = std::chrono::steady_clock::now();
 
-  if (next_loop_time > cur_time) {
+  if (next_loop_tm > cur_tm) {
     /* Get sleep interval in micro seconds. We use
     ut_min() to avoid long sleep in case of
     wrap around. */
-    os_thread_sleep(std::min(1000000UL, (next_loop_time - cur_time) * 1000));
+    auto sleep_tm = std::chrono::duration_cast<std::chrono::microseconds>(
```
laurynas-biveinis:
```suggestion
    const auto sleep_tm = std::chrono::duration_cast<std::chrono::microseconds>(
```

storage/innobase/buf/buf0flu.cc
```diff
@@ -3305,18 +3311,22 @@ void buf_flush_sync_all_buf_pools(void) {
...
   /* If this is the server shutdown buffer pool flushing phase, skip the
   sleep to quit this thread faster */
   if (srv_shutdown_state == SRV_SHUTDOWN_FLUSH_PHASE) return;
 
-  const auto cur_time = ut_time_ms();
+  const auto cur_tm = std::chrono::steady_clock::now();
 
-  if (next_loop_time > cur_time) {
+  if (next_loop_tm > cur_tm) {
     /* Get sleep interval in micro seconds. We use
```
laurynas-biveinis: Redundant comment (and ut_min bit outdated)

storage/innobase/buf/buf0flu.cc
```diff
@@ -2565,16 +2565,19 @@ static ulint page_cleaner_flush_pages_recommendation(lsn_t *lsn_limit,
  @param next_loop_time	time when next loop iteration should start
  @param sig_count	zero or the value returned by previous call of
                          os_event_reset() */
-static ulint pc_sleep_if_needed(ulint next_loop_time, int64_t sig_count) {
-  ulint cur_time = ut_time_ms();
+static ulint pc_sleep_if_needed(
+    const std::chrono::steady_clock::time_point &next_loop_tm,
```
laurynas-biveinis, line 2569: next_loop_tm in the header comment too (or just don't remember the arguments, here and elsewhere - except for _ms ones, these are better to be renamed)

storage/innobase/buf/buf0flu.cc
```diff
@@ -3305,18 +3311,22 @@ void buf_flush_sync_all_buf_pools(void) {
 /** Make a LRU manager thread sleep until the passed target time, if it's not
 already in the past.
 @param[in]	next_loop_time	desired wake up time */
```
laurynas-biveinis:
```suggestion
 @param[in]	next_loop_tm	desired wake up time */
```

storage/innobase/buf/buf0flu.cc
```diff
@@ -3325,26 +3335,30 @@ the last flush result
 @param[in]	buf_pool	buffer pool whom we are flushing
 @param[in]	lru_n_flushed	last LRU flush page count
 @param[in,out]	lru_sleep_time	LRU manager thread sleep time */
```
laurynas-biveinis:
```suggestion
 @param[in,out]	lru_sleep_tm	LRU manager thread sleep time */
```

storage/innobase/buf/buf0flu.cc
```diff
@@ -3325,26 +3335,30 @@ the last flush result
...
-static void buf_lru_manager_adapt_sleep_time(const buf_pool_t *buf_pool,
-                                             ulint lru_n_flushed,
-                                             ulint *lru_sleep_time) {
+static void buf_lru_manager_adapt_sleep_time(
+    const buf_pool_t *buf_pool, ulint lru_n_flushed,
+    std::chrono::milliseconds &lru_sleep_tm) {
```
laurynas-biveinis, line 3336: Rename in the header comment too

storage/innobase/buf/buf0lru.cc
```diff
@@ -1213,19 +1213,19 @@ has failed
...
 @param[out]	started_monitor	whether InnoDB monitor print has been requested
 */
-static void buf_LRU_handle_lack_of_free_blocks(ulint n_iterations,
-                                               ulint started_ms,
-                                               ulint flush_failures,
-                                               bool *mon_value_was,
-                                               bool *started_monitor) {
-  static ulint last_printout_ms = 0;
+static void buf_LRU_handle_lack_of_free_blocks(
+    ulint n_iterations, const std::chrono::steady_clock::time_point &started_tm,
```
laurynas-biveinis, line 1217: Rename in the header comment too

storage/innobase/include/ut0ut.h
```diff
@@ -49,6 +49,7 @@ this program; if not, write to the Free Software Foundation, Inc.,
 #include <iterator>
 #include <ostream>
 #include <sstream>
+#include <chrono>
```
laurynas-biveinis: Too broad to include here?
Also smells of not being clang-format'ed (it reorders includes alphabetically in a single block)

storage/innobase/ut/ut0ut.cc
```diff
@@ -58,6 +58,9 @@ this program; if not, write to the Free Software Foundation, Inc.,
 #include "mysql/components/services/log_builtins.h"
 #include "sql/derror.h"
 
+static_assert(std::chrono::steady_clock::is_steady,
```
laurynas-biveinis, line 61: I believe is_steady is more useful to query the properties of all other clocks, here IMHO it's 100% redundant.

laurynas-biveinis (changes requested):

storage/innobase/ut/ut0ut.cc
george-lorch: Actually, no, the reason is that on certain platforms and older glibc versions , the steady_clock/monotonic implementation does not exist and steady_clock is implemented as system_clock.  In that case steady_clock::is_steady will be false.  For our platforms, I have already verified that we are fully covered and added this "just in case" someone tries to build where the real steady_clock implementation is missing.  See https://stackoverflow.com/questions/15032462/c11-clocks-g-steady-clockis-steady-false/15033898#15033898 for details.

storage/innobase/include/ut0ut.h
george-lorch: Yes, I need to reformat this all.  I accidentally committed and pushed with the formatting filter disabled.

storage/innobase/ut/ut0ut.cc
laurynas-biveinis: Great to know, thanks for pointing this out!

george-lorch:
OK, finally got regressions fixed, new jenkins https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/220/ and other minor changes.  @laurynas-biveinis if you care to do one more pass over when you have a moment, please do, otherwise I will go ahead and merge.

storage/innobase/buf/buf0flu.cc
laurynas-biveinis: s/remember/rename in my above comment

laurynas-biveinis (approved):

storage/innobase/buf/buf0flu.cc
george-lorch: It is, correct?  next_loop_tm is present in both the comment and the function signature.  Maybe github is not rendering it clearly?

PR #2970, opened 2019-01-30
PS-5221 : Page Cleaner will sleep for long time if clock changes

george-lorch:
- Added monotonic clock source for transient comparisons.

- Changed InnoDB page flusher, lru, and bp load throttle to use new monotonic
  clock source.

### DIFF
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -2567,27 +2567,21 @@ page_cleaner_flush_pages_if_needed(void)
 /*********************************************************************//**
 Puts the page_cleaner thread to sleep if it has finished work in less
 than a second */
-static
-void
-page_cleaner_sleep_if_needed(
-/*=========================*/
-	ulint	next_loop_time)	/*!< in: time when next loop iteration
-				should start */
-{
+static void
+page_cleaner_sleep_if_needed(ut_monotonic_time next_loop_time) {
 	/* No sleep if we are cleaning the buffer pool during the shutdown
 	with everything else finished */
 	if (srv_shutdown_state == SRV_SHUTDOWN_FLUSH_PHASE)
 		return;
 
-	ulint	cur_time = ut_time_ms();
+	const ut_monotonic_time cur_time = ut_monotonic_time_ms();
 
-	if (next_loop_time > cur_time) {
+	if (next_loop_time.ms > cur_time.ms) {
 		/* Get sleep interval in micro seconds. We use
 		ut_min() to avoid long sleep in case of
 		wrap around. */
-		os_thread_sleep(ut_min(1000000,
-				(next_loop_time - cur_time)
-				 * 1000));
+		os_thread_sleep(ut_min(
+		    1000000, (next_loop_time.ms - cur_time.ms) * 1000));
 	}
 }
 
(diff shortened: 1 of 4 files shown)


george-lorch:
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/46/

Please double check proper formatting.   I forgot how bizarre InnoDB formatting used to be in down versions and am pretty sure I have something wrong.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2630,36 +2624,34 @@ buf_get_total_LRU_list_length(void)
...
-void
+static void
 page_cleaner_adapt_lru_sleep_time(
-/*==============================*/
-	ulint*	lru_sleep_time,	/*!< in/out: desired page cleaner thread sleep
-				time for LRU flushes  */
-	ulint	lru_n_flushed) /*!< in: number of flushed in previous batch */
-
+    ut_monotonic_time
+	* lru_sleep_time, /*!< in/out: desired page cleaner thread
```
laurynas-biveinis: Suspicious formatting (just clang-format, there is a format file at storage/innobase in 5.6)

storage/innobase/buf/buf0flu.cc
```diff
@@ -2630,36 +2624,34 @@ buf_get_total_LRU_list_length(void)
...
+                                      sleep time for LRU flushes  */
+    ulint lru_n_flushed)  /*!< in: number of flushed in previous batch */
```
laurynas-biveinis: Should not have been touched

storage/innobase/buf/buf0flu.cc
```diff
@@ -2630,36 +2624,34 @@ buf_get_total_LRU_list_length(void)
...
 {
-        ulint free_len = buf_get_total_free_list_length();
-        ulint max_free_len = ut_min(buf_get_total_LRU_list_length(),
-                        srv_LRU_scan_depth * srv_buf_pool_instances);
+	ulint free_len = buf_get_total_free_list_length();
```
laurynas-biveinis: Should not have been touched

storage/innobase/buf/buf0flu.cc
```diff
@@ -2724,8 +2717,7 @@ DECLARE_THREAD(buf_flush_page_cleaner_thread)(
 	buf_page_cleaner_is_active = TRUE;
 
 	while (srv_shutdown_state == SRV_SHUTDOWN_NONE) {
-
-		ulint	page_cleaner_sleep_time;
+		ut_monotonic_time page_cleaner_sleep_time;
```
laurynas-biveinis: This one should stay as ulint (it's not a timestamp from a monotonic clock, it's duration, result of arithmetic from two monotonic clock timestamps)

laurynas-biveinis (changes requested):

storage/innobase/buf/buf0flu.cc
george-lorch: this _is_ the result of clang-format, this is why I was asking about formatting because the result doesn't look like the surrounding code.

storage/innobase/buf/buf0flu.cc
george-lorch: Result of clang-format of the function signature due to the changed type of lru_sleep_time

storage/innobase/buf/buf0flu.cc
george-lorch: Hmm, odd, not sure why that changed unless clang-format picked it up and changed it as being adjacent/part of the diff from the signature change :-\

storage/innobase/buf/buf0flu.cc
george-lorch: Yup, good catch, no need to change for this purpose.

storage/innobase/buf/buf0flu.cc
george-lorch: So on the clang-format bits, as discussed, we ack that there is no circle ci, but, there _is_ a clang-format test in place in Travis.  If you drill down through, you can see the result (clean):

> $ git diff -U0 --no-color HEAD^1 *.c *.cc *.cpp *.h *.hpp *.i *.ic *.ih | ./clang-format-diff.py -binary=clang-format-5.0 -style=file -p1 >_GIT_DIFF
> The command "git diff -U0 --no-color HEAD^1 *.c *.cc *.cpp *.h *.hpp *.i *.ic *.ih | ./clang-format-diff.py -binary=clang-format-5.0 -style=file -p1 >_GIT_DIFF" exited with 0.

With that, my guess is that the 5.6 innodb clang-format file is probably not exactly correct for the 5.6 innodb format specification.

PR #2971, opened 2019-01-31
PS-4892 : Re-implement expanded fast index creation in 8.0 DD

satya-bodapati:
Expanded fast index creation is reimplemented in 8.0.

The strategy remains same as 5.7 i.e. for ALTER ALGORITHM=COPY,

1. On the intermediate table, remove secondary indexes
2. copy data
3. Create secondary indexes.

Specifically 8.0 challenges are with DD to handle the
intermediate DD table object with removed secondary indexes.

To handle this, an in-memory flag is set in DD objects. Note that
this intermediate table is not accessible by other threads or statements.
So it is safe.

In InnoDB, dd checks for the disabled for secondary indexes which are marked
disabled. When the secondary indexes are restored, full checks are done.

Apart from that, for temporary tables, there is intermediate full commit
after copy stage.

Since inplace alter APIs do not commit the transaction in 8.0, we have trx
in active state. Because of the full intermediate commit in between, future
attempts to commit will not succeed.

It looks safe to disable the intermediate full commit for all temp tables but
we play safe and disable only when expanded fast index creation is enabled.

This commit reverts PS-5195 and PS-5124 (they are about disabling
expanded fast index creation until it is reimplemented).


sql/dd/impl/types/index_impl.h
```diff
@@ -121,6 +121,14 @@ class Index_impl : public Entity_object_impl, public Index {
 
   virtual void set_hidden(bool hidden) { m_hidden = hidden; }
 
+  /////////////////////////////////////////////////////////////////////////
+  // is_disabled.
+  /////////////////////////////////////////////////////////////////////////
+
+  virtual bool is_disabled() const { return m_disabled; }
```
laurynas-biveinis:
```suggestion
  virtual bool is_disabled() const noexcept { return m_disabled; }
```

sql/dd/impl/types/index_impl.h
```diff
@@ -121,6 +121,14 @@ class Index_impl : public Entity_object_impl, public Index {
...
+
+  virtual void set_disabled(bool disable) { m_disabled = disable; }
```
laurynas-biveinis:
```suggestion
  virtual void set_disabled(bool disable) noexcept { m_disabled = disable; }
```

sql/dd/impl/types/partition_index_impl.h
```diff
@@ -73,6 +73,16 @@ class Partition_index_impl : public Weak_object_impl, public Partition_index {
 
   virtual bool store_attributes(Raw_record *r);
 
+  /////////////////////////////////////////////////////////////////////////
+  // is_disabled.
+  /////////////////////////////////////////////////////////////////////////
+
+  virtual bool is_disabled() const { return m_index->is_disabled(); }
```
laurynas-biveinis:
```suggestion
  virtual bool is_disabled() const noexcept { return m_index->is_disabled(); }
```

sql/dd/impl/types/partition_index_impl.h
```diff
@@ -73,6 +73,16 @@ class Partition_index_impl : public Weak_object_impl, public Partition_index {
...
+
+  virtual void set_disabled(bool disable) { m_index->set_disabled(disable); }
```
laurynas-biveinis:
```suggestion
  virtual void set_disabled(bool disable) noexcept { m_index->set_disabled(disable); }
```

sql/dd/impl/types/partition_index_impl.h
```diff
@@ -73,6 +73,16 @@ class Partition_index_impl : public Weak_object_impl, public Partition_index {
...
+
+  virtual Index::enum_index_type type() const { return m_index->type(); }
```
laurynas-biveinis:
```suggestion
  virtual Index::enum_index_type type() const noexcept { return m_index->type(); }
```

sql/dd/types/index.h
```diff
@@ -100,6 +100,13 @@ class Index : virtual public Entity_object {
   virtual bool is_hidden() const = 0;
   virtual void set_hidden(bool hidden) = 0;
 
+  /////////////////////////////////////////////////////////////////////////
+  // disabled.
+  /////////////////////////////////////////////////////////////////////////
+
+  virtual bool is_disabled() const = 0;
```
laurynas-biveinis:
```suggestion
  virtual bool is_disabled() const noexcept = 0;
```

sql/dd/types/index.h
```diff
@@ -100,6 +100,13 @@ class Index : virtual public Entity_object {
...
+  virtual void set_disabled(bool disable) = 0;
```
laurynas-biveinis:
```suggestion
  virtual void set_disabled(bool disable) noexcept = 0;
```

sql/dd/types/partition_index.h
```diff
@@ -69,6 +69,14 @@ class Partition_index : virtual public Weak_object {
 
   const String_type &name() const { return index().name(); }
 
+  /////////////////////////////////////////////////////////////////////////
+  // disabled.
+  /////////////////////////////////////////////////////////////////////////
+
+  virtual bool is_disabled() const = 0;
```
laurynas-biveinis:
```suggestion
  virtual bool is_disabled() const noexcept = 0;
```

sql/dd/types/partition_index.h
```diff
@@ -69,6 +69,14 @@ class Partition_index : virtual public Weak_object {
...
+  virtual void set_disabled(bool disable) = 0;
```
laurynas-biveinis:
```suggestion
  virtual void set_disabled(bool disable) noexcept = 0;
```

sql/dd/types/partition_index.h
```diff
@@ -69,6 +69,14 @@ class Partition_index : virtual public Weak_object {
...
+  virtual Index::enum_index_type type() const = 0;
```
laurynas-biveinis:
```suggestion
  virtual Index::enum_index_type type() const noexcept = 0;
```

sql/sql_table.cc
```diff
@@ -13933,6 +13935,17 @@ static bool remove_secondary_keys(THD *thd, HA_CREATE_INFO *create_info,
       HA_ALTER_INPLACE_NOT_SUPPORTED)
     DBUG_RETURN(true);
 
+  for (auto index : *altered_table_def->indexes()) {
```
laurynas-biveinis:
```suggestion
  for (const auto index : *altered_table_def->indexes()) {
```

sql/sql_table.cc
```diff
@@ -13983,6 +13996,10 @@ static bool restore_secondary_keys(THD *thd, HA_CREATE_INFO *create_info,
   ha_alter_info.index_add_buffer =
       (uint *)thd->alloc(sizeof(uint) * alter_info->delayed_key_count);
 
+  for (auto index : *dd_disabled_sec_keys) {
```
laurynas-biveinis:
```suggestion
  for (const auto index : *dd_disabled_sec_keys) {
```

storage/innobase/dict/dict0dd.cc
```diff
@@ -1909,10 +1916,16 @@ void dd_write_table(dd::Object_id dd_space_id, Table *dd_table,
...
+    if (table->skip_alter_undo && dd_index->is_disabled()) {
+      ut_ad(dd_index->type() == dd::Index::IT_MULTIPLE);
+      continue;
+    }
+
     /* Don't assume the index orders are the same, even on
     CREATE TABLE. This could be called from TRUNCATE path,
     which would do some adjustment on FULLTEXT index, thus
     the out-of-sync order */
+
```
laurynas-biveinis: Spurious

storage/innobase/handler/handler0alter.cc
```diff
@@ -3807,7 +3807,16 @@ static void dd_commit_inplace_alter_table(
                                  FTS_DOC_ID_INDEX_NAME, col);
     }
 
-    dd_space_id = dd_first_index(old_dd_tab)->tablespace_id();
+    /* This can happen only with expanded fast index creation. On the
+    intermediate table during ALTER COPY, we drop secondary indexes using
+    inplace alter APIs. The old definition here is old copy of table. Hence we
+    should use new_dd_tab here for updating the dd::Indexes */
+    if (new_table->skip_alter_undo) {
```
laurynas-biveinis, line 3814: Maybe
`dd_space_id = dd_first_index(new_table->skip_alter_undo ? new_dd_tab : old_dd_tab)->tablespace_id();`
?

laurynas-biveinis (changes requested):
I like this approach very much.
My only "big" question is how it crash recovers for a failure while secondary index definitions are dropped? If any work is needed there, it can be filed as a follow-up issue

storage/innobase/handler/handler0alter.cc
satya-bodapati: Dont like the readability of this suggestion.

satya-bodapati:
Regarding the crash recovery, if there is crash after secondary indexes are removed, there is no loss. This is because, the entire ALTER is now one transaction and the secondary indexes removed are from the "newer version"/"interemdiate version" of table.

The original table remains intact. On crash, the original table is available. The question is whether this intermediate table is removed or not. I think it should be. The entire DD trx is rolled back.

Neverthelss, I will test this and if there is issue, will create a bug (follow-up)

storage/innobase/handler/handler0alter.cc
laurynas-biveinis: Fair enough

satya-bodapati:
Ready for review

laurynas-biveinis (approved):

satya-bodapati:
https://ps.cd.percona.com/job/percona-server-8.0-param/210/

satya-bodapati:
Recorded all_persisted_variables.test

PR #2973, opened 2019-01-31
Ps 5250.selinux 5.7

borys-belinsky:


doc/source/installation/yum_repo.rst
```diff
@@ -51,15 +51,24 @@ Installing |Percona Server| from Percona ``yum`` repository
...
+         1:percona-release        ########################################### [100%]
+
+   To install |Percona Server| with SELinux policies, you also need the :program:`Percona-Server-selinux-*.noarch.rpm` package:
 
    .. code-block:: bash
 
-     Retrieving https://repo.percona.com/yum/percona-release-latest.noarch.rpm
-     Preparing...                ########################################### [100%]
-        1:percona-release        ########################################### [100%]
+      $ yum install http://repo.percona.com/centos/7/RPMS/x86_64/Percona-Server-selinux-56-5.6.42-rel84.2.el7.noarch.rpm
```
laurynas-biveinis, line 70: 5.7 example?

hrvojem: AFAIK there is no SElinux support for 5.7. So the 5.7 docs shouldn't be updated (yet).

EvgeniyPatlan: for 5.7 there is no separate package. policies are included into server package, so we just follow upstream way to provide them

borys-belinsky: This instruction for installing *selinux* package will be removed: it is not needed for 5.7. selinux is supported in the percona-server-server-5.7 package

PR #2976, opened 2019-01-31
Ps 4688.ssl certificate compatibility matrix 5.7

borys-belinsky:


laurynas-biveinis (commented):
All the items in the JIRA issue have applying versions - please review and remove the content which does not apply to 5.7

PR #2977, opened 2019-01-31
PS-4688: ssl improvement - add section

borys-belinsky:


laurynas-biveinis (changes requested):
Please create 5.7 version first

PR #2987, opened 2019-02-05
Fix PS-1827 (LP #1736921: Inconsistent and unsafe FLUSH behavior in terms of replication) (8.0)

percona-ysorokin:


percona-ysorokin:
Jenkins build link
https://ps2.cd.percona.com/view/8.0/job/percona-server-8.0-param/19/

`binlog_nogtid.binlog_persist_only_variables`, `binlog_nogtid.binlog_persist_variables` and `main.all_persisted_variables` already fixed and re-recorded.

mysql-test/r/all_persisted_variables.result
```diff
@@ -52,9 +52,9 @@ include/assert.inc ['Expect 413 persisted variables in the table. Due to open Bu
 ************************************************************
 # restart
 
-include/assert.inc ['Expect 353 persisted variables in persisted_variables table.']
-include/assert.inc ['Expect 353 persisted variables shown as PERSISTED in variables_info table.']
-include/assert.inc ['Expect 353 persisted variables with matching peristed and global values.']
+include/assert.inc ['Expect 410 persisted variables in persisted_variables table.']
```
laurynas-biveinis, line 55: 60 variable difference?!

mysql-test/suite/binlog_gtid/t/percona_binlog_skip_flush_commands.test
```diff
@@ -0,0 +1,204 @@
+--source include/have_log_bin.inc
```
laurynas-biveinis, line 1: Ideally there could be a binlog_nogtid test too that drops the check for gtids, but I am not sure whether effort/payoff ratio is good enough for that

sql/mysqld.cc
```diff
@@ -1036,6 +1036,7 @@ bool binlog_gtid_simple_recovery;
 ulong binlog_error_action;
 const char *binlog_error_action_list[] = {"IGNORE_ERROR", "ABORT_SERVER",
                                           NullS};
+bool opt_binlog_skip_flush_commands = 0;
```
laurynas-biveinis: !!!

```suggestion
bool opt_binlog_skip_flush_commands = false;
```

laurynas-biveinis (changes requested):

mysql-test/r/all_persisted_variables.result
percona-ysorokin: Here, assert **message** hasn't been updated for a long time. For some reason, upon adding a new var, developers were updating only assert **condition**. Before this commit the message should have been set to 409, but it wasn't.

sql/mysqld.cc
percona-ysorokin: Fixed

percona-ysorokin:
Updated Jenkins build link with selected tests
https://ps2.cd.percona.com/view/8.0/job/percona-server-8.0-param/23/

PR #2988, opened 2019-02-06
PS-5395: Huge memory usage for audit plugin when large queries are

gl-sergei:
executed

Given 10 threads each running 10 MiB queries audit plugin allocates 400
MiB of RAM.

This is because of the following:

1. Large buffer is allocated per thread
2. Its size is 4x record size
3. Record size is greater than 10 MiB
4. 10 threads are running queries

10 MiB * 4 * 10 = 400 MiB

This patch makes audit plugin to reserve less memory for the record
buffer (roughly equal the query size) down total memory to 100 MiB in
our case.

### DIFF
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -636,7 +636,8 @@ char *audit_log_general_record(char *buf, size_t buflen,
     endptr= endbuf;
     query= escape_string(event->general_query, event->general_query_length,
                          endptr, endbuf - endptr, &endptr, &full_outlen);
-    full_outlen+= full_outlen * my_charset_utf8mb4_general_ci.mbmaxlen;
+    full_outlen*= my_charset_utf8mb4_general_ci.mbmaxlen;
+    full_outlen+= query_length * my_charset_utf8mb4_general_ci.mbmaxlen;
   }
 
   user= escape_string(event->general_user, event->general_user_length,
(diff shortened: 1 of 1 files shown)


gl-sergei:
https://ps2.cd.percona.com/view/5.6/job/percona-server-5.6-param/8/

plugin/audit_log/audit_log.c
```diff
@@ -636,7 +636,8 @@ char *audit_log_general_record(char *buf, size_t buflen,
     endptr= endbuf;
     query= escape_string(event->general_query, event->general_query_length,
                          endptr, endbuf - endptr, &endptr, &full_outlen);
-    full_outlen+= full_outlen * my_charset_utf8mb4_general_ci.mbmaxlen;
+    full_outlen*= my_charset_utf8mb4_general_ci.mbmaxlen;
```
laurynas-biveinis, line 639: Can you explain this change?

gl-sergei: It makes the `else` branch estimate `full_outlen` more accuratelty. After executing this code:
```
  if (query_length < (size_t) (endbuf - endptr))
  {
    uint errors;
    query_length= my_convert(endptr, query_length,
                             &my_charset_utf8mb4_general_ci,
                             event->general_query,
                             event->general_query_length,
                             event->general_charset, &errors);
    query= endptr;
    endptr+= query_length;

    full_outlen+= query_length;

    query= escape_string(query, query_length, endptr, endbuf - endptr,
                         &endptr, &full_outlen);
  }
```
```
full_outlen = length(utf-8 encoded query) + length(utf8-encoded escaped query)
```

this is what we do now in "else" branch:
```
full_length = query_length * 4 + escaped_query_length * 4
```
It was a bug which was mitigated by multiplying total estimate by 4. It actually showed up in audit_log_charset test. In that test utf8 encoded query was roughly 4 times longer than source query.

laurynas-biveinis (approved):

PR #2990, opened 2019-02-06
Fix PS-5395 in 8.0

gl-sergei:
https://ps2.cd.percona.com/view/8.0/job/percona-server-8.0-param/22/

Looks like audit log tests are not run by default. Test case failures don't seem to be related to this patch. Something is wrong with logging connects and disconnects it seems.

### DIFF
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -636,7 +636,8 @@ char *audit_log_general_record(char *buf, size_t buflen,
     endptr= endbuf;
     query= escape_string(event->general_query, event->general_query_length,
                          endptr, endbuf - endptr, &endptr, &full_outlen);
-    full_outlen+= full_outlen * my_charset_utf8mb4_general_ci.mbmaxlen;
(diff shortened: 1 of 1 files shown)


laurynas-biveinis:
@gl-sergei, audit_log is in the default MTR suite list, it is bad if it's not run and we don't know it. Perhaps you have a Jenkins link showing tests not being run?

gl-sergei:
I was trying to run single test from the test suite, but mtr complained that the test was not in the known test suites. Then I specified audit_log.test_name and mtr has found it. Could be just some local glitch maybe?

laurynas-biveinis:
This looks like a recent MTR logic change that I don't really understand (check sub main). If no testcase option is passed, then "default" suites are used, including i.e. audit_log. If --do-test, --do-suite, or a test list of passed, then "all" suites are used, which is something 1) different from "default", 2) populated by walking the mysql-test/ dir - skipping the plugin dirs, hence not really "all".

laurynas-biveinis (approved):

PR #3000, opened 2019-02-13
Add thd_wait_begin & thd_wait_end call for BINLOG dump

anysql:
To decrease the active thread count under thread pool mode, else active_thread_pool may not be zero for specific thread group

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -1929,15 +1929,19 @@ bool dispatch_command(THD *thd, const COM_DATA *com_data,
       break;
     case COM_BINLOG_DUMP_GTID:
       // TODO: access of protocol_classic should be removed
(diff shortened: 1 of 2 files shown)


laurynas-biveinis:
Can you elaborate? Why binlog dump thread should not count towards thread pool group size? And if it should not, perhaps it should be implemented through checking the thread type, as thd_wait_begin/end has different semantics?

anysql:
if requests arrived and there is now active thread (active_thread_count = 0), thread pool will create a new worker for new message right now. But if the binlog dump thread is running (it will never return unless binlog dump end), the active_thread_count will never be 0, and there will be no worker thread for newly arrived requests.

anysql:
thd_wait_begin means that I am waiting io resource, don't think I am a active thread for other requests.

PR #3003, opened 2019-02-14
[Feature] File Purge Asynchronously

xpchild07:
Summary:
--------
InnoDB big data file unlink will cause serious stable problem on POSIX file system,
So InnoDB launch a background thread to truncate data file little by little.

The data file will firstly renamed into a temporary file when single tablespace is deleted,
and write a new DDL log to promise  atomicity,
then File Purge thread will truncate file asynchronously and slowly.

Mainly for DROP/TRUNCATE command;

Several variables control the behaviors:

INNODB_DATA_FILE_PURGE
	Whether enable the async purge strategy
INNODB_DATA_FILE_PURGE_DIR
	Temporary file directory
INNODB_DATA_FILE_PURGE_INTERVAL
	Purge time interval (by milliseconds)
INNODB_DATA_FILE_PURGE_MAX_SIZE
	Purge max size every time (by MB)
INNODB_PRINT_DATA_FILE_PURGE_PROCESS
	Print the process of file purge worker


xpchild07:
Hi， guys
I didn't know how to contribute to percona-server, so just pull a request here. If any wrong with it,
pls give some clue.

Thanks a lot!

Jianwei.zhao

satya-bodapati (changes requested):
Some comments and questions

1. "InnoDB big data file unlink will cause serious stable problem on POSIX file system"
     - can you elaborate the stability problem? Any references? bug#? is this problem with all linux kernels?

2. Avoid using the word "purge", it is so frequently used in InnoDB, it can be confusing to users.

3. I think this commit breaks Atomicity of DDLs. The thread_id used to write DDL log records is very important concept.  You can't just use one fixed thread id (the thread id of the background thread).

4. Incase you solve it (the design issue), please add testcases to simulate crash during delete and show DDL recovers fine.

xpchild07:
> Some comments and questions
> 
> 1. "InnoDB big data file unlink will cause serious stable problem on POSIX file system"
>    
>    * can you elaborate the stability problem? Any references? bug#? is this problem with all linux kernels?
> 2. Avoid using the word "purge", it is so frequently used in InnoDB, it can be confusing to users.
> 3. I think this commit breaks Atomicity of DDLs. The thread_id used to write DDL log records is very important concept.  You can't just use one fixed thread id (the thread id of the background thread).
> 4. Incase you solve it (the design issue), please add testcases to simulate crash during delete and show DDL recovers fine.


Hi, Satya

Thanks for your comments, I try to answer questions.

1. On Alibaba Cloud platform, we deployed many RDS MySQL instances on a machine,  When use EXT4 on Linux, if some instance operated a big table drop/truncate, other instances will be affected. Maybe there is a lock contention on EXT4 space allocation/deallocation.

2. 'Purge' maybe have confusion on codes,  I can change a new word.

3. The thread id is the reserved thread id, no user thread will use it.

4.  Sure,  I will strengthen the test cases to prove the atomicity.

then, soon I will print the Linux kernel stack about EXT4 FS lock contention when unlink big file.


Jianwei.zhao

laurynas-biveinis:
@satya-bodapati , on question 1), see also https://dom.as/2016/05/02/on-removing-files/

xpchild07:
> @satya-bodapati , on question 1), see also https://dom.as/2016/05/02/on-removing-files/

yeah，we often encountered this problem on product environment, we used Linux 2.6 kernel mostly.

xpchild07:
image


image

@laurynas-biveinis  @satya-bodapati   see this perf picture.

PR #3008, opened 2019-02-17
GPL version of Data Masking for MySQL v8.0

fmbiete:
This is a GPL implementation of Oracle MySQL Enterprise Data Masking plugin. 
It follows the API described in the documentation.


fmbiete:
Travis jobs are failing compiling other features. As far as I can see, clang and gcc 4.8 in Linux don't include the required support to compile MySQL server.

bajrang0789:
are there any progress on this one, to support the `data_masking.so` plugin on Official Percona Repo. (for MySQL 5.7 as well) ?

fmbiete:
> are there any progress on this one, to support the `data_masking.so` plugin on Official Percona Repo. (for MySQL 5.7 as well) ?

Waiting for Percona review, after that I will add 5.7 support

laurynas-biveinis:
Thank you for this pull request! It is still in our queue and we will process it, unfortunately at the moment I cannot give an ETA

plugin/data_masking/src/udf/udf_gen_blacklist.cc
```diff
@@ -0,0 +1,120 @@
...
+ */
+static std::string _gen_blacklist(const char *str, const char *dictionary_name,
+                                  const char *replacement_dictionary_name) {
+  std::string res(str);
+  std::string s_dictname_a(dictionary_name);
+  mysql::plugins::tolower(s_dictname_a);
+  std::string s_dictname_b(replacement_dictionary_name);
+  mysql::plugins::tolower(s_dictname_b);
+
+  // TODO: do we need to lock the mutex to read the data??? this would be a
```
satya-bodapati: Yes, otherwise how do you prevent a concurrent removal of the dictionary?
Other simple idea to improve performance is to use rwlocks

plugin/data_masking/include/udf/udf_gen_range.h
```diff
@@ -0,0 +1,23 @@
...
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software Foundation,
+   51 Franklin Street, Suite 500, Boston, MA 02110-1335 USA */
+
+#define _PLUGIN_DATAMASKING_GENRANGE_H
+
+#include "plugin/data_masking/include/udf/udf_descriptor.h"
+
+udf_descriptor udf_gen_range();
```
satya-bodapati, line 21: I would suggest to aggegrate all the  generate functions (udf_*gen_*()) into one .h and .cc file.

plugin/data_masking/src/udf/udf_gen_dictionary.cc
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * @return A random term from the dictionary as a string, or NULL if the
+ * dictionary name is not in the dictionary registry.
+ */
+static std::string _gen_dictionary(const char *dictionary_name) {
+  std::string res = "";
+  std::string s_dictname(dictionary_name);
+  mysql::plugins::tolower(s_dictname);
+
+  // TODO: do we need to lock the mutex to read the data??? this would be a
```
satya-bodapati: Same as above. Answer is yes.
Preferrably use RW locks.

plugin/data_masking/src/udf/udf_gen_dictionary.cc
```diff
@@ -0,0 +1,94 @@
...
+}
+
+static char *gen_dictionary(UDF_INIT *, UDF_ARGS *args, char *result,
+                            unsigned long *length, char *is_null, char *) {
+  DBUG_ENTER("gen_dictionary");
+
+  std::string res = _gen_dictionary(args->args[0]);
+  *length = res.size();
+  if (!(*is_null = (*length == 0))) {
+    strcpy(result, res.c_str());
```
satya-bodapati: How are we sure that result has memory of  "length" bytes allocated?

plugin/data_masking/src/udf/udf_gen_rnd_us_phone.cc
```diff
@@ -0,0 +1,74 @@
...
+static char *gen_rnd_us_phone(UDF_INIT *initid,
+                              UDF_ARGS *args MY_ATTRIBUTE((unused)),
+                              char *result MY_ATTRIBUTE((unused)),
+                              unsigned long *length, char *is_null,
+                              char *is_error) {
+  DBUG_ENTER("gen_rnd_us_phone");
+
+  std::string phone = mysql::plugins::random_us_phone();
+  *length = phone.size();
+  initid->ptr = new char[*length + 1];
```
satya-bodapati, line 64: When is this memory released?

plugin/data_masking/src/udf/udf_mask_ssn.cc
```diff
@@ -0,0 +1,89 @@
...
+    *is_null = 1;
+  } else {
+    const char masking_char = 'X';
+    const unsigned int unmasked_chars_end = 4;
+
+    std::string s(args->args[0]);
+    s = mysql::plugins::mask_inner(args->args[0], args->lengths[0], 0,
+                                   unmasked_chars_end, masking_char);
+    *length = s.length();
+    initid->ptr = new char[*length + 1];
```
satya-bodapati, line 76: is there memory leak here?
Lets say the first mask_ssn() set the initd->ptr to some memory location of 10 bytes (say). The next mask_ssn() would set the initd->ptr to new location without freeing the old one?

satya-bodapati: Or would the denit() be called for every UDF call()?

plugin/data_masking/src/udf/udf_mask_ssn.cc
```diff
@@ -0,0 +1,89 @@
...
+  initid->maybe_null = 1;
+  initid->ptr = NULL;
+
+  DBUG_RETURN(false);
+}
+
+static void mask_ssn_deinit(UDF_INIT *initid) {
+  DBUG_ENTER("mask_ssn_deinit");
+
+  if (initid->ptr) free(initid->ptr);
```
satya-bodapati: memory allocated with new(), do not use free().

Just use  delete initd->ptr

satya-bodapati (commented):
Few more requests
1. Add a test with view example as shown here at end https://docs.oracle.com/cd/E17952_01/mysql-8.0-en/data-masking-usage.html

2. Does this work with multi-byte character sets? If so, add some tests with non-ascii characters. If not, add tests to ensure it doesn't crash server.

plugin/data_masking/src/udf/udf_gen_rnd_us_phone.cc
fmbiete: deinit is automatically called after each call

plugin/data_masking/src/udf/udf_mask_ssn.cc
fmbiete: yes, deinit is automatically called after each UDF call to clean up

plugin/data_masking/src/udf/udf_mask_ssn.cc
fmbiete: Thank you, done!

plugin/data_masking/src/udf/udf_gen_dictionary.cc
fmbiete: Thank you, done!

plugin/data_masking/src/udf/udf_gen_blacklist.cc
fmbiete: Thank you, done!

plugin/data_masking/include/udf/udf_gen_range.h
fmbiete: It started like that, but it was a huge file. I think it's more easy to maintain and extend if the methods are in different smaller files.

plugin/data_masking/src/udf/udf_gen_dictionary.cc
fmbiete: Good catch. result has a fixed length. I replaced this with initid->ptr to allow using bigger dictionary terms.

satya-bodapati:
Thank you for addressing the review comments. Your patch looks good. :+1: 

You might have to squash this into single commit. I will check and get back to you.

satya-bodapati (approved):

plugin/data_masking/src/udf/udf_registration.cc
```diff
@@ -0,0 +1,139 @@
+/* Copyright (c) 2018, 2019 Francisco Miguel Biete Banon. All rights reserved.
```
dutow: This file (and some others) seems to be the same as in the group replication plugin?

Please retain the Oracle copyright for files copied from Oracle plugins, and include Oracle in the plugin authors.

dutow (changes requested):

plugin/data_masking/src/udf/udf_registration.cc
satya-bodapati: May I know why this conversation is marked resolved? Isn't the copyright requirement not applicable anymore?

satya-bodapati:
Request to squash the changes into single commit. After this QA will test the PR

plugin/data_masking/src/udf/udf_registration.cc
fmbiete: Sorry, I forgot to commit-push

plugin/data_masking/include/udf/udf_registration.h
```diff
@@ -0,0 +1,22 @@
+#ifndef _PLUGIN_DATA_MASKING_UDFREGISTRATION_H
+/* Copyright (c) 2018, 2019 Francisco Miguel Biete Banon. All rights reserved.
```
dutow: Shouldn't this file have the same copyright as the cc file?

plugin/data_masking/src/plugin.cc
```diff
@@ -0,0 +1,75 @@
...
+struct st_mysql_daemon data_masking_plugin = {MYSQL_DAEMON_INTERFACE_VERSION};
+
+//#define mysql_declare_plugin_end   ,{0,0,0,0,0,0,0,0,0,0,0,0,0,0}} <- 14 v8.0
+//#define mysql_declare_plugin_end   ,{0,0,0,0,0,0,0,0,0,0,0,0,0}}   <- 13 v5.7
+
+mysql_declare_plugin(data_masking){
+    MYSQL_DAEMON_PLUGIN,             // Type
+    &data_masking_plugin,            // info struct for plugin type
+    "data_masking",                  // Short name
+    "Francisco Miguel Biete Banon",  // Author
```
dutow, line 64: There's an author field here too

plugin/data_masking/include/udf/udf_registration.h
fmbiete: Added Oracle

plugin/data_masking/src/plugin.cc
fmbiete: I'm not sure about this one, wouldn't this that mean that Oracle is the author of this plugin? but added anyway.

satya-bodapati: I don't think Oracle should be put here.. @dutow what was your concern here?

satya-bodapati: You can remove Oracle here

satya-bodapati:
Hello @fmbiete 

Just want to update that your patch is under testing now

plugin/data_masking/src/plugin.cc
fmbiete: Removed. Thank you

dutow:
@fmbiete I updated your branch. Please update your local checkout and check it.

Changes:
* Added missing DBUG_RETURN macros
* Fixed the tests so they work when executed from the build directory
* Rerecorded the view mtr result, as it was failing
* Squashed your two commits
* Rebased on the current 8.0 head

dutow:
@fmbiete @satya-bodapati I rebased on 8.0.17 and reformatted with clang-format-8.

I'll merge the code once jenkins completes.

Jenkins job: https://ps80.cd.percona.com/job/percona-server-8.0-param/377/

satya-bodapati:
> @fmbiete @satya-bodapati I rebased on 8.0.17 and reformatted with clang-format-8.
> 
> I'll merge the code once jenkins completes.
> 
> Jenkins job: https://ps80.cd.percona.com/job/percona-server-8.0-param/377/

Thanks @dutow

dutow:
I've fixed another issue I didn't notice locally: the install command for the tests was incorrect and unnecessary.

New jenkins: https://ps80.cd.percona.com/job/percona-server-8.0-param/380/

PR #3012, opened 2019-02-19
JEN-795 Failed connect to vault.public-ci.percona.com:8600

robgolebiowski:
vault.public-ci.percona was deemed to be unsecure and thus removed.
keyring_vault MTR test should use https://vault.percona.com, which is
maintained by IIT team. The tests were amended to use this new server.
The policy was changed to more strict one in comparision to
vault.public-ci.percona - the tests were also amended to cope with the
new policy.

Also new tests were added to test the separation of keyrings inside a
Vault server on directory level. For those tests only one new mount
point is created on the server and the separation is done by creating
separate directories inside the mount point - one directory per each
Percona server keyring_vault keyring.

The old tests are still checking the separation of keyrings with use of
separate mount points.

The not_repeat.inc was added to check whether the tests are not run with
--repeat > 1. This is because the tests are creating mount points at
the beginning of a run and delete when at the end of the run. When repeat is
used server will remember that it has already created master key and will
try to fetch the key. Unfortunately the key will be gone because the
mount point was deleted and re-created.

Changed the number of times the Master Key is rotated in
keyring_vault.table_encrypt_3.test and
keyring_vault.table_encrypt_3_directory to 100, so not to DoS the Vault Server.

Also clang-format-5.0 was applied.


plugin/keyring_vault/tests/mtr/delete_secret_directory.inc
```diff
@@ -0,0 +1,119 @@
...
+  # double quotes and delete the secret. When all secrets are removed from a directory Vault will remove the directory itself.
+
+  my $CURL_LIST_RESULT;
+  open(CURL_LIST_RESULT, "$vardir/tmp/curl_list_result") or die("Could not open file with list of secrets in a mount point. Please remove directory $secret_mount_point manualy.\n");
+  while (my $row = <CURL_LIST_RESULT>)
+  {
+    print $help_file $row;
+    if ($row =~ m/keys":\[(.*)\]/)
+    {
+      #print $1;
```
laurynas-biveinis:
```suggestion
```

laurynas-biveinis (approved):
I have one comment but I am not sure re-pushing / re-upmerging / re-testing is worth addressing it

plugin/keyring_vault/tests/mtr/innodb_online_alter_encryption.test
```diff
@@ -1,6 +1,7 @@
 --source include/have_keyring_vault_plugin.inc
+--source include/not_repeat.inc
```
percona-ysorokin, line 2: Did you try adding `--force-restart` to the `.opt` file instead of this `not_repeat.inc`?

plugin/keyring_vault/tests/mtr/table_encrypt_3_directory-master.opt
```diff
@@ -0,0 +1 @@
+--testcase-timeout=60
```
percona-ysorokin, line 1: Is this a valid option for a `.opt` file?

cmake/curl.cmake
```diff
@@ -51,3 +51,15 @@ MACRO(MYSQL_CHECK_CURL)
    ENDIF()
   ENDIF()
 ENDMACRO()
+
```
percona-ysorokin, line 54: @robgolebiowski Once again, do we really need this RTMP checking? At least on Ubuntu Bionic unit tests work fine even though curl library there depends on rtmp.

robgolebiowski: @percona-ysorokin It is possible that they fixed it already. I remember it was failing a lot on centos6. There is JIRA for it PS-4757. Will try to take care of that as a background task.

plugin/keyring_vault/tests/mtr/table_encrypt_3_directory-master.opt
robgolebiowski: yes, it is used quite a lot

laurynas-biveinis:
As discussed on Slack, merging now for better Vault testing, Yura's review comments will be addressed once Robert returns

PR #3014, opened 2019-02-19
PS-5044 : Merge MySQL 8.0.14 & 8.0.15

george-lorch:
- Fixed minor Travis CI complilation issues for clang and gcc 4.8.5 introduced
  by upstream.

### DIFF
--- a/sql/dd/impl/properties_impl.h
+++ b/sql/dd/impl/properties_impl.h
@@ -83,8 +83,14 @@ class Properties_impl : public Properties {
  public:
   Properties_impl() = default;
 
+  Properties_impl(const Properties_impl &o)
+      : Properties(o), m_map(o.m_map), m_keys(o.m_keys) {}
+
   /* Constructor accepting a set of valid keys. */
-  Properties_impl(const std::set<String_type> &keys) : m_keys(keys) {}
+  Properties_impl(const std::set<String_type> &keys)
+      : Properties(), m_keys(keys) {}
+
+  Properties_impl &operator=(const Properties_impl &) = delete;
 
   virtual const Properties_impl *impl() const { return this; }
 
--- a/plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/task_net.c
+++ b/plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/task_net.c
@@ -84,8 +84,8 @@ int checked_getaddrinfo(const char *nodename, const char *servname,
   _hints.ai_family = AF_UNSPEC;
   _hints.ai_socktype = SOCK_STREAM;  // TCP stream sockets
   if (hints == NULL) hints = &_hints;
-  for (int attempt_nr = 0;
-       errval == EAI_AGAIN && attempt_nr < NR_GETADDRINFO_ATTEMPTS;
+  int attempt_nr = 0;
+  for (; errval == EAI_AGAIN && attempt_nr < NR_GETADDRINFO_ATTEMPTS;
        attempt_nr++) {
     if (*res) {
       freeaddrinfo(*res);
--- a/sql/dd/properties.h
+++ b/sql/dd/properties.h
@@ -72,6 +72,10 @@ namespace dd {
 
 class Properties {
  public:
+  Properties() = default;
+
+  Properties(const Properties &) = default;
+
   Properties &operator=(const Properties &) = delete;
 
   /**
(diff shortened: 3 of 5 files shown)


router/src/harness/include/mysql/harness/mpsc_queue.h
```diff
@@ -43,7 +43,7 @@ class MPSCQueueDV {
  public:
   using value_type = T;
 
-  MPSCQueueDV()
+  MPSCQueueDV() noexcept
```
laurynas-biveinis: What does this fix and why is this correct - given "new Node" below?

sql/dd/impl/properties_impl.h
```diff
@@ -83,8 +83,14 @@ class Properties_impl : public Properties {
  public:
   Properties_impl() = default;
 
+  Properties_impl(const Properties_impl &o)
```
laurynas-biveinis, line 86: = default;
?

laurynas-biveinis:
The majority of (or all?) the issues here look upstream and should be reported there too

breisig:
I am assuming the new fixes with fix the compiling. I guess the tests would need to be restarted again before it can be merged?  It would be nice to get it working so we can get a 8.0.15 released as soon as possible. :-)

laurynas-biveinis:
@breisig, these are non-default compiler warning fixes, not on the critical path for the release

sql/dd/impl/properties_impl.h
george-lorch: "= default" was not sufficient to silence the "error: definition of implicit copy constructor for 'Properties' is deprecated because it has a user-declared destructor" "Rule of 3" error on all platforms.  Only adding an explicit copy constructor solved this.

router/src/harness/include/mysql/harness/mpsc_queue.h
george-lorch: It addresses this error emitted from gcc 4.8 only:
/home/travis/build/percona/percona-server/router/src/harness/src/loader.cc:27:/home/travis/build/percona/percona-server/router/src/harness/include/mysql/harness/mpsc_queue.h: In instantiation of ‘mysql_harness::MPSCQueueDV<T>::MPSCQueueDV() [with T = std::__exception_ptr::exception_ptr]’:/home/travis/build/percona/percona-server/router/src/harness/include/mysql/harness/waiting_queue_adaptor.h:38:7:   required from here/home/travis/build/percona/percona-server/router/src/harness/include/mysql/harness/mpsc_queue.h:111:10: error: use of deleted function ‘std::atomic<_Tp*>::atomic(const std::atomic<_Tp*>&) [with _Tp = mysql_harness::MPSCQueueDV<std::__exception_ptr::exception_ptr>::Node]’   struct Node {          ^In file included from /home/travis/build/percona/percona-server/router/src/harness/include/mysql/harness/mpsc_queue.h:28:0,                 from /home/travis/build/percona/percona-server/router/src/harness/include/mysql/harness/loader.h:626,                 from /home/travis/build/percona/percona-server/router/src/harness/src/loader.cc:27:/usr/include/c++/4.8/atomic:298:7: error: declared here       atomic(const atomic&) = delete;       ^In file included from /home/travis/build/percona/percona-server/router/src/harness/include/mysql/harness/loader.h:626:0,                 from /home/travis/build/percona/percona-server/router/src/harness/src/loader.cc:27:/home/travis/build/percona/percona-server/router/src/harness/include/mysql/harness/mpsc_queue.h:47:69: note: synthesized method ‘mysql_harness::MPSCQueueDV<std::__exception_ptr::exception_ptr>::Node::Node()’ first required here        : head_{new Node}, tail_{head_.load(std::memory_order_relaxed)} {}

I am unsure why this only trips on gcc 4.8.  Stolen from a stackoverflow on this identical question
> std::atomic's default constructor is noexcept, so when you try to instantiate it with some type, that type also needs to be noexcept default constructible so that the exception specifications match.

laurynas-biveinis: OK, but the 2nd half of my question stands: why is this correct, given "new Node" in constructor body which may throw std::bad_alloc?

george-lorch: I have no idea why, only that this fixed the error.  Considering that his was limited to gcc 4.8, I suspect that this is a compiler implementation fail.

manuelm: Just ran into this. Looks like https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60258#c4

Adding `noexcept` doesn't fix the compile error here.

george-lorch:
OK, I re-pushed with the "noexcept" reverted.  TravicCI now runs clean for all but gcc 4.8.

laurynas-biveinis (approved):

PR #3016, opened 2019-02-25
PS-5333: Remove need for innodb_encrypt_tables to ON/FORCE when

robgolebiowski:
mysql.ibd is encrypted

PS-5117 introduced requirement that innodb_encrypt_tables need to be set
to ON/FORCE on server startup when mysql.ibd is encrypted. This
requirement was added because there was a need for mysql.ibd space flags
validation against hardcoded, predefined flags. In PS mysql.ibd can be
encrypted and have encrypted flag set - thus generating a mismatch with
predefined flags. In order to check whether mysql.ibd is encrypted we
will check flags from the first page of the tablespace instead of
relaying on innodb_encrypt_tables. Apart from improvement from user
point of view it will also ease keyring encryption threads development
in 8.0 as encryption threads can also change the mysql.ibd's encryption
flag.

In keyring encryption there is a possibility that update to mysql.ibd's
page0 will reside only inside redo log (due to the crash).
Added code for reading page0 encryption flag from redo log and using it
to update mysql.ibd's DD flag. This code will be tested when keyring
encryption gets fixed.

### DIFF
--- a/sql/dd/types/init_mode.h
+++ b/sql/dd/types/init_mode.h
@@ -1,0 +1,37 @@
+/*
+   Copyright (c) 2000, 2018, Oracle and/or its affiliates. All rights reserved.
+   Copyright (c) 2019 Percona LLC and/or its affiliates. All rights reserved.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License, version 2.0,
+   as published by the Free Software Foundation.
+
+   This program is also distributed with certain software (including
+   but not limited to OpenSSL) that is licensed under separate terms,
+   as designated in a particular file or component or in included license
+   documentation.  The authors of MySQL hereby grant you an additional
+   permission to link the program and your derivative works with the
+   separately licensed software that they have included with MySQL.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License, version 2.0, for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA
+*/
+
+#ifndef DD_INIT_MODE_T_H_INCLUDED
+#define DD_INIT_MODE_T_H_INCLUDED
+
+/** Mode for initializing the data dictionary. */
+enum dict_init_mode_t {
+  DICT_INIT_CREATE_FILES,      //< Create all required SE files
+  DICT_INIT_CHECK_FILES,       //< Verify existence of expected files
+  DICT_INIT_UPGRADE_57_FILES,  //< Used for upgrade from mysql-5.7
+  DICT_INIT_IGNORE_FILES       //< Don't care about files at all
+};
+
+#endif  // DD_INIT_MODE_T_H_INCLUDED
--- a/storage/innobase/dict/dict0dd.cc
+++ b/storage/innobase/dict/dict0dd.cc
@@ -68,6 +68,7 @@ Data dictionary interface */
 #include "query_options.h"
 #include "sql_base.h"
 #include "sql_table.h"
+#include "thd_raii.h"
 #endif /* !UNIV_HOTBACKUP */
 
 const char *DD_instant_col_val_coder::encode(const byte *stream, size_t in_len,
@@ -6082,4 +6083,119 @@ bool dd_is_table_in_encrypted_tablespace(const dict_table_t *table) {
     return false;
   }
 }
+
+/* Updates tablespace's DD flags.
+@param[in] Thread       THD
+@param[in] space_name   name of the space that DD flags are to be updated
+@param[in] update       function object that will be invoked for updating DD
+flags
+@return false on success */
+static bool dd_update_tablespace_dd_flags(
+    THD *thd, const char *space_name, std::function<void(uint32 &)> update) {
+  Disable_autocommit_guard autocommit_guard(thd);
+  dd::cache::Dictionary_client *client = dd::get_dd_client(thd);
+  dd::cache::Dictionary_client::Auto_releaser releaser(client);
+
+  dd::Tablespace *dd_space = nullptr;
+
+  trx_t *trx = check_trx_exists(thd);
+  trx_start_if_not_started(trx, true);
+
+  if (dd::acquire_exclusive_tablespace_mdl(thd, space_name, false)) {
+    ut_a(false);
+  }
+
+  if (client->acquire_for_modification<dd::Tablespace>(space_name, &dd_space) ||
+      dd_space == nullptr) {
+    return (true);
+  }
+
+  uint32_t dd_space_flags = 0;
+
+  if (dd_space->se_private_data().get(dd_space_key_strings[DD_SPACE_FLAGS],
+                                      &dd_space_flags)) {
+    return (true);
+  }
+
+  update(dd_space_flags);
+
+  /* Update DD flags for tablespace */
+  dd_space->se_private_data().set(dd_space_key_strings[DD_SPACE_FLAGS],
+                                  static_cast<uint32>(dd_space_flags));
+
+  if (dd::commit_or_rollback_tablespace_change(thd, dd_space, false)) {
+    return (true);
+  }
+
+  return (false);
+}
+
+bool dd_set_encryption_flag(THD *thd, const char *space_name) {
+  auto update_func = [](uint32_t &dd_space_flags) {
+    dd_space_flags |= (1U << FSP_FLAGS_POS_ENCRYPTION);
+  };
+  return dd_update_tablespace_dd_flags(thd, space_name, update_func);
+}
+
+bool dd_clear_encryption_flag(THD *thd, const char *space_name) {
+  auto update_func = [](uint32_t &dd_space_flags) {
+    dd_space_flags &= ~(1U << FSP_FLAGS_POS_ENCRYPTION);
+  };
+  return dd_update_tablespace_dd_flags(thd, space_name, update_func);
+}
+
+static bool dd_get_tablespace_flags(THD *thd, const char *space_name,
+                                    uint32_t &dd_space_flags) {
+  const dd::Tablespace *dd_space = nullptr;
+  dd::cache::Dictionary_client::Auto_releaser releaser(thd->dd_client());
+
+  return thd->dd_client()->acquire(space_name, &dd_space) ||
+         dd_space == nullptr ||
+         dd_space->se_private_data().get(dd_space_key_strings[DD_SPACE_FLAGS],
+                                         &dd_space_flags);
+}
+
+/* Sets tablespace's DD flags.
+@param[in] Thread       THD
+@param[in] space_name   name of the space for which DD flags are to be set
+@param[in] space_flags  DD flags that are to be assigned to space */
+static bool dd_set_flags(THD *thd, const char *space_name,
+                         const uint32_t space_flags) {
+  auto set_flags = [](uint32_t &dd_space_flags, uint32_t space_flags) {
+    // currently we are using this function only for correcting encryption flag
+    ut_ad(dd_space_flags == space_flags ||
+          FSP_FLAGS_GET_ENCRYPTION(dd_space_flags) !=
+              FSP_FLAGS_GET_ENCRYPTION(space_flags));
+    dd_space_flags = space_flags;
+  };
+  auto update_func = std::bind(set_flags, std::placeholders::_1, space_flags);
+  return dd_update_tablespace_dd_flags(thd, space_name, update_func);
+}
+
+bool dd_fix_mysql_ibd_encryption_flag_if_needed(THD *thd,
+                                                uint32_t space_flags) {
+  uint32_t dd_space_flags;
+  if (dd_get_tablespace_flags(thd, dict_sys_t::s_dd_space_name,
+                              dd_space_flags)) {
+    return true;
+  }
+  if (FSP_FLAGS_GET_ENCRYPTION(dd_space_flags) ==
+      FSP_FLAGS_GET_ENCRYPTION(space_flags)) {
+    return false;
+  }
+  // exclude encryption flag from validation
+  dd_space_flags &= ~FSP_FLAGS_MASK_ENCRYPTION;
+  space_flags &= ~FSP_FLAGS_MASK_ENCRYPTION;
+  if (dd_space_flags != space_flags) {
+    // this should not happen - some other flags other than encryption flag are
+    // mismatched
+    ib::error(ER_IB_MSG_394)
+        << "Flags read from mysql.ibd file (" << space_flags
+        << ") are different from the flags read from DD (" << dd_space_flags
+        << ") This comparission does *not* include the encryption flag.";
+    return true;
+  }
+  return dd_set_flags(thd, dict_sys_t::s_dd_space_name, space_flags);
+}
+
 #endif /* !UNIV_HOTBACKUP */
--- a/storage/innobase/dict/dict0dict.cc
+++ b/storage/innobase/dict/dict0dict.cc
@@ -7450,27 +7450,130 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
   return err;
 }
 
-bool dict_detect_encryption(bool is_upgrade, space_id_t mysql_plugin_space) {
-  bool encrypt_mysql = false;
-  if (is_upgrade) {
-    if (mysql_plugin_space == SYSTEM_TABLE_SPACE) {
-      return (srv_sys_space.is_encrypted());
-    }
-
-    space_id_t space_id = fil_space_get_id_by_name("mysql/plugin");
-    if (space_id == SPACE_UNKNOWN) {
-      return (false);
-    }
-
-    fil_space_t *space = fil_space_get(space_id);
-    ut_ad(space != nullptr);
-
-    if (space == nullptr) {
-      return (false);
-    }
-    encrypt_mysql = FSP_FLAGS_GET_ENCRYPTION(space->flags);
+/** Detect if database has encrypted mysql plugin space. This is
+ *  indication that system tablespace was encrypted.
+@param[in] mysql_plugin_space   space_id of mysql/plugin table.
+@return true if encrypted, false if not (or fail to open mysql/plugin table) */
+static bool dict_is_mysql_plugin_space_encrypted(
+    space_id_t mysql_plugin_space) {
+  if (mysql_plugin_space == SYSTEM_TABLE_SPACE) {
+    return (srv_sys_space.is_encrypted());
   }
 
-  return (encrypt_mysql || srv_encrypt_tables == SRV_ENCRYPT_TABLES_ON ||
-          srv_encrypt_tables == SRV_ENCRYPT_TABLES_FORCE);
+  space_id_t space_id = fil_space_get_id_by_name("mysql/plugin");
+  if (space_id == SPACE_UNKNOWN) {
+    return (false);
+  }
+
+  const fil_space_t *space = fil_space_get(space_id);
+  ut_ad(space != nullptr);
+
+  if (space == nullptr) {
+    return false;
+  }
+  return FSP_FLAGS_GET_ENCRYPTION(space->flags);
+}
+
+/** Detect if innodb_encrypt_tables is set either to ON or FORCE or
+ONLINE_TO_KEYRING*
+@return true if innodb_encrypt_tables is equal to ON or FORCE or
+ONLINE_TO_KEYRING* */
+static bool dict_should_be_keyring_encrypted() {
+  return srv_encrypt_tables == SRV_ENCRYPT_TABLES_ON ||
+         srv_encrypt_tables == SRV_ENCRYPT_TABLES_FORCE ||
+         srv_encrypt_tables == SRV_ENCRYPT_TABLES_ONLINE_TO_KEYRING ||
+         srv_encrypt_tables == SRV_ENCRYPT_TABLES_ONLINE_TO_KEYRING_FORCE;
+}
+
+/** Reads mysql.ibd's page0 from buffer
+@return page0 or nullptr */
+static buf_frame_t *get_mysql_ibd_page_0_from_buffer() {
+  fil_space_t *space = fil_space_acquire_silent(dict_sys_t::s_space_id);
+  if (space == nullptr) {
+    return nullptr;
+  }
+
+  const page_size_t page_size(space->flags);
+  mtr_t mtr;
+  mtr_start(&mtr);
+  buf_block_t *block = buf_page_get(page_id_t(dict_sys_t::s_space_id, 0),
+                                    univ_page_size, RW_X_LATCH, &mtr);
+  mtr_commit(&mtr);
+
+  fil_space_release(space);
+
+  return !block ? nullptr : buf_block_get_frame(block);
+}
+
+/** Reads mysql.ibd's page0 directly from disk
+@param[in] buf - buffer for reading page0 into
+@return page0 or nullptr */
+static buf_frame_t *get_mysql_ibd_page_0_io(byte *buf) {
+  bool successful_read = false;
+
+  pfs_os_file_t file = os_file_create_simple_no_error_handling(
+      innodb_data_file_key, dict_sys_t::s_dd_space_file_name, OS_FILE_OPEN,
+      OS_FILE_READ_ONLY, srv_read_only_mode, &successful_read);
+
+  if (!successful_read) {
+    return nullptr;
+  }
+
+  buf_frame_t *page = static_cast<buf_frame_t *>(ut_align(buf, UNIV_PAGE_SIZE));
+
+  ut_ad(page == page_align(page));
+
+  IORequest request(IORequest::READ);
+  dberr_t err = os_file_read_first_page(request, file, page, UNIV_PAGE_SIZE);
+
+  os_file_close(file);
+
+  return DB_SUCCESS == err ? page : nullptr;
+}
+
+/** Detect if mysql.ibd's page0 has encryption flag set.
+ *  The page 0 either read from buffer (if available) or
+ *  directly from disk.
+@return tuple <0> success - true if no error
+              <1> true if encryption flag is set, false otherwise */
+static std::tuple<bool, bool> dict_mysql_ibd_page_0_has_encryption_flag_set() {
+  //<0> element - success, <1> element - encryption flag
+  auto result = std::make_tuple(false, false);
+  auto buf = ut_make_unique_ptr_null();
+
+  /* read from buffer */
+  buf_frame_t *page0 = get_mysql_ibd_page_0_from_buffer();
+  if (!page0) {
+    /* page0 of mysql.ibd is not in the buffer, try direct io */
+    buf = ut_make_unique_ptr_nokey(2 * UNIV_PAGE_SIZE);
+    if (buf == nullptr || !(page0 = get_mysql_ibd_page_0_io(buf.get()))) {
+      return result;
+    }
+  }
+
+  const ulint flags = fsp_header_get_flags(page0);
+  std::get<0>(result) = true;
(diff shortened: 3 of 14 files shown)


robgolebiowski:
This PR is exactly the same as https://github.com/percona/percona-server/pull/2937. Here, on top of 2937, I also implemented checks for mysql.ibd's page 0 updates in redo log suggested by Satya. This code is not currently tested as there is no scenario in MK encryption where there are updates to encryption flag stored in page0 of mysql.ibd. However, this code is needed for keyring encryption and testes for this code will be added later, when keyring encryption test suite will be able to be run in jenkins.

storage/innobase/dict/dict0dict.cc
```diff
@@ -7450,27 +7450,85 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
+}
+
+static bool dict_should_mysql_ibd_be_created_encrypted() {
+  return srv_encrypt_tables == SRV_ENCRYPT_TABLES_ON ||
+         srv_encrypt_tables == SRV_ENCRYPT_TABLES_FORCE;
+}
+
+static bool dict_should_mysql_ibd_be_opened_as_encrypted(const char *filepath,
+                                                         bool &success) {
+  pfs_os_file_t file = os_file_create_simple_no_error_handling(
```
laurynas-biveinis, line 7514: suspicious: pfs_ types should not be used directly: os_file_t ?

storage/innobase/dict/dict0dict.cc
```diff
@@ -7450,27 +7450,85 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
+      srv_read_only_mode, &success);
+
+  if (!success) {
+    return (false);
   }
 
-  return (encrypt_mysql || srv_encrypt_tables == SRV_ENCRYPT_TABLES_ON ||
-          srv_encrypt_tables == SRV_ENCRYPT_TABLES_FORCE);
+  /* Read the first page of the tablespace */
+  std::unique_ptr<byte, std::function<void(byte *)>> buf(
```
laurynas-biveinis: This looks useful enough to have it in e.g. ut0ut.h: unique_ptr with ut_free deleter (but with struct ut_free_deleter w/ operator() instead of std::function<>/lambda to avoid having state in the smart pointer)

storage/innobase/srv/srv0start.cc
```diff
@@ -2689,6 +2691,20 @@ dberr_t srv_start(bool create_new_db, const std::string &scan_directories) {
 
       log_sys_close();
 
+      // We have set mysql's dd flags based on what we read from page0. This was
```
```suggestion
      // We have set mysql's dd flags based on what we read from page 0. This was
```

laurynas-biveinis (changes requested):

storage/innobase/dict/dict0dict.cc
robgolebiowski: Seems that pfs_os_file_t is used everywhere where os_file_create_simple_no_error_handling is used ...

laurynas-biveinis: Huh. OK then

storage/innobase/dict/dict0dict.cc
robgolebiowski: added to ut0new.h

laurynas-biveinis (approved):

storage/innobase/dict/dict0dict.cc
```diff
@@ -7450,27 +7450,83 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
   return err;
 }
 
-bool dict_detect_encryption(bool is_upgrade, space_id_t mysql_plugin_space) {
-  bool encrypt_mysql = false;
-  if (is_upgrade) {
-    if (mysql_plugin_space == SYSTEM_TABLE_SPACE) {
-      return (srv_sys_space.is_encrypted());
-    }
+static bool dict_should_mysql_ibd_be_created_encrypted_due_to_upgrade(
```
satya-bodapati: Please keep the function names precise. Exact details of function can be added as function comment.

storage/innobase/dict/dict0dict.cc
```diff
@@ -7450,27 +7450,83 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
+    return (false);
+  }
+
+  const ulint flags = fsp_header_get_flags(page);
+  os_file_close(file);
+  success = true;
+  return (FSP_FLAGS_GET_ENCRYPTION(flags));
+}
+
+bool dict_detect_encryption_of_mysql_ibd(dict_init_mode_t dict_init_mode,
```
satya-bodapati, line 7560: Please add doxygen style comments for all new functions

storage/innobase/srv/srv0start.cc
```diff
@@ -2689,6 +2691,20 @@ dberr_t srv_start(bool create_new_db, const std::string &scan_directories) {
...
 
+      // We have set mysql's dd flags based on what we read from page 0. This
+      // was before redo log was applied. Now we have applied changes from redo
+      // log. If there were changes in redo for page 0 of mysql.ibd we check if
+      // encryption flag has changed. If it did we amend mysql's dd encryption
+      // flags accordingly.
+      if (mysql_ibd_page_0_present_in_redo) {
+        fil_space_t *mysql_space = fil_space_get(dict_sys_t::s_space_id);
+        if (mysql_space == nullptr ||
+            dd_fix_mysql_ibd_encryption_flag_if_needed(current_thd,
```
satya-bodapati, line 2705: you are not fixing the fil_space_t flags of mysql tablespace. Only fixing DD is not sufficient.

satya-bodapati (changes requested):

storage/innobase/dict/dict0dict.cc
robgolebiowski: I disagree. Comments then need to change when function changes. I took this from clean code and I will stand by it. Self explanatory names are much better - imho.

storage/innobase/srv/srv0start.cc
robgolebiowski: mysql tablespace fil_space_t flags was already updated by redo log applier.

storage/innobase/dict/dict0dict.cc
robgolebiowski: I disagree. Function names should describe what the function is doing. Comments are additional maintenance burden and thus should be avoided if possible. Thus the code should be self explanatory. I have changed the names to be more generic and describe what the function is doing instead of describing what it is used for.

storage/innobase/srv/srv0start.cc
robgolebiowski: mysql tablespace fil_space_t flags was already updated by redo log applier.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -5392,16 +5386,21 @@ static int innobase_init_files(dict_init_mode_t dict_init_mode,
     btr_search_enabled = old_btr_search_value;
   }
 
-  bool do_encrypt =
-      dict_detect_encryption(srv_is_upgrade_mode, upgrade_mysql_plugin_space);
+  bool do_encrypt = false;
+  bool ret = dict_detect_encryption_of_mysql_ibd(
+      dict_init_mode, upgrade_mysql_plugin_space,
+      dict_sys_t::s_dd_space_file_name, do_encrypt);
```
satya-bodapati: Why do you need to pass a global variable?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -5392,16 +5386,21 @@ static int innobase_init_files(dict_init_mode_t dict_init_mode,
...
+  if (!ret) {
+    ib::error(ER_XB_MSG_2) << "Failed to determine if mysql.ibd is encrypted. "
```
satya-bodapati: You are not supposed to re-use ER_ messages. Every message should introduce a unique id.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -5392,16 +5386,21 @@ static int innobase_init_files(dict_init_mode_t dict_init_mode,
...
+                              "Have you not deleted it?";
```
satya-bodapati: Did you mean "have you deleted it?"

sql/dd/types/init_mode.h
```diff
@@ -0,0 +1,37 @@
+/*
+   Copyright (c) 2000, 2018, Oracle and/or its affiliates. All rights reserved.
+   Copyright (c) 2019 Percona LLC and/or its affiliates. All rights reserved.
```
satya-bodapati, line 3: Why Percona copyright? This is just Oracle code in different file

storage/innobase/include/dict0dd.h
```diff
@@ -1222,5 +1222,29 @@ bool dd_tablespace_update_cache(THD *thd);
...
+ * one from space_flags.
+@param[in] Thread       THD
+@param[in] space_flags  with correct encryption flag */
+bool dd_fix_mysql_ibd_encryption_flag_if_needed(THD *thd, uint32_t space_flags);
+
+/* Sets tablespace's DD flags.
+@param[in] Thread       THD
+@param[in] space_name   name of the space for which DD flags are to be set
+@param[in] space_flags  DD flags that are to be assigned to space */
+bool dd_set_flags(THD *thd, const char *space_name, const uint32_t space_flags);
```
satya-bodapati: Shouldn't this be static function?

storage/innobase/include/dict0dd.h
```diff
@@ -1222,5 +1222,29 @@ bool dd_tablespace_update_cache(THD *thd);
...
+@param[in] Thread       THD
+@param[in] space_name   name of the space for which DD encryption flag is to be
+set */
+bool dd_set_encryption_flag(THD *thd, const char *space_name);
+
+/* Clears tablespace's DD encryption flag.
+@param[in] Thread       THD
+@param[in] space_name   name of the space for which DD encryption flag is to be
+cleared */
+bool dd_clear_encryption_flag(THD *thd, const char *space_name);
```
satya-bodapati, line 1235: Remove this function. Dead code

storage/innobase/dict/dict0dd.cc
```diff
@@ -6082,4 +6083,111 @@ bool dd_is_table_in_encrypted_tablespace(const dict_table_t *table) {
...
+                                  static_cast<uint32>(dd_space_flags));
+
+  if (dd::commit_or_rollback_tablespace_change(thd, dd_space, false)) {
+    return (true);
+  }
+
+  return (false);
+}
+
+bool dd_set_encryption_flag(THD *thd, const char *space_name) {
```
satya-bodapati, line 6133: Remove this. Dead code

storage/innobase/dict/dict0dd.cc
```diff
@@ -6082,4 +6083,111 @@ bool dd_is_table_in_encrypted_tablespace(const dict_table_t *table) {
...
+                  const uint32_t space_flags) {
+  auto set_flags = [](uint32_t &dd_space_flags, uint32_t space_flags) {
+    // currently we are using this function only for correcting encryption flag
+    ut_ad(dd_space_flags == space_flags ||
+          FSP_FLAGS_GET_ENCRYPTION(dd_space_flags) !=
+              FSP_FLAGS_GET_ENCRYPTION(space_flags));
+    dd_space_flags = space_flags;
+  };
+  auto update_func = std::bind(set_flags, std::placeholders::_1, space_flags);
+  return dd_update_tablespace_dd_flags(thd, space_name, update_func);
```
satya-bodapati, line 6172: With the clear and set functions unused, only dd_set_flags is remaining and there is no real reason to use std::function.

storage/innobase/dict/dict0dict.cc
```diff
@@ -7450,27 +7450,93 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
+                                         space_id_t mysql_plugin_space,
+                                         const char *filepath,
+                                         bool &encrypt_mysql) {
+  bool success = false;
+  switch (dict_init_mode) {
+    case DICT_INIT_CREATE_FILES:
+      encrypt_mysql = dict_is_encrypt_tables_either_on_or_force();
```
satya-bodapati: I see that you really like long names (function name that can go upto 60-80 characters).

Currently this handles only  ON or FORCE. So you made a choice to a "self describing" function which mentions value as part of function name. So if this function extends to handle below values,
SRV_ENCRYPT_TABLES_KEYRING_ON:
SRV_ENCRYPT_TABLES_KEYRING_FORCE:
SRV_ENCRYPT_TABLES_ONLINE_TO_KEYRING:
SRV_ENCRYPT_TABLES_ONLINE_TO_KEYRING_FORCE:

the function name should be:
dict_is_encrypt_tables_either_on_or_force_or_keyring_on_or_keyring_force_or_online_keyring_to_keyring_or_online_to_keyring_force() ? isn't it ;)

IMHO, function name should be precise and clear. This doesn't mean "Long explaining function names".

storage/innobase/dict/dict0dd.cc
```diff
@@ -6082,4 +6083,111 @@ bool dd_is_table_in_encrypted_tablespace(const dict_table_t *table) {
...
+}
+
+bool dd_fix_mysql_ibd_encryption_flag_if_needed(THD *thd,
```
satya-bodapati, line 6175: General suggestion "if_needed" can be avoided most times.

storage/innobase/dict/dict0dict.cc
```diff
@@ -7450,27 +7450,93 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
-    fil_space_t *space = fil_space_get(space_id);
-    ut_ad(space != nullptr);
+/** Detect if mysql.ibd's page0 has encryption flag set.
+ *  The page 0 is read from disk.
+@param[in] filepath path of mysql.ibd
+@param[out] success true if no error
+@return true if encryption flag is set, false otherwise */
+static bool dict_mysql_ibd_page_0_has_encryption_flag_set(const char *filepath,
+                                                          bool &success) {
+  pfs_os_file_t file = os_file_create_simple_no_error_handling(
```
satya-bodapati, line 7514: You should avoid direct reads if possible. (may be buf_page_get_gen()?)

Is the double write buffer recovery over at this stage? Does the double write buffer do "sync" io on the tablespace? -> If the answer is yes to both of them, then you may use direct read.

storage/innobase/include/ut0new.h
```diff
@@ -1211,4 +1211,36 @@ class aligned_array_pointer : public aligned_memory<T_Type, T_Align_to> {
   size_t m_size;
 };
 
+auto ut_unique_ptr_deleter = [](byte *buf) { ut_free(buf); };
+
+using ut_unique_ptr = std::unique_ptr<byte, decltype(ut_unique_ptr_deleter)>;
+
+inline ut_unique_ptr ut_make_unique_ptr_nokey(const size_t size) {
```
satya-bodapati, line 1223: Only this variant is required for now. Rest is dead code.

satya-bodapati (changes requested):

storage/innobase/dict/dict0dict.cc
satya-bodapati: Ok, I disagree too.

Function names should be precise. It doesn't need to describe the entire functionality in "name". That is why we have function comments. 

Fixing function comments when function does extra is not "burden". Actually it is more burden when you function names like this. Let say function does more. So we have to fix function name  fomr "it_does_this()" to "it_does_this_and_more()".

So you have to fix all the callers as well.

See my comment in the place where you use "is_on_or_force()"

storage/innobase/srv/srv0start.cc
satya-bodapati: Ok, is it possible to add assertion that flags have encrypted bit set?

storage/innobase/dict/dict0dict.cc
robgolebiowski: I think it was agreed that we are not code reviewing individual codying styles. I opt for longer names describing what a function is doing. You can disagree, but you cannot hold PR because of that.

storage/innobase/dict/dict0dict.cc
robgolebiowski: Doxygen comments were added to header file in this case. I do not think comments repetition makes sense - just add maintenance cost.

storage/innobase/srv/srv0start.cc
robgolebiowski: Why encrypted bit set ? it might be that the bit is not set. It's checking if mysql table was changed by redo applier and then it sets encryption flag in DD to the one from fil_space_t.

sql/dd/types/init_mode.h
robgolebiowski: But the file is ours, so there was a change. Laurynas asked me to put Oracle copyrights here, next to Percona one.

storage/innobase/dict/dict0dd.cc
robgolebiowski: clear and set functions are used by the encryption threads. I have taken the DD flag update code from there. Turns out that your comment about fixing DD encryption flag for mysql.ibd is only valid for encryption threads. I do not want to put effort in making code changes that will need to be reverted once encryption threads code is merged in. There I will also test the mysql.ibd DD code - so this PR depends on encryption threads anyways.

storage/innobase/dict/dict0dict.cc
robgolebiowski: I disagree, but this is code style and I will stay with longer names.

storage/innobase/dict/dict0dd.cc
robgolebiowski: Well it gives you information that mysql_ibd will not always need flag fixing - without studying the body of dd_fix_mysql_ibd_encryption_flag

sql/dd/types/init_mode.h
satya-bodapati: @laurynas-biveinis Are you sure with Percona copyright here? There is nothing new by percona here.

storage/innobase/dict/dict0dd.cc
satya-bodapati: ok, similar discussion in other thread. Many function has "if" statements. So this means the decision "if needed" can be written to all.  So functions
"dict_set_flags_if_needed()",  "dict_update_information_if_val_is_true()"
I know  this is 'your' preferred way. Had to say anyway.

storage/innobase/dict/dict0dict.cc
satya-bodapati: If it is in header, it is fine.

storage/innobase/dict/dict0dict.cc
satya-bodapati: Ok. Enjoy using them.

Other than that, why are you handling only ON and FORCE? what about the other enum values?

storage/innobase/srv/srv0start.cc
satya-bodapati: Ok, I see

storage/innobase/dict/dict0dict.cc
satya-bodapati: Anyways, you can keep this as you wish.

Am not holding this PR for this reason. There are other comments, concerns raised. If those are solved, it can proceed further.

storage/innobase/dict/dict0dd.cc
satya-bodapati: Ok, Do you have testcase for exercising all this code (by encryption threads)?
Should be easy to write one (not as part of this PR )
disable flush, encryption thread encrypt mysql.ibd, crash and ensure startup works fine.

sql/dd/types/init_mode.h
laurynas-biveinis: Our new files should carry our copyright. Let's keep both copyrights and let someone better versed in these matters to change it later if needed

satya-bodapati: Ok

storage/innobase/dict/dict0dd.cc
robgolebiowski: yes, this will be added in encryption threads test suite.

storage/innobase/dict/dict0dict.cc
robgolebiowski: I have implemented reading from the buffer, as you said, but it is not possible to do so at this stage - because the fil_system is not yet ready. When I tried to acquire the tablespace I got missing tablespace error and when I tried to read page from a buffer without acquiring space I got a core dump - as assert that page's space must exists in the fil_system was false.

storage/innobase/include/ut0new.h
robgolebiowski: @laurynas-biveinis wanted to see those functions for future use

storage/innobase/include/dict0dd.h
robgolebiowski: is used in keyring encryption. Will be used when it gets merged in.

storage/innobase/include/dict0dd.h
robgolebiowski: is used in keyring encryption. Will be used when it gets merged in.

storage/innobase/dict/dict0dd.cc
satya-bodapati: Does it exist in your encryption branch?

robgolebiowski: one of the tests (more) was failing due to this. I have restarted the server after all tables were encrypted with encryption threads and I got a crash as the DD was not expecting mysql.ibd to be encrypted.

storage/innobase/dict/dict0dict.cc
satya-bodapati: I have done lot of debugging and code study and I now am in better position :)

In my debugging, I found that recovery (recv_recovery_from_checkpoint_finish()) is over before dict_detect_encryption_of_mysql_ibd().

So there shouldn't be any issues.

1. Why is fil_system not ready. If redo recovery is over, then fil_system is fully initialized.
2. Why is mysql.ibd tablespace missing in fil_system cache?
3. buf_page_get_gen() should work fine at this stage. Infact recovery also uses buffer pool read to apply redo on pages. See buf_read_recv_pages().

And to answer the questions I asked.
"Is the double write buffer recovery over at this stage? 
Yes, the dblwr buffer recovery is over before even redo starts applying/scanning.

Does the double write buffer do "sync" io on the tablespace?
Yes, The IO is sync IO and also fsync() is done.

Neverthless, based on my debugging and observations, it is not required to do direct IO. A simple buffer pool read should work.

satya-bodapati: Ok, after our discussion our on slack and then verifying with the branch, there is *indeed* a case where we cannot do buffer pool read. This is the non-recovery case. But there is still one issue.

If recovery fixed mysq.ibd, all the changes to page 0 of mysql are done via buffer pool and regular IO.

If you do direct IO in this case, you may read a stale copy of page. The updated page 0 copy is in buffer pool (and IO threads may write at leisure).

storage/innobase/srv/srv0start.cc
```diff
@@ -2689,6 +2691,20 @@ dberr_t srv_start(bool create_new_db, const std::string &scan_directories) {
...
       log_sys_close();
 
+      // We have set mysql's dd flags based on what we read from page 0. This
+      // was before redo log was applied. Now we have applied changes from redo
+      // log. If there were changes in redo for page 0 of mysql.ibd we check if
+      // encryption flag has changed. If it did we amend mysql's dd encryption
+      // flags accordingly.
+      if (mysql_ibd_page_0_present_in_redo) {
+        fil_space_t *mysql_space = fil_space_get(dict_sys_t::s_space_id);
+        if (mysql_space == nullptr ||
```
satya-bodapati, line 2704: if the tablespace is opened by redo, it should not be nullptr.
Can you add ut_ad(mysql_space != nullptr)?

storage/innobase/dict/dict0dict.cc
robgolebiowski: @satya-bodapati I will then implement first reading from buffer pool and then if the page does not exist I will read with regular IO

satya-bodapati: Yes, May be you should use  is_mysql_ibd_page_0_in_redo() (save state). If this is true, then fil_space_t* should be present and you should do buffer pool read, else { do direct IO read)

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/72/

storage/innobase/dict/dict0dd.cc
```diff
@@ -6082,4 +6083,115 @@ bool dd_is_table_in_encrypted_tablespace(const dict_table_t *table) {
...
+  }
+  if (FSP_FLAGS_GET_ENCRYPTION(dd_space_flags) ==
+      FSP_FLAGS_GET_ENCRYPTION(space_flags)) {
+    return false;
+  }
+  // exclude encryption flag from validation
+  dd_space_flags &= ~FSP_FLAGS_MASK_ENCRYPTION;
+  space_flags &= ~FSP_FLAGS_MASK_ENCRYPTION;
+  if (dd_space_flags != space_flags) {
+    // this should not happen - some other flags other than encryption flag are
```
satya-bodapati, line 6190: Can you please add ib::error() here with the flags values? Also mention this is with encryption flag excluded.

storage/innobase/dict/dict0dict.cc
```diff
@@ -7450,27 +7450,130 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
+  IORequest request(IORequest::READ);
+  dberr_t err = os_file_read_first_page(request, file, page, UNIV_PAGE_SIZE);
+
+  os_file_close(file);
+
+  return DB_SUCCESS == err ? page : nullptr;
+}
+
+/** Detect if mysql.ibd's page0 has encryption flag set.
+ *  The page 0 either read from buffer (if available) or
```
satya-bodapati, line 7535: Can you please format into doxygen style used by other functions? Only this one seems a bit different

storage/innobase/dict/dict0dict.cc
```diff
@@ -7450,27 +7450,130 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
+    directly from disc.
```
satya-bodapati: s/disc/ disk

satya-bodapati (approved):

storage/innobase/dict/dict0dict.cc
```diff
@@ -7450,27 +7450,130 @@ dberr_t dict_get_dictionary_info_by_id(ulint dict_id, char **name,
...
+
+  const page_size_t page_size(space->flags);
+  mtr_t mtr;
+  mtr_start(&mtr);
+  buf_block_t *block = buf_page_get(page_id_t(dict_sys_t::s_space_id, 0),
+                                    univ_page_size, RW_X_LATCH, &mtr);
+  mtr_commit(&mtr);
+
+  fil_space_release(space);
+
```
satya-bodapati, line 7504: block should never be nullptr. another debug assert here please

robgolebiowski: actually with PS implementation it can be nullptr. If srv_pass_corrupt_tables is ON or there is a decryption error. For MySQL it would just crash.

satya-bodapati: Ok, does it mean all buf_page_get_gen() in PS checks for nullptr?

robgolebiowski: I cannot say for sure all of them, but we should check for null when we add new code. In encryption threads branch I have added quite a lot of them - to see if decryption passed.

robgolebiowski: sorry, when decryption fails I am returning error though an [out] variable. nullptr is only for srv_pass_corrupt_tables and it should be check with macro that will assert if the srv_pass_corrupt_tables != ON. Will update the patch with this assert.

robgolebiowski: OK, I have confused myself for a moment - both for decryption errors and for srv_pass_corrupt_tables the page can be nullptr.

PR #3017, opened 2019-02-25
PS-5263: Follow up fix for binlog_cache_write_failure testcase failure

satya-bodapati:
Problem:
--------
PS-5263 added extra my_ok() in MYSQL_BIN_LOG::handle_binlog_flush_or_sync_error
to handle binlog_error_action = ABORT_SERVER for autocommit SQLs (like INSERT, DELETE, UPDATE etc)

Because of this my_ok(), SQLs like COMMIT, ROLLBACK, etc failed with assert failure in my_ok().
These SQLs do extra my_ok() after executing the commands.

Double my_ok() is not allowed. the DBUG_ASSERT(!is_set()) fails for second my_ok().

Fix:
----
Avoid the extra my_ok() in MYSQL_BIN_LOG::handle_binlog_flush_or_sync_error() for SQLs
COMMIT, ROLLBACK, ROLLBACK TO SAVEPOINT

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -8775,7 +8775,14 @@ void MYSQL_BIN_LOG::handle_binlog_flush_or_sync_error(THD *thd,
     */
     if (check_write_error(thd)) { /* we have DA_ERROR */
       thd->clear_error(); /* sets thd->get_stmt_da()->status() to DA_EMPTY */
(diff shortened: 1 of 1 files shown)


satya-bodapati:
Approved by @laurynas-biveinis on slack

PR #3022, opened 2019-02-28
PS-5450 - InnoDB's MVCC has O(N^2) behaviors (Also fixes PS-4712 - As…

satya-bodapati:
…sertion after bug#84958 fix on FB tree)

MySQL Bug#84958 -  InnoDB's MVCC has O(N^2) behaviors
Problem:

If there're multiple row versions in InnoDB, reading one row
from PK may have O(N) complexity and reading from secondary keys may have
O(N^2) complexity.

Fix:

The problem occurs when there are many pending versions of the same row,
meaning that the primary key is the same, but a secondary key is different.
The slowdown occurs when the secondary index is traversed. This patch
creates a helper class for the function row_sel_get_clust_rec_for_mysql()
which can remember and re-use cached_clust_rec & cached_old_vers so that
rec_get_offsets() does not need to be called over and over for the
clustered record.

PS-4712 : Assertion after bug#84958 fix on FB tree
(Although this bug was found on FB tree, the fix is applicable for Percona
Server too when porting)

Problem:

The function row_sel_build_prev_vers_for_mysql() goes through the old versions
of the clustered index record and identifies the one which can be viewed by the
current transaction (MVCC).  The result is then cached in an object of
Row_sel_get_clust_rec_for_mysql.

The two things that are cached by the helper class
Row_sel_get_clust_rec_for_mysql are cached_clust_rec and cached_old_vers.  If
the latest record is cached_clust_rec, its old version for the given
transaction that is viewable is cached_old_vers. The important point here is
that the record offsets for these two records are not cached.

The record offsets for cached_clust_rec and the cached_old_vers need not be the
same.  This is confirmed by looking at row_sel_build_prev_vers_for_mysql().
The 6th argument which contains the record offsets is an input/output argument.
The input is the offsets of the latest version of clust_rec, and the output is
the record offsets of the older version.  So the record offsets of different
versions of clust_rec can change.

Solution:

Re-calculate the record offsets for the cached old_vers.

### DIFF
--- a/mysql-test/suite/innodb/r/percona_bug5450.result
+++ b/mysql-test/suite/innodb/r/percona_bug5450.result
@@ -1,0 +1,51 @@
+create procedure insert_n(n int)
+begin
+declare i int default 0;
+while i < n do
+insert into t1 values (1, 2, 3) on duplicate key update c= c + 1;
+set i = i + 1;
+end while;
+end~~
+create function bprrs()
+returns int
+begin
+declare ret int;
+select variable_value
+from information_schema.global_status
+where
+variable_name = 'innodb_buffer_pool_read_requests'
+    into ret;
+return ret;
+end~~
+create table t1 (a int, b int, c int, primary key(a,b), key (b,c)) engine=InnoDB;
+begin;
+select * from t1;
+a	b	c
+call insert_n(100);
+set @rr1= bprrs();
+select * from t1 force index (b);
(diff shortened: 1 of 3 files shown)


laurynas-biveinis (approved):

satya-bodapati:
https://ps.cd.percona.com/job/percona-server-5.7-pipeline/2660/

PR #3024, opened 2019-03-01
Ps 8.0 5450

satya-bodapati:
Merge PS-5.7-5450 to PS-8.0

Merge only the testcase as the fix is already present in 8.0 (by upstream)


laurynas-biveinis (approved):

satya-bodapati:
https://ps.cd.percona.com/job/percona-server-8.0-param/243/

PR #3039, opened 2019-03-04
PS-4654: installation - add note

borys-belinsky:


EvgeniyPatlan (commented):
Sometimes if my.cnf exists on the system it works with it but prints message during package installation with steps which should be done to update cnf. Maybe it also should be mentioned?

borys-belinsky:
Ready for review

doc/source/installation.rst
```diff
@@ -32,6 +32,25 @@ Following guides describe the installation process for using the official Percon
...
    installation/yum_repo 
 
+.. important::
+
+   When you install |Percona Server| from repositories, the installer creates a
+   backup of the :file:`/etc/mysql.cnf` file. Then, :file:`/etc/mysql.cnf`
+   becomes a soft link that ultimately (via :file:`/etc/alternatives/mysql.cnf`)
+   points to :file:`/etc/percona-server.cnf` on `RPM` based systems and to
+   :file:`/etc/mysql/percona-server.cnf` on `DEB` based systems.
+
```
hrvojem, line 42: AFAIK this is different on DEB and RPM based repos, for example there is no `/etc/mysql.cnf` on debian based systems:

`root@stretch:~# ls -lah /etc/mysql.cnf
ls: cannot access '/etc/mysql.cnf': No such file or directory`

EvgeniyPatlan: yes for rpms and debs the message is different and paths are different

PR #3040, opened 2019-03-05
PS-5446: FLUSH CHANGED_PAGE_BITMAPS leaves gaps between the last written

gl-sergei:
bitmap LSN and InnoDB checkpoint LSN

Analysis.

xtrabackup determines the checkpoint lsn at the beginning of the backup
and starts the redo log copying. Then it invokes FLUSH
CHANGED_PAGE_BITMAPS.

FLUSH CHANGED_PAGE_BITMAPS implementation simply runs
log_online_follow_redo_log which may finish scanning before reaching
checkpoint lsn.

xtrabackup will than see that tracked lsn is less than checkpoint lsn
and changed page bitmaps are unusable.

Fix.

Fix is to save checkpoint lsn at the beginning of the flush
command. This lsn will be greater or equal to the one xtrabackup
has. Then run log_online_follow_redo_log until tracked lsn is greater or
equal to saved checkpoint lsn.

### DIFF
--- a/storage/innobase/include/log0online.h
+++ b/storage/innobase/include/log0online.h
@@ -52,6 +52,14 @@ void log_online_shutdown(void) noexcept;
 changed page bitmap which is then written to disk.
 
 @return true if log tracking succeeded, false if bitmap write I/O error */
+MY_NODISCARD
+bool log_online_follow_redo_log_one_pass(void);
+
+/** Read and parse redo log for thr FLUSH CHANGED_PAGE_BITMAPS command.
+Make sure that the checkpoint LSN measured at the beginning of the command
+is tracked.
+
+@return true if log tracking succeeded, false if bitmap write I/O error */
 bool log_online_follow_redo_log(void);
 
 /** Delete all the bitmap files for data less than the specified LSN.
--- a/storage/innobase/log/log0online.cc
+++ b/storage/innobase/log/log0online.cc
@@ -761,7 +761,7 @@ static void log_online_track_missing_on_startup(
     ib::info() << "Reading the log to advance the last tracked LSN.";
 
     log_bmp_sys->start_at(std::max(last_tracked_lsn, MIN_TRACKED_LSN));
-    if (!log_online_follow_redo_log()) {
+    if (!log_online_follow_redo_log_one_pass()) {
       exit(1);
     }
     ut_ad(log_bmp_sys->end_lsn == tracking_start_lsn);
@@ -1317,12 +1317,10 @@ static void log_online_parse_complete_recs_past_previous_checkpoint() noexcept {
 page bitmap which is then written to disk.
 
 @return true if log tracking succeeded, false if bitmap write I/O error */
-bool log_online_follow_redo_log(void) {
+bool log_online_follow_redo_log_one_pass(void) {
   ut_ad(!srv_read_only_mode);
 
   if (!srv_track_changed_pages) return true;
-
-  DEBUG_SYNC_C("log_online_follow_redo_log");
 
   mutex_enter(&log_bmp_sys_mutex);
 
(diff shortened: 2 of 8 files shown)


gl-sergei:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/250/

storage/innobase/log/log0online.cc
```diff
@@ -1350,13 +1350,34 @@ bool log_online_follow_redo_log(void) {
...
+  ut_ad(!srv_read_only_mode);
+
+  const auto last_checkpoint_lsn = log_get_checkpoint_lsn(*log_sys);
+  bool result = true;
+
+  DEBUG_SYNC_C("log_online_follow_redo_log_past_checkpoint_lsn");
+
+  while (result && srv_track_changed_pages &&
+         last_checkpoint_lsn > log_sys->tracked_lsn.load()) {
+    result = log_online_follow_redo_log();
```
laurynas-biveinis: if (!result) break?

storage/innobase/include/log0online.h
```diff
@@ -54,6 +54,13 @@ changed page bitmap which is then written to disk.
 @return true if log tracking succeeded, false if bitmap write I/O error */
 bool log_online_follow_redo_log(void);
 
+/** Read and parse redo log past the last checkpoint lsn for FLUSH
+CHANGED_PAGE_BITMAPS command
+
+@return true if log tracking succeeded, false if bitmap write I/O error
+*/
+bool log_online_follow_redo_log_past_checkpoint_lsn(void);
```
laurynas-biveinis: This is the main entry point into the changed page tracking - IMHO it should have a short name, such as original log_online_follow_redo_log, and that one should be renamed instead (i.e. log_online_follow_redo_log_one_pass or something else.)

storage/innobase/include/log0online.h
```diff
@@ -54,6 +54,13 @@ changed page bitmap which is then written to disk.
 @return true if log tracking succeeded, false if bitmap write I/O error */
 bool log_online_follow_redo_log(void);
```
laurynas-biveinis: Is it now feasible to mark this one MY_NODISCARD?

laurynas-biveinis (changes requested):

storage/innobase/log/log0online.cc
gl-sergei: `if (!result) break'` is not needed, because the loop is `while (result)`

storage/innobase/include/log0online.h
gl-sergei: makes sense, will rename

storage/innobase/log/log0online.cc
```diff
@@ -1317,13 +1317,11 @@ static void log_online_parse_complete_recs_past_previous_checkpoint() noexcept {
 page bitmap which is then written to disk.
 
 @return true if log tracking succeeded, false if bitmap write I/O error */
-bool log_online_follow_redo_log(void) {
+bool log_online_follow_redo_log_one_pass(void) {
```
laurynas-biveinis, line 1320: Also, can this one be made static?

storage/innobase/log/log0online.cc
laurynas-biveinis: Right, sorry

storage/innobase/log/log0online.cc
gl-sergei: `srv_redo_log_follow_thread` is using `...one_pass` and I did not want change it. My original idea was to only change FLUSH command....

laurynas-biveinis: Sure, no need to change that

gl-sergei:
Updated:
- `log_online_follow_redo_log renamed` to `log_online_follow_redo_log_one_pass`, added MY_NODISCARD

https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/251/

percona_changed_page_bmp_log_resize failed again. Note, previous failure was in different PR.

laurynas-biveinis (approved):

PR #3041, opened 2019-03-05
PS-5447: Changed page tracking is missing pages changed by the in-place

gl-sergei:
DDL

In-place DDL does not redo log changes. The only redo log record it logs
is the MLOG_INDEX_LOAD with space_id and the root page of the changed
index. Fix is to mark all pages of affected tablespace as changed.


gl-sergei:
https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/124/

laurynas-biveinis (approved):

PR #3053, opened 2019-03-06
PS-4606: PXB with keyring vault plugin - remove incorrect note

borys-belinsky:


patrickbirch:
Close this PR - information is updated in 6075

PR #3057, opened 2019-03-08
5.7 ps 3969

altmannmarcelo:
This commit adds a variable to track when an audit log entry was either
dropped or written directly to the file due to its size being bigger
than audit_log_buffer_size

https://ps.cd.percona.com/job/percona-server-5.7-param/128/


plugin/audit_log/audit_log.c
```diff
@@ -70,6 +70,7 @@ static char *audit_log_exclude_databases= NULL;
 static char *audit_log_include_databases= NULL;
 static char *audit_log_exclude_commands= NULL;
 static char *audit_log_include_commands= NULL;
+int64 audit_log_buffer_size_overflow = 0;
```
altmannmarcelo, line 73: @laurynas-biveinis  previous build was failing because `my_atomic_add64` expects the first parameter to be an int64 (not uint64 like it was before)

altmannmarcelo:
Ubuntu cosmic builds were failing(related to myrocks). Compiled again without it
https://ps.cd.percona.com/job/percona-server-5.7-pipeline/2942/
https://ps.cd.percona.com/job/percona-server-5.7-pipeline/2941/

laurynas-biveinis (approved):

PR #3067, opened 2019-03-13
PS-5473 : innodb_system tablespace information missing from I_S innodb_tablespaces view

satya-bodapati:
Problem:
--------
innodb system tablespace entry missing from
SELECT * FROM INFORMATION_SCHEMA.INNODB_TABLESPACES;

1)
With no such entry, we miss information like whether the system tablespace is
encrypted or not.

2)
The row format shows "Compact or Redundant" for innodb system
tablespace. This is wrong. Dynamic is allowed too. So it should show
"any".

3)
FS_BLOCK_SIZE, FILE_SIZE, ALLOCATED_SIZE are empty for undo tablespaces.
Undo tablespace is also single-file tablespace and above stats can be
retrieved.

Fix:
----
Make sure system tablespace entry exists in I_S.INNODB_TABLESPACES output
Also fix 2) and 3).

### DIFF
--- a/storage/innobase/handler/i_s.cc
+++ b/storage/innobase/handler/i_s.cc
@@ -6525,6 +6525,42 @@ static ST_FIELD_INFO innodb_tablespaces_fields_info[] = {
 
 };
 
+/** Get the filepath of tablespace. For multi-file tablespaces (like
+system and temporary), return nullptr. Callers owns the responsibility
+for freeing the memory of the path returned and should use ut_free()
+to free the memory
+@param[in]	space_id	tablespace id
+@param[in]	flags		tablespace flags
+@param[in]	name		tablespace name
+@return filepath or nullptr */
+MY_NODISCARD
+static char *i_s_get_filepath(space_id_t space_id, ulint flags,
+                              const char *name) {
+  char *filepath = nullptr;
+
+  if (fsp_is_system_or_temp_tablespace(space_id)) {
+    return (filepath);
+  }
+
+  if (FSP_FLAGS_HAS_DATA_DIR(flags) || FSP_FLAGS_GET_SHARED(flags) ||
+      fsp_is_undo_tablespace(space_id)) {
+    mutex_enter(&dict_sys->mutex);
+    filepath = fil_space_get_first_path(space_id);
+    mutex_exit(&dict_sys->mutex);
+  }
+
+  if (filepath == nullptr) {
+    if (strstr(name, dict_sys_t::s_file_per_table_name) != 0) {
+      mutex_enter(&dict_sys->mutex);
+      filepath = fil_space_get_first_path(space_id);
+      mutex_exit(&dict_sys->mutex);
+    } else {
+      filepath = Fil_path::make_ibd_from_table_name(name);
+    }
+  }
+  return (filepath);
+}
+
 /** Function to fill INFORMATION_SCHEMA.INNODB_TABLESPACES with information
 collected by scanning INNODB_TABLESPACESS table.
 @param[in]      thd             thread
@@ -6560,9 +6596,8 @@ static int i_s_dict_fill_innodb_tablespaces(
 
   if (fsp_is_undo_tablespace(space_id)) {
     row_format = "Undo";
-  } else if (fsp_is_system_or_temp_tablespace(space_id)) {
-    row_format = "Compact or Redundant";
-  } else if (fsp_is_shared_tablespace(flags) && !is_compressed) {
+  } else if (fsp_is_system_or_temp_tablespace(space_id) ||
+             (fsp_is_shared_tablespace(flags) && !is_compressed)) {
     row_format = "Any";
   } else if (is_compressed) {
     row_format = "Compressed";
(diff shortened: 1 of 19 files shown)


satya-bodapati:
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/293/

storage/innobase/handler/i_s.cc
```diff
@@ -6525,6 +6525,39 @@ static ST_FIELD_INFO innodb_tablespaces_fields_info[] = {
 
 };
 
+/** Get the filepath of tablespace. For multi-file tablespaces (like
+system and temporary), return nullptr. Callers owns the responsibility
+for freeing the memory of the path returned
```
laurynas-biveinis: Please also specify which API to use

storage/innobase/handler/i_s.cc
```diff
@@ -6525,6 +6525,39 @@ static ST_FIELD_INFO innodb_tablespaces_fields_info[] = {
...
+@param[in]	space_id	tablespace id
+@param[in]	flags		tablespace flags
+@param[in]	name		tablespace name
+@return filepath or nullptr */
+static char *i_s_get_filepath(space_id_t space_id, ulint flags,
```
laurynas-biveinis, line 6537: MY_NODISCARD

storage/innobase/handler/i_s.cc
```diff
@@ -6607,22 +6640,7 @@ static int i_s_dict_fill_innodb_tablespaces(
...
-  if (filepath == NULL) {
-    if (strstr(name, dict_sys_t::s_file_per_table_name) != 0) {
-      mutex_enter(&dict_sys->mutex);
-      filepath = fil_space_get_first_path(space_id);
-      mutex_exit(&dict_sys->mutex);
-    } else {
-      filepath = Fil_path::make_ibd_from_table_name(name);
-    }
-  }
+  char *filepath = i_s_get_filepath(space_id, flags, name);
```
laurynas-biveinis, line 6645:
```suggestion
  const char *filepath = i_s_get_filepath(space_id, flags, name);
```

storage/innobase/handler/i_s.cc
```diff
@@ -6721,7 +6739,7 @@ static int i_s_innodb_tablespaces_fill_table(THD *thd, TABLE_LIST *tables,
     mtr_commit(&mtr);
     mutex_exit(&dict_sys->mutex);
 
-    if (ret && space != 0) {
+    if (ret) {
```
laurynas-biveinis, line 6744: Strictly speaking this is the fix, and i_s_get_filepath is refactoring only?

laurynas-biveinis (changes requested):
Encryption aside, this seems to be a specific case of an upstream bug: that space id = 0 is not present in innodb_tablespaces, it should be reported there

storage/innobase/handler/i_s.cc
satya-bodapati: My initial fix was only this :) and then it was crashing because of assert in fil_space_get_first_path().

  ut_a(space_id != TRX_SYS_SPACE);

The assert is valid. It says, you shouln't use this API for multi-file tablespaces. It is wrong to consider only first file when calculating size, block_size etc.

So the above refactoring avoids calls to fil_space_get_first_path() for system tablespace.

laurynas-biveinis: Ah yes, now I see it

satya-bodapati: Even better fix (future proof) would be to iterate all files of system tablespace and get the total size. We have to assume all files of system tablespace reside in one partition(or different partitions but of same block size).

But currently it is not seen necessary as Page IO compression is not enabled on system tablespace.

satya-bodapati:
ready for review

storage/innobase/handler/i_s.cc
```diff
@@ -6652,7 +6672,7 @@ static int i_s_dict_fill_innodb_tablespaces(
         break;
     }
 
-    ut_free(filepath);
+    ut_free(const_cast<char *>(filepath));
```
laurynas-biveinis: Sorry, it was a mistake to suggest const then :(

storage/innobase/handler/i_s.cc
```diff
@@ -6607,22 +6642,7 @@ static int i_s_dict_fill_innodb_tablespaces(
...
-  if (filepath == NULL) {
-    if (strstr(name, dict_sys_t::s_file_per_table_name) != 0) {
-      mutex_enter(&dict_sys->mutex);
-      filepath = fil_space_get_first_path(space_id);
-      mutex_exit(&dict_sys->mutex);
-    } else {
-      filepath = Fil_path::make_ibd_from_table_name(name);
-    }
-  }
+  const char *filepath = i_s_get_filepath(space_id, flags, name);
```
laurynas-biveinis:
```suggestion
  char *filepath = i_s_get_filepath(space_id, flags, name);
```

laurynas-biveinis (changes requested):

storage/innobase/handler/i_s.cc
satya-bodapati: No problem

satya-bodapati:
Addressed the const char* issue, ready for review.

laurynas-biveinis (approved):

satya-bodapati:
contributed the same patch to upstream as https://github.com/mysql/mysql-server/pull/256 Lets see how it goes

PR #3072, opened 2019-03-14
Post test fix for PS-5473

satya-bodapati:


satya-bodapati:
self approval for only result content mismatches.

PR #3076, opened 2019-03-14
PS-5489 Fix double dash

EvgeniyPatlan:


EvgeniyPatlan:
Merging as the same changes were approved for 5.7

PR #3079, opened 2019-03-14
PS-4842 no default listed corrupt action 8.0

pbirch999:


laurynas-biveinis (approved):

laurynas-biveinis (changes requested):
This fix applies to the 5.6 and 5.7 versions too, so it should follow the process of fixing lower then higher versions

PR #3080, opened 2019-03-15
PS-5492: bootstrap failure with innodb_encrypt_tables = 1

satya-bodapati:
Problem:
--------
Bootstrap fails (asserts in debug build) with innodb_encrypt_tables = 1

This happens because server requests to create temporary table before UUID
is ready. InnoDB allocates session temp tablespace and tries to set encryption
info. The assert detects that encryption is requested before UUID is ready.

Fix:
----

1. We allow to create encrypted session temp tablespaces with empty uuid for a brief
   period of time.

2. We fix the assert to rely on a new debug variable 'srv_is_uuid_ready' instead
   of 'innodb_inited'. So any tablespaces that are encrypted at startup before
   uuid is ready, should be fixed in hanlderton API innobase_fix_tablespaces_empty_uuid()

3. For session temp tablespaces, it is observed that they are destroyed before we attempt
   fix the empty uuid. Add debug checks to catch if there are any encrypted session temp
   tablespaces.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -4180,6 +4180,11 @@ a new master key and do key rotation. These tablespaces if encrypted
 during startup, will be encrypted with tablespace key which has empty UUID
 @return false on success, true on failure */
 bool innobase_fix_tablespaces_empty_uuid() {
+#ifdef UNIV_DEBUG
+  /* This API is called only after uuid is ready */
+  srv_is_uuid_ready = true;
+#endif /* UNIV_DEBUG */
+
   /* If we are in read only mode, we cannot do rotation but it
   is OK */
   if (srv_read_only_mode) {
@@ -4231,6 +4236,19 @@ bool innobase_fix_tablespaces_empty_uuid() {
 
   space_ids.push_back(srv_sys_space.space_id());
   space_ids.push_back(srv_tmp_space.space_id());
+
+#ifdef UNIV_DEBUG
+  /* Currently all session temp tablespaces that use empty uuid
+  are destroyed. So if there is encrypted sesion temp tablespace
+  we assert here */
+  const auto find_encrypted = [&](const ibt::Tablespace *ts) {
+    if (ts->is_encrypted()) {
+      ut_ad(0);
+    }
+  };
+
+  ibt::tbsp_pool->iterate_active_tbsp(find_encrypted);
+#endif /* UNIV_DEBUG */
 
   undo::spaces->s_lock();
   for (auto undo_space : undo::spaces->m_spaces) {
--- a/storage/innobase/include/srv0tmp.h
+++ b/storage/innobase/include/srv0tmp.h
@@ -135,8 +135,32 @@ class Tablespace {
   /** @return complete path including filename */
   std::string path() const;
 
-  /** @return true if session tablespace is encrypted, else false */
+  /** Encrypt a session temporary tablespace
+  @return true if session tablespace is encrypted, else false */
   bool encrypt();
+
+  /** @return true for encrypted purpose, else false */
+  bool is_encrypted() const { return (is_encrypted(m_purpose)); }
+
+  /** @return true for encrypted purpose, else false
+  @param[in]  purpose  the purpose of session temp tablespace */
+  static bool is_encrypted(enum tbsp_purpose purpose) {
+    switch (purpose) {
+      case TBSP_USER:
+      case TBSP_INTRINSIC:
+      case TBSP_SLAVE:
+        return (false);
+      case TBSP_ENC_USER:
+      case TBSP_ENC_INTRINSIC:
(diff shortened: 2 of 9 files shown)


satya-bodapati:
Ready for review with new approach.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -4232,6 +4237,17 @@ bool innobase_fix_tablespaces_empty_uuid() {
   space_ids.push_back(srv_sys_space.space_id());
   space_ids.push_back(srv_tmp_space.space_id());
 
+  /* Currently all session temp tablespaces that use empty uuid
+  are destroyed. So if there is encrypted sesion temp tablespace
+  we assert here */
+  auto find_encrypted = [&](const ibt::Tablespace *ts) {
+    if (ts->is_encrypt()) {
+      ut_ad(0);
```
laurynas-biveinis, line 4246: The debug build assert is in the innermost loop, so for release build this is a loop with no-op body. I'd surround the whole piece with #ifdef UNIV_DEBUG

storage/innobase/include/srv0tmp.h
```diff
@@ -218,6 +242,19 @@ class Tablespace_pool {
     release();
   }
 
+  /** Iterate through the list of "active" tablespaces and perform specified
+  operation on the tablespace on every iteration.
+  @param[in]    f                Function pointer for the function to be
+  executed on every iteration */
+  template <typename F>
+  void iterate_active_tbsp(F &&f) {
```
laurynas-biveinis, line 250: Hmm, that lambda is not actually moved into this, is it? Not moving it also should allow to declare it as const auto

storage/innobase/include/srv0tmp.h
```diff
@@ -135,9 +135,33 @@ class Tablespace {
   /** @return complete path including filename */
   std::string path() const;
 
-  /** @return true if session tablespace is encrypted, else false */
+  /** Encrypt a session temporary tablespace
+  @return true if session tablespace is encrypted, else false */
   bool encrypt();
 
+  /** @return true for encrypted purpose, else false */
+  bool is_encrypt() const { return (is_encrypt(m_purpose)); }
```
laurynas-biveinis:
```suggestion
  bool is_encrypted() const { return (is_encrypt(m_purpose)); }
```

storage/innobase/include/srv0tmp.h
```diff
@@ -135,9 +135,33 @@ class Tablespace {
...
+
+  /** @return true for encrypted purpose, else false
+  @param[in]  purpose  the purpose of session temp tablespace */
+  static bool is_encrypt(enum tbsp_purpose purpose) {
```
laurynas-biveinis:
```suggestion
  static bool is_encrypted(enum tbsp_purpose purpose) {
```

storage/innobase/srv/srv0tmp.cc
```diff
@@ -186,7 +167,7 @@ bool Tablespace::encrypt() {
 }
 
 void Tablespace::decrypt() {
-  if (!is_encrypt(m_purpose)) {
+  if (!is_encrypt()) {
```
laurynas-biveinis:
```suggestion
  if (!is_encrypted()) {
```

storage/innobase/srv/srv0tmp.cc
```diff
@@ -273,7 +254,7 @@ Tablespace *Tablespace_pool::get(my_thread_id id, enum tbsp_purpose purpose) {
   }
 
   ts = m_free->back();
-  if (is_encrypt(purpose)) {
+  if (Tablespace::is_encrypt(purpose)) {
```
laurynas-biveinis:
```suggestion
  if (Tablespace::is_encrypted(purpose)) {
```

laurynas-biveinis (changes requested):

satya-bodapati:
addressed review comments ready for review @laurynas-biveinis

satya-bodapati:
run all innodb suites: https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/367/

storage/innobase/include/srv0tmp.h
laurynas-biveinis: What was the resolution here?

laurynas-biveinis (approved):

storage/innobase/include/srv0tmp.h
satya-bodapati: Isn't this request to use const auto for the lamba function? 
I have used
const auto find_encrypted = [&]...

laurynas-biveinis: That was half of it yes, the other half was - is it required to pass by rvalue ref?

satya-bodapati: Hmm, probably not. Just used same as original implementation.

PR #3091, opened 2019-03-20
PS-5243 SHOW BINLOG EVENT ASSERTION

altmannmarcelo:

also fixes mysqlbinlog.

### DIFF
--- a/include/my_sys.h
+++ b/include/my_sys.h
@@ -516,7 +516,7 @@ static inline int my_b_inited(const IO_CACHE *info)
 MY_NODISCARD
 static inline int my_b_read(IO_CACHE *info, uchar *Buffer, size_t Count)
 {
-  if (info->read_pos + Count <= info->read_end)
+  if ((uint)(info->read_end - info->read_pos) >= Count)
   {
     memcpy(Buffer, info->read_pos, Count);
     info->read_pos+= Count;


altmannmarcelo:
https://ps.cd.percona.com/job/percona-server-5.7-param/129/

include/my_sys.h
```diff
@@ -516,7 +516,7 @@ static inline int my_b_inited(const IO_CACHE *info)
 MY_NODISCARD
 static inline int my_b_read(IO_CACHE *info, uchar *Buffer, size_t Count)
 {
-  if (info->read_pos + Count <= info->read_end)
+  if ((uint)(info->read_end - info->read_pos) >= Count)
```
laurynas-biveinis, line 519: Why cast to uint is required?

laurynas-biveinis (changes requested):
The change looks correct, unfortunately, there is some related yak shaving to be done: 5.6 should also be affected (please check), but running the test on a 32bit machine will not necessarily reveal it because testcase improvements of https://jira.percona.com/browse/PS-5126 were not backported to 5.6) - thus, please backport the testcase to 5.6

include/my_sys.h
altmannmarcelo: It was introducing a lot of new warnings:

```
/tmp/ps/include/my_sys.h: In function 'my_b_read':
/tmp/ps/include/my_sys.h:519:41: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if ((info->read_end - info->read_pos) >= Count)
```

However, since Count can be of any size one may pass an int to it, which will lead to the same problem. I think it makes more sense to cast Count instead.

```c++
if (info->read_end - info->read_pos >= (int)Count)
```

percona-ysorokin: The result type of subtracting 2 pointers is `ptrdiff_t` which is a signed integral type of the same size as `size_t` (https://en.cppreference.com/w/cpp/types/ptrdiff_t). Moreover, we know that `info->read_end` > `info->read_pos` by definition, so in my opinion the safest and the most portable solution which would cover both x86 and x64 platforms would be
```c++
  if ((size_t)(info->read_end - info->read_pos) >= Count)
```

PR #3102, opened 2019-03-26
PS-5243 SHOW BINLOG EVENT ASSERTION

altmannmarcelo:
Cherry-pick the testcase[1] from 5.7 and fixed the bug itself:

my_b_write uses the same logic as my_b_read. Extending
the patch to my_b_write as well.

also fixes mysqlbinlog.

[1]:
commit ac038cc5ed04675c90f3921bcd4c72e89979cbd2
Author: Przemek Skibinski <przemyslaw.skibinski@percona.com>
Date:   Mon Dec 24 13:11:53 2018 +0100

    PS-5126: SHOW BINLOG EVENTS FROM <bad offset> is not diagnosed
    inside Format_description_log_events

    Backport improved rpl.bug75480 test from 8.0

### DIFF
--- a/include/my_sys.h
+++ b/include/my_sys.h
@@ -536,13 +536,13 @@ extern my_error_reporter my_charset_error_reporter;
 #define my_b_EOF INT_MIN
 
 #define my_b_read(info,Buffer,Count) \
-  ((info)->read_pos + (Count) <= (info)->read_end ?\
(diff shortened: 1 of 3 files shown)


altmannmarcelo:
https://ps.cd.percona.com/job/percona-server-5.6-param/61/

laurynas-biveinis (approved):
OK if GCA rev

PR #3109, opened 2019-03-28
8.0 PS-5243 SHOW BINLOG EVENT ASSERTION

altmannmarcelo:

my_b_write uses the same logic as my_b_read. Extending
the patch to my_b_write as well.

also fixes mysqlbinlog.


altmannmarcelo:
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/260/

laurynas-biveinis (approved):

PR #3111, opened 2019-03-29
PS-5516 [RHEL 8] when installing 5.7 packages, if sysbench/PT is already installed MariaDB config will be used

EvgeniyPatlan:
This commit should fix build issues on rhel8 connected with mecab_libs
    and dependency issue. Also it fix situation when mariadb-connector-c-config
    is installed and during installation of percona-server old my.cnf will be used.
    Additionally we will create backup of old my.cnf on first installation

### DIFF
--- a/build-ps/percona-server-5.7_builder.sh
+++ b/build-ps/percona-server-5.7_builder.sh
@@ -102,6 +102,13 @@ add_percona_yum_repo(){
 
 add_percona_apt_repo(){
   if [ ! -f /etc/apt/sources.list.d/percona-dev.list ]; then
+    if [ "x$RHEL" = "x8" ]; then
+      echo -e '[main]\nenabled=0\n' > /etc/yum/pluginconf.d/subscription-manager.conf
+      echo 'strict=0' >> /etc/dnf/dnf.conf
+      echo 'strict=0' >> /etc/yum/yum.conf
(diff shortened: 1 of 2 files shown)


hors (approved):

build-ps/percona-server-5.7_builder.sh
```diff
@@ -464,6 +471,11 @@ build_mecab_lib(){
     make
     make check
     make DESTDIR=${MECAB_INSTALL_DIR} install
+    cd ../${MECAB_INSTALL_DIR}
+    if [ -d usr/lib64 ]; then
+	mkdir -p usr/lib
+        mv usr/lib64/* usr/lib
```
laurynas-biveinis, line 477: Indentation

laurynas-biveinis (changes requested):
Please use descriptive:
- JIRA issue title 
- commit title
- commit message, which should list what is changed and why. It seems that this commit changes many things at once, and PS-5516 is only about part of them

build-ps/percona-server-5.7_builder.sh
EvgeniyPatlan: fixed. thanks

EvgeniyPatlan:
@laurynas-biveinis thanks for the info. I will try to keep this style now and in future.

PR #3112, opened 2019-04-01
PS-5508 : native_part::Partition_base::<...> -> handler::ha_open -> _…

satya-bodapati:
…_cxa_pure_virtual () -> ABORT

PS-5525 : TokuDB and MyRocks native partitioning never releases memory

Problem:
--------
1)

Crashes due to memory corruption. This happens because mem_root of TABLE_SHARE* of partition handler
(table_share->mem_root) is shared between multiple threads.

MEM_ROOT documentation says

"The MEM_ROOT is thread-compatible but not thread-safe. This means you cannot
use the same instance from multiple threads at the same time without external
synchronization, but you can use different MEM_ROOTs concurrently in
different threads"

Because of mem_root corruption, two handlers objects are created at same address.

2)
Native partitioning never releases memory (or releases only at shutdown). This
because partition handlers on open/create are using the TABLE_SHARE* mem_root.
TABLE_SHARE* mem_root is freed only at shutdown.

Fix:
----
Use TABLE*->mem_root for partition handler memory requirements.

TABLE* object is one per handler and not shared across threads/handlers.
Hence it is safe across threads/handlers.

Also TABLE* are destroyed based on TABLE open cache management and hence
memory is also released during runtime. (see Table_cache::free_unused_tables_if_necessary())


laurynas-biveinis (approved):
Thanks, what a nasty bug it was

laurynas-biveinis:
Hmm how come our native partitioning suites were passing before this fix, probably too much focus on single-threaded workload? Possible to create an MTR testcase with two connections operating on the same native-partitioned table crashing without the fix?

satya-bodapati:
Yes! very nasty. So much effort had to be put to find the root cause.
Thanks for the quick approval @laurynas-biveinis

satya-bodapati:
The only multi connections tests we have are doing ALTER+SELECT/UPDATE and another LOCK TABLE FOR READ/WRITE+ SELECT UPDATE. Rest all are single threaded.

that is why testsuite never caught this.(checked tokudb.parts test suite)

mysql-test/suite/parts/inc/multi_con_read.inc
```diff
@@ -0,0 +1,50 @@
...
+--source include/count_sessions.inc
+eval CREATE TABLE `t1` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `k` int(11) NOT NULL DEFAULT '0',
+  `c` char(120) NOT NULL DEFAULT '',
+  `pad` char(60) NOT NULL DEFAULT '',
+  PRIMARY KEY (`id`),
+  KEY `k_1` (`k`)
+) ENGINE=$engine AUTO_INCREMENT=10001 DEFAULT CHARSET=utf8mb4
+/*!50100 PARTITION BY RANGE (`id`)
```
laurynas-biveinis: Let's assume this is indeed a server no earlier than 5.1:
```suggestion
PARTITION BY RANGE (`id`)
```

mysql-test/suite/parts/inc/multi_con_read.inc
```diff
@@ -0,0 +1,50 @@
...
+SUBPARTITION BY HASH (`id`)
+SUBPARTITIONS 20
+(PARTITION p0 VALUES LESS THAN (333333),
+ PARTITION p1 VALUES LESS THAN (666666),
+ PARTITION p3 VALUES LESS THAN MAXVALUE) */;
```
laurynas-biveinis:
```suggestion
 PARTITION p3 VALUES LESS THAN MAXVALUE);
```

laurynas-biveinis (approved):
OK to push with minor testcase update, no further reviews needed

PR #3113, opened 2019-04-01
Ps 8.0 5508

satya-bodapati:
Merge PS-5.7-5508 to 8.0


laurynas-biveinis (commented):
Is it possible to create an MTR testcase with two connections operating on the same native-partitioned table crashing without the fix?

satya-bodapati:
@laurynas-biveinis without the fix, two connections using same native partition table, need not result in crash because it depends on how mem root allocations are interleaved. Only if two objects fall under same address, corruptions are much more visible.

Nevertheless, I am thinking if a test fails with --repeat=10, it is worth. I will give it a shot but outcome may or may not be positive ;)

PR #3115, opened 2019-04-03
PS-5476: Report error when redo log encryption is requested without a keyring (5.7)

dutow:
Issue: when the server is started without a keyring, the redo log encryption
    checks were never performed, redo log encrytpion wasn't initialized, but the
    setting was left on, misleading the user. Later operations which triggered
    this check could possibly fail.

    Issue #2: redo log encryption didn't work properly when it was turned on using
    the command line: in some cases, the redo log remained unencrypted (until
    another operation triggered the redo log encryption setup methods to be called
    again)

    Both issues were caused by the refactoring of the periodic, once every
    second encryption checks in upstream, in PS-5189.

    This commit:

    * changes the code so redo log encryption routienes are called at least twice
    during every startup with the encryption settings on. This fixes issue #2.
    * changes the code so redo log encryption routines are called at least once
    during evrey startup, even without a keyring, or a read only server: this fixes
    issues #1
    * adds a test ensuring that in an invalid configuration (without a keyring) the
    server remains running, but the redo log encryption setting correctly displays the
    OFF status
    * adds a test ensuring that when the redo log is correctly configured (using the
    dynamic variable or a command line parameter) the log is encrypted, and no unencrypted
    data is present in it.

### DIFF
--- a/mysql-test/suite/innodb/t/percona_log_encrypt_failure.test
+++ b/mysql-test/suite/innodb/t/percona_log_encrypt_failure.test
@@ -1,0 +1,11 @@
+call mtr.add_suppression("Encryption can't find master key, please check the keyring plugin is loaded.");
+call mtr.add_suppression("Can't set redo log tablespace to be encrypted.");
+select @@innodb_redo_log_encrypt;
+
+
+--let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let SEARCH_PATTERN=Can't set redo log tablespace to be encrypted.
+--source include/search_pattern.inc
+--let SEARCH_PATTERN=Encryption can't find master key, please check the keyring plugin is loaded.
+--source include/search_pattern.inc
+
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -3910,6 +3910,8 @@ static bool innobase_is_supported_system_table(const char *db,
 /* mutex protecting the master_key_id */
 ib_mutex_t	master_key_id_mutex;
 
+void srv_enable_undo_encryption_if_set();
+
 /** Fix the empty UUID of tablespaces like system, temp etc by generating
 a new master key and do key rotation. These tablespaces if encrypted
 during startup, will be encrypted with tablespace key which has empty UUID
@@ -3917,20 +3919,25 @@ during startup, will be encrypted with tablespace key which has empty UUID
 bool
 innobase_fix_tablespaces_empty_uuid()
 {
+	if (Encryption::master_key_id == 0) {
+		/* We have to call srv_enable_redo_encryption during every
+		 startup, to report errors generated in this function correctly.
+		 Without this call here, some illegal configurations, such as
+		 enabling encryption without a keyring are silently accepted,
+		 and result in errors later during the server run.
+		 These functions are also called later, when the 
+		 master key is correctly set up, later in this function. */
+		if(srv_enable_redo_encryption()) {
+			srv_redo_log_encrypt = REDO_LOG_ENCRYPT_OFF;
+		} else {
+			redo_rotate_default_key();
+		}
+	}
+
 	/* If we are in read only mode, we cannot do rotation but it
 	is OK */
 	if (srv_read_only_mode) {
 		return(false);
-	}
-
-	if (Encryption::master_key_id == 0 && srv_redo_log_encrypt == REDO_LOG_ENCRYPT_RK) {
-		/* redo log can be encrypted with keyring_key, which has to be 
-		   initialized even when nothing's encrypted with master_key - in
-		   which case, master_key_id == 0, so this condition succeeds. 
-		   We call this function here, because otherwise, if the redo log
-		   is encrypted with master_key, log encryption has to be initialized
-		   after Encryption::create_master_key. */
-		log_enable_encryption_if_set();
 	}
 
 	/* We only need to handle the case when an encrypted tablespace
@@ -3963,7 +3970,11 @@ innobase_fix_tablespaces_empty_uuid()
 		return(true);
 	}
 
-	log_enable_encryption_if_set();
+	if (srv_enable_redo_encryption()) {
+		srv_redo_log_encrypt = REDO_LOG_ENCRYPT_OFF;
+	} else {
+		redo_rotate_default_key();
+	}
 
 	/** Check if sys, temp need rotation to fix the empty uuid */
 	/* Also, in future, it is possible to fix empty uuid for redo & undo
@@ -21437,12 +21448,36 @@ innodb_temp_tablespace_encryption_update(
 @param[in]	save	immediate result from check function */
 static
 void
-innodb_redo_encryption_update(
+update_innodb_redo_log_encrypt(
 	THD*				thd,
 	struct st_mysql_sys_var*	var,
 	void*				var_ptr,
 	const void*			save)
 {
+	const ulong target = *static_cast<const ulong *>(save);
+
+	if (srv_redo_log_encrypt == target) {
+		/* No change */
+		return;
+	}
+
+	if (target == REDO_LOG_ENCRYPT_OFF) {
+  		srv_redo_log_encrypt = REDO_LOG_ENCRYPT_OFF;
+ 		return;
+	}
+
+	if (srv_redo_log_encrypt != REDO_LOG_ENCRYPT_OFF
+	    && srv_redo_log_encrypt != target) {
+		push_warning_printf(
+			thd, Sql_condition::SL_WARNING,
+			ER_WRONG_ARGUMENTS,
+			" Redo log encryption mode"
+			" can't be switched without stopping the server and"
+			" recreating the redo logs.");
+	    return;
+	}
+
+
 	if (srv_read_only_mode) {
 		push_warning_printf(
 			thd, Sql_condition::SL_WARNING,
@@ -21452,10 +21487,30 @@ innodb_redo_encryption_update(
 		return;
 	}
 
-	*static_cast<ulong*>(var_ptr) =
-		*static_cast<const ulong*>(save);
+	byte iv[ENCRYPTION_KEY_LEN];
+	Encryption::random_value(iv);
 
-	log_enable_encryption_if_set();
+	if (target == REDO_LOG_ENCRYPT_MK) {
+		ut_ad(strlen(server_uuid) > 0);
+		if(srv_enable_redo_encryption_mk()) {
+			return;
+		}
+
+		srv_redo_log_encrypt = target;
+		return;
+	}
+
+	if (target == REDO_LOG_ENCRYPT_RK) {
+		ut_ad(strlen(server_uuid) > 0);
+		if(srv_enable_redo_encryption_rk()) {
+			return;
+		}
+
+		srv_redo_log_encrypt = target;
+		return;
+	}
+
+	ut_ad(0);
 }
 
 static SHOW_VAR innodb_status_variables_export[]= {
@@ -22521,7 +22576,7 @@ static MYSQL_SYSVAR_ENUM(default_row_format, innodb_default_row_format,
 static MYSQL_SYSVAR_ENUM(redo_log_encrypt, srv_redo_log_encrypt,
   PLUGIN_VAR_OPCMDARG,
   "Enable or disable Encryption of REDO tablespace. Possible values: OFF, MASTER_KEY, KEYRING_KEY.",
-  NULL, innodb_redo_encryption_update, REDO_LOG_ENCRYPT_OFF, &redo_log_encrypt_typelib);
+  NULL, update_innodb_redo_log_encrypt, REDO_LOG_ENCRYPT_OFF, &redo_log_encrypt_typelib);
 
 #ifdef UNIV_DEBUG
 static MYSQL_SYSVAR_UINT(trx_rseg_n_slots_debug, trx_rseg_n_slots_debug,
--- a/storage/innobase/include/srv0srv.h
+++ b/storage/innobase/include/srv0srv.h
@@ -355,12 +355,6 @@ extern ulong	srv_n_log_files;
 
 extern ulong	srv_redo_log_encrypt;
 
-enum redo_log_encrypt_enum {
-	REDO_LOG_ENCRYPT_OFF = 0,
-	REDO_LOG_ENCRYPT_MK = 1,
-	REDO_LOG_ENCRYPT_RK = 2,
-};
-
 /** At startup, this is the current redo log file size.
 During startup, if this is different from srv_log_file_size_requested
 (innodb_log_file_size), the redo log will be rebuilt and this size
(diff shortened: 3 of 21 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -3909,22 +3909,23 @@ during startup, will be encrypted with tablespace key which has empty UUID
 bool
 innobase_fix_tablespaces_empty_uuid()
 {
+	if (Encryption::master_key_id == 0) {
+		/* We have to call log_enable_encryption_if_set during every
+		 startup, to report errors genrated in this function correctly.
+		 Without this call here, some illegal configurations, such as
+		 enabling encryption without a keyring are silently accepted,
+		 and result in errors later during the servers run.
+		   log_enable_encryption_if_set is also called later, when the
```
laurynas-biveinis: Formatting?

laurynas-biveinis (changes requested):

satya-bodapati (changes requested):
Am not able to understand how this fix works.
Shouldn't there be a Encryption::check_keyring() call on redo log path instead of this change?

dutow:
> Am not able to understand how this fix works.
> Shouldn't there be a Encryption::check_keyring() call on redo log path instead of this change?

There are multiple problems here:

* log_enable_encrpytion_if_set does many things. Because of this, it has to be called for every startup, otherwise, the server doesn't report some errors, and silently continues in an invalid configuration. In the current version of this function, we never call it in some cases, such as when there's no keyring specified. (after a long discussion with you about this in the previous PR changing this method - I missed this issue)
* log_enable_encryption_if_set has to be called multiple times. Originally I thought that it's enough if we do it only for rotated keys, but I was wrong - if we call it only once, the master_key encryption also has issues. (upstream calls this method in every second during the servers entire run - so that's why there's "no" issue there, and that's why they have short sleeps in the log encryption tests)
* log_enable_encryption_if_set is also upstream in 8.0, and backported to 5.7. If we refactor it (split it into multiple methods, called at different places), we'll have more trouble maintaining upstream changes.

So I see 3 ways to fix the redo log encryption issues I see:
* call it twice in this method (once early, once after we have the server uuid)
* call it once elsewhere (early), and once here after we have the server uuid
* refactor that method completely (and call different parts at different places)

To me, the first seemed the simplest.

satya-bodapati:
@dutow thanks for detailed reply. The return type of the log_encryption_if_set() is ignored. How does it abort incase of configuration issues? Does it do _exit() ?

dutow:
@satya-bodapati It's ignored because it's void. When it encounters an error, it simply displays an error message, and sets srv_redo_log_encrpyt to false. So the server continues to run, but if it later actually encounters an encrypted redo log block, it fails there. The testcase for ps-5476 is also a good example for this behavior: if there isn't anything encrypted in the log right now, the error message is displayed, but the server will continue to run, without encryption.

satya-bodapati:
@dutow Hmm, I guess this is the behavior of all encryption features to just throw error or make tablespace unavailable.

Please add two sections (Problem and Fix) in all your (non-trivial) commit messages. Makes it easier for reviewers or future coders.

mysql-test/suite/innodb/t/percona_bug_ps5476.test
```diff
@@ -0,0 +1,4 @@
+call mtr.add_suppression("Encryption can't find master key, please check the keyring plugin is loaded.");
+call mtr.add_suppression("Can't set redo log tablespace to be encrypted.");
```
satya-bodapati: Please add grep inc for these error messages.

Instead of number for test file names, it is better to use text. may be redo_log_encrypt_failure.test  or something like that..

satya-bodapati: ofcourse dont forget percona_ prefix for our test :)

storage/innobase/handler/ha_innodb.cc
```diff
@@ -3909,22 +3909,23 @@ during startup, will be encrypted with tablespace key which has empty UUID
 bool
 innobase_fix_tablespaces_empty_uuid()
 {
+	if (Encryption::master_key_id == 0) {
```
satya-bodapati, line 3922: Why do you want to move this condition above srv_read_only mode? is there any effect?

dutow: Yes. log_enable_encryption_if_set reports an error redo log encryption is on in read only mode.

satya-bodapati: Ah right, encryption can be enabled in read-only mode as well..

mysql-test/include/log_encrypt_content_test.inc
```diff
@@ -0,0 +1,35 @@
...
+INSERT INTO t1 select * from t1;
+INSERT INTO t1 select * from t1;
+INSERT INTO t1 select * from t1;
+
+SELECT * FROM t1 LIMIT 10;
+
+# Check file content
+--let $MYSQLD_DATADIR= `select @@datadir`
+--let SEARCH_PATTERN= asdfghjkl
+#--let SEARCH_RANGE= 12582912
```
laurynas-biveinis:
```suggestion
```

laurynas-biveinis (changes requested):

dutow:
Updated with backported upstream changes from 8.0.

https://ps57.cd.percona.com/job/percona-server-5.7-param/34/

laurynas-biveinis:
@dutow do you have the list of backported commits?

dutow:
Yes, I mentioned it in the commit message: 92198f46da210b2db946192391b14e7345364b18

laurynas-biveinis:
Sorry for missing it there

storage/innobase/handler/ha_innodb.cc
```diff
@@ -3963,7 +3970,11 @@ innobase_fix_tablespaces_empty_uuid()
 		return(true);
 	}
 
-	log_enable_encryption_if_set();
+	if(srv_enable_redo_encryption()) {
```
laurynas-biveinis: Formatting

storage/innobase/include/srv0srv.h
```diff
@@ -1053,6 +1047,13 @@ bool
 srv_is_undo_tablespace(
 	ulint	space_id);
 
+bool
+srv_enable_redo_encryption_mk();
+bool
+srv_enable_redo_encryption_rk();
+bool
```
laurynas-biveinis: New style for new functions (return type and function name on the same line)
Also while otherwise it could be possible to look the other way re. missing doc comments, now the relationship between three functions needs to be clarified.

Or maybe _mk _rk functions can be made statis srv0srv.cc?

storage/innobase/srv/srv0srv.cc
```diff
@@ -3524,6 +3525,160 @@ srv_is_undo_tablespace(
 			      + srv_undo_tablespaces_open));
 }
 
+bool
+srv_enable_redo_encryption()
+{
+	if(srv_redo_log_encrypt == REDO_LOG_ENCRYPT_MK) {
```
laurynas-biveinis: Formatting

storage/innobase/srv/srv0srv.cc
```diff
@@ -3524,6 +3525,160 @@ srv_is_undo_tablespace(
...
+bool
+srv_enable_redo_encryption_mk()
+{
+	fil_space_t *space = fil_space_get(dict_sys_t::s_log_space_first_id);
+	if (FSP_FLAGS_GET_ENCRYPTION(space->flags))
+	{
+		return false;
+	}
+	byte key[ENCRYPTION_KEY_LEN];
+        byte iv[ENCRYPTION_KEY_LEN];
```
laurynas-biveinis: Formatting

storage/innobase/srv/srv0srv.cc
```diff
@@ -3524,6 +3525,160 @@ srv_is_undo_tablespace(
...
+		return false;
+	}
+
+	byte key[ENCRYPTION_KEY_LEN];
+        byte iv[ENCRYPTION_KEY_LEN];
+	uint version;
+
+	Encryption::random_value(key);
+
+        // load latest key & write version
```
laurynas-biveinis: Formatting

storage/innobase/srv/srv0srv.cc
```diff
@@ -3524,6 +3525,160 @@ srv_is_undo_tablespace(
...
+
+	char *redo_key_type = NULL;
+	byte *rkey = NULL;
+	size_t   klen = 0;
```
laurynas-biveinis: Formatting

storage/innobase/srv/srv0srv.cc
```diff
@@ -3524,6 +3525,160 @@ srv_is_undo_tablespace(
...
+				my_free(redo_key_type);
+			}
+			if (rkey != NULL) {
+				my_free(rkey);
+			}
+			return true;
+		} else {
+			DBUG_ASSERT(rkey != NULL);
+			byte *rkey2 = NULL;
+			size_t   klen2 = 0;
```
laurynas-biveinis: Formatting

storage/innobase/srv/srv0srv.cc
```diff
@@ -3524,6 +3525,160 @@ srv_is_undo_tablespace(
...
+	byte *rkey = NULL;
+	size_t   klen = 0;
+
+	if (my_key_fetch(PERCONA_REDO_KEY_NAME, &redo_key_type, NULL,
+				reinterpret_cast<void**>(&rkey), &klen) || rkey == NULL)
+	{
+		if (my_key_generate(PERCONA_REDO_KEY_NAME, "AES", NULL, ENCRYPTION_KEY_LEN)) {
+			ib::error() << "Redo log key generation failed.";
+			if (redo_key_type != NULL) {
+				my_free(redo_key_type);
```
laurynas-biveinis: my_free can handle NULL itself, will help for readability here and elsewhere

laurynas-biveinis (changes requested):
Please expand the commit message with what was taken from 92198f46da210b2db946192391b14e7345364b18 and how it was changed, because this is clearly not a 1:1 backport

storage/innobase/include/srv0srv.h
dutow: `_mk` and `_rk` is also used directly outside this file, I'll document them.

mysql-test/include/log_encrypt_content_test.inc
```diff
@@ -0,0 +1,34 @@
+
+CREATE TABLE t1(c1 INT, c2 char(20)) ENGINE = InnoDB;
+
+SHOW CREATE TABLE t1;
+INSERT INTO t1 VALUES(0, "asdfghjkl");
+INSERT INTO t1 VALUES(1, "qwertyuio");
+INSERT INTO t1 VALUES(2, "zxcvbnm");
+INSERT INTO t1 select * from t1;
```
satya-bodapati: Do we need these many insert selects? will FLUSH ENGINE LOGS help?

mysql-test/suite/innodb/t/percona_log_encrypt_failure.test
```diff
@@ -0,0 +1,4 @@
+call mtr.add_suppression("Encryption can't find master key, please check the keyring plugin is loaded.");
+call mtr.add_suppression("Can't set redo log tablespace to be encrypted.");
```
satya-bodapati, line 2: Please convert these suppressions to grep.inc

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21437,12 +21448,36 @@ innodb_temp_tablespace_encryption_update(
...
+		/* No change */
+		return;
+	}
+
+	if (target == REDO_LOG_ENCRYPT_OFF) {
+  		srv_redo_log_encrypt = REDO_LOG_ENCRYPT_OFF;
+ 		return;
+	}
+
+	if (srv_redo_log_encrypt != REDO_LOG_ENCRYPT_OFF
```
satya-bodapati, line 21469: Can we add a test for this?

storage/innobase/log/log0log.cc
```diff
@@ -1279,7 +1279,8 @@ It just need to flush the file header block with current master key.
 bool
 log_write_encryption(
 	byte*	key,
-	byte*	iv)
+	byte*	iv,
+	redo_log_encrypt_enum redo_log_encrypt)
```
satya-bodapati, line 1284: Please document this  parameter in the doxygen comments.

Also why do we have this variable when there is global variable srv_redo_log_encrypt?

storage/innobase/srv/srv0srv.cc
```diff
@@ -3524,6 +3525,160 @@ srv_is_undo_tablespace(
...
 }
 
+bool
+srv_enable_redo_encryption()
+{
+	if(srv_redo_log_encrypt == REDO_LOG_ENCRYPT_MK) {
+		return srv_enable_redo_encryption_mk();
+	}
+
+	if(srv_redo_log_encrypt == REDO_LOG_ENCRYPT_RK) {
```
satya-bodapati: formatting here too. Laurynas missed :)

satya-bodapati (changes requested):

storage/innobase/log/log0log.cc
dutow: Because this function can be called before srv_redo_log_encryption is set. This was a change introduced in 8.0.16, where it's a non-issue (it can only be on and off, and this method is only called when it's on), but we added additional options.

storage/innobase/handler/ha_innodb.cc
dutow: Yes, but this is only a partial fix, which I actually encountered because there was a related typo in one of the existing tests (which I rerecorded in this pull request).

There's also the case when the setting is set to a different encryption mode than what's in the currently existing redo log, which I didn't include in this commit (that's PS-5149). I'll add a testcase along with that fix. (I can move this check to that diff too, if that makes more sense).

mysql-test/include/log_encrypt_content_test.inc
dutow: I simply used the same amount of inserts as in upstream log_encrypt_2 test, with different texts. I can change it if you think that's better.

satya-bodapati: yes, please try with fewer and then FLUSH ENGINE LOGS.

storage/innobase/handler/ha_innodb.cc
satya-bodapati: Ok to be deferred to PS-5149

storage/innobase/log/log0log.cc
satya-bodapati: Confused here. How can log_write_encryption() be called  before if srv_red_log_encryption is  SET?

dutow: That's how upstream refactored the setter: now when the user changes the redo_log_encryption setting, first the initialization is completed (including writing out the encryption header), and after that, the variable is changed.

dutow:
Updated.8

storage/innobase/srv/srv0srv.cc
```diff
@@ -3524,10 +3524,145 @@ srv_is_undo_tablespace(
...
+
+	return false;
+}
+
+bool
+srv_enable_redo_encryption_mk()
+{
+	fil_space_t *space = fil_space_get(dict_sys_t::s_log_space_first_id);
+	if (FSP_FLAGS_GET_ENCRYPTION(space->flags))
+	{
```
satya-bodapati: Move the brace to above line

storage/innobase/srv/srv0srv.cc
```diff
@@ -3524,10 +3524,145 @@ srv_is_undo_tablespace(
...
+
+		ib::error() << "Can't set redo log tablespace to be encrypted.";
+		return true;
+	}
+
+	space->flags |= FSP_FLAGS_MASK_ENCRYPTION;
+
+	const dberr_t err = fil_set_encryption(space->id, Encryption::AES, key, iv);
+	if (err != DB_SUCCESS) {
+		ib::error() << "Can't set redo log tablespace to be encrypted.";
```
satya-bodapati, line 3565: Can you print redo tablespace name or file name here?

mysql-test/include/log_encrypt_content.inc
```diff
@@ -0,0 +1,34 @@
+# InnoDB transparent tablespace data encryption
```
satya-bodapati: please add percona prefix to .inc files added by us(got this from laurynas :))

satya-bodapati (approved):
Looks good. Please address the review comments and push

PR #3116, opened 2019-04-03
PS-5174 - Compressed column validation on partition

altmannmarcelo:
Engine support validation for compressed columns is not performed
when creating tables with partition. Partitioned tables allow multiple
combinations of storage engines, which is syntactically correct.
When omitting table SE on non partition table, @@default_storage_engine
will be used as default. But for partitioned tables,
you can omit default SE, and pass a SE on partition clause.
In this case, instead of using @@default_storage_engine as table SE,
the SE passed on partition clause will be passed as table SE.
This commit implements the compressed column check after the partition
info has been analyzed.


altmannmarcelo:
https://ps.cd.percona.com/job/percona-server-8.0-param/261/

satya-bodapati (changes requested):
In 8.0, you cannot have partitions with combination of different storage engines. So the comment in commit message is wrong. Please remove that.

1. Can you add testcase?

2. If you enable native partitioning, does this bug appear in 5.7 too?

sql/sql_table.cc
```diff
@@ -8106,6 +8106,14 @@ static bool create_table_impl(
       }
       create_info->db_type = engine_type;
     }
+    if (alter_info->has_compressed_columns() &&
```
satya-bodapati, line 8109: Is it possible to do this check in native Partitioning code? (see sql/partitioning/partition_base.cc) Partition_base::create() ?

satya-bodapati (changes requested):

mysql-test/suite/rocksdb/t/partition.test
```diff
@@ -658,6 +658,21 @@ INSERT INTO trb1 VALUES
 
 ALTER TABLE trb1 ADD PRIMARY KEY (id);
 
+--echo #
+--echo # PS-5174: Check COMPRESSED COLUMNS support on partitioned tables
+--echo #
+CREATE COMPRESSION_DICTIONARY numbers ('');
```
satya-bodapati: Hmm, we allow  empty dictionary?  we shouldn't have :(
Suggest to use some text.

mysql-test/suite/rocksdb/t/partition.test
```diff
@@ -658,6 +658,21 @@ INSERT INTO trb1 VALUES
 
 ALTER TABLE trb1 ADD PRIMARY KEY (id);
 
+--echo #
```
laurynas-biveinis: Please make a common test include file shared between RocksDB and TokuDB (please check; there must be existing native partitioning examples).

Please then also start at 5.7, where only the testcase would be present.

mysql-test/suite/rocksdb/t/partition.test
laurynas-biveinis: @satya-bodapati can you please file a bug re. empty dictionaries?

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb/t/partition.test
satya-bodapati: @laurynas-biveinis done! https://jira.percona.com/browse/PS-5546

sql/sql_table.cc
altmannmarcelo: Hi @satya-bodapati . The issue is not specific for native partition. On 5.7 code we have the same validation in `create_table_impl() ` which is why it was ported to 8.0 in this PR. If we move the check into `partition_base.cc` upstream storage engines won't be checked. For example, InnoDB code doesn't enter `Partition_base::create()` . IMHO we should keep compatibility with 5.7. Here are the tests you requested:

```
#toku without native partition
mysql [localhost:5725] {msandbox} ((none)) > use test;
Database changed
mysql [localhost:5725] {msandbox} (test) > SELECT @@tokudb_enable_native_partition;
+----------------------------------+
| @@tokudb_enable_native_partition |
+----------------------------------+
|                                0 |
+----------------------------------+
1 row in set (0.00 sec)

mysql [localhost:5725] {msandbox} (test) > CREATE TABLE t2(id INT UNSIGNED,a VARCHAR(1)COLUMN_FORMAT COMPRESSED WITH COMPRESSION_DICTIONARY numbers) ENGINE=TokuDB PARTITION BY HASH (id) PARTITIONS 2;
ERROR 1478 (HY000): Table storage engine 'TokuDB' does not support the create option 'COMPRESSED COLUMNS'


#Toku with native partition

mysql [localhost:5725] {msandbox} ((none)) > use test;
Database changed
mysql [localhost:5725] {msandbox} (test) > SELECT @@tokudb_enable_native_partition;
+----------------------------------+
| @@tokudb_enable_native_partition |
+----------------------------------+
|                                1 |
+----------------------------------+
1 row in set (0.00 sec)

mysql [localhost:5725] {msandbox} (test) > CREATE TABLE t2(id INT UNSIGNED,a VARCHAR(1)COLUMN_FORMAT COMPRESSED WITH COMPRESSION_DICTIONARY numbers) ENGINE=TokuDB PARTITION BY HASH (id) PARTITIONS 2;
ERROR 1478 (HY000): Table storage engine 'TokuDB' does not support the create option 'COMPRESSED COLUMNS'


#Rocks without native partition
mysql [localhost:5725] {msandbox} ((none)) > use test;
Database changed
mysql [localhost:5725] {msandbox} (test) > SELECT @@rocksdb_enable_native_partition;
+-----------------------------------+
| @@rocksdb_enable_native_partition |
+-----------------------------------+
|                                 0 |
+-----------------------------------+
1 row in set (0.01 sec)

mysql [localhost:5725] {msandbox} (test) > CREATE TABLE t1(id INT UNSIGNED,a VARCHAR(1)COLUMN_FORMAT COMPRESSED WITH COMPRESSION_DICTIONARY numbers) ENGINE=RocksDB PARTITION BY HASH (id) PARTITIONS 2;
ERROR 1478 (HY000): Table storage engine 'ROCKSDB' does not support the create option 'COMPRESSED COLUMNS'

#Rocks with native partition 
mysql [localhost:5725] {msandbox} ((none)) > use test
Database changed
mysql [localhost:5725] {msandbox} (test) > SELECT @@rocksdb_enable_native_partition;
+-----------------------------------+
| @@rocksdb_enable_native_partition |
+-----------------------------------+
|                                 1 |
+-----------------------------------+
1 row in set (0.00 sec)

mysql [localhost:5725] {msandbox} (test) > CREATE TABLE t1(id INT UNSIGNED,a VARCHAR(1)COLUMN_FORMAT COMPRESSED WITH COMPRESSION_DICTIONARY numbers) ENGINE=RocksDB PARTITION BY HASH (id) PARTITIONS 2;
ERROR 1478 (HY000): Table storage engine 'ROCKSDB' does not support the create option 'COMPRESSED COLUMNS'
```

mysql-test/suite/rocksdb/t/partition.test
altmannmarcelo: @laurynas-biveinis Done - #3132

sql/sql_table.cc
satya-bodapati: Ok, better to leave validation to individual SE's

satya-bodapati:
I think you misunderstood Laurynas comment. What he meant is

1. have an inc file in parts suite (this will have test with $engine parameter)
2. Create one test in rocksdb suite using the .inc file
3. Create another test in tokudb suite using the same .inc file.

For example see https://github.com/percona/percona-server/pull/3112/

laurynas-biveinis:
@satya-bodapati, https://github.com/percona/percona-server/pull/3132 looks good to me?

satya-bodapati:
> @satya-bodapati, #3132 looks good to me?

I see, there is new PR, that works too

satya-bodapati (approved):

altmannmarcelo:
@laurynas-biveinis . Added the test cases from #3132. Ready for next review.

https://ps80.cd.percona.com/job/percona-server-8.0-param/75/

altmannmarcelo:
@laurynas-biveinis . Added the test cases from #3132. Ready for next review.

https://ps80.cd.percona.com/job/percona-server-8.0-param/75/

Latest force push I changed the commit message only

laurynas-biveinis (approved):

PR #3117, opened 2019-04-03
PS-5148 : [PS8QA] compression dictionary operations not handled under innodb-force-recovery

satya-bodapati:
Problem:
--------
Compression dictionary operations are not handled when innodb_force_recovery > 0.

Fix:
----
We treat compression dictionary operations like DDL operations (The reason to allow DDL is to be able
to drop corrupted table).

For value <  3, compression dictionary operations are allowed, DDLs are allowed
For value == 3, compression dictionary operations are not allowed, DDLs are allowed (On why compression dictionary operation are not allowed, see below)
For values > 3, InnoDB is in read-only mode

Reason for disallowing compression dictionary operation in innodb_force_recovery == 3 
--------------
other DDLs would need rsegs too. In case of other DDLs, the trx is not active and the rsegs are assigned via different path.

For compression dictionary delete operation, first there is search operation on the row, makes the trx active (without rsegs assigned) and then the delete operation, cannot assign rsegs anymore.

Infact, this should have affected regular BEGIN; SELECT FOR UPDATE; DELETE;. What prevents the issue is,  DELETE on user tables is not allowed. If the check is skipped, we get same issue/assert failure: ut_ad(trx_is_rseg_assigned(trx));

### DIFF
--- a/sql/sql_zip_dict.cc
+++ b/sql/sql_zip_dict.cc
@@ -426,10 +426,18 @@ int create_zip_dict(THD *thd, const char *name, ulong name_len,
         break;
       case HA_ERR_RECORD_FILE_FULL:
         error = ER_RECORD_FILE_FULL;
-        my_error(error, MYF(0), "compression_dictionary");
+        my_error(error, MYF(0), COMPRESSION_DICTIONARY_TABLE);
         break;
       case HA_ERR_TOO_MANY_CONCURRENT_TRXS:
         error = ER_TOO_MANY_CONCURRENT_TRXS;
+        my_error(error, MYF(0));
+        break;
+      case HA_ERR_TABLE_READONLY:
+        error = ER_OPEN_AS_READONLY;
+        my_error(error, MYF(0), COMPRESSION_DICTIONARY_TABLE);
+        break;
+      case HA_ERR_INNODB_FORCED_RECOVERY:
+        error = ER_INNODB_FORCED_RECOVERY;
         my_error(error, MYF(0));
         break;
       default:
@@ -578,6 +586,14 @@ int drop_zip_dict(THD *thd, const char *name, ulong name_len, bool if_exists) {
       break;
     case HA_ERR_TOO_MANY_CONCURRENT_TRXS:
       error = ER_TOO_MANY_CONCURRENT_TRXS;
+      my_error(error, MYF(0));
+      break;
+    case HA_ERR_TABLE_READONLY:
+      error = ER_OPEN_AS_READONLY;
+      my_error(error, MYF(0), COMPRESSION_DICTIONARY_TABLE);
+      break;
+    case HA_ERR_INNODB_FORCED_RECOVERY:
+      error = ER_INNODB_FORCED_RECOVERY;
       my_error(error, MYF(0));
       break;
     default:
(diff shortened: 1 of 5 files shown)


satya-bodapati:
@laurynas-biveinis ready for review

sql/sql_zip_dict.cc
```diff
@@ -426,12 +426,20 @@ int create_zip_dict(THD *thd, const char *name, ulong name_len,
...
       case HA_ERR_TOO_MANY_CONCURRENT_TRXS:
         error = ER_TOO_MANY_CONCURRENT_TRXS;
         my_error(error, MYF(0));
         break;
+      case HA_ERR_TABLE_READONLY:
+        error = ER_OPEN_AS_READONLY;
+        my_error(error, MYF(0), COMPRESSION_DICTIONARY_TABLE);
+        break;
+      case HA_ERR_INNODB_FORCED_RECOVERY:
+	error = ER_INNODB_FORCED_RECOVERY;
```
laurynas-biveinis: Formatting

sql/sql_zip_dict.h
```diff
@@ -57,6 +57,13 @@ inline bool is_hardcoded(const dd::String_type &schema,
                                 table_name == "compression_dictionary_cols"));
 }
 
+/** @return true if the table is hardcoded compression dictionary tables
```
laurynas-biveinis: "is a hardcoded compression dictionary table"
or
"is one of hardcoded compression dictionary tables"

storage/innobase/row/row0mysql.cc
```diff
@@ -2698,7 +2701,19 @@ static dberr_t row_update_for_mysql_using_upd_graph(const byte *mysql_rec,
...
       !(srv_force_recovery < SRV_FORCE_NO_UNDO_LOG_SCAN &&
-        dict_sys_t::is_dd_table_id(prebuilt->table->id))) {
+        (dict_sys_t::is_dd_table_id(prebuilt->table->id) ||
+         compression_dict::is_hardcoded(prebuilt->table->name.m_name)))) {
+    ib::error(ER_IB_MSG_985) << MODIFICATIONS_NOT_ALLOWED_MSG_FORCE_RECOVERY;
+    DBUG_RETURN(DB_READ_ONLY);
+  }
+
+  /* For compression dictionary delete, trx is already active because of the
+  search before delete . With the trx active, we cannot assign rseg with
```
laurynas-biveinis:
```suggestion
  search before delete. With the trx active, we cannot assign rseg with
```

laurynas-biveinis (commented):
Why at innodb_force_recovery = 3, other DDLs work? Don't they need rseg too for DD update?

satya-bodapati:
> Why at innodb_force_recovery = 3, other DDLs work? Don't they need rseg too for DD update?

Good question. Right, they would need rsegs too. In case of other DDLs, the trx is not active and the rsegs are assigned via different path.

For compression dictionary delete operation, first there is search operation on the row, makes the trx  active (without rsegs assigned) and then the delete operation, cannot assign rsegs anymore.

Infact, this should have affected egular BEGIN; SELECT FOR UPDATE; DELETE;. What prevents the issue is, DELETE on user tables is not allowed. If the check is skipped, we get same issue/assert failure:  ut_ad(trx_is_rseg_assigned(trx));

satya-bodapati:
@laurynas-biveinis addressed review comments

laurynas-biveinis (approved):

satya-bodapati:
Thanks @laurynas-biveinis!

PR #3118, opened 2019-04-03
PS-5537 : Installation of 'keyring_vault' test for Centos6

george-lorch:
- Moved CMake test suite install directives before all checks to build or not
  to ensure tests are always installed regardless if the plugin is built or not.


george-lorch:
Here is jenkins at the GCA point which this branch is based upon https://ps57.cd.percona.com/job/percona-server-5.7-param/5/

And here is jenkins for what it would look like after merge to HEAD https://ps57.cd.percona.com/job/percona-server-5.7-param/6/

laurynas-biveinis (approved):
Please add a short comment before the moved snippet so that somebody does not come later to optimize things by moving it back to its former place. No need for further reviews

PR #3121, opened 2019-04-04
PS-5485 : exposing undo log tablespace encryption status through inno…

satya-bodapati:
…db_tablespaces

PS-5490 : enabling temp-tablespace encryption doesn't mark innodb_temporary tablespace encryption flag

Problem:
--------
When temporary tablespace and undo tablespaces are encrypted, I_S views does not show proper status in "encryption" column output.

Fix:
----
We will use cached tablespace flags to retreive encryption status.
When upstream fix for Bug#94665 arrives, remove the fsp_is_undo_tablespace() check introduced
in this patch.

### DIFF
--- a/mysql-test/suite/innodb/r/temp_table_encrypt.result
+++ b/mysql-test/suite/innodb/r/temp_table_encrypt.result
@@ -24,6 +24,7 @@ Warning	1210	 Temporary tablespace couldn't be encrypted. Check if keyring plugi
 SELECT @@innodb_temp_tablespace_encrypt;
 @@innodb_temp_tablespace_encrypt
 0
+include/assert.inc [Temporary tablespace should be unencrypted]
 CREATE TEMPORARY TABLE t01 (a INT) ENGINE=InnoDB ENCRYPTION='Y';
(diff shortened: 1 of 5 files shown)


satya-bodapati:
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/780/ (with innodb only suites, Debug build)

laurynas-biveinis (commented):
No encrypted undo tablespace tests to update?

satya-bodapati:
> No encrypted undo tablespace tests to update?

There are two tests but since the undo encryption is from background threads, the assert.inc might fail randomly.

laurynas-biveinis:
Then it should be possible to write a test with wait_condition.inc waiting for background threads to complete their actions?

satya-bodapati:
> Then it should be possible to write a test with wait_condition.inc waiting for background threads to complete their actions?

right, I can do that. Currently a sleep is used. Then there is no point in using 'assert.inc' right?

laurynas-biveinis:
Right, in such test wait_condition.inc both synchronizes and tests.

satya-bodapati:
Added undo test. Ready for review @laurynas-biveinis

laurynas-biveinis (approved):

PR #2729, opened 2018-11-27
Merge upstream 5.7.24 into Percona Server

percona-ysorokin:
https://jira.percona.com/browse/PS-4709

PR #3122, opened 2019-04-04
Null-merge branch 5.7 (commit fc5fca4) into 8.0

percona-ysorokin:
This is a null merge after merging upstream 5.7.25 and 8.0.15.


mysql-test/include/log_encrypt_4.inc
```diff
@@ -70,7 +70,7 @@ DROP TABLE tab1,tab2;
 let NEW_CMD = $MYSQLD $extra_args --initialize-insecure --datadir=$MYSQLD_DATADIR2 --init-file=$BOOTSTRAP_SQL  --secure-file-priv="" --innodb_redo_log_encrypt=$redo_log_mode >>$MYSQLD_LOG 2>&1;
 
 --echo # Run the bootstrap command of datadir2, it should fail since the keyring is not loaded.
---error 1,42
```
laurynas-biveinis, line 73: But 5.7 trunk is also "1,42" here - please check that the null merge does not revert 8.0 to some earlier 5.7 state

mysql-test/include/log_encrypt_6.inc
```diff
@@ -85,7 +85,7 @@ DROP DATABASE tde_db;
 
 --echo # Try starting without keyring : Error
 let NEW_CMD = $MYSQLD --no-defaults --innodb_dedicated_server=OFF --innodb_page_size=$START_PAGE_SIZE --innodb_log_file_size=$LOG_FILE_SIZE --basedir=$MYSQLD_BASEDIR --datadir=$MYSQLD_DATADIR1  --secure-file-priv="" --console </dev/null>>$MYSQL_TMP_DIR/wl9290.err 2>&1;
---error 1,42
```
laurynas-biveinis, line 88: Likewise

mysql-test/suite/innodb/t/alter_kill.test
```diff
@@ -141,7 +141,7 @@ XA PREPARE 'x';
 
 -- echo # Attempt to start without an *.ibd file.
 let SEARCH_FILE= $MYSQLTEST_VARDIR/log/my_restart.err;
--- error 1,42
```
laurynas-biveinis, line 144: Please check 5.7 HEAD, where it is "-- error 3"

mysql-test/suite/innodb/t/bootstrap.test
```diff
@@ -35,7 +35,7 @@ EOF
 write_file $bugdir/data/mysql/plugin.frm;
```
laurynas-biveinis, line 35: What is the current path of this file in 5.7 HEAD?

mysql-test/suite/innodb/t/log_corruption_1.test
```diff
@@ -30,7 +30,7 @@ EOF
 # platform sporadically
```
laurynas-biveinis, line 30: What is the path of this file in 5.7 HEAD?

mysql-test/suite/innodb/t/tablespace_encrypt_9.test
```diff
@@ -79,7 +79,7 @@ SET GLOBAL innodb_master_thread_disabled_debug = 1;
 --error 0,CR_SERVER_LOST,ER_INTERNAL_ERROR
```
laurynas-biveinis, line 79: What is the path of this file in 5.7 HEAD?

mysql-test/suite/innodb/t/temporary_table.test
```diff
@@ -633,7 +633,7 @@ drop table t1;
 #
```
laurynas-biveinis, line 633: What is the path of this file in 5.7 HEAD?

mysql-test/suite/innodb/t/undo_encrypt_bootstrap.test
```diff
@@ -42,7 +42,7 @@ EOF
 let NEW_CMD = $MYSQLD --no-defaults --innodb_dedicated_server=OFF --initialize-insecure --secure-file-priv="" --innodb_log_file_size=$LOG_FILE_SIZE --innodb_page_size=$START_PAGE_SIZE --innodb_data_home_dir=$MYSQLD_HOME_DATA_DIR --innodb_undo_directory=$MYSQLD_UNDO_DATADIR --basedir=$MYSQLD_BASEDIR --datadir=$MYSQLD_DATADIR --init-file=$BOOTSTRAP_SQL --innodb_undo_log_encrypt=ON </dev/null>>$MYSQLTEST_VARDIR/tmp/bootstrap.log 2>&1;
```
laurynas-biveinis, line 42: What is the path of this file in 5.7 HEAD?

plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xdr_utils.h
```diff
@@ -33,7 +33,6 @@ extern "C" {
     Initialize an array
  */
 #define def_init_xdr_array(name) \
-  MY_ATTRIBUTE((unused))         \
```
laurynas-biveinis, line 36: No compilation warning regression?

storage/heap/ha_heap.h
```diff
@@ -63,6 +63,7 @@ class ha_heap : public handler {
                 : HA_ONLY_WHOLE_INDEX | HA_KEY_SCAN_NOT_ROR);
   }
   uint max_supported_keys() const { return HP_MAX_KEY; }
+  uint max_supported_key_length() const { return HP_MAX_KEY_LENGTH; }
```
laurynas-biveinis, line 66: How did this appear here? Which commit in 5.7?

storage/innobase/btr/btr0scrub.cc
```diff
@@ -52,22 +52,6 @@ static btr_scrub_stat_t scrub_stat;
 static ib_mutex_t scrub_stat_mutex;
 mysql_pfs_key_t scrub_stat_mutex_key;
 
-#ifdef UNIV_DEBUG
```
laurynas-biveinis, line 55: Heads-up @robgolebiowski, you might need to rebase once this is merged

mysql-test/suite/innodb/t/alter_kill.test
percona-ysorokin: This is how it is in Oracle MySQL: 5.7 has `--error 3`, whereas 8.0 has `--error 1`

plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xdr_utils.h
percona-ysorokin: Yes, everything is clear. Removing `MY_ATTRIBUTE((unused)) ` was our quick and dirty way of suppressing modern gcc / clang warnings, then Oracle fixed this properly.
https://github.com/mysql/mysql-server/commit/1f610f7
We reverted these changes in 5.7, but never in 8.0

storage/heap/ha_heap.h
percona-ysorokin:
```
'storage/heap/ha_heap.h'
"uint max_supported_key_length() const { return HP_MAX_KEY_LENGTH; }"
method override in 'ha_heap' class never got to 8.0.
See "Import memory_dynamic_rows.patch"
```
(commit 51b5cb2).

mysql-test/suite/innodb/t/bootstrap.test
percona-ysorokin: In 5.7 it used to be `mysql-test/suite/innodb/t/log_file.test`

mysql-test/include/log_encrypt_4.inc
percona-ysorokin: Double checked with `mtr --debug-server --force log_encrypt_4_rk log_encrypt_4_mk --sanitize`.
42 here is needed to suppress the same `sync_latch_meta_init()` memory leak.

mysql-test/include/log_encrypt_6.inc
percona-ysorokin: Double checked with `mtr --debug-server --force log_encrypt_6_rk log_encrypt_6_mk --sanitize`.
42 here is needed to suppress the same `sync_latch_meta_init()` memory leak.

mysql-test/suite/innodb/t/alter_kill.test
percona-ysorokin: The same `sync_latch_meta_init()` memory leak.

mysql-test/suite/innodb/t/bootstrap.test
percona-ysorokin: The same `sync_latch_meta_init()` memory leak. Possibly others as well.

mysql-test/suite/innodb/t/log_corruption_1.test
percona-ysorokin: This is a new 8.0 file but the problem is the same `sync_latch_meta_init()` memory leak.

mysql-test/suite/innodb/t/tablespace_encrypt_9.test
percona-ysorokin: This is a new 8.0 file but the problem is the same sync_latch_meta_init() memory leak.

mysql-test/suite/innodb/t/temporary_table.test
percona-ysorokin: This is a new 8.0 file but the problem is the same `sync_latch_meta_init()` memory leak.

mysql-test/suite/innodb/t/undo_encrypt_bootstrap.test
percona-ysorokin: This test case does not have memory leaks. So, 42 is not needed here even for suppression

percona-ysorokin:
Jenkins build link
https://ps80.cd.percona.com/job/percona-server-8.0-param/59/

percona-ysorokin:
Generally speaking, `--error ...,42` should never be used in MTR test cases in both 5.7 and 8.0.

mysql-test/include/log_encrypt_4.inc
laurynas-biveinis: Then the state of the current 5.7 trunk and this PR suggests that in 5.7 we reverted sync_latch_meta_init fix (due to upstream, right?), and in 8.0 there is a better fix?

storage/heap/ha_heap.h
laurynas-biveinis: That's the original patch import to 5.5. But this being null merge suggests that this should have appeared in the 5.7 commit range being merged. ATM I cannot exclude max_supported_key_length being dropped on purpose from 8.0 on purpose, for example

percona-ysorokin: @laurynas-biveinis 
5.5, 5.6 and 5.7 branches have this have this `max_supported_key_length()` method overridden in `ha_heap` class.
The only commit I found that removes this override is your Initial "Percona Server 8.0.11 tree"
`git show 4ad1d0ccd476ae4990dac354b382d0e2d7220be8 -- storage/heap/ha_heap.h`

```diff
...
-   uint max_supported_keys() const { return MAX_KEY; }
-   uint max_supported_key_part_length() const { return MAX_KEY_LENGTH; }
 -  const key_map *keys_to_use_for_scanning() { return &btree_keys; }
 -  uint max_supported_keys()          const { return HP_MAX_KEY; }
 -  uint max_supported_key_length() const { return HP_MAX_KEY_LENGTH; }
++  uint max_supported_keys() const { return HP_MAX_KEY; }
+   uint max_supported_key_part_length() const { return HP_MAX_KEY_LENGTH; }
 -  double scan_time()
 -  { return (double) (stats.records+stats.deleted) / 20.0+10; }
 -  double read_time(uint index, uint ranges, ha_rows rows)
 -  { return (double) rows /  20.0+1; }
 +  double scan_time() {
 +    return (double)(stats.records + stats.deleted) / 20.0 + 10;
 +  }
 +  double read_time(uint, uint, ha_rows rows) { return (double)rows / 20.0 + 1; }
...
```
Unfortunately, there is nothing related to this file in the comments for this commit.

Another version of this weirdness is Oracle force-pushing 5.7.24. There were 2 orphan commits, which I reverted as part of the upstream 5.7.24 merge (https://github.com/percona/percona-server/pull/2729). One of them have changes very close to `max_supported_key_length()` method
(https://github.com/percona/percona-server/pull/2729/commits/e13e014b0f65ce4142f519a256655d6c38a8042b#diff-bb6ecd31cb792b2abb4bc7b806e7b5ff).

mysql-test/include/log_encrypt_4.inc
percona-ysorokin: No, @laurynas-biveinis . The story was slightly different.
We had `sync_latch_meta_init()` memory leak problem reported in one of our PS-XXX issues.
In https://jira.percona.com/browse/PS-5008 in particular.
I also reported this to upstream (https://bugs.mysql.com/bug.php?id=93165), but they never agreed that this is a real problem and despite all my arguments never agreed to fix this.
Instead in one of the upcoming 5.7 releases they brutally suppressed all these issues by adding `42` to `--error` for a bunch of MTR test cases (I believe the same was done in 8.0 as well).
When we were merging this 5.7 upstream release to Percona Server, we just reverted those suppressions.
Actual memory leak has never been fixed neither in 5.7 nor in 8.0. We still need to do this.

Personally, I am 100% confident that this is the right way to go. MTR code should never include these `--error xxx, 42` ASan suppressions. The main reasons is: once we add this `42` to mark that this test is OK to leak memory, we will never be able to detect if our new changes introduce a new (previously non-existing) leak.

laurynas-biveinis: OK, then the only remaining issue is that 5.7 HEAD still has --error 42, at least in log_encrypt_4.inc. This also has to be fixed in the scope of this upstream/null merge work.

storage/heap/ha_heap.h
laurynas-biveinis: The most likely reason I did this was because the parent class implementation is suitable for HEAP, ie. the key length is not restricted further. I'd leave this out but go ahead if you believe this must go back in

laurynas-biveinis (approved):
OK with ha_heap change reverted or left in.
Follow-up to remove last instances of --error 42 from 5.7 should be done too

PR #3126, opened 2019-04-08

robgolebiowski:


robgolebiowski:
It did not pick the new commit with amended message

PR #3127, opened 2019-04-08

robgolebiowski:
The problem was that FORCE is defined in global parser scope as a boolean
variable. The fix was to add a FORCE string option in set variable scope,
which overrides the global FORCE.


robgolebiowski:
I had to created a new branch as the old PR kept referencing the old commit with incorrect commit message.

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-trunk/12/

laurynas-biveinis (approved):
And please stop using trunk jobs for param testing! :)

robgolebiowski:
@laurynas-biveinis  this is the old job, I used yesterday. Since that I am not using trunk :)

PR #3128, opened 2019-04-08
PS-5547: redo log encryption is not working.

robgolebiowski:
The encryption was disabled for all logs. This problem was overlooked as
upstream MTR tests does not GREP the redo log for plaintext, only check
for encryption markers. Thus although the upstream tests passed, the
redo itself was not encrypted.

The tests for GREPing the redo log will be commited as part of PS-5476.

### DIFF
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -5926,9 +5926,12 @@ fil_io_set_encryption(
 		return;
 	}
 
-	/* Don't encrypt the log, page 0 of all tablespaces, all pages
+	/* Don't encrypt the page 0 of all tablespaces,
 	   don't encrypt TRX_SYS_SPACE.TRX_SYS_PAGE_NO as it contains address to dblwr buffer */
-	if (!req_type.is_log() && page_id.page_no() > 0 && (TRX_SYS_SPACE != page_id.space() || TRX_SYS_PAGE_NO != page_id.page_no())
+	if ((req_type.is_log()
+	     || (page_id.page_no() > 0
+		 && (TRX_SYS_SPACE != page_id.space()
+		     || TRX_SYS_PAGE_NO != page_id.page_no())))
 	    && space->encryption_type != Encryption::NONE) {
 		if (space->encryption_type == Encryption::KEYRING)  {
 			ut_ad(space->crypt_data != NULL);


robgolebiowski:
https://ps57.cd.percona.com/job/percona-server-5.7-param/7/

robgolebiowski:
https://ps57.cd.percona.com/job/percona-server-5.7-param/7/

storage/innobase/fil/fil0fil.cc
```diff
@@ -5928,7 +5928,7 @@ fil_io_set_encryption(
 
 	/* Don't encrypt the log, page 0 of all tablespaces, all pages
 	   don't encrypt TRX_SYS_SPACE.TRX_SYS_PAGE_NO as it contains address to dblwr buffer */
-	if (!req_type.is_log() && page_id.page_no() > 0 && (TRX_SYS_SPACE != page_id.space() || TRX_SYS_PAGE_NO != page_id.page_no())
+	if ((req_type.is_log() || (page_id.page_no() > 0 && (TRX_SYS_SPACE != page_id.space() || TRX_SYS_PAGE_NO != page_id.page_no())))
```
satya-bodapati: First I was wondering how long this bug has been there? i.e. since backport of redo encryption to 5.7?  git analysis showed that the regression is introduced as part of  this commit:
```
commit c7f44ee3974df2ffe9938f673d25cd5e47beeae1 (HEAD)
Author: Robert Golebiowski <robert.golebiowski@percona.com>
Date:   Thu Nov 16 17:53:11 2017 +0100

    PS-3829 : Innodb key rotation. ALPHA
    
    This PR implements Encryption threads for
    encryption/re-encryption/decryption of
    innodb tablespaces. It is based on MariaDB implementation ported from
    Google's
    patch. With this WL user can:
```

Is there another bug in this condition?

1. For system tablespace, the intention is to not encrypt pages upto TRX_SYS_PAGE_NO. But your condition also encrypts page 1, 2 ,3 and 4

2. This System tablespace check is done above. Can't we use that? Remove "space->crypt_data == NULL" ? Independent of MK or Keyring key encryption, we don't want these pages encrypted right?

3. Can you please reformat code to occupy < 80 chars. It has to be consistent with
InnoDB coding style.  In fact, entire commit "c7f44ee3974df2ffe9938f673d25cd5e47beeae1" didn't care about any coding guidelines or style issue.

satya-bodapati (changes requested):

storage/innobase/fil/fil0fil.cc
laurynas-biveinis: For encryption, only the touched lines in fixing PRs should be formatted OK, and the rest should used a separate commit with nothing else but clang-format. The reason is that a lot of unformatted code was pushed, fixing which would drown out any real changes

satya-bodapati: Right, I meant fixing the formatting of touched code.

dutow: Yes, since the merge. We merged keyring encryption and redo/undo encryption in quick succession (with only a few hour difference), and there were merge conflicts in several places. This was one of them, and somehow I resolved it incorrectly.

dutow: @robgolebiowski I'm not sure if the log condition is needed at all. The original redo diff simply deletes these from most places - I think we should do that here too. Upstream also removed it from this line in the log encryption diff.

dutow:
Also, this code seems broken in 8.0 too, I think we need a PR for that also.

* When keyring encryption is set, this function in 8.0 turns off log encryption
* When using master key encryption, this entire if block is skipped

storage/innobase/fil/fil0fil.cc
robgolebiowski: @dutow ack, will change that. @satya-bodapati thanks for comments, but they are regarding keyring encryption and we will deal with them when I start pushing fixes for it. I hope this is OK.

robgolebiowski: @dutow on the second look, they actually are checking for log write when they check the page ids:      (page_id.page_no() == 0 && !req_type.is_log())) {
    req_type.clear_encrypted();
I think this should stay as is then.

robgolebiowski:
Will prepare the path for 8.0 to then and update the affected versions in jira ticket.

storage/innobase/fil/fil0fil.cc
dutow: @robgolebiowski You are right, they have it there. (And also, looks like os0file makes sure that the first few log blocks are never encrypted, so it doesn't really matter what's here)

satya-bodapati: @robgolebiowski I need answer to the question if there is bug in TRX_SYS_PAGE check? Can this be simplified?

robgolebiowski:
As discuss on slack. The other issue is that keyring encryption is disabled for redo. This will be addressed in keyring workload.

storage/innobase/fil/fil0fil.cc
robgolebiowski: Pasting here the discussion from slack about TRX_SYS_PAGE:

Robert Golebiowski [9:25 AM]
hey @satya. To answer your question about TRX_SYS_PAGE_NO being not encrypted, while 1-4 pages being encrypted. This is the comment from MariaDB encryption implementation:
if (space_id == TRX_SYS_SPACE && offset == TRX_SYS_PAGE_NO) {
        /* don't encrypt this as it contains address to dblwr buffer */
        return;
    }
looks that when reaching out for the double write buffer the system space is still not encrypted - thus thus page needs to stay decrypted

Satya Bodapati [9:28 AM]
hi @robert.golebiowski I see. We have different path in system tablespace encryption. all pages  <= TRX_SYS_PAGE_NO can be left unencrypted.

Robert Golebiowski [9:29 AM]
yes, I have seen it, maybe we should consider encrypting them too.
anyways, system tablespace encryption is coming soon from upstream, we will see what they do

Satya Bodapati [9:30 AM]
I prefer we treat system tablespace in same way.
nope, upstream encryption is about mysql.ibd encryption

Robert Golebiowski [9:31 AM]
ah, right

Satya Bodapati [9:31 AM]
it is better if you can use same condition (<= TRX_SYS_PAGE_NO) in both MK and Keyringkey (edited) 

Robert Golebiowski [9:33 AM]
in my opinion we should have as smallest number of pages left unencrypted as possible. Thus I do not see reason to not encrypting 1-4. If any I would encrypt them for MK too. (edited) 

Satya Bodapati [9:35 AM]
well, we cannot change it now. there are headaches with minor upgrades if we make it too strict (only encrpt page 5).  I think there are issues with encrypting IBUF header pages. Dont remember details exactly

Robert Golebiowski [9:38 AM]
that would be surprising if there were as MariaDB is encrypting them. I think there should not be a problem if we start encrypting 1-4 now. The decrypt function will check the page's header and if it sees that the page is not encrypted then it will just return - so no problem with setting decryption information for not encrypted pages.

Satya Bodapati [9:41 AM]
ok, I guess we have to live the different types of sys encryption. both encrypt different header pages.

Robert Golebiowski [9:41 AM]
@satya I think we do not and we can encrypt pages 1-4. I will create a bug issue for it and have a look at this later

Satya Bodapati [9:42 AM]
ok, go ahead

robgolebiowski: PS-5557 was created for this issue.

robgolebiowski:
@laurynas-biveinis @satya-bodapati @dutow I think all the comments were answered (no need to re-run jenkins), please accept this PR if no more comments.

laurynas-biveinis (approved):

dutow (approved):

storage/innobase/fil/fil0fil.cc
satya-bodapati: The comment has to be fixed. "Don't encrypt the log, "

and one suggestion for better readability.

Please make it this way:

     /* For writing temporary tablespace, if encryption for temporary
          tablespace is disabled, skip setting encryption. */
          if (fsp_is_system_temporary(space->id)
              && !srv_tmp_tablespace_encrypt && req_type.is_write()) {
                  req_type.clear_encrypted();
                  return;
          }
  
          if (keyring && all cases to exempt) {
                   req_type.clear_encrypted();
                   return;
          }

          if (keyring_key) {             
           fil_io_set_keyring_key_encryption();
          } else {
           fil_io_set_mk_encryption():
          }

robgolebiowski: Thanks @satya-bodapati I think the refactoring is needed. I have only a problem with adding it as a part of this bug fix. Two things here 1) should not the bug fix resolve the bug and refactoring be a separate task ? 2) This problem affects only 5.7 - if I do the refactoring and merge it with 8.0 the footprint will be increased and it might seems that this fix was also fixing something in 8.0. I think this refactoring should be part of @dutow PR. @laurynas-biveinis what do you think ?

laurynas-biveinis: I agree re. splitting work into fixing and refactoring separately

robgolebiowski: Thanks - @dutow I can do it as a part of encryption threads refactoring should be cleaner.

satya-bodapati: here refactoring is actually a big word. This is very small one and indeed 8.0 has exactly what was needed. Am sure its get lost in *big* scheme of things..

Anyways..
1. Can you atleast fix the long line issue (only the code touched here?)
2. And the wrong comment

robgolebiowski: @satya-bodapati done

satya-bodapati (approved):
Ok with the formatting fixed

PR #3137, opened 2019-04-11
PS-5554: System keys not thread safe

robgolebiowski:
System keys are lazy initialized. This means creating the key's body
only when it is requested. The same key can be requested by multiple
threads. Which means the initialization can be requested by multiple threads
simultaneously. To cope with this a lock-free CAS was used. The same
system key can be requested by multiple threads but only one thread
will be first and thus will set the key body.

### DIFF
--- a/plugin/keyring/common/system_key_adapter.cc
+++ b/plugin/keyring/common/system_key_adapter.cc
@@ -16,6 +16,7 @@
 
 #include "system_key_adapter.h"
 #include "secure_string.h"
+#include "my_atomic.h"
 
 namespace keyring
 {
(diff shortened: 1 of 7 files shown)


plugin/keyring/common/system_key_adapter.cc
```diff
@@ -25,27 +26,62 @@ namespace keyring
...
+        keyring_key->get_key_data(), keyring_key->get_key_data_size());
+    
+    // need to "de"-xor keying key data
+    keyring_key->xor_data(system_key_data_candidate + system_key_data_version_prefix.length(),
+                          keyring_key->get_key_data_size());
+    // next xor system key data as a whole
+    keyring_key->xor_data(system_key_data_candidate, system_key_data_candidate_size);
+
+    void *null_system_key_data = NULL;
+    void *addr= & this->system_key_data.key_data;
```
laurynas-biveinis, line 50: Formatting (is it foo = or foo= ?)

plugin/keyring/common/system_key_adapter.cc
```diff
@@ -25,27 +26,62 @@ namespace keyring
...
+    void * volatile * typed_addr= static_cast<void * volatile *>(addr);
+
+    if(my_atomic_casptr(typed_addr, &null_system_key_data, system_key_data_candidate)) {
```
laurynas-biveinis: Formatting
```suggestion
    if (my_atomic_casptr(typed_addr, &null_system_key_data, system_key_data_candidate)) {
```

laurynas-biveinis (changes requested):
Only formatting comments, please resolve and push

PR #3140, opened 2019-04-12

robgolebiowski:
The problem was that FORCE is defined in global parser scope as a boolean
variable. The fix was to add a FORCE string option in set variable scope,
which overrides the global FORCE.

(cherry picked from commit eba1da1e3676e24675ae489efd8f025a81345035)


robgolebiowski:
Forgot that 5.7 also needs fixing, cherry-picked from 8.0

https://ps57.cd.percona.com/job/percona-server-5.7-param/13

PR #3141, opened 2019-04-16
Fix PS-5566 (INFORMATION_SCHEMA.GLOBAL_TEMPORARY_TABLES queries may c…

laurynas-biveinis:
…rash if online ALTER TABLE is running in parallel | Test percona_bug1294190 always failing)

8.0 introduced dd::Table argument on the common
handler::clone/handler::open code path, which is also used in querying
INFORMATION_SCHEMA.GLOBAL_TEMPORARY_TABLES. There, it causes a debug
build crash and almost certainly a release build failure in certain
scenario, i.e. by querying the I_S table while a parallel ALTER TABLE
is running. This happens because handler::clone always passes nullptr for
dd::Table, which then makes InnoDB DD code fail. Apparently such I_S
queries are the only case this happens.

We cannot fix this by making handler::clone e.g. query DD for the
needed dd::Table object because there is none: the temp table
dd::Table is registered with server DD only at the end of ALTER
TABLE. Until then, the dd::Table object only lives on the call
stack. Thus, fix by introducing tmp_dd_table_ptr field and
set_tmp_dd_table_ptr / get_tmp_dd_table_ptr setter/getter in
TABLE. Invoke the setter right before adding the temp table to THD
temp table list (which makes the table visible to I_S queries), and
get the dd::Table object in handler::clone.

This fixes main.percona_bug1294190 regression in 8.0.

https://ps80.cd.percona.com/job/percona-server-8.0-param/77/
>1000 failures due to recent debug build regression, reviewed the failure list that I_S.GLOBAL_TEMP_TABLES-related tests are absent

### DIFF
--- a/sql/handler.cc
+++ b/sql/handler.cc
@@ -2712,7 +2712,8 @@ handler *handler::clone(const char *name, MEM_ROOT *mem_root) {
     the same table instance. The ha_open call is not cachable for clone.
   */
   if (new_handler->ha_open(table, name, table->db_stat,
-                           HA_OPEN_IGNORE_IF_LOCKED, NULL))
+                           HA_OPEN_IGNORE_IF_LOCKED,
+                           table->get_tmp_dd_table_ptr()))
     goto err;
 
   DBUG_RETURN(new_handler);
--- a/sql/sql_base.cc
+++ b/sql/sql_base.cc
@@ -6966,6 +6966,7 @@ TABLE *open_table_uncached(THD *thd, const char *path, const char *db,
                                            : NON_TRANSACTIONAL_TMP_TABLE);
 
   if (add_to_temporary_tables_list) {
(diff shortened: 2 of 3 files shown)


satya-bodapati:
Can you please check if dict_table_t of temporary table created by ha_innobase::open() is destroyed? May be this issue exists with this I_S code in 5.7 too?

laurynas-biveinis:
> Can you please check if dict_table_t of temporary table created by ha_innobase::open() is destroyed? May be this issue exists with this I_S code in 5.7 too?

Unfortunately, in all active branches, 5.6 to 8.0 with this fix, the above question does not come into relevance because clone in this particular testcase always returns nullptr, because ha_innobase::open fails with table not found, and never creates any new dict_table_t object.

As discussed on Slack, this PR is approved to restore parity with 5.6/5.7, and a follow-up bug https://jira.percona.com/browse/PS-5572 has been created.

PR #3144, opened 2019-04-16
Fixed PS-5353 (Wrong binlog entry for BLOB on a blackhole intermediary master) (5.6)

percona-ysorokin:
https://jira.percona.com/browse/PS-5353

Unlike normal storage engine (e.g. 'InnoDB') in which
'rnd_next()' overload actually reads all data in BLOB fields
from real storage into internal SE memory (like 'mem_heap' in InnoDB)
and updates packed representation('table->record[0]') of the BLOB field
values with pointers to this internal SE memory, 'Blackhole' engine does
not do this.

In case when a database with 'Blackhole' tables serves just as an
intermediate binlog server in a replication chain, this may cause
data corruption.

In particular, when 'Update_rows_log_event' is processed on a
'Blackhole' table, calling 'Update_rows_log_event::do_exec_row()' will
first copy Before Image (BI) found in 'record[0]' into 'record[1]' and
then unpack After Image (AI) into 'record[0]'. The problem is that this
record copying is shallow (just 'memcpy()') and for packed BLOB fields
it just copies pointer values. In other words, before calling
'unpack_current_row()' we end up in a situation when packed BLOB field
values in 'record[0]' (AI) and 'record[1]' (BI) point to exactly the
same memory location and calling 'unpack_current_row()' for AI
overwrites BLOB data in BI.

To prevent this we do the same trick as for virtual generated columns
in 5.7 - keeping old BLOB value inside 'old_value' field in
'Field_blob' class by calling 'keep_old_value()' for all BLOB fields
currently marked for update in 'table->write_set'.

Added 'rpl.bug93917' MTR test case which simulates this scenario for
various column types (all flavors of BLOB, all flavors of TEXT, VARCHAR
and VARBINARY).

Partially backported 'old_value' extension of the 'Field_blob' class
from the
WL #8149 "B-tree Index Support on non-materialized virtual columns"
(commit mysql/mysql-server@c47e175).

### DIFF
--- a/mysql-test/suite/rpl/t/bug93917.test
+++ b/mysql-test/suite/rpl/t/bug93917.test
@@ -1,0 +1,73 @@
+--echo #
+--echo # Bug #93917 "Wrong binlog entry for BLOB on a blackhole intermediary master"
+--echo # (https://bugs.mysql.com/bug.php?id=93917)
+--echo #
+
+--source include/have_innodb.inc
+--source include/have_blackhole.inc
+--source include/have_binlog_format_row.inc
+
+--source include/master-slave.inc
+
+--let $safe_load_data_dir = `SELECT @@global.secure_file_priv`
+--let $server_1_binlog_record_file = $safe_load_data_dir/bug93917_server_1.txt
+--let $server_2_binlog_record_file = $safe_load_data_dir/bug93917_server_2.txt
+
+--let $data_type_list = 'TINYBLOB', 'BLOB', 'MEDIUMBLOB', 'LONGBLOB', 'TINYTEXT', 'TEXT', 'MEDIUMTEXT', 'LONGTEXT', 'VARCHAR(64)', 'VARBINARY(64)'
+--let $list_sentinel = 'ABCDEF'
+--let $number_of_data_types = `SELECT FIND_IN_SET($list_sentinel, CONCAT_WS(',', $data_type_list, $list_sentinel)) - 1`
+--let $data_type_index = 1
+
+--let $server_1_datadir = `SELECT @@datadir`
+--let $server_1_log_file = query_get_value(SHOW MASTER STATUS, File, 1)
+--source include/rpl_connection_slave.inc
+--let $server_2_datadir = `SELECT @@datadir`
+--let $server_2_log_file = query_get_value(SHOW MASTER STATUS, File, 1)
+--source include/rpl_connection_master.inc
+
+while($data_type_index <= $number_of_data_types)
+{
+  --let $data_type = `SELECT ELT($data_type_index, $data_type_list)`
+  --echo ********************************************************************************
+  --echo Testing '$data_type' data type
+  --echo ********************************************************************************
+  --echo
+
+  --eval CREATE TABLE t1(id BIGINT, col $data_type) ENGINE=InnoDB
+
+  --source include/sync_slave_sql_with_master.inc
+  ALTER TABLE t1 ENGINE=Blackhole;
+
+  --source include/rpl_connection_master.inc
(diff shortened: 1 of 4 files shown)


percona-ysorokin:
Jenkins build link:
https://ps56.cd.percona.com/job/percona-server-5.6-param/3/

mysql-test/suite/rpl/t/bug93917.test
```diff
@@ -0,0 +1,73 @@
...
+  --echo
+
+  --eval CREATE TABLE t1(id BIGINT, col $data_type) ENGINE=InnoDB
+
+  --source include/sync_slave_sql_with_master.inc
+  ALTER TABLE t1 ENGINE=Blackhole;
+
+  --source include/rpl_connection_master.inc
+  INSERT INTO t1 VALUES(1, 'testblob_1');
+  FLUSH BINARY LOGS;
```
laurynas-biveinis: Why so many binlog flushes are required and why they are required at all?

storage/blackhole/ha_blackhole.cc
```diff
@@ -156,7 +156,50 @@ int ha_blackhole::rnd_next(uchar *buf)
                        TRUE);
   THD *thd= ha_thd();
   if (pretend_for_slave(*thd))
+  {
+    /*
+      Unlike normal storage engine (e.g. 'InnoDB') in which
```
laurynas-biveinis:
```suggestion
      Unlike a normal storage engine (e.g. 'InnoDB') in which
```

storage/blackhole/ha_blackhole.cc
```diff
@@ -156,7 +156,50 @@ int ha_blackhole::rnd_next(uchar *buf)
...
+      'rnd_next()' overload actually reads all data in BLOB fields
+      from real storage into internal SE memory (like 'mem_heap' in InnoDB)
+      and updates packed representation('table->record[0]') of the BLOB field
```
laurynas-biveinis:
```suggestion
      and updates packed representation ('table->record[0]') of the BLOB field
```

mysql-test/suite/rpl/t/bug93917.test
percona-ysorokin: The idea was to have a separate, currently closed binlog.nnn file which would contain only one `Update_rows` event. This helps to add stability/predictability when we extract event data via
`--exec $MYSQL_BINLOG -vv --base64-output=decode-rows ... | grep ... > ...`

laurynas-biveinis: Maybe mysqlbinlog arguments to select the desired event only (--start-position/--stop-position) would be more robust? The current way looks like it could be broken by GTID events, for example

storage/blackhole/ha_blackhole.cc
percona-ysorokin: fixed

storage/blackhole/ha_blackhole.cc
percona-ysorokin: fixed

mysql-test/suite/rpl/t/bug93917.test
percona-ysorokin: Removed all "FLUSH BINARY LOG". Reworked record extraction - now we use --start-position/--stop-position.

percona-ysorokin:
Updated Jenkins build link: https://ps56.cd.percona.com/job/percona-server-5.6-param/4/ (rpl.bug93917 rpl.bug93917 --repeat=2)

laurynas-biveinis (approved):

PR #3150, opened 2019-04-17
PS-5541: Do not try to decrypt blocks in changed page tracking. (5.7)

dutow:
Issue: changed page tracking uses the LOG flag during read
operations to follow the correct code path. Redo log encryption
backported from 8.0 tries to decrypt pages with a certain bit set,
and fails. While the end result is the same, the error log gets
flooded with decryption errors.

Solution: the NO_ENCRYPTION flag, previously ignored by log
encryption now disables decryption during log reads too.

### DIFF
--- a/mysql-test/t/percona_changed_pages_big-master.opt
+++ b/mysql-test/t/percona_changed_pages_big-master.opt
@@ -1,0 +1,2 @@
+--innodb_track_changed_pages=1 --innodb_max_bitmap_file_size=1G
+
--- a/mysql-test/t/percona_changed_pages_big.test
+++ b/mysql-test/t/percona_changed_pages_big.test
@@ -1,0 +1,45 @@
+-- source include/big_test.inc
+-- source include/have_innodb.inc
+
+--echo #
+--echo # PS-5541: Changed page tracking floods the error log with decryption errors
+--echo #
+
+CREATE TABLE `joinit` (
+  `i` int(11) NOT NULL AUTO_INCREMENT,
+  `s` varchar(64) DEFAULT NULL,
+  `t` time NOT NULL,
+  `g` int(11) NOT NULL,
+  PRIMARY KEY (`i`),
+  KEY key_g (`g`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+INSERT INTO joinit VALUES (NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60));
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60) FROM joinit;
+
+--disable_result_log
+SELECT COUNT(*) AS count_innodb_changed_pages FROM INFORMATION_SCHEMA.INNODB_CHANGED_PAGES;
+--enable_result_log
+
+# Success: if there are no errors in the log
+
+DROP TABLE joinit;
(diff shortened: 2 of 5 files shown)


storage/innobase/log/log0online.cc
```diff
@@ -290,7 +290,7 @@ log_online_read_bitmap_page(
 	     <= bitmap_file->size - MODIFIED_PAGE_BLOCK_SIZE);
 	ut_a(bitmap_file->offset % MODIFIED_PAGE_BLOCK_SIZE == 0);
 
-	IORequest io_request(IORequest::LOG | IORequest::READ);
+	IORequest io_request(IORequest::LOG | IORequest::READ | IORequest::NO_ENCRYPTION);
```
laurynas-biveinis, line 293: This works, but that still keeps the IORequest::LOG in mix, which is not entirely correct. Have you evaluated effort needed to move away from that?

mysql-test/t/percona_bug_ps5541.test
```diff
@@ -0,0 +1,47 @@
+-- source include/big_test.inc
```
laurynas-biveinis: If I understand correctly, the test does *not* require redo log encryption itself to be enabled, correct? In which case this scenario should be test by existing innodb.percona_changed* testcases, please check, whether the new testcase is required at all

storage/innobase/log/log0online.cc
dutow: I did - and that would require changing multiple conditions in os0file, and adding a new enum value. The `NO_ENCRYPTION` tag is already used in a similar case for keyring encryption, so it seemed a better choice.

mysql-test/t/percona_bug_ps5541.test
dutow: That's right, encryption is not needed - but a huge amount of data is. If I insert fewer records than what we have in this test case, the issue won't show up without this commit. And with this amount, it runs for 15-20 minutes on a debug build.

laurynas-biveinis: OK, then the existing bitmap testcases have predetermined name prefix (percona_changed_page_bmp in general, percona_changes_pages for I_S queries like this one), so please rename to percona_changed_pages_big and add a header comment --echo # / --echo # PS-5541: blah blah / --echo #

mysql-test/t/percona_bug_ps5541.test
```diff
@@ -0,0 +1,47 @@
...
+
+--disable_result_log
+show status like 'Innodb_page%';
+--enable_result_log
+
+USE test;
```
laurynas-biveinis: Redundant

mysql-test/t/percona_bug_ps5541.test
```diff
@@ -0,0 +1,47 @@
+-- source include/big_test.inc
+
+--disable_result_log
+show status like 'Innodb_page%';
```
laurynas-biveinis: Why?

mysql-test/t/percona_bug_ps5541.test
```diff
@@ -0,0 +1,47 @@
+-- source include/big_test.inc
+
```
laurynas-biveinis: --source include/have_innodb.inc

mysql-test/t/percona_bug_ps5541.test
```diff
@@ -0,0 +1,47 @@
...
+USE test;
+CREATE TABLE `joinit` (
+  `i` int(11) NOT NULL AUTO_INCREMENT,
+  `s` varchar(64) DEFAULT NULL,
+  `t` time NOT NULL,
+  `g` int(11) NOT NULL,
+  PRIMARY KEY (`i`),
+  KEY key_g (`g`)
+) ENGINE=InnoDB  DEFAULT CHARSET=latin1;
```
laurynas-biveinis:
```suggestion
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```

mysql-test/t/percona_bug_ps5541.test
```diff
@@ -0,0 +1,47 @@
...
+
+INSERT INTO joinit VALUES (NULL, uuid(), time(now()),  (FLOOR( 1 + RAND( ) *60 )));
```
laurynas-biveinis: Everywhere
```suggestion
INSERT INTO joinit VALUES (NULL, uuid(), time(now()), (FLOOR(1 + RAND() * 60)));
```

mysql-test/t/percona_bug_ps5541.test
```diff
@@ -0,0 +1,47 @@
...
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  (FLOOR( 1 + RAND( ) *60 )) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  (FLOOR( 1 + RAND( ) *60 )) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  (FLOOR( 1 + RAND( ) *60 )) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  (FLOOR( 1 + RAND( ) *60 )) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  (FLOOR( 1 + RAND( ) *60 )) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  (FLOOR( 1 + RAND( ) *60 )) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  (FLOOR( 1 + RAND( ) *60 )) FROM joinit;
+
+--disable_result_log
+show status like 'Innodb_page%';
```
laurynas-biveinis: Why?

laurynas-biveinis (changes requested):

mysql-test/t/percona_changed_pages_big-master.opt
```diff
@@ -0,0 +1,2 @@
+--innodb_track_changed_pages=1 --innodb_max_bitmap_file_size=1G --innodb_buffer_pool_size=1G
```
laurynas-biveinis: Is 1G buffer pool size necessary?

mysql-test/t/percona_changed_pages_big.test
```diff
@@ -0,0 +1,45 @@
...
+CREATE TABLE `joinit` (
+  `i` int(11) NOT NULL AUTO_INCREMENT,
+  `s` varchar(64) DEFAULT NULL,
+  `t` time NOT NULL,
+  `g` int(11) NOT NULL,
+  PRIMARY KEY (`i`),
+  KEY key_g (`g`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+INSERT INTO joinit VALUES (NULL, uuid(), time(now()),  FLOOR(1 + RAND() *60 ));
```
laurynas-biveinis: Comment not addressed

mysql-test/t/percona_changed_pages_big.test
```diff
@@ -0,0 +1,45 @@
...
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  FLOOR(1 + RAND() *60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  FLOOR(1 + RAND() *60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  FLOOR(1 + RAND() *60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  FLOOR(1 + RAND() *60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  FLOOR(1 + RAND() *60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  FLOOR(1 + RAND() *60) FROM joinit;
+INSERT INTO joinit SELECT NULL, uuid(), time(now()),  FLOOR(1 + RAND() *60) FROM joinit;
+
+--disable_result_log
+select count(*) as count_innodb_changed_pages from information_schema.INNODB_CHANGED_PAGES;
```
laurynas-biveinis: Make SQL case consistent

laurynas-biveinis (changes requested):

mysql-test/t/percona_changed_pages_big.test
```diff
@@ -0,0 +1,45 @@
...
+CREATE TABLE `joinit` (
+  `i` int(11) NOT NULL AUTO_INCREMENT,
+  `s` varchar(64) DEFAULT NULL,
+  `t` time NOT NULL,
+  `g` int(11) NOT NULL,
+  PRIMARY KEY (`i`),
+  KEY key_g (`g`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+INSERT INTO joinit VALUES (NULL, uuid(), time(now()),  FLOOR(1 + RAND() * 60));
```
laurynas-biveinis: Strictly speaking
```suggestion
INSERT INTO joinit VALUES (NULL, uuid(), time(now()), FLOOR(1 + RAND() * 60));
```
still not done but of course not enough to cause a PR respin by itself

laurynas-biveinis (approved):

PR #3151, opened 2019-04-18
PS-5541: Do not try to decrypt blocks in changed page tracking. (8.0)

dutow:
Issue: changed page tracking uses the LOG flag during read
operations to follow the correct code path. Redo log encryption
backported from 8.0 tries to decrypt pages with a certain bit set,
and fails. While the end result is the same, the error log gets
flooded with decryption errors.

Solution: the NO_ENCRYPTION flag, previously ignored by log
encryption now disables decryption during log reads too.

(cherry picked from commit 48b67f85cba1c54d02e129ac862096c786324a5b)

### DIFF
--- a/mysql-test/r/percona_changed_pages_big.result
+++ b/mysql-test/r/percona_changed_pages_big.result
@@ -1,0 +1,34 @@
+#
+# PS-5541: Changed page tracking floods the error log with decryption errors
+#
+CREATE TABLE `joinit` (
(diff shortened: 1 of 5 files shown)


laurynas-biveinis (changes requested):
Same comments as for 5.7

laurynas-biveinis (approved):

PR #3152, opened 2019-04-19
PS-5550 : RHEL8 runners in 5.7 seem to be missing python

george-lorch:
- Changed TokuDB test generator makefiles to invoke python2 explicitly rather
  than generic python.
- Changed TokuDB *.py scripts to she-bang with python2 instead of python.

### DIFF
--- a/mysql-test/suite/innodb_stress/include/innodb_stress.inc
+++ b/mysql-test/suite/innodb_stress/include/innodb_stress.inc
@@ -41,7 +41,7 @@ if ($do_checksum)
 {
     # populate the table and store its checksum before any load.
     let $exec =
-python $MYSQL_BASEDIR/mysql-test/suite/innodb_stress/t/load_generator.py $pid_file $kill_db_after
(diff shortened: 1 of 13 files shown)


george-lorch:
No jenkins as these changes are not run as part of any mtr suite, only to python scripts that are used to generate mtr tests.

laurynas-biveinis (changes requested):
Please take a look at innodb_stress suite too, it's even more acute (here we have test generators whereas there it's tests themselves)

george-lorch:
> Please take a look at innodb_stress suite too, it's even more acute (here we have test generators whereas there it's tests themselves)

Ack, yes, we discussed this and it slipped my mind while backporting to 5.6 and re-creating work based on GCA.  Will get these touched up and jenkins-ized if possible.

george-lorch:
OK, got the innodb_stress suite changes and a jenkins of just the innodb_stress suite run https://ps56.cd.percona.com/job/percona-server-5.6-param/7/ with what appear to be spurious failures unrelated to this change.

laurynas-biveinis (approved):

PR #3153, opened 2019-04-19
Ps 5.7 5550

george-lorch:


george-lorch:
Conflicts are due to incorrect changes made to some .py scripts after the GCA point that changed the she-bang to python3 without actually fixing the scripts to be python3 compatible.  On Merge these will be changed back to python2.

Here is what it would look like after merge https://github.com/percona/percona-server/compare/5.7...georgelorchpercona:ps-5.7-5550-post-merge and here is a jenkins run of the post-merge branch https://ps57.cd.percona.com/job/percona-server-5.7-param/20/

laurynas-biveinis (changes requested):
Please take a look at innodb_stress suite too, it's even more acute (here we have test generators whereas there it's tests themselves)

george-lorch:
OK, got innodb_stress and remerged up from 5.6.  New jenkins on a post-merge branch here https://ps57.cd.percona.com/job/percona-server-5.7-param/23/

laurynas-biveinis (approved):

PR #3154, opened 2019-04-19
PS-5360 : Implement server side changes for XtraBackup MyRocks

george-lorch:
PS-5362 : Create new session variable rocksdb_create_temporary_checkpoint
PS-5362 : Expose RocksDB WAL position through I_S.log_status
PS-5563 : Create new session variable rocksdb_disable_file_deletions
PS-5583 : Lock tables for backup blocks DML for RocksDB

- Temporarily forked RocksDB into percona/rocksdb and created temp branch
  (ps-5362) pointing to same commit that 8.0 currently points to
  (516c7df05dd267e4725e7dd061a6767de7aayyac0d).

- Cherry-picked Sergeis work onto new RocksDB branch (ps-5362) in new temp fork.

- Repointed Percona Server RocksDB git submodule commit pointer to point to new
  fork (percona/rocksdb) at same commit that was previously pointed to
  (516c7df05dd267e4725e7dd061a6767de7aayyac0d) +1 commit to Sergeis
  cherry-picked work (3f5c1236026233b34be64d597fe660d7e698f2f5).

- Added new session variable to MyRocks rocksdb_create_temporary_checkpoint
  which will create a temporary RocksDB 'checkpoint' or 'snapshot'.  If the
  session ends with an existingcheckpoint, or if the variable is reset to
  another value, the checkpoint will get removed.

- Added new session variable to MyRocks rocksdb_disable_file_deletions, which
  when set, will tell RocksDB to disable the deletion of old WAL and .sst files.
  If the session ends without re-enabling file deletion, file deletions will be
  explicitly re-enabled.

- Expose rocksdb WAL position in i_s.log_status table.

- Added tests to MyRocks for new options and behaviors.
    
- Added HTON_SUPPORTS_ONLINE_BACKUPS to MyRocks handlerton flags.
    
- Split main.backup_locks and rpl.rpl_backup_locks tests into common includes
  with engine and feature controlling options and updating to use modern
  semantics.  Implemented InnoDB and MyRocks tests based on these new common
  tests and recorded.
    
- Updated documentation to reflect MyRocks support/inclusion in LTFB.


george-lorch:
https://ps80.cd.percona.com/job/percona-server-8.0-param/90/

When reviewing, please verify the proper submodule repo (percona/rocksdb) and that the commit is exactly the same as the previous facebook/rocksdb pointer + 1 commit containing Sergeis work.  Once merged, we will create a new 8.0.15-6 release branch based off the 8.0.15-5 tag and cherry-pick this over.

mysql-test/suite/rocksdb/t/rocksdb_disable_file_deletions.test
```diff
@@ -0,0 +1,55 @@
...
+# make sure the WAL has not been removed
+file_exists $ROCKSDB_DATADIR/$wal;
+
+connection con1;
+disconnect con1;
+--source include/wait_until_disconnected.inc
+# deletions should be enabled now
+connection default;
+
+# write little more data data
```
gl-sergei: data data

gl-sergei:
Github actually showing nice diff for changes in submodules, it looks ok. Travis reporting that formatting changes are needed

mysql-test/suite/rocksdb.sys_vars/t/rocksdb_create_temporary_checkpoint_basic.test
```diff
@@ -0,0 +1,57 @@
+--source include/have_rocksdb.inc
+
+SET @orig_value = @@session.rocksdb_create_temporary_checkpoint;
```
laurynas-biveinis: No need to restore session variable to original value

mysql-test/suite/rocksdb.sys_vars/t/rocksdb_create_temporary_checkpoint_basic.test
```diff
@@ -0,0 +1,57 @@
...
+disconnect con1;
+--source include/wait_until_disconnected.inc
+
+connection default;
+
+# error, checkpoint removed upon disconnect
+--error 1
+file_exists $MYSQL_TMP_DIR/ghj/CURRENT;
+
+--eval SET SESSION ROCKSDB_CREATE_TEMPORARY_CHECKPOINT = @orig_value
```
laurynas-biveinis: Remove or set to default

mysql-test/suite/rocksdb.sys_vars/t/rocksdb_disable_file_deletions_basic.test
```diff
@@ -0,0 +1,18 @@
+--source include/have_rocksdb.inc
+
+SET @orig_value = @@session.rocksdb_disable_file_deletions;
```
laurynas-biveinis: No need to restore session variable to original value

mysql-test/suite/rocksdb.sys_vars/t/rocksdb_disable_file_deletions_basic.test
```diff
@@ -0,0 +1,18 @@
...
+SET SESSION rocksdb_disable_file_deletions = ON;
+SET SESSION rocksdb_disable_file_deletions = OFF;
+SET SESSION rocksdb_disable_file_deletions = FALSE;
+
+--error ER_WRONG_VALUE_FOR_VAR
+SET SESSION rocksdb_disable_file_deletions = "abc";
+--error ER_WRONG_VALUE_FOR_VAR
+SET SESSION rocksdb_disable_file_deletions = 12;
+
+SET SESSION ROCKSDB_DISABLE_FILE_DELETIONS = @orig_value;
```
laurynas-biveinis: Remove or set to default

mysql-test/suite/rocksdb/t/rocksdb_disable_file_deletions.test
```diff
@@ -0,0 +1,55 @@
...
+    inc $i;
+}
+enable_query_log;
+
+# make sure the WAL has not been removed
+file_exists $ROCKSDB_DATADIR/$wal;
+
+connection con1;
+disconnect con1;
+--source include/wait_until_disconnected.inc
```
laurynas-biveinis: count_sessions.inc / wait_until_count_sessions.inc

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -204,51 +205,79 @@ static std::string rdb_normalize_dir(std::string dir) {
...
 
-static int
-rocksdb_create_checkpoint(THD *const thd MY_ATTRIBUTE((__unused__)),
-                          struct SYS_VAR *const var MY_ATTRIBUTE((__unused__)),
-                          void *const save MY_ATTRIBUTE((__unused__)),
-                          struct st_mysql_value *const value) {
+static int rocksdb_create_checkpoint(const char *checkpoint_dir_raw) {
+  DBUG_ASSERT(checkpoint_dir_raw);
+
+  std::string checkpoint_dir = rdb_normalize_dir(checkpoint_dir_raw);
```
laurynas-biveinis: const (const auto while at that)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -204,51 +205,79 @@ static std::string rdb_normalize_dir(std::string dir) {
...
+  LogPluginErrMsg(INFORMATION_LEVEL, 0,
+                  "creating checkpoint in directory : %s\n",
```
laurynas-biveinis:
```suggestion
                  "creating checkpoint in directory: %s\n",
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -204,51 +205,79 @@ static std::string rdb_normalize_dir(std::string dir) {
...
+                  checkpoint_dir.c_str());
+  rocksdb::Checkpoint *checkpoint;
+  auto status = rocksdb::Checkpoint::Create(rdb, &checkpoint);
+  // We can only return HA_EXIT_FAILURE/HA_EXIT_SUCCESS here which is why
+  // the return code is ignored, but by calling into rdb_error_to_mysql,
+  // it will call my_error for us, which will propogate up to the client.
+  int rc MY_ATTRIBUTE((__unused__));
```
laurynas-biveinis: IMHO cast to void would be more readable

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -204,51 +205,79 @@ static std::string rdb_normalize_dir(std::string dir) {
...
+    }
+  } else {
+    rc = ha_rocksdb::rdb_error_to_mysql(status);
+  }
+
+  return HA_EXIT_FAILURE;
+}
+
+static int rocksdb_remove_checkpoint(const char *checkpoint_dir_raw) {
+  std::string checkpoint_dir = rdb_normalize_dir(checkpoint_dir_raw);
```
laurynas-biveinis: const (const auto while at that)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -204,51 +205,79 @@ static std::string rdb_normalize_dir(std::string dir) {
...
+  LogPluginErrMsg(INFORMATION_LEVEL, 0,
+                  "deleting temporary checkpoint in directory : %s\n",
+                  checkpoint_dir.c_str());
+  auto status = rocksdb::DestroyDB(checkpoint_dir, rocksdb::Options());
```
laurynas-biveinis:
```suggestion
  const auto status = rocksdb::DestroyDB(checkpoint_dir, rocksdb::Options());
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -204,51 +205,79 @@ static std::string rdb_normalize_dir(std::string dir) {
...
+  int rc MY_ATTRIBUTE((__unused__));
```
laurynas-biveinis: IMHO cast to void would be more readable

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3009,9 +3052,40 @@ void Rdb_snapshot_notifier::SnapshotCreated(
...
 mysql_mutex_t Rdb_transaction::s_tx_list_mutex;
 
-static Rdb_transaction *&get_tx_from_thd(THD *const thd) {
-  return *reinterpret_cast<Rdb_transaction **>(
+struct rocksdb_ha_data {
+  char *checkpoint_dir;
+  Rdb_transaction *trx;
+  bool disable_file_deletions;
+  rocksdb_ha_data() : checkpoint_dir(), trx(), disable_file_deletions() {}
+  const char *get_checkpoint_dir() const { return checkpoint_dir; }
```
laurynas-biveinis, line 3058: Redunndant: either remove the getter either make checkpoint_dir private.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3090,9 +3164,73 @@ static int rocksdb_close_connection(handlerton *const hton, THD *const thd) {
...
+  DBUG_ASSERT(thd != nullptr);
+
+  char buf[FN_REFLEN];
+  int len = sizeof(buf);
+
+  const char *current_checkpoint_dir = get_ha_data(thd)->get_checkpoint_dir();
+  const char *new_checkpoint_dir = value->val_str(value, buf, &len);
+
+  if (current_checkpoint_dir != nullptr && new_checkpoint_dir != nullptr) {
+    std::string errmsg = "Checkpoint already exists in directory '" +
```
laurynas-biveinis: Local var which is never used?..

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3090,9 +3164,73 @@ static int rocksdb_close_connection(handlerton *const hton, THD *const thd) {
...
+                         std::string(current_checkpoint_dir) + "'";
+    *reinterpret_cast<const char **>(save) = current_checkpoint_dir;
+    return HA_EXIT_FAILURE;
+  } else if (new_checkpoint_dir != nullptr) {
+    auto res = rocksdb_create_checkpoint(new_checkpoint_dir);
```
laurynas-biveinis:
```suggestion
    const auto res = rocksdb_create_checkpoint(new_checkpoint_dir);
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3090,9 +3164,73 @@ static int rocksdb_close_connection(handlerton *const hton, THD *const thd) {
...
+    if (res == HA_EXIT_SUCCESS) {
+      *reinterpret_cast<const char **>(save) = new_checkpoint_dir;
+      get_ha_data(thd)->set_checkpoint_dir(new_checkpoint_dir);
+    } else {
+      return res;
+    }
+  } else if (current_checkpoint_dir != nullptr) {
+    auto res = rocksdb_remove_checkpoint(current_checkpoint_dir);
```
laurynas-biveinis:
```suggestion
    const auto res = rocksdb_remove_checkpoint(current_checkpoint_dir);
```

laurynas-biveinis (changes requested):

storage/rocksdb/ha_rocksdb.cc
gl-sergei: I lost the following line somewhere between iterations:
```
    my_error(ER_GET_ERRMSG, MYF(0), HA_ERR_ROCKSDB_STATUS_INVALID_ARGUMENT,
             errmsg.c_str(), rocksdb_hton_name);
```

george-lorch:
OK, made all of the suggested changes and went a little further to C++-ize the Rdb_ha_info and make it a little more MyRocks-ish.  Probably could have refactored some more but this I think is good enough for the purposes here.  New Jenkins https://ps80.cd.percona.com/job/percona-server-8.0-param/97/

mysql-test/suite/rocksdb/t/rocksdb_disable_file_deletions.test
laurynas-biveinis: What I meant was to replace this wait_until_disconnected.inc with pulled-up wait_until_count_sessions.inc, which will both sync here and no second wait_until_count_sessions.inc will be needed at the end (wait_until_disconnected.inc by itself is not a sufficient replacement for wait_until_count_sessions.inc)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -204,51 +205,81 @@ static std::string rdb_normalize_dir(std::string dir) {
...
+  LogPluginErrMsg(INFORMATION_LEVEL, 0,
+                  "creating checkpoint in directory: %s\n",
+                  checkpoint_dir.c_str());
+  rocksdb::Checkpoint *checkpoint;
+  auto status = rocksdb::Checkpoint::Create(rdb, &checkpoint);
+  // We can only return HA_EXIT_FAILURE/HA_EXIT_SUCCESS here which is why
+  // the return code is ignored, but by calling into rdb_error_to_mysql,
+  // it will call my_error for us, which will propogate up to the client.
+  int rc;
+  (void)rc;
```
laurynas-biveinis: no no no :)
My comment was to remove rc variable altogether and cast to void its assignments below

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -204,51 +205,81 @@ static std::string rdb_normalize_dir(std::string dir) {
...
+}
+
+static int rocksdb_remove_checkpoint(const char *checkpoint_dir_raw) {
+  const auto checkpoint_dir = rdb_normalize_dir(checkpoint_dir_raw);
+  LogPluginErrMsg(INFORMATION_LEVEL, 0,
+                  "deleting temporary checkpoint in directory : %s\n",
+                  checkpoint_dir.c_str());
+  const auto status = rocksdb::DestroyDB(checkpoint_dir, rocksdb::Options());
+  int rc;
+  (void)rc;
```

laurynas-biveinis (changes requested):

mysql-test/suite/rocksdb/t/rocksdb_disable_file_deletions.test
george-lorch: Ahh yeah, good point, in this context, wait_until_count_sessions will satisfy the same need as wait_until_disconnected as well as ensure that all secondary connections are closed before ending the test.

storage/rocksdb/ha_rocksdb.cc
george-lorch: `(void)some_function(asdf);` will still raise the `error: ignoring return value of` error due to the prototype explicitly containing `MY_ATTRIBUTE((__warn_unused_result__))` which overrides the void cast.  See interesting discussion here https://gcc.gnu.org/bugzilla/show_bug.cgi?id=25509

It could be argued in this case that the `MY_ATTRIBUTE((__warn_unused_result__))` is not necessarily correct for this method or, probably more correct, is that we shouldn't be using this method here and instead just directly calling `my_error(ER_GET_ERRMSG, MYF(0), status.code(), status.ToString().c_str(), rocksdb_hton_name);` which is all that `ha_rocksdb::rdb_error_to_mysql` is effectively doing here.

george-lorch:
OK, pushed again https://ps80.cd.percona.com/job/percona-server-8.0-param/103/
- Removed call to rdb_error_to_mysql and replaced with direct calls to my_error
- Fixed wait_until_disconnected/wait_until_count_sessions
- Added HTON_SUPPORTS_ONLINE_BACKUPS to MyRocks hton::flags for `LOCK TABLES FOR BACKUP`
- Added Myrocks to main.backup_locks test.  This may not be where/how we want to do this but I did it here for now anyway to see how this functionality tests out in MyRocks.  If desirable, we can either clone that test into the rocksdb suite with the added MyRocks specifics or somehow refactor the whole test into a .inc with a $engine specifier or something.  That may be a little difficult because not all engines support all statements/features that the test is executing.  Note new jenkins run includes "main" suite to capture main.backup_locks.

gl-sergei:
Will we also fix LTFB for tokudb? for 5.7?

george-lorch:
> Will we also fix LTFB for tokudb? for 5.7?

Maybe, but I am not sure where I see any real value there, perhaps for mysqldump?  TokuDB is deprecated in 8.0 so we will not be pursuing a backup solution for it going forward and the MyRocks backup is 8.0 only, so what value exists for backport of this to 5.7?  If Laurynas and Tyler think there is any value here, this can easily be backported to 5.7 for MyRocks and/or cross and backported for TokuDB from 5.6 onward via GCA.  For now, we need to get this landed in the best way possible to meet the release time frame demands.

gl-sergei:
> Maybe, but I am not sure where I see any real value there, perhaps for mysqldump?

I am not sure if there are users who using both InnoDB and TokuDB and using xtrabackup
for InnoDB and TokuBackup for TokuDB. Doesn't sound like sane idea, but I will not be
surprised if there are some.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: Oh right I forgot about this. Then the original code looks the best to me :/

mysql-test/t/backup_locks.test
```diff
@@ -5,6 +5,7 @@
 --source include/have_myisam.inc
 --source include/have_archive.inc
 --source include/have_blackhole.inc
+--source include/have_rocksdb.inc
```
laurynas-biveinis: I don't agree with this change. Lots of folks build server without MyRocks, and now they would get test disabled. Also, this is not the only one relevant backup locks test, IIRC there was rpl_backup_locks too at least, needs grepping. Please clone into a MyRocks suite with a TODO to work out the currently non-obvious details and convert it to .inc scheme

laurynas-biveinis (changes requested):

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3863,6 +4020,82 @@ static bool rocksdb_show_status(handlerton *const hton, THD *const thd,
...
+}
+
+/*
+  Implements Log_resource collect_info.
+
+  Produces JSON object with following structure:
+
+  "RocksDB": {
+    "wal_files": [
+      { "log_number": N, "path_name": "...", "file_size_bytes": K },
```
gl-sergei, line 4050: do you guys think that {"number": N, "path", "...", "size": K} would sound better? I just copied RocksDB method names here, but we don't really need to follow their naming...

george-lorch: I think this is fine as-is.

george-lorch:
OK @laurynas-biveinis I did a little refactoring of the backup_locks and rpl_backup_locks tests to move their content to include and extra/rpl_tests with some control options and pull them in from the existing tests and from new rocksdb tests.  I also did a little wee bit of straightening out of the orders of the series of mini-test sets within backup_locks so that they reflect the same ordering from table creation all the way  through.  Here is a jenkins run of just these changes to the PR https://ps80.cd.percona.com/job/percona-server-8.0-param/113/

mysql-test/extra/rpl_tests/rpl_backup_locks.inc
```diff
@@ -0,0 +1,127 @@
+########################################################################
+# Replication-related tests for backup locks
+#
+# Variables to control teste execution:
```
laurynas-biveinis:
```suggestion
# Variables to control test execution:
```

laurynas-biveinis (approved):
One comment typo, no need for new test cycle, just fix & push
I refrained from commenting on the moved test content

gl-sergei (approved):

george-lorch:
OK, as discussed offline, I went ahead and 'modernized' the rpl_backup_lock test and re-recorded.  Jenkins here https://ps80.cd.percona.com/job/percona-server-8.0-param/117/

I also did a minor update to the doc that states how MyRocks plays with LTFB.

laurynas-biveinis (approved):

PR #3156, opened 2019-04-24
Fix PS-5578 (Test innodb.percona_log_slow_innodb is unstable)

laurynas-biveinis:
While the actual test failures are observed on 8.0, clean up the test
in lower branches as well: for table and row lock waits, only slow-log
the actual statements that wait for the locks, removing test
coverage (and potential instability issues) from other commands.

https://ps56.cd.percona.com/job/percona-server-5.6-param/5/


percona-ysorokin (approved):
LGTM

PR #3160, opened 2019-04-26
PS-5574: innodb.log_corruption / innodb.log_file_size_checkpoint

robgolebiowski:
regression on trunk

The innodb.log_corruption error:
dict_detect_encryption_of_mysql_ibd calls os_file_read_first_page. This
function causes server to abort when the first page cannot be read.
Introduced new function os_file_read_first_page_noexit, which does not
abort the server when first page cannot be read.

The innodb.log_file_size_checkpoint error:
The redo was checked for any log records that apply to the first page of
mysql.ibd. This was done as part of PS-5333. In PS-5333 we did:
- set mysql's DD encryption flag to the value read from its page0
- execute redo
- correct mysql's DD encryption flag if there were changes to mysql's
page0 during redo
It turns out that the last part is not needed. System tablespace
encryption is a bootstrap only option for MK encryption and for keyring
encryption the encryption flag will be corrected by encryption threads.

### DIFF
--- a/storage/innobase/dict/dict0dict.cc
+++ b/storage/innobase/dict/dict0dict.cc
@@ -7528,13 +7528,13 @@ static std::tuple<bool, bool> get_mysql_ibd_page_0_io() {
   /* page0 of mysql.ibd is not in the buffer, try direct io */
   auto buf = ut_make_unique_ptr_nokey(2 * UNIV_PAGE_SIZE);
 
-  bool successful_read = false;
+  bool successfully_opened = false;
 
   pfs_os_file_t file = os_file_create_simple_no_error_handling(
       innodb_data_file_key, dict_sys_t::s_dd_space_file_name, OS_FILE_OPEN,
-      OS_FILE_READ_ONLY, srv_read_only_mode, &successful_read);
+      OS_FILE_READ_ONLY, srv_read_only_mode, &successfully_opened);
 
-  if (!successful_read) {
+  if (!successfully_opened) {
     return (result);
   }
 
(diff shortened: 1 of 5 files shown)


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/101/

robgolebiowski:
The code for redo was removed. It was a better approach this way imho, as it would give us correct encryption flag even if the encryption threads were turned off. However, after seeing how many problems this is causing, I do not think it is worth it. If you disagree I can look into fixing this otherwise we can merge this removal.

laurynas-biveinis (approved):

satya-bodapati:
Ok, ideally "mysql.ibd" should always be present. If it is missing, server wouldn't start anyway. I think the zip files tries only apply redo without any datafiles. So the fix is mostly about silencing/satisfying the testcase. Right?

Regarding "innodb.log_file_size_checkpoint ", I agree that it is redundant but what exactly caused the issue/failure?

robgolebiowski:
@satya-bodapati yes, the mysql.ibd created in that was just empty. However, if we stayed with os_file_read_first_page causing the server to abort we would get error that mysql.ibd file read  failed instead of nice MySQL DD initalization failed with all surrounding information. 

The thd->dd_client()->acquire(space_name, &dd_space) called in dd_get_tablespace_flags would sometimes returned dd_space == nullptr without any error. Seems that mysql is not yet in the DD cache here.

satya-bodapati:
Ok, I debugged to find out why DD API failed. This code can never work. It is because DD initialization is not complete yet. The DD APIs didn't even call InnoDB to fetch the object. So the code to update DD is in wrong place.

You are lucky that you need not find a new place for this :). Just one small request.

Please mention in commit message that: the test failed because dd_get_tablespace_flags() tried to access DD before it is fully initialized.

satya-bodapati (approved):
Please see my request in comment. Ok to push with that.

PR #3162, opened 2019-04-26
PS-5562 : index_merge + partitioning + MyRocks crashes the server

satya-bodapati:
Problem:
--------
With native paritioning and index_merge, the native partition handler
is cloned.

One important point missed when fixing PS-5206 is that both the original
and clone share the parition_info object (m_part_info). It has bitmap of
partitions read and used.

The cloned handler, which has reference to the shared to partition_info
object, resets the partition bitmap during creation. This causes problem
during the actual index scan.

Fix:
----
1. Create context to identify if the native partition handler creation is
from clone or brand new.

2. When it is clone, do not reset the partition bitmaps (i.e avoid calling
m_part_info->set_partition_bitmaps()). This is the core fix.

3. Additionally, we fix the cloned object creation to use clone mem_root
instead of TABLE*->mem_root. Optimizer can destroy the cloned handler
immediately after its use. (See QUICK_RANGE_SELECT::~QUICK_RANGE_SELECT).


satya-bodapati:
8.0 version will have testcase because the optimizer hint to force /*+ INDEX_MERGE */ is introduced only in 8.0. This is not present in 5.7.

To reproduce the crash,  we need native partitioning + index_merge + *only a subset of partitions* should be used for query. Tried for two days before giving up.

8.0 version has simple testcase.

sql/partitioning/partition_base.cc
```diff
@@ -392,12 +392,15 @@ Partition_base::Partition_base(handlerton *hton, TABLE_SHARE *share)
...
 Partition_base::Partition_base(handlerton *hton, TABLE_SHARE *share,
-    Handler_share** ha_share)
+    Partition_base* clone_base,
+    MEM_ROOT* clone_mem_root)
     : handler(hton, share), Partition_helper(this)
 {
   DBUG_ENTER("Partition_base::Partition_base(table, ha_share)");
   init_handler_variables();
-  set_ha_share_ref(ha_share);
+  m_clone_base = clone_base;
```
laurynas-biveinis: Use constructor initializer list

sql/partitioning/partition_base.cc
```diff
@@ -1966,15 +1979,30 @@ int Partition_base::open(const char *name, int mode, uint test_if_locked)
...
+  }
 
   if (!foreach_partition([&](const partition_element *parent_part_elem,
                              const partition_element *part_elem) -> bool {
         char name_buff[FN_REFLEN];
         part_name(name_buff, name, parent_part_elem, part_elem);
+
+	if (m_clone_base != nullptr) {
+	  uint ref_length = (*clone_base_file)->ref_length;
+          (*file)->ref= (uchar *)alloc_root(m_clone_mem_root,
```
laurynas-biveinis, line 1991: Formatting

sql/partitioning/partition_base.h
```diff
@@ -111,6 +111,14 @@ class Partition_base : public handler,
   /** keep track of partitions to call ha_reset */
   MY_BITMAP m_partitions_to_reset;
 
+  /** Used when Partiton_base is cloned. Specifically it is used to retreive
```
laurynas-biveinis:
```suggestion
  /** Used when Partiton_base is cloned. Specifically it is used to retrieve
```

sql/partitioning/partition_base.h
```diff
@@ -111,6 +111,14 @@ class Partition_base : public handler,
...
+  each handler ref_length */
+  Partition_base* m_clone_base;
+  /** For clone operation, memory will be allocated from this mem_root instead
+  of TABLE*->mem_root. The clone can be destroyed by optimizer immediately. So
```
laurynas-biveinis:
```suggestion
  of TABLE->mem_root. The clone can be destroyed by optimizer immediately. So
```

laurynas-biveinis (changes requested):

sql/partitioning/partition_base.cc
satya-bodapati: well, init_handler_variables() would reset the value.

Ideally, init_handler_variables() should be removed and all variables should be moved to constructor initalizer list.

laurynas-biveinis (approved):

satya-bodapati:
```
./mtr --mem --suite=tokudb.parts --parallel=4 --force --retry-failure=0|tee mtr.log
 93%] tokudb.parts.upgrade_from_prev_ver 'native-partitioning' w4 [ pass ]  13338
[ 94%] tokudb.parts.partition_float_tokudb 'native-partitioning' w4 [ pass ]   1185
[ 95%] tokudb.parts.partition_syntax_tokudb 'native-partitioning' w4 [ pass ]    165
[ 96%] tokudb.parts.partition_value_tokudb 'native-partitioning' w4 [ skipped ]  # CAST() in partitioning function is currently not supported.
[ 97%] tokudb.parts.partition_exchange_tokudb 'non-native-partitioning' w1 [ pass ]  40126
[ 98%] tokudb.parts.partition_exchange_tokudb 'native-partitioning' w2 [ pass ]  40175
[100%] tokudb.parts.partition_special_tokudb 'native-partitioning' w3 [ pass ]  27387
The servers were restarted 12 times
Spent 181.175 of 80 seconds executing testcases

Completed: All 55 tests were successful.

36 tests were skipped, 25 by the test itself.


===============================
./mtr --mem --suite=rocksdb --parallel=4 --force --retry-failure=0|tee mtr.log


[100%] rocksdb.autoinc_crash_safe_partition 'write_committed' w2 [ pass ]  80331
The servers were restarted 119 times
Spent 3606.150 of 1161 seconds executing testcases

Completed: Failed 5/495 tests, 98.99% were successful.

Failing test(s): rocksdb.multi_con_read rocksdb.persistent_cache

```

satya-bodapati:
multi_con_read is result content mismatch. I hve recorded this as part of this commit.

PR #3164, opened 2019-04-29
(8.0)PS-5562 : index_merge + partitioning + MyRocks crashes the server

satya-bodapati:
Problem:
--------
With native paritioning and index_merge, the native partition handler
is cloned.

One important point missed when fixing PS-5206 is that both the original
and clone share the parition_info object (m_part_info). It has bitmap of
partitions read and used.

The cloned handler, which has reference to the shared to partition_info
object, resets the partition bitmap during creation. This causes problem
during the actual index scan.

Fix:
----
1. Create context to identify if the native partition handler creation is
from clone or brand new.

2. When it is clone, do not reset the partition bitmaps (i.e avoid calling
m_part_info->set_partition_bitmaps()). This is the core fix.

3. Additionally, we fix the cloned object creation to use clone mem_root
instead of TABLE->mem_root. Optimizer can destroy the cloned handler
immediately after its use. (See QUICK_RANGE_SELECT::~QUICK_RANGE_SELECT).
So the memory used by cloned handler objects is freed immediately at end
of query execution.
(cherry picked from commit 744f4e1c44d8debcdebd5a83f127cbbc28c261e4)

### DIFF
--- a/mysql-test/suite/rocksdb/r/partition_handler_clone.result
+++ b/mysql-test/suite/rocksdb/r/partition_handler_clone.result
@@ -7,3 +7,35 @@ INSERT INTO t1  VALUES (5,'Bozo', NULL),(6,'Ronald', NULL),(7,'Kinko', NULL),(8,
 SELECT * FROM t1 WHERE b IS NULL or a IS NULL;
 a	b	c
 DROP TABLE t1;
+#
+# PS-5562 - index_merge + partitioning + MyRocks crashes the server
+#
+CREATE TABLE `t1` (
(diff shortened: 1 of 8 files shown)


george-lorch:
Minor clang-format failure.  Excellent diagnosis.  Thank you @satya-bodapati!

satya-bodapati:
> Minor clang-format failure. Excellent diagnosis. Thank you @satya-bodapati!

Yw George! :)

laurynas-biveinis (approved):

PR #3167, opened 2019-05-02
Post-merge fix to PS-5044 (Merge MySQL 8.0.14 & 8.0.15)

laurynas-biveinis:
Remove GCC 4.9 from Travis-CI test matrix because of this version not
being used for any of official packages, and having bugs whose
workarounds are not trivial. Moreover, MySQL 8.0.16 is C++14, which is
fully supported starting with GCC 5 only.


laurynas-biveinis:
https://travis-ci.org/percona/percona-server/builds/527209481

PR #3169, opened 2019-05-03
PS-5563 : Create new session variable rocksdb_disable_file_deletions

george-lorch:
- Fixed rocksdb_disable_file_deletions-master.opt to properly prefix rocksdb
  options with --loose


george-lorch:
Trivial fix, self approving.

PR #3171, opened 2019-05-06
Merge upstream 5.7.26 into Percona Server

percona-ysorokin:
Blueprint:
https://jira.percona.com/browse/PS-5101


percona-ysorokin:
Jenkins build link:
https://ps57.cd.percona.com/job/percona-server-5.7-param/24/

laurynas-biveinis (approved):

percona-ysorokin:
`auth_sec.keyring_file_data_qa` - failing on all platforms (only on Jenkins, not locally)
When this test is run under a user who can access '/', additional error might be generated in the error log file.
Fixed by adding a new 'mtr.add_suppression()'.

`main.plugin-load-add-with-path` used to be skipped in 5.7.25 because of the Bugs #26115002, #29178542.
This bug was fixed by Oracle in 5.7.26 (The binary file for the udf_example user-defined
function was omitted from binary distributions).
Now, when the test is not skipped any more, it requires modifications to support testing environments where '$PLUGIN_AUTH' and '$EXAMPLE_PLUGIN' are located in the same directory (Like Jenkins).

`rocksdb.multi_con_read` - already fixed by @satya-bodapati as part of the https://jira.percona.com/browse/PS-5562 "index_merge + partitioning + MyRocks crashes the server"

laurynas-biveinis:
@percona-ysorokin the latest changes are OK but ideally we'd like upstream branch to pass testing in our CI environment too, by changing the environment as needed

percona-ysorokin:
Additional Jenkins build link (`main.plugin-load-add-with-path --repeat=2`)
https://ps57.cd.percona.com/job/percona-server-5.7-param/33/

PR #3172, opened 2019-05-07
PS-5571: Merge encryption threads to 8.0.15 trunk

robgolebiowski:
This is umbrella projects that covers lots of bug fixes to make
encryption threads runnable and almost stable. The testcases can still
fail with assertion failure (PS-5593 this == space->crypt_data). Also
PS-5320 (Keyring encrypted tablespace import/export is broken) needs to
be fixed. This is what is known for now, however more bug fixes may be
needed.

The list of fixes this commit provide:

The DD encryption flag setting/clearing by encryption threads was added.

dd_update_tablespace_dd_flags was locking table for modification in
blocking mode. This was changed to timeout mode. The locking may fail
if the tablespace is being dropped.

encryption_key_id is saved in DD.

The innodb decryption errors were splitted into two:
HA_ERR_ENCRYPTION_KEY_MISSING and HA_ERR_DECRYPTION_FAILED. The first
one is generated when given encryption key (or version) cannot be found
in the keyring. The second one is generated when a key was found, but
page decryption failed. The most common scenario when this could happen
would be when keyring is loaded with needed key id, but the key itself
is not the correct one. The server code was amended so it should cope
with such decryption errors and generate appropriate errors instead of
crashing.

error ER_XB_MSG_4 was introduced for reporting decryption errors (when
checksum of the page do not match after decryption).
error ER_XB_MSG_5 was introduced for reporting that tablespace's
encryption key is missing.

There were two decryption errors codes - one coming from MariaDB
(DB_DECRYPTION_FAILED) and one coming from MySQL (DB_IO_DECRYPT_FAIL).
The were unified and now only DB_IO_DECRYPT_FAIL exists.

When there was a decryption error the server would go into infinite loop
when reading the page. This is now fixed.

keyring encrypted zip compressed pages had their checksums checked after
they were decompressed - thus reporting decryption errors as corruption
errors. Now when decompression fails it is checked if the page was
encrypted - if this happens the decryption error is generated.

Unencrypted pages could have been encrypted when a key was rotated. On
key rotation all the tablespaces with min_key_version == 0 (unencrypted)
would be encrypted with the latest key version (upstream bug). This was
fixed - key rotation no longer encrypts unencrypted
tables.

Removed generating error when encryption threads cannot flush changes to
a tablespace when tablespace is being removed.

Fixed starting and deactivating encryption threads.

Fixed updating variable innodb_pages0_read.

Fixed PS-5567 : Encryption tests should not leave any tables encrypted.

Fixed PS-5568 : New pages from buffer pool, that belong to encrypted
space can be mistaken for unencrypted ones by encryption threads

Fixed PS-5551 : Undo tablespace keyring encryption information gets
removed.

Fixed race problem with race condition for fil_crypt_threads_event. It
could have been reset while being in a set state before the wait on this
variable
occurred in fil_crypt_thread().

Added encryption module to innodb monitor with two variables:
pages_encrypted and pages_decrypted

Read applier now fails if there is a redo log that need to be applied to
a page for which decryption failed.

Fixed problem that when setting innodb-encryption-threads variable.
Increasing this variable will cause new threads to be created. During
THD object creation LOCK_global_system_variables is acquired. This will
cause the deadlock as LOCK_global_system_variable is already locked at
this point by variables framework. Resolved by adding additional
variable lock_global_system_var to create_thd (with default value ==
true).


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/123/

PR #3174, opened 2019-05-08
PS-5576 : Server stalls because ALTER TABLE on partitioned table hold…

satya-bodapati:
…s dict mutex

Why there is stall?
-------------------
1. Truncate holds dict_mutex and tries to free indexes. When indexes are huge (in
tera bytes), the dropping will take time.

See fseg_free_extent()->btr_search_drop_page_hash_when_freed()->buf_page_hash_get_low().

For every extent, for every page that is not marked free, there is check if page exists
in buffer pool (only peek, doesn't bring the page if it is not in buffer pool). This
check is mainly for AHI. Unfortunately, even if AHI is disabled, this check exists.

This iteration of extents * page in extents times hash searches take time. The dictionary
mutex is held for the entire duration.

Why concurrent inserts need dict_mutex?

Ideally not all inserts would need dict_mutex. This would have been terrible.
Only those inserts, where the TABLE* of table is not in cache, server has to open the table
from SE. InnoDB SE open tries to acquire dict_mutex to open table.

So all such inserts get blocked and they wait for dict_mutex. Remember they can be on different
tables. The inserts on truncated table are anyway blocked because of MDL.

Block of unrelated inserts on different tables is the core problem here. TRUNCATE TABLE t1 can
cause INSERT INTO t2 ..() or SELECT on t2 can also be blocked.

Fix:
----
Avoid buffer pool page hash lookups when AHI is disabled. This will reduce the time taken for
dict_sys mutex. So this patch solves only the TRUNCATE problem with innodb_adaptive_hash_index=OFF.

### DIFF
--- a/mysql-test/suite/innodb/t/percona_truncate_table_ahi.test
+++ b/mysql-test/suite/innodb/t/percona_truncate_table_ahi.test
@@ -1,0 +1,51 @@
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+--source include/count_sessions.inc
+
+CREATE TABLE t1(a INT, b INT, KEY k1(b));
+INSERT INTO t1 VALUES (1,1),(2,2),(3,3),(4,4),(5,5);
+CREATE TABLE t2 AS SELECT * FROM t1;
+FLUSH TABLES;
+
+SET GLOBAL innodb_limit_optimistic_insert_debug=2;
+SET GLOBAL innodb_adaptive_hash_index = OFF;
+
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+SELECT COUNT(*) FROM t1;
+
+SET DEBUG='+d, simulate_long_ahi';
+SET DEBUG_SYNC = 'truncate_table SIGNAL truncate_started WAIT_FOR resume_truncate';
+--send TRUNCATE TABLE t1;
+
+connect (con1,localhost,root,,);
+--connection con1
+--echo #connection con1
+SET DEBUG_SYNC = 'now WAIT_FOR truncate_started';
+SET DEBUG_SYNC = 'open_dict_table SIGNAL resume_truncate';
+SET @t_start:=unix_timestamp(now(6));
+SELECT * FROM t2;
+SET @t_end:=unix_timestamp(now(6));
+SET @duration:=@t_end-@t_start;
+
+--let $assert_text= The time taken for SELECT should be less than 5 seconds
+--let $assert_cond= [SELECT @duration] < 5
+--source include/assert.inc
+
+--connection default
+--echo # connection default
+--reap
+SET DEBUG_SYNC = 'RESET';
+--disconnect con1
+
+DROP TABLE t1;
+DROP TABLE t2;
+SET GLOBAL innodb_limit_optimistic_insert_debug=default;
+SET GLOBAL innodb_adaptive_hash_index = default;
+--source include/wait_until_count_sessions.inc
--- a/storage/innobase/btr/btr0btr.cc
+++ b/storage/innobase/btr/btr0btr.cc
@@ -921,17 +921,52 @@ btr_page_get_father(
 	mem_heap_free(heap);
 }
 
+/** Check whether AHI is enabled or not. Not this is only applicable
+for dropping of indexes during truncate.
+@param[in]	is_truncate 	true for truncate operation
+@return true always for non-truncate operations, for truncate, check
+the AHI status under lock */
+static
+bool
+btr_check_ahi_status(bool is_truncate) {
+
+	if (!is_truncate) {
+		return(true);
+	}
+
+	/* At drop index stage (only for truncate), concurrently AHI can only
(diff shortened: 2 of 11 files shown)


laurynas-biveinis (approved):
(please hold merging until 8.0 version is ready for merge too)

satya-bodapati:
@laurynas-biveinis I am preparing a patch with completely different approach.

satya-bodapati:
updated patch with new approach. dict_sys mutex is not released. Instead the path causing the longer acquisition time is fixed.

mysql-test/suite/innodb/t/percona_truncate_table_ahi.test
```diff
@@ -0,0 +1,50 @@
+--source include/have_debug.inc
```
laurynas-biveinis, line 1: Nit: have_debug_sync.inc as well

mysql-test/suite/innodb/t/percona_truncate_table_ahi.test
```diff
@@ -0,0 +1,50 @@
...
+--connection con1
+--echo #connection con1
+SET DEBUG_SYNC = 'now WAIT_FOR truncate_started';
+SET DEBUG_SYNC = 'open_dict_table SIGNAL resume_truncate';
+SET @t_start:=unix_timestamp(now(6));
+SELECT * FROM t2;
+SET @t_end:=unix_timestamp(now(6));
+SET @duration:=@t_end-@t_start;
+
+--let $assert_text= The time taken for SELECT should be less than 5 seconds
```
laurynas-biveinis, line 37: This will be unstable in our Jenkins

storage/innobase/btr/btr0btr.cc
```diff
@@ -921,17 +921,46 @@ btr_page_get_father(
...
+
+	/* At drop index stage (only for truncate), concurrently AHI can only
+	be enabled. It cannot be disabled because disabling needs dict_sys
+	mutex. See btr_search_disable(). We already own dict_sys mutex here.
+	Lock and get status of AHI. If we read it as disabled under
+	lock, it can become enabled after lock release. But it doesn't
+	matter because no one can add AHI entries at this stage. So
+	what we read under lock can be trusted */
+	ut_ad(mutex_own(&dict_sys->mutex));
+	btr_search_s_lock_all();
```
laurynas-biveinis: Don't need to lock all: any single partition should be enough

laurynas-biveinis (changes requested):
- I liked the original patch too. It seems that the patches are orthogonal, and even could be combined. What was the reason to change the approach?
- Does fixing AHI=off case address customer issue or further AHI=on fixing is needed?
- In general, we don't do lower-version-only fix if the issue exists on a higher version too, even if the fix would need reimplementation.

satya-bodapati:
>     * I liked the original patch too. It seems that the patches are orthogonal, and even could be combined. What was the reason to change the approach?

Good question. Two reasons
1) by releasing dict_sys mutex, one thing that I missed is the need to serialize operations on SYS_* dictionary tables. 

2) When I released dict_sys mutex, reacquiring at later stage caused latch order violations.  index->lock was already acquired and acquring dict_sys mutex after that is not allowed. The violations if ignored can lead to hang or deadlocks.

> 
>     * Does fixing AHI=off case address customer issue or further AHI=on fixing is needed?

With OFF, it should address the customer issue.  If AHI=on and drop/truncate partition has to be handled, PS-5634 should handle it.

> 
>     * In general, we don't do lower-version-only fix if the issue exists on a higher version too, even if the fix would need reimplementation.

In this case 8.0 is not affected. This is because truncate is implemented as drop+ create. DROP of file_per_table tablespace is different in 8.0. it is removal of AHI entries directly from buffer pool + no btr_free() of individual indexes + direct removal .ibd file.

mysql-test/suite/innodb/t/percona_truncate_table_ahi.test
satya-bodapati: I don't think so. in the test, it is finishing less than second. the extra time is buffer for slow systems.

satya-bodapati: If there is any failure, the time should be adjusted

storage/innobase/btr/btr0btr.cc
satya-bodapati: True, it can be improved. index_id should be passed from different paths. Since it is only about reading status, this should be quick.

mysql-test/suite/innodb/t/percona_truncate_table_ahi.test
laurynas-biveinis: I have seen overloaded Jenkins builds where threads were not scheduled to run on the order of tens of seconds. It really shakes out missing synchronization in otherwise good tests. Maybe things are better now, we can live with this until the 1st failure

storage/innobase/btr/btr0btr.cc
laurynas-biveinis: You don't even need index_id. [0] or any other ID will work.

mysql-test/suite/innodb/t/percona_truncate_table_ahi.test
satya-bodapati: Ok, Without the fix, it should take approximately 180 seconds. So we can adjust this later.

storage/innobase/btr/btr0btr.cc
satya-bodapati: Hmm, since it only about reading the variable under lock, I guess there is no harm in using 'any' index_id.

Will try to implement this

satya-bodapati: done

satya-bodapati:
ready for review

laurynas-biveinis (approved):

PR #3176, opened 2019-05-09
PS-5610: Fix keyring redo encryption in 32 bit builds (5.7)

dutow:
The location of the IV was dependent on a sizeof, resulting in an
incorrect data layout written in 32 bit builds.

### DIFF
--- a/storage/innobase/fsp/fsp0fsp.cc
+++ b/storage/innobase/fsp/fsp0fsp.cc
@@ -919,7 +919,7 @@ fsp_header_fill_encryption_info(
 
 	/* Write master key id. */
 	mach_write_to_4(ptr, key_version);
-	ptr += sizeof(ulint);
+	ptr += 4;
 
 	/* Write server uuid. */
 	memcpy(ptr, Encryption::uuid, ENCRYPTION_SERVER_UUID_LEN);
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -1139,7 +1139,7 @@ log_read_encryption()
 		unsigned char* info_ptr = log_block_buf + LOG_HEADER_CREATOR_END + ENCRYPTION_MAGIC_SIZE;
 	        uint version = mach_read_from_4(info_ptr);
 
-		memcpy(iv, info_ptr + ENCRYPTION_SERVER_UUID_LEN + 8, ENCRYPTION_KEY_LEN);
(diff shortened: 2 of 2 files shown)


dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/35/

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -919,7 +919,7 @@ fsp_header_fill_encryption_info(
 
 	/* Write master key id. */
 	mach_write_to_4(ptr, key_version);
-	ptr += sizeof(ulint);
+	ptr += 8;
```
laurynas-biveinis: That looks suspicious. mach_write_to_4 writes four bytes, why the pointer is advanced by 8 and not by 4?

satya-bodapati: For the entire proper fix. I was reviewer for this commit. See

commit f565013b7f31f225c349a6484ccab8cf4b5c8f52
Author: Allen Lai <zheng.lai@oracle.com>
Date:   Mon Dec 25 10:30:03 2017 +0800

    Bug#22740362 ENCRYPTED INNODB DATA FILES ARE NOT PORTABLE BETWEEN 32-BIT AND 64-BIT
    
    Problem:

Not sure if this can be ported as is. That is V3 format created by Percona Server might be unreadable by Oracle MySQL 5.7. 

AFAIR, the fix went only to 8.0

dutow: @satya-bodapati This is our own code - RK encryption isn't in upstream. I ended up copying a similar code from upstream (or mariadb?) using this sizeof, and only changed it to a fixed size on another part of the code.

@laurynas-biveinis Because this is only the part that writes it, the part that reads it already uses a fixed number, which is 8. This expression also evaluates to 8 on 64 bit builds, where we had no issues with this encryption mode. So I used 8 to keep the format the same as before.

laurynas-biveinis: @dutow in which case please don't leave uninitialized bytes, this will cause both Valgrind errors and might leak something what is not supposed to be leaked

dutow: @laurynas-biveinis the buffer is zero initialized in log_file_header_fill_encryption. (it is larger than what this function fills, so this isn't the only 4 unused byte in it)

dutow: Updated.

laurynas-biveinis (approved):

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/36/

PR #3195, opened 2019-05-23
8.0 PS-5642 - page tracker thread won't exist if startup fails

altmannmarcelo:
If page tracker is enabled and innodb fail to start,
srv_shutdown_all_bg_threads won't signal page tracker thread to exit.

This fix ensure srv_redo_log_follow_thread will be signaled.

### DIFF
--- a/mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread.test
+++ b/mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread.test
@@ -1,0 +1,20 @@
+#
+# PS-5642 - page tracker thread won't exist if startup fails
+#
+
+--source include/have_debug.inc
+--source include/not_embedded.inc
+
+SELECT 1;
+--source include/shutdown_mysqld.inc
+--error 1
+--exec $MYSQLD_CMD --innodb_track_changed_pages=TRUE --debug=d,ib_dic_boot_error --log-error=$MYSQLTEST_VARDIR/tmp/percona_changed_page_bmp_shutdown_thread.err
+
+--echo # Asserting that the error was not thrown.
+--let log_file_full_path=$MYSQLTEST_VARDIR/tmp/percona_changed_page_bmp_shutdown_thread.err
+--let grep_pattern=threads created by InnoDB had not exited at shutdown
+--let log_expected_matches=0
+--source include/log_grep.inc
+
+--source include/start_mysqld.inc
+--remove_file $MYSQLTEST_VARDIR/tmp/percona_changed_page_bmp_shutdown_thread.err
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -2506,7 +2506,8 @@ files_checked:
 		works for space 0. */
 
 		err = dict_boot();
-
+		DBUG_EXECUTE_IF("ib_dic_boot_error",
+				err = DB_ERROR;);
 		if (err != DB_SUCCESS) {
 			return(err);
 		}
--- a/mysql-test/suite/innodb/r/percona_changed_page_bmp_shutdown_thread.result
+++ b/mysql-test/suite/innodb/r/percona_changed_page_bmp_shutdown_thread.result
@@ -1,0 +1,6 @@
+SELECT 1;
+1
+1
+# Asserting that the error was not thrown.
+[log_grep.inc] file:  pattern: threads created by InnoDB had not exited at shutdown
+[log_grep.inc] lines:   0
--- a/mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread-master.opt
+++ b/mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread-master.opt
@@ -1,0 +1,1 @@
+--innodb_track_changed_pages=TRUE


altmannmarcelo:
https://ps80.cd.percona.com/job/percona-server-8.0-param/139/

mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread.test
```diff
@@ -0,0 +1,37 @@
+#
+# PS-5642 - page tracker thread won't exist if startup fails
+#
+
+--source include/have_debug.inc
```
laurynas-biveinis, line 5: Can you make this a non-debug test by poking a few bytes into actual redo log header while server is shutdown?

laurynas-biveinis (changes requested):
Please start with 5.6 to add the testcase to lower versions

mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread.test
altmannmarcelo: @laurynas-biveinis you mean replace some bytes at some specific position? such as last checkpoint on redo log header?
What I'm finding hard here is that page_tracking only starts after the redo log has been initialized. Even though I mess with redo log it will fail on a point where page tracking threads are not started yet.

laurynas-biveinis: Right. Perhaps it's possible to disrupt InnoDB startup at a point after redo initialization, maybe by deleting data file or setting some later incompatible startup option. If such alternative does not come up with little effort, then OK to leave as a debug test

altmannmarcelo: Hi @laurynas-biveinis , I've not been able to come up with a non-debug test that did not affect innodb start-up.

laurynas-biveinis: OK then

mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread.test
```diff
@@ -0,0 +1,37 @@
...
+
+SELECT 1;
+--source include/shutdown_mysqld.inc
+--error 1
+--exec $MYSQLD_CMD --innodb_track_changed_pages=TRUE --debug=d,ib_redo_log_corruption --log-error=$MYSQLTEST_VARDIR/tmp/percona_changed_page_bmp_shutdown_thread.err
+
+--echo # Asserting that the server tried to start
```
laurynas-biveinis: Probably can omit this assertion altogether - "error 1 / exec $MYSQLD_CMD" above should be good enough

mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread.test
```diff
@@ -0,0 +1,37 @@
...
+--let $grep_output= print_count
+--source include/grep_pattern.inc
+
+--echo # Asserting that the server aborted.
+--let $grep_file=$MYSQLTEST_VARDIR/tmp/percona_changed_page_bmp_shutdown_thread.err
+--let $grep_pattern=\[Server\] Aborting
+--let $grep_output= print_count
+--source include/grep_pattern.inc
+
+--echo # Asserting that the server did not start.
```
laurynas-biveinis: Likewise

storage/innobase/srv/srv0start.cc
```diff
@@ -2566,6 +2572,8 @@ dberr_t srv_start(bool create_new_db, const std::string &scan_directories) {
         recv_apply_hashed_log_recs(*log_sys, true);
       }
 
+      DBUG_EXECUTE_IF("ib_redo_log_corruption",
+                      recv_sys->found_corrupt_log = TRUE;);
```
```suggestion
                      recv_sys->found_corrupt_log = true;);
```

laurynas-biveinis (changes requested):
Please start with 5.6 and upmerge.
For the actual 8.0 version, if 8.0.17 lands in trunk by the time you make PR, re-check the srv0start.cc part - 8.0.17 reworked background threads

altmannmarcelo:
5.6 is actually not required.
`innobase_shutdown_for_mysql` will call `logs_empty_and_mark_files_at_shutdown` which wake up redo follow thread at:

https://github.com/percona/percona-server/blob/5.6/storage/innobase/log/log0log.cc#L3586
and 
https://github.com/percona/percona-server/blob/5.6/storage/innobase/log/log0log.cc#L3672

altmannmarcelo:
5.7 is not affected as well.
If InnoDB finds a corruption, it will only send a warning while in 8.0 it aborts the start-up:

5.7 - https://github.com/percona/percona-server/blob/5.7/storage/innobase/srv/srv0start.cc#L2578
8.0 - https://github.com/percona/percona-server/blob/8.0/storage/innobase/srv/srv0start.cc#L2623

altmannmarcelo:
https://ps80.cd.percona.com/job/percona-server-8.0-param/297/

@laurynas-biveinis ready for next review.

laurynas-biveinis (changes requested):
The work for lower versions is not to fix but rather add the regression testcase

altmannmarcelo:
https://ps80.cd.percona.com/job/percona-server-8.0-param/301/

mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread.test
```diff
@@ -0,0 +1,25 @@
...
+--error 1
+--exec $MYSQLD_CMD --innodb_track_changed_pages=TRUE --debug=d,ib_dic_boot_error --log-error=$MYSQLTEST_VARDIR/tmp/percona_changed_page_bmp_shutdown_thread.err
+
+--echo # Asserting that the error was not thrown.
+--let log_file_full_path=$MYSQLTEST_VARDIR/tmp/percona_changed_page_bmp_shutdown_thread.err
+--let grep_pattern=threads created by InnoDB had not exited at shutdown
+--let log_expected_matches=0
+--source include/log_grep.inc
+
+--echo # Asserting that the server aborted.
```
laurynas-biveinis: I still believe this assert is redundant and tested by error 1/exec MYSQLD_CMD above. Any particular reason to keep it?

altmannmarcelo:
ready for next review

altmannmarcelo:
@dutow Ready for next review. I've rebased my branch with the latest 8.0 commit.

dutow (approved):
Approved, but please wait with the merging until the 8.0.17 merge lands.

PR #3196, opened 2019-05-24
PS-5310 : Test innodb.percona_session_temp_tablespaces_encrypt is uns…

satya-bodapati:
…table

Problem:
--------
The test sometimes fails with difference in tablespace id used by connection id 11.

This is because connection disconnect is async. In regular case, connection 10
disconnects first and then connection 9.

Each connection disconnect releases session temporary tablespaces back to pool.
So the order of connection disconnect will decide the tablespace used by next connection.

On stress, it is observed that connection 9 disconnects first and then the connection 10.
Hence there is result content mismatch on the "tablespace id" used by connection 11.

Fix:
----
Use assert.inc on the number of expected active and inactive session temporary tablespaces
instead of displaying all tablespaces in session temporary tablespaces pool.

### DIFF
--- a/mysql-test/suite/innodb/t/percona_session_temp_tablespaces_encrypt.test
+++ b/mysql-test/suite/innodb/t/percona_session_temp_tablespaces_encrypt.test
@@ -95,6 +95,7 @@ SELECT * FROM INFORMATION_SCHEMA.INNODB_SESSION_TEMP_TABLESPACES WHERE ID = conn
 connection default;
 disconnect con1;
 disconnect con2;
+--source include/wait_until_count_sessions.inc
 
 SET GLOBAL innodb_temp_tablespace_encrypt=default;
 SET GLOBAL innodb_encrypt_tables=default;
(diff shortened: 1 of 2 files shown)


mysql-test/suite/innodb/t/percona_session_temp_tablespaces_encrypt.test
```diff
@@ -104,9 +104,13 @@ DROP TABLE t6;
 connect (con3, localhost, root,,);
 connection con3;
 
---replace_column 1 ID
---replace_regex /\\#innodb_temp\\temp/\/#innodb_temp\/temp/
-SELECT * FROM INFORMATION_SCHEMA.INNODB_SESSION_TEMP_TABLESPACES ORDER BY SPACE;
+--let $assert_text = Active session temp tablespaces count should be 2
```
laurynas-biveinis, line 108: The asserts should sync with disconnect con1/con2 by wait_until_count_sessions.inc with a proper parameter

laurynas-biveinis (changes requested):

mysql-test/suite/innodb/t/percona_session_temp_tablespaces_encrypt.test
satya-bodapati: Fixed.

laurynas-biveinis (approved):

PR #3208, opened 2019-05-29
PS-5149: Redo log encryption type shouldn't be changed on a running server (5.7)

dutow:
We support two redo log encryption modes:
* master_key, as upstream
* keyring_key, a per-page versioned encryption mode

The active encryption mode is chosen using the innodb-redo-log-encypt variable. This currently allows the user to select one mode, then change to the other, resulting in errors if the server has to decrypt earlier redo log blocks. To fix this, we disallow changing redo log encryption modes.

* First, during startup, if the log is already encrypted, we detect the existing encryption mode
* Later, when the user changes this variable, we check against the existing encryption mode. If the logs aren't encrypted, or if they were encrypted in the same mode, we allow the change. Otherwise, we report a warning and leave the variable as is.
* If we changed the encryption mode in the previous step, we also update the existing encryption mode to match it.

This way we disallow both direct (e.g. master_key -> keyring_key) and indirect (e.g. master_key -> off -> server restart -> keyring_key) mode changes, but allow changes back to the same mode (e.g. master_key -> off (-> server restarts, etc) -> master_key).

To change the encryption mode, we have to:

* shutdown the server
* delete the redo log files
* start up the server with the new encryption settings


mysql-test/include/percona_log_encrypt_change.inc
```diff
@@ -0,0 +1,28 @@
+# InnoDB transparent tablespace data encryption
+# This test case will verify that log encryption type can't be changed
+
+call mtr.add_suppression("Redo log encryption mode can't be switched without stopping the server and recreating the redo logs");
```
laurynas-biveinis: Suppress after checking innodb and not embedded

mysql-test/include/percona_log_encrypt_change.inc
```diff
@@ -0,0 +1,28 @@
...
+--eval SET GLOBAL innodb_redo_log_encrypt=$LOG_ENCRYPT_OTHER_TYPE
+SELECT @@innodb_redo_log_encrypt;
+
+# Test: changing when it is turned off, but the logs ARE encrypted
+
+--eval SET GLOBAL innodb_redo_log_encrypt=OFF
+--eval SET GLOBAL innodb_redo_log_encrypt=$LOG_ENCRYPT_OTHER_TYPE
+SELECT @@innodb_redo_log_encrypt;
+
+--let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err
```
laurynas-biveinis: Error log grep antipattern here, the error log contents will have all the previous test content, please look into assert_only_after=CURRENT_TEST: test_name of assert_grep.inc, if not practical, then use a custom error log path

storage/innobase/log/log0log.cc
```diff
@@ -62,7 +62,7 @@ Created 12/9/1995 Heikki Tuuri
 
 #include "system_key.h"
 
-static redo_log_encrypt_enum found_log_encryption_mode = REDO_LOG_ENCRYPT_OFF;
+redo_log_encrypt_enum existing_redo_encryption_mode = REDO_LOG_ENCRYPT_OFF;
```
laurynas-biveinis, line 65: IMHO "current" is a better fit than "existing", but this is up to you

laurynas-biveinis (changes requested):

storage/innobase/log/log0log.cc
dutow: This could be true even if the actual setting is off (if the log has previously encrypted records) - I think current would be misleading.

satya-bodapati, line 65: lets follow the running style used. So have the name log_existing_encryption_type /mode.

Also add a declaration in header log0log.h

storage/innobase/srv/srv0srv.cc
```diff
@@ -3509,6 +3509,8 @@ srv_fatal_error()
 	exit(3);
 }
 
+extern redo_log_encrypt_enum existing_redo_encryption_mode;
```
satya-bodapati: remove this. if u add the declaration in log0log.h it should work

storage/innobase/srv/srv0srv.cc
```diff
@@ -3542,6 +3544,15 @@ srv_enable_redo_encryption()
 bool
 srv_enable_redo_encryption_mk()
 {
+	if (existing_redo_encryption_mode != REDO_LOG_ENCRYPT_MK
```
satya-bodapati: Can we rewrite it as if( existing_redo_encryption_mode == REDO_LOG_ENCRYPT_RK)

storage/innobase/srv/srv0srv.cc
```diff
@@ -3542,6 +3544,15 @@ srv_enable_redo_encryption()
...
+	    && existing_redo_encryption_mode != REDO_LOG_ENCRYPT_OFF)
+	{
+		ib::warn() << "Redo log encryption mode can't be switched without "
```
satya-bodapati: I think SET command should throw an error. Not the ib::error(), Send error to client. Can you please check this.

Also startup should fail with clear message. "Current redo logs are encrypted with MK. Please start with innodb_log_encrypt=MK to continue startup with MK Or delete redo logs blah blah.."

satya-bodapati (changes requested):

storage/innobase/srv/srv0srv.cc
dutow: Everything else redo encryption related only results in a warning and continues without doing anything - including the server start. (and this is upstream behavior, we shouldn't change it)

I'll update the sending to the client part.

storage/innobase/srv/srv0srv.cc
dutow: We can, but I don't think that would be better. This condition now exits on explicitly anything not handled by this function - that would return on one specific something not handled by it, which is more fragile if we ever add a third option.

satya-bodapati: May be switch case then? that way new types are forced to handle in switch case

storage/innobase/srv/srv0srv.cc
satya-bodapati: well, upstream doesn't have two types, isn't it? So can you please explain what you mean by upstream behaviour here?

dutow: I mean every other possible problem with redo encryption. Like the test case I've added in my previous PR - if there is no keyring, we display a warning and continue without encryption.
That's how basically any error handling works in redo encryption, and unless the server actually ends up doing recovery, and fails to decrypts some pages, it continues to run.

storage/innobase/srv/srv0srv.cc
```diff
@@ -3542,6 +3542,16 @@ srv_enable_redo_encryption()
 bool
 srv_enable_redo_encryption_mk()
 {
+	switch (existing_redo_encryption_mode) {
+	case REDO_LOG_ENCRYPT_RK:
+		ib::error() << "Redo log encryption mode can't be switched without "
```
satya-bodapati: Add error code IB_XB_ and also change the message to print the current and desired redo modes. "Cannot change from X to Y"

IMHO, the warning sent to client should be converted to error.

storage/innobase/srv/srv0srv.cc
```diff
@@ -3575,6 +3585,16 @@ srv_enable_redo_encryption_mk()
 bool
 srv_enable_redo_encryption_rk()
 {
+	switch (existing_redo_encryption_mode) {
+	case REDO_LOG_ENCRYPT_MK:
+		ib::error() << "Redo log encryption mode can't be switched without "
```
satya-bodapati: same as above

satya-bodapati (changes requested):

storage/innobase/srv/srv0srv.cc
dutow: I don't think 5.7 ib::error supports this - all existing uses are with hardcoded strings, error codes were only added in 8.0.

mysql-test/suite/innodb/r/percona_log_encrypt_change_mk.result
```diff
@@ -0,0 +1,17 @@
...
+@@innodb_redo_log_encrypt
+master_key
+SET GLOBAL innodb_redo_log_encrypt=KEYRING_KEY;
+Warnings:
+Warning	1210	 Redo log encryption mode can't be switched without stopping the server and recreating the redo logs.
+SELECT @@innodb_redo_log_encrypt;
+@@innodb_redo_log_encrypt
+master_key
+SET GLOBAL innodb_redo_log_encrypt=OFF;
+SET GLOBAL innodb_redo_log_encrypt=KEYRING_KEY;
```
satya-bodapati, line 12: THis is supposed to throw warning right? Current is MK. MK->Keying warning is there
MK to OFF (no warning) and after OFF, to KEYRING_KEY (with MK existing) should throw warning right?

dutow: Ops, and it does - but it passes the first check in ha_innodb, and the second check in srv0srv places the warning in the error log. I'll fix this

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21468,14 +21468,17 @@ update_innodb_redo_log_encrypt(
...
-		push_warning_printf(
-			thd, Sql_condition::SL_WARNING,
-			ER_WRONG_ARGUMENTS,
+	if (existing_redo_encryption_mode != REDO_LOG_ENCRYPT_OFF
+	    && existing_redo_encryption_mode != target) {
+		static const char err_msg[] = 
 			" Redo log encryption mode"
 			" can't be switched without stopping the server and"
-			" recreating the redo logs.");
+			" recreating the redo logs.";
```
satya-bodapati: I think, it is better to show(and print) the current detected redo log encryption mode (note that even though variable is OFF, detected mode could be MK or Keyring) and the desired mode.

satya-bodapati (changes requested):

storage/innobase/handler/ha_innodb.cc
dutow: I can change it, but this message is only shown when the user wants to set keyring or master key - so obliviously the existing mode is the other.

storage/innobase/log/log0log.cc
```diff
@@ -1098,6 +1098,18 @@ log_group_file_header_flush(
...
 }
 
+const char* log_encrypt_name(redo_log_encrypt_enum val) {
+	switch(val) {
+	case REDO_LOG_ENCRYPT_OFF:
+		return "off";
+	case REDO_LOG_ENCRYPT_MK:
+		return "master_key";
+	case REDO_LOG_ENCRYPT_RK:
+		return "rotated_key";
```
satya-bodapati: Although we have chosen the variable name as RK, current terminology is "keyring_key" I think.

Optional comment.

satya-bodapati (approved):

storage/innobase/log/log0log.cc
dutow: Right, I should have changed that `RK` constant, I always mix up things because of it. Fixed.

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/48/#showFailuresLink

PR #3209, opened 2019-05-31
PS-5668 : Ensure keyring encryption header doesn't cross MK header en…

satya-bodapati:
…cryption size

Ensure that the header size written by keyring key encryption doesn't cross the
encryption header size used by Master Key encryption.
Once the header size is crossed, it will corrupt the data written for other features.

Size of MK encryption header size: ENCRYPTION_INFO_MAX_SIZE ( 115 bytes)

Size of Keyring encryption header size: fil_get_encrypt_info_size(16) = 105.

Fix:
---
Added assert that checks the size of keyring key header size (written at Page 0) is
less than the Master Key encryption header size.

### DIFF
--- a/storage/innobase/fil/fil0crypt.cc
+++ b/storage/innobase/fil/fil0crypt.cc
@@ -118,19 +118,27 @@ uint get_global_default_encryption_key_id_value();
 
 #define DEBUG_KEYROTATION_THROTTLING 0
 
-uint fil_get_encrypt_info_size(const uint iv_len) {
-  return ENCRYPTION_MAGIC_SIZE + 2  //length of iv
-         + 4                        // space id
-         + 2                        // offset
-         + 1                        // type
-         + 4                        // min_key_version
-         + 4                        // key_id
-         + 1                        // encryption
-         + iv_len                   // iv
-         + 4                        // encryption rotation type
-         + ENCRYPTION_KEY_LEN       // tablespace key
-         + ENCRYPTION_KEY_LEN;      // tablespace iv
-}
+static constexpr uint KERYING_ENCRYPTION_INFO_MAX_SIZE =
+    ENCRYPTION_MAGIC_SIZE + 2  // length of iv
+    + 4                        // space id
+    + 2                        // offset
+    + 1                        // type
+    + 4                        // min_key_version
+    + 4                        // key_id
+    + 1                        // encryption
+    + CRYPT_SCHEME_1_IV_LEN    // iv (16 bytes)
+    + 4                        // encryption rotation type
+    + ENCRYPTION_KEY_LEN       // tablespace key
+    + ENCRYPTION_KEY_LEN;      // tablespace iv
+
+/* The size of keyring key encrption header cannot cross the Master
+Key header. This is because the bytes followed by MK header are used
+by other features like SDI, encryption progress (MK) etc. If we cross
+the MK encryption header size ENCRYPTION_INFO_MAX_SIZE, we corrupt the
+header and other features will not work */
+static_assert(KERYING_ENCRYPTION_INFO_MAX_SIZE < ENCRYPTION_INFO_MAX_SIZE,
+              "Keyring key encryption header crosses Master Key encryption"
+              " header size");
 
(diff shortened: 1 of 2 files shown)


storage/innobase/fil/fil0crypt.cc
```diff
@@ -118,18 +118,28 @@ uint get_global_default_encryption_key_id_value();
...
-         + 4                        // key_id
-         + 1                        // encryption
-         + iv_len                   // iv
-         + 4                        // encryption rotation type
-         + ENCRYPTION_KEY_LEN       // tablespace key
-         + ENCRYPTION_KEY_LEN;      // tablespace iv
+/* Return the size of encryption header used Keyring key
+encryption.
+@return length of encryption header */
+static uint fil_get_keyring_encrypt_info_size() {
```
laurynas-biveinis: Mark the function constexpr

storage/innobase/fil/fil0crypt.cc
```diff
@@ -118,18 +118,28 @@ uint get_global_default_encryption_key_id_value();
...
+                    + CRYPT_SCHEME_1_IV_LEN    // iv (16 bytes)
+                    + 4                        // encryption rotation type
+                    + ENCRYPTION_KEY_LEN       // tablespace key
+                    + ENCRYPTION_KEY_LEN;      // tablespace iv
+  /* The size of keyring key encrption header cannot cross the Master
+  Key header. This is because the bytes followed by MK header are used
+  by other features like SDI, encryption progress (MK) etc. If we cross
+  the MK encryption header size ENCRYPTION_INFO_MAX_SIZE, we corrupt the
+  header and other features will not work */
+  ut_a(size < ENCRYPTION_INFO_MAX_SIZE);
```
laurynas-biveinis: static_assert, if both are constants, compile time checking is sufficient and efficient

storage/innobase/fil/fil0crypt.cc
```diff
@@ -118,18 +118,28 @@ uint get_global_default_encryption_key_id_value();
...
-         + 1                        // encryption
-         + iv_len                   // iv
-         + 4                        // encryption rotation type
-         + ENCRYPTION_KEY_LEN       // tablespace key
-         + ENCRYPTION_KEY_LEN;      // tablespace iv
+/* Return the size of encryption header used Keyring key
+encryption.
+@return length of encryption header */
+static uint fil_get_keyring_encrypt_info_size() {
+  const uint size = ENCRYPTION_MAGIC_SIZE + 2  //length of iv
```
laurynas-biveinis:
```suggestion
  const uint size = ENCRYPTION_MAGIC_SIZE + 2  // length of iv
```

laurynas-biveinis (changes requested):

storage/innobase/fil/fil0crypt.cc
satya-bodapati: leads to compile error:

variable declaration in a constexpr function is a C++14 extension

laurynas-biveinis: Just fold const uint size ... ; return size; to return expression, and move static_assert to file scope

satya-bodapati: Ok, then this function make any sense. Just removed it. Please check now

storage/innobase/fil/fil0crypt.cc
```diff
@@ -118,19 +118,27 @@ uint get_global_default_encryption_key_id_value();
...
-         + 1                        // type
-         + 4                        // min_key_version
-         + 4                        // key_id
-         + 1                        // encryption
-         + iv_len                   // iv
-         + 4                        // encryption rotation type
-         + ENCRYPTION_KEY_LEN       // tablespace key
-         + ENCRYPTION_KEY_LEN;      // tablespace iv
-}
+static const uint KERYING_ENCRYPTION_INFO_MAX_SIZE =
```
laurynas-biveinis:
```suggestion
static constexpr uint KERYING_ENCRYPTION_INFO_MAX_SIZE =
```

laurynas-biveinis (changes requested):
OK to push with the change, no further reviews needed

PR #3210, opened 2019-05-31
PS-5565: Better versioning with keyring redo encryption (5.7)

dutow:
Previously keyring redo encryption only selected a version when the
redo log was created - changing the key version required recreating
the redo log.

With this change, individual pages all have separate versions. To
accomplish this, we reuse the checksum field: during encryption, we
assume that the checksum is correct, and replace it with the
checksum of the encrpyted data, plus the key version.
During reads, we calculate the checksum again, and assuma that the
difference between the written data and the calcalated value is the
key version.
An incorrect checksum will result in an incorrect key version,
resulting in quite likely a decryption error.
In the unlikely case that the checksum is wrong, but the resulting
number refers to an existing key, the decrypted data will be
garbage, and log recovery will abort later.

Since the redo key version now dynamic, it is exported as the
innodb_encryption_redo_key_version status variable.

To support this change, keyring handling of the redo log keys were
extracted to the redo_log_key_mgr global variable. This variable is
now the owner of all used keyring redo keys, and everything else
uses pointers to keys managed by it.
This way:
* keyring key handling is extracted to a common place, without
  too much code duplicated in multiple files
* key-copying races between threads changing the keys and others
writing data to the redo log are prevented, as it is a simple
pointer change now

The commit also includes minor fixes for issues previously hidden
by the limited usability of keyring redo encryption.

### DIFF
--- a/mysql-test/include/log_encrypt_2.inc
+++ b/mysql-test/include/log_encrypt_2.inc
@@ -32,15 +32,15 @@ INSERT INTO t1 VALUES(6, "ggggg");
 INSERT INTO t1 VALUES(7, "hhhhh");
 INSERT INTO t1 VALUES(8, "iiiii");
 INSERT INTO t1 VALUES(9, "jjjjj");
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
 
 SELECT * FROM t1 LIMIT 10;
 
@@ -50,7 +50,14 @@ SELECT * FROM t1 LIMIT 10;
 
 SELECT * FROM t1 LIMIT 10;
 
-# Key rotation.
+SELECT rotate_system_key("percona_redo");
+
+if ($wait_for_key_version)
+{
+--let $wait_condition = select variable_value = 2 from performance_schema.global_status where variable_name = 'innodb_encryption_redo_key_version'
+--source include/wait_condition.inc
+}
+
 ALTER INSTANCE ROTATE INNODB MASTER KEY;
 
 DROP TABLE t1;
@@ -68,15 +75,15 @@ INSERT INTO t1 VALUES(6, "ggggg");
 INSERT INTO t1 VALUES(7, "hhhhh");
 INSERT INTO t1 VALUES(8, "iiiii");
 INSERT INTO t1 VALUES(9, "jjjjj");
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
-INSERT INTO t1 select * from t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
 
 # Restart to confirm the encryption info can be retrieved properly.
 --let restart_parameters="restart:$KEYRING_PARAMS"
@@ -84,6 +91,13 @@ INSERT INTO t1 select * from t1;
 --source include/kill_and_restart_mysqld.inc
 
 SELECT * FROM t1 LIMIT 10;
+
+let $restart_parameters = restart: $KEYRING_PARAMS --general-log --log-output=FILE --general_log_file=$MYSQL_TMP_DIR/keyring_query_log;
+--replace_result $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT --plugin-dir=KEYRING_PLUGIN_PATH
+--replace_regex /\.dll/.so/
+--source include/restart_mysqld_no_echo.inc
+
+INSERT INTO t1 SELECT * FROM t1;
 DROP TABLE t1;
 
 let $restart_parameters = restart: $KEYRING_PARAMS --general-log --log-output=FILE --general_log_file=$MYSQL_TMP_DIR/keyring_query_log;
--- a/mysys/system_key.c
+++ b/mysys/system_key.c
@@ -32,7 +32,7 @@ struct Valid_percona_system_key
 */
 struct Valid_percona_system_key valid_percona_system_keys[] = { {PERCONA_BINLOG_KEY_NAME, 16, FALSE},
                                                                 {PERCONA_INNODB_KEY_NAME, 32, TRUE},
-								{PERCONA_REDO_KEY_NAME, 16, FALSE}};
+								{PERCONA_REDO_KEY_NAME, 32, FALSE}};
 const size_t valid_percona_system_keys_size = array_elements(valid_percona_system_keys);
 
 my_bool is_valid_percona_system_key(const char *key_name, size_t *key_length)
--- a/storage/innobase/fil/fil0crypt.cc
+++ b/storage/innobase/fil/fil0crypt.cc
@@ -24,6 +24,7 @@ Modified           Jan Lindström jan.lindstrom@mariadb.com
 *******************************************************/
 
 #include "fil0fil.h"
+#include "system_key.h"
 #include "mtr0types.h"
 #include "mach0data.h"
 #include "page0size.h"
@@ -3144,3 +3145,181 @@ fil_space_verify_crypt_checksum(
 
 	return(encrypted);
 }
+
+redo_log_key*
+redo_log_keys::load_latest_key(bool generate) {
+	size_t klen = 0;
+	char*   key_type = NULL;
+	byte*   rkey = NULL;
+
+	if (my_key_fetch(PERCONA_REDO_KEY_NAME, &key_type, NULL,
+			 reinterpret_cast<void **>(&rkey), &klen) ||
+	    rkey == NULL || strncmp(key_type, "AES", 4) != 0) {
+		/* There is no key yet, we'll try to generate one */
+		my_free(rkey);
+		if (!generate) {
+			return(NULL);
+		}
+		return(generate_and_store_new_key());
+	}
+
+
+	uint       version = 0;
+	byte*       rkey2 = NULL;
+	size_t     klen2 = 0;
+	const bool err =
+	    (parse_system_key(rkey, klen, &version, &rkey2, &klen2) ==
+	     reinterpret_cast<uchar *>(NullS));
+	if (err) {
+		my_free(rkey);
+		my_free(rkey2);
+		my_free(key_type);
+		return(NULL);
+	}
+
+	ut_ad(klen2 == ENCRYPTION_KEY_LEN);
+
+	key_iterator it = m_keys.find(version);
+
+	if (it != m_keys.end() && it->second.present) {
+		ut_ad(memcmp(it->second.key, rkey2, 32) == 0);
+		return(&it->second);
+	}
+
+	redo_log_key* rk = &m_keys[version];
+	rk->version = version;
+	rk->present = true;
+	memcpy(rk->key, rkey2, ENCRYPTION_KEY_LEN);
+
+	my_free(rkey);
+	my_free(rkey2);
+	my_free(key_type);
+
+	return(rk);
+}
+
+redo_log_key*
+redo_log_keys::load_key_version(uint version) {
+	key_iterator it = m_keys.find(version);
+
+	if (it != m_keys.end() && it->second.present) {
+		return(&it->second);
+	}
+
+	size_t klen = 0;
+	char*  key_type = NULL;
+	byte*  rkey = NULL;
+
+	std::ostringstream percona_redo_with_ver_ss;
+	percona_redo_with_ver_ss << PERCONA_REDO_KEY_NAME << ':' << version;
+	if (my_key_fetch(percona_redo_with_ver_ss.str().c_str(), &key_type,
+			 NULL, reinterpret_cast<void **>(&rkey), &klen) ||
+	    rkey == NULL || strncmp(key_type, "AES", 4) != 0) {
+		my_free(rkey);
+		my_free(key_type);
+		ib::error() << "Failed to load redo key version " << version;
+		return(NULL);
+	}
+
+	ut_ad(klen == ENCRYPTION_KEY_LEN);
+
+	redo_log_key *rk = &m_keys[version];
+	rk->version = version;
+	rk->present = true;
+	memcpy(rk->key, rkey, ENCRYPTION_KEY_LEN);
+
+	my_free(rkey);
+	my_free(key_type);
+
+	return(rk);
+}
+
+redo_log_key*
+redo_log_keys::generate_and_store_new_key() {
+	if (my_key_generate(PERCONA_REDO_KEY_NAME, "AES", NULL,
+			    ENCRYPTION_KEY_LEN)) {
+		ib::error() << "Redo log key generation failed.";
+		return(NULL);
+	}
+
+	char*  redo_key_type = NULL;
+	byte*  rkey = NULL;
+	size_t klen = 0;
+
+	if (my_key_fetch(PERCONA_REDO_KEY_NAME, &redo_key_type, NULL,
+			 reinterpret_cast<void **>(&rkey), &klen)) {
+		ib::error() << "Couldn't fetch newly generated redo key.";
+		my_free(redo_key_type);
+		my_free(rkey);
+		return(NULL);
+	}
+
+	DBUG_ASSERT(rkey != NULL);
+	byte*  rkey2 = NULL;
+	size_t klen2 = 0;
+	uint   version = 0;
+
+	bool err = (parse_system_key(rkey, klen, &version, &rkey2, &klen2) ==
+		    reinterpret_cast<uchar *>(NullS));
+
+	ut_ad(klen2 == ENCRYPTION_KEY_LEN);
+
+	if (err) {
+		ib::error() << "Couldn't parse system key: " << rkey;
+		my_free(redo_key_type);
+		my_free(rkey);
+		return(NULL);
+	}
+
+	redo_log_key* rk = &m_keys[version];
+	rk->version = version;
+	memcpy(rk->key, rkey2, ENCRYPTION_KEY_LEN);
+	rk->present = true;
+
+	my_free(redo_key_type);
+	my_free(rkey);
+	my_free(rkey2);
+
+	return(rk);
+}
+
+redo_log_key*
+redo_log_keys::generate_new_key_without_storing() {
+	ut_ad(m_keys.empty());
+	Encryption::random_value(reinterpret_cast<byte *>(&m_keys[0].key));
+	return(&m_keys[0]);
+}
+
+bool
+redo_log_keys::store_used_keys() {
+	/* This is a for loop, but it really only should store a key
+	   with current version 0 */
+	for (key_iterator it = m_keys.begin(); it != m_keys.end(); ++it) {
+		if (!it->second.persisted()) {
+			ut_ad(it->first == 0);
+			if (my_key_store(PERCONA_REDO_KEY_NAME, "AES", NULL,
+					 it->second.key,
+					 ENCRYPTION_KEY_LEN)) {
+				return(false);
+			}
+		}
+	}
+
+	return(true);
+}
+
+void
+redo_log_keys::unload_old_keys() {
+	if (m_keys.size() == 0) {
+		return;
+	}
+	redo_log_key* last = &(--m_keys.end())->second;
+	for (key_iterator it = m_keys.begin(); it != m_keys.end(); ++it) {
+		if (&it->second != last) {
+			it->second.present = false;
+			memset(it->second.key, 0, ENCRYPTION_KEY_LEN);
+		}
+	}
+}
+
+redo_log_keys redo_log_key_mgr;
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -5868,6 +5868,34 @@ fil_io_set_keyring_encryption(IORequest& req_type,
 	mutex_exit(&space->crypt_data->mutex);
 }
 
+static void
+fil_io_set_mk_encryption(IORequest& req_type, fil_space_t* space)
+{
+	unsigned char* key = space->encryption_redo_key != NULL
+		? reinterpret_cast<unsigned char*>(space->encryption_redo_key->key)
+		: space->encryption_key;
+	uint version = space->encryption_redo_key != NULL
+		? space->encryption_redo_key->version
+		: space->encryption_key_version;
+	req_type.encryption_key(key,
+				32,
+				false,
+				space->encryption_iv,
+				version, 0, NULL, NULL);
+
+	req_type.encryption_rotation(Encryption::NO_ROTATION);
+}
+
+static
+bool
+fil_keyring_skip_encryption(const page_id_t& page_id) {
+
+	/* Don't encrypt TRX_SYS_SPACE.TRX_SYS_PAGE_NO as it contains address
+	to dblwr buffer */
+	return((page_id.space() == TRX_SYS_SPACE
+		&& page_id.page_no() == TRX_SYS_PAGE_NO) ? true : false);
+}
+
 /** Set encryption information for IORequest.
 @param[in,out]	req_type	IO request
 @param[in]	page_id		page id
@@ -5926,30 +5954,59 @@ fil_io_set_encryption(
 		return;
 	}
 
-	/* Don't encrypt the page 0 of all tablespaces,
-	   don't encrypt TRX_SYS_SPACE.TRX_SYS_PAGE_NO as it contains address to dblwr buffer */
-	if ((req_type.is_log()
-	     || (page_id.page_no() > 0
-		 && (TRX_SYS_SPACE != page_id.space()
-		     || TRX_SYS_PAGE_NO != page_id.page_no())))
-	    && space->encryption_type != Encryption::NONE) {
-		if (space->encryption_type == Encryption::KEYRING)  {
-			ut_ad(space->crypt_data != NULL);
-
-			fil_io_set_keyring_encryption(req_type, space, page_id);
-		} else {
-			ut_ad(space->encryption_type == Encryption::AES);
-			req_type.encryption_key(space->encryption_key,
-						32,
-						false,
-						space->encryption_iv,
-						0, 0, NULL, NULL); // not relevant for Master Key encryption
-
-			req_type.encryption_rotation(Encryption::NO_ROTATION);
-		}
-		req_type.encryption_algorithm(space->encryption_type);
-	} else {
+	if (space->encryption_type == Encryption::NONE) {
 		req_type.clear_encrypted();
+		return;
+	}
+
+        if (req_type.is_log())  {
+		/* redo log encryption */
+		ut_ad(page_id.space() == SRV_LOG_SPACE_FIRST_ID);
+
+		switch (space->encryption_type) {
+		case Encryption::AES:
+		case Encryption::KEYRING:
+			// Both MK and Keyrig key use same style for redo log
+			fil_io_set_mk_encryption(req_type, space);
+		        req_type.encryption_algorithm(space->encryption_type);
+			return;
+		case Encryption::NONE:
+			// Already handled above
+		default:
+			ut_a(0);
+		}
+
+	} else {
+		/* tablespace encryption */
+
+		/* Don't encrypt the page 0 of all tablespaces */
+		if (page_id.page_no() == 0) {
+		    req_type.clear_encrypted();
+		    return;
+		}
+
+		switch (space->encryption_type) {
+		case Encryption::KEYRING:
+			if (fil_keyring_skip_encryption(page_id)) {
+				req_type.clear_encrypted();
+				return;
+			} else {
+				fil_io_set_keyring_encryption(
+					req_type, space, page_id);
+				req_type.encryption_algorithm(
+						space->encryption_type);
+				return;
+			}
+
+		case Encryption::AES:
+			fil_io_set_mk_encryption(req_type, space);
+		        req_type.encryption_algorithm(space->encryption_type);
+			return;
+		case Encryption::NONE:
+			// Already handled above
+		default:
+			ut_a(0);
+		}
 	}
 }
 
@@ -7957,7 +8014,6 @@ fil_set_encryption(
 
 	if (key == NULL) {
 		Encryption::random_value(space->encryption_key);
-		space->encryption_key_version = REDO_LOG_ENCRYPT_NO_VERSION;
 	} else {
 		memcpy(space->encryption_key,
 		       key, ENCRYPTION_KEY_LEN);
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -1247,7 +1247,9 @@ static SHOW_VAR innodb_status_variables[]= {
   {"num_pages_decrypted",
    (char*) &export_vars.innodb_pages_decrypted,
    SHOW_LONGLONG, SHOW_SCOPE_GLOBAL },  
-  {NullS, NullS, SHOW_LONG, SHOW_SCOPE_GLOBAL}
+  {"encryption_redo_key_version",
+  (char*) &export_vars.innodb_redo_key_version, SHOW_LONGLONG, SHOW_SCOPE_GLOBAL},
+   {NullS, NullS, SHOW_LONG, SHOW_SCOPE_GLOBAL}
 };
 
 /************************************************************************//**
@@ -3930,7 +3932,7 @@ innobase_fix_tablespaces_empty_uuid()
 		if(srv_enable_redo_encryption()) {
 			srv_redo_log_encrypt = REDO_LOG_ENCRYPT_OFF;
 		} else {
-			redo_rotate_default_key();
+			log_rotate_default_key();
 		}
 	}
 
@@ -3973,7 +3975,7 @@ innobase_fix_tablespaces_empty_uuid()
 	if (srv_enable_redo_encryption()) {
 		srv_redo_log_encrypt = REDO_LOG_ENCRYPT_OFF;
 	} else {
-		redo_rotate_default_key();
+		log_rotate_default_key();
 	}
 
 	/** Check if sys, temp need rotation to fix the empty uuid */
--- a/storage/innobase/include/fil0crypt.h
+++ b/storage/innobase/include/fil0crypt.h
@@ -612,6 +612,58 @@ fil_space_get_scrub_status(
 	const fil_space_t*		space,
 	fil_space_scrub_status_t*	status);
 
+struct redo_log_key {
+	uint  version;
+	char  key[ENCRYPTION_KEY_LEN];
+	ulint read_count;
+	ulint write_count;
+	bool  present;
+
+	bool
+	persisted() const {
+		return version != 0;
+	}
+};
+
+/** Handles the fetching/generation/storing/etc of keyring redo log keys.
+
+ This class is *NOT* thread safe, as thread safety is not required.
+ Data is only accessed/modified on the following points:
+ * When the redo space is created, at startup
+ * During redo log recovery, at startup
+ * When the server UUID is generated, at startup
+ * When the user requests a new key version, checked periodically in the
+   master thread
+
+ As these can't happen in parallel, no lock is used. */
+class redo_log_keys {
+       public:
+	MY_NODISCARD redo_log_key*
+	load_latest_key(bool generate);
+	MY_NODISCARD redo_log_key*
+	load_key_version(uint version);
+
+	MY_NODISCARD redo_log_key*
+	generate_and_store_new_key();
+
+	/* These two methods are used during bootstrap encryption,
+	when wo do not yet have an uuid */
+	MY_NODISCARD redo_log_key*
+	generate_new_key_without_storing();
+	MY_NODISCARD bool
+	store_used_keys();
+
+	void
+	unload_old_keys();
+
+       private:
+	typedef std::map<ulint, redo_log_key> key_map;
+	typedef key_map::iterator	     key_iterator;
+	key_map				      m_keys;
+};
+
+extern redo_log_keys redo_log_key_mgr;
+
 //#include "fil0crypt.ic"
 #endif /* !UNIV_INNOCHECKSUM */
 
--- a/storage/innobase/include/fil0fil.h
+++ b/storage/innobase/include/fil0fil.h
@@ -52,6 +52,8 @@ Created 10/25/1995 Heikki Tuuri
 /** Structure containing encryption specification */
 struct fil_space_crypt_t;
 
+struct redo_log_key;
+
 #define REDO_LOG_ENCRYPT_NO_VERSION 0
 
 #ifdef UNIV_HOTBACKUP
@@ -255,6 +257,8 @@ struct fil_space_t {
 
 	/** Encrypt key version*/
 	ulint			encryption_key_version;
+	/** Only used for redo log encryption: the currently active key handle */
+	redo_log_key*           encryption_redo_key;
 
 	/** Encrypt initial vector */
         byte			encryption_iv[ENCRYPTION_KEY_LEN];
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -1114,25 +1114,26 @@ const char* log_encrypt_name(redo_log_encrypt_enum val) {
 3rd block.
 @return true if success */
 bool
-log_read_encryption()
-{
-	byte		key[ENCRYPTION_KEY_LEN];
-	byte		iv[ENCRYPTION_KEY_LEN];
+log_read_encryption() {
+	byte key[ENCRYPTION_KEY_LEN];
+	byte iv[ENCRYPTION_KEY_LEN];
 
-	byte* log_block_buf_ptr = static_cast<byte*>(ut_malloc_nokey(
-		2 * OS_FILE_LOG_BLOCK_SIZE));
+	byte* log_block_buf_ptr =
+	    static_cast<byte *>(ut_malloc_nokey(2 * OS_FILE_LOG_BLOCK_SIZE));
 	memset(log_block_buf_ptr, 0, 2 * OS_FILE_LOG_BLOCK_SIZE);
-	byte* log_block_buf = static_cast<byte*>(
-		ut_align(log_block_buf_ptr, OS_FILE_LOG_BLOCK_SIZE));
+	byte* log_block_buf = static_cast<byte *>(
+	    ut_align(log_block_buf_ptr, OS_FILE_LOG_BLOCK_SIZE));
 
 	ulint		log_space_id = SRV_LOG_SPACE_FIRST_ID;
-	const page_id_t	page_id(log_space_id, 0);
+	const page_id_t page_id(log_space_id, 0);
 	fil_io(IORequestLogRead, true, page_id, univ_page_size,
 	       LOG_CHECKPOINT_1 + OS_FILE_LOG_BLOCK_SIZE,
 	       OS_FILE_LOG_BLOCK_SIZE, log_block_buf, NULL);
 
-	bool encryption_magic = false;
-	bool encrypted_log = false;
+	bool		 encryption_magic = false;
+	bool		 encrypted_log = false;
+	redo_log_key*    mkey = NULL;
+	Encryption::Type encryption_type = Encryption::NONE;
 
 	if (memcmp(log_block_buf + LOG_HEADER_CREATOR_END,
 		   ENCRYPTION_KEY_MAGIC_RK, ENCRYPTION_MAGIC_SIZE) == 0) {
@@ -1142,39 +1143,31 @@ log_read_encryption()
 		/* Make sure the keyring is loaded. */
 		if (!Encryption::check_keyring()) {
 			ut_free(log_block_buf_ptr);
-			ib::error()
-				<< "Redo log was encrypted,"
-				<< " but keyring plugin is not loaded.";
+			ib::error() << "Redo log was encrypted,"
+				    << " but keyring plugin is not loaded.";
 			return(false);
 		}
 
-		unsigned char* info_ptr = log_block_buf + LOG_HEADER_CREATOR_END + ENCRYPTION_MAGIC_SIZE;
-	        uint version = mach_read_from_4(info_ptr);
+		unsigned char* info_ptr = log_block_buf +
+					  LOG_HEADER_CREATOR_END +
+					  ENCRYPTION_MAGIC_SIZE;
+		uint version = mach_read_from_4(info_ptr);
 
-		memcpy(iv, info_ptr + ENCRYPTION_SERVER_UUID_LEN + 4, ENCRYPTION_KEY_LEN);
+		memcpy(iv, info_ptr + ENCRYPTION_SERVER_UUID_LEN + 4,
+		       ENCRYPTION_KEY_LEN);
 
 #ifdef UNIV_ENCRYPT_DEBUG
-			fprintf(stderr, "Using redo log encryption key version: %u\n", version);
+		fprintf(stderr, "Using redo log encryption key version: %u\n",
+			version);
 #endif
-		
-		char *key_type = NULL;
-		char* rkey = NULL;
-		std::ostringstream percona_redo_with_ver_ss;
-		percona_redo_with_ver_ss << PERCONA_REDO_KEY_NAME << ':' << version;
-		size_t klen;
-		if (my_key_fetch(percona_redo_with_ver_ss.str().c_str(), &key_type, NULL,
-					reinterpret_cast<void**>(&rkey), &klen) ||
-				rkey == NULL)
-		{
-			ib::error() << "Couldn't fetch redo log encryption key: " << percona_redo_with_ver_ss.str() << ".";
-		} else if(key_type == NULL || strncmp(key_type, "AES", 3) != 0) {
-			ib::error() << "Unknown redo log encryption type: " << key_type << ".";
-		} else {
+
+		mkey = redo_log_key_mgr.load_key_version(version);
+		if (mkey != NULL) {
 			encrypted_log = true;
-			memcpy(key, rkey, ENCRYPTION_KEY_LEN);
+			memcpy(key, mkey->key, ENCRYPTION_KEY_LEN);
+			encryption_type = Encryption::KEYRING;
+			srv_redo_log_key_version = mkey->version;
 		}
-		my_free(key_type);
-		my_free(rkey);
 	}
 
 	if (memcmp(log_block_buf + LOG_HEADER_CREATOR_END,
@@ -1184,15 +1177,14 @@ log_read_encryption()
 
 		/* Make sure the keyring is loaded. */
 		if (!Encryption::check_keyring()) {
-			ib::error()
-				<< "Redo log was encrypted,"
-				<< " but keyring plugin is not loaded.";
+			ib::error() << "Redo log was encrypted,"
+				    << " but keyring plugin is not loaded.";
 		} else if (Encryption::decode_encryption_info(
-				key, iv,
-				log_block_buf + LOG_HEADER_CREATOR_END)) {
+			       key, iv,
+			       log_block_buf + LOG_HEADER_CREATOR_END)) {
 			encrypted_log = true;
+			encryption_type = Encryption::AES;
 		}
-
 	}
 	if (encrypted_log) {
 		if (existing_redo_encryption_mode != srv_redo_log_encrypt &&
@@ -1215,30 +1207,29 @@ log_read_encryption()
 		   space flag. Otherwise, we just fill the encryption
 		   information to space object for decrypting old
 		   redo log blocks. */
-		fil_space_t*	space = fil_space_get(log_space_id);
+		fil_space_t* space = fil_space_get(log_space_id);
+		space->encryption_redo_key = mkey;
 		space->flags |= FSP_FLAGS_MASK_ENCRYPTION;
-		dberr_t err = fil_set_encryption(space->id,
-				Encryption::AES,
-				key, iv);
+		dberr_t err =
+		    fil_set_encryption(space->id, encryption_type, key, iv);
 
 		if (err == DB_SUCCESS) {
 			ut_free(log_block_buf_ptr);
 			ib::info() << "Read redo log encryption"
-				<< " metadata successful.";
+				   << " metadata successful.";
 			return(true);
 		} else {
 			ut_free(log_block_buf_ptr);
-			ib::error()
-				<< "Can't set redo log tablespace"
-				<< " encryption metadata.";
+			ib::error() << "Can't set redo log tablespace"
+				    << " encryption metadata.";
 			return(false);
 		}
 	} else if (encryption_magic) {
 		ut_free(log_block_buf_ptr);
 		ib::error() << "Cannot read the encryption"
-			" information in log file header, please"
-			" check if keyring plugin loaded and"
-			" the key file exists.";
+			       " information in log file header, please"
+			       " check if keyring plugin loaded and"
+			       " the key file exists.";
 		return(false);
 	}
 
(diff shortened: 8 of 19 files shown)


mysql-test/include/log_encrypt_2.inc
```diff
@@ -84,6 +91,13 @@ INSERT INTO t1 select * from t1;
 --source include/kill_and_restart_mysqld.inc
 
 SELECT * FROM t1 LIMIT 10;
+
+let $restart_parameters = restart: $KEYRING_PARAMS --general-log --log-output=FILE --general_log_file=$MYSQL_TMP_DIR/keyring_query_log;
+--replace_result $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT --plugin-dir=KEYRING_PLUGIN_PATH
+--replace_regex /\.dll/.so/
+--source include/restart_mysqld_no_echo.inc
+
+INSERT INTO t1 select * from t1;
```
laurynas-biveinis: Consistent case

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
 
 	return(encrypted);
 }
+
+redo_log_key* redo_log_keys::load_latest_key() {
+	size_t   klen = 0;
+	size_t   klen2 = 0;
```
laurynas-biveinis: Declare at use (klen2, rkey2, others as needed)

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+	uint    version = 0;
+	char *key_type = NULL;
+	byte *rkey = NULL;
+	byte *rkey2 = NULL;
+
+	if (my_key_fetch(PERCONA_REDO_KEY_NAME, &key_type, NULL,
+		reinterpret_cast<void**>(&rkey), &klen) || rkey == NULL ||
+		strncmp(key_type, "AES", 3) != 0) {
```
laurynas-biveinis: If only 3 chars are compared, this does prefix comparison only?
```suggestion
		strncmp(key_type, "AES", 4) != 0) {
```

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+
+
+	const bool err = (parse_system_key(rkey, klen, &version, &rkey2, &klen2) == reinterpret_cast<uchar*>(NullS));
+	if (err) {
+		my_free(rkey);
+		my_free(rkey2);
+		my_free(key_type);
+	}
+
+	ut_ad(klen2 ==  ENCRYPTION_KEY_LEN);
```
laurynas-biveinis:
```suggestion
	ut_ad(klen2 == ENCRYPTION_KEY_LEN);
```

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+
+	{
```
laurynas-biveinis: Why extra block? Is key_iterator very expensive to have around until outer scope end?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+
+	my_free(rkey);
+	my_free(rkey2);
+	my_free(key_type);
+
+	return rk;
+}
+
+redo_log_key* redo_log_keys::load_key_version(uint version) {
+	{
```
laurynas-biveinis: Same question re. extra block

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+
+	size_t klen = 0;
+	char *key_type = NULL;
+	byte *rkey = NULL;
+
+	std::ostringstream percona_redo_with_ver_ss;
+	percona_redo_with_ver_ss << PERCONA_REDO_KEY_NAME << ':' << version;
+	if (my_key_fetch(percona_redo_with_ver_ss.str().c_str(), &key_type, NULL,
+		reinterpret_cast<void**>(&rkey), &klen) || rkey == NULL ||
+		strncmp(key_type, "AES", 3) != 0) {
```
laurynas-biveinis:
```suggestion
		strncmp(key_type, "AES", 4) != 0) {
```

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+		my_free(rkey);
+		my_free(key_type);
+		ib::error() << "Failed to load redo key version " << version;
+		return NULL;
+	}
+
+	ut_ad(klen ==  ENCRYPTION_KEY_LEN);
```
laurynas-biveinis:
```suggestion
	ut_ad(klen == ENCRYPTION_KEY_LEN);
```

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+	memcpy(rk->key, rkey2, ENCRYPTION_KEY_LEN);
+
+	my_free(rkey);
+	my_free(rkey2);
+	my_free(key_type);
+
+	return rk;
+}
+
+redo_log_key* redo_log_keys::load_key_version(uint version) {
```
laurynas-biveinis: It looks like it should be possible to extract common code from these two functions

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+	memcpy(rk->key, rkey, ENCRYPTION_KEY_LEN);
+
+	my_free(rkey);
+	my_free(key_type);
+
+	return rk;
+}
+
+redo_log_key* redo_log_keys::generate_and_store_new_key() {
+        char *redo_key_type = NULL;
```
laurynas-biveinis: Declare at first use

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+
+	my_free(redo_key_type);
+	my_free(rkey);
+	my_free(rkey2);
+
+	return rk;
+}
+
+redo_log_key* redo_log_keys::generate_new_key_without_storing() {
+        ut_ad(m_keys.empty());
```
laurynas-biveinis: Formatting

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+	Encryption::random_value(reinterpret_cast<byte*>(&m_keys[0].key));
+	return &m_keys[0];
+}
+
+bool redo_log_keys::store_used_keys() {
+    /* This is a for loop, but it really only should store a key with current version 0 */
+    for(key_iterator it = m_keys.begin();it != m_keys.end(); ++it) {
```
laurynas-biveinis: Formatting

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+	    if(!it->second.persisted()) {
+                    ut_ad(it->first == 0);
```
laurynas-biveinis: Formatting

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+}
+
+redo_log_key* redo_log_keys::generate_new_key_without_storing() {
+        ut_ad(m_keys.empty());
+	Encryption::random_value(reinterpret_cast<byte*>(&m_keys[0].key));
+	return &m_keys[0];
+}
+
+bool redo_log_keys::store_used_keys() {
+    /* This is a for loop, but it really only should store a key with current version 0 */
```
laurynas-biveinis: Formatting

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+		    {
+			    return false;
+		    }
+	    }
+    }
+
+    return true;
+}
+
+void redo_log_keys::unload_old_keys() {
```
laurynas-biveinis: Formatting

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+    if(m_keys.size() == 0) {
+	    return;
+    }
+    redo_log_key* last = &(m_keys.end()--)->second;
```
laurynas-biveinis: Use prefix --

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+    for(key_iterator it = m_keys.begin();it != m_keys.end(); ++it) {
```
laurynas-biveinis: Formatting

storage/innobase/fil/fil0fil.cc
```diff
@@ -5933,17 +5933,25 @@ fil_io_set_encryption(
 		 && (TRX_SYS_SPACE != page_id.space()
 		     || TRX_SYS_PAGE_NO != page_id.page_no())))
 	    && space->encryption_type != Encryption::NONE) {
-		if (space->encryption_type == Encryption::KEYRING)  {
+		if (space->encryption_type == Encryption::KEYRING
+		    && page_id.space() != SRV_LOG_SPACE_FIRST_ID)  {
```
laurynas-biveinis:
```suggestion
		    && page_id.space() != SRV_LOG_SPACE_FIRST_ID) {
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1247,6 +1247,9 @@ static SHOW_VAR innodb_status_variables[]= {
   {"num_pages_decrypted",
    (char*) &export_vars.innodb_pages_decrypted,
    SHOW_LONGLONG, SHOW_SCOPE_GLOBAL },  
+  {"encryption_redo_key_version",
+   (char*) &export_vars.innodb_redo_key_version,
+   SHOW_LONGLONG, SHOW_SCOPE_GLOBAL },
```
laurynas-biveinis: Formatting

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,41 @@ fil_space_get_scrub_status(
...
 	fil_space_scrub_status_t*	status);
 
+/** Handles the fetching/generation/storing/etc of keyring redo log keys.
+ 
+ This class is *NOT* thread safe, as thread safety is not required.
+ Data is only accessed/modified on the following points:
+ * When the redo space is created, at startup
+ * During redo log recovery, at startup
+ * When the server UUID is generated, at startup
+ * When the user requests a new key version, in the setter
```
laurynas-biveinis: Why it is OK to be thread unsafe, if two users can request new key versions in parallel?

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,41 @@ fil_space_get_scrub_status(
...
+ This class is *NOT* thread safe, as thread safety is not required.
+ Data is only accessed/modified on the following points:
+ * When the redo space is created, at startup
+ * During redo log recovery, at startup
+ * When the server UUID is generated, at startup
+ * When the user requests a new key version, in the setter 
+ 
+ As these can't happen in parallel, no lock is used.
+ */
+class redo_log_keys {
```
laurynas-biveinis, line 639: Formatting

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,41 @@ fil_space_get_scrub_status(
...
+	public:
+
+		redo_log_key* load_latest_key();
```
laurynas-biveinis: MY_NODISCARD all the non-void methods

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,41 @@ fil_space_get_scrub_status(
...
+
+		/** These two methods are used during bootstrap encryption,
+		 when wo do not yet have an uuid */
+		redo_log_key* generate_new_key_without_storing();
+		bool store_used_keys();
+
+		void unload_old_keys();
+
+	private:
+		std::map<ulint, redo_log_key> m_keys;
```
laurynas-biveinis: typedef std::map too, use it in the 2nd typedef

storage/innobase/os/os0file.cc
```diff
@@ -10197,6 +10201,13 @@ Encryption::encrypt_log_block(
 	/* Set the encrypted flag. */
 	log_block_set_encrypt_bit(dst_ptr, true);
 
+	if(m_type == Encryption::KEYRING) {
```
laurynas-biveinis: Formatting

storage/innobase/os/os0file.cc
```diff
@@ -10197,6 +10201,13 @@ Encryption::encrypt_log_block(
...
+		const ulint crc = log_block_calc_checksum_crc32(dst_ptr);
+		log_block_set_checksum(dst_ptr, crc + m_key_version);
+		fprintf(stderr, "Encrypted with version %u\n", m_key_version);
```
laurynas-biveinis: Huh?

storage/innobase/os/os0file.cc
```diff
@@ -10601,16 +10612,38 @@ dberr_t
...
 	byte*			src,
-	byte*			dst)
+	byte*			dst,
+	ulint                   offset)
 {
 	byte		remain_buf[MY_AES_BLOCK_SIZE * 2];
 
-	const ulint data_len = OS_FILE_LOG_BLOCK_SIZE - LOG_BLOCK_HDR_SIZE;
+	const ulint unencrypted_trailer_size =
+		(m_type == Encryption::KEYRING) ?  LOG_BLOCK_TRL_SIZE : 0;
```
laurynas-biveinis: Formatting

storage/innobase/os/os0file.cc
```diff
@@ -10601,16 +10612,38 @@ dberr_t
...
 	byte*		ptr = src + LOG_BLOCK_HDR_SIZE;
 
 	switch (m_type) {
+	case Encryption::KEYRING:
+		{
+			const ulint block_crc = log_block_calc_checksum_crc32(src);
+			const ulint written_crc = log_block_get_checksum(src);
+
+			const ulint enc_key_version = written_crc - block_crc;
+				fprintf(stderr, "Using key version: %lu\n", enc_key_version);
```
laurynas-biveinis: Huh? And some more fprintf(stderr) below

storage/innobase/os/os0file.cc
```diff
@@ -10667,6 +10700,7 @@ Encryption::decrypt_log_block(
 		/* Copy the remaining bytes. */
 		memcpy(ptr + main_len, dst + main_len, data_len - main_len);
 
+
```
laurynas-biveinis: Remove spurious diff

storage/innobase/os/os0file.cc
```diff
@@ -10689,6 +10723,12 @@ Encryption::decrypt_log_block(
 	/* Reset the encrypted flag. */
 	log_block_set_encrypt_bit(ptr, false);
 
+	/* recalculate the checksum */
```
laurynas-biveinis: This comment is obvious
If there has to be a comment here, better to explain, why the checksum has to be recalculated at this point

laurynas-biveinis (changes requested):
Please use clang-format
Please also pass the commit message through a spell checker

storage/innobase/fil/fil0crypt.cc
dutow: It isn't I only added these to limit the scope of the variable. I'll remove it.

storage/innobase/fil/fil0crypt.cc
dutow: Uh, and I only copied this check from elsewhere in the original redo encryption commit, so it's incorrect at other places too.

storage/innobase/include/fil0crypt.h
dutow: New key versions are added only in the master thread, which checks it periodically.

storage/innobase/os/os0file.cc
dutow: Uh, sorry about this and the formatting issues - I removed this and fixed most formatting problems based on the github diff, but looks like I forgot to push the changes or simply didn't add force...

storage/innobase/fil/fil0crypt.cc
dutow: I don't think I could extract too many parts - a few lines at most, otherwise, there are small differences everywhere. And that would make the code less readable.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+redo_log_key* redo_log_keys::load_latest_key() {
+	size_t klen = 0;
+	char *key_type = NULL;
+	byte *rkey = NULL;
+
+	if (my_key_fetch(PERCONA_REDO_KEY_NAME, &key_type, NULL,
+		reinterpret_cast<void**>(&rkey), &klen) || rkey == NULL ||
+		strncmp(key_type, "AES", 4) != 0) {
+
+                /* There is no key yet, we'll try to generate one */
```
laurynas-biveinis: Still needs clang-format throughout the diff

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+
+	my_free(redo_key_type);
+	my_free(rkey);
+	my_free(rkey2);
+
+	return rk;
+}
+
+redo_log_key* redo_log_keys::generate_new_key_without_storing() {
+	ut_ad(m_keys.empty());
```
laurynas-biveinis, line 3288: Are we using an out-of-bounds map element in this method?

storage/innobase/include/fil0crypt.h
laurynas-biveinis: Please clarify this in the comment

laurynas-biveinis (changes requested):

storage/innobase/fil/fil0crypt.cc
dutow: Uh, this is for 5.7, but I'll go through it once more.

storage/innobase/fil/fil0crypt.cc
dutow: Yes, but operator[] calls the default constructor for non existing elements, so it works as it should. The key field is intialiezed here, and the other fields in `store_used_keys`.

laurynas-biveinis: Of course, sorry

storage/innobase/fil/fil0crypt.cc
dutow: Fixed.

storage/innobase/include/fil0crypt.h
dutow: Done

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,169 @@ fil_space_verify_crypt_checksum(
...
+		reinterpret_cast<void**>(&rkey), &klen) || rkey == NULL ||
+		strncmp(key_type, "AES", 4) != 0) {
+		
+		/* There is no key yet, we'll try to generate one */
+		my_free(rkey);
+		return generate_and_store_new_key();
+	}
+
+
+	uint    version = 0;
```
laurynas-biveinis: Formatting. I asked for specifically clang-format at the first review already, there are simply too many things to fix manually. 5.7 has .clang-format. If the problem is that the rest of fil0crypt.cc is unformatted, surround with "noformat" comments before clang-format invocation as needed

storage/innobase/os/os0file.cc
```diff
@@ -10601,16 +10611,32 @@ dberr_t
...
 {
 	byte		remain_buf[MY_AES_BLOCK_SIZE * 2];
 
-	const ulint data_len = OS_FILE_LOG_BLOCK_SIZE - LOG_BLOCK_HDR_SIZE;
+	const ulint unencrypted_trailer_size =(m_type == Encryption::KEYRING) ?  LOG_BLOCK_TRL_SIZE : 0;
+	const ulint data_len = OS_FILE_LOG_BLOCK_SIZE - LOG_BLOCK_HDR_SIZE - unencrypted_trailer_size;
 	const ulint main_len = (data_len / MY_AES_BLOCK_SIZE) * MY_AES_BLOCK_SIZE;
 	ulint remain_len = data_len - main_len;
-	byte*		ptr = src + LOG_BLOCK_HDR_SIZE;
+	byte* ptr = src + LOG_BLOCK_HDR_SIZE;
```
laurynas-biveinis, line 10608: Spurious diff

laurynas-biveinis (changes requested):

storage/innobase/fil/fil0crypt.cc
dutow: There is a clang format, but it doesn't match the code exactly. The clang-format pre-commit hook script results in a 73k line diff file for this patch, and 99% of that isn't my change.

For encryption related parts, it's even worse than usual.

laurynas-biveinis: Right, because all of fil0crypt is a dumpster fire as far as formatting is concerned, and we will have to commit that 73K diff sooner or later.

That's why I suggested noformat as a workaround

dutow: I'm trying to do that now, but it's not only that. Clang-format would reformat huge parts of most innodb files I touched (e.g. 2k line diff for os0file.h), and for some files (such as the system_key header) would completely rearrange the code.

dutow: I've reformatted everything redo log encryption related with clang-format. I left a few things (such as system_key) alone. I think clang-format also formats class/struct declarations incorrectly for innodb, but I didn't change that.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1247,6 +1247,8 @@ static SHOW_VAR innodb_status_variables[]= {
   {"num_pages_decrypted",
    (char*) &export_vars.innodb_pages_decrypted,
    SHOW_LONGLONG, SHOW_SCOPE_GLOBAL },  
+  {"encryption_redo_key_version",
+  (char*) &export_vars.innodb_redo_key_version, SHOW_LONGLONG, SHOW_SCOPE_GLOBAL},
```
laurynas-biveinis, line 1251:
```suggestion
   (char*) &export_vars.innodb_redo_key_version, SHOW_LONGLONG, SHOW_SCOPE_GLOBAL},
```

storage/innobase/os/os0file.cc
```diff
@@ -10128,104 +10128,106 @@ Encryption::encrypt_log_block(
...
-		       - LOG_BLOCK_HDR_SIZE - len);
+			/* Copy remaining bytes. */
+			memcpy(dst_ptr + LOG_BLOCK_HDR_SIZE + len,
+			       src_ptr + LOG_BLOCK_HDR_SIZE + len,
+			       OS_FILE_LOG_BLOCK_SIZE - LOG_BLOCK_HDR_SIZE -
+				   len);
 
-		/* Encrypt the remaining bytes. Since my_aes_encrypt
+			/* Encrypt the remaining bytes. Since my_aes_encrypt
 		request the content which need to encrypt is
```
laurynas-biveinis: Should this be indented too?

storage/innobase/os/os0file.cc
```diff
@@ -10598,98 +10594,118 @@ Encryption::encrypt(
...
+			    enc_key_version != REDO_LOG_ENCRYPT_NO_VERSION) {
+				redo_log_key *mkey =
+				    redo_log_key_mgr.load_key_version(
+					enc_key_version);
+				m_key_version = mkey->version;
+				m_key = reinterpret_cast<unsigned char *>(
+				    mkey->key);
+			}
+		}
+			/* FALLTHROUGH */
```
laurynas-biveinis: Should this be unindented?

laurynas-biveinis (approved):
Some minor formatting questions but LGTM with or without them addressed as you see fit. No need for further reviews from me

satya-bodapati:
few questions on design (before looking into implementation)

1. is theren't sufficient space in redo log header to write key_version?
2. When you OR the checksum with key_version, will  older version of xtrabackup fail to read this redo log?
3. Is key rotation applicable on redo log? i.e with innodb_rotate_encrypt_age work on redo log also?

dutow:
1 - there's some space in the log header, but only for a few keys. If the user would call rotate too many times, we would run out of space to store versions. So that's not a good option.
2 - Yes. But it would anyway, as it doesn't support this encryption schema.
3 - no. Key versioning only means that we encrypt new blocks with new versions, we never actually rotate old blocks. That can be added in the future if needed, but old redo blocks shouldn't live for long anyway. And if a user wants to remove uses of old keys log scrubbing could make more sense, to simply delete no longer needed log blocks.

dutow:
@satya-bodapati I opened the 8.0 version of this PR, that's easier to review - it only contains logic changes, no reformatting. The effective code change is the same.

satya-bodapati:
> 1 - there's some space in the log header, but only for a few keys. If the user would call rotate too many times, we would run out of space to store versions. So that's not a good option.

how many bytes used for tablespace page key versions?

> 2 - Yes. But it would anyway, as it doesn't support this encryption schema.

Aha, xtrabackup doesn't work with keyring_key encrypted tablespaces. So safe :D 

> 3 - no. Key versioning only means that we encrypt new blocks with new versions, we never actually rotate old blocks. That can be added in the future if needed, but old redo blocks shouldn't live for long anyway. And if a user wants to remove uses of old keys log scrubbing could make more sense, to simply delete no longer needed log blocks.

Ok, got it

dutow:
1 - if we would store the versions in the log header, we would have to store the version and the start lsn at minimum, so that's 12 / version. The longest free continuous block we have in the log header for encryption is 36 bytes, so that would be enough for 3 versions.

2 - I'm not saying it won't work in the future. But for that, it needs the code to decrypt the redo log first.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,176 @@ fil_space_verify_crypt_checksum(
 
 	return(encrypted);
 }
+
+redo_log_key *
```
satya-bodapati:
```suggestion
redo_log_key*
```

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,176 @@ fil_space_verify_crypt_checksum(
...
+redo_log_keys::load_latest_key() {
+	size_t klen = 0;
+	char * key_type = NULL;
```
satya-bodapati: Please fix this  space and pointer everywhere. I think this style is also not allowed in clang-format ;)

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,176 @@ fil_space_verify_crypt_checksum(
...
+
+	ut_ad(klen2 == ENCRYPTION_KEY_LEN);
+
+	key_iterator it = m_keys.find(version);
+
+	if (it != m_keys.end() && it->second.present) {
+		return &it->second;
+	}
+
+	redo_log_key *rk = &m_keys[version];
```
satya-bodapati: This is server style ;)

mysys/system_key.c
```diff
@@ -32,7 +32,7 @@ struct Valid_percona_system_key
 */
 struct Valid_percona_system_key valid_percona_system_keys[] = { {PERCONA_BINLOG_KEY_NAME, 16, FALSE},
                                                                 {PERCONA_INNODB_KEY_NAME, 32, TRUE},
-								{PERCONA_REDO_KEY_NAME, 16, FALSE}};
+								{PERCONA_REDO_KEY_NAME, 32, FALSE}};
```
satya-bodapati, line 35: Why the length change now?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,176 @@ fil_space_verify_crypt_checksum(
...
+	size_t klen = 0;
+	char * key_type = NULL;
+	byte * rkey = NULL;
+
+	if (my_key_fetch(PERCONA_REDO_KEY_NAME, &key_type, NULL,
+			 reinterpret_cast<void **>(&rkey), &klen) ||
+	    rkey == NULL || strncmp(key_type, "AES", 4) != 0) {
+		/* There is no key yet, we'll try to generate one */
+		my_free(rkey);
+		return generate_and_store_new_key();
```
satya-bodapati: Please return(x) format everywhere

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,52 @@ fil_space_get_scrub_status(
...
+ * During redo log recovery, at startup
+ * When the server UUID is generated, at startup
+ * When the user requests a new key version, checked periodically in the
+   master thread
+ 
+ As these can't happen in parallel, no lock is used.
+ */
+class redo_log_keys {
+       public:
+	MY_NODISCARD
```
satya-bodapati: Does it have be seriously on three lines? One line is better IMHO

Applicable to all lines below

storage/innobase/include/fil0fil.h
```diff
@@ -52,6 +52,19 @@ Created 10/25/1995 Heikki Tuuri
 /** Structure containing encryption specification */
 struct fil_space_crypt_t;
 
+struct redo_log_key {
```
satya-bodapati: This doesn't seem right. Redo log key defined in fil0fil.h.

Please put it in log0log.h

satya-bodapati (changes requested):

mysys/system_key.c
dutow: 16 was incorrect (it worked, but the second part of the key was always empty), but I didn't notice it before.

storage/innobase/fil/fil0crypt.cc
dutow: This is how clang format formats it... (for every pointer style comment)

satya-bodapati: Zsolt, this doesn't seem right. It is OK atleast in 8.0.
Do you know why it does differently in 5.7? Am not sure if we should clang-format for innodb codebase. It looks inconsistent. Whats your opinion @laurynas-biveinis ?

dutow: In 8.0, every file uses the same clang format file, and the entire source code was reformatted.

For 5.7, @inikep tried to create clang format configs for both the base source and innodb that mimics how it looks like.

The issue with the pointer symbols is that it first aligns the pointer symbols to the variable name, and then it alings the variable names - but leaves the pointer symbols where they were.

mysys/system_key.c
satya-bodapati: so it is wrong for BINLOG_KEY_NAME too?

Is my understanding correct. This length 32 is about the total length "keyname:key_version". This is  "string:number"?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,176 @@ fil_space_verify_crypt_checksum(
...
+		my_free(key_type);
+		return NULL;
+	}
+
+	ut_ad(klen2 == ENCRYPTION_KEY_LEN);
+
+	key_iterator it = m_keys.find(version);
+
+	if (it != m_keys.end() && it->second.present) {
+		return &it->second;
```
satya-bodapati: Can we assert here? 
What we found in the cache should be equal to the one  from keyring

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,176 @@ fil_space_verify_crypt_checksum(
...
+
+	return true;
+}
+
+void
+redo_log_keys::unload_old_keys() {
+	if (m_keys.size() == 0) {
+		return;
+	}
+	redo_log_key *last = &(--m_keys.end())->second;
```
satya-bodapati: So we assume last is the latest? Can you add version check? if last key version is always greater than the old ones.

storage/innobase/fil/fil0fil.cc
```diff
@@ -5928,22 +5928,32 @@ fil_io_set_encryption(
...
-		 && (TRX_SYS_SPACE != page_id.space()
-		     || TRX_SYS_PAGE_NO != page_id.page_no())))
-	    && space->encryption_type != Encryption::NONE) {
-		if (space->encryption_type == Encryption::KEYRING)  {
+	if ((req_type.is_log() || (page_id.page_no() > 0 &&
+				   (TRX_SYS_SPACE != page_id.space() ||
+				    TRX_SYS_PAGE_NO != page_id.page_no()))) &&
+	    space->encryption_type != Encryption::NONE) {
+		if (space->encryption_type == Encryption::KEYRING &&
+		    page_id.space() != SRV_LOG_SPACE_FIRST_ID) {
```
satya-bodapati: Why this check on SRV_LOG_SPACE_FIRST_ID? What does it mean?

satya-bodapati: IMHO, this can be made simpler. The outer check has to be first about space->encryption_type. 

8.0 version looks better. Can you improve (make similar to that) in 5.7 too?

storage/innobase/fil/fil0fil.cc
```diff
@@ -5928,22 +5928,32 @@ fil_io_set_encryption(
...
+			fil_io_set_keyring_encryption(req_type, space,
+						      page_id);
 		} else {
-			ut_ad(space->encryption_type == Encryption::AES);
-			req_type.encryption_key(space->encryption_key,
-						32,
-						false,
-						space->encryption_iv,
-						0, 0, NULL, NULL); // not relevant for Master Key encryption
+			ut_ad(space->encryption_type == Encryption::AES ||
```
satya-bodapati: This piece of code should ideally be fil_io_set_mk_encryption() ?

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,52 @@ fil_space_get_scrub_status(
...
 	fil_space_scrub_status_t*	status);
 
+/** Handles the fetching/generation/storing/etc of keyring redo log keys.
+ 
+ This class is *NOT* thread safe, as thread safety is not required.
+ Data is only accessed/modified on the following points:
+ * When the redo space is created, at startup
+ * During redo log recovery, at startup
+ * When the server UUID is generated, at startup
+ * When the user requests a new key version, checked periodically in the
```
satya-bodapati, line 635: You mean rotate("percona_redo"). Which background thread is accessing this?
Aren't there two threads that can access this vector?

1. User thread that does rotation of key and inserts into vector
2. Background thread that reads the vector

storage/innobase/log/log0log.cc
```diff
@@ -1351,82 +1321,69 @@ redo log file header.
...
-			}
-			if (rkey2 != NULL) {
-				my_free(rkey2);
-			}
-			if (redo_key_type != NULL) {
-				my_free(redo_key_type);
-			}
+	if (space->encryption_type != Encryption::NONE &&
+	    Encryption::master_key_id == ENCRYPTION_DEFAULT_MASTER_KEY_ID &&
+	    !srv_read_only_mode && strlen(server_uuid) > 0 &&
```
satya-bodapati: I think this code is no longer called from background thread (different from upstream).
Can you remove the uuid check and assert it is  > 0.

storage/innobase/srv/srv0srv.cc
```diff
@@ -3585,54 +3589,18 @@ srv_enable_redo_encryption_rk()
...
-				return true;
-			} else {
-				memcpy(key, rkey2, ENCRYPTION_KEY_LEN);
-			}
-			my_free(rkey2);
-		}
-	} else {
-		memcpy(key, rkey, ENCRYPTION_KEY_LEN);
+        redo_log_key* mkey = redo_log_key_mgr.load_latest_key();
+	if(mkey == NULL) {
```
satya-bodapati: if (mkey ==

storage/innobase/srv/srv0srv.cc
```diff
@@ -2881,6 +2883,8 @@ DECLARE_THREAD(srv_master_thread)(
 		}
 
 		srv_enable_undo_encryption_if_set();
+
+		log_check_new_key_version();
```
satya-bodapati, line 2887: Ok, master thread is used for this. For regular tabespace, to do this version checking, encryption threads are used right?

Why is this not part of encryption threads?

storage/innobase/srv/srv0start.cc
```diff
@@ -465,25 +465,33 @@ create_log_files(
...
+
+		Encryption::Type alg =
+		    srv_redo_log_encrypt == REDO_LOG_ENCRYPT_RK
+			? Encryption::KEYRING
+			: Encryption::AES;
+
+		log_space->flags |= FSP_FLAGS_MASK_ENCRYPTION;
+
+		redo_log_key *mkey =
+		    redo_log_key_mgr.generate_new_key_without_storing();
```
satya-bodapati: Whats the idea of "non-stored" version? I mean the advantage of using this variant?

 It seems always the first element in vector is used for this. When will this be stored?

satya-bodapati (changes requested):
Check the entire patch.. so thats good news.. next review rounds should be faster.

mysys/system_key.c
dutow: I don't know why binlog uses 16, and 32 is the actual length of the key.

storage/innobase/fil/fil0fil.cc
dutow: SRV_LOG_SPACE_FIRST_ID is the id of the redo log space, and it has to be handled differently than other keyring encrypted tablespaces.

storage/innobase/include/fil0crypt.h
dutow: * We check the latest version (read, possibly insert) periodically in the master_thread
* We possibly load earlier versions during recovery, but then the master thread isn't running yet
* We load an initial version at startup, the master thread is not yet running
* We insert a first entry in the update function in a user thread (this is missing from the comment, I'll add), but then the log is not yet encrypted, so the master thread doesn't access the vector yet

storage/innobase/srv/srv0srv.cc
dutow: Because encryption threads do more and differently: those also reencrypt earlier records with the new key.

With redo, that's not needed, and this way it works without encryption threads on.

storage/innobase/srv/srv0start.cc
dutow: During bootstrap with redo encryption, we have to encrypt the log before we can communicate with the keyring plugin. This is why we generate the first key directly and store later when the keyring is guaranteed to be available.

storage/innobase/include/fil0fil.h
dutow: I've put the rest of the code to fil0crypt.h, where everything related to keyring encryption is. The issue here is that fil0crypt includes fil0fil, and fil0fil defines fil_space_t, which needs a redo_log_key* data member. I could move everything to log0log, but I'm not sure it would be better than fil0crypt? Or maybe I could only forward declare it in fil0fil, I only need a pointer to it?

storage/innobase/fil/fil0crypt.cc
dutow: What assert would that be? At this point, we didn't even talk to the keyring, and we avoid is, as we have the entry in the cache. I don't think a debug build should contact the keyring one additional time only for an assertion.

storage/innobase/fil/fil0crypt.cc
dutow: This is an std::map with a default comparator, the last element will always be the largest.

mysys/system_key.c
satya-bodapati: how is 32 bytes used? 16 character for name and 15 character for number and 1char for ":"?

storage/innobase/include/fil0fil.h
satya-bodapati: just use forward declaration and put the redo_log_key code in log0log.h/.cc

mysys/system_key.c
dutow: 32 for the actual key - the name, version and the ":" is above that.

storage/innobase/fil/fil0crypt.cc
satya-bodapati: Few lines back, we have key and the version See
```
+	uint       version = 0;
+	byte *     rkey2 = NULL;
+	size_t     klen2 = 0;
+	const bool err =
+	    (parse_system_key(rkey, klen, &version, &rkey2, &klen2) ==
+	     reinterpret_cast<uchar *>(NullS));
+	if (err) {
+		my_free(rkey);
+		my_free(rkey2);
+		my_free(key_type);
+		return NULL;
```

We even parsed it. Am asking you to assert the key parsed to the one in cache.

storage/innobase/fil/fil0crypt.cc
satya-bodapati: Ok

storage/innobase/fil/fil0fil.cc
satya-bodapati: ok, I understand now but I think this needs a bit rewrite. This is made complex with keyring, now with additional requirement for redo with keyring, it is hard to understand.

Please try to make it simple. Similar request I made to robert and it was discarded as 'improvement or formatting later'. Now I feel, it is time to fix this. It is hard to understand and there can be bugs creeping in.

storage/innobase/include/fil0crypt.h
satya-bodapati: So rotate("percona_redo") doesn't lead to accessing/updating this vector?

dutow: It doesn't. rotate adds a new key to the keyring. Then a bit later the master thread notices this, and loads the new key.

storage/innobase/srv/srv0start.cc
satya-bodapati: What you wrote is a good example of a function comment for generate_new_key_without_storing(). 

If you are open to accepting suggestions on name, my suggestion is 

generate_key_for_bootstrap().

One more doubt, will this be used when redo files are recreated on startup? In that case we shouldn't use this function?

mysys/system_key.c
satya-bodapati: If we agree 16 is wrong, it is OK to change it to 32 for BINGLOG_KEY_NAME too.

storage/innobase/srv/srv0start.cc
dutow: It's called in `create_log_files()`, so we use it every time when:

* there is no existing redo log
* redo log encryption is turned on

I could add more conditions / paths, but it only would make the code more complex, the key is written to the keyring a bit later anyway.

mysys/system_key.c
dutow: But not as part of this PR, and it is a question for @robgolebiowski

storage/innobase/include/fil0crypt.h
satya-bodapati: Ok, thanks

storage/innobase/srv/srv0start.cc
satya-bodapati: Ok, lets assume the recreate case. We first create the key assuming keyring is not accessible, Will the key change once we know the keyring is available? When and how it will use a key generated by using keyring plugin?  (i.e my_key_generate())?

dutow: Change: no, we upload this key to the keyring, and keep using it.

Key generated by keyring: every other case generates a key with a keyring, which means:
* rotate_system_key
* if the redo log initially isn't encrypted, and we request encryption later, we start by using a keyring generated key

satya-bodapati: Ok, regarding the second point, what if that request is startup option?

Am trying to understand the pros and cons here.Can you please list out? 
For MK encryption too, such default key exists but that is changed once uuid is ready.

dutow: If it is a startup option, and the log isn't encrypted, but it exists? Then we simply generate a key with the keyring. The issue is only when the logs are created during bootstrap, in every other case, the normal way works.

For MK, we *never* change the key. MK generates a redo key at encryption setup, and stores the key in the log file header. This key can't possibly be changed later, as it is needed to read older encrypted records.
But this key isn't stored in plain text, as that would make no sense: MK encrypts the redo key using the innodb master key in the log file header. Every time the master key is rotated, MK reencrypts the log file header. And if MK encryption is turned on before an uuid exists, it encrypts the file header using the default key first (tbh I'm not sure why, because it would never be able to read that header, it doesn't make any sense to write it at all).

satya-bodapati: Thanks, now it is clear what MK does. so let me be specific

After using a default generated key, and assume no rotation initiated, when will redo use a  "percona_redo" key (created by my_key_generate())

dutow: By "default generated key", you mean a key generated with this without_storing method? If yes, then that key will be saved as percona_redo in the fix_empty_uuid method, so it *is* a percona_redo key. Unless the user rotates it, it won't be changed.

satya-bodapati: ok, checked the code redo_rotate_default_key(). Here we are storing the default key as percona_redo key.

So we cannot generate and new encryption key here (in fix_empty_uuid) and use it for redo tablespace (store it in fil_space_t) and write the key encryption in redo header?

dutow: This "default" key is as random as any other key, not something fixed, why would we generate a new key?

With keyring encryption, every key we used to encrypt blocks in the redo log has to be present in the keyring. Otherwise we won't be able to decrypt those blocks. Since it's possible that we encrypted blocks with the server generated key, we have to store it.

After that, we could rotate the key - but I don't see the point of it.

satya-bodapati: ok, got it. Here it doesn't make sense (yet). Incase of MK, we have UUID in tablespace encryption header. So in MK, the rotation at fix_empty_uuid,  fixes all tablespaces that wrote tablespace without UUID to headers with UUID.

storage/innobase/fil/fil0fil.cc
```diff
@@ -5933,19 +5968,14 @@ fil_io_set_encryption(
 		 && (TRX_SYS_SPACE != page_id.space()
 		     || TRX_SYS_PAGE_NO != page_id.page_no())))
 	    && space->encryption_type != Encryption::NONE) {
-		if (space->encryption_type == Encryption::KEYRING)  {
+		if (should_be_keyring_encrypted(page_id, space)) {
```
satya-bodapati: Well, I didn't mean to add wrapper. More like rewrite.
```
diff --git a/storage/innobase/fil/fil0fil.cc b/storage/innobase/fil/fil0fil.cc
index ec00dce4612..342b12d4fad 100644
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -5886,23 +5886,16 @@ fil_io_set_mk_encryption(IORequest& req_type, fil_space_t* space)
 	req_type.encryption_rotation(Encryption::NO_ROTATION);
 }
 
-static bool
-should_be_keyring_encrypted(const page_id_t& page_id, fil_space_t* space)
-{
-	if (page_id.space() == SRV_LOG_SPACE_FIRST_ID) {
-		/* Redo log always follows the MK encryption codepath,
-		   even when using keyring_key. */
-		return(false);
-	}
-
-	if (space->encryption_type == Encryption::KEYRING) {
-		return(true);
-	}
+static
+bool
+fil_keyring_skip_encryption(const page_id_t& page_id) {
 
-	return(false);
+	/* Don't encrypt TRX_SYS_SPACE.TRX_SYS_PAGE_NO as it contains address
+	to dblwr buffer */
+	return((page_id.space() == TRX_SYS_SPACE
+		&& page_id.page_no() == TRX_SYS_PAGE_NO) ? true : false);
 }
 
-
 /** Set encryption information for IORequest.
 @param[in,out]	req_type	IO request
 @param[in]	page_id		page id
@@ -5961,25 +5954,58 @@ fil_io_set_encryption(
 		return;
 	}
 
-	/* Don't encrypt the page 0 of all tablespaces,
-	   don't encrypt TRX_SYS_SPACE.TRX_SYS_PAGE_NO as it contains address to dblwr buffer */
-	if ((req_type.is_log()
-	     || (page_id.page_no() > 0
-		 && (TRX_SYS_SPACE != page_id.space()
-		     || TRX_SYS_PAGE_NO != page_id.page_no())))
-	    && space->encryption_type != Encryption::NONE) {
-		if (should_be_keyring_encrypted(page_id, space)) {
-			ut_ad(space->crypt_data != NULL);
+	if (space->encryption_type == Encryption::NONE) {
+		return;
+	}
 
-			fil_io_set_keyring_encryption(req_type, space, page_id);
-		} else {
-			ut_ad(space->encryption_type == Encryption::AES || page_id.space() == SRV_LOG_SPACE_FIRST_ID);
-			
+        if (req_type.is_log())  {
+		/* redo log encryption */
+		ut_ad(page_id.space() == SRV_LOG_SPACE_FIRST_ID);
+
+		switch (space->encryption_type) {
+		case Encryption::AES:
+		case Encryption::KEYRING:
+			// Both MK and Keyrig key use same style for redo log
 			fil_io_set_mk_encryption(req_type, space);
+		        req_type.encryption_algorithm(space->encryption_type);
+			return;
+		case Encryption::NONE:
+			// Already handled above
+		default:
+			ut_a(0);
 		}
-		req_type.encryption_algorithm(space->encryption_type);
+
 	} else {
-		req_type.clear_encrypted();
+		/* tablespace encryption */
+
+		/* Don't encrypt the page 0 of all tablespaces */
+		if (page_id.page_no() == 0) {
+		    req_type.clear_encrypted();
+		    return;
+		}
+
+		switch (space->encryption_type) {
+		case Encryption::KEYRING:
+			if (fil_keyring_skip_encryption(page_id)) {
+				req_type.clear_encrypted();
+				return;
+			} else {
+				fil_io_set_keyring_encryption(
+					req_type, space, page_id);
+				req_type.encryption_algorithm(
+						space->encryption_type);
+			}
+
+		case Encryption::AES:
+			fil_io_set_mk_encryption(req_type, space);
+		        req_type.encryption_algorithm(space->encryption_type);
+			return;
+		case Encryption::NONE:
+			// Already handled above
+		default:
+			ut_a(0);
+		}
 	}
 }
 ```

Comment (satya-bodapati) [reply storage/innobase/fil/fil0fil.cc]: Please test this

File: mysys/system_key.c (hunk 40 above)
Comment (satya-bodapati) [reply mysys/system_key.c:35]: Please ask him on slack and if possible fix it here (no need for separate PR for single line change) You can add in description for clarity.

File: storage/innobase/fil/fil0crypt.cc [hunk 54]
```diff
@@ -3144,3 +3145,177 @@ fil_space_verify_crypt_checksum(
...
+		my_free(key_type);
+		return(NULL);
+	}
+
+	ut_ad(klen2 == ENCRYPTION_KEY_LEN);
+
+	key_iterator it = m_keys.find(version);
+
+	if (it != m_keys.end() && it->second.present) {
+		return(&it->second);
```
satya-bodapati, line 3186: can we compare the key with the cached key here?

storage/innobase/fil/fil0crypt.cc
satya-bodapati: Please fix this

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,177 @@ fil_space_verify_crypt_checksum(
...
+	my_free(key_type);
+
+	return(rk);
+}
+
+redo_log_key*
+redo_log_keys::generate_and_store_new_key() {
+	if (my_key_generate(PERCONA_REDO_KEY_NAME, "AES", NULL,
+			    ENCRYPTION_KEY_LEN)) {
+		ib::error() << "Redo log key generation failed.";
```
satya-bodapati, line 3241: I see that this is suppressed in mtr testcase. If background threads get more turns before keyring plugin is ready, we could see multiple error messages and this is not good.

Is there a way to avoid this?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3144,3 +3145,177 @@ fil_space_verify_crypt_checksum(
...
+redo_log_key *
+redo_log_keys::generate_new_key_without_storing() {
+	ut_ad(m_keys.empty());
+	Encryption::random_value(reinterpret_cast<byte *>(&m_keys[0].key));
+	return(&m_keys[0]);
+}
+
+bool
+redo_log_keys::store_used_keys() {
+	/* This is a for loop, but it really only should store a key
```
satya-bodapati, line 3295: If the map is sorted by version number, can't you just use the first element?

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,59 @@ fil_space_get_scrub_status(
...
 
+struct redo_log_key {
+	uint  version;
+	char  key[ENCRYPTION_KEY_LEN];
+	ulint read_count;
+	ulint write_count;
+	bool  present;
+
+	bool
+	persisted() const {
```
satya-bodapati, line 623: can be one line?

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,59 @@ fil_space_get_scrub_status(
...
+       public:
+	MY_NODISCARD redo_log_key*
+	load_latest_key();
+	MY_NODISCARD redo_log_key*
+	load_key_version(uint version);
+
+	MY_NODISCARD redo_log_key*
+	generate_and_store_new_key();
+
+	/** These two methods are used during bootstrap encryption,
```
satya-bodapati: Improper alignment of comment

storage/innobase/include/os0file.h
```diff
@@ -707,7 +707,8 @@ struct Encryption {
 	dberr_t decrypt_log_block(
 		const IORequest&	type,
 		byte*			src,
-		byte*			dst);
+		byte*			dst,
+		ulint                   offset);
```
satya-bodapati: Add doxygen comment for new parameter

storage/innobase/log/log0log.cc
```diff
@@ -1251,49 +1242,36 @@ log_read_encryption()
...
-	byte*		iv)
-{
-	byte		encryption_info[ENCRYPTION_INFO_SIZE_V2] = {};
+static bool
+log_file_header_fill_encryption(byte *buf, ulint key_version, byte *iv) {
+	byte encryption_info[ENCRYPTION_INFO_SIZE_V2] = {};
 
 	if (!fsp_header_fill_encryption_info(key_version, iv,
-				      encryption_info)) {
+					     encryption_info)) {
```
satya-bodapati, line 1267: Is the indentation correct? (tabs followed by space for alignment)

satya-bodapati (changes requested):
Am not happy with the clang-format for 5.7 codebase (InnoDB).
In you new code, some times the pointer is with variable, sometimes it is with type (inconsistent)

And old InnoDB code, pointer is with type. Please ensure it is with "type" everywhere.

storage/innobase/log/log0log.cc
dutow: Yes, it's exactly at the right position with this.

storage/innobase/include/fil0crypt.h
dutow: It's on two line every other place in the file.

storage/innobase/fil/fil0crypt.cc
dutow: With the loop, I can also add the assert ensuring that other versions are always stored. And this is only called once per run, it's not performance critical.

storage/innobase/fil/fil0crypt.cc
dutow: That suppression was a bug, I've fixed it.

dutow:
Re: pointer alignment.

For some reason, `storage/innobase/.clang-format` has `PointerAlignment: right`. Now looking at the existing code, it really should be aligned to the left. @inikep do you remember the reason behind this?

mysys/system_key.c
dutow: He is also working on binlog encryption (uuid) changes now, so if it has to be changed, it would be better as part of that PR.

satya-bodapati: Ok, please ensure this is not lost. best way: create bug and assign it to him.

storage/innobase/fil/fil0crypt.cc
satya-bodapati: is this comment missed by mistake?

storage/innobase/fil/fil0crypt.cc
satya-bodapati: Cool!

storage/innobase/fil/fil0crypt.cc
satya-bodapati: Ok to leave as is

storage/innobase/log/log0log.cc
```diff
@@ -1385,20 +1355,40 @@ redo log file header.
...
+	}
 }
 
 /** Check the redo log encryption is enabled or not.
 It will try to enable the redo log encryption and write the metadata to
 redo log file header. */
 void
-redo_rotate_default_key()
-{
+redo_rotate_default_key() {
```
satya-bodapati: minor request: only this function name convention looks different. Can we change to log_rotate_default_key(). (I know it exists before too)

satya-bodapati (changes requested):
Looks close to approval.  Please address one comment about assert to compare cached key with retrieved key from keyring

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,59 @@ fil_space_get_scrub_status(
...
+	bool  present;
+
+	bool
+	persisted() const {
+		return version != 0;
+	}
+};
+
+/** Handles the fetching/generation/storing/etc of keyring redo log keys.
+
```
satya-bodapati: remove the trailing space here

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,59 @@ fil_space_get_scrub_status(
...
+/** Handles the fetching/generation/storing/etc of keyring redo log keys.
+ 
+ This class is *NOT* thread safe, as thread safety is not required.
+ Data is only accessed/modified on the following points:
+ * When the redo space is created, at startup
+ * During redo log recovery, at startup
+ * When the server UUID is generated, at startup
+ * When the user requests a new key version, checked periodically in the
+   master thread
+
```
satya-bodapati: remove trailing space here too

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,59 @@ fil_space_get_scrub_status(
...
+	MY_NODISCARD redo_log_key*
+	load_latest_key();
+	MY_NODISCARD redo_log_key*
+	load_key_version(uint version);
+
+	MY_NODISCARD redo_log_key*
+	generate_and_store_new_key();
+
+	/** These two methods are used during bootstrap encryption,
+	 when wo do not yet have an uuid */
```
satya-bodapati: shift "when.."  to one position left

storage/innobase/log/log0log.cc
```diff
@@ -1407,60 +1397,32 @@ redo_rotate_default_key()
...
 
+		log_write_encryption(NULL, NULL, REDO_LOG_ENCRYPT_MK);
+	}
+
+	if (space->encryption_type != Encryption::NONE &&
+	    space->encryption_key_version == REDO_LOG_ENCRYPT_NO_VERSION &&
+	    !srv_read_only_mode &&
+	    srv_redo_log_encrypt == REDO_LOG_ENCRYPT_RK) {
+		/* This only happens when the server uuid was just generated, so we can
+			 * save the key to the keyring */
```
satya-bodapati: This is not aligned properly

storage/innobase/os/os0file.cc
```diff
@@ -10701,16 +10718,11 @@ Encryption::decrypt_log_block(
...
-	const IORequest&	type,
-	byte*			src,
-	ulint			src_len,
-	byte*			dst,
-	ulint			dst_len)
-{
-	Block*		block;
-	byte*		ptr = src;
-	dberr_t		ret;
+Encryption::decrypt_log(const IORequest &type, byte* src, ulint src_len,
```
satya-bodapati, line 10720: doxygen comment for offset parameter

storage/innobase/os/os0file.cc
```diff
@@ -10743,25 +10755,25 @@ Encryption::decrypt_log(
 #endif
 
 		/* Decrypt block */
-		ret = decrypt_log_block(type, ptr, dst);
+		ret = decrypt_log_block(type, ptr, dst, offset);
```
satya-bodapati: add doxygen comment for decrypt_log_block too (offset param)

satya-bodapati: Another interesting thing I found is that offset is not used at all by decrypt_log_block? :(

What is the significance of this parameter? It seems it caused rippling effect (to callers) and only to be unused?

satya-bodapati (changes requested):
Unfortunately there is new issue about "offset" parameter being unused. Please check.

storage/innobase/include/fil0crypt.h
```diff
@@ -612,6 +612,58 @@ fil_space_get_scrub_status(
...
+	persisted() const {
+		return version != 0;
+	}
+};
+
+/** Handles the fetching/generation/storing/etc of keyring redo log keys.
+
+ This class is *NOT* thread safe, as thread safety is not required.
+ Data is only accessed/modified on the following points:
+ * When the redo space is created, at startup
```
satya-bodapati, line 632: Remove the "*" at beginning of this line and below

storage/innobase/srv/srv0srv.cc
```diff
@@ -3604,85 +3608,53 @@ srv_enable_redo_encryption_rk()
...
-
-	ut_ad(redo_key_type && strcmp(redo_key_type, "AES") == 0);
-
-	my_free(redo_key_type);
-	my_free(rkey);
-
+	version = mkey->version;
+	srv_redo_log_key_version = version;
+	memcpy(key, mkey->key, ENCRYPTION_KEY_LEN);
+
```
satya-bodapati: Remove spaces in this line

storage/innobase/srv/srv0start.cc
```diff
@@ -465,23 +465,30 @@ create_log_files(
...
+		Encryption::Type alg = srv_redo_log_encrypt == REDO_LOG_ENCRYPT_RK
+		       ? Encryption::KEYRING : Encryption::AES;
+
+		log_space->flags |= FSP_FLAGS_MASK_ENCRYPTION;
+
+		redo_log_key* mkey = redo_log_key_mgr.generate_new_key_without_storing();
                err = fil_set_encryption(log_space->id,
-                                        Encryption::AES,
-                                        NULL,
+			       		alg,
```
satya-bodapati: Spurious space in this line

satya-bodapati (approved):
No need further approval. Fix those style comments and OK to push/merge

storage/innobase/include/fil0crypt.h
dutow: That's a markdown style list there.

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/81/

PR #3213, opened 2019-06-04
PS-5665 Documentation missing for binlog_space_limit in 8.0

patrickbirch:
Added missing page, updated index and checked ps-versions-comparison doc.

### DIFF
--- a/doc/source/flexibility/maximum_binlog_size.rst
+++ b/doc/source/flexibility/maximum_binlog_size.rst
@@ -1,0 +1,67 @@
+.. _maximum_binlog_size:
+
+========================================
+ Restricting the number of binlog files
+========================================
+
+The task of estimating the number and size of binary logs and the disk space use
+is a challenge. |Percona Server| has the variable
+:variable:`binlog_space_limit`, which you can use to
+define the maximum amount of available space for all binary logs. A limit of
+`0`, the default value, disables the variable, and the binary log files have no limit. 
+
+.. note::
+
+    The value specified is not a hard limit. A single
+    transaction is written to a single binary log, and the size 
+    may exceed the limit. To store a large transaction, the :variable:`binlog_space_limit` removes old binary logs as needed, up to the three oldest ones, to remain under the limit.
+    
+    The last binary log is never pruned.
+
+Example
+=======
+
+The following code is an example of configuring the size of a binary log file
+and the space allowed for all binary file logs in my.cnf:
+
+.. code-block:: guess
+
+[mysqld]
+max_binlog_size = 1G
+binlog_space_limit = 10G
+
+System Variables
+================
+
+.. variable:: max_binlog_size
+
+     :cli: Yes
+     :conf: Yes
+     :scope: Global
+     :dyn: Yes
+     :vartype: INTEGER
+     :default: 1GB
+     :range: 4096 - 1GB
+
+This option specifies the maximum size of the current log file. 
+
+.. seealso::
+
+    MySQL Documentation
+    `maximum_binlog_size <https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#sysvar_max_binlog_size>`__ 
+
+.. variable:: binlog_space_limit
+
+     :cli: Yes
+     :conf: Yes
+     :scope: Global
+     :dyn: No
+     :vartype: ULONG
+     :default: 0 (unlimited)
+     :range: 0-102400
+
+This option places an upper limit on the total size in bytes of all binary logs.
+
+.. note:: You should not set ``--binlog-space-limit`` to less or equal than the value of
+          ``--max-binlog-size`` because if the max-binlog-size limit is reached, the logs are
+          rotated and immediately pruned by binlog-space-limit.
--- a/doc/source/ps-variables.rst
+++ b/doc/source/ps-variables.rst
@@ -77,6 +77,11 @@ System Variables
      - Yes
      - Global
      - No
+   * - :variable:`binlog_space_limit`
+     - Yes
+     - Yes
+     - Global
+     - No
    * - :variable:`csv_mode`
      - Yes
      - Yes
--- a/doc/source/ps-versions-comparison.rst
+++ b/doc/source/ps-versions-comparison.rst
@@ -41,14 +41,14 @@ List of features available in |Percona Server| releases
      - :ref:`Ability to change database for mysqlbinlog <ps80:mysqlbinlog_change_db>`
    * - :ref:`Fixed Size for the Read Ahead Area <ps56:buff_read_ahead_area>`
      - :ref:`Fixed Size for the Read Ahead Area <ps57:buff_read_ahead_area>`
-     - :ref:`Fixed Size for the Read Ahead Area <ps80:buff_read_ahead_area>`      
+     - :ref:`Fixed Size for the Read Ahead Area <ps80:buff_read_ahead_area>`
    * - :ref:`Improved MEMORY Storage Engine <ps56:improved_memory_engine>`
      - :ref:`Improved MEMORY Storage Engine <ps57:improved_memory_engine>`
      - :ref:`Improved MEMORY Storage Engine <ps80:improved_memory_engine>`
    * - :ref:`Restricting the number of binlog files <ps56:maximum_binlog_files>`
      - :ref:`Restricting the number of binlog files <ps57:maximum_binlog_files>`
-     - :ref:`Restricting the number of binlog files <ps80:maximum_binlog_files>`
-   * - :ref:`Ignoring missing tables in mysqldump <ps56:mysqldump_ignore_create_error>`
+     - :ref:`Restricting the size of binlog files <ps80:maximum_binlog_size>`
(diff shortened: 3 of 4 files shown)


laurynas-biveinis (changes requested):
It is max_binlog_size that needs to be documented
The max_binlog_files variable has been removed from 8.0 and only should be mentioned, if ever, historically as a 5.7 feature that's no longer present.

Please also read the internal Confluence page on commit message formatting

patrickbirch:
deleted source/flexibility/max_binlog_files.rst
renamed, rewrote, and updated source/flexibility/maximum_binlog_size.rst
updated source/index.rst replacing max_binlog_files with maximum_binlog_size.rst
updated source/ps-variables with variables
updated the link for source/ps-versions-comparison

doc/source/flexibility/maximum_binlog_size.rst
```diff
@@ -0,0 +1,57 @@
...
+
+[mysqld]
+max_binlog_size = 1G
+binlog_space_limit = 10G
+
+Version Specific Information
+============================
+
+  * 5.7.10-1:
+    Variable :variable:`max_binlog_files` ported from |Percona Server| 5.6.
```
dutow: Why are there 5.7 versions mentioned in an 8.0 documentation? Especially about a removed variable.

doc/source/flexibility/maximum_binlog_size.rst
```diff
@@ -0,0 +1,57 @@
...
+
+     :cli: Yes
+     :conf: Yes
+     :scope: Global
+     :dyn: Yes
+     :vartype: INTEGER
+     :default: 1GB
+     :range: 4096 - 1GB
+
+This option is a limit to the current log file size.
```
dutow: This is an upstream variable? https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#sysvar_max_binlog_size

doc/source/ps-variables.rst
```diff
@@ -217,6 +222,11 @@ System Variables
      - Yes
      - Global
      - Yes
+   * - :variable:`max_binlog_size`
```
dutow: This is an upstream variable.

doc/source/ps-versions-comparison.rst
```diff
@@ -41,14 +41,14 @@ List of features available in |Percona Server| releases
...
-     - :ref:`Fixed Size for the Read Ahead Area <ps80:buff_read_ahead_area>`      
+     - :ref:`Fixed Size for the Read Ahead Area <ps80:buff_read_ahead_area>`
    * - :ref:`Improved MEMORY Storage Engine <ps56:improved_memory_engine>`
      - :ref:`Improved MEMORY Storage Engine <ps57:improved_memory_engine>`
      - :ref:`Improved MEMORY Storage Engine <ps80:improved_memory_engine>`
    * - :ref:`Restricting the number of binlog files <ps56:maximum_binlog_files>`
      - :ref:`Restricting the number of binlog files <ps57:maximum_binlog_files>`
-     - :ref:`Restricting the number of binlog files <ps80:maximum_binlog_files>`
-   * - :ref:`Ignoring missing tables in mysqldump <ps56:mysqldump_ignore_create_error>`
+     - :ref:`Restricting the number of binlog files <ps80:maximum_binlog_size>`
```
dutow: We no longer restring the number of files - only the size of the files.

doc/source/flexibility/maximum_binlog_size.rst
```diff
@@ -0,0 +1,57 @@
+.. _maximum_binlog_size:
+
+========================================
+ Restricting the number of binlog files
+========================================
+
+Estimating the number of binary logs and disk space use is a challenge. |Percona Server| has the variable :variable:`binlog_space_limit`, which sets the maximum amount of space to use for all binary logs. With the :variable:`binlog_space_limit` set to `0`, the default value, the variable is turned off and no limit is set. You set the :variable:`binlog_space_limit` on and provide a size limit. The value specified is not a hard limit. A single transaction is written to a single binary log, and the size of this transaction may exceed the limit.
```
dutow: * Not only a single transaction: as the documentation later describes, the last binlog file will never be pruned.
* Please use consistent grammar, e.g. With variable set..., you set, ...

doc/source/flexibility/maximum_binlog_size.rst
```diff
@@ -0,0 +1,57 @@
...
+
+For example, the :variable:`max_binlog_size` is set to 1GB, which is the default value and the maximum value and the :variable:`binlog_space_limit` is set to 10GB. A large transaction changes 5GB of data. To store the transaction, the :variable:`binlog_space_limit` removes old binary logs as needed, up to the three oldest ones, to remain under the limit.
```
dutow: "default value and the maximum value and the <variable>" - there are definitely some extra words here.

I also don't understand the example - why three? Why does it have to be one transaction? And it doesn't mention the current binlog size, only that the maximum is 10g - so if currently the sum of binlog size is 3 or 4g, this variable will do nothing.

doc/source/ps-versions-comparison.rst
```diff
@@ -41,14 +41,14 @@ List of features available in |Percona Server| releases
...
+   * - :ref:`Ignoring missing tables in mysqldump  <ps56:mysqldump_ignore_create_error>`
```
dutow, line 51: Why changed?

dutow (changes requested):

dutow:
Please also change your commit message. The first line should contain the jira ticket id, otherwise jira won't pick it up. E.g. `PS-5665: ...` or `PS-5665 ...`

doc/source/flexibility/maximum_binlog_size.rst
patrickbirch: removed.

doc/source/flexibility/maximum_binlog_size.rst
patrickbirch: Added a link to the MySQL documentation.

doc/source/ps-variables.rst
patrickbirch: removed from the list of ps-variables.

doc/source/flexibility/maximum_binlog_size.rst
patrickbirch: rewritten.

doc/source/flexibility/maximum_binlog_size.rst
patrickbirch: Removed most of the example text.

doc/source/ps-versions-comparison.rst
patrickbirch: Changed the "number" to "size" since you said that with 8.0 "we are no longer restricting the number"..."only the size"

patrickbirch:
Updated commit message and PR heading.

doc/source/ps-versions-comparison.rst
patrickbirch: changed to restricting the size of binlog files

doc/source/flexibility/maximum_binlog_size.rst
```diff
@@ -0,0 +1,67 @@
...
+
+.. variable:: binlog_space_limit
+
+     :cli: Yes
+     :conf: Yes
+     :scope: Global
+     :dyn: No
+     :vartype: ULONG
+     :default: 0 (unlimited)
+     :range: 0-102400
```
dutow, line 61: I don't see an upper limit in the code, it's ULONG_MAX

doc/source/flexibility/maximum_binlog_size.rst
```diff
@@ -0,0 +1,67 @@
...
+
+.. variable:: max_binlog_size
+
+     :cli: Yes
+     :conf: Yes
+     :scope: Global
+     :dyn: Yes
+     :vartype: INTEGER
+     :default: 1GB
+     :range: 4096 - 1GB
```
dutow, line 44: 4096-1GB isn't realy clear (as 4096 is larger than 1). It would be better to keep this the same as upstream, using 1073741824 instead of 1GB.

doc/source/flexibility/maximum_binlog_size.rst
patrickbirch: fixed.

doc/source/flexibility/maximum_binlog_size.rst
patrickbirch: Fixed.

patrickbirch:
waiting for either approval or comments.

doc/source/flexibility/maximum_binlog_size.rst
dutow: This wasn't fixed, range still shows "0-102400" 
There is no limit other than the vartype.

Please see how upstream documents these, like binlog_cache_size: https://dev.mysql.com/doc/refman/5.6/en/replication-options-binary-log.html#sysvar_binlog_cache_size

dutow (changes requested):

PR #3215, opened 2019-06-05
PS-5659-RHEL8-package-install-requires-mysql-mod-disabled - added RHE…

patrickbirch:
…L 8 to supported releases and added a note to installing PS using rpm packages - step 2.


laurynas-biveinis (changes requested):
The fix branches must be GCA-based

laurynas-biveinis (changes requested):
The fix branches must be GCA-based

patrickbirch:
Fixed branch to GCA

george-lorch (approved):

PR #3216, opened 2019-06-05
PS-5662-MyRocks-limitation-partial-updates 

patrickbirch:
added text about Partial update of LOB in InnoDB to MyRocks Limitations


george-lorch (approved):
+1

PR #3219, opened 2019-06-05
PS-5656-doc-incorrect-encryption-8.0  

patrickbirch:
added text to note line 225 and fixed note header line 641

### DIFF
--- a/doc/source/management/data_at_rest_encryption.rst
+++ b/doc/source/management/data_at_rest_encryption.rst
@@ -222,8 +222,21 @@ Trying to add unencrypted table to this table space will result in an error:
 
 .. note::
 
-   |Percona XtraBackup| currently doesn't support backup of encrypted general
-   tablespaces.
+    |Percona XtraBackup| version 2.4 and version 8.0 support the backup of encrypted general
+    tablespaces.
+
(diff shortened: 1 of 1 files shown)


laurynas-biveinis (commented):
Same comment as for the 5.7 version

doc/source/management/data_at_rest_encryption.rst
```diff
@@ -222,8 +222,21 @@ Trying to add unencrypted table to this table space will result in an error:
 
 .. note::
 
-   |Percona XtraBackup| currently doesn't support backup of encrypted general
-   tablespaces.
+    |Percona XtraBackup| version 2.4 and version 8.0 support the backup of encrypted general
```
george-lorch, line 225: Only PXB 8.0 works on PS 8.0, 2.4 does not support PS/MySQL 8.0.  Should remove all references to 2.4 in 8.0 docs other than maybe to state the fact that only PXB 8.0 supports PS/MySQL 8.0.

george-lorch (changes requested):

patrickbirch:
No longer needed. Encryption docs have been updated.

PR #3220, opened 2019-06-05
PS-5636-innodb_log_block_size-note-5.6  …

patrickbirch:
removed note that the feature is considered BETA quality. The settings are the same as version 5.5.


laurynas-biveinis (changes requested):
5.6 docs are correct. It is the 5.5 docs - which are EOL but still deployed - that would need correction

patrickbirch:
No change needed.

PR #3221, opened 2019-06-07
PS-5149: Redo log encryption type shouldn't be changed on a running server (8.0)

dutow:
We support two redo log encryption modes:
* master_key, as upstream
* keyring_key, a per-page versioned encryption mode

The active encryption mode is chosen using the innodb-redo-log-encypt variable. This currently allows the user to select one mode, then change to the other, resulting in errors if the server has to decrypt earlier redo log blocks. To fix this, we disallow changing redo log encryption modes.

* First, during startup, if the log is already encrypted, we detect the existing encryption mode
* Later, when the user changes this variable, we check against the existing encryption mode. If the logs aren't encrypted, or if they were encrypted in the same mode, we allow the change. Otherwise, we report a warning and leave the variable as is.
* If we changed the encryption mode in the previous step, we also update the existing encryption mode to match it.

This way we disallow both direct (e.g. master_key -> keyring_key) and indirect (e.g. master_key -> off -> server restart -> keyring_key) mode changes, but allow changes back to the same mode (e.g. master_key -> off (-> server restarts, etc) -> master_key).

To change the encryption mode, we have to:

* shutdown the server
* delete the redo log files
* start up the server with the new encryption settings


storage/innobase/handler/ha_innodb.cc
```diff
@@ -21841,7 +21841,7 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
 
   if (srv_redo_log_encrypt != REDO_LOG_ENCRYPT_OFF &&
       srv_redo_log_encrypt != target) {
-    push_warning_printf(thd, Sql_condition::SL_WARNING, ER_WRONG_ARGUMENTS,
+    push_warning_printf(thd, Sql_condition::SL_WARNING, ER_REDO_ENCRYPTION_CANT_BE_CHANGED,
```
laurynas-biveinis: Error not warning should be returned?

laurynas-biveinis (changes requested):

storage/innobase/handler/ha_innodb.cc
dutow: I kept it as a warning, as other similar messages (for example, if the user tries to turn on encryption on a read only server) are warnings too, and those are upstream.

share/errmsg-utf8.txt
```diff
@@ -18983,6 +18983,9 @@ ER_RDB_TTL_DURATION_FORMAT
 ER_PER_INDEX_CF_DEPRECATED
   eng "The per-index column family option has been deprecated."
 
+ER_REDO_ENCRYPTION_CANT_BE_CHANGED
```
laurynas-biveinis: This is not a MyRocks-specific message - wrong file section then?

laurynas-biveinis (changes requested):

share/errmsg-utf8.txt
dutow: Right, moved.

dutow:
Updated based on @satya-bodapati 's 5.7 comments.

laurynas-biveinis (approved):

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21839,12 +21839,20 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
 
-  if (srv_redo_log_encrypt != REDO_LOG_ENCRYPT_OFF &&
-      srv_redo_log_encrypt != target) {
-    push_warning_printf(thd, Sql_condition::SL_WARNING, ER_WRONG_ARGUMENTS,
-                        " Redo log encryption mode"
-                        " can't be switched without stopping the server and"
-                        " recreating the redo logs.");
+  if (existing_redo_encryption_mode != REDO_LOG_ENCRYPT_OFF &&
+      existing_redo_encryption_mode != target) {
+    push_warning_printf(
```
satya-bodapati: may be use ib_senderrf(thd, IB_LOG_LEVEL_WARN, ER_REDO_ENCRYPTION_CANT_BE_CHANGED, log_encrypt_name(existing_redo_encryption_mode), log_encrypt_name(static_cast<redo_log_encrypt_enum>(target))

satya-bodapati: ib::warn() could be unnecessary then? please check

satya-bodapati (changes requested):

satya-bodapati (approved):

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/162/

PR #3231, opened 2019-06-11
PS-5593: Assertion failure this == space->crypt_data in encryption

robgolebiowski:
threads

When KEYRING encryption is started for a space by encryption thread
a crypt_data is generated for this space. However there is a race
when setting this crypt data for space (changing from
space->crypt_data == nullptr).
fil_crypt_read_crypt_data can be called at the moment when
crypt_data is being set. If it discovers that crypt_data is nullptr
it will try to read crypt_data from page0 and assign it to space. It can
read it as a nullptr and override the crypt_data created by
start_encrypting_space - as there could be a race here.
Normally this is secured by taking a lock on crypt_data->mutex.
Which is not possible here, as crypt_data doest not exist yet.

Other solution would be to add some additional synchronization here,
however since fil_crypt_read_crypt_data can be called periodically
by encryption threads it may add additional contention.
start_encrypting_space could also try to set this crypt_data in
loop, however this seems like better solution.

### DIFF
--- a/storage/innobase/fil/fil0crypt.cc
+++ b/storage/innobase/fil/fil0crypt.cc
@@ -832,8 +832,11 @@ fil_crypt_read_crypt_data(fil_space_t* space) {
 					      page_size, RW_S_LATCH, &mtr)) {
 		mutex_enter(&fil_system->mutex);
 		if (!space->crypt_data) {
-			space->crypt_data = fil_space_read_crypt_data(
-				page_size, block->frame);
+			fil_space_crypt_t *crypt_data =
+				fil_space_read_crypt_data(page_size, block->frame);
+			if (crypt_data != NULL) {
+				space->crypt_data = crypt_data;
+			}
 		}
 		mutex_exit(&fil_system->mutex);
 	}


laurynas-biveinis (commented):
The commit message should describe the solution taken, too (now it describes the problem and other alternative solutions, but not the actual solution).

Where in the code the other thread accesses this?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -832,8 +832,11 @@ fil_crypt_read_crypt_data(fil_space_t* space) {
 					      page_size, RW_S_LATCH, &mtr)) {
 		mutex_enter(&fil_system->mutex);
 		if (!space->crypt_data) {
-			space->crypt_data = fil_space_read_crypt_data(
-				page_size, block->frame);
+			fil_space_crypt_t *crypt_data =
```
satya-bodapati, line 835: This fix doesn't seem to be right

We have fil_space_t->crypt_data,  it can go NULL<->crypt_data and multiple threads operating on it makes decision based on fil_space_t->crypt_data value. This need proper synchronization. IMHO, u should consider using il_space_t rw latch. Readers take S and Writers (those set the variable) take X.


crypt_data->mutex is only about "crypt_data" content. But not about crypt_data being set to NULL. Infact I wonder what prevents this?

thread  operating on fil_space_t took crypt_data->mutex and reading crypt_data. This is is fine. No concurrent modifications of crypt_data.  what prevents from the another thread to set fil_space_t->crypt_data = NULL? Does that thread take crypt_data->mutex too?

satya-bodapati (changes requested):

storage/innobase/fil/fil0crypt.cc
satya-bodapati: btw, one more minor thing that I think should be improved. The function names:
 fil_crypt_read_crypt_data() and fil_space_read_crypt_data()

both seem to  achieve the same thing. In fact for a moment, I thought it is recursive call ;)

robgolebiowski:
I have changed the commit message so it would describe the fix.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: @satya, pasting here our conversation from slack (the last part of it), for documentation:
Satya Bodapati [10:58 AM]
@robert.golebiowski while I think your fix can work..  but it can be done better way.  The reader  fil_crypt_read_crypt_data() is reading space->crypt_data after page 0 S latch. The writer, fil_crypt_start_encrypting_space(), if it can call fil_space_set_crypt_data() after page0 X latch, then it should solve the issue. essentially we are using Page 0 latch to solve the conncurrency issue and it seems cleaner to me.
Dont know if it works.. but try and see

Robert Golebiowski [11:11 AM]
@satya - It was by design - crypt_data cannot go to NULL. If we change it here from this assumption to latches - we will have to add latches in other places that are accessing page0 - to be consistent. It is cleaner, I agree. However, as I you probably have seen - they added a lot of such silent assumption in the code (probably known only to the programmer who wrote this) - maybe to get some performance gain. I am positive that if we start to introduce new latches etc. - we get into some other problems. Because of that I would leave the fix as minimal as it is now and probably have a separate re-design request - about protecting crypt_data struct and page0 in encryption threads .... it would be cleaner to me. The tests are stable, I would not like to make some bigger changes as part of bug fix - unless necessary. We can do it separately.

Satya Bodapati [11:24 AM]
ideally you should use some latch or mutex. This page 0 latch is still a compromiised solution because space->crypt_data cannot go to NULL. Infact my suggestion is to use fil_space_t->rw_latch.  I dont agree on the performance gain by the current code/patch. I am seeing uncontrolled use of fil_system mutex(wider scope latch) for setting fil_space_t->crypt_data. The suggestion is to use  fine graned latch (smaller scope) and consistently. Am  sure it will become one of those Low prio feature enhancements.. :wink:

Robert Golebiowski [11:32 AM]
I will add those comments into PR and we will ask @laurynas to decide - whether we should start changing those latches and mutexes as part of this bug fix or as a separate bug/feature requests. If it ever becomes low prio feature, then it should not be done now too.

Satya Bodapati [11:35 AM]
my point is that amount of code to do "moving fil_space_set_crypt_data() after page 0 X latch"  would be minimal and is considered as feature request :wink:

Robert Golebiowski [11:42 AM]
ok, I get you - but do not you think this needs to be added into all places where page0 is accessed and drop fil_system mutex -since we are locking the page0? If I disable fil_system mutex hell knows what will happen :slightly_smiling_face: If you insists I can add this latches - without removing fil_system mutex. But I can imagine people wondering - why in this place latch is taken and in other place it is not ? why do I need fil_system mutex, since the latch should be enough. I think if we start changing how this works we should be consistent - and do not change it only in one place. That's why I think it is a bigger change and should be at least a separate bug.

Satya Bodapati [11:45 AM]
Hmm.. Better to do it properly and cleanly (fix at all places).. please create enhancement/bug. Also please explain how your fix works and is safe in PR page( and commit msg)

Robert Golebiowski [11:47 AM]
great :handshake:

Satya Bodapati [11:49 AM]
:+1::skin-tone-3:

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/165/

laurynas-biveinis (commented):
In the commit message: "reads crypt_data from page0 and find it missing, it tries to read it from page0." - same action twice?

robgolebiowski:
corrected commit message

laurynas-biveinis (approved):
OK as an interim fix as discussed

storage/innobase/fil/fil0crypt.cc
robgolebiowski: Created PS-5858.

PR #3232, opened 2019-06-11
Assert problem 8.0

robgolebiowski:


laurynas-biveinis (commented):
Same comment as for 5.7

PR #3235, opened 2019-06-12
PS-5682-extra_port  

patrickbirch:
removed extra_port reference from ps-variables.rst and extra_port and the extra_max_connections from performance/threadpool.rst

### DIFF
--- a/doc/source/performance/threadpool.rst
+++ b/doc/source/performance/threadpool.rst
@@ -4,9 +4,9 @@
  Thread Pool
 =============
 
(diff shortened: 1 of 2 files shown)


laurynas-biveinis (changes requested):
The docs have to be "versioned", that is, these variables have to be mentioned in .rst, but as removed in such and such version, for example check binlog_encryption https://www.percona.com/doc/percona-server/LATEST/management/data_at_rest_encryption.html#binary-log-encryption

Its replacement - upstream admin_port - should be mentioned too, with a doc link

PR #3237, opened 2019-06-13
PS-5322 Keyring encrypted tablespace import/export is broken

robgolebiowski:
Introduced IB_EXPORT_CFG_VERSION_V3_WITH_RK for marking tablespaces
exported with keyring encryption. Please note that V2(IB_EXPORT_CFG_VERSION_V2)
was skipped because when V2 was added by upstream the import/export
with keyring encryption was not functional. Thus it is not possible
for keyring encrypted tablespaces to be exported with V2.

The error handling was improved, but not all paths were covered and some
decryption errors can manifest as "missing tablespace" range of errors.
Please have a look at diff for innodb-bad-key-change2.test.
The keyring encryption error handling becomes a big part of every
upstream merge. To mitigate this problem an improvement PS-5635 was
proposed.

### DIFF
--- a/mysql-test/suite/encryption/t/innodb-bad-key-change2-master.opt
+++ b/mysql-test/suite/encryption/t/innodb-bad-key-change2-master.opt
@@ -1,0 +1,4 @@
+$KEYRING_PLUGIN_OPT
+--early-plugin-load="keyring_file=$KEYRING_PLUGIN"
+--loose-keyring_file_data=$MYSQLTEST_VARDIR/std_data/keys2.txt
+--innodb-stats-persistent=OFF
--- a/mysql-test/suite/encryption/t/innodb-bad-key-change3.test
+++ b/mysql-test/suite/encryption/t/innodb-bad-key-change3.test
@@ -1,12 +1,9 @@
---skip #has import
 -- source include/not_valgrind.inc
 
-call mtr.add_suppression("cannot be decrypted. Are you using correct keyring that contain the");
-call mtr.add_suppression("InnoDB: Tablespace for table .* is set as discarded.");
-call mtr.add_suppression("InnoDB: Cannot calculate statistics for table .* because the .ibd file is missing. Please refer to .* for how to resolve the issue.");
-call mtr.add_suppression("InnoDB: Trying to access missing tablespace [0-9]+");
-call mtr.add_suppression("InnoDB: Cannot save statistics for table .* because the .ibd file is missing.");
-
+call mtr.add_suppression("\\[InnoDB\\] Tablespace id [0-9]+ in file t1\.ibd is encrypted but keyring or used key_id 4 is not available\. Can't continue reading table\. Please provide the correct keyring\.");
+call mtr.add_suppression("\\[InnoDB\\] Tablespace for table `test`\.`t1` is set as discarded\.");
+call mtr.add_suppression("\\[InnoDB\\] Cannot calculate statistics for table .* because the \.ibd file is missing\. Please refer to .* for how to resolve the issue\.");
+call mtr.add_suppression("\\[InnoDB\\] Encryption can't find tablespace key_id = 4");
 
 CREATE TABLE t1 (pk INT PRIMARY KEY, f VARCHAR(255)) ENGINE=InnoDB ROW_FORMAT=COMPRESSED ENCRYPTION='KEYRING' ENCRYPTION_KEY_ID=4;
 SHOW WARNINGS;
@@ -37,7 +34,7 @@ EOF
 --let MYSQLD_DATADIR =`SELECT @@datadir`
 --let t1_IBD = $MYSQLD_DATADIR/test/t1.ibd
 
---error ER_GET_ERRMSG
+--error ER_IMPORT_TABLESPACE_MISSING_KEY
 ALTER TABLE t1 IMPORT TABLESPACE;
 SHOW CREATE TABLE t1;
 --error ER_TABLESPACE_DISCARDED
--- a/mysql-test/suite/encryption/t/innodb_encryption_discard_import.test
+++ b/mysql-test/suite/encryption/t/innodb_encryption_discard_import.test
@@ -1,9 +1,14 @@
---skip
 --source include/not_valgrind.inc
 
-call mtr.add_suppression("\\[Warning\\] InnoDB: Tablespace for table `[^`]+`.`[^`]+` is set as discarded\\.");
-call mtr.add_suppression("\\[Warning\\] InnoDB: Page [[:digit:]]+ at offset [[:digit:]]+ looks corrupted in file");
-call mtr.add_suppression("\\[ERROR\\] InnoDB: os_file_read\\(\\) failed");
+call mtr.add_suppression("\\[InnoDB\\] Tablespace id .* in file t[1-8]\.ibd is encrypted but keyring or used key_id [4-5] is not available\. Can't continue reading table\. Please provide the correct keyring\.");
+call mtr.add_suppression("\\[InnoDB\\] Tablespace for table `test`\.`t[1-8]` is set as discarded\.");
+call mtr.add_suppression("\\[InnoDB\\] Cannot calculate statistics for table .* because the \.ibd file is missing\. Please refer to .* for how to resolve the issue\.");
+call mtr.add_suppression("\\[InnoDB\\] Encryption can't find tablespace key_id = 5");
+
+CREATE TABLE t2 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB ENCRYPTION='Y';
+
+SET GLOBAL default_table_encryption = ONLINE_TO_KEYRING;
+SET GLOBAL innodb_encryption_threads = 4;
 
 --let $number_of_records = 10000
 --let MYSQLD_DATADIR = `SELECT @@datadir`
@@ -19,7 +24,6 @@ call mtr.add_suppression("\\[ERROR\\] InnoDB: os_file_read\\(\\) failed");
 --let $t8_ibd = $MYSQLD_DATADIR/$DB_NAME/t8.ibd
 
 CREATE TABLE t1 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB ENCRYPTION='KEYRING';
-CREATE TABLE t2 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB ENCRYPTION='Y';
 CREATE TABLE t3 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB COMPRESSION="zlib" ENCRYPTION='KEYRING';
 CREATE TABLE t4 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB COMPRESSION="zlib";
 CREATE TABLE t5 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB ROW_FORMAT=COMPRESSED ENCRYPTION='KEYRING';
@@ -52,14 +56,12 @@ eval CALL innodb_insert_proc($number_of_records);
 COMMIT;
 SET autocommit = 1;
 
---let $tables_count = `SELECT COUNT(*) FROM INFORMATION_SCHEMA.INNODB_SYS_TABLESPACES`
+--let $tables_count=`SELECT COUNT(*) FROM INFORMATION_SCHEMA.INNODB_TABLESPACES`
 
 --echo # Wait max 2 min for key encryption threads to encrypt all spaces
 --let $wait_timeout = 120
-# All tables should get encrypted. Exactly ($tables_count) because
-# INNODB_TABLESPACES_ENCRYPTION contains artificial 'innodb_system' (+1)
-# table and 't7' is unencrypted (-1).
---let $wait_condition = SELECT COUNT(*) = $tables_count FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION = 1
+# All tables should get encrypted. Apart from temporary tablespace and t7.
+--let $wait_condition = SELECT COUNT(*) = $tables_count - 2 FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION = 1
 --source include/wait_condition.inc
 
 --let $assert_text = Make sure t7 is not encrypted
@@ -191,10 +193,17 @@ ALTER TABLE t8 IMPORT TABLESPACE;
 --source include/search_pattern_in_file.inc
 
 
-ALTER TABLE t1 ENGINE InnoDB;
-SHOW CREATE TABLE t1;
+SET GLOBAL innodb_encryption_threads = 0;
+SET GLOBAL default_table_encryption = OFF;
+
 ALTER TABLE t2 ENGINE InnoDB;
 SHOW CREATE TABLE t2;
+
+SET GLOBAL default_table_encryption = ONLINE_TO_KEYRING;
+SET GLOBAL innodb_encryption_threads = 4;
+
+ALTER TABLE t1 ENGINE InnoDB;
+SHOW CREATE TABLE t1;
 ALTER TABLE t3 ENGINE InnoDB;
 SHOW CREATE TABLE t3;
 ALTER TABLE t4 ENGINE InnoDB;
@@ -209,6 +218,7 @@ ALTER TABLE t8 ENGINE InnoDB;
 SHOW CREATE TABLE t8;
 
 --echo # Restarting server
+--let $restart_hide_args=1
 --source include/restart_mysqld.inc
 --echo # Done restarting server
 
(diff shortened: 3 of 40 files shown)


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/161/
failing bug innodb.percona_changed_page_bmp_requests_debug => as in https://ps80.cd.percona.com/job/percona-server-8.0-trunk/50/
failing bug innodb.blob_page_reserve => as in https://ps80.cd.percona.com/job/percona-server-8.0-trunk/51/

mysql-test/suite/encryption/t/innodb-bad-key-change2-master.opt
```diff
@@ -0,0 +1,6 @@
+$KEYRING_PLUGIN_OPT
+--early-plugin-load="keyring_file=$KEYRING_PLUGIN"
+--loose-keyring_file_data=$MYSQLTEST_VARDIR/std_data/keys2.txt
+--innodb-encrypt-tables=OFF
+--innodb-stats-persistent=OFF
+--innodb-file-per-table=ON
```
laurynas-biveinis: A default already

mysql-test/suite/encryption/t/innodb-bad-key-change2-master.opt
```diff
@@ -0,0 +1,6 @@
+$KEYRING_PLUGIN_OPT
+--early-plugin-load="keyring_file=$KEYRING_PLUGIN"
+--loose-keyring_file_data=$MYSQLTEST_VARDIR/std_data/keys2.txt
+--innodb-encrypt-tables=OFF
```
laurynas-biveinis: A default already

mysql-test/suite/encryption/t/innodb-bad-key-change3.test
```diff
@@ -1,10 +1,7 @@
---skip #has import
 -- source include/not_valgrind.inc
 
-call mtr.add_suppression("cannot be decrypted. Are you using correct keyring that contain the");
-call mtr.add_suppression("InnoDB: Tablespace for table .* is set as discarded.");
-call mtr.add_suppression("InnoDB: Cannot calculate statistics for table .* because the .ibd file is missing. Please refer to .* for how to resolve the issue.");
-
+call mtr.add_suppression("\\[InnoDB\\] Tablespace id 4 in file t1\.ibd is encrypted but keyring or used key_id 4 is not available\. Can't continue reading table\. Please provide the correct keyring\.");
```
laurynas-biveinis: Is the tablespace ID stable if the same worker instance has run more or fewer tests before this one?

mysql-test/suite/encryption/t/innodb_encryption_discard_import-master.opt
```diff
@@ -1,5 +1,5 @@
 $KEYRING_PLUGIN_OPT
 --early-plugin-load="keyring_file=$KEYRING_PLUGIN"
 --loose-keyring_file_data=$MYSQL_TMP_DIR/mysecret_keyring
---innodb-encrypt-tables=ONLINE_TO_KEYRING
+--innodb-encrypt-tables=OFF
```
laurynas-biveinis: Already a default

mysql-test/suite/encryption/t/innodb_encryption_discard_import.test
```diff
@@ -1,9 +1,9 @@
---skip
 --source include/not_valgrind.inc
 
-call mtr.add_suppression("\\[Warning\\] InnoDB: Tablespace for table `[^`]+`.`[^`]+` is set as discarded\\.");
-call mtr.add_suppression("\\[Warning\\] InnoDB: Page [[:digit:]]+ at offset [[:digit:]]+ looks corrupted in file");
-call mtr.add_suppression("\\[ERROR\\] InnoDB: os_file_read\\(\\) failed");
+call mtr.add_suppression("\\[InnoDB\\] Tablespace id .* in file t[1-8]\.ibd is encrypted but keyring or used key_id [4-5] is not available\. Can't continue reading table\. Please provide the correct keyring\.");
```
laurynas-biveinis, line 3: Same question as for the other test, is this range enough?

mysql-test/suite/encryption/t/innodb_encryption_discard_import.test
```diff
@@ -52,14 +52,12 @@ eval CALL innodb_insert_proc($number_of_records);
 COMMIT;
 SET autocommit = 1;
 
---let $tables_count = `SELECT COUNT(*) FROM INFORMATION_SCHEMA.INNODB_SYS_TABLESPACES`
+--let $tables_count=`select count(*) from INFORMATION_SCHEMA.INNODB_TABLESPACES`
```
laurynas-biveinis: Inconsistent SQL keyword case

storage/innobase/buf/checksum.cc
```diff
@@ -746,25 +752,18 @@ std::ostream &operator<<(std::ostream &out, const page_id_t &page_id) {
...
-		? ut_crc32_legacy_big_endian
-		: ut_crc32;
-
-	const uint32_t	c1 = crc32_func(
-		page + FIL_PAGE_OFFSET,
-		FIL_PAGE_FILE_FLUSH_LSN - FIL_PAGE_OFFSET);
+uint32_t buf_calc_page_crc32_encrypted_with_keyring(
+    const byte *page, uint page_size,
+    bool use_legacy_big_endian /* = false */) {
+  ut_crc32_func_t crc32_func =
```
laurynas-biveinis: Can this legacy business be dropped? Not in this PR, but please make a note

storage/innobase/row/row0import.cc
```diff
@@ -3739,12 +3742,13 @@ dberr_t row_import_for_mysql(dict_table_t *table, dd::Table *table_def,
...
   }
 
-  /*a For encrypted tablespace, set encryption information. */
-  if (FSP_FLAGS_GET_ENCRYPTION(fsp_flags)) {
-    err = fil_set_encryption(
-        table->space,
-        cfg.m_is_keyring_encrypted ? Encryption::KEYRING : Encryption::AES,
-        table->encryption_key, table->encryption_iv);
+  /* For encrypted tablespace, set encryption information.
+   * If it is keyring encryption encryption iv was already read in fil_ibd_open
```
laurynas-biveinis: Leading * column comment style not common in InnoDB

storage/innobase/row/row0mysql.cc
```diff
@@ -1948,11 +1948,17 @@ or is tablespace .ibd file missing.
...
+                                     when tablespace is missing, could
+                                     be either DB_TABLESPACE_NOT_FOUND
+                                     or DB_TABLE_NOT_FOUND
 @retval DB_IO_DECRYPT_FAIL    table is encrypted but decryption failed
 @retval DB_CORRUPTION           table is corrupted
-@retval DB_TABLESPACE_NOT_FOUND tablespace .ibd file not found */
-static dberr_t row_mysql_get_table_status(const dict_table_t *table, trx_t *trx,
-                                          bool push_warning = true) {
+@retval DB_TABLESPACE_NOT_FOUND tablespace .ibd file not found
+        or DB_TABLE_NOT_FOUND   depends on err_on_tablespace_missing */
```
laurynas-biveinis: Split DB_TABLE_NOT_FOUND to its own retval

laurynas-biveinis (changes requested):

mysql-test/suite/encryption/t/innodb-bad-key-change3.test
robgolebiowski: I think it should be stable. All tests have to pass MTR validation that they leave the server in the same state that they found it. Thus all tables created by a test should have been removed by it. Then the tablespace id should be always the same - in theory ...

robgolebiowski: I think it should be stable. All tests have to pass MTR validation that they leave the server in the same state that they found it. Thus all tables created by a test should have been removed by it. Then the tablespace id should be always the same - in theory ...

laurynas-biveinis: And are the freed tablespace IDs reused?

mysql-test/suite/encryption/t/innodb_encryption_discard_import-master.opt
robgolebiowski: done

mysql-test/suite/encryption/t/innodb_encryption_discard_import.test
robgolebiowski: this should be fine as the relation tablespace t1 -> t1.ibd is fixed

mysql-test/suite/encryption/t/innodb-bad-key-change3.test
robgolebiowski: Ah, right, they are not reused, fixed.

mysql-test/suite/encryption/t/innodb_encryption_discard_import.test
robgolebiowski: fixed

storage/innobase/row/row0import.cc
robgolebiowski: done

storage/innobase/row/row0mysql.cc
robgolebiowski: done

mysql-test/suite/encryption/t/innodb-bad-key-change2-master.opt
robgolebiowski: done

mysql-test/suite/encryption/t/innodb-bad-key-change2-master.opt
robgolebiowski: done

storage/innobase/buf/checksum.cc
robgolebiowski: Created PS-5705: Get rid of legacy_big_endian_checksum

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/179/

storage/innobase/api/api0api.cc
```diff
@@ -2960,7 +2960,8 @@ dberr_t ib_sdi_set(uint32_t tablespace_id, const ib_sdi_key_t *ib_sdi_key,
                                   << " Error returned: " << err
                                   << " by trx->id: " << trx->id;);
 
-    ut_ad(err == DB_SUCCESS || trx_is_interrupted(trx) || !"sdi_insert_failed");
+    ut_ad(err == DB_SUCCESS || err == DB_IO_DECRYPT_FAIL ||
+          trx_is_interrupted(trx) || !"sdi_insert_failed");
```
laurynas-biveinis, line 2985: preexisting of course, but !"sdi_insert_failed" - is this an attempt to provide some sort of message in the output? I am not sure if it's worthy enough to keep

laurynas-biveinis (approved):

storage/innobase/api/api0api.cc
satya-bodapati: Hey its me who has added this message :) I can give the background around this.
The assert initially is about crashing when SDI operations fail. They cannot fail under normal circumstances.

But with some debug options (innodb_trx_undo_n_slots_debug or something like that), the assert here crashed. Many bugs are raised by percona (Roel). With that settings, it crashed at other places too apart from SDI operations. (IMHO, that debug setting was a very specific purpose)

The fix is tolerate such failures for those weird cases.
So that part of message shouldn't be removed. The string is to provide a unique signature for the assert instead of  regular Assertion 0 failed.

storage/innobase/dict/dict0dd.cc
```diff
@@ -1080,7 +1080,6 @@ dberr_t dd_rename_tablespace(dd::Object_id dd_space_id,
   }
 
   bool fail = client->update(new_space);
-  ut_ad(!fail);
```
satya-bodapati, line 1146: Why this assert is removed?

storage/innobase/include/row0quiesce.h
```diff
@@ -46,6 +46,7 @@ struct trx_t;
 /** The v3 .cfg writes instant column default values in metadata section. */
 #define IB_EXPORT_CFG_VERSION_V3 0x3UL
 #define IB_EXPORT_CFG_VERSION_V1_WITH_RK 0xFFFFFFFF
+#define IB_EXPORT_CFG_VERSION_V3_WITH_RK 0xFFFFFFFD
```
satya-bodapati: RK should be avoided. I see that there is hangover (zsolt also uses this at some places). If it is at only few places in the row0queisce* files, please fix them as part of this PR.

If there are many, please create separate PR for removal of RK in entire code base.

storage/innobase/row/row0import.cc
```diff
@@ -3739,12 +3742,13 @@ dberr_t row_import_for_mysql(dict_table_t *table, dd::Table *table_def,
...
-  if (FSP_FLAGS_GET_ENCRYPTION(fsp_flags)) {
-    err = fil_set_encryption(
-        table->space,
-        cfg.m_is_keyring_encrypted ? Encryption::KEYRING : Encryption::AES,
-        table->encryption_key, table->encryption_iv);
+  /* For encrypted tablespace, set encryption information.
+  If it is keyring encryption encryption iv was already read in fil_ibd_open
+  and the encryption key will be loaded when keys version is read from pages. */
+
+  if (FSP_FLAGS_GET_ENCRYPTION(fsp_flags) && !cfg.m_is_keyring_encrypted) {
```
satya-bodapati: If it is keyring encrypted, where is the encryption information set?

satya-bodapati (changes requested):
Request to add this scenario as part of testcase

CREATE TABLE t1(a INT) ENCRYPTION_KEY_ID=5;
CREATE TABLE t2(a INT) ENCRYPTION_KEY_ID=10;
FLUSH TABLE t2 FOR EXPORT;
(backup t2.ibd, t2.cfg);
ALTER TABLE t1 DISCARD TABLESPACE;
move t2.ibd as t1.ibd
move t2.cfg as t1.cfg
ALTER TABLE t1 IMPORT TABLESPACE;

SHOW CREATE TABLE t1;

satya-bodapati:
Also, do we have a testcase with tablespace has mix type of pages. ie. some MK encrypted + some KEYRING encrypted?

satya-bodapati:
> Introduced IB_EXPORT_CFG_VERSION_V3_WITH_RK for marking tablespaces
> exported with keyring encryption. Please note that V2(IB_EXPORT_CFG_VERSION_V2)
> was skipped because when V2 was added by upstream the import/export
> with keyring encryption was not functional. Thus it is not possible
> for keyring encrypted tablespaces to be exported with V2.
> 
> The error handling was improved, but not all paths were covered and some
> decryption errors can manifest as "missing tablespace" range of errors.
> Please have a look at diff for innodb-bad-key-change2.test.
> The keyring encryption error handling becomes a big part of every
> upstream merge. To mitigate this problem an improvement PS-5635 was
> proposed.

I would like to know what is extra/different with V3 with RK compared to normal MK V3 formats

robgolebiowski:
@satya-bodapati The test you have asked for produces:
SHOW CREATE TABLE t1;
Table>--Create Table
t1>-CREATE TABLE `t1` (
  `a` int(11) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci ENCRYPTION_KEY_ID=5
ENCRYPTION_KEY_ID is read from DD, the same as ENCRYPTION clause.

storage/innobase/dict/dict0dd.cc
robgolebiowski: Because in test : encryption.innodb-bad-key-change2 I am testing : RENAME TABLE t1 TO t1new. t1 is encrypted and keyring with incorrect encryption key is loaded. Decryption of t1 fails and client->update fails. The path where client->update can fail were not tested prior to this test.

storage/innobase/include/row0quiesce.h
robgolebiowski: Changed and created PS-5862: Make sure that rotation key is not mentioned in code base.

storage/innobase/row/row0import.cc
robgolebiowski: It's in fil_ibd_open - few lines above. For Keyring all needed information is read from page0 - thus it can be assigned in fil_ibd_open. For Master Key the transition key is kept in cfp file - it is used to decrypt the encryption key. It is done from row0import level.

robgolebiowski:
@satya-bodapati I have added new tests: import_half_mk_rotated, import_half_rotated, import_while_threads_run.

robgolebiowski:
The difference is that for MK V3 (and earlier versions) server adds cfp file with transition MK. For Keyring encrypted V3_WITH_KEYRING is just like plain V3. It contains all the information that V3 contains + encryption flag. When V3_WITH_KEYRING is read import knows that it should set decryption information for the tablespace and check that crypt_data is written in page0. I also store explicitly encryption flag. We must set encryption flag in fsp_flags to match the one read from page0 when fil_ibd_open is called.

satya-bodapati:
Request to add link to Jenkins result here

robgolebiowski:
@satya-bodapati https://ps80.cd.percona.com/job/percona-server-8.0-param/344/

satya-bodapati:
> The difference is that for MK V3 (and earlier versions) server adds cfp file with transition MK. For Keyring encrypted V3_WITH_KEYRING is just like plain V3. It contains all the information that V3 contains + encryption flag. When V3_WITH_KEYRING is read import knows that it should set decryption information for the tablespace and check that crypt_data is written in page0. I also store explicitly encryption flag. We must set encryption flag in fsp_flags to match the one read from page0 when fil_ibd_open is called.

From V2 onwards, cfg has tablespace flags written (fsp_flags). Isn't this enough to determine if tabelspace is encrypted or not?

storage/innobase/dict/dict0dd.cc
satya-bodapati: Can you retain the assert and add checks if tablespace is readable (decryptable) or not early before doing DD update?

Relying on DD code to know decryption failure isn't recommended. Generally DD calls shouldn't fail and the assert should be retained.

satya-bodapati:
Other questions/requests

1. How will the UUID PR affect this one? Don't you need to update the page header again and create new keys?

 2) Can you also bring this one to latest 8.0.17 

3) Do we really need to handle the mix type import and export (instead of just throwing error "hey retry later") less code is gold :slightly_smiling_face:

may be we should come to this import/export PR after UUID PR is merged

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/755

PR #3243, opened 2019-06-14
Implement PS-5688 (Find a way to make filesort() generate temp files and add a new DEBUG_SYNC point for analyzing their content) (5.7)

percona-ysorokin:
https://jira.percona.com/browse/PS-5688

Added new DEBUG_SYNC point 'after_find_all_keys' inside 'filesort()'
function which allows suspending execution and analyzing the content of the
generated merge sort temporary files.

'main.percona_encrypt_tmp_files' MTR test case extended with a new section
which performs executing SELECT statement on a large table with ORDER BY
clause upon a column without a key. This is executed twice (with and without
'--encrypt-tmp-files' enabled). In both cases the server is suspended in the
middle of the 'filesort()' call and the content of the temporary merge sort
file is grepped for one of the known row values.

### DIFF
--- a/mysql-test/r/percona_encrypt_tmp_files_debug.result
+++ b/mysql-test/r/percona_encrypt_tmp_files_debug.result
@@ -1,0 +1,66 @@
+include/assert.inc [Temporary file encryption must be enabled]
+include/assert.inc [Sort buffer size must be 4M]
+CREATE TEMPORARY TABLE id_table(id INT UNSIGNED);
+LOAD DATA LOCAL INFILE '<pid_file>' INTO TABLE id_table;
+TRUNCATE TABLE id_table;
+CREATE TABLE t1(
+id BIGINT UNSIGNED NOT NULL PRIMARY KEY,
+value CHAR(32) NOT NULL,
+dummy VARCHAR(1024) NOT NULL
+) ENGINE=InnoDB;
+INSERT INTO t1 VALUES (0, MD5(0), REPEAT(MD5(0), 16));
+INSERT INTO t1 SELECT 1 + id, MD5(1 + id), REPEAT(MD5(1 + id), 16) FROM t1 ORDER BY id;
+INSERT INTO t1 SELECT 2 + id, MD5(2 + id), REPEAT(MD5(2 + id), 16) FROM t1 ORDER BY id;
+INSERT INTO t1 SELECT 4 + id, MD5(4 + id), REPEAT(MD5(4 + id), 16) FROM t1 ORDER BY id;
+INSERT INTO t1 SELECT 8 + id, MD5(8 + id), REPEAT(MD5(8 + id), 16) FROM t1 ORDER BY id;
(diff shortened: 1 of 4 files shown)


mysql-test/t/percona_encrypt_tmp_files.test
```diff
@@ -1,5 +1,148 @@
 --source include/have_innodb.inc
 --source include/have_myisam.inc
+--source include/have_grep.inc
+--source include/have_debug_sync.inc
```
laurynas-biveinis: We would still test the feature code under the release build too, even if we don't have any sync guarantees there (and cannot grep for plaintext). Consider splitting a testcase into two, maybe with a shared include file

mysql-test/t/percona_encrypt_tmp_files.test
```diff
@@ -1,5 +1,148 @@
...
+  # when executing the following 'SELECT sql_big_result' statement.
+  SET DEBUG_SYNC = 'after_find_all_keys SIGNAL after_find_all_keys_reached WAIT_FOR after_find_all_keys_signaled';
+  --send SELECT sql_big_result * FROM t1 ORDER BY dummy
+
+  # In the main connection waiting for the SELECT from the aux connection to
+  # reach 'after_find_all_keys' DEBUG_SYNC point and suspend.
+  --connection default
+  SET DEBUG_SYNC = 'now WAIT_FOR after_find_all_keys_reached';
+
+  # Looking for a file descriptor (symbolic link) from '/proc/$pid_no/fd'
```
laurynas-biveinis: Whoa we are deep in --source include/linux.inc territory. Another argument for splitting the testcase, to let mac devs know locally if they broke something too badly

mysql-test/t/percona_encrypt_tmp_files.test
```diff
@@ -1,5 +1,148 @@
...
+  {
+    --let $assert_text = filesort() temporary file must contain unencrypted key
+    --let $assert_count = 1
+  }
+  --source include/assert_grep.inc
+
+  # Letting SLECT from the aux connection continue execution.
+  SET DEBUG_SYNC = 'now SIGNAL after_find_all_keys_signaled';
+
+  # Finalizing executionof the SELECT in the aux connection.
```
laurynas-biveinis:
```suggestion
  # Finalizing execution of the SELECT in the aux connection.
```

laurynas-biveinis (changes requested):

percona-ysorokin:
Updated.

New test scenario added to a separate `main.percona_encrypt_tmp_files_debug` MTR test case with `--source include/linux.inc` and `--source include/have_debug_sync.inc`

laurynas-biveinis (approved):

percona-ysorokin:
Jenkins build link (percona_encrypt_tmp_files_debug --repeat=2)
https://ps57.cd.percona.com/job/percona-server-5.7-param/58/

PR #3246, opened 2019-06-17
PS-5323: Align Keyring encryption with Master Key encryption.

robgolebiowski:
Tablespaces, system tablespace, undo tablespaces can be either KEYRING
encrypted or Master Key encrypted, but cannot be both.

Tablespaces cannot be created with Master Key encryption when online to
KEYRING encryption is turned on. Otherwise the tablespaces would be
immediately re-encrypted with KEYRING encryption, so allowing creating
Master Key encrypted tablespaces in such case does not make sense.

System tablespace Master Key is bootstrap only option. We disallow
enabling online to KEYRING re-encryption (when system tablespace is Master
Key encrypted) unless innodb-sys-tablespace-encrypt is set to
RE_ENCRYPTING_TO_KEYRING.

Undo tablespace Master Key encryption cannot be turned on when online to
KEYRING encryption is turned on, as it would immediately be re-encrypted
with KEYRING encryption. Undo tablespace encryption can be dynamically
changed from Master Key encryption to KEYRING encryption and in the
opposite direction. However, to change from KEYRING encryption to Master
Key encryption the undo tablespaces must first be decrypted by
encryption threads. Otherwise KEYRING encryption information stored in
page0 of undo tablespace would be overwritten by Master Key encryption
information. These would make logs already encrypted with KEYRING
encryption unreadable. This follows the current behavior that online
from Master Key encryption to KEYRING encryption is supported by not the
other way around.


mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
```diff
@@ -0,0 +1,262 @@
...
+
+--echo # Check also that undo tablespace encryption cannot be turned on if undo tablespace
+--echo # is already encrypted with encryption threads and keyring encryption is off.
+--error ER_WRONG_VALUE_FOR_VAR
+SET GLOBAL innodb_undo_log_encrypt=ON;
+SHOW WARNINGS;
+
+SET GLOBAL innodb_encrypt_tables=ONLINE_FROM_KEYRING_TO_UNENCRYPTED;
+
+--echo # Wait max 10 min for key encryption threads to encrypt all spaces, apart from
```
laurynas-biveinis:
```suggestion
--echo # Wait max 10 min for key encryption threads to decrypt all spaces, apart from
```

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
```diff
@@ -0,0 +1,262 @@
+let $MYSQLD_BASEDIR= `select @@basedir`;
```
laurynas-biveinis, line 1: The test has several restarts, what is its typical runtime on an unloaded machine? Fast enough not to be a big test?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,180 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+@param[in]	value	incoming bool
+@return 0 for valid innodb_track_changed_pages */
+static int innodb_undo_log_encryption_validate(THD *thd, SYS_VAR *var,
+                                               void *save,
+                                               struct st_mysql_value *value) {
+  const char *undo_log_encryption_input;
+  char buff[4];
+  int len = sizeof(buff);
+
+  ut_a(save != NULL);
```
laurynas-biveinis:
```suggestion
  ut_a(save != nullptr);
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,180 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+  ut_a(value != NULL);
```
laurynas-biveinis:
```suggestion
  ut_a(value != nullptr);
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,180 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+
+  undo_log_encryption_input = value->val_str(value, buff, &len);
+
+  if (len > 3) return 1;
+
+  bool is_undo_log_encryption_on =
+      !innobase_strcasecmp(undo_log_encryption_input, "ON") ||
+      !innobase_strcasecmp(undo_log_encryption_input, "1");
+
+  if (is_undo_log_encryption_on == false) {
```
laurynas-biveinis:
```suggestion
  if (!is_undo_log_encryption_on) {
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,180 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+    if (innobase_strcasecmp(undo_log_encryption_input, "OFF") != 0 &&
+        innobase_strcasecmp(undo_log_encryption_input, "0") != 0) {
+      return 1;
+    }
+  } else if (Encryption::is_online_encryption_on()) {
+    push_warning_printf(thd, Sql_condition::SL_WARNING, ER_WRONG_ARGUMENTS,
+                        " Undo log cannot be"
```
laurynas-biveinis:
```suggestion
                        "Undo log cannot be"
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,180 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+    undo::spaces->s_lock();
+    for (auto undo_space : undo::spaces->m_spaces) {
+      fil_space_t *space = fil_space_acquire_silent(undo_space->id());
+      ut_ad(space != nullptr);
+      fil_space_crypt_t *crypt_data = space->crypt_data;
+      if (crypt_data != nullptr &&
+          crypt_data->type != CRYPT_SCHEME_UNENCRYPTED) {
+        push_warning_printf(
+            thd, Sql_condition::SL_WARNING, ER_WRONG_ARGUMENTS,
+            " Undo log cannot be"
```
laurynas-biveinis:
```suggestion
            "Undo log cannot be"
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,180 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+  return 0;
+}
+
+static int innodb_sys_tablespace_encyption_validate(
+    THD *thd, SYS_VAR *var, void *save, struct st_mysql_value *value) {
+  const char *innodb_sys_tablespace_encryption_input;
+  char buff[STRING_BUFFER_USUAL_SIZE];
+  int len = sizeof(buff);
+
+  ut_a(save != NULL);
```
laurynas-biveinis:
```suggestion
  ut_a(save != nullptr);
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,180 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+  ut_a(value != NULL);
```
laurynas-biveinis:
```suggestion
  ut_a(value != nullptr);
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,180 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+  uint use = 0;
+  for (; use < array_elements(sys_tablespace_encrypt_names); use++) {
+    if (!innobase_strcasecmp(innodb_sys_tablespace_encryption_input,
+                             sys_tablespace_encrypt_names[use])) {
+      legit_value = true;
+      break;
+    }
+  }
+
+  if (legit_value == false) return 1;
```
laurynas-biveinis:
```suggestion
  if (!legit_value) return 1;
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,180 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+
+  const auto change_to_sys_tablespace_encryption =
+      static_cast<srv_sys_tablespace_encrypt_enum>(use);
+
+  if (change_to_sys_tablespace_encryption == SYS_TABLESPACE_ENCRYPT_OFF) {
```
laurynas-biveinis, line 21981: Maybe make some 2D arrays indexed by srv_sys_tablespace_encrypt and change_to_sys_tablespace_encryption, one bool whether it's allowed, another for the warning message

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,180 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+      return 1;
+    }
+  } else if (Encryption::is_online_encryption_on()) {
+    push_warning_printf(thd, Sql_condition::SL_WARNING, ER_WRONG_ARGUMENTS,
+                        " Undo log cannot be"
+                        " encrypted with Master Key encryption"
+                        " when online to KEYRING encryption is turned ON.");
+    return 1;
+  } else {
+    undo::spaces->s_lock();
```
laurynas-biveinis: Code shared with check_mk_and_keyring_encrypt_exclusion_for_undo, possible to refactor?

laurynas-biveinis (changes requested):

storage/innobase/handler/ha_innodb.cc

storage/innobase/handler/ha_innodb.cc

storage/innobase/handler/ha_innodb.cc

storage/innobase/handler/ha_innodb.cc

storage/innobase/handler/ha_innodb.cc

storage/innobase/handler/ha_innodb.cc

storage/innobase/handler/ha_innodb.cc

storage/innobase/handler/ha_innodb.cc

storage/innobase/handler/ha_innodb.cc
robgolebiowski: Hm, I am not sure. Good idea to be sure that all options are covered, on the other hand, there is a special case for change == ON and srv_sys_tablespace_encrypt == RE_ENCRYPT and it is a quite big one.

laurynas-biveinis: if (special case) special_case(); else matrix(); ?

robgolebiowski: special case means half of the code stays, half goes into matrix. Special case is half of existing code. But if you as a side reader would find the matrix more readable with this "half the code" :) special case then I will implement the matrix. What do you say ?

laurynas-biveinis: No, all I ask you is to consider, and it looks like you have considered :) OK to leave as is

storage/innobase/handler/ha_innodb.cc

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
robgolebiowski: 91511 [ms]

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
robgolebiowski: done

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/177/

laurynas-biveinis (approved):

satya-bodapati (changes requested):
Few requests at higher level

1. Better name for bug. Align X with Y is too generic
2. Re-arrange commit and PR message. Please divide into problem and fix sections. You can describe what is disallowed now with PR
3. Since we choose disallow MK related operations, please consider disallowing ALTER INSTANCE INNODB ROTATE MASTER KEY

satya-bodapati (changes requested):
To turn innodb_encrypt_tables=ONLINE_TO_KEYRING, please mention (in PR/commit) all the MK encryption settings that should be OFF.

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
```diff
@@ -0,0 +1,262 @@
...
+--echo # fault that sys_tablespace_encrypt cannot be changed.
+SET GLOBAL innodb_encryption_threads=0;
+SET GLOBAL innodb_encrypt_tables = OFF;
+
+--error ER_WRONG_VALUE_FOR_VAR
+SET GLOBAL innodb_sys_tablespace_encrypt=ON;
+SHOW WARNINGS;
+
+--error ER_WRONG_VALUE_FOR_VAR
+SET GLOBAL innodb_sys_tablespace_encrypt=OFF;
```
satya-bodapati, line 210: Why is this disallowed? system tablespace is now unencrypted?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -473,6 +473,12 @@ ibool meb_get_checksum_algorithm_enum(const char *algo_name,
 }
 #endif /* !UNIV_HOTBACKUP */
 
+static const char *sys_tablespace_encrypt_names[] = {
+    "OFF", "ON", "RE_ENCRYPTING_TO_KEYRING", NullS};
```
satya-bodapati, line 477: I would suggest to use "MASTER_KEY_TO_KEYRING" (or MK_TO_KEYRING) instead "RE_ENCRYPTING_TO_KEYRING"

storage/innobase/handler/ha_innodb.cc
```diff
@@ -12167,6 +12173,16 @@ bool create_table_info_t::create_option_encryption_is_valid() const {
     return (false);
   }
 
+  if (!table_is_keyring &&
+      Encryption::is_master_key_encryption(m_create_info->encrypt_type.str) &&
+      Encryption::is_online_encryption_on()) {
+    my_printf_error(ER_ILLEGAL_HA_CREATE_OPTION,
+                    "InnoDB: ENCRYPTED='Y' not supported for table because "
+                    "online encryption to KEYRING is turned ON.",
```
satya-bodapati, line 12181: I feel mentioning the variable name here is better and user need not search for the variable to fix. 

My suggestion is "because innodb_encrypt_tables=get_string(srv_encrypt_tables)"

satya-bodapati: Ok to have both too (above text and this)

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,161 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+@param[in]	value	incoming bool
+@return 0 for valid innodb_track_changed_pages */
+static int innodb_undo_log_encryption_validate(THD *thd, SYS_VAR *var,
+                                               void *save,
+                                               struct st_mysql_value *value) {
+  const char *undo_log_encryption_input;
+  char buff[4];
+  int len = sizeof(buff);
+
+  ut_a(save != nullptr);
```
satya-bodapati, line 21924: is it same other update functions? asserts could be debug and on release build, if they are nullptr, we can return 1.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,161 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+  ut_a(value != nullptr);
+
+  undo_log_encryption_input = value->val_str(value, buff, &len);
+
+  if (len > 3) return 1;
+
+  bool is_undo_log_encryption_on =
+      !innobase_strcasecmp(undo_log_encryption_input, "ON") ||
```
satya-bodapati, line 21932: How is this validation done before your change? by Server?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,161 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+      push_warning_printf(
+          thd, Sql_condition::SL_WARNING, ER_WRONG_ARGUMENTS,
+          "System tablespace Master Key encryption can be turned ON only "
+          "at bootstrap. You can still encrypt System tablespace encryption "
+          "with "
+          "KEYRING encryption.");
+      return 1;
+    } else if (srv_sys_tablespace_encrypt ==
+               SYS_TABLESPACE_RE_ENCRYPTING_TO_KEYRING) {
+      // The system tablespace encryption was marked to be re-encrypted to
```
satya-bodapati, line 22014: This comment style is not allowed. Please use  /*  */

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21877,6 +21906,161 @@ static void update_innodb_redo_log_encrypt(THD *thd, SYS_VAR *var,
...
+      if (Encryption::is_online_encryption_on()) {
+        push_warning_printf(thd, Sql_condition::SL_WARNING, ER_WRONG_ARGUMENTS,
+                            "System tablespace cannot be marked as encrypted "
+                            "with Master Key encryption "
+                            "when Online to keyring encryption is turned ON or "
+                            "when system tablespace "
+                            "was already encrypted with KEYRING encryption.");
+        return 1;
+      } else {
+        fil_space_t *space = fil_space_acquire_silent(TRX_SYS_SPACE);
```
satya-bodapati, line 22027: fil_space_get() is sufficient. There is no eviction concept for fil_space_t and this tablespace is never dropped.

storage/innobase/include/srv0srv.h
```diff
@@ -360,8 +360,14 @@ extern bool srv_undo_log_encrypt;
 /** Enable or disable encryption of temporary tablespace.*/
 extern bool srv_tmp_tablespace_encrypt;
 
+enum srv_sys_tablespace_encrypt_enum {
+  SYS_TABLESPACE_ENCRYPT_OFF = 0,
+  SYS_TABLESPACE_ENCRYPT_ON = 1,
```
satya-bodapati, line 365: I think it now makes sense to have three different values as "OFF, MASTER_KEY, KEYRING_KEY" @laurynas-biveinis  what do you think?

storage/innobase/srv/srv0start.cc
```diff
@@ -1902,6 +1902,90 @@ static lsn_t srv_prepare_to_delete_redo_log_files(ulint n_files) {
...
+                             : fil_space_get(undo_space->id());
+    ut_ad(space != nullptr);
+    fil_space_crypt_t *crypt_data = space->crypt_data;
+    if (crypt_data != nullptr && crypt_data->type != CRYPT_SCHEME_UNENCRYPTED) {
+      static const char *err_msg =
+          " Undo log cannot be"
+          " encrypted with Master Key encryption"
+          " while there are undo tablespaces encrypted with "
+          "KEYRING encryption."
+          " Please decrypt them first with encryption threads.";
```
satya-bodapati, line 1968: How? which option to use to decrypt with encryption threads?

storage/innobase/srv/srv0start.cc
```diff
@@ -2786,6 +2889,14 @@ dberr_t srv_start(bool create_new_db, const std::string &scan_directories) {
       return (srv_init_abort(err));
     }
 
+    if (srv_undo_log_encrypt) {
+      err = check_mk_and_keyring_encrypt_exclusion_for_undo(false, nullptr);
```
satya-bodapati, line 2893: I think it makes sense to directly call this function and read the value of srv_undo_log_encrypt inside it

storage/innobase/include/srv0start.h
```diff
@@ -188,4 +188,7 @@ void srv_fatal_error() MY_ATTRIBUTE((noreturn));
 Shutdown all background threads created by InnoDB. */
 void srv_shutdown_all_bg_threads();
 
+dberr_t check_mk_and_keyring_encrypt_exclusion_for_undo(
```
satya-bodapati, line 191: Please add srv_ prefix for all functions introduced in srv0*
add doxygen comments

storage/innobase/srv/srv0start.cc
```diff
@@ -1902,6 +1902,90 @@ static lsn_t srv_prepare_to_delete_redo_log_files(ulint n_files) {
   return (flushed_lsn);
 }
 
+static dberr_t check_online_to_keyring_and_mk_encrypt_exclusions(
+    bool create_new_db) {
+  if (Encryption::is_online_encryption_on()) {
+    if (srv_sys_tablespace_encrypt == SYS_TABLESPACE_ENCRYPT_ON) {
```
satya-bodapati, line 1908: can this if condition and the above if condition be combined?

storage/innobase/srv/srv0start.cc
```diff
@@ -1902,6 +1902,90 @@ static lsn_t srv_prepare_to_delete_redo_log_files(ulint n_files) {
   return (flushed_lsn);
 }
 
+static dberr_t check_online_to_keyring_and_mk_encrypt_exclusions(
```
satya-bodapati, line 1905: srv_ prefix

satya-bodapati (changes requested):
Majority of the effort in this PR seems to be on system tablespace encryption and efforts to handle it as the encryption option is bootstrap only option.

Hopefully it will become dynamic (from upstream) and all these restrictions will be removed :)

storage/innobase/handler/ha_innodb.cc
laurynas-biveinis: That's news to me? Especially when not having the trailing */ saves a line break

storage/innobase/include/srv0srv.h
laurynas-biveinis: @robgolebiowski call

storage/innobase/handler/ha_innodb.cc
satya-bodapati: Yes, this is upstream style. Rarely I have seen such comments with "//".

We can choose to allow it ;) [add to list of acceptable deviations]

PR #3247, opened 2019-06-17
PS-5639 : Slow shutdown with many tables

satya-bodapati:
Problem:
--------
On a database with 2Million tables and with buffer pool size of 20GB or more,
shutdown takes approximately 12mins.

Lot of tables should have auto-inc mutexes.

Analysis:
---------
Perf data collected during shutdown revealead about 82% of time is spent in
mutex_destroy().

With a debug instrumentation patch, it is observed that the mutexes destroyed
are AUTOINC.

Still if all tables with AUTOINC mutex are accessed, there would be cached table
objects (dict_table_t). With good amount of 'table_definition_cache',
'table_open_cache_instances' and 'open_files_limit', many tables could remain
in the cache.

The important observation here is about 1.6Million tables in cache and
Count one mutex per each of those.

Still why ~ 10mins (82% of time) for destruction of these 2.6 Million mutexes?

These are custom mutexes. Implemented by InnoDB and not pthread_mutexes.

For these mutexes, to collect statistics, a Counter is created for each mutex.
Counter has information about spins, waits, calls. One counter is created for
every mutex instance. So for AUTOINC Mutex type, there is one vector which
will hold all Counters.

For AUTOINC, the vector would hold 1.6Million Count elements

During shutdown, each mutex is destroyed and so that removes the stat object(count)
from the huge vector. This is one-by-one removal.

Lets see the complexity of removal of an element from vector of size N. To remove
one element, it is O(N). And to remove N elements, it is ~ O(N^2). Lets calculate
this for 1.6Million. So it would be O(1.6Million ^ 2) = 2.5Trillion operations.

Now we know why it took 82% of shutdown time :)

Fix:
----
Similar problem doesn't exist for Block Mutexes because for stats, it uses aggregrated
counters. See BlockMutexPolicy.

For AUTOINC too, which can be in Millions, it doesn't make sense to collect mutex
stats individually. So use the same strategy as BlockMutex. i.e to use Aggregate Stats

Use ib_bpmutex_t instead of ib_mutex_t for Autoinc mutex of table.

With ib_bpmutex, there is only element(for mutex metric) in LatchMetaData Vector for all autoinc mutexes. So this solution doesn't suffer the consequences of 'huge' vector and removal of elements from it.

### DIFF
--- a/storage/innobase/include/sync0policy.ic
+++ b/storage/innobase/include/sync0policy.ic
@@ -32,12 +32,29 @@ std::string GenericPolicy<Mutex>::to_string() const
 }
 
 template <typename Mutex>
-std::string BlockMutexPolicy<Mutex>::to_string() const
+std::string AggregateMutexStatsPolicy<Mutex>::to_string() const
 {
-	/* I don't think it makes sense to keep track of the file name
-	and line number for each block mutex. Too much of overhead. Use the
-	latch id to figure out the location from the source. */
-	return(sync_mutex_to_string(get_id(), "buf0buf.cc:0"));
+	switch (m_id) {
+
+	case LATCH_ID_BUF_BLOCK_MUTEX:
+		/* I don't think it makes sense to keep track of the file name
+		and line number for each block mutex. Too much of overhead.
+		Use the latch id to figure out the location from the source. */
+		return(sync_mutex_to_string(get_id(), "buf0buf.cc:0"));
+
+	case LATCH_ID_AUTOINC:
+		return(sync_mutex_to_string(
+			get_id(), "dict_table_autoinc_alloc():0"));
+
+	default:
+		/* Currently only block mutexes and autoinc mutexes use aggregrate
+		Latch Stat counters. If any new mutex uses this aggregrate,
+		add entry here */
+		ut_ad(0);
+	}
+
+	ut_ad(0);
+	return("unknown");
 }
 
 #ifdef UNIV_DEBUG
--- a/storage/innobase/dict/dict0dict.cc
+++ b/storage/innobase/dict/dict0dict.cc
@@ -750,7 +750,7 @@ dict_table_autoinc_alloc(
 	void*	table_void)
 {
 	dict_table_t*	table = static_cast<dict_table_t*>(table_void);
-	table->autoinc_mutex = UT_NEW_NOKEY(ib_mutex_t());
+	table->autoinc_mutex = UT_NEW_NOKEY(AutoIncMutex());
 	ut_a(table->autoinc_mutex != NULL);
 	mutex_create(LATCH_ID_AUTOINC, table->autoinc_mutex);
 }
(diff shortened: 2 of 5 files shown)


satya-bodapati:
Shutdown time improved from 10-12mins to 33secs (Results from internal testing)

storage/innobase/include/sync0policy.ic
```diff
@@ -31,13 +31,29 @@ std::string GenericPolicy<Mutex>::to_string() const
 	return(sync_mutex_to_string(get_id(), sync_file_created_get(this)));
 }
 
+/** BlockMutexPolicy should be renamed as AggregateMutexStatsPolcy. This Policy
```
laurynas-biveinis:
```suggestion
/** BlockMutexPolicy should be renamed as AggregateMutexStatsPolicy. This Policy
```

laurynas-biveinis: Grepping suggests about a dozen places to be fixed if we actually proceeded with the rename. Maybe it's not too bad for maintenance?

laurynas-biveinis (commented):
What is the tradeoff? Do we lose the actual ability to pinpoint the exact table with hot autoinc?

satya-bodapati:
> What is the tradeoff? Do we lose the actual ability to pinpoint the exact table with hot autoinc?

SHOW ENGINE INNODB MUTEX output is different with aggregated metrics.

Before patch, there is one row per autoinc mutex. (Note that it doesn't show table name)
After patch, there is only one row for all AUTOINC mutexes

storage/innobase/include/sync0policy.ic
satya-bodapati: Ok, If you think there is no merge trouble, I can go ahead.

IMO too, this code doesn't change much. Meaning not many new mutexes will use this BlockMutexPolicy. So we can rename this.

laurynas-biveinis:
> Before patch, there is one row per autoinc mutex. (Note that it doesn't show table name)
> After patch, there is only one row for all AUTOINC mutexes

Perfect then

storage/innobase/include/sync0policy.ic
```diff
@@ -32,12 +32,26 @@ std::string GenericPolicy<Mutex>::to_string() const
...
+		and line number for each block mutex. Too much of overhead.
+		Use the latch id to figure out the location from the source. */
+		return(sync_mutex_to_string(get_id(), "buf0buf.cc:0"));
+
+	case LATCH_ID_AUTOINC:
+		return(sync_mutex_to_string(
+			get_id(), "dict_table_autoinc_alloc():0"));
+
+	default:
+		/* Currently only block mutexes and autoinc mutexes uses aggregrate
```
laurynas-biveinis:
```suggestion
		/* Currently only block mutexes and autoinc mutexes use aggregrate
```

laurynas-biveinis (approved):
A comment typo found, no need for further reviews

PR #3248, opened 2019-06-17
PS-5689: Add an MTR testcase to test encryption of temporary tables c…

gl-sergei:
…reated during online alter

Added test which produce temporary table during online ALTER and
checks that table is encrypted by looking at tablespace flags.


gl-sergei:
https://ps57.cd.percona.com/view/5.7/job/percona-server-5.7-param/53/

laurynas-biveinis (approved):

PR #3249, opened 2019-06-17
Fix PS-5689 in 8.0

gl-sergei:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/174/


gl-sergei:
self approve since 5.7 version is approved [here](https://github.com/percona/percona-server/pull/3248) and there are'n any changes between 8.0 and 5.7

PR #3251, opened 2019-06-18
PS-5697: Improve redo log encryption error reporting (8.0)

dutow:
* If encryption can't be turned on during startup, the server will abort
* If encryption can't be turned on when it's requested in a connection,
  an error message will be returned to the client.

This results in a different behavior than MySQL 8.0, which silently
ignores these scenarios.

### DIFF
--- a/mysql-test/include/percona_log_encrypt_change.inc
+++ b/mysql-test/include/percona_log_encrypt_change.inc
@@ -24,19 +24,47 @@ SELECT @@innodb_redo_log_encrypt;
 
 # Test: using a different parameter during restart
 
---let $restart_parameters = restart: $KEYRING_PARAMS --innodb-redo_log_encrypt=$LOG_ENCRYPT_OTHER_TYPE
---replace_result $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT --plugin-dir=KEYRING_PLUGIN_PATH
---replace_regex /\.dll/.so/
---source include/restart_mysqld_no_echo.inc
+--let MYSQLD_DATADIR = `SELECT @@datadir`
+--let MYSQLD_LOG = $MYSQL_TMP_DIR/server.log
 
-SELECT @@innodb_redo_log_encrypt;
+--source include/shutdown_mysqld.inc
+
+--let NEW_CMD = $MYSQLD $extra_args $KEYRING_PARAMS --datadir=$MYSQLD_DATADIR --innodb_redo_log_encrypt=$LOG_ENCRYPT_OTHER_TYPE >$MYSQLD_LOG 2>&1;
+--error 2
+--exec $NEW_CMD
+
+# Test: crash the server AND restart with a different parameter
+--source include/start_mysqld_no_echo.inc
+
+--let $assert_text = Check that there is a warning in the error log
+--let $assert_select = Redo log encryption mode can't be switched without stopping the server and recreating the redo logs
+--let $assert_file = $MYSQLD_LOG
+--let $assert_count = 1
+--source include/assert_grep.inc
+
+--remove_file $MYSQLD_LOG
+
+CREATE TABLE t1(c1 INT, c2 char(20)) ENCRYPTION="Y" ENGINE = InnoDB;
+INSERT INTO t1 VALUES(0, "aaaaa");
+INSERT INTO t1 VALUES(1, "bbbbb");
+INSERT INTO t1 VALUES(2, "ccccc");
+
+--source include/kill_mysqld.inc
+
+--let NEW_CMD = $MYSQLD $extra_args $KEYRING_PARAMS --datadir=$MYSQLD_DATADIR --innodb_redo_log_encrypt=$LOG_ENCRYPT_OTHER_TYPE >$MYSQLD_LOG 2>&1;
+--error 2
+--exec $NEW_CMD
+
+--source include/start_mysqld_no_echo.inc
 
 --let $assert_text=Check that there is a warning in the error log
 --let $assert_select=Redo log encryption mode can't be switched without stopping the server and recreating the redo logs
---let $assert_file=$MYSQLTEST_VARDIR/log/mysqld.1.err
---let $assert_count=3
+--let $assert_file=$MYSQLD_LOG
+--let $assert_count=1
 --let $assert_only_after=CURRENT_TEST: innodb.percona_log_encrypt_change_$LOG_ENCRYPT_TEST_END
 --source include/assert_grep.inc
 
 # Cleanup
+DROP TABLE t1;
 --eval SET GLOBAL innodb_redo_log_encrypt=$LOG_ENCRYPT_TYPE
+--remove_file $MYSQLD_LOG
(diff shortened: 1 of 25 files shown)


mysql-test/include/percona_log_encrypt_change.inc
```diff
@@ -20,23 +20,48 @@ SELECT @@innodb_redo_log_encrypt;
 --let $assert_file=$MYSQLTEST_VARDIR/log/mysqld.1.err
 --let $assert_count=2
 --let $assert_only_after=CURRENT_TEST:innodb.percona_log_encrypt_change_$LOG_ENCRYPT_TEST_END
---source include/assert_grep.inc
+#--source include/assert_grep.inc
```
laurynas-biveinis: Don't comment out, remove

mysql-test/suite/innodb/t/percona_log_encrypt_failure.test
```diff
@@ -1,11 +1,20 @@
-call mtr.add_suppression("Encryption can't find master key, please check the keyring plugin is loaded.");
-call mtr.add_suppression("Can't set redo log tablespace to be encrypted.");
-select @@innodb_redo_log_encrypt;
 
 
---let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err
+
+--let MYSQLD_DATADIR = `SELECT @@datadir`
+--let MYSQLD_LOG=$MYSQL_TMP_DIR/server.log
```
laurynas-biveinis, line 2: Inconsistent spaces around =

storage/innobase/include/ut0ut.h
```diff
@@ -691,7 +691,8 @@ class fatal : public logger {
   @param[in]	err		Error code from errmsg-*.txt.
   @param[in]	args		Variable length argument list */
   template <class... Args>
-  explicit fatal(int err, Args &&... args) : logger(ERROR_LEVEL, err) {
+  explicit fatal(int err, Args &&... args) : logger(ERROR_LEVEL) {
```
laurynas-biveinis, line 708: Please explain in the commit message

share/errmsg-utf8.txt
```diff
@@ -17967,16 +17967,16 @@ ER_IB_MSG_1241
   eng "%s"
 
 ER_IB_MSG_1242
-  eng "%s"
+  eng "Can't set redo log tablespace to be encrypted in read-only mode."
```
laurynas-biveinis: Why?

laurynas-biveinis (changes requested):

share/errmsg-utf8.txt
dutow: This is from 8.0.16, which added these error codes.

mysql-test/include/percona_log_encrypt_change.inc
```diff
@@ -24,19 +24,44 @@ SELECT @@innodb_redo_log_encrypt;
 
 # Test: using a different parameter during restart
 
---let $restart_parameters = restart: $KEYRING_PARAMS --innodb-redo_log_encrypt=$LOG_ENCRYPT_OTHER_TYPE
---replace_result $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT --plugin-dir=KEYRING_PLUGIN_PATH
---replace_regex /\.dll/.so/
---source include/restart_mysqld_no_echo.inc
+--let MYSQLD_DATADIR = `SELECT @@datadir`
+--let MYSQLD_LOG=$MYSQL_TMP_DIR/server.log
```
laurynas-biveinis:
```suggestion
--let MYSQLD_LOG = $MYSQL_TMP_DIR/server.log
```

mysql-test/include/percona_log_encrypt_change.inc
```diff
@@ -24,19 +24,44 @@ SELECT @@innodb_redo_log_encrypt;
...
+
+--source include/kill_mysqld.inc
+
+--let NEW_CMD = $MYSQLD $extra_args $KEYRING_PARAMS --datadir=$MYSQLD_DATADIR --innodb_redo_log_encrypt=$LOG_ENCRYPT_OTHER_TYPE >$MYSQLD_LOG 2>&1;
+--error 2
+--exec $NEW_CMD
+
+--source include/start_mysqld_no_echo.inc
+
+--let $assert_text=Check that there is a warning in the error log
```
laurynas-biveinis, line 60: Spaces around =

mysql-test/suite/innodb/t/percona_log_encrypt_failure.test
```diff
@@ -1,11 +1,16 @@
...
+--let MYSQLD_DATADIR=`SELECT @@datadir`
+--let MYSQLD_LOG=$MYSQL_TMP_DIR/server.log
 
+--source include/shutdown_mysqld.inc
 
---let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err
---let SEARCH_PATTERN=Can't set redo log tablespace to be encrypted.
---source include/search_pattern.inc
---let SEARCH_PATTERN=Encryption can't find master key, please check the keyring plugin is loaded.
+--let NEW_CMD = $MYSQLD $extra_args $KEYRING_PARAMS --datadir=$MYSQLD_DATADIR --innodb_redo_log_encrypt=MASTER_KEY >$MYSQLD_LOG 2>&1;
```
laurynas-biveinis:
```suggestion
--let NEW_CMD=$MYSQLD $extra_args $KEYRING_PARAMS --datadir=$MYSQLD_DATADIR --innodb_redo_log_encrypt=MASTER_KEY >$MYSQLD_LOG 2>&1;
```

share/errmsg-utf8.txt
laurynas-biveinis: Ack, but please explain in the commit message

laurynas-biveinis (changes requested):

laurynas-biveinis (changes requested):
8.0.16 is now in trunk - please rebase and edit commit message accordingly (no more "backporting from 8.0.16")

laurynas-biveinis (approved):

mysql-test/suite/innodb/r/log_encrypt_1_mk.result
```diff
@@ -1,4 +1,6 @@
 SET GLOBAL innodb_redo_log_encrypt = MASTER_KEY;
+Warnings:
+Warning	13068	InnoDB: Can't set redo log tablespace to be encrypted.
```
satya-bodapati, line 3: Can we add a bit more?
These messages  "Blah X failed" are not helpful to user.

it should be more like

"Blah X failed because of Y reason. Please check keyring plugin is loaded or not" Optionally some variable names can also be mentioned.

dutow: We could... with even more refactoring. There are two limitations here:

* most methods only return a boolean status
* actual error could be several levels below in the call chain (in this case: log_write_encryption -> log_file_header_fill_encryption -> fsp_header_fill_encryption_info)

In end-user error cases, I would give it 99% that it's the missing keyring - but we don't know that for sure when we generate this error message.

So I would leave this as is for now, and maybe change it later in a different PR? Changing error messages wouldn't be API breaking.

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/239/

PR #3257, opened 2019-06-21
Fix PS-5722 (rocksdb.ttl_primary testcase sometimes failed)

louishust:
update ttl column may be ignored if the ttl column value not changed.

### DIFF
--- a/mysql-test/suite/rocksdb/t/ttl_primary.test
+++ b/mysql-test/suite/rocksdb/t/ttl_primary.test
@@ -487,6 +487,9 @@ INSERT INTO t1 values (3, UNIX_TIMESTAMP());
 INSERT INTO t1 values (5, UNIX_TIMESTAMP());
 INSERT INTO t1 values (7, UNIX_TIMESTAMP());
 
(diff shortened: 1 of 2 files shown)


mysql-test/suite/rocksdb/t/ttl_primary.test
```diff
@@ -487,6 +487,8 @@ INSERT INTO t1 values (3, UNIX_TIMESTAMP());
 INSERT INTO t1 values (5, UNIX_TIMESTAMP());
 INSERT INTO t1 values (7, UNIX_TIMESTAMP());
 
+# sleep 1 second to make sure column b value changed
+--sleep 1
 set global rocksdb_debug_ttl_rec_ts = 300;
```
george-lorch, line 493: Using sleeps is not the ideal way to make a test deterministic.  Use something like wait_condition.inc to pause the test until the condition is met.

louishust: Thanks for the advice, changed the patch using wait_condition, 
plz review it @georgelorchpercona

george-lorch (approved):

PR #3269, opened 2019-06-28
WIP PS-5102: MyRocks errors after 8.0.16 merge

dutow:
* Regular expressions are now based on std::regex
* Added include my_systime.h to get declaration of my_sleep
* Fix initialization of PSI_stage_info_v1 structs that are now C++14-ized
* Fix use of Item::print as it now needs a const THD* passed in.
* Added include of mysql/psi/mysql_rwlock.h to re-gain rwlock types/definitions.
* Added include of sql/sql_table.h to re-gain declaration of ST_FIELD_INFO
* Fix initialization of PSI thread, mutex, rwlock, cond, and memory structs that all had new members added.
* Rerecorded many testcases where warnings now refer to different rows, but are still correct.
* Rerecorded 2 tests (collation_exceptions_lctn_0, collation_exceptions_lctn_1) where additional warnings are reported now
* Rerecorded rocksdb.rocksdb, where an explain changed (impossible where) (?)
* Rerecorded rocksdb_cf_per_partition: information_schema.rocksdb_cfstats now has an upper case value
* Rerecorded type_decimal: warning formatting changed
* Fixed rocksdb.create_table: ENCRYPTION="y" returns a different error code
* Fixed a syntax error in rocksdb.select and rerecorded it

Remaining MTR failures:

* rocksdb.unique_sec_rev_cf - Using Gap Lock without full unique key in multi-table or multi-statement transactions is not allowed.
* rocksdb.unique_sec - Using Gap Lock without full unique key in multi-table or multi-statement transactions is not allowed.
* rocksdb.bulk_load_sk - select count(*) with force index returns a different result
* upgrade_parts_from_prev_ver - mysql_upgrade deprecated


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6186,21 +6180,20 @@ void ha_rocksdb::free_key_buffers() {
...
+    std::regex_constants::syntax_option_type flags =
+        std::regex::nosubs | std::regex::extended;
+    if (lower_case_table_names) {
+      flags |= std::regex::icase;
+    }
+    std::regex regex_handler(wl, flags);
 
-  if (!regex_handler.compile(wl, flags, table_alias_charset)) {
-    warn_about_bad_patterns(regex_handler, "read_free_rpl_tables");
+    std::smatch matches;
```
laurynas-biveinis: Move out of try block so that it contains only potentially throwing operations (initializing flags is OK)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4280,12 +4280,6 @@ static int rocksdb_init_func(void *const p) {
   mysql_mutex_init(rdb_mem_cmp_space_mutex_key, &rdb_mem_cmp_space_mutex,
                    MY_MUTEX_INIT_FAST);
 
-#if defined(HAVE_PSI_INTERFACE)
-  rdb_collation_exceptions = new Regex(key_rwlock_collation_exception_list);
```
laurynas-biveinis, line 4284: Is this key ever used then?

laurynas-biveinis (changes requested):

storage/rocksdb/ha_rocksdb.cc
dutow: Done.

storage/rocksdb/ha_rocksdb.cc
dutow: No, I removed it.

george-lorch:
With my_regex gone, we also need to remove the wrapper file for my_regexp at sql/sql_regex.h

george-lorch (approved):
The remaining failing mtr tests are known issues that need to be diagnosed and fixed separately from the 8.0.16 merge.  Everything looks good, please merge.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: I don't see any difference from previous code?

laurynas-biveinis:
sql/sql_regex.h removal can happen in this PR or separately

PR #3275, opened 2019-07-01
WIP PS-5102: Fix tokudb errors after 8.0.16 merge

dutow:
(WIP because the remaining failures)

* tokudb_backup now uses std::regex
* rerecorded two date related tests because of different timestamp formatting
* removed invalid date tests from the date test, as those result in an error now
* updated the tokubackup cmake minimum version to 3.0.0 to suppress a CMP0048 policy warning                                                                                    

Remaining failures:
* tokudb.change_column_all_1000_10 timeouts
* tokudb.change_column_all_1000_1 timeouts
* tokudb.upgrade_row_format_from_prev_ver fails because mysql_upgrade is now deprecated


mysql-test/suite/tokudb/t/type_date.test
```diff
@@ -180,52 +180,6 @@ EXPLAIN SELECT COUNT(*) FROM t1 WHERE a = '2015-12-08 02:00:09';
 
 DROP TABLE t1;
 
-#
-# Bug#28687: Search fails on '0000-00-00' date after sql_mode change
```
laurynas-biveinis: What happened to this test snippet in the core server?

plugin/tokudb-backup-plugin/tokudb_backup.cc
```diff
@@ -1141,15 +1144,17 @@ static void tokudb_backup_run(THD *thd, const char *dest_dir) {
...
-    if (regr) {
+    try {
+      exclude_re =
+          new std::regex(exclude_string, std::regex_constants::extended);
+    } catch (std::regex_error const &re) {
       error = EINVAL;
       char reg_error[100 + strlen(exclude_string)];
       snprintf(reg_error, sizeof reg_error,
-               "tokudb backup exclude %s regcomp %d", exclude_string, regr);
+               "tokudb backup exclude %s regcomp %s", exclude_string,
```
laurynas-biveinis: Not longer correct to refer to "regcomp" in the error message?

plugin/tokudb-backup-plugin/tokudb_backup.cc
```diff
@@ -1183,7 +1188,7 @@ static void tokudb_backup_run(THD *thd, const char *dest_dir) {
       tokudb_backup_before_stop_capt_fun, thd,
       tokudb_backup_after_stop_capt_fun, &asce);
 
-  if (exclude_string) my_regfree(&exclude_re);
+  if (exclude_copy_extra.re) delete exclude_copy_extra.re;
```
laurynas-biveinis: delete nullptr is fine

```suggestion
  delete exclude_copy_extra.re;
```

laurynas-biveinis (changes requested):

mysql-test/suite/tokudb/t/type_date.test
dutow: I've changed it to match it

mysql-test/suite/tokudb/t/type_date.test
```diff
@@ -187,16 +187,18 @@ DROP TABLE t1;
 CREATE TABLE t1 (a DATE);
 CREATE TABLE t2 (a DATE);
 CREATE INDEX i ON t1 (a);
---error ER_TRUNCATED_WRONG_VALUE
-INSERT INTO t1 VALUES ('0000-00-00'),('0000-00-00');
---error ER_TRUNCATED_WRONG_VALUE
-INSERT INTO t2 VALUES ('0000-00-00'),('0000-00-00');
+INSERT IGNORE INTO t1 VALUES ('0000-00-00'),('0000-00-00');
+INSERT IGNORE INTO t2 VALUES ('0000-00-00'),('0000-00-00');
+SET sql_mode = 'ALLOW_INVALID_DATES';
```
george-lorch, line 192: I am not sure this is right.  This is changing the nature of this test sequence.  It seems that we want to see that an error is being kicked back out when we provide an invalid date, this change inverts that and now explicitly allows the invalid dates to be inserted and they later show up in the result for the next two SELECTs.

mysql-test/suite/tokudb/t/type_date.test
```diff
@@ -207,19 +209,21 @@ DROP TABLE t1,t2;
 # Bug #31928: Search fails on '1000-00-00' date after sql_mode change
 #
 
+SET sql_mode = 'ALLOW_INVALID_DATES';
```
george-lorch, line 212: Same here, we are no explicitly allowing invalid dates where before we wanted to see the behavior of using an invalid date.

mysql-test/suite/tokudb/t/type_date.test
```diff
@@ -207,19 +209,21 @@ DROP TABLE t1,t2;
...
-INSERT INTO t1 VALUES ('1000-00-00'),('1000-00-00');
---error ER_TRUNCATED_WRONG_VALUE
-INSERT INTO t2 VALUES ('1000-00-00'),('1000-00-00');
+INSERT IGNORE INTO t1 VALUES ('1000-00-00'),('1000-00-00');
+INSERT IGNORE INTO t2 VALUES ('1000-00-00'),('1000-00-00');
 SELECT * FROM t1 WHERE a = '1000-00-00';
 SELECT * FROM t2 WHERE a = '1000-00-00';
 SET SQL_MODE=TRADITIONAL;
 replace_column 9 NA;
+--error ER_WRONG_VALUE
```
george-lorch, line 222: And here same/similar thing, here I think is where we would want to allow invalid values so that we can get the EXPLAIN result rather than an error.

george-lorch (changes requested):

mysql-test/suite/tokudb/t/type_date.test
dutow: This only changes the mode after the inserts. The block below this is different, there the inserts should succeed. This is how main.type_date looks like now (at least for these two blocks - there are more differences I didn't copy)

george-lorch: OK, that's fine then.

george-lorch (approved):

dutow:
Update: updated the tokubackup cmake minimum version to 3.0.0 to suppress a CMP0048 policy warning

laurynas-biveinis (approved):

laurynas-biveinis:
> Update: updated the tokubackup cmake minimum version to 3.0.0 to suppress a CMP0048 policy warning

This change is not present in the diff

dutow:
Edit: uh, that's in a submodule. It's in https://github.com/percona/Percona-TokuBackup/pull/89

laurynas-biveinis (approved):

PR #3280, opened 2019-07-02
PS-5736 - Make innodb_temp_tablespace_encrypt truly dynamic

gl-sergei:
PS-5734 - Disabling temp tablespace encryption at runtime does
not create un-encrypted file-per-table temp tables

Problem:

  innodb_temp_tablespace_encrypt behavior is confusing in two ways:

  1. User can turn it ON dynamically, but cannot turn it off
  2. File-per-table temporary tables obey the setting for system
     temporary tablespace

Fix:

  Make innodb_temp_tablespace_encrypt dynamic. It means that:

  1. User can turn it ON and OFF any time.
  2. Once turned ON, it generates the encryption key for system
     temporary tablespace and starts encrypting all pages written into
     system temporary tablespace.
  3. Once turned OFF, all pages are written to system temporary
     tablespace without encryption. Tablespace keys are not erased so
     that already encrypted pages can be decrypted.
  4. Changing innodb_temp_tablespace_encrypt affects CREATE TEMPORARY
     TABLE:
     - when ON: CREATE TEMPORARY TABLE t ENCRYPTION='n' fails
     - when ON: CREATE TEMPORARY TABLE t fails
     - when OFF: CREATE TEMPORARY TABLE t ENCRYPTION='y' fails
  5. Changing innodb_temp_tablespace_encrypt does not affect
     - CREATE TEMPORARY TABLE t ROW_FORMAT=COMPRESSED
     - CREATE TEMPORARY TABLE t KEY_BLOCK_SIZE=n
     File-per-table temporary tables will be encrypted when
     ENCRYPTION='y' is specified.

### DIFF
--- a/mysql-test/suite/innodb/include/temp_table_encrypt.inc
+++ b/mysql-test/suite/innodb/include/temp_table_encrypt.inc
@@ -17,11 +17,10 @@ INSERT INTO t01 VALUES ('Lorem ipsum dolor sit amet, consectetur adipiscing elit
 --source include/wait_condition.inc
 
 # make sure ibtmp1 is not encrypted
-
---let $grep_pattern= Lorem ipsum dolor sit amet
---let $grep_file= $MYSQL_DATA_DIR/ibtmp1
---let $grep_output= boolean
---source include/grep_pattern.inc
+--let SEARCH_PATTERN= Lorem ipsum dolor sit amet
+--let SEARCH_FILE= $MYSQL_DATA_DIR/ibtmp1
+--let ABORT_ON= NOT_FOUND
+--source include/search_pattern_in_file.inc
 
 
 # this table created in separate file per table tablespace, make sure it not
@@ -36,21 +35,14 @@ INSERT INTO t04 VALUES ('Praesent tristique eros a tempus fringilla');
 --let $t04_fn= `SELECT CONCAT(@@tmpdir, '/', NAME, '.ibd') FROM INFORMATION_SCHEMA.INNODB_TEMP_TABLE_INFO WHERE NAME LIKE '#%';`
 
 # make sure t04 is not encrypted
-
---let $grep_pattern= Praesent tristique eros a tempus fringilla
---let $grep_file= $t04_fn
---let $grep_output= boolean
---source include/grep_pattern.inc
+--let SEARCH_PATTERN= Praesent tristique eros a tempus fringilla
+--let SEARCH_FILE= $t04_fn
+--let ABORT_ON= NOT_FOUND
+--source include/search_pattern_in_file.inc
 
 DROP TABLE t04;
 
 SET GLOBAL innodb_temp_tablespace_encrypt = ON;
-
---let $innodb_master_loops= `SELECT SUM(VARIABLE_VALUE) FROM performance_schema.global_status WHERE VARIABLE_NAME LIKE 'innodb_master%loops'`
-
-# wait for couple of master thread turnarounds
---let $wait_condition= SELECT SUM(VARIABLE_VALUE) > $innodb_master_loops + 2 FROM performance_schema.global_status WHERE VARIABLE_NAME LIKE 'innodb_master%loops'
---source include/wait_condition.inc
 
 CREATE TEMPORARY TABLE t02 (a INT) ENGINE=InnoDB ENCRYPTION='Y';
 
@@ -62,11 +54,10 @@ INSERT INTO t03 VALUES ('Curabitur laoreet, velit non interdum venenatis');
 --source include/wait_condition.inc
 
 # make sure ibtmp1 is encrypted now
-
---let $grep_pattern= Curabitur laoreet, velit non interdum venenatis
---let $grep_file= $MYSQL_DATA_DIR/ibtmp1
---let $grep_output= boolean
---source include/grep_pattern.inc
+--let SEARCH_PATTERN= Curabitur laoreet, velit non interdum venenatis
+--let SEARCH_FILE= $MYSQL_DATA_DIR/ibtmp1
+--let ABORT_ON= FOUND
+--source include/search_pattern_in_file.inc
 
 # this table created in separate file per table tablespace, make sure it is
 # encrypted as well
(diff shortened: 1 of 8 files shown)


gl-sergei:
@robgolebiowski @satya-bodapati @mohitj1988 please check if this behavior is what we want

gl-sergei:
https://ps57.cd.percona.com/job/percona-server-5.7-param/80/

mysql-test/suite/innodb/include/temp_table_encrypt.inc
```diff
@@ -181,3 +204,60 @@ INSERT INTO t1 VALUES (3);
...
+CREATE TEMPORARY TABLE tmp3 (a TEXT);
+INSERT INTO tmp3 VALUES ('Sed sodales ligula sed enim condimentum');
+
+--let $wait_condition= SELECT variable_value = 0 FROM performance_schema.global_status WHERE LOWER(variable_name) = 'innodb_buffer_pool_pages_dirty'
+--source include/wait_condition.inc
+
+--let $grep_pattern= Nam vestibulum mauris massa
+--let $grep_file= $MYSQL_DATA_DIR/ibtmp1
+--let $grep_output= boolean
+--source include/grep_pattern.inc
```
satya-bodapati: Please use include/search_pattern_in_file.inc instead of grep_pattern.inc.
search_pattern_in_file has cool feature "let ABORT_ON=FOUND;" or NOT_FOUND

With search_pattern_in_file.inc aborts instead of showing result content mismatch (with grep_pattern.inc). When the test aborts with all logs .ibd file, it much more useful in debugging.

Please fix all instances of grep_pattern.inc

satya-bodapati (changes requested):

storage/innobase/fil/fil0fil.cc
```diff
@@ -7948,66 +7948,72 @@ fil_set_encryption(
...
-{
-	bool success = true;
-	if (space->encryption_type == Encryption::AES) {
-		mtr_t mtr;
-		mtr_start(&mtr);
-
-		if (fsp_is_system_temporary(space->id)) {
-			mtr_set_log_mode(&mtr, MTR_LOG_NO_REDO);
-		}
+	static bool fil_encryption_rotate_low(const fil_space_t *space) {
```
satya-bodapati: The indentation of the entire function is wrong now

satya-bodapati: Actually why this function formatting is touched? Seems no new changes?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -12273,6 +12279,9 @@ create_table_info_t::create_option_encryption_is_valid() const
 	ulint		fsp_flags = space->flags;
 
 	bool tablespace_is_encrypted = FSP_FLAGS_GET_ENCRYPTION(fsp_flags);
+	if (fsp_is_system_temporary(space_id)) {
```
satya-bodapati, line 12282: Why can't we have proper fsp_flags for temp tablespace? then we don't need this special condition

storage/innobase/srv/srv0srv.cc
```diff
@@ -2774,29 +2774,15 @@ srv_temp_encryption_update(bool enable)
...
-		}
-
-	} else {
-		if (!is_encrypted) {
-			/* Encryption already disabled */
-			return(DB_SUCCESS);
-		} else {
-			// TODO: Disabling encryption is not allowed yet
-			return(DB_SUCCESS);
+	if (enable != is_encrypted) {
```
satya-bodapati, line 2777: Good work on the simplification!

storage/innobase/fil/fil0fil.cc
gl-sergei: ¯\_(ツ)_/¯ probably pressed TAB in the wrong place, will revert this change

storage/innobase/handler/ha_innodb.cc
gl-sergei: Then we will need similar condition in other places, when we trying to read from temporary tablespace, because pages may be or may be not encrypted regardless of the tablespace flag.

mysql-test/suite/innodb/include/temp_table_encrypt.inc
gl-sergei: OK. Will use search_pattern_in_file instead

gl-sergei:
Updated:
- reverted formatting of fil_encryption_rotate_low
- use search_pattern_in_file in the test case

https://ps57.cd.percona.com/job/percona-server-5.7-param/85/

storage/innobase/fil/fil0fil.cc
```diff
@@ -7948,33 +7948,42 @@ fil_set_encryption(
...
 {
-	/* Make sure the keyring is loaded. */
-	if (!Encryption::check_keyring()) {
-		ib::error() << "Can't set temporary tablespace"
-			<< " to be encrypted because"
-			<< " keyring plugin is not"
-			<< " available.";
-			return(DB_ERROR);
-	}
+	if (enable && !FSP_FLAGS_GET_ENCRYPTION(space->flags)) {
```
satya-bodapati: I would rewrite it as if (!enable) {return (DB_SUCCESS) } at top. and keep the rest same. No indentation change required and fewer lines changed.

storage/innobase/handler/ha_innodb.cc
satya-bodapati: Ok, the core change to make write to temp tablespace unencrypted is already there before your PR. 

fil_io_set_encryption()
```
  /* For writing temporary tablespace, if encryption for temporary
        tablespace is disabled, skip setting encryption. */
        if (fsp_is_system_temporary(space->id)
            && !srv_tmp_tablespace_encrypt && req_type.is_write()) {
                req_type.clear_encrypted();
                return;
        }
```
The writes are already handled(see above) and this is the main change to allow table creation?

gl-sergei: Yep, writes were already handled in the `fil_io_set_encryption` and it was actually a bug, because it did not match the documented behaviour. I thought it was obvious and didn't mention. Now we turned it into a feature. And this PR only adds handling of CREATE TABLE statement.

satya-bodapati: It is a "hidden feature" :laughing:

storage/innobase/fil/fil0fil.cc
gl-sergei: done and force pushed

satya-bodapati (approved):

PR #3288, opened 2019-07-03
Bug #94747: 4GB Limit on large_pages shared memory set-up (5.7)

dutow:
mysqld --large-pages --innodb_buffer_pool_chunk_size=4G result in 1G chunk allocations instead of 4G.

This patch was originally provided by Daniel Black at https://bugs.mysql.com/bug.php?id=94747

No testcase provided: the server setup requires a specific environment.

### DIFF
--- a/storage/innobase/include/os0proc.h
+++ b/storage/innobase/include/os0proc.h
@@ -45,7 +45,7 @@ extern ulint	os_total_large_mem_allocated;
 extern my_bool	os_use_large_pages;
 
 /** Large page size. This may be a boot-time option on some platforms */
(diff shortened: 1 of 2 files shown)


laurynas-biveinis (commented):
Have you confirmed the fix by manual testing?

dutow:
Yes. I also wanted to add a test case, but this can't be tested without at least a 4GB hugetlb setup, which is not realistic on most test machines.

laurynas-biveinis (approved):

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/95/

PR #3297, opened 2019-07-05
PS-3845, Bug#80821: replication fails with multi-delete and cascade FKs (5.7)

dutow:
Issue:

When a multiple delete statement deletes both the parent and the
child of a cascade foreign key, both deletes are written to the
binary log. When the slave reads this log, it tries to delete
the dependent row twice: first because the foreign key, and after
because of the binlog entry.

This bug was caused by WL#6987, dbb6e38c : before this refactoring,
multi statement deletes executed deletes for one table immediately,
then delayed the remaining entries. After the refactoring, foreign
keys weren't handled.

Fix:

multi statement deletes fall back to deleyed mode when there is a
cascade foreign key dependency between the tables.

### DIFF
--- a/mysql-test/suite/rpl/t/bug80821.test
+++ b/mysql-test/suite/rpl/t/bug80821.test
@@ -1,0 +1,38 @@
+--source include/have_innodb.inc
+--source include/have_binlog_format_row.inc
+--source include/master-slave.inc
+
+CREATE TABLE t0 (
+f0 INT PRIMARY KEY
+) ENGINE=innodb;
+
+CREATE TABLE t1 (
+f1 INT PRIMARY KEY,
+f0 INTEGER,
+FOREIGN KEY (f0)
+REFERENCES t0(f0)
+ON DELETE CASCADE
+) ENGINE=innodb;
+
+INSERT INTO t0 VALUES (0);
+INSERT INTO t0 VALUES (1);
+INSERT INTO t1 VALUES (0, 0);
+INSERT INTO t1 VALUES (1, 0);
+
+# Should delete both t1 values, and the 0 value from t0
+DELETE t0a.*, t1.* FROM t0 AS t0a, t1 WHERE t0a.f0 = 0 AND t1.f1 = 0;
+
+SELECT COUNT(*) = 1 FROM t0;
+SELECT COUNT(*) = 0 FROM t1;
+
+--source include/sync_slave_sql_with_master.inc
+
+SELECT COUNT(*) = 1 FROM t0;
+SELECT COUNT(*) = 0 FROM t1;
+
+--source include/rpl_connection_master.inc
+
+DROP TABLE t1;
+DROP TABLE t0;
+
+--source include/rpl_end.inc
--- a/sql/sql_delete.cc
+++ b/sql/sql_delete.cc
@@ -868,6 +868,58 @@ int Query_result_delete::prepare(List<Item> &values, SELECT_LEX_UNIT *u)
   DBUG_RETURN(0);
 }
 
+/**
+  Test that the two strings are equal, accoding to the lower_case_table_names
+  setting.
+
+  @param  a     A table or database name
+  @param  b     A table or database name
+  @retval bool  True if the two names are equal
+*/
+static bool db_or_table_name_equals(const char* a, const char *b) {
+  return lower_case_table_names ?
+    my_strcasecmp(files_charset_info, a, b) == 0 :
+    strcmp(a, b) == 0;
+}
+
+/**
+  Test that the subject table (of DELETE) has a cascade foreign key
+  parent present in the query.
+
+  @param  thd        thread handle
+  @param  table      table to be checked (must be updatable base table)
+  @param  table_list List of tables to check against
+
+  @retval bool       True if cascade parent found.
+*/
+static bool has_cascade_dependency(THD *thd, const TABLE_LIST &table,
+                                   TABLE_LIST *table_list)
+{
+  DBUG_ASSERT(&table == const_cast<TABLE_LIST &>(table).updatable_base_table());
+
+  List <st_handler_tablename> fk_table_list;
+  List_iterator<st_handler_tablename> fk_table_list_it(fk_table_list);
+
+  table.table->file->get_cascade_foreign_key_table_list(thd, &fk_table_list);
+
+  st_handler_tablename *tbl_name;
+  while ((tbl_name= fk_table_list_it++))
+  {
+    for (TABLE_LIST* curr= table_list; curr; curr= curr->next_local)
+    {
+      const bool same_table_name= 
+        db_or_table_name_equals(curr->table_name, tbl_name->tablename);
+      const bool same_db_name=
+        db_or_table_name_equals(curr->db, tbl_name->db);
+      if (same_table_name && same_db_name)
+      {
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
 
 bool Query_result_delete::initialize_tables(JOIN *join)
 {
@@ -893,7 +945,8 @@ bool Query_result_delete::initialize_tables(JOIN *join)
     TABLE_LIST *const ref= walk->correspondent_table->updatable_base_table();
     delete_table_map|= ref->map();
     if (delete_while_scanning &&
-        unique_table(thd, ref, join->tables_list, false))
+        (unique_table(thd, ref, join->tables_list, false) ||
+         has_cascade_dependency(thd, *ref, join->tables_list)))
     {
       /*
         If the table being deleted from is also referenced in the query,
--- a/mysql-test/suite/rpl/r/bug80821.result
+++ b/mysql-test/suite/rpl/r/bug80821.result
@@ -1,0 +1,37 @@
+include/master-slave.inc
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
+Note	####	Storing MySQL user name or password information in the master info repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START SLAVE; see the 'START SLAVE Syntax' in the MySQL Manual for more information.
+[connection master]
+CREATE TABLE t0 (
+f0 INT PRIMARY KEY
+) ENGINE=innodb;
+CREATE TABLE t1 (
+f1 INT PRIMARY KEY,
+f0 INTEGER,
+FOREIGN KEY (f0)
+REFERENCES t0(f0)
+ON DELETE CASCADE
+) ENGINE=innodb;
+INSERT INTO t0 VALUES (0);
+INSERT INTO t0 VALUES (1);
+INSERT INTO t1 VALUES (0, 0);
+INSERT INTO t1 VALUES (1, 0);
+DELETE t0a.*, t1.* FROM t0 AS t0a, t1 WHERE t0a.f0 = 0 AND t1.f1 = 0;
+SELECT COUNT(*) = 1 FROM t0;
+COUNT(*) = 1
+1
+SELECT COUNT(*) = 0 FROM t1;
+COUNT(*) = 0
+1
+include/sync_slave_sql_with_master.inc
+SELECT COUNT(*) = 1 FROM t0;
+COUNT(*) = 1
+1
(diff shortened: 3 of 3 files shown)


dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/99/

mysql-test/suite/rpl/t/bug80821.test
```diff
@@ -0,0 +1,45 @@
+--source include/not_ndb_default.inc
```
laurynas-biveinis: Why?

mysql-test/suite/rpl/t/bug80821.test
```diff
@@ -0,0 +1,45 @@
...
+--source include/not_group_replication_plugin.inc
```
laurynas-biveinis: Why?

mysql-test/suite/rpl/t/bug80821.test
```diff
@@ -0,0 +1,45 @@
...
+--source include/have_binlog_format_row.inc
+--source include/master-slave.inc
+
+--source include/rpl_connection_master.inc
```
laurynas-biveinis: Likely to be redundant

mysql-test/suite/rpl/t/bug80821.test
laurynas-biveinis: --source include/have_innodb.inc (for 5.7 only)

mysql-test/suite/rpl/t/bug80821.test
```diff
@@ -0,0 +1,45 @@
...
+
+CREATE TABLE t0 (
+f0 INT PRIMARY KEY
+) engine=innodb;
```
laurynas-biveinis: Consistent case here and for CREATE TABLE t1 too please

mysql-test/suite/rpl/t/bug80821.test
```diff
@@ -0,0 +1,45 @@
...
+INSERT INTO t1 VALUES (0, 0);
+INSERT INTO t1 VALUES (1, 0);
+
+# Should delete both t1 values, and the 0 value from t0
+DELETE t0a.*, t1.* FROM t0 AS t0a, t1 WHERE t0a.f0 = 0 AND t1.f1 = 0;
+
+--source include/sync_slave_sql_with_master.inc
+--source include/rpl_connection_master.inc
+
+SELECT COUNT(*) = 1 FROM t0;
```
laurynas-biveinis, line 25: Consider assert.inc

laurynas-biveinis, line 25: "sync slave to master, switch to master, check master, switch to slave, check slave"
could be simplified to
"check master, sync slave to master, check slave"

sql/sql_base.cc
```diff
@@ -2245,6 +2245,43 @@ static TABLE_LIST* find_dup_table(THD *thd, const TABLE_LIST *table,
...
+
+  @param  thd        thread handle
+  @param  table      table to be checked (must be updatable base table)
+  @param  table_list List of tables to check against
+
+  @retval non-NULL The table list element for the table that
+                   represents the cascade parent.
+  @retval NULL     No cascade parents found.
+*/
+TABLE_LIST *no_cascade_dependency(THD *thd, const TABLE_LIST *table,
```
laurynas-biveinis:
```suggestion
TABLE_LIST *no_cascade_dependency(THD *thd, const TABLE_LIST &table,
```

sql/sql_base.cc
```diff
@@ -2245,6 +2245,43 @@ static TABLE_LIST* find_dup_table(THD *thd, const TABLE_LIST *table,
...
+                                  TABLE_LIST *table_list)
+{
+  DBUG_ASSERT(table == ((TABLE_LIST *)table)->updatable_base_table());
```
laurynas-biveinis: const_cast

sql/sql_base.cc
```diff
@@ -2245,6 +2245,43 @@ static TABLE_LIST* find_dup_table(THD *thd, const TABLE_LIST *table,
...
+
+  List <st_handler_tablename> fk_table_list;
+  List_iterator<st_handler_tablename> fk_table_list_it(fk_table_list);
+  st_handler_tablename *tbl_name;
```
laurynas-biveinis: Declare right above the while loop

sql/sql_base.cc
```diff
@@ -2245,6 +2245,43 @@ static TABLE_LIST* find_dup_table(THD *thd, const TABLE_LIST *table,
...
+
+  table->table->file->get_cascade_foreign_key_table_list(thd, &fk_table_list);
+
+  while ((tbl_name= fk_table_list_it++))
+  {
+    for (TABLE_LIST* curr = table_list; curr; curr= curr->next_local)
```
laurynas-biveinis:
```suggestion
    for (TABLE_LIST* curr= table_list; curr; curr= curr->next_local)
```

sql/sql_base.cc
```diff
@@ -2245,6 +2245,43 @@ static TABLE_LIST* find_dup_table(THD *thd, const TABLE_LIST *table,
...
+    {
+      if (strcmp(curr->table_name, tbl_name->tablename) == 0 &&
```
laurynas-biveinis: This is not going to work with different lowercase_table_name settings - check for how that variable is used for canonical way to compare names which takes case insensitivity into account

sql/sql_base.h
```diff
@@ -302,6 +302,8 @@ void free_io_cache(TABLE *entry);
 void intern_close_table(TABLE *entry);
 void close_thread_table(THD *thd, TABLE **table_ptr);
 bool close_temporary_tables(THD *thd);
+TABLE_LIST *no_cascade_dependency(THD *thd, const TABLE_LIST *table,
```
laurynas-biveinis: Is it possible to have it static to sql_delete.cc instead?

sql/sql_delete.cc
```diff
@@ -893,7 +893,9 @@ bool Query_result_delete::initialize_tables(JOIN *join)
     TABLE_LIST *const ref= walk->correspondent_table->updatable_base_table();
     delete_table_map|= ref->map();
     if (delete_while_scanning &&
-        unique_table(thd, ref, join->tables_list, false))
+        (unique_table(thd, ref, join->tables_list, false) ||
+        no_cascade_dependency(thd, ref, join->tables_list)
+	))
```
laurynas-biveinis: ))) in the previous line

sql/sql_base.cc
laurynas-biveinis: The actual TABLE_LIST* value is not used, the return value is immediately cast to bool, so return bool directly.
Consider removing "no_" from function name and return value logic

laurynas-biveinis (changes requested):

mysql-test/suite/rpl/t/bug80821.test
dutow: I've used the testcase from the jira ticket, and these were present in it. I didn't know why, so I kept them. I'll remove them.

mysql-test/suite/rpl/t/bug80821.test
dutow: These selects aren't really required, without the patch the testcase fails at the sync. I kept them here as they were in the original testcase, but I could also simply remove them.

laurynas-biveinis: It's good to keep the selects for correct behavior confirmation on slave at least

dutow:
I kept the selects, as for a simple additional verification I think they are simpler. Everything else updated.

sql/sql_delete.cc
```diff
@@ -869,6 +869,44 @@ int Query_result_delete::prepare(List<Item> &values, SELECT_LEX_UNIT *u)
...
 
+/**
+  Test that the subject table (of DELETE) has a cascade foreign key
+  parent present in the query.
+
+  @param  thd        thread handle
+  @param  table      table to be checked (must be updatable base table)
+  @param  table_list List of tables to check against
+
+  @retval NULL     No cascade parents found.
```
laurynas-biveinis: false/true

sql/sql_delete.cc
```diff
@@ -869,6 +869,44 @@ int Query_result_delete::prepare(List<Item> &values, SELECT_LEX_UNIT *u)
...
+
+  st_handler_tablename *tbl_name;
+  while ((tbl_name= fk_table_list_it++))
+  {
+    for (TABLE_LIST* curr= table_list; curr; curr= curr->next_local)
+    {
+      const bool same_table_name= lower_case_table_names ?
+        my_strcasecmp(files_charset_info, curr->table_name, tbl_name->tablename) :
+        strcmp(curr->table_name, tbl_name->tablename);
+      if (same_table_name && strcmp(curr->db, tbl_name->db) == 0)
```
laurynas-biveinis: As far as I can tell by existing use, lower_case_table_names govern DB names too.
Also I see some inconsistency in existing code whether files_charset_info or system_charset_info should be used, please double-check which is which.

laurynas-biveinis (changes requested):

sql/sql_delete.cc
```diff
@@ -869,6 +869,47 @@ int Query_result_delete::prepare(List<Item> &values, SELECT_LEX_UNIT *u)
...
+  List_iterator<st_handler_tablename> fk_table_list_it(fk_table_list);
+
+  table.table->file->get_cascade_foreign_key_table_list(thd, &fk_table_list);
+
+  st_handler_tablename *tbl_name;
+  while ((tbl_name= fk_table_list_it++))
+  {
+    for (TABLE_LIST* curr= table_list; curr; curr= curr->next_local)
+    {
+      const bool same_table_name= lower_case_table_names ?
```
laurynas-biveinis: IMHO this is now asking for bool db_or_table_name_equal(const char *, const char *) helper (it's fascinating that server code does not have it already)

sql/sql_delete.cc
```diff
@@ -893,7 +934,8 @@ bool Query_result_delete::initialize_tables(JOIN *join)
     TABLE_LIST *const ref= walk->correspondent_table->updatable_base_table();
     delete_table_map|= ref->map();
     if (delete_while_scanning &&
-        unique_table(thd, ref, join->tables_list, false))
+        (unique_table(thd, ref, join->tables_list, false) ||
+        has_cascade_dependency(thd, *ref, join->tables_list)))
```
laurynas-biveinis:
```suggestion
         has_cascade_dependency(thd, *ref, join->tables_list)))
```

laurynas-biveinis (changes requested):

sql/sql_delete.cc
dutow: It is, but it would be either 4 string parameter, or many overloads (TABLE, TABLE_LIST, st_handler_table_name all combinations, for example...)

laurynas-biveinis: I had something in mind with the above signature: 
bool db_or_table_name_equal(const char *, const char *);
which compares only two given DB or table names and would be invoked twice here.
Perhaps the "db_or_table" name part can be improved

dutow: That's a good idea, I thought you meant "and" there.

sql/sql_delete.cc
```diff
@@ -868,6 +868,58 @@ int Query_result_delete::prepare(List<Item> &values, SELECT_LEX_UNIT *u)
...
+  Test that the two strings are equal, accoding to the lower_case_table_names
+  setting.
+
+  @param  a     A table or database name
+  @param  b     A table or database name
+  @retval bool  True if the two names are equal
+*/
+static bool db_or_table_name_equals(const char* a, const char *b) {
+  return lower_case_table_names ?
+    my_strcasecmp(files_charset_info, a, b) == 0:
```
laurynas-biveinis:
```suggestion
    my_strcasecmp(files_charset_info, a, b) == 0 :
```

laurynas-biveinis (changes requested):
One formatting change, otherwise OK if testing OK, no need for further 5.7 reviews

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/108/

PR #3301, opened 2019-07-05
PS-5724: Travis-CI: 8.0.16 MySQL Router does not build cleanly with clang 4 & 5 (8.0)

inikep:
Turn off warnings for MySQL Router for clang-4 and clang-5 using the `-Wno-missing-braces` compiler flag.


percona-ysorokin (approved):
LGTM

PR #3302, opened 2019-07-05
PS-5697: Improve redo log encryption error reporting (5.7)

dutow:
* If encryption can't be turned on during startup, the server will abort
* If encryption can't be turned on when it's requested in a connection,
  an error message will be returned to the client.

This results in a different behavior than MySQL 8.0, which silently
ignores these scenarios.


mysql-test/suite/perfschema/r/show_sanity.result
```diff
@@ -692,6 +692,7 @@ where (show_mode = "JUNK: GLOBAL-ONLY")
 and source = "I_S.SESSION_STATUS")
 order by show_mode, source, variable_name;
 SHOW_MODE	SOURCE	VARIABLE_NAME
+5.7	P_S.SESSION_STATUS	Innodb_encryption_redo_key_version
```
laurynas-biveinis: show_sanity.test needs updating instead?

mysql-test/suite/innodb/t/percona_log_encrypt_failure.test
```diff
@@ -1,11 +1,19 @@
...
+--let NEW_CMD=$MYSQLD $extra_args --datadir=$MYSQLD_DATADIR --innodb_redo_log_encrypt=MASTER_KEY >$MYSQLD_LOG 2>&1;
+--error 1
+--exec $NEW_CMD
+
+--let SEARCH_FILE=$MYSQLD_LOG
+--let SEARCH_PATTERN=Redo log encryption is enabled, but keyring plugin is not loaded.
 --source include/search_pattern.inc
 
+--source include/start_mysqld.inc
+#--remove_file $MYSQLD_LOG
```
laurynas-biveinis: No commented out code, no extra newlines at the end please

storage/innobase/srv/srv0start.cc
```diff
@@ -486,6 +486,13 @@ create_log_files(
 					 alg,
 					 reinterpret_cast<byte*>(mkey->key),
 					 NULL);
+
+
```
```suggestion
```

storage/innobase/log/log0log.cc
```diff
@@ -1143,7 +1143,7 @@ log_read_encryption() {
 		/* Make sure the keyring is loaded. */
 		if (!Encryption::check_keyring()) {
 			ut_free(log_block_buf_ptr);
-			ib::error() << "Redo log was encrypted,"
+			ib::fatal() << "Redo log was encrypted,"
```
laurynas-biveinis, line 1146: I like ib::fatal() here and elsewhere in the patch, just am wondering whether the previous ib::error()/return(false); resulted in eventual server startup error too?

laurynas-biveinis (changes requested):

storage/innobase/log/log0log.cc
dutow: It did, as in this case, this method returns an error and innodb initialization aborts. There were already mtr tests about this.

laurynas-biveinis (approved):

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/110/

PR #3304, opened 2019-07-08
PS-5581: Bloom_filter allocates too much memory (5.7)

dutow:
In the 5.7 port, the extended slow query log was changed: the
Bloom_filter uses a MEM_ROOT instead of new/malloc, reallocating
the filter on clearing. For complex statements, such as stored
procedures calling many statements, or updates calling user functions
for each row, this results in a huge amount of allocations, which
will only be freed at the end of the entire statemment.
(8c9fa78)

This commit changes how clear works: instead of allocating a new
bitset on the next use, the current bitset is cleared.

### DIFF
--- a/sql/sql_class.h
+++ b/sql/sql_class.h
@@ -1598,16 +1598,10 @@ class Bloom_filter {
  public:
   Bloom_filter() : bit_set(NULL) {}
 
-  ~Bloom_filter()
-  {
-    // Do not delete, just destruct, due to MEM_ROOT allocation
-    bit_set->~bitset();
-  }
-
   void clear()
   {
-    // Do not delete, just force new MEM_ROOT allocation on the next use
-    bit_set= NULL;
+    if (bit_set != NULL)
+      bit_set->reset();
   }
 
   /**
@@ -1627,7 +1621,7 @@ class Bloom_filter {
       void *bit_set_place= alloc_root(mem_root, sizeof(Bit_set));
       // FIXME: memory allocation failure is eaten silently. Nonexact stats are
       // the least of the concerns then.
-      if (!bit_set_place)
+      if (bit_set_place == NULL)
         return false;
       bit_set= new (bit_set_place) Bit_set();
     }


sql/sql_class.h
```diff
@@ -1624,10 +1627,10 @@ class Bloom_filter {
   {
     if (!bit_set)
     {
-      void *bit_set_place= alloc_root(mem_root, sizeof(Bit_set));
+      void *bit_set_place= my_malloc(PSI_INSTRUMENT_ME, sizeof(Bit_set), 0);
```
percona-ysorokin: No need to have `mem_root` as an input parameter.

dutow: We can actually keep the mem_root, I retested it again. I think I forgot a make install before: the memroot itself isn't the issue, the recreateing in clear is.

sql/sql_class.h
```diff
@@ -1601,13 +1601,14 @@ class Bloom_filter {
   ~Bloom_filter()
   {
     // Do not delete, just destruct, due to MEM_ROOT allocation
-    bit_set->~bitset();
+    if (bit_set != NULL)
+      bit_set->~bitset();
```
laurynas-biveinis: Now I'm wondering whether it was correct for me to substitute delete with destructor call in the first place. The std::bitset apparently has no destructor. Perhaps it should be clear() here?

dutow: That's why it was accidentally working (but I think it was still UB) - it was possible that bit_set was null at this point.

I don't see the point of calling clear here - the next variables reusing the same place in memory will reinitialize anyway, and this Bloom_filter is destroyed at this point.

laurynas-biveinis: Since std::bitset destructor is default, then nothing has to be done here, and Bloom_filter destructor may default as well?

laurynas-biveinis (approved):

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/109/

PR #3306, opened 2019-07-11
PS-5747 : Remove AHI overhead for DDLs when AHI is disabled

satya-bodapati:
Problem:
--------
Even when AHI is disabled, there is overhead because of AHI.
This is because of checks required for AHI to delete hash entries
of pages being dropped.

These checks are not necessary if we know for sure that AHI is already
disabled.

Fix:
----
1. Take extra dict_sys mutex when enabling AHI (btr_search_enable()). This
   is now similar to AHI disable (btr_search_disable())

2. btr_search_enabled variable read under dict_sys mutex is now accurate
   and can be used to know if AHI is disabled for sure and can't be enabled
   concurrently.

Fix is similar to PS-5576 - Server stalls because ALTER TABLE on partitioned
                             table holds dict mutex

This fix is more generic and should cover DROP TABLE, DROP INDEX also (apart
from TRUNCATE)


storage/innobase/btr/btr0btr.cc
```diff
@@ -1192,17 +1192,26 @@ btr_create(
 /** Free a B-tree except the root page. The root page MUST be freed after
 this by calling btr_free_root.
 @param[in,out]	block		root page
-@param[in]	log_mode	mtr logging mode */
+@param[in]	log_mode	mtr logging mode
+@param[in]	is_ahi_allowed	false for intrinsic tables else true */
```
laurynas-biveinis: argument name and description differ. Perhaps the description explains when AHI is allowed, but that should be made clear

storage/innobase/btr/btr0sea.cc
```diff
@@ -429,16 +429,28 @@ btr_search_disable(
 	btr_search_mem_accounting_validate();
 }
 
-/** Enable the adaptive hash search system. */
+/** Enable the adaptive hash search system
+@param[in]	need_mutex	need to acquire dict_sys->mutex */
```
laurynas-biveinis: Add that the mutex must be already held if false

laurynas-biveinis (changes requested):
Nice. Minor comments only

storage/innobase/btr/btr0btr.cc
satya-bodapati: ahi is not applicable/allowed for intrinsic tables, For non-intrinsic tables, AHI is applicable/allowed.

Didn't get the question. What should I change here?

laurynas-biveinis (changes requested):
As discussed on Slack

satya-bodapati:
Ready for review

laurynas-biveinis (approved):

satya-bodapati:
I will submit a 8.0 version patch too. It will be done via cherry-pick( This is because PS-5576 was done on latest 5.7)

PR #3308, opened 2019-07-12
PS-5772: 8.0.16 encryption reconciliation.

robgolebiowski:
Upgrade:
All tables from encrypted tablespace need to have ENCRYPTION='Y' clause
assigned and DD'table option encryption set to Y. In previous versions
we have assigned encryption clause and DD encryption option only to
tablespaces. Thus on update now we need to assign ENCRYPTION='Y' and set
DD option encryption to Y for all tables that resided in encrypted
tablespace. Also we can be upgrading from PS with encrypted system
tablespace (ibdata).

EXPLICIT ENCRYPTION clause:
MySQL now adds clause ENCRYPTION=Y/N to all tables. However, in MySQL,
ENCRYPTION=N it is shown to the user in SHOW CREATE TABLE statement only
when it resides in schema with DEFAULT ENCRYPTION=Y. Thus only when it
was explicilty specified that table/space should stay decrypted. We were
using explicit ENCRYPTION=N clause to mark table/spaces that are to be
excluded from encryption threads. Thus we also show ENCRYPTION=N when it
was explicitly specified. This means that in PS - ENCRYPTION=N will show
up in SHOW CREATE TABLE even if SCHEMA has DEFAULT ENCRYPTION=N. This
behavior was retained to be able to tests encryption threads - we may
drop this behavior in favor of ENCRYPTION=NO_ONLINE.

--innodb-encrypt-tables to --default-table-encryption migration:
innodb-encrypt-tables has been removed. Now all encryption is directed
by database's default encryption and default_table_encryption variable.
Tablespaces and schema's inherit encryption clause from
default_table_encryption. Tables created in schema must obey schema's
default encryption when table_encryption_privilege_check - unless user
has privilege to overwrite table_encryption_privilege_check. Tables
created in tablespace - of course - must obey tablespace encryption and
when table_encryption_privilege_check is on - schema's default
encryption. encryption='y' is no longer added to all tables/tablespaces.
The code from adjust_encryption_options that was doing that was
commented out.


sql/dd/dd_table.cc
```diff
@@ -2656,6 +2661,46 @@ inline void report_error_as_tablespace_missing(const String_type name) {
...
+  const Tablespace *tsp = nullptr;
+  if (thd->dd_client()->acquire("innodb_system", &tsp)) {
+    return {true, false};
+  }
+  DBUG_ASSERT(tsp);
+
+  if (tsp->options().exists("encryption")) {
+    String_type e;
+    (void)tsp->options().get("encryption", &e);
+    DBUG_ASSERT(e.empty() == false);
```
laurynas-biveinis:
```suggestion
    DBUG_ASSERT(!e.empty());
```

sql/dd/dd_table.h
```diff
@@ -425,6 +425,9 @@ inline bool is_encrypted(const LEX_STRING &type) {
   return is_encrypted(String_type(type.str, type.length));
 }
 
+using Encrypt_result = ResultType<bool>;
+Encrypt_result is_system_tablespace_encrypted(THD *thd);
+
 using Encrypt_result = ResultType<bool>;
```
laurynas-biveinis:
```suggestion
```

sql/dd/impl/bootstrap/bootstrap_ctx.h
```diff
@@ -171,6 +176,11 @@ class DD_bootstrap_ctx {
   bool is_initialize() const {
     return opt_initialize && (m_actual_dd_version == dd::DD_VERSION);
   }
+
+  bool is_system_tablespace_encrypted() {
```
laurynas-biveinis:
```suggestion
  bool is_system_tablespace_encrypted() const noexcept {
```

sql/dd/impl/bootstrap/bootstrap_ctx.h
```diff
@@ -105,6 +106,10 @@ class DD_bootstrap_ctx {
     m_actual_dd_version = actual_dd_version;
   }
 
+  void set_system_tablespace_encrypted() {
```
laurynas-biveinis:
```suggestion
  void set_system_tablespace_encrypted() noexcept {
```

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -748,6 +748,18 @@ bool DDSE_dict_init(THD *thd, dict_init_mode_t dict_init_mode, uint version) {
                            &ddse_tablespaces))
     return true;
 
+  dd::Properties *p = dd::Properties_impl::parse_properties(ddse_tablespaces.begin()->get_options());
+
+  if (p->exists("encryption")) {
+    dd::String_type tablespace_encryption;
+    p->get("encryption", &tablespace_encryption);
+    if (my_strcasecmp(system_charset_info, tablespace_encryption.c_str(),
+                      "y")  == 0) {
```
laurynas-biveinis: clang-format
```suggestion
                      "y") == 0) {
```

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -748,6 +748,18 @@ bool DDSE_dict_init(THD *thd, dict_init_mode_t dict_init_mode, uint version) {
...
+        bootstrap::DD_bootstrap_ctx::instance().set_system_tablespace_encrypted();
+        //from now on all system tablespaces will be created with encryption='y'
```
laurynas-biveinis:
```suggestion
        // From now on all system tablespaces will be created with encryption='y'
```

sql/dd/upgrade_57/table.cc
```diff
@@ -1455,6 +1457,36 @@ static bool fix_fk_parent_key_names(THD *thd, const String_type &schema_name,
...
+  @param  name        Name of tablespace
+
+  @retval false - Not encrypted
+  @retval true  - Encrytpted
+*/
+
+bool is_tablespace_encrypted(const char *name) {
+
+  THD *thd = current_thd;
+  dd::cache::Dictionary_client *client = dd::get_dd_client(thd);
```
laurynas-biveinis: prefer auto for locals

sql/dd/upgrade_57/table.cc
```diff
@@ -1598,6 +1630,15 @@ static bool migrate_table_to_dd(THD *thd, const String_type &schema_name,
       key_info->is_algorithm_explicit = true;
   }
 
+  // if table is part of tablespace - check if tablespace is encrypted
```
laurynas-biveinis:
```suggestion
  // If table is a part of a tablespace - check if the tablespace is encrypted
```

sql/dd/upgrade_57/table.cc
```diff
@@ -1598,6 +1630,15 @@ static bool migrate_table_to_dd(THD *thd, const String_type &schema_name,
...
+  // if it is - assing encryption clause to the table
```
laurynas-biveinis:
```suggestion
  // If it is - assign encryption clause to the table
```

sql/dd/upgrade_57/table.cc
```diff
@@ -1598,6 +1630,15 @@ static bool migrate_table_to_dd(THD *thd, const String_type &schema_name,
...
+  if ( ( (share.tablespace && is_tablespace_encrypted(share.tablespace)) ||
```
laurynas-biveinis: clang-format

sql/sys_vars.cc
```diff
@@ -6961,13 +6961,36 @@ static bool check_set_default_table_encryption_access(
...
-static Sys_var_bool Sys_default_table_encryption(
+static const char *default_table_encryption_names[] = {
+    "OFF", "ON", "KEYRING_ON", "ONLINE_TO_KEYRING", "ONLINE_FROM_KEYRING_TO_UNENCRYPTED"};
+
+bool Sys_var_enum_default_table_encryption::global_update(THD *thd, set_var *var) { //TODO: Test if it deos not overflow
+
+  if (thd->variables.default_table_encryption == var->save_result.ulonglong_value) 
+    return 0;
+
+  global_var(ulong) = var->save_result.ulonglong_value;
```
laurynas-biveinis, line 6975: Not very common syntax

sql/sys_vars.cc
```diff
@@ -6961,13 +6961,36 @@ static bool check_set_default_table_encryption_access(
...
+
+  LEX_STRING innodb_engine{C_STRING_WITH_LEN("innodb")};
```
laurynas-biveinis:
```suggestion
  static const LEX_STRING innodb_engine{C_STRING_WITH_LEN("innodb")};
```

laurynas-biveinis: Check if LEX_CSTRING works too

sql/sys_vars.cc
```diff
@@ -6961,13 +6961,36 @@ static bool check_set_default_table_encryption_access(
...
+
+  plugin_ref plugin;
+  if ((plugin = ha_resolve_by_name(NULL, &innodb_engine, false))) {
```
laurynas-biveinis: s/NULL/nullptr/g

sql/sys_vars.cc
```diff
@@ -6961,13 +6961,36 @@ static bool check_set_default_table_encryption_access(
...
+
+static Sys_var_enum_default_table_encryption Sys_default_table_encryption(
     "default_table_encryption",
     "Database and tablespace created with this default encryption property"
     "unless, the user specify a explicit encryption property.",
     HINT_UPDATEABLE SESSION_VAR(default_table_encryption), CMD_LINE(OPT_ARG),
-    DEFAULT(false), NO_MUTEX_GUARD, IN_BINLOG,
-    ON_CHECK(check_set_default_table_encryption_access), ON_UPDATE(0));
+    default_table_encryption_names, DEFAULT(DEFAULT_TABLE_ENC_OFF), NO_MUTEX_GUARD,
+    IN_BINLOG, ON_CHECK(check_set_default_table_encryption_access));
```
laurynas-biveinis: Why not have ON_UPDATE callback instead of having to create Sys_var_enum_default_table_encryption class?

sql/sys_vars.h
```diff
@@ -2096,6 +2096,21 @@ class Sys_var_enum_binlog_checksum : public Sys_var_enum {
   virtual bool global_update(THD *thd, set_var *var);
 };
 
+class Sys_var_enum_default_table_encryption : public Sys_var_enum {
```
laurynas-biveinis, line 2099: See above

storage/innobase/dict/dict0dd.cc
```diff
@@ -6095,6 +6095,38 @@ bool dd_is_table_in_encrypted_tablespace(const dict_table_t *table) {
   }
 }
 
+/* Check if the table belongs to an encrypted tablespace.
+@param[in]	name	name of table for which check is to be done
+@return true if it does. */
+bool dd_is_table_in_encrypted_tablespace(const char *name) {
+  space_id_t space_id = fil_space_get_id_by_name(name);
```
laurynas-biveinis: Prefer auto
(me, having a s/ulint/space_id_t or auto patch in progress ;)

storage/innobase/dict/dict0dd.cc
```diff
@@ -6095,6 +6095,38 @@ bool dd_is_table_in_encrypted_tablespace(const dict_table_t *table) {
...
+    Dictionary/ */
+    THD *thd = current_thd;
+    dd::cache::Dictionary_client *client = dd::get_dd_client(thd);
+    dd::cache::Dictionary_client::Auto_releaser releaser(client);
+    dd::Tablespace *dd_space = nullptr;
+
+    if (!client->acquire_uncached_uncommitted<dd::Tablespace>(
+            space_id, &dd_space) &&
+        dd_space != nullptr) {
+      uint32 flags;
```
laurynas-biveinis:
```suggestion
      uint32_t flags;
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -12118,32 +12125,22 @@ bool create_table_info_t::create_option_tablespace_is_valid() {
...
-        my_printf_error(ER_ILLEGAL_HA_CREATE_OPTION,
-                        "InnoDB : ENCRYPTION=Y is not accepted"
-                        " for system tablespace.",
-                        MYF(0));
-        return false;
+      if (m_create_info->encrypt_type.str != nullptr) {
+        if (is_temp) {
+          /* Temporary tablespace is being used for table */
+          my_printf_error(ER_ILLEGAL_HA_CREATE_OPTION,
+                          "InnoDB : ENCRYPTION=Y is not accepted"
```
laurynas-biveinis:
```suggestion
                          "InnoDB: ENCRYPTION=Y is not accepted"
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -12118,32 +12125,22 @@ bool create_table_info_t::create_option_tablespace_is_valid() {
...
-      /* System or temporary tablespace is being used for table */
-      if (m_create_info->encrypt_type.str != nullptr &&
-          !Encryption::is_none(m_create_info->encrypt_type.str)) {
-        /* Encryption is not allowed for system tablespace. */
-        my_printf_error(ER_ILLEGAL_HA_CREATE_OPTION,
-                        "InnoDB : ENCRYPTION=Y is not accepted"
-                        " for system tablespace.",
-                        MYF(0));
-        return false;
+      if (m_create_info->encrypt_type.str != nullptr) {
```
laurynas-biveinis: Join nested ifs (at least Percona ones) to && ?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -12671,42 +12638,39 @@ void ha_innobase::adjust_encryption_options(HA_CREATE_INFO *create_info,
...
-        break;
-      case SRV_ENCRYPT_TABLES_OFF:
-      case SRV_ENCRYPT_TABLES_ONLINE_FROM_KEYRING_TO_UNENCRYPTED:
-      case SRV_ENCRYPT_TABLES_ONLINE_TO_KEYRING:
-        break;
-      default:
-        ut_ad(0);
-    }
-  }
+  //if (create_info->encrypt_type.length == 0 &&
```
laurynas-biveinis: Remove

storage/innobase/handler/ha_innodb.cc
```diff
@@ -12671,42 +12638,39 @@ void ha_innobase::adjust_encryption_options(HA_CREATE_INFO *create_info,
...
+        //break;
+      //default:
+        //ut_ad(0);
+    //}
+  //}
 
   adjust_encryption_key_id(create_info,
                            table_def ? &(table_def->options()) : nullptr);
 
   /* Add encryption attribute only to file_per_table table */
```
laurynas-biveinis: Remove

storage/innobase/srv/srv0srv.cc
```diff
@@ -532,7 +532,7 @@ ulong srv_n_purge_threads = 4;
 /* the number of pages to purge in one batch */
 ulong srv_purge_batch_size = 20;
 
-ulong srv_encrypt_tables = 0;
+enum_default_table_encryption srv_default_table_encryption;
```
laurynas-biveinis, line 535: Can InnoDB use the server variable instead?

laurynas-biveinis (changes requested):

storage/innobase/handler/ha_innodb.cc
```diff
@@ -11196,7 +11196,11 @@ and set the encryption flag in table flags
 dberr_t create_table_info_t::enable_master_key_encryption(dict_table_t *table) {
   const char *encrypt = m_create_info->encrypt_type.str;
 
-  if (Encryption::is_none(encrypt)) return (DB_SUCCESS);
+  /* if table is part of tablespace - no need for retrieving
+   * master key as tablespace key was already decrypted
```
laurynas-biveinis: Please don't use this leading * column comment style

laurynas-biveinis (changes requested):

storage/innobase/handler/ha_innodb.cc
```diff
@@ -12138,7 +12138,7 @@ bool create_table_info_t::create_option_tablespace_is_valid() {
           /* Temporary tablespace is being used for table */
           my_printf_error(
               ER_ILLEGAL_HA_CREATE_OPTION,
-              "InnoDB : ENCRYPTION=Y is not accepted"
+              "InnoDB : ENCRYPTION is not accepted"
```
laurynas-biveinis: s/InnoDB :/InnoDB: - already suggested in a previous review, will save you a rerecord :)

sql/dd/dd_table.h
robgolebiowski: done

sql/dd/impl/bootstrap/bootstrap_ctx.h
robgolebiowski: done

sql/dd/impl/bootstrap/bootstrap_ctx.h
robgolebiowski: done

sql/dd/impl/bootstrap/bootstrapper.cc
robgolebiowski: done

sql/dd/impl/bootstrap/bootstrapper.cc
robgolebiowski: done

sql/dd/upgrade_57/table.cc
robgolebiowski: done

sql/dd/upgrade_57/table.cc
robgolebiowski: done

sql/dd/upgrade_57/table.cc
robgolebiowski: done

sql/sys_vars.cc
robgolebiowski: ha_resolve_by_name requires LEX_STRING

sql/sys_vars.cc
robgolebiowski: done

sql/sys_vars.cc
robgolebiowski: That's because regular ON_UPDATE do not give you access to the value to which variable is being changed to (ser_var *). In order to get this access classes override global_update function.

storage/innobase/dict/dict0dd.cc
robgolebiowski: done

storage/innobase/dict/dict0dd.cc
robgolebiowski: done

storage/innobase/handler/ha_innodb.cc

storage/innobase/handler/ha_innodb.cc

storage/innobase/srv/srv0srv.cc
robgolebiowski: I put this as a separate copy as it is frequently checked by encryption threads.
Also because anyways I need to call fil_crypt_set_thread_cnt each time default_table_encryption is set. - I am re- using this to set this innodb copy of server variable.

storage/innobase/handler/ha_innodb.cc
robgolebiowski: sorry, keep forgetting

storage/innobase/handler/ha_innodb.cc

mysql-test/suite/encryption/r/mysqldump.result
```diff
@@ -189,7 +189,7 @@ Table	Create Table
 i_part3	CREATE TABLE `i_part3` (
   `id` int(11) DEFAULT NULL,
   `name` varchar(50) DEFAULT NULL
-) /*!50100 TABLESPACE `innodb_file_per_table` */ ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
+) /*!50100 TABLESPACE `innodb_file_per_table` */ ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci ENCRYPTION='n'
```
satya-bodapati, line 192: Shouldn't this output be like /*!80016 ENCRYPTION='n' */

mysql-test/std_data/dd/sdi/innodb_sdi/mysql.json
```diff
@@ -24735,7 +24735,7 @@
         "created": NNN,
         "last_altered": NNN,
         "hidden": 1,
-        "options": "avg_row_length=0;explicit_tablespace=1;key_block_size=0;keys_disabled=0;pack_record=0;stats_auto_recalc=0;stats_persistent=0;stats_sample_pages=0;",
+        "options": "avg_row_length=0;encrypt_type=N;explicit_tablespace=1;key_block_size=0;keys_disabled=0;pack_record=0;stats_auto_recalc=0;stats_persistent=0;stats_sample_pages=0;was_encrypt_type_set_explicitly=1;",
```
satya-bodapati, line 24738: I would go with shorter option and similar one to upstream (explicit_tablespace). Suggestion: "explicit_encryption"

satya-bodapati: Why is this attribute set ? I mean I don't think mysql.host was created with explicit encryption type.

mysql-test/suite/innodb/r/percona_sys_tablespace_encrypt.result
```diff
@@ -79,154 +85,179 @@ SELECT @@innodb_sys_tablespace_encrypt;
 @@innodb_sys_tablespace_encrypt
 1
 include/assert.inc [System tablespace should be encrypted]
+CREATE DATABASE db1;
+CREATE USER u1@localhost;
+GRANT ALL ON db1.* TO u1@localhost;
+GRANT CREATE TABLESPACE, PROCESS, SYSTEM_VARIABLES_ADMIN ON *.* TO u1@localhost;
+# System is encrypted, test with DEFAULT ENCRYPTION=N
+# and table_encryption_privilege_check = OFF
```
satya-bodapati, line 96: Either set this variable "table_encryption_privilege_check" again here or show the value of the varible

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/243

mysql-test/std_data/dd/sdi/innodb_sdi/mysql.json
```diff
@@ -15,7 +15,7 @@
         "created": NNN,
         "last_altered": NNN,
         "hidden": 2,
-        "options": "avg_row_length=0;explicit_tablespace=1;key_block_size=0;keys_disabled=0;pack_record=1;row_type=2;stats_auto_recalc=0;stats_persistent=0;stats_sample_pages=0;",
+        "options": "avg_row_length=0;encrypt_type=N;explicit_tablespace=1;key_block_size=0;keys_disabled=0;pack_record=1;row_type=2;stats_auto_recalc=0;stats_persistent=0;stats_sample_pages=0;",
```
satya-bodapati, line 18: why do we have this DD change? AFAIU, upstream adds encrypt_type=N for all non-encrypted tables.

scripts/mysql_system_tables_fix.sql
```diff
@@ -1002,62 +1002,213 @@ INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.session', 'localhost', 'SE
...
+set @is_mysql_encrypted = (select ENCRYPTION from information_schema.INNODB_TABLESPACES where NAME='mysql');
+
+SET @str="ALTER TABLE mysql.db ENCRYPTION='Y'";
+SET @cmd = IF(STRCMP(@is_mysql_encrypted,'Y'), 'SET @dummy = 0', @str);
+PREPARE stmt FROM @cmd;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
 # Move all system tables with InnoDB storage engine to mysql tablespace.
 SET @cmd="ALTER TABLE mysql.db TABLESPACE = mysql";
```
satya-bodapati, line 1014: Is it possible to combine ENCRYPTION='Y' and TABLESPACE=mysql ?

sql/dd/dd_table.cc
```diff
@@ -2107,12 +2107,17 @@ static bool fill_dd_table_from_create_info(
       encrypt_type.assign(create_info->encrypt_type.str,
                           create_info->encrypt_type.length);
     }
+    table_options->set("encrypt_type", encrypt_type);
```
satya-bodapati, line 2110: "MySQL now adds clause ENCRYPTION=Y/N to all tables." I thought this is done by upstream. Why are 'we' doing it here?

sql/dd/upgrade_57/table.cc
```diff
@@ -1598,6 +1628,15 @@ static bool migrate_table_to_dd(THD *thd, const String_type &schema_name,
       key_info->is_algorithm_explicit = true;
   }
 
+  // if table is part of tablespace - check if tablespace is encrypted
+  // if it is - assing encryption clause to the table
+  if (((share.tablespace && is_tablespace_encrypted(share.tablespace)) ||
+       (schema_name == "mysql" && bootstrap::DD_bootstrap_ctx::instance()
+                                      .is_system_tablespace_encrypted())) &&
+      !is_encrypted(share.encrypt_type)) {
+    share.encrypt_type.str = C_STRING_WITH_LEN("Y");
```
satya-bodapati: Shouldn't this be done at InnoDB layer? the table attributes can be fixed/updated in one of DD upgrade APIs?

sql/dd/impl/bootstrap/bootstrap_ctx.h
```diff
@@ -171,6 +176,10 @@ class DD_bootstrap_ctx {
   bool is_initialize() const {
     return opt_initialize && (m_actual_dd_version == dd::DD_VERSION);
   }
+
+  bool is_system_tablespace_encrypted() {
```
satya-bodapati: Although this works, this doesn't seem right. Server code shouldn't ideally care of InnoDB artifacts.

As i understand, server cannot handle encrypted hardcoded tabelspaces? but it should be done in 'generic' way..

sql/dd/upgrade_57/table.cc
```diff
@@ -1455,6 +1457,34 @@ static bool fix_fk_parent_key_names(THD *thd, const String_type &schema_name,
...
+
+  if (!client->acquire<dd::Tablespace>(name, &dd_space) &&
+      dd_space != nullptr) {
+    if (!dd_space->options().exists("encryption")) {
+      return false;
+    }
+
+    dd::String_type tablespace_encryption;
+    (void)dd_space->options().get("encryption", &tablespace_encryption);
+    return tablespace_encryption == "Y";
```
satya-bodapati: you are sure that it is never smaller case?

sql/handler.h
```diff
@@ -1836,6 +1836,8 @@ typedef bool (*rotate_encryption_master_key_t)(void);
 */
 using fix_tablespaces_empty_uuid_t = bool (*)(void);
 
+using fix_default_table_encryption_t = void (*)(ulong);
```
satya-bodapati, line 1845: Is the new handlerton added to handle encryption attribute during upgrade?

sql/sql_table.cc
```diff
@@ -8453,15 +8454,23 @@ static bool validate_table_encryption(THD *thd, HA_CREATE_INFO *create_info) {
             create_info->tablespace, &tt)) {
       return true;
     }
-    uses_general_tablespace = (tt != Tablespace_type::SPACE_TYPE_IMPLICIT);
+    uses_general_tablespace = (tt != Tablespace_type::SPACE_TYPE_IMPLICIT &&
+                               tt != Tablespace_type::SPACE_TYPE_SHARED);
+    uses_system_tablespace = tt == Tablespace_type::SPACE_TYPE_SYSTEM;
+    if (uses_system_tablespace) {
+      dd::Encrypt_result result = dd::is_system_tablespace_encrypted(thd);
```
satya-bodapati, line 8461: IMHO, server code shouldn't be talking about InnoDB specific details.

satya-bodapati: am referring to is_system-tablespace_encrypted()

sql/sql_zip_dict.cc
```diff
@@ -103,13 +104,27 @@ bool bootstrap(THD *thd) {
   DBUG_EXECUTE_IF("skip_compression_dict_create", skip_bootstrap = true;
                   return false;);
 
+  dd::String_type dict_table_str_enc{dict_table_str};
+
+  if (dd::bootstrap::DD_bootstrap_ctx::instance()
+          .is_system_tablespace_encrypted()) {
+    dict_table_str_enc += " ENCRYPTION='Y'";
```
satya-bodapati, line 110: Why is this required? I mean which tables need this?

sql/sys_vars.cc
satya-bodapati: yeah, this looks unusual.. :)

storage/innobase/dict/dict0dd.cc
```diff
@@ -6095,6 +6095,38 @@ bool dd_is_table_in_encrypted_tablespace(const dict_table_t *table) {
   }
 }
 
+/* Check if the table belongs to an encrypted tablespace.
+@param[in]	name	name of table for which check is to be done
+@return true if it does. */
+bool dd_is_table_in_encrypted_tablespace(const char *name) {
```
satya-bodapati: Does it work for partitioned table?

storage/innobase/dict/dict0upgrade.cc
```diff
@@ -937,18 +937,6 @@ bool dd_upgrade_table(THD *thd, const char *db_name, const char *table_name,
 
   dd_set_table_options(dd_table, ib_table);
 
-  /* Tables in encrypted general tablespace from PS-5.7 have
-  encryption attribute. Remove this attribute as tables in
-  general tablespace shouldn't have it. General tablespace
-  should have encryption attribute */
-  if (!dict_table_is_file_per_table(ib_table) &&
```
satya-bodapati, line 944: ok, so it is reverse now.. :D  You should fix DD here.  ie. add encrypt_type to dd::Table here..

satya-bodapati: As I understand, you are doing it in server

satya-bodapati (changes requested):
Partial review done..

mysql-test/suite/innodb/t/percona_sys_tablespace_encrypt.test
```diff
@@ -154,107 +178,149 @@ SELECT @@innodb_sys_tablespace_encrypt;
...
---error ER_ILLEGAL_HA_CREATE_OPTION
-CREATE TABLE t5_5(a TEXT) TABLESPACE = innodb_system, ENCRYPTION='Y';
+--error ER_INVALID_ENCRYPTION_REQUEST
+CREATE TABLE t1(a TEXT) TABLESPACE = innodb_system;
+
+--error ER_INVALID_ENCRYPTION_REQUEST
+CREATE TABLE t1(a TEXT) TABLESPACE = innodb_system, ENCRYPTION='N';
+
+--error ER_CANNOT_SET_TABLE_ENCRYPTION
+CREATE TABLE t1(a TEXT) TABLESPACE = innodb_system, ENCRYPTION='Y';
```
satya-bodapati, line 250: system is encrypted, why encryption='y' fails?

sql/sql_zip_dict.cc
satya-bodapati: Ok, nevermind, found the definition above. But I don't think it is right strategy.

We shouldn't add "ENCRYPTION=Y" to each create table statement. 

can we set mysql database default encryption as encrypted instead of fixing each "create table mysql. *" ?

scripts/mysql_system_tables_fix.sql
```diff
@@ -1002,62 +1002,213 @@ INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.session', 'localhost', 'SE
...
 PREPARE stmt FROM @cmd;
 EXECUTE stmt;
 DROP PREPARE stmt;
 
+SET @str="ALTER TABLE mysql.user ENCRYPTION='Y'";
```
satya-bodapati, line 1019: We should avoid modifying all create table statements for mysql database. Instead do ALTER DATABASE mysql default encryption='y' if mysql tablespace is encrypted.

mysql-test/suite/encryption/r/mysqldump.result
robgolebiowski: this comment is SHOW CREATE DATABASE specific. From WL#12261 HLA:
1. SHOW CREATE DATABASE uses versioned mysql-specific
    commenting style like /*!80014 DEFAULT ENCRYPTION */ for
    portability, so the older server would ignore parsing this
    new clause.

scripts/mysql_system_tables_fix.sql
robgolebiowski: From documentation:
Changing the default encryption of the database does not alter the encryption of existing tables associated with the schema. Only newly-created tables inherit the default database encryption.

Thus I need to set encryption explicitly for all tables inside mysql schema.

satya-bodapati: Yes, if this can be set before tables are created, we can avoid the explicit change for all create tables.

mysql-test/suite/encryption/r/mysqldump.result
satya-bodapati: Right, my point is that it is not enclosed in such comment. I just see plain "ENCRYPTION='n'" in the diff

robgolebiowski: This is upstream behavior. For tables 'innodb_file_per_table' the ENCRYPTION clause is shown without comment. However for tables inside general tablespace the ENCRYPTION clause is put into comment as tables inside general tablespaces could not have ENCRYPTION clause assigned in the previous versions.

mysql-test/std_data/dd/sdi/innodb_sdi/mysql.json
robgolebiowski: Upstream did - but we not have fully merged upstream changes before this PR. This comes from upstream and it was merged in from upstream in this PR.

sql/dd/dd_table.cc
robgolebiowski: This comes from upstream and it was merged in from upstream in this PR.

scripts/mysql_system_tables_fix.sql
robgolebiowski: but this is script for upgrade - so it works on already existing tables.

mysql-test/std_data/dd/sdi/innodb_sdi/mysql.json
robgolebiowski: In this diff there is no mysql.host. Other than that actually those tables are created with explicit encryption type. Have a look at mysql_system_tables.sql. It would make sense to remove the encryption clause from there so the ENCRYPTION clause was not added in case table is not encrypted. Otherwise encryption threads will not pick up those tables. I do not know if it is worth the effort since we will probably go with NO_ONLINE.

mysql-test/suite/innodb/r/percona_sys_tablespace_encrypt.result
robgolebiowski: done

scripts/mysql_system_tables_fix.sql
robgolebiowski: I do not see a convenient way of combining those two without explicitly specifying ENCRYPTION=N clause also. Also I do not see much benefit in refactoring this.

sql/dd/upgrade_57/table.cc
robgolebiowski: This is for upgrade from 5.7. In this function .frm is read and based on the information from .frm CREATE_INFO is created - that should contain encrypt_type information. Function name:
/**
  Read .frm files and enter metadata for tables/views.
*/

static bool migrate_table_to_dd(THD *thd, const String_type &schema_name,

since it is reading frm files it is clearly interacting with INNODB.

sql/sql_table.cc
robgolebiowski: I think it is hard to draw a line what is innodb specific and what is not. validate_table_encryption is already calling dd::is_tablespace_encrypted - which returns information whether table is in general tablespace - is this InnoDB specific or not ? I can make is_tablespace_encrypted return uses_general_tablespace = true when system tablespace is used.

sql/sql_zip_dict.cc
robgolebiowski: I guess we can change DEFAULT ENCRYPTION on start of a bootstrap and then change it back to the original DEFAULT ENCRYPTION. However, check mysql_system_tables.sql - they are assigning ENCRYPTION per clause - so only advantage we will get are those two tables - I do not think it is worth the effort. Note that mysql_system_tables_fix.sql you are referring to later is about upgrade and for this we need explicit ENCRYPTION clause - DEFAULT ENCRYPTION will not help.

storage/innobase/dict/dict0upgrade.cc
robgolebiowski: The code from Oracle that sets the default encryption is here:
sql/dd/impl/bootstrap/bootstrapper.cc
-    // All the predefined tablespace are unencrypted (atleast for now).
-    tablespace->options().set("encryption", "N");
And the validation of encryption clause is done in sql/sql_table validate_table_encryption. From MySQL code I understood the intention was to implement all encryption stuff in server code, not in innodb

robgolebiowski: Also scripts update encryption flags before the server start works on them - The encryption validation is done mysql_create_table_no_lock - before innodb is called - thus the encryption must be fixed in the server.

sql/sql_table.cc
robgolebiowski: The encryption validation is done mysql_create_table_no_lock - before innodb is called - thus the encryption must be fixed in the server.

mysql-test/suite/innodb/t/percona_sys_tablespace_encrypt.test
robgolebiowski: Because SCHEMA's DEFAULT ENCRYPTION is 'N' and table_encryption_privilege_check is ON.

sql/dd/impl/bootstrap/bootstrap_ctx.h
robgolebiowski: Now it is the server which validates the encryption, before it even goes into InnoDB. Thus the encryption needs to be fixed before the server code will validate it.

sql/handler.h
robgolebiowski: no, it is for handling changing default_table_encryption variable. The encryption threads in the background check this variable repeatedly and I need to set fil_crypt_threads_event when it is set. Thus I need this handler anyways - and I am also reusing it for changing local copy of this variable inside innodb - so to not have to reach out to the server all the time.

sql/dd/upgrade_57/table.cc
satya-bodapati: reading .frms doesn't mean it is interacting with InnoDB. Please consider doing this in InnoDB. Specially the dd_upgrade_table() API.

It should be simpler too
something like 
```
if (table->space == 0 && srv_sys_space.is_encrypted() (or some flags check))
 {
       add encryption attribute
}
```

sql/dd/impl/bootstrap/bootstrap_ctx.h
satya-bodapati: this cannot be an excuse to put InnoDB specific details in server.

SE should convey the intention to upper layer via APIs. and as far as server is concerned it should only care that it is operating on tablespace (and not the specific type).

But as I understand from @georgelorch and @laurynas-biveinis, we are happy to have something 'working'. We can do it post reconciliation. Please create JIRA issue for this.

sql/sql_table.cc
satya-bodapati: General Tablespace is not InnoDB specific concept. Other engines can have genral tablespace too.

I see your point but I think we can avoid specific details here.
From DD perspective, system tablespace is also a general tablespace. A general tablespace is tablespace that can hold multiple tables. Can't we treat system as general type and remove this 'system' specific checks?

satya-bodapati: This also falls under 'get-this-working-and-do-later' category.. So we can definitely improve this. Please file JIRA issue if u want to handle this later.

sql/sql_zip_dict.cc
satya-bodapati: Why are we checking "system tablespace" encryption status? Is this the ibdata1 status or mysql.ibd status?

storage/innobase/dict/dict0upgrade.cc
satya-bodapati: may be true for regular case. During upgrade, server doesn't even know what InnoDB tablespaces are. It is SE (InnoDB) that reads SYS_TABLESPACES and register in DD. (dd_upgrade_tablespace())

Similarly, for table, .frm gives most details about table via TABLE and TABLE_SHARE and InnoDB DD has some more details about the table. It uses that too.. My point is that for upgrade, it is much easier and cleaner to register encryption attribute in DD.

sql/handler.h
satya-bodapati: can you please document in the code it is used for encryption threads only?

scripts/mysql_system_tables_fix.sql
satya-bodapati: Ok, fine..

sql/dd/upgrade_57/table.cc
satya-bodapati: slight change of mind:

during upgrade, server has to add "encrypt_type=n" right for all the tables upgraded from 5.7. Where does this happen? if it happens in server, am OK if you can do at same place.

satya-bodapati (changes requested):

storage/innobase/handler/handler0alter.cc
```diff
@@ -4565,15 +4569,6 @@ static MY_ATTRIBUTE((warn_unused_result)) bool prepare_inplace_alter_table_dict(
 
     const char *encrypt;
     encrypt = ha_alter_info->create_info->encrypt_type.str;
-    /* If encryption option is specified, then it must be
-    innodb-file-per-table tablespace. Otherwise case would
-    have already been blocked at
-    create_option_tablespace_is_valid(). */
-    if (encrypt) {
```
satya-bodapati, line 4572: why is this block removed?

storage/innobase/os/os0file.cc
```diff
@@ -8709,7 +8709,6 @@ void Encryption::get_master_key(ulint *master_key_id, byte **master_key) {
   memset(key_name, 0x0, sizeof(key_name));
 
   if (s_master_key_id == 0) {
-    ut_ad(strlen(server_uuid) > 0);
```
satya-bodapati, line 8709: This removal looks suspicious

storage/innobase/srv/srv0srv.cc
satya-bodapati: How do you syncronize two copies?

satya-bodapati (changes requested):
Full review done..

sql/sql_zip_dict.cc
robgolebiowski: changed the name of the variable to is_dd_encrypted. Now DD does not see InnoDB internals. Whether to encrypt or not is based on encryption flag from first tablespace in bootstrap.cc:DDSE_dict_init. First Plugin_tablespace is DD tablespace.

robgolebiowski: this variable should be named is_mysql_tablespace_encrypted in the first place - now is is_dd_encrypted

sql/dd/upgrade_57/table.cc
robgolebiowski: The code there is not needed. This is because sql/dd/impl/upgrade/dd.cc:update_meta_data is already updating table's encrypt_type based on general tablespace encrypt_type. Removing.

sql/dd/impl/bootstrap/bootstrap_ctx.h
robgolebiowski: this variable should be named is_mysql_tablespace_encrypted() in the first place - now is is_dd_encrypted() - to not show SE internals - based on that we know if mysql is encrypted from 5.7 during bootstrap.

storage/innobase/dict/dict0dd.cc
robgolebiowski: Dead code - removed.

sql/sql_table.cc
robgolebiowski: When tablespace is empty and it is not an implicit tablespace (DD treats this as a special case) - is_tablespace_encrypted returns false - so it already discovers that this is system tablespace and it knows that this tablespace is not encrypted. It assumes thus that InnoDB is used - as some other SE could be encrypting system tablespace. 
The only way to determine on DD level whether system tablespace is encrypted or not is by acquiring innodb_system tablespace. So the name innodb_system will shine through the cracks anyways.
So if we want this to be totally transparent - and for any general/system tablespace get answer encrypted yes or no we should delegate all the logic to handlers - which MySQL fails to do, because it already assumes that system tablespace is not encrypted.
However, this is quite big effort and I do not think we should do it now. Probably upstream will implement ibdata1 encryption sooner than later and we would end up removing this code anyways.

storage/innobase/handler/handler0alter.cc
robgolebiowski: upstream change - It is no longer is valid that only innodb-file-per-table tablespaces can be encrypted.

storage/innobase/os/os0file.cc
robgolebiowski: This check was added by 4d747cf709910ae0b6ed4626b0fca843c37d10e1 : Bug#29006275: ENCRYPTION MASTER KEY IS GENERATED WITH BLANK UUID. We have a better fix for it with fix_empty_uuid handler - thus I removed this. I should have added this to the commit message and will do that now.

storage/innobase/srv/srv0srv.cc
robgolebiowski: It can be only updated from the server level. The server's update function is:
Sys_var_enum_default_table_encryption::global_update . There I call handler function fix_default_table_encryption that will update the innodb's copy of this variable. Also this handler is called in mysqld.cc, right after fix_tablespace_empty_uuid.

sql/sql_table.cc
satya-bodapati: Ok, fine to defer this

satya-bodapati: Thanks for explaining the current logic.

storage/innobase/handler/handler0alter.cc
satya-bodapati: if upstream, then OK

storage/innobase/os/os0file.cc
satya-bodapati: ok, thnx

satya-bodapati (approved):
The only question left is "how do you syncronize two copies?" but it is fine to get answer later.

PR #3310, opened 2019-07-12
PS-3845, Bug#80821: replication fails with multi-delete and cascade FKs (8.0)

dutow:


sql/sql_delete.cc
```diff
@@ -835,14 +835,60 @@ bool Query_result_delete::prepare(THD *thd, List<Item> &, SELECT_LEX_UNIT *u) {
...
+/**
+  Test that the two strings are equal, accoding to the lower_case_table_names
+  setting.
+
+  @param  a     A table or database name
+  @param  b     A table or database name
+  @retval bool  True if the two names are equal
+*/
+static bool db_or_table_name_equals(const char *a, const char *b) {
+  return lower_case_table_names ? my_strcasecmp(files_charset_info, a, b) == 0
```
laurynas-biveinis, line 847: Do this comparison idiom and the need to have the helper function still hold in 8.0?

dutow: I still see some similar comparisons in the code, and couldn't find a better comparator, so I guess yes.

laurynas-biveinis (approved):

laurynas-biveinis:
LGTM, please upstream the fix

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/237/

PR #3312, opened 2019-07-12
Ps 5749 5.7

laurynas-biveinis:
https://ps57.cd.percona.com/job/percona-server-5.7-param/107/


sql/binlog_crypt_data.cc
```diff
@@ -112,7 +112,8 @@ bool Binlog_crypt_data::init_with_loaded_key(uint sch, const uchar* nonce)
 {
   scheme= sch;
 #ifdef MYSQL_SERVER
-  DBUG_ASSERT(key != NULL && nonce != NULL);
+  DBUG_ASSERT(key != NULL);
+  DBUG_ASSERT(nonce != NULL);
```
robgolebiowski, line 116: all the ASSERTs where split - why ?

storage/innobase/btr/btr0scrub.cc
```diff
@@ -15,9 +15,6 @@ Scrubbing of btree pages
 #include "dict0dict.h"
 #include "mtr0mtr.h"
 
-/* used when trying to acquire dict-lock */
-bool fil_crypt_is_closing(ulint space);
-
```
robgolebiowski, line 20: @dutow - scrubbing part - seems legit ?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -892,13 +892,13 @@ fil_crypt_start_encrypting_space(
 
 	if (space->encryption_type == Encryption::AES) {// We are re-encrypting space from MK encryption to RK encryption
 		crypt_data->encryption_rotation = Encryption::MASTER_KEY_TO_KEYRING;
-		ut_ad(space->encryption_key != NULL && space->encryption_iv != NULL);
```
robgolebiowski, line 896: :+1:

robgolebiowski: memory for space is allocated with ut_zalloc_nokey - can we add here :
static char zeros_test_block [ENCRYPTION_KEY_LEN];
memset (zeros_test_block, 0, sizeof testblock);
ut_ad(memcmp(space->encryption_key, zeros_test_block, ENCRYPTION_KEY_LEN) == 0); ?
space->encryption_iv in MySQL is ENCRYPTION_KEY_LEN, although the real size is 128 bits. We can skip checking it - encryption_key should be enough.

robgolebiowski (changes requested):

george-lorch (approved):

sql/binlog_crypt_data.cc
laurynas-biveinis: so that if they fail the diagnostics pinpoint the failing part instead of everything (i.e. "C" instead of "A && B && C && D")

storage/innobase/fil/fil0crypt.cc
laurynas-biveinis: I am testing
`
#ifdef UNIV_DEBUG
		static const char all_zero_key[ENCRYPTION_KEY_LEN] = {0};
		ut_ad(memcmp(space->encryption_key, all_zero_key, ENCRYPTION_KEY_LEN) == 0);
#endif
`

laurynas-biveinis:
https://ps57.cd.percona.com/job/percona-server-5.7-param/113/

storage/innobase/btr/btr0scrub.cc
dutow: Yes, it's not used at all there.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: great, thanks

robgolebiowski (approved):

PR #3315, opened 2019-07-16
PS-5775: Improve redo log encryption error messages (5.7)

dutow:
If redo log encryption fails becasuse a keyring errors, this should be
reported to the user instead of a generic error message.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -21528,11 +21528,20 @@ update_innodb_redo_log_encrypt(
 		return;
 	}
 
-	byte iv[ENCRYPTION_KEY_LEN];
-	Encryption::random_value(iv);
+	ut_ad(strlen(server_uuid) > 0);
+
+	if (!Encryption::check_keyring()) {
+		ib_senderrf(
+			thd, IB_LOG_LEVEL_WARN,
+			ER_REDO_ENCRYPTION_ERROR,
+			"Redo log cannot be encrypted"
+			" if the keyring plugin is not loaded.");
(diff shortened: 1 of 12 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -21528,11 +21528,20 @@ update_innodb_redo_log_encrypt(
 		return;
 	}
 
-	byte iv[ENCRYPTION_KEY_LEN];
-	Encryption::random_value(iv);
+	ut_ad(strlen(server_uuid) > 0);
+
+	if (!Encryption::check_keyring()) {
+		ib_senderrf(
+			thd, IB_LOG_LEVEL_WARN,
```
satya-bodapati, line 21535: same as other PR comment, see if we can make it ERROR instead of WARN

satya-bodapati (changes requested):

storage/innobase/handler/ha_innodb.cc
dutow: As we discussed on slack: I can make most of the messages errors, but not all of them - we can't report errors from the update function, only from validate. I could move most of the checks to validate, but it'll be more complex there, and I still would have to keep some warnings. So it's best to leave these as is.

satya-bodapati: It doesn't make sense to move the code to validate() because of the limitation.

This is upstream limitation that we cannot return errors from SET's update() function. Can you please create bug report as Feature Request?

satya-bodapati (approved):

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/123/#showFailuresLink

PR #3316, opened 2019-07-16
PS-5776: Report undo encryption errors to the user (8.0)

dutow:
If we can't set the undo tablespace to be encrypted, and it was requested in a
MySQL session, errors should be also reported to the sessior requesting encryption.

If encryption fails because there is no keyring plugin loaded, this should be
mentioned in the error message.

### DIFF
--- a/mysql-test/suite/innodb/r/log_encrypt_kill.result
+++ b/mysql-test/suite/innodb/r/log_encrypt_kill.result
@@ -6,6 +6,8 @@ SET GLOBAL innodb_redo_log_encrypt = 1;
 Warnings:
 Warning	49014	InnoDB: Redo log cannot be encrypted if the keyring plugin is not loaded.
 SET GLOBAL innodb_undo_log_encrypt = 1;
+Warnings:
+Warning	49015	InnoDB: Undo log can't be encrypted if the keyring plugin is not loaded.
 UNINSTALL PLUGIN keyring_file;
 ERROR 42000: PLUGIN keyring_file does not exist
 CREATE TABLE tne_1(c1 INT, c2 varchar(2000)) ENGINE = InnoDB;
--- a/storage/innobase/srv/srv0srv.cc
+++ b/storage/innobase/srv/srv0srv.cc
@@ -2886,7 +2886,7 @@ bool srv_enable_redo_encryption_rk(THD *thd) {
 }
 
 /* Enable UNDO tablespace encryption */
-bool srv_enable_undo_encryption(bool is_boot) {
+bool srv_enable_undo_encryption(THD *thd, bool is_boot) {
   /* Traverse over all UNDO tablespaces and mark them encrypted. */
   undo::spaces->s_lock();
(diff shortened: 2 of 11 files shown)


mysql-test/suite/innodb/r/log_encrypt_kill.result
```diff
@@ -6,6 +6,8 @@ SET GLOBAL innodb_redo_log_encrypt = 1;
 Warnings:
 Warning	13068	InnoDB: Can't set redo log tablespace to be encrypted.
 SET GLOBAL innodb_undo_log_encrypt = 1;
```
satya-bodapati, line 8: Can you do SELECT on the global variable after this operation?

two things
1. it should be error instead of warning (if possible)
2. the value should be OFF/0 after the SET failed.

dutow: 1 - all other similar messages are warnings (because that's how existing upstream messages worked around this)
2 - fixed

satya-bodapati: Right, I think we agreed that we will deviate this from upstream? And warning to error deviation makes sense IMHO.

dutow: I mean, all other messages in our codebase (for redo encryption) are warnings currently. Upstream reports a warning if the server is read only, and doesn't notify the user for any other failure. We changed the code so we report warnings for other cases too.
edit: hm, upstream doesn't report anything even for read only mode - not sure from where I took this then.

satya-bodapati: If it is simple change, I would like you to try it in this PR

dutow: As we discussed on slack: I can make most of the messages errors, but not all of them - we can't report errors from the update function, only from validate. I could move most of the checks to validate, but it'll be more complex there, and I still would have to keep some warnings. So it's best to leave these as is.

storage/innobase/srv/srv0srv.cc
```diff
@@ -2929,6 +2929,9 @@ bool srv_enable_undo_encryption(bool is_boot) {
     /* Fill up encryption info to be set */
     if (!Encryption::fill_encryption_info(key, iv, encrypt_info, is_boot)) {
       ib::error(ER_IB_MSG_1052, undo_space->space_name());
+      if (thd != nullptr)
```
satya-bodapati: Please use 
 if (X) {
blah
}
format in InnoDB

satya-bodapati (approved):

storage/innobase/srv/srv0srv.cc
satya-bodapati: Applies to other lines as well

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/284/

PR #3317, opened 2019-07-17
Fix PS-5780 (The release build (5.6) does not work on Debian buster) (5.6)

percona-ysorokin:
https://jira.percona.com/browse/PS-5780

Partially cherry-picked 5.7 upstream fixes for gcc-8 - releated warnings.
"Bug#27558169 BACKPORT TO 5.7 BUG #26826272: REMOVE GCC 8 WARNINGS [noclose]"
(commit mysql/mysql-server@1ffd796)
(commit mysql/mysql-server@a091d6b)

Fixed Percona-specific 'strncpy()'-releated
'output may be truncated copying NNN bytes from a string of length MMM [-Werror=stringop-truncation]'
warning inside 'log_online_setup_bitmap_file_range()' function in
'storage/innobase/log/log0online.cc'.

### DIFF
--- a/mysys/my_gethwaddr.c
+++ b/mysys/my_gethwaddr.c
@@ -108,7 +108,7 @@ my_bool my_gethwaddr(uchar *to)
     {
       /* Reset struct, copy interface name */
       memset(&ifr, 0, sizeof(ifr));
-      strncpy(ifr.ifr_name, ifri->ifr_name, sizeof(ifr.ifr_name));
+      memcpy(ifr.ifr_name, ifri->ifr_name, sizeof(ifr.ifr_name));
 
       /* Get HW address, break if not 0 */
       if (ioctl(fd, SIOCGIFHWADDR, &ifr) >= 0)
--- a/storage/innobase/log/log0online.cc
+++ b/storage/innobase/log/log0online.cc
@@ -1468,7 +1468,7 @@ log_online_setup_bitmap_file_range(
 		if (file_seq_num > bitmap_files->files[array_pos].seq_num) {
 
 			bitmap_files->files[array_pos].seq_num = file_seq_num;
-			strncpy(bitmap_files->files[array_pos].name,
+			memcpy(bitmap_files->files[array_pos].name,
 				bitmap_dir_file_info.name, FN_REFLEN);
 			bitmap_files->files[array_pos].name[FN_REFLEN - 1]
 				= '\0';
--- a/client/mysql_upgrade.c
+++ b/client/mysql_upgrade.c
@@ -1004,7 +1004,7 @@ static int check_version_match(void)
 
 int main(int argc, char **argv)
 {
-  char self_name[FN_REFLEN];
+  char self_name[FN_REFLEN + 1];
 
   MY_INIT(argv[0]);
 
--- a/sql/sql_connect.cc
+++ b/sql/sql_connect.cc
@@ -186,11 +186,17 @@ void init_user_stats(USER_STATS *user_stats,
   DBUG_PRINT("info",
              ("Add user_stats entry for user %s - priv_user %s",
               user, priv_user));
-  strncpy(user_stats->user, user, sizeof(user_stats->user));
-  strncpy(user_stats->priv_user, priv_user, sizeof(user_stats->priv_user));
+  user_stats->user_len=               strlen(user);
+  if (user_stats->user_len >= sizeof(user_stats->user))
+    user_stats->user_len= sizeof(user_stats->user) - 1;
+  memcpy(user_stats->user, user, user_stats->user_len);
+  user_stats->user[user_stats->user_len]= '\0';
 
-  user_stats->user_len=               strlen(user_stats->user);
-  user_stats->priv_user_len=          strlen(user_stats->priv_user);
+  user_stats->priv_user_len=          strlen(priv_user);
+  if (user_stats->priv_user_len >= sizeof(user_stats->priv_user))
+    user_stats->priv_user_len= sizeof(user_stats->priv_user) - 1;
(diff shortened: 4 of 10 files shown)


percona-ysorokin:
Jenkins build links:
https://ps56.cd.percona.com/job/percona-server-5.6-param/27/
https://ps3.cd.percona.com/view/5.6/job/percona-server-5.6-pipeline/1276/ (Debian Buster Release)
https://ps3.cd.percona.com/view/5.6/job/percona-server-5.6-pipeline/1277/ (Debian Buster Debug)

Known OpenSSL 1.1.1 issues and Toku crashes on newer Debian/Ubuntu.

mysys/my_gethwaddr.c
```diff
@@ -108,7 +108,7 @@ my_bool my_gethwaddr(uchar *to)
     {
       /* Reset struct, copy interface name */
       memset(&ifr, 0, sizeof(ifr));
-      strncpy(ifr.ifr_name, ifri->ifr_name, sizeof(ifr.ifr_name));
+      memcpy(ifr.ifr_name, ifri->ifr_name, sizeof(ifr.ifr_name));
```
laurynas-biveinis, line 111: What was the warning fixed?
The only difference between strncpy and memcpy here should have been that strncpy might have stopped earlier on seeing NUL. Since the destination is memset to zero, the result should have been the same

storage/innobase/log/log0online.cc
```diff
@@ -1468,7 +1468,7 @@ log_online_setup_bitmap_file_range(
 		if (file_seq_num > bitmap_files->files[array_pos].seq_num) {
 
 			bitmap_files->files[array_pos].seq_num = file_seq_num;
-			strncpy(bitmap_files->files[array_pos].name,
+			memcpy(bitmap_files->files[array_pos].name,
```
laurynas-biveinis, line 1471: What was the fixed warning?
Our code, not an upstream backport - the commit message should mention this.
This will need an upmerge, right?

laurynas-biveinis (changes requested):

mysys/my_gethwaddr.c
percona-ysorokin:
```
/home/yura/ws/percona-server/mysys/my_gethwaddr.c: In function ‘my_gethwaddr’:
/home/yura/ws/percona-server/mysys/my_gethwaddr.c:111:7: error: ‘strncpy’ specified bound 16 equals destination size [-Werror=stringop-truncation]
       strncpy(ifr.ifr_name, ifri->ifr_name, sizeof(ifr.ifr_name));
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
As far as I understood from https://en.cppreference.com/w/cpp/string/byte/strncpy the main reason for this warning is
`If count is reached before the entire string src was copied, the resulting character array is not null-terminated.`

This fix is from the upstream https://github.com/mysql/mysql-server/commit/a091d6b#diff-8b61af0eaadf2d4f8128d3c5122fecf5, btw

storage/innobase/log/log0online.cc
percona-ysorokin:
```
/home/yura/ws/percona-server/storage/innobase/log/log0online.cc: In function ‘ulint log_online_setup_bitmap_file_range(log_online_bitmap_file_range_t*, lsn_t, lsn_t)’:
/home/yura/ws/percona-server/storage/innobase/log/log0online.cc:1471:11: error: ‘char* strncpy(char*, const char*, size_t)’ output may be truncated copying 512 bytes from a string of length 3999 [-Werror=stringop-truncation]
    strncpy(bitmap_files->files[array_pos].name,
    ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     bitmap_dir_file_info.name, FN_REFLEN);
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Yes, this is our code. And this is the biggest mystery. For some reason, this fragment does not trigger any warnings in 5.7 code.

laurynas-biveinis: @percona-ysorokin check if the array lengths (most likely the destination one) did not change between 5.6 and 5.7

percona-ysorokin: @laurynas-biveinis 
Actually `log_online_bitmap_file_range_struct` definitions in both 5.6 and 5.7 are identical
https://github.com/percona/percona-server/blob/5.6/storage/innobase/include/log0online.h#L145
https://github.com/percona/percona-server/blob/5.7/storage/innobase/include/log0online.h#L144
`FN_REFLEN` in both branches is 512.

`os_file_stat_t` definitions are also identical
https://github.com/percona/percona-server/blob/5.6/storage/innobase/include/os0file.h#L567
https://github.com/percona/percona-server/blob/5.7/storage/innobase/include/os0file.h#L1286
`OS_FILE_MAX_PATH` in both cases is 4000.

I believe this is just a limitation of GCC to detect this warning (5.7 code could be more complicated from its point of view).
In any case, I believe the `bitmap_files->files[array_pos].name[FN_REFLEN - 1] = '\0';` at the next line makes both statements (`strncpy` and `memcpy`) safe here. So, if `memcpy` makes GCC 8.3 happier in 5.6, I would do this change only for 5.6 and leave 5.7 as is.

laurynas-biveinis: @percona-ysorokin I agree, please proceed

percona-ysorokin:
No new Jenkins build links as only commit message updated.

PR #3319, opened 2019-07-17
Fix PS-5780 (The release build (5.6) does not work on Debian buster) (8.0)

percona-ysorokin:


laurynas-biveinis (commented):
Why log0online.cc changes did not need upmerging?

PR #3320, opened 2019-07-18
 PS-5785: MTR issues with changed page tracking tests (5.6)

dutow:
I've added this as two commits to keep the mysql test run change separate (and easily upstreamable).

The issue did not occur on 5.6 or 5.7, I based the changes here to:
* have the same MTR improvement on all branches
* keep the changed tests the same on all branches

#### PS-5785: Fix MTR error reporting when a worker dies.

Previously if for some reason an MTR worker crashed, MTR would continue, but it wouldn't summarize the failures at the end. Checking which checks failed could only be done by analyzing the output, if it was saved.

This change modifies mysql-test-run so:

* reporting continues even if not all tests were completed (a worker crashed, for example because the die call in split_options during a server restart)
* the tests without a status set will be reported at the end. One line per missing testcase, with the format: "Missing result for testcase: <X>"

#### PS-5785: Fix incorrect restart parameters in changed page tests

### DIFF
--- a/mysql-test/mysql-test-run.pl
+++ b/mysql-test/mysql-test-run.pl
@@ -388,6 +388,8 @@ sub main {
 
   mtr_report("Collecting tests...");
   my $tests= collect_test_cases($opt_reorder, $opt_suites, \@opt_cases, \@opt_skip_test_list);
+  my $all_tests;
+  @$all_tests = @$tests;
   mark_time_used('collect');
 
   if ( $opt_report_features ) {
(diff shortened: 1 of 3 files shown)


laurynas-biveinis (approved):
Amazing

laurynas-biveinis:
@dutow on the second thought, what exactly is incorrect with debug options? According to https://dev.mysql.com/doc/refman/5.6/en/dbug-package.html, both syntaxes are correct

dutow:
Edited the commit message of the changed page test changes to make it clearer.

laurynas-biveinis (approved):

dutow:
https://ps56.cd.percona.com/job/percona-server-5.6-param/41/parameters/

peterschwaller:
touch

PR #3324, opened 2019-07-19
PS-5427 Document the correct distribution upgrade process for Percona…

patrickbirch:
… Server

Added the process to upgrade the distribution in-place with Percona Server already installed - 5.7

### DIFF
--- a/doc/source/upgrading_guide_55_56.rst
+++ b/doc/source/upgrading_guide_55_56.rst
@@ -223,3 +223,18 @@ If it can't find the pid file, kill the server and start it normally: ::
 
   $ killall /usr/sbin/mysqld
   $ /sbin/service mysql start
+
+Performing a Distribution upgrade in-place on a System with installed Percona packages
+--------------------------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


laurynas-biveinis:
(changed the base branch from 8.0 to 5.7)

laurynas-biveinis (changes requested):
If 5.6 PR followed GCA, this one does not: there is only one commit, so there is no merge from 5.6

patrickbirch:
i will fix. thanks

patrickbirch:
I have merged with 5.6 and updated the 5.7 upgrading doc. It still comes out as one commit for two files. Is that correct, or should I commit them separately? Thanks.

laurynas-biveinis:
This is not a GCA process, please check the Confluence

patrickbirch:
Thanks. I will double-check and see what I missed.

dutow (approved):
For GCA pull requests, please wait until all versions are reviewed , and merge them together.

Will there be an 8.0 version? If yes, that needs a PR too. If no, it needs a null merge PR.

PR #3325, opened 2019-07-24
PS-5794: Crash when innodb-redo-log-encrypt=ON specified during boots…

dutow:
…trap

### DIFF
--- a/mysql-test/suite/innodb/t/percona_bug_ps5794.test
+++ b/mysql-test/suite/innodb/t/percona_bug_ps5794.test
@@ -1,0 +1,36 @@
+SHOW VARIABLES LIKE "%innodb_redo_log_encrypt%"; 
+
+let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/data1;
+
+let BOOTSTRAP_SQL=$MYSQL_TMP_DIR/boot.sql;
+
+--echo # create bootstrap file
+write_file $BOOTSTRAP_SQL;
+CREATE DATABASE test;
+EOF
+
+--echo # Stop the MTR default DB server
+--source include/shutdown_mysqld.inc
+
+--let $KEYRING_PARAMS=--early-plugin-load="keyring_file=$KEYRING_PLUGIN" --loose-keyring_file_data=$MYSQL_TMP_DIR/mysecret_keyring $KEYRING_PLUGIN_OPT
+let extra_args=--no-defaults --console --loose-skip-auto_generate_certs --loose-skip-sha256_password_auto_generate_rsa_keys --skip-ssl --basedir=$BASEDIR --lc-messages-dir=$MYSQL_SHAREDIR;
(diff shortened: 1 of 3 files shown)


mysql-test/suite/innodb/t/percona_bug_ps5794.test
```diff
@@ -0,0 +1,38 @@
+# restart
```
laurynas-biveinis: Remove

mysql-test/suite/innodb/t/percona_bug_ps5794.test
```diff
@@ -0,0 +1,38 @@
...
+--echo # Stop the MTR default DB server
+--source include/shutdown_mysqld.inc
+
+--let $KEYRING_PARAMS=--early-plugin-load="keyring_file=$KEYRING_PLUGIN" --loose-keyring_file_data=$MYSQL_TMP_DIR/mysecret_keyring $KEYRING_PLUGIN_OPT
+let extra_args=--no-defaults --console --loose-skip-auto_generate_certs --loose-skip-sha256_password_auto_generate_rsa_keys --skip-ssl --basedir=$BASEDIR --lc-messages-dir=$MYSQL_SHAREDIR;
+let NEW_CMD = $MYSQLD $extra_args --innodb_dedicated_server=OFF --initialize-insecure $KEYRING_PARAMS --datadir=$MYSQLD_DATADIR1 --init-file=$BOOTSTRAP_SQL --secure-file-priv="" --innodb-redo-log-encrypt >>$MYSQL_TMP_DIR/bootstrap1.log 2>&1;
+
+--echo # Run the bootstrap command of datadir1
+--exec $NEW_CMD
+--sleep 5
```
laurynas-biveinis: Please sync properly

laurynas-biveinis (changes requested):
5.7?

satya-bodapati:
This looks like wrong fix. Enum shouldn't allow optional arguments. Option value is must.

for
  static MYSQL_SYSVAR_ENUM(redo_log_encrypt, srv_redo_log_encrypt,

you should use PLUGIN_VAR_RQCMDARG instead of PLUGIN_VAR_OPCMDARG

dutow:
Upstream allows it to be empty - so we should allow it to be empty. Also, even if I disallow the empty version, the issue would be there with `=ON`, as that's basically the empty option.

dutow:
In 5.7,  we don't allow the "ON" option, it's not affected.

satya-bodapati (approved):
Please move this testcase to existing redo log encrypt testcase

dutow:
@satya-bodapati I kept the test where it is - other test either have a specific topic or have mk/rk variants both.

laurynas-biveinis (approved):

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/274/

PR #3329, opened 2019-07-29
PS-5803, PS-5804: innodb_redo_log_encrypt shouldn't be persisted

dutow:
Persisted variables are handled differently (after the redo log is already opened),
    and startup continues even with incorrect values, resulting in encryption turning off
    and writing a message to the error log.

    This caused the following issues:

    * PS-5803: changing encryption mode with set persist fails, but we do persist the current
    value
    * PS-5804: changing encryption mode with set persist_only succeeds, and we silently turn
    off encryption at next startup

    As a workaround, we disable persisting this variable.

### DIFF
--- a/mysql-test/r/all_persisted_variables.result
+++ b/mysql-test/r/all_persisted_variables.result
@@ -45,7 +45,7 @@ include/assert.inc ['Expect 625 variables in the table. Due to open Bugs, we are
 
 # Test SET PERSIST
 
-include/assert.inc [Expect 437 persisted variables in the table. Due to open Bugs, we are checking for 426]
+include/assert.inc [Expect 436 persisted variables in the table. Due to open Bugs, we are checking for 425]
 
 ************************************************************
 * 3. Restart server, it must preserve the persisted variable
@@ -53,9 +53,9 @@ include/assert.inc [Expect 437 persisted variables in the table. Due to open Bug
 ************************************************************
 # restart
 
-include/assert.inc [Expect 426 persisted variables in persisted_variables table.]
-include/assert.inc [Expect 426 persisted variables shown as PERSISTED in variables_info table.]
-include/assert.inc [Expect 426 persisted variables with matching peristed and global values.]
+include/assert.inc [Expect 425 persisted variables in persisted_variables table.]
+include/assert.inc [Expect 425 persisted variables shown as PERSISTED in variables_info table.]
+include/assert.inc [Expect 425 persisted variables with matching peristed and global values.]
 
 ************************************************************
 * 4. Test RESET PERSIST IF EXISTS. Verify persisted variable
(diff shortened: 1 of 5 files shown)


laurynas-biveinis:
main.all_persisted_vars?

satya-bodapati (commented):
1. Is this a bug in SET PERSIST framework? To allow invalid values? 
2. Is it because redo_log_encrypt missed validate() implementation?
3. With invalid values, on next startup, innodb doesn't see it?

dutow:
The issue is that persistent values are loaded later than "normal" startup parameters.

Normal startup parameters are set before the redo log is read, so when we open it, we detect the issue and abort startup.

Persistent values are set later after we already opened the redo log. So at that point, our code sees that a "user" (the persistence framework) requests a different value, and turns off encryption with an error message. But I don't see a way to abort startup at this point.

satya-bodapati:
Am not sure if I understand the issue completely. why does this statement pass?
set persist innodb_redo_log_encrypt=keyring_key;

The same statement without persist will fail right? (existing redo encryption mode is master key).

dutow:
@satya-bodapati I've updated the commit message to explain it better. I think the PS-5803 issue is generic, and could be also an issue with other variables, but PS-5804 (persist_only) is specific to redo log encryption.

satya-bodapati:
I realized that I faced similar issue with innodb_sys_tablespace_encrypt. The constraint I had is read-only variable is not allowed to change after it is used for bootstrap. It is also similar here.

I think the real fix is to allow to persist only the current_value and not any other values. Am not sure though if it can be done easily. if you think it is possible to do, please create JIRA feature request.

satya-bodapati:
One more concern is that we are disallowing persist of variable which upstream allows.

dutow:
Yeah, I realize that - but it's either that, or we'll have to ship it with this bug. Maybe @laurynas-biveinis  should decide if this is okay or not?

I also can't see how could we easily fix it - moving as much as we can to the validate method again would make things a bit better, but not perfect.

dutow:
Even if I do manage to move everything to validate, somebody could still do:

* start with an unencrypted redo log
* set persist_only redo log encryption to master_key
* set redo log encryption to keyring_key
* restart the server

And some other similar edge cases...

satya-bodapati (approved):
I think we can work on enabling the persistence later.

laurynas-biveinis:
There must be an upstream bug in here. The logic running too late should mean that upstream should be affected by OFF/ON persisting the same

dutow:
But ON/OFF isn't that bad, most likely they didn't even notice it - the issue here is our code handling keyring_key and master_key changes (not allowing it)

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/271/

PR #3332, opened 2019-07-30
PS-5790: keyring_vault tests are failing on Jenkins

robgolebiowski:
There is a bug in curl on Debian and it seems to not support http v2.
Because of that we should fall back to HTTP 1.1 in keyring_vault.
We should check when new fixes come around on Debian and re-check the
problem. Re-enable HTTP v2 in keyring_vault when the problem goes away
on Debian. There should not be any negative consequences security-wise
as all communication happens over TLS.

### DIFF
--- a/plugin/keyring_vault/vault_curl.cc
+++ b/plugin/keyring_vault/vault_curl.cc
@@ -208,6 +208,9 @@ bool Vault_curl::setup_curl_session(CURL *curl)
       (curl_res= curl_easy_setopt(curl, CURLOPT_CONNECTTIMEOUT, timeout)) !=
           CURLE_OK ||
       (curl_res= curl_easy_setopt(curl, CURLOPT_TIMEOUT, timeout)) !=
+          CURLE_OK ||
+      (curl_res = curl_easy_setopt(curl, CURLOPT_HTTP_VERSION,
+                                   (long)CURL_HTTP_VERSION_1_1)) !=
           CURLE_OK)
   {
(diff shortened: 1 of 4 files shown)


robgolebiowski:
https://ps57.cd.percona.com/job/percona-server-5.7-param/119/

laurynas-biveinis (commented):
Is there a Debian bug #?

robgolebiowski:
I do not see any bug reported against it. If you want me I can report it to Vault and ask them to investigate it?

laurynas-biveinis:
Please do - stuff like this needs some end game plan

robgolebiowski:
@laurynas-biveinis Created bug against Vault: https://github.com/hashicorp/vault/issues/7320

laurynas-biveinis (approved):

plugin/keyring_vault/vault_curl.cc
```diff
@@ -208,6 +208,9 @@ bool Vault_curl::setup_curl_session(CURL *curl)
       (curl_res= curl_easy_setopt(curl, CURLOPT_CONNECTTIMEOUT, timeout)) !=
           CURLE_OK ||
       (curl_res= curl_easy_setopt(curl, CURLOPT_TIMEOUT, timeout)) !=
+          CURLE_OK ||
```
laurynas-biveinis, line 211: Please reference the bug report at least there, and maybe in the MTR patch too

laurynas-biveinis (changes requested):

PR #3336, opened 2019-07-30
PS-5655 Change default of rocksdb_large_prefix in 8.0 to TRUE like In…

patrickbirch:
…noDB

Changed the default value and added a note that states the change was implemented in version 8.0.16-7 and later.


george-lorch (approved):
LGTM for real this time ;)

PR #3338, opened 2019-07-31
PS-5775: Improve redo log encryption error messages (8.0)

dutow:
* If redo log encryption fails becasuse a keyring errors, this should be
reported to the user instead of a generic error message.
* User facing errors are simple ib::errors now instead of ib::fatal, to
avoid the debug assertion there

(cherry picked from commit 1fd5c9def0b19489ef7e685b7aca771eadc53bc6)


satya-bodapati (approved):

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/270/ - I've rerecorded the irrodb_redo_log_encrypt sys_vars test, others seem to be already existing on trunk.

PR #3339, opened 2019-08-01
PS-5584 Update PS-5.7 feature comparison

patrickbirch:
Changed layout of tables to list-table
Added items from spreadsheet made during the 8.0 GA

### DIFF
--- a/doc/source/feature_comparison.rst
+++ b/doc/source/feature_comparison.rst
@@ -20,111 +20,263 @@
 
 We provide these benefits by significantly enhancing |Percona Server| as compared to the standard |MySQL| database server:
 
-+---------------------------------------+----------------------+--------------+
-|Features                               |Percona Server 5.7.10 |MySQL 5.7.10  |
-+=======================================+======================+==============+
-|Open source                            | Yes                  | Yes          |      
-+---------------------------------------+----------------------+--------------+
-|ACID Compliance                        | Yes                  | Yes          |                     
-+---------------------------------------+----------------------+--------------+
-|Multi-Version Concurrency Control      | Yes                  | Yes          |                      
-+---------------------------------------+----------------------+--------------+
-|Row-Level Locking                      | Yes                  | Yes          |
-+---------------------------------------+----------------------+--------------+
-|Automatic Crash Recovery               | Yes                  | Yes          |
-+---------------------------------------+----------------------+--------------+
-|Table Partitioning                     | Yes                  | Yes          |
-+---------------------------------------+----------------------+--------------+
-|Views                                  | Yes                  | Yes          |
-+---------------------------------------+----------------------+--------------+
-|Subqueries                             | Yes                  | Yes          |
-+---------------------------------------+----------------------+--------------+
-|Triggers                               | Yes                  | Yes          |
-+---------------------------------------+----------------------+--------------+
-|Stored Procedures                      | Yes                  | Yes          |
-+---------------------------------------+----------------------+--------------+
-|Foreign Keys                           | Yes                  | Yes          |
-+---------------------------------------+----------------------+--------------+
-|GTID Replication                       | Yes                  | Yes          |
-+---------------------------------------+----------------------+--------------+
+.. tabularcolumns:: |p{5cm}|p{5cm}|p{5cm}|
 
-+---------------------------------------+----------------------+--------------+
-|Extra Features for Developers          |Percona Server 5.7.10 |MySQL 5.7.10  |
-+=======================================+======================+==============+
-|NoSQL Socket-Level Interface           | Yes                  | Yes          |
-+---------------------------------------+----------------------+--------------+
-|Extra Hash/Digest Functions            | Yes                  |              |
-+---------------------------------------+----------------------+--------------+
+.. list-table::
+    :header-rows: 1
 
-+---------------------------------------+----------------------+--------------+
-|Extra Diagnostic Features              |Percona Server 5.7.10 |MySQL 5.7.10  |
-+=======================================+======================+==============+
-|INFORMATION_SCHEMA Tables              | 75                   | 61           |
-+---------------------------------------+----------------------+--------------+
-|Global Performance and Status Counters | 385                  | 355          |
-+---------------------------------------+----------------------+--------------+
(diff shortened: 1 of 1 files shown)


doc/source/feature_comparison.rst
```diff
@@ -20,111 +20,260 @@
...
+   :header-rows: 1
+
+   * - Instrumentation and Troubleshooting Features
+     - Percona Server 5.7.24
+     - MySQL 5.7.24
+   * - INFORMATION_SCHEMA Tables
+     - 93
+     - 61
+   * - Global Performance and Status Counters
+     - 789
```
dutow: This was 389 in the original table? And 789 looks too much to me, are you sure?

doc/source/feature_comparison.rst
```diff
@@ -20,111 +20,260 @@
...
+.. tabularcolumns:: |p{5cm}|p{5cm}|p{5cm}|
+
+.. list-table::
+   :header-rows: 1
+
+   * - Instrumentation and Troubleshooting Features
+     - Percona Server 5.7.24
+     - MySQL 5.7.24
+   * - INFORMATION_SCHEMA Tables
+     - 93
```
dutow: This was 75 in the original table - did we implement 18 new since then?

doc/source/feature_comparison.rst
```diff
@@ -20,111 +20,260 @@
...
-+---------------------------------------+----------------------+--------------+
-|Enforcing the Specific Storage Engine  | Yes                  |              |
-+---------------------------------------+----------------------+--------------+
+.. tabularcolumns:: |p{5cm}|p{5cm}|p{5cm}|
+
+.. list-table::
+    :header-rows: 1
+
+    * - Features
+      - Percona Server 5.7.24
```
dutow: Shouldn't this be .27 now?

doc/source/feature_comparison.rst
```diff
@@ -20,111 +20,260 @@
...
+   * - Encrypt InnoDB logs
+     - Yes
+     - No
+   * - Encrypt Binary logs
+     - Yes
+     - No
+   * - Encrypt temporary files
+     - Yes
+     - No
+   * - Key Rotation with scrubbing
```
dutow: Scrubbing - I wouldn't even call it experimental yet. Also, key rotation and scrubbing are separate things. For several other features here (log encryption for example): these are experimental, not yet GA, I don't think we should write a definite "yes" here.

dutow (changes requested):

doc/source/feature_comparison.rst
patrickbirch: According to the PS Feature Comparison spreadsheet made during 8.0 GA (Tyler)
Global Performance and Status Counters - 789

doc/source/feature_comparison.rst
patrickbirch: According to the PS Feature Comparison spreadsheet made during 8.0 GA (Tyler)
INFORMATION_SCHEMA Tables - 93

doc/source/feature_comparison.rst
patrickbirch: The information for the PS Feature Comparison was for PS 5.7.24 compared to MySQL 5.7.24.

doc/source/feature_comparison.rst
dutow: I've just executed `show tables` on the current 5.7 turnk, and I see 71 tables - not even 75. So it would be great for somebody who actually does release testing of our software (QA?) to verify these numbers.

doc/source/feature_comparison.rst
dutow: Similarly here, I'm getting 432 on the 5.7 trunk...

doc/source/feature_comparison.rst
dutow: And did anything change since then? I don't think so, upstream only did bugfixes, same for us. So I would go with .27, since that's the latest version. @georgelorchpercona ?

doc/source/feature_comparison.rst
patrickbirch: Updated with Experimental.

doc/source/feature_comparison.rst
george-lorch: Agreed, I don't think anything changed so should probably be 5.7.27

doc/source/feature_comparison.rst
george-lorch: Yes, we should have some doc QA, but realistically, this to me seems like 100% useless information anyway.  x number of  i_s tables vs y makes not one bit of difference to an end user.  What they do care about is what those tables do.

dutow: The other number (432  or 789) is more significant, this one is a smaller difference, but still. I would go with the smaller numbers then - either the original, or what I see with running the server.

doc/source/feature_comparison.rst
patrickbirch: updated to 5.7.27

doc/source/feature_comparison.rst
patrickbirch: changed to 71.

doc/source/feature_comparison.rst
patrickbirch: changed to 432

doc/source/feature_comparison.rst
patrickbirch: Key Rotation has its own row. Scrubbing has its own row.

doc/source/feature_comparison.rst
```diff
@@ -20,111 +20,263 @@
...
+   * - Encrypt Binary logs
+     - Experimental
+     - No
+   * - Encrypt temporary files
+     - Experimental
+     - No
+   * - Key Rotation
+     - Experimental
+     - No
+   * - scrubbing
```
dutow: **S**crubbing

dutow (approved):
One small thing otherwise looks good

doc/source/feature_comparison.rst
patrickbirch: fixed. Thanks.

PR #3341, opened 2019-08-01
PS-5820: Crashes with the slow query log enabled (5.7)

dutow:
Issue: there is a race condition between different threads calling
Bloom_filter::clear and Bloom_filter::test_and_set.
Before the fix for PS-5581, races always resulted in test_and_set
returning false. After the fix, clear is no longer a single pointer
change, and test_and_set could also return true in these situations.

Root cause: PS-5581, 715f78f

Workaround: restore the previous behavior by switching the single
MEM_ROOT allocated Bitset per THD to shared_ptrs. The race is
still there, but the results should be the same as before PS-5581.

### DIFF
--- a/sql/sql_class.h
+++ b/sql/sql_class.h
@@ -67,6 +67,7 @@ typedef struct st_mysql_lex_string LEX_STRING;
 typedef struct user_conn USER_CONN;
 
 extern ulong kill_idle_transaction_timeout;
+extern PSI_mutex_key key_LOCK_bloom_filter;
 
 /**
   The meat of thd_proc_info(THD*, char*), a macro that packs the last
@@ -1588,20 +1589,32 @@ class Bloom_filter {
 
   typedef std::bitset<SIZE> Bit_set;
 
-  /** The bit set, which is allocated in a MEM_ROOT */
-  Bit_set *bit_set;
+
+  Bit_set* bit_set;
+  // Protects the bit set from concurrent assignment
+  mysql_mutex_t LOCK_bit_set;
 
   // Non-copyable
   Bloom_filter (const Bloom_filter &);
   Bloom_filter & operator = (const Bloom_filter &);
 
  public:
-  Bloom_filter() : bit_set(NULL) {}
+  Bloom_filter() : bit_set(NULL) {
+    mysql_mutex_init(key_LOCK_bloom_filter,
+                     &LOCK_bit_set, MY_MUTEX_INIT_FAST);
+  }
+
+  ~Bloom_filter() {
+    clear();
+    mysql_mutex_destroy(&LOCK_bit_set);
+  }
 
   void clear()
   {
-    if (bit_set != NULL)
-      bit_set->reset();
+    mysql_mutex_lock(&LOCK_bit_set);
+    delete bit_set;
+    bit_set= NULL;
+    mysql_mutex_unlock(&LOCK_bit_set);
   }
 
   /**
@@ -1614,23 +1627,22 @@ class Bloom_filter {
      @return if true, the key might be a member of the set. If false, the key
      is definitely not a member of the set.
   */
-  bool test_and_set(MEM_ROOT *mem_root, ulong key)
+  bool test_and_set(ulong key)
   {
-    if (!bit_set)
+    mysql_mutex_lock(&LOCK_bit_set);
+    if (bit_set == NULL)
     {
-      void *bit_set_place= alloc_root(mem_root, sizeof(Bit_set));
-      // FIXME: memory allocation failure is eaten silently. Nonexact stats are
-      // the least of the concerns then.
-      if (bit_set_place == NULL)
-        return false;
-      bit_set= new (bit_set_place) Bit_set();
+      bit_set= new Bit_set();
     }
     // Duplicating ut_hash_ulint calculation
     const ulong pos= (key ^ 1653893711) % SIZE;
     DBUG_ASSERT(pos < SIZE);
-    if (bit_set->test(pos))
+    if (bit_set->test(pos)) {
+      mysql_mutex_unlock(&LOCK_bit_set);
       return false;
+    }
     bit_set->set(pos);
+    mysql_mutex_unlock(&LOCK_bit_set);
     return true;
   }
 };
@@ -2229,7 +2241,7 @@ public:
 
   void access_distinct_page(ulong page_id)
   {
-    if (approx_distinct_pages.test_and_set(&main_mem_root, page_id))
+    if (approx_distinct_pages.test_and_set(page_id))
       innodb_page_access++;
   }
 
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -9189,6 +9189,7 @@ static PSI_mutex_info all_server_mutexes[]=
   { &key_RELAYLOG_LOCK_xids, "MYSQL_RELAY_LOG::LOCK_xids", 0},
   { &key_hash_filo_lock, "hash_filo::lock", 0},
   { &Gtid_set::key_gtid_executed_free_intervals_mutex, "Gtid_set::gtid_executed::free_intervals_mutex", 0 },
+  { &key_LOCK_bloom_filter, "Bloom_filter", 0},
   { &key_LOCK_crypt, "LOCK_crypt", PSI_FLAG_GLOBAL},
   { &key_LOCK_error_log, "LOCK_error_log", PSI_FLAG_GLOBAL},
   { &key_LOCK_global_user_client_stats,
--- a/sql/sql_class.cc
+++ b/sql/sql_class.cc
@@ -65,6 +65,7 @@ using std::min;
 using std::max;
 
 ulong kill_idle_transaction_timeout= 0;
+PSI_mutex_key key_LOCK_bloom_filter;
 
 /*
   The following is used to initialise Table_ident with a internal


dutow:
Also: this restores the previous behavior before PS-5581, with the memory usage reduced, but I think this is only a workaround if we want accurate statistics.

sql/sql_class.h
```diff
@@ -1614,23 +1615,21 @@ class Bloom_filter {
...
+    if (!bit_set_copy)
     {
-      void *bit_set_place= alloc_root(mem_root, sizeof(Bit_set));
-      // FIXME: memory allocation failure is eaten silently. Nonexact stats are
-      // the least of the concerns then.
-      if (bit_set_place == NULL)
-        return false;
-      bit_set= new (bit_set_place) Bit_set();
+      // TODO: possible inaccurate statistics because of races?
+      bit_set_copy= Bit_set_ptr(new Bit_set());
```
percona-ysorokin: I believe there is also boost::make_shared
```suggestion
      bit_set_copy= boost::make_shared<Bit_set>();
```

sql/sql_class.h
```diff
@@ -1614,23 +1615,21 @@ class Bloom_filter {
...
+      bit_set= bit_set_copy;
```
percona-ysorokin: This call is not thread-safe. In c++11 I would use std::atomic_exchange here
https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic
I believe there must be something similar in boost

sql/sql_class.h
```diff
@@ -1614,23 +1615,21 @@ class Bloom_filter {
...
     }
     // Duplicating ut_hash_ulint calculation
     const ulong pos= (key ^ 1653893711) % SIZE;
     DBUG_ASSERT(pos < SIZE);
-    if (bit_set->test(pos))
+    if (bit_set_copy->test(pos))
       return false;
-    bit_set->set(pos);
+    bit_set_copy->set(pos);
```
percona-ysorokin: `set()` method on an `std::bitset<N>` class is not atomic either - it still does
```
mov <register>, internal_array[index]
or <register>, <mask>
mov internal_array[index], <register>
```
internally.
Unless we can ignore some inaccuracies in the statistics, this needs to be reworked as well. I would implement a simple  helper class `atomic_bitset` or something. which would internally hold `std::array<std::atomic_ullong, N>`.

sql/sql_class.h
dutow: Before C++20, there is no atomic_shared_ptr in any of the standard libraries - using atomic_exchange basically means a mutex+lock. (and I'm not sure if now that C++20 requires it, modern compilers will or will not "backport" it to earlier standards)

So if we need thread safety here, we could also use a mysql lock like everywhere else. (but I'm not sure if we need. To me, it looks like two innodb workers can't call this function simultaneously, but I could be wrong here)

sql/sql_class.h
dutow: We have to ignore some inaccuracies - the entire issue is an inaccuracy: if test_and_set is called after clear, we are inaccurate - we miss some statistics for the current statement, and we assign some to the next executed statement... but it happens, as we report these stats from some background workers.

And again, I'm not sure if this method is called concurrently. But if we want to, I can change this, there's a boost::array for 5.7 too.

sql/sql_class.h
dutow: One possible multi-worker issue is that if workers are way too behind (at least 2 statements/commits). I'm not sure if that can happen in a real scenario, but theoretically should be possible. Will update with locking here.

sql/sql_class.h
```diff
@@ -1588,20 +1590,29 @@ class Bloom_filter {
 
   typedef std::bitset<SIZE> Bit_set;
 
-  /** The bit set, which is allocated in a MEM_ROOT */
-  Bit_set *bit_set;
+  typedef boost::shared_ptr<Bit_set> Bit_set_ptr;
```
percona-ysorokin: @dutow I have a strange feeling that using `shared_ptr` here will case us more race condition troubles despite the fact that it may potentially simplify the code.
Although I don't like this either, for 5.7 I would go with canonical volatile pointer approach.
```suggestion
  typedef Bit_set* Bit_set_ptr;
```

sql/sql_class.h
```diff
@@ -1588,20 +1590,29 @@ class Bloom_filter {
...
 
  public:
-  Bloom_filter() : bit_set(NULL) {}
+  Bloom_filter() {
+    mysql_mutex_init(key_LOCK_bloom_filter,
+                   &LOCK_bit_set, MY_MUTEX_INIT_FAST);
+  }
+
+  ~Bloom_filter() {
+    mysql_mutex_destroy(&LOCK_bit_set);
```
percona-ysorokin, line 1609:
```suggestion
    clear();
    mysql_mutex_destroy(&LOCK_bit_set);
```

sql/sql_class.h
```diff
@@ -1588,20 +1590,29 @@ class Bloom_filter {
 
   typedef std::bitset<SIZE> Bit_set;
 
-  /** The bit set, which is allocated in a MEM_ROOT */
-  Bit_set *bit_set;
+  typedef boost::shared_ptr<Bit_set> Bit_set_ptr;
+
+  Bit_set_ptr bit_set;
```
percona-ysorokin:
```suggestion
  volatile Bit_set_ptr bit_set;
```

sql/sql_class.h
```diff
@@ -1588,20 +1590,29 @@ class Bloom_filter {
...
+  // Protects the bit set from concurrent assignment
+  mysql_mutex_t LOCK_bit_set;
 
   // Non-copyable
   Bloom_filter (const Bloom_filter &);
   Bloom_filter & operator = (const Bloom_filter &);
 
  public:
-  Bloom_filter() : bit_set(NULL) {}
+  Bloom_filter() {
```
percona-ysorokin:
```suggestion
  Bloom_filter() : bit_set(NULL) {
```

sql/sql_class.h
```diff
@@ -1614,23 +1625,26 @@ class Bloom_filter {
      @return if true, the key might be a member of the set. If false, the key
      is definitely not a member of the set.
   */
-  bool test_and_set(MEM_ROOT *mem_root, ulong key)
+  bool test_and_set(ulong key)
   {
-    if (!bit_set)
+    Bit_set_ptr bit_set_copy= bit_set;
```
percona-ysorokin:
```suggestion
    Bit_set_ptr local_bit_set = my_atomic_loadptr(&bit_set);
```

sql/sql_class.h
```diff
@@ -1614,23 +1625,26 @@ class Bloom_filter {
...
+    if (!bit_set_copy)
```
percona-ysorokin:
```suggestion
    if (local_bit_set != NULL)
```

sql/sql_class.h
```diff
@@ -1614,23 +1625,26 @@ class Bloom_filter {
...
     {
-      void *bit_set_place= alloc_root(mem_root, sizeof(Bit_set));
-      // FIXME: memory allocation failure is eaten silently. Nonexact stats are
-      // the least of the concerns then.
-      if (bit_set_place == NULL)
-        return false;
-      bit_set= new (bit_set_place) Bit_set();
+      mysql_mutex_lock(&LOCK_bit_set);
+      // TODO: possible inaccurate statistics because of races?
+      if (!bit_set_copy)
```
percona-ysorokin:
```suggestion
      local_bit_set = my_atomic_loadptr(&bit_set);
      if (local_bit_set != NULL)
```

sql/sql_class.h
```diff
@@ -1614,23 +1625,26 @@ class Bloom_filter {
...
+      {
+        bit_set_copy= Bit_set_ptr(new Bit_set());
```
percona-ysorokin:
```suggestion
        local_bit_set_copy= new Bit_set;
```

sql/sql_class.h
```diff
@@ -1614,23 +1625,26 @@ class Bloom_filter {
...
+        bit_set= bit_set_copy;
```
percona-ysorokin:
```suggestion
        my_atomic_storeptr(&bit_set, local_bit_set);
```

sql/sql_class.h
```diff
@@ -1614,23 +1625,26 @@ class Bloom_filter {
...
+      }
+      mysql_mutex_unlock(&LOCK_bit_set);
     }
     // Duplicating ut_hash_ulint calculation
     const ulong pos= (key ^ 1653893711) % SIZE;
     DBUG_ASSERT(pos < SIZE);
-    if (bit_set->test(pos))
+    if (bit_set_copy->test(pos))
```
percona-ysorokin:
```suggestion
    if (local_bit_set->test(pos))
```

sql/sql_class.h
```diff
@@ -1614,23 +1625,26 @@ class Bloom_filter {
...
       return false;
-    bit_set->set(pos);
+    bit_set_copy->set(pos);
```
percona-ysorokin:
```suggestion
    local_bit_set->set(pos);
```

CMakeLists.txt
```diff
@@ -534,7 +534,10 @@ INCLUDE_DIRECTORIES(
   ${CMAKE_CURRENT_BINARY_DIR}/include
   ${CMAKE_SOURCE_DIR}/extra/rapidjson/include
   ${CMAKE_CURRENT_BINARY_DIR}/libbinlogevents/include
-  ${CMAKE_SOURCE_DIR}/libbinlogevents/export)
+  ${CMAKE_SOURCE_DIR}/libbinlogevents/export
```
percona-ysorokin: Not needed if shared_ptr is not used.
See below

sql/sql_class.h
```diff
@@ -50,6 +50,7 @@
 
 #include <bitset>
 #include <memory>
+#include <boost/shared_ptr.hpp>
```
percona-ysorokin:
```suggestion
```

sql/sql_class.h
```diff
@@ -1588,20 +1590,29 @@ class Bloom_filter {
...
+
+  ~Bloom_filter() {
+    mysql_mutex_destroy(&LOCK_bit_set);
+  }
 
   void clear()
   {
-    if (bit_set != NULL)
-      bit_set->reset();
+    bit_set.reset();
```
percona-ysorokin: Here, I would also protect this with a mutex.
```suggestion
    mysql_mutex_lock(&LOCK_bit_set);
    Bit_set_ptr local_bit_set = my_atomic_loadptr(&bit_set);
    delete local_bit_set;
    my_atomic_storeptr(&bit_set, NULL);
    mysql_mutex_unlock(&LOCK_bit_set);
```

percona-ysorokin (changes requested):
All of this makes sense only if there cannot be concurrent calls between `test_and_set()` and `clear()`. Otherwise, I don't see any way to implement this other than just always protecting everything with a mutex.

sql/sql_class.h
dutow: And for 8.0, should I keep shared pointers?

dutow:
Updated.

percona-ysorokin (approved):
LGTM

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/128/

percona-ysorokin:
@dutow `perfschema.dml_setup_instruments` MTR test case should have been re-recorded as part of this commit because of the new `key_LOCK_bloom_filter` being instrumented.

PR #3342, opened 2019-08-01
PS-5820: Crashes with the slow query log enabled (8.0)

dutow:
Issue: there is a race condition between different threads calling
Bloom_filter::clear and Bloom_filter::test_and_set.
Before the fix for PS-5581, races always resulted in test_and_set
returning false. After the fix, clear is no longer a single pointer
change, and test_and_set could also return true in these situations.

Root cause: PS-5581, 715f78f

Workaround: restore the previous behavior by switching the single
MEM_ROOT allocated Bitset a per THD locked raw pointer. The race is
still there, but the results should be the same as before PS-5581.

(cherry-picked from commit bde9bf2)


dutow:
Also: this restores the previous behavior before PS-5581, with the memory usage reduced, but I think this is only a workaround if we want accurate statistics.

dutow:
Updated based on 5.7

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/285/

percona-ysorokin (approved):
LGTM

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/285/

percona-ysorokin:
@dutow `perfschema.dml_setup_instruments` MTR test case should have been re-recorded as part of this commit because of the new `key_LOCK_bloom_filter` being instrumented.

laurynas-biveinis:
Not that it could be fixed now, but is the commit message current?

dutow:
The commit message is, the PR description wasn't. I edited it.

PR #3343, opened 2019-08-02
PS-5821: Revert explicit encryption from 8.0.16 added by PS-5772 8.0.16

robgolebiowski:
encryption reconciliation

The explicit_encryption added to DD should be removed. The information
whether tablespace will get encrypted with encryption threads
should be visible via I_S (see PS-5819). After this change
SHOW CREATE TABLE will be fully compatible with upstream - i.e. showing
ENCRYPTION='N' only when unenencrypted table exists in schema with
DEFAULT_ENCRYPTION='Y'. Previous to this revert - explicit
ENCRYPTION='N' was always shown in SHOW CREATE TABLE.


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/278 (full encryption suite with all encryption threads tests enabled) --suite=encryption --big-test
https://ps80.cd.percona.com/job/percona-server-8.0-param/277/ (rest of the tests) --suite=encryption,innodb,main

sql/dd/impl/types/abstract_table_impl.cc
```diff
@@ -61,35 +61,33 @@ namespace dd {
...
-     "stats_persistent",
-     "stats_sample_pages",
-     "storage",
-     "tablespace",
-     "timestamp",
-     "view_valid",
-     "encryption_key_id",     // Added by Percona InnoDB rotated key encryption
-     "explicit_encryption"};  // Added by Percona InnoDB rotated key
-                              // encryption
+static const std::set<String_type> default_valid_option_keys = {
```
laurynas-biveinis, line 64: OK, since CircleCI passes, the formatting must have been wrong before

laurynas-biveinis (approved):

PR #3344, opened 2019-08-02
Merge upstream 5.7.27 into Percona Server (outdated)

inikep:
https://jira.percona.com/browse/PS-5532

***
VERSION raised to "5.7.27-30".
univ.i version raised to "30".

***
PS-4907: Inconsistent ResultSet is obtained for different Execution Plans
Our fix for PS-4907 left intact as it is identical to the upstream fix (only EOL added by upstream).
https://jira.percona.com/browse/PS-4907
https://bugs.mysql.com/bug.php?id=92809

***
PS-5221: Page Cleaner will sleep for long time if clock changes
Our fix for PS-5221 was reverted in a separate commit at 90a212c
https://jira.percona.com/browse/PS-5221
Upstream is fix at 6ce62a4

***
Re-recorded MTR test because of changed `errmsg-utf8.txt`:
mysql-test/suite/innodb/r/percona_log_encrypt_change_mk.result
mysql-test/suite/innodb/r/percona_log_encrypt_change_rk.result
mysql-test/suite/innodb/r/xtradb_compressed_columns_mysqldump.result
mysql-test/suite/innodb/r/xtradb_compressed_columns_with_dictionaries.result
mysql-test/suite/innodb/r/percona_log_encrypt_change_mk.result
mysql-test/suite/innodb/r/percona_log_encrypt_change_rk.result
mysql-test/suite/innodb/r/xtradb_compressed_columns_mysqldump.result
mysql-test/suite/innodb/r/xtradb_compressed_columns_with_dictionaries.result
mysql-test/suite/sys_vars/r/innodb_redo_log_encrypt_basic.result

****
Add `class nonassignable` in `sql/field.h` to solve the issue with gcc-9 (c++11 mode) warning -Wdeprecated-copy.

***
TokuDB (storage/tokudb/CMakeLists.txt) - disable warnings for gcc-9 or newer: -Waddress-of-packed-member -Wdeprecated-copy

***
Fix "PS-5836: clang-8 reports broken source encoding in scripts/sql_commands_help_data.h".
scripts/comp_sql.c: utf-8 encoded characters shouldn't be split into two separate lines
Fixes clang-8 warning `-Winvalid-source-encoding`.

***
os_event::os_event(void) is run twice for mutexes declared statically (second call with mutex_create).
Implement lazy init pattern - make `global_initialized` atomic and call os_event_global_init if it’s not set.
In storage/innobase/os/os0event.cc:
- os_event::global_initialized changed to atomic
- os_event::os_event(void) calls os_event_global_init() which is run only once using os_event::global_initialized

***
Fix `query_response_time.install_query_response_time` and `audit_log_default_db` by reconnect of `default` connection after mysql/mysql-server@6a893b8

***
Fix "PS-5828: incorrect length in read_ok_ex() in sql-common/client.c:805".
Change `sizeof(charset_name - 1)` to `(sizeof(charset_name) - 1)` in `read_ok_ex()`.
The issue was reported by clang-8.

***
RocksDB: Do not treat warnings as errors for gcc-8+ or clang-8+

***
include/m_string.h: added `my_strncpy_trunc()` to fix many gcc-9 -Wstringop-truncation warnings in Release compilation.


libbinlogevents/include/table_id.h
```diff
@@ -39,6 +39,7 @@ class Table_id
 public:
   Table_id() : m_id(0) {}
   explicit Table_id(unsigned long long id) : m_id(id) {}
+  Table_id(const Table_id& tid) { m_id = tid.m_id; }
```
percona-ysorokin, line 42: Is this necessary to get rid of a warning? Otherwise auto-generated copy constructor looks fine.

sql-common/client.c
```diff
@@ -802,7 +802,7 @@ void read_ok_ex(MYSQL *mysql, ulong length) {
             if (is_charset == 1) {
               char charset_name[MY_CS_NAME_SIZE * 8]; // MY_CS_BUFFER_SIZE
               size_t length = data->length > (sizeof(charset_name) - 1)
-                                  ? sizeof(charset_name - 1)
+                                  ? (sizeof(charset_name) - 1)
```
percona-ysorokin, line 805: This looks like an upstream bug - should be reported separately.
Please, create a PS-NNN bug, an upstream bug and specify in the commit message `this merge commit also fixes PS-NNN`.

sql-common/client.c
```diff
@@ -2700,75 +2696,30 @@ static int ssl_verify_server_cert_san(X509 *server_cert,
 /*
   Check the server's (subject) Common Name against the
   hostname we connected to
-||||||| merged common ancestors
```
percona-ysorokin: This does not look right. `ssl_verify_server_cert` contains our Percona-specific extensions for validating host names with SAN-enabled SSL certificates. So, basically, we do better than upstream on platforms where there is no OpenSSL with native SAN validation. So, these changes should be preserved. See https://jira.percona.com/browse/PS-1076 for more details.

However, Oracle did really bad in https://github.com/mysql/mysql-server/commit/253d846f5904983bb4a4a64baa9237955c060237 with both clang-formatting `sql-common/client.c` /  `sql/session_tracker.cc` and fixing real problem `Bug #29297652: STACK OVERFLOW IN READ_OK_EX() WITH SESSION_STATE_CHANGED` in the same commit.
So, It looks like the proper way to do the merge would be to create one more prerequisite commit in this PR (just before the merge commit) in which you would just clang-format those 2 files.

sql/field.h
```diff
@@ -566,10 +566,19 @@ class Proto_field
...
+{
+protected:
+    nonassignable() {}
+    ~nonassignable() {}
+    nonassignable( const nonassignable& ) {}
+private:  /* emphasize the following member is private */
+    nonassignable& operator=( const nonassignable& );
+};
+
+class Field: public Proto_field, public nonassignable
```
percona-ysorokin: Please, reorder the classes being inherited. If you have `nonassignable` (an empty class) going first in the list the compiler will be able to use empty base class optimization and therefore the size of the `Field` class will not grow.
```suggestion
class Field: private nonassignable, public Proto_field
```

sql/share/errmsg-utf8.txt
```diff
@@ -7777,7 +7777,26 @@ ER_XA_REPLICATION_FILTERS
...
+
+#
+#  End of 5.7 error messages.
+#
+
+#
+# Start of Percona Server 5.7 error messages
+#
+
+start-error-number 11000
```
percona-ysorokin, line 7794: Do you think we should try to synchronize this with 8.0 Percona-specific error code numbers as well?
@georgelorch do you have any experience with this?

percona-ysorokin (changes requested):

percona-ysorokin:
In the commit message, please also specify which upstream commit covered our fix for
```
PS-5221: Page Cleaner will sleep for long time if clock changes
Our fix for PS-5221 was reverted in a separate commit at 90a212c
https://jira.percona.com/browse/PS-5221
```

percona-ysorokin:
In case you decide to re-apply your merge diff with that new clang-format prerequisite commit, please also change the commit message and title for
`Revert "Merge remote-tracking branch 'local-5.6/ps-5.6-5221' into ps-5.7-5221".
It would help a lot for those who would be browsing the code to see that this is also a part of Upstream 5.7.27 merge.
Something like
```
Implemented PS-5532 (Merge MySQL 5.7.27) (reverted commit XXXX)

https://jira.percona.com/browse/PS-5532
```

percona-ysorokin:
There is also no MTR test case for https://bugs.mysql.com/bug.php?id=94176.

sql/share/errmsg-utf8.txt
george-lorch: I believe they are/should already be synced.  MyRocks specific code start at 12000 for both 5.7 and 8.0

dutow:
> In case you decide to re-apply your merge diff with that new clang-format prerequisite commit, please also change the commit message and title for

That's also doable without re-applying the diff manually with `git rebase -i` (and that also could help with injecting that reformatting commit)

libbinlogevents/include/table_id.h
inikep: It fixes gcc-9 warning:
```
/data/mysql-server/5.7/libbinlogevents/include/table_id.h:64:12: error: implicitly-declared ‘constexpr Table_id::Table_id(const Table_id&)’ is deprecated [-Werror=deprecated-copy]
```

sql-common/client.c
inikep: Reported at:
https://bugs.mysql.com/bug.php?id=96433
https://jira.percona.com/browse/PS-5828

sql-common/client.c
inikep: These conflicts are badly detected because of clang reformatting. All conflicts are resolved taking `HEAD` what keeps original Percona code.

sql/share/errmsg-utf8.txt
inikep: Errors in 8.0 are much more complicated:
```
# 5.6 error messages
start-error-number 1000

# Error numbers 2000 to 2999 are reserved for the client libraries
reserved-error-section 2000 2999

# 5.7 error messages
start-error-number 3000

# 8.0 error messages
start-error-number 3500

# Error numbers 5000 to 5999 are reserved for X Plugin. 
reserved-error-section 5000 5999

# Start of 8.0 error messages intended to be logged to the server error log.
start-error-number 10000

# Start of Percona Server 8.0 server error log messages
start-error-number 48000

# Start of Percona Server 8.0 error messages to be sent to client
start-error-number 49000

# Start of MyRocks specific messages to be sent to client
start-error-number 50000
```

inikep: Errors in 5.7 are simple:
```
# 5.6 error messages
start-error-number 1000

# 5.7 error messages
start-error-number 3000

# Start of Percona Server 5.7 error messages
start-error-number 11000

# Start of MyRocks specific messages
start-error-number 12000
```

plugin/audit_log/tests/mtr/audit_log_default_db.test
```diff
@@ -31,6 +31,10 @@ GRANT ALL PRIVILEGES ON ąžąžąžą.* TO 'user2'@'%';
 
 # truncate audit log
 UNINSTALL PLUGIN audit_log;
+# disconnect required after https://github.com/mysql/mysql-server/commit/6a893b8b
+--source include/disconnect_connections.inc
+set names utf8;
```
percona-ysorokin, line 36: I believe this is necessary because original `default` connection before disconnect also used to have `set names utf8;`, right?

scripts/comp_sql.c
```diff
@@ -107,6 +107,18 @@ static void print_query(FILE *out, const char *query)
       column++;
       break;
     default:
+      /* Fix clang-8 warning -Winvalid-source-encoding.
```
percona-ysorokin: Please, report this to both Percona Jira and upstream as well.

plugin/audit_log/tests/mtr/audit_log_default_db.test
inikep: You are right

scripts/comp_sql.c
inikep: Reported at:
https://jira.percona.com/browse/PS-5836
https://bugs.mysql.com/bug.php?id=96449

percona-ysorokin:
Jenkins Build link:
https://ps57.cd.percona.com/job/percona-server-5.7-param/133/

Jenkins Build link (ASan):
https://ps57.cd.percona.com/job/percona-server-5.7-asan-param/3/

percona-ysorokin (approved):
LGTM

libbinlogevents/include/table_id.h
laurynas-biveinis, line 42: Why not a ctor initializer list?

sql/field.h
```diff
@@ -566,10 +566,19 @@ class Proto_field
   virtual bool send_text(Protocol *protocol)= 0;
 };
 
-class Field: public Proto_field
+/* auxiliary class to solve the issue with gcc-9 warning -Wdeprecated-copy */
+class nonassignable
```
laurynas-biveinis, line 570: Our pre-C++11 codebase already uses boost::noncopyable for this purpose. Perhaps we should replace the Boost use with this, but in any case let's stay consistent first

storage/innobase/include/trx0trx.h
```diff
@@ -1034,7 +1031,7 @@ class trx_stats {
 	private:
 	// FIXME: only used for reclock-waiting threads, thus could be moved to
 	// e.g. que_thr_t if there's benefit
-	ib_uint64_t lock_que_wait_ustarted;
+	ib_time_monotonic_ms_t lock_que_wait_ustarted;
```
laurynas-biveinis, line 1034: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/buf/buf0buf.cc
```diff
@@ -3790,7 +3790,7 @@ buf_page_get_zip(
 		/* Let us wait until the read operation
 		completes */
 
-		const ib_uint64_t start_time =
+		const ib_time_monotonic_ms_t start_time =
```
laurynas-biveinis, line 3793: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/buf/buf0buf.cc
```diff
@@ -4068,7 +4068,7 @@ buf_wait_for_read(
 
 		/* Wait until the read operation completes */
 
-		const ib_uint64_t start_time =
+		const ib_time_monotonic_ms_t start_time =
```
laurynas-biveinis, line 4071: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/include/trx0trx.h
```diff
@@ -924,7 +924,7 @@ class trx_stats {
 	@return value to be passed to end_io_read
 	*/
 	MY_NODISCARD
-	static ib_uint64_t
+	static ib_time_monotonic_ms_t
```
laurynas-biveinis, line 927: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/include/trx0trx.h
```diff
@@ -936,7 +936,7 @@ class trx_stats {
...
```
laurynas-biveinis, line 939: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/include/trx0trx.h
```diff
@@ -946,15 +946,15 @@ class trx_stats {
 	@param	start_time	return value of start_io_read
 	*/
 	static void
-	end_io_read(trx_t *trx, ib_uint64_t start_time);
+	end_io_read(trx_t *trx, ib_time_monotonic_ms_t start_time);
```
laurynas-biveinis, line 949: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/include/trx0trx.h
```diff
@@ -946,15 +946,15 @@ class trx_stats {
...
 
 	/**
 	Register end of an I/O read or wait.
 
 	@param	trx		transaction to account I/O to
 	@param	start_time	return value of start_io_read*/
 	static void
-	end_io_read(const trx_t &trx, ib_uint64_t start_time);
+	end_io_read(const trx_t &trx, ib_time_monotonic_ms_t start_time);
```
laurynas-biveinis, line 957: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/include/trx0trx.h
```diff
@@ -1445,19 +1442,17 @@ trx_is_started(
 	       && trx->state != TRX_STATE_FORCED_ROLLBACK);
 }
 
-inline ib_uint64_t
+inline ib_time_monotonic_ms_t
```
laurynas-biveinis, line 1445: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/include/trx0trx.h
```diff
@@ -1445,19 +1442,17 @@ trx_is_started(
...
 				       MYSQL_TRX_STAT_IO_READ_BYTES, bytes);
 	}
-	ulint sec, ms;
-	ut_usectime(&sec, &ms);
-	return static_cast<ib_uint64_t>(sec * 1000000 + ms);
+	return ut_time_monotonic_us();
 }
 
-inline ib_uint64_t
+inline ib_time_monotonic_ms_t
```
laurynas-biveinis, line 1455: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/include/trx0trx.h
```diff
@@ -1466,19 +1461,17 @@ trx_stats::start_io_read(trx_t *trx, ulint bytes)
 }
 
 inline void
-trx_stats::end_io_read(const trx_t& trx, ib_uint64_t start_time)
+trx_stats::end_io_read(const trx_t& trx, ib_time_monotonic_ms_t start_time)
```
laurynas-biveinis, line 1464: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/include/trx0trx.h
```diff
@@ -1466,19 +1461,17 @@ trx_stats::start_io_read(trx_t *trx, ulint bytes)
...
 {
-	ulint sec, ms;
-	ut_usectime(&sec, &ms);
-	const ib_uint64_t finish_time
-		= static_cast<ib_uint64_t>(sec * 1000000 + ms);
+	const ib_time_monotonic_ms_t finish_time
```
laurynas-biveinis, line 1466: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/include/trx0trx.h
```diff
@@ -1466,19 +1461,17 @@ trx_stats::start_io_read(trx_t *trx, ulint bytes)
...
+		= ut_time_monotonic_us();
 	thd_report_innodb_stat(trx.mysql_thd, trx.id,
 			       MYSQL_TRX_STAT_IO_READ_WAIT_USECS,
 			       finish_time - start_time);
 }
 
 inline void
-trx_stats::end_io_read(trx_t *trx, ib_uint64_t start_time)
+trx_stats::end_io_read(trx_t *trx, ib_time_monotonic_ms_t start_time)
```
laurynas-biveinis, line 1474: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/include/trx0trx.h
```diff
@@ -1512,9 +1505,7 @@ trx_stats::stop_lock_wait(const trx_t& trx)
 {
 	if (UNIV_LIKELY(!take_stats))
 		return;
-	ulint sec, ms;
-	ut_usectime(&sec, &ms);
-	const ib_uint64_t now = static_cast<ib_uint64_t>(sec * 1000000 + ms);
+	const ib_time_monotonic_ms_t now = ut_time_monotonic_us();
```
laurynas-biveinis, line 1508: Type mismatch: ib_time_monotonic_us_t should be used here

storage/innobase/os/os0file.cc
```diff
@@ -6168,7 +6160,7 @@ os_file_pread(
 {
 	++os_n_file_reads;
 
-	const ib_uint64_t start_time = trx_stats::start_io_read(trx, n);
+	const ib_time_monotonic_ms_t start_time = trx_stats::start_io_read(trx, n);
```
laurynas-biveinis, line 6163: Type mismatch: ib_time_monotonic_us_t should be used here

storage/rocksdb/CMakeLists.txt
```diff
@@ -35,6 +35,12 @@ IF ((CMAKE_CXX_COMPILER_ID STREQUAL GNU) AND
   RETURN ()
 ENDIF ()
 
+# Do not treat warnings as errors for gcc-8+ or clang-8+
```
laurynas-biveinis, line 38: This still results in noisy build log - why not disable selected warnings instead?

laurynas-biveinis (changes requested):

sql/field.h
percona-ysorokin: @laurynas-biveinis This was a tricky one. We originally tried `boost::noncopyable` but it turned out that some derived classes need copy constructor for the `clone()` method. Moreover, in 8.0 source tree there is a comment
```
class Field : public Proto_field {
 public:
  /*
    Field(const Item &) = delete;
    The original intention was seemingly for Field to be non-copyable,
    but due to a typo, this was never enforced, and now there's lots of
    code that copies Field objects around. Thus, the default copy
    constructor needs to stay (assignment is blocked off), but it's probably
    better not to write code that depends on it.
   */
  Field(const Field &) = default;
  void operator=(Field &) = delete;
```
So, this `nonassignable` class is the smartest thing we managed to invent to suppress the warning - it has trivial public copy constructor but hidden private undefined assignment operator.

include/m_string.h
```diff
@@ -109,6 +109,31 @@ static inline char *my_stpncpy(char *dst, const char *src, size_t n)
...
+
+   @param dst   Destination
+   @param src   Source
+   @param n     Maximum number of characters to copy.
+
+   @return pointer to Destination is returned.
+*/
+static inline char *my_strncpy_trunc(char *dst, const char *src, size_t num)
+{
+  size_t len = strlen(src);
```
laurynas-biveinis, line 126: len=

sql/field.h
laurynas-biveinis: Oof. Please add a comment explaining "non-assignable BUT COPYABLE"

laurynas-biveinis (changes requested):

PR #3355, opened 2019-08-09
PS-5835 : Sig 6 Assertion failure: fsp0fsp.cc:914:space_id == page_ge…

satya-bodapati:
…t_space_id(buf_block_get_frame(block))

Problem:
--------
When ALTER INSTANCE ROTATE INNODB MASTER KEY is rotating keys of
encrypted session temporary tablespaces, session disconnect are
concurrently truncating the IBT files. This causes the assert

Fix:
----
Add synchronization mechanism (tablespace mutex) to handle ALTER INSTANCE ROTATE MASTER KEY
and session temp tablespace truncate.

Also encrypted session temporary tablespaces are now rotated by
iterating all active tablespaces from IBT pool.


mysql-test/suite/innodb/r/percona_session_temp_tablespaces_encrypt.result
```diff
@@ -93,6 +93,7 @@ SELECT * FROM INFORMATION_SCHEMA.INNODB_SESSION_TEMP_TABLESPACES WHERE ID = conn
 ID	SPACE	PATH	SIZE	STATE	PURPOSE
 ID	4294501260	./#innodb_temp/temp_4.ibt	98304	ACTIVE	INTRINSIC ENCRYTPED
 ID	4294501261	./#innodb_temp/temp_5.ibt	147456	ACTIVE	USER ENCRYPTED
+ALTER INSTANCE ROTATE INNODB MASTER KEY;
```
robgolebiowski, line 96: This test does not test the crash situation. Some Debug sync between one connection issuing ALTER INSTANCE ROTATE and the other performing SESSION disconnection or temp table truncation would be needed.

robgolebiowski (changes requested):
As discussed on slack to speed the conversation up:
I would prefer using existing mechanism for blocking new operation on tablespace being truncated - and the other way around - blocking truncation when there are operations running on tablespace. Satya prefers new mutex, which is cleaner, but duplicates already existing solution.

Full conversation:
Robert Golebiowski 2:17 PM
@satya shouldn't you try acquiring mutex on shard, get space, check its stop_new_ops, if false increase space's n_pending_ops and continue with rotation. This will block truncation on it. Instead of introducing new mutex ?
Satya Bodapati 2:18 PM
@robert.golebiowski no, there is no point in allowing key rotation on tabelspace being truncated. See the comment for undo tablespace truncation
/* Take MDL on UNDO tablespace to make it mutually exclusive with
   UNDO tablespace truncation. For other tablespaces MDL is not required
   here. */
   MDL_ticket *mdl_ticket = nullptr;
   if (fsp_is_undo_tablespace(space->id)) {
     THD *thd = current_thd;
infact I tried to used MDLs for session temp tablespaces too
there is problem with that. session temp truncation can happen very very early
at THD destruction from bootstrap threads
above code is pasted from Fil_system::encryption_rotate_in_a_shard()
Robert Golebiowski 2:22 PM
It will not allow key rotation on tablespace being truncated. The behavior is the same as with new mutex. If key rotation starts (acquires the new mutex) before truncation. Truncation will have to wait for key rotation to complete. This is the same scenario as with stop_new_ops. When stop_new_ops is set key rotation will not be performed on space.
Satya Bodapati 2:23 PM
I don't see using stop_new_ops is better approach.. any positives?
Robert Golebiowski 2:24 PM
no new mutex - truncating is checking for n_peding_ops to be 0. This is how currently the synchronization with truncate is done.
Satya Bodapati 2:25 PM
encrypt threads is doing that?
the best approach is to take MDLs
Robert Golebiowski 2:25 PM
yes, but not only
Satya Bodapati 2:25 PM
at high level view
infact undo truncation is also done similar in background
and it relies on MDLs
as you can see above
using IO stat variable is mainly for background operations which couldn't take locks/MDLs
came for 5.7 era
8.0 generally avoids this
or should
Robert Golebiowski 2:29 PM
fil_space_acquire_low in 8.0 is doing n_pending_ops++ - to disallow truncation. The mechanism is already there - I do not see a reason to introduce a new one.
Satya Bodapati 2:29 PM
IMHO, this is cleaner approach to handle the session temp tablespaces. There are other issues too
let me explain
Robert Golebiowski 2:30 PM
what's more if stop_new_ops is set fil_space_acquire_low will return null
Satya Bodapati 2:30 PM
rotation can start working on session temp tablespace immediately as long as in-mem encryption flag is set
this truncate is equivalent to DROP.. session temp tablespace is truncated, freed and returned to pool and used for other session (edited) 
and the fil_space_acquire_low() also has comment which I wrote
/** Acquire a tablespace when it could be dropped concurrently.
Used by background threads that do not necessarily hold proper locks
for concurrency control.
there are no background threads
not*
above comment is from fil_space_acquire_low() (edited) 
Robert Golebiowski 2:37 PM
I am seeing it used all over the place, not only by background threads
fil_space_acquire
Satya Bodapati 2:38 PM
who all sets stop_new_ops is the question
Robert Golebiowski 2:38 PM
only truncate
Satya Bodapati 2:38 PM
fil_space_acquire can be freely used just to get fil_space_t with ref_count increase
which truncate?
because truncate is 8.0 is drop + create
Robert Golebiowski 2:39 PM
and delete
Satya Bodapati 2:40 PM
like I said, these are mechanisms from 5.7 era, to handle  drop table from background threads (there was one drop table queue threads)
not sure if that still exists
also there is no perf issue with new sync mechanism too
generally session temp tabelspaces are private to each THD
nobody sees them except two guys
1. I_S
2. ALTER INSTANCE
Robert Golebiowski 2:44 PM
So what's the problem with using MDL locks ?
Satya Bodapati 2:46 PM
I think I explained somewhere above
"infact I tried to used MDLs for session temp tablespaces too
[5:50 PM]
there is problem with that. session temp truncation can happen very very early
[5:50 PM]
at THD destruction from bootstrap threads"
acquire MDLs  at this stage is problematic (edited) 
Robert Golebiowski 2:48 PM
must have overlooked.
Satya Bodapati 2:48 PM
ok, np
190809 11:55:45 [ 33%] innodb.innodb_scrub_background           w8  [ fail ] with 2019-08-09T11:55:41.989811Z 0 [ERROR] [MY-000067] [Server] unknown variable 'innodb-encrypt-tables=KEYRING_ON'.
Robert Golebiowski 2:51 PM
I do not see why we introduce new mutex, if there is already a solution to this. But what can I do
Satya Bodapati 2:52 PM
and that existing solution was not meant to be generic
for all uses cases
and not the right one too
even purge threads started using THD + MDLs
also there is no performance issue
Robert Golebiowski 2:53 PM
but we are not using MDLs :slightly_smiling_face:
Satya Bodapati 2:54 PM
yes, an alternate is used
Robert Golebiowski 2:54 PM
3rd mechanism :slightly_smiling_face:
Satya Bodapati 2:56 PM
yes and nothing wrong at all :wink:

PR #3356, opened 2019-08-11
GPL Authentication LDAP Simple Plugin

fmbiete:
GPL implementation of Oracle MySQL Enterprise Authentication LDAP Simple Plugin


plugin/auth_ldap/.gitignore
```diff
@@ -0,0 +1 @@
+/DEV_NOTES
```
dutow: Why is this folder ignored?

plugin/auth_ldap/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+IF(WITH_LDAP)
+  INCLUDE (CheckLibraryExists)
+  CHECK_LIBRARY_EXISTS(ldap ldap_initialize "" HAVE_LDAP)
+
+  IF(NOT HAVE_LDAP)
+  	MESSAGE(FATAL_ERROR "Required LDAP dev library not found. Please install LDAP development files!")
+  ENDIF(NOT HAVE_LDAP)
+
+  IF(HAVE_LDAP)
+#    SET(ALP_SOURCES_SASL
```
dutow: Please don't leave commented out code (at multiple places)

plugin/auth_ldap/include/auth_ldap_connection.h
```diff
@@ -0,0 +1,72 @@
...
+                     std::string bind_dn, std::string bind_pwd);
+  ~AuthLDAPConnection();
+  bool bind(std::string bind_dn, std::string bind_pwd,
+            bool initial_bind = false);
+  inline AuthLDAPConnection *borrow() {
+    this->borrowed = true;
+    this->borrowed_ts = std::time(nullptr);
+    return this;
+  };
+  inline unsigned int borrowed_for_secs() {
```
dutow: Please use `const` where a function is const. Here also assert that it's borrowed.

plugin/auth_ldap/include/auth_ldap_connection.h
```diff
@@ -0,0 +1,72 @@
...
+namespace auth_ldap {
+class AuthLDAPConnection {
+ public:
+  AuthLDAPConnection(std::string server_host, unsigned int server_port,
+                     bool ssl, bool tls, std::string ca_path, bool initial_bind,
+                     std::string bind_dn, std::string bind_pwd);
+  ~AuthLDAPConnection();
+  bool bind(std::string bind_dn, std::string bind_pwd,
+            bool initial_bind = false);
+  inline AuthLDAPConnection *borrow() {
```
dutow: inline unnecessary (at many places), please remove them.
Also should be noexcept (at many places, where it is)
And please assert that it's not borrowed yet.

plugin/auth_ldap/include/auth_ldap_connection.h
```diff
@@ -0,0 +1,72 @@
...
+    this->borrowed = true;
+    this->borrowed_ts = std::time(nullptr);
+    return this;
+  };
+  inline unsigned int borrowed_for_secs() {
+    return std::time(nullptr) - this->borrowed_ts;
+  };
+  bool is_alive();
+  inline bool is_borrowed() { return this->borrowed; }
+  inline int get_error() { return this->error; }
```
dutow: This function is never called

plugin/auth_ldap/include/auth_ldap_connection.h
```diff
@@ -0,0 +1,72 @@
...
+  std::string get_ldap_uri(std::string server_host, unsigned int server_port,
+                           bool ssl);
+  bool initiate(std::string server_host, unsigned int server_port, bool ssl,
+                bool tls, std::string ca_path);
+
+ private:
+  bool borrowed;
+  std::time_t borrowed_ts;
+  std::time_t created_ts;
+  int error;
```
dutow: This variable is never set

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+std::string AuthLDAPConnection::get_ldap_uri(std::string server_host,
+                                             unsigned int server_port,
+                                             bool ssl) {
+  std::stringstream str_stream;
+  str_stream << (ssl ? "ldaps://" : "ldap://") << server_host << ":"
+             << server_port;
+  return str_stream.str();
+}
+
+bool AuthLDAPConnection::initiate(std::string server_host,
```
dutow: This is only called from the constructor, it would be better to make it part of it.

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+                                  unsigned int server_port, bool ssl, bool tls,
+                                  std::string ca_path) {
+  this->created_ts = std::time(nullptr);
+  this->borrowed = false;
```
dutow: These should be set in the member initializer list

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+
+  if (server_host.empty()) {
+    // log_srv_error("ERROR: server_host is empty");
+    return false;
```
dutow: These return codes are ignored in the constructor, and not all places log errors.

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+    err = ldap_set_option(this->ldap, LDAP_OPT_X_TLS_REQUIRE_CERT, &reqCert);
+    if (err != LDAP_OPT_SUCCESS) {
+      std::stringstream log_stream;
+      log_stream << "ERROR: ldap_set_option(LDAP_OPT_X_TLS_REQUIRE_CERT) "
+                 << ldap_err2string(err);
+      log_srv_error(log_stream.str());
+      return false;
+    }
+  } else {
+    char *cca_path = strdup(ca_path.c_str());
```
dutow: Why `strdup` is necessary?

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+    err = ldap_set_option(this->ldap, LDAP_OPT_X_TLS_CACERTFILE,
+                          static_cast<void *>(cca_path));
+    delete cca_path;
```
dutow: `free(cca_path)`, this was allocated by `strdup`

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+  log_stream << "search_dn(" << base_dn << ", " << filter << ")";
+  log_srv_dbg(log_stream.str());
+  log_stream.str("");
+
+  LDAPMessage *res;
+  char *attrs[] = {strdup("dn"), nullptr};
+  struct timeval search_timeout;
+  memset(&search_timeout, 0, sizeof(struct timeval));
+  search_timeout.tv_sec = 5;
+  int searchlimit = 1;
```
dutow: Please use `const` (at other places too, when possible)

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+      &search_timeout /*timeout*/, searchlimit /*searchlimit*/,
+      &res /*ldapmessage*/);
+  if (err == LDAP_SUCCESS) {
+    // Verify an entry was found
+    if (ldap_count_entries(this->ldap, res) == 0) {
+      log_stream << "ldap_search_ext_s(" << base_dn << ", " << filter
+                 << ") returned no matching entries";
+      log_srv_warn(log_stream.str());
+      log_stream.str("");
+      // Only free up res if there are no items
```
dutow: When is `res` freed in the other case?

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+          vals = ldap_get_values_len(this->ldap, entry, attribute);
+          for (int pos = 0; pos < ldap_count_values_len(vals); pos++) {
+            list.push_back(std::string(vals[pos]->bv_val));
+          }
+          ldap_value_free_len(vals);
+          attribute = ldap_next_attribute(this->ldap, entry, ber);
+        }
+        ber_free(ber, 0);
+        ldap_memfree(entry);
+        entry = ldap_next_entry(this->ldap, entry);
```
dutow: Shouldn't attribute be freed too?

plugin/auth_ldap/include/auth_ldap_connection.h
```diff
@@ -0,0 +1,72 @@
...
+#include <ctime>
+#include <list>
+#include <string>
+
+#include <ldap.h>
+
+namespace mysql {
+namespace plugin {
+namespace auth_ldap {
+class AuthLDAPConnection {
```
dutow: This type should be noncopyable? (possibly others too)

plugin/auth_ldap/src/auth_ldap_connection_pool.cc
```diff
@@ -0,0 +1,169 @@
...
+}
+
+void AuthLDAPConnectionPool::reconfigure(
+    unsigned int initsize, unsigned int maxsize, std::string server_host,
+    unsigned int server_port, bool ssl, bool tls, std::string bind_dn,
+    std::string bind_pwd, std::string ca_path) {
+  if (this->server_port != server_port || this->ca_path != ca_path ||
+      this->server_host != server_host || this->ssl != ssl ||
+      this->tls != tls || this->bind_dn != bind_dn ||
+      this->bind_pwd != bind_pwd) {
```
dutow: Some of these values are uninitialized in the constructor, which will result in sanitizer/valgrind warnings. Please initialize everything in the constructor.

plugin/auth_ldap/src/auth_ldap_connection_pool.cc
```diff
@@ -0,0 +1,169 @@
...
+             << server_port << "]";
+  log_srv_dbg(log_stream.str());
+  log_stream << "ssl [" << ssl << "] tls [" << tls << "] ca_path [" << ca_path
+             << "]";
+  log_srv_dbg(log_stream.str());
+  log_stream << "bind_dn [" << bind_dn << "]";
+  log_srv_dbg(log_stream.str());
+}
+
+void AuthLDAPConnectionPool::reconfigure(
```
dutow: This function/class is not thread safe. Based on other parts of the plugin, it's possible that:

* somebody tries to log on while reconfigure is running
* settings are changed from two different sessions, resulting in two concurrent reconfigure runs

Please use mutexes for everything affected by this.

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+
+  log_stream << "search_dn(" << base_dn << ", " << filter << ")";
+  log_srv_dbg(log_stream.str());
+  log_stream.str("");
+
+  LDAPMessage *res;
+  char *attrs[] = {strdup("dn"), nullptr};
+  struct timeval search_timeout;
+  memset(&search_timeout, 0, sizeof(struct timeval));
+  search_timeout.tv_sec = 5;
```
dutow: Please use a named constant, especially since this is used at multiple places.

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+  std::string str;
+  std::stringstream log_stream;
+  std::string filter = user_search_attr + "=" + user_name;
+
+  log_stream << "search_dn(" << base_dn << ", " << filter << ")";
+  log_srv_dbg(log_stream.str());
+  log_stream.str("");
+
+  LDAPMessage *res;
+  char *attrs[] = {strdup("dn"), nullptr};
```
dutow: When is this duped `dn` freed?

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+  std::stringstream log_stream;
+  std::string filter = std::regex_replace(group_search_filter,
+                                          std::regex("\\{UA\\}"), user_name);
+  filter = std::regex_replace(filter, std::regex("\\{UD\\}"), user_dn);
+  log_stream << "search_groups() - filter: " << filter;
+  log_srv_dbg(log_stream.str());
+  log_stream.str("");
+
+  LDAPMessage *res;
+  char *attrs[] = {strdup(group_search_attr.c_str()), nullptr};
```
dutow: When is this duped string freed?

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+  struct timeval search_timeout;
+  memset(&search_timeout, 0, sizeof(struct timeval));
+  search_timeout.tv_sec = 5;
```
dutow: Please use a named constant.

plugin/auth_ldap/src/auth_ldap_connection_pool.cc
```diff
@@ -0,0 +1,169 @@
...
+    this->server_port = server_port;
+    this->server_host = server_host;
+    this->ssl = ssl;
+    this->tls = tls;
+    this->bind_dn = bind_dn;
+    this->bind_pwd = bind_pwd;
+    this->ca_path = ca_path;
+
+    // Create a new pool of connections
+    for (unsigned int i = 0; i < initsize; i++) {
```
dutow: This loop (and the one below it) will continue (and block) even if it's clear that the server didn't respond to the first request. If we get a timeout, this will result in a quite long wait time before we return from here.

Also, when a user changes the server, it is likely that multiple parameters (hosts, ports, etc) will be changed together. With the current implementation, this will potentially result in multiple long connection trials, which is far from optimal.

I think it would be better to:

(1) stop at the first error, so that the error log isn't flooded, and the function returns sooner
(2) defer actual pool creation to later, instead of for every single change

plugin/auth_ldap/src/auth_ldap_connection_pool.cc
```diff
@@ -0,0 +1,169 @@
...
+    house_keeping_pool(maxsize);
+  }
+  this->initsize = initsize;
+  this->maxsize = maxsize;
+}
+
+AuthLDAPConnection *AuthLDAPConnectionPool::get_connection() {
+  AuthLDAPConnection *obj = nullptr;
+  for (AuthLDAPConnection *conn : this->list) {
+    if (obj == nullptr) {
```
dutow: Shouldn't the loop break when a connection was already found?

plugin/auth_ldap/src/auth_ldap_connection_pool.cc
```diff
@@ -0,0 +1,169 @@
...
+      if (conn->is_borrowed()) {
+        // Checking for zombie connections
+        if (destroy_if_expired(conn)) {
```
dutow: There is another possible threading issue here:
* the lookups in the connection class specify 5 seconds as the timeout, plus some additional time needed for processing
* this zombie check also uses 5 seconds, without additional margins

So it's possible that one thread is still running a lookup with a slow responding server, and another will destroy the object it uses, resulting in a crash.

Adding a safety margin around this would probably fix the issue, but it would be better if "borrowing" would work with a RAII helper class and a shared_ptr. (this function would return a BorrowedLdapConnection, which would store a shared_ptr to the actual ldap connection. BorrowedLdapConnection could borrow in its constructor, and unborrow in the destructor.)

plugin/auth_ldap/include/auth_ldap_connection_pool.h
```diff
@@ -0,0 +1,67 @@
...
+  void house_keeping_pool(unsigned int maxsize);
+
+ private:
+  const unsigned int lost_connection_secs = 5;
+
+  std::string bind_dn;
+  std::string bind_pwd;
+  std::string ca_path;
+  unsigned int initsize;
+  std::list<AuthLDAPConnection *> list;
```
dutow: Consider using `std::list<std::shared_ptr<AuthLDAPConnection>>` (see my comment below about borrowing)

plugin/auth_ldap/include/auth_ldap_connection.h
```diff
@@ -0,0 +1,72 @@
...
+    this->borrowed_ts = std::time(nullptr);
+    return this;
+  };
+  inline unsigned int borrowed_for_secs() {
+    return std::time(nullptr) - this->borrowed_ts;
+  };
+  bool is_alive();
+  inline bool is_borrowed() { return this->borrowed; }
+  inline int get_error() { return this->error; }
+  std::string search_dn(std::string user_name, std::string user_search_attr,
```
dutow: Please use `std::string const&` for parameters, at other places too

plugin/auth_ldap/include/auth_ldap_connection_pool.h
```diff
@@ -0,0 +1,67 @@
...
+class AuthLDAPConnectionPool {
+ public:
+  AuthLDAPConnectionPool(unsigned int initsize, unsigned int maxsize,
+                         std::string server_host, unsigned int server_port,
+                         bool ssl, bool tls, std::string bind_dn,
+                         std::string bind_pwd, std::string ca_path);
+  ~AuthLDAPConnectionPool();
+  void debug_info();
+  AuthLDAPConnection *get_connection();
+  AuthLDAPConnection *new_connection(bool initial_bind);
```
dutow: I don't think this function should be public (see later comments). Also consider returning a shared_ptr (using make_shared)

plugin/auth_ldap/include/auth_ldap_connection_pool.h
```diff
@@ -0,0 +1,67 @@
...
+namespace auth_ldap {
+class AuthLDAPConnectionPool {
+ public:
+  AuthLDAPConnectionPool(unsigned int initsize, unsigned int maxsize,
+                         std::string server_host, unsigned int server_port,
+                         bool ssl, bool tls, std::string bind_dn,
+                         std::string bind_pwd, std::string ca_path);
+  ~AuthLDAPConnectionPool();
+  void debug_info();
+  AuthLDAPConnection *get_connection();
```
dutow: Consider returning `std::shared_ptr<AuthLdapConnection>`, see later comments about borrowing.

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -0,0 +1,245 @@
...
+  this->pool = pool;
+}
+
+AuthLDAPImpl::~AuthLDAPImpl() {
+  this->mappings.clear();
+  this->pool = nullptr;
+}
+
+bool AuthLDAPImpl::bind(const std::string user_dn, const std::string password) {
+  log_srv_dbg("AuthLDAPImpl::bind()");
```
dutow, line 53: Please use `DBUG_ENTER` / related macros instead (at other places too)

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -0,0 +1,245 @@
...
+  }
+  log_srv_dbg("MySQL mapping not found for existing LDAP groups");
+  return "";
+}
+
+AuthLDAPConnection *AuthLDAPImpl::_get_ldap_connection(bool is_from_pool) {
+  log_srv_dbg("AuthLDAPImpl::_get_ldap_connection()");
+  AuthLDAPConnection *conn = nullptr;
+
+  if (is_from_pool && this->pool->max_size() > 0) {
```
dutow: This condition is a possible memory leak: if a caller specifies is_from_pool, it won't ever delete the connection.

Please consider using shared_ptr.

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -0,0 +1,245 @@
...
+    log_srv_dbg("Get connection from pool");
+    conn = pool->get_connection();
+    if (!conn) {
+      log_srv_error("There are no connections available in the LDAP pool");
+      pool->debug_info();
+    }
+  } else {
+    conn = pool->new_connection(is_from_pool /*initial_bind = is_from_pool*/);
```
dutow: Why is a non pooled connection required for bind?

plugin/auth_ldap/src/plugin_common.cc
```diff
@@ -0,0 +1,120 @@
...
+  if (!impl->get_ldap_uid(&user_dn)) {
+    delete impl;
+    log_stream << "LDAP user DN not found for [" << STR_NULL(info->user_name)
+               << "]";
+    log_srv_warn(log_stream.str());
+    return CR_AUTH_USER_CREDENTIALS;
+  }
+
+  // Authenticate agains ldap
+  if (!impl->bind(user_dn, STR_NULL(password))) {
```
dutow: bind is used only here, and_get_ldap_connection at this point, the pool is already set up.

plugin/auth_ldap/tests/mtr/auth_ldap_simple.t
```diff
@@ -0,0 +1,10 @@
+INSTALL PLUGIN authentication_ldap_simple SONAME 'authentication_ldap_simple.so';
+SELECT PLUGIN_NAME, PLUGIN_STATUS FROM INFORMATION_SCHEMA.PLUGINS WHERE PLUGIN_NAME LIKE 'authentication_ldap_simple%';
+SHOW GLOBAL VARIABLES LIKE 'authentication_ldap_simple%';
+SET GLOBAL authentication_ldap_simple_bind_base_dn = 'cn=users,cn=accounts,dc=example,dc=test';
+SET GLOBAL authentication_ldap_simple_server_host = 'freeipa';
+SET GLOBAL authentication_ldap_simple_server_port = 389;
+SHOW GLOBAL VARIABLES LIKE 'authentication_ldap_simple%';
+CREATE USER user1 IDENTIFIED WITH authentication_ldap_simple BY 'uid=user1,cn=users,cn=accounts,dc=example,dc=test';
```
dutow: Please include a test which at least logs on the user.

plugin/auth_ldap/include/auth_ldap_connection_pool.h
```diff
@@ -0,0 +1,67 @@
...
+
+ private:
+  AuthLDAPConnection *create_connection(bool borrow);
+  void destroy(AuthLDAPConnection *con);
+  void destroy_all();
+  bool destroy_if_expired(AuthLDAPConnection *con);
+  void house_keeping_pool(unsigned int maxsize);
+
+ private:
+  const unsigned int lost_connection_secs = 5;
```
dutow: This should be more than 5, as 5 is used as the connection timeout for ldap requests. Also should be static and constexpr.

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,76 @@
...
+#include "plugin/auth_ldap/include/auth_ldap_connection_pool.h"
+
+#include <list>
+#include <string>
+#include <vector>
+
+namespace mysql {
+namespace plugin {
+namespace auth_ldap {
+typedef struct {
```
dutow: C-style definition, please use C++ style.

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,76 @@
...
+ public:
+  AuthLDAPImpl(const std::string user_name, const std::string auth_string,
+               const std::string user_search_attr,
+               const std::string group_search_filter,
+               const std::string group_search_attr,
+               const std::string bind_base_dn, AuthLDAPConnectionPool *pool);
+  ~AuthLDAPImpl();
+
+  bool bind(const std::string user_dn, const std::string password);
+  bool get_ldap_uid(std::string *user_dn);
```
dutow, line 42: Add a comment that user_dn is an out parameter.  Consider non const ref instead, as it can't be null.

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,76 @@
...
+
+  bool get_mysql_uid(std::string *user_mysql, const std::string user_dn);
```
dutow: Add a comment that user_mysql is an out parameter. Consider non const ref instead, as it can't be null. Output parameters also should be after input parameters, consider switching them.

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,76 @@
...
+
+ private:
+  const std::string _calc_ldap_uid();
+
+  void _calc_mappings(const std::string group_str);
+  const std::string _calc_mysql_user(const std::list<std::string> groups);
```
dutow: Please use `const&`, do not copy a list if not necessary. Other places too.

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,76 @@
...
+
+  bool _matched_map(const t_group_mapping map,
+                    const std::list<std::string> groups);
+
+  std::list<std::string> _search_ldap_groups(const std::string user_dn);
+
+  const std::string _search_ldap_uid();
+
+ private:
+  AuthLDAPConnectionPool *pool;
```
dutow: Shouldn't one `AuthLDAPImpl` stick to one connection? Based on later code, using a pool here means that in theory it is possible that:

* some of these functions are completed
* ldap settings are changed, the server connects to a different ldap server
* later function calls here access a different ldap server

plugin/auth_ldap/include/plugin_common.h
```diff
@@ -0,0 +1,51 @@
...
+#include "my_dbug.h"
+#include "my_sys.h"
+#include "mysql/components/my_service.h"
+#include "mysql/components/service_implementation.h"
+#include "mysql/plugin_auth.h"
+#include "mysqld_error.h"
+
+#include "plugin/auth_ldap/include/auth_ldap_connection_pool.h"
+
+#define STR_NULL(x) (x == nullptr ? "" : x)
```
dutow: maybe `STR_OR_EMPTY`? Also, consider making it a simple inline function, the compiler will optimize it the same way

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+
+namespace mysql {
+namespace plugin {
+namespace auth_ldap {
+AuthLDAPConnection::AuthLDAPConnection(std::string server_host,
+                                       unsigned int server_port, bool ssl,
+                                       bool tls, std::string ca_path,
+                                       bool initial_bind, std::string bind_dn,
+                                       std::string bind_pwd) {
+  this->ldap = nullptr;
```
dutow: use initializer list (for other classes too)

plugin/auth_ldap/src/plugin_common.cc
```diff
@@ -0,0 +1,120 @@
...
+    MYSQL_PLUGIN_VIO *vio MY_ATTRIBUTE((unused)), MYSQL_SERVER_AUTH_INFO *info,
+    const char *password,
+    mysql::plugin::auth_ldap::AuthLDAPConnectionPool *pool,
+    const char *user_search_attr, const char *group_search_attr,
+    const char *group_search_filter, const char *bind_base_dn) {
+  std::stringstream log_stream;
+
+  log_srv_dbg("auth_ldap_common_authenticate_user()");
+
+  mysql::plugin::auth_ldap::AuthLDAPImpl *impl =
```
dutow: Please use a unique_ptr

plugin/auth_ldap/src/plugin_simple.cc
```diff
@@ -0,0 +1,138 @@
...
+  MYSQL_PLUGIN_VAR_HEADER;
+};
+
+template <typename Copy_type>
+void update_sysvar(THD *, SYS_VAR *var, void *var_ptr, const void *value) {
+  // Update the value
+  *(Copy_type *)var_ptr = *(Copy_type *)value;
+
+  if (strcmp(var->name, "authentication_ldap_simple_log_status") == 0)
+    g_logger_server->set_log_level(static_cast<ldap_log_level>(log_status));
```
dutow: Please create a specialized function for this instead of the runtime check.

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+      return false;
+    }
+  }
+
+  return true;
+}
+
+bool AuthLDAPConnection::is_alive() {
+  int id;
+  return ldap_whoami(this->ldap, nullptr, nullptr, &id) == LDAP_SUCCESS;
```
dutow: Can ldap_whoami handle NULLs? And similarly for other functions: shouldn't a null ldap connection result in an early exit?

plugin/auth_ldap/src/auth_ldap_connection.cc
```diff
@@ -0,0 +1,288 @@
...
+  struct timeval search_timeout;
+  memset(&search_timeout, 0, sizeof(struct timeval));
+  search_timeout.tv_sec = 5;
+  int err = ldap_search_ext_s(
+      this->ldap, base_dn.c_str() /* base */, LDAP_SCOPE_SUBTREE /*scope*/,
+      filter.c_str() /*filter*/, attrs /*attrs*/, 0 /*attrsonly*/,
+      nullptr /*serverctrls*/, nullptr /*clientctrls*/,
+      &search_timeout /*timeout*/, 0 /*searchlimit*/, &res /*ldapmessage*/);
+  if (err == LDAP_SUCCESS) {
+    // Verify an entry was found
```
dutow: Please use `/**/` for comments

plugin/auth_ldap/include/auth_ldap_connection.h
```diff
@@ -0,0 +1,72 @@
...
+                        std::string base_dn);
+  std::list<std::string> search_groups(std::string user_name,
+                                       std::string bind_user,
+                                       std::string group_search_attr,
+                                       std::string group_search_filter,
+                                       std::string base_dn);
+  inline void unborrow() { this->borrowed = false; };
```
dutow: Assert that it's borrowed.

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -0,0 +1,245 @@
...
+  if (parts.size() == 2) {
+    std::string raw_group_mappings = boost::algorithm::trim_copy(parts[1]);
+    if (!raw_group_mappings.empty()) _calc_mappings(raw_group_mappings);
+  }
+  this->pool = pool;
+}
+
+AuthLDAPImpl::~AuthLDAPImpl() {
+  this->mappings.clear();
+  this->pool = nullptr;
```
dutow: Both unnecessary could use a default destructor.

dutow (changes requested):
Thank you for the pull request!

I've added some comments - I did not review the SASL part yet, but for everything else.

Most of them are about the code style, but a few are questions about how this works in a multithreaded environment, I think there are a few possible issues there.

plugin/auth_ldap/src/auth_ldap_connection_pool.cc
dutow: There's also a thread safety issue here too:
* if one thread finds a zombie, and destroys it (removing from the list)
* while another is also iterating on the same list

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: I'm doing a static cast to void in the ldap call, I need a char* that is not const

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: Keeping it in a function make it easier to cleanup and control the errors without goto or nested ifs

plugin/auth_ldap/src/auth_ldap_impl.cc
fmbiete: A pooled connection is authenticated with a common user/password defined in the global settings. User authentication should overwrite that or use another connection to perform the authentication.

plugin/auth_ldap/src/plugin_common.cc
fmbiete: Yes, this is user authentication. It doesn't use a pre-authenticated connection from the pool

plugin/auth_ldap/tests/mtr/auth_ldap_simple.t
fmbiete: If I include a login test that would only succeed if there is a ldap backend. Is that ok?

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: Yes, it can. Here we are checking that the connection is alive, since that would be the only case where ldap_whoami returns a non-success code

plugin/auth_ldap/.gitignore
fmbiete: Removed

plugin/auth_ldap/CMakeLists.txt
fmbiete: Removed

plugin/auth_ldap/include/auth_ldap_connection.h
fmbiete: Fixed

plugin/auth_ldap/include/auth_ldap_connection.h
fmbiete: Changed

plugin/auth_ldap/include/auth_ldap_connection.h
fmbiete: Removed

plugin/auth_ldap/include/auth_ldap_connection.h
fmbiete: Removed

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: Changed

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: Changed

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: Changed

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: Changed

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: Fixed

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: Fixed

plugin/auth_ldap/include/auth_ldap_connection.h
fmbiete: Added non-copyable

plugin/auth_ldap/src/auth_ldap_connection_pool.cc
fmbiete: Changed

plugin/auth_ldap/src/auth_ldap_connection_pool.cc
fmbiete: Protected with rw_locks

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: it cannot be a constant or the ldap call won't compile

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: Removed

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: Removed

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: It cannot be a constant

plugin/auth_ldap/src/auth_ldap_impl.cc
fmbiete: Oracle auth_ldap plugin supports custom log including debug level. By that reason I have not used DBUG_ENTER or related macros.

plugin/auth_ldap/include/auth_ldap_impl.h
fmbiete: Removed typedef

plugin/auth_ldap/include/plugin_common.h
fmbiete: Chnaged

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: Changed

plugin/auth_ldap/src/plugin_common.cc
fmbiete: Changed

plugin/auth_ldap/src/auth_ldap_connection.cc
fmbiete: ?? I can see multiple // within MySQL. Is this something upstream is trying to change?

dutow:
Thanks for the updates!

Let's focus on the multithreaded part for now: I see that you created `pool_rwlock`, but it only protects the pool destructor and reconfigure now. This should fix the issue when multiple sessions change the pool parameters simultaneously.

However, as I mentioned in some of my previous comments, this isn't the full issue. With this authentication plugin, there are:
* many variables, which possibly can be set from different sessions at the same time
* and many users, who want to log in at the same time, possibly when somebody else is changing the parameter
* there's also the fact that even a simple connection getter could delete zombie collections, modifying the list

Or looking at this from a different viewpoint:
* if a variable is protected by a lock somewhere, there's a good chance it should always be used behind that lock (even for reading)
* Most STL data structures (including lists, vectors, maps) aren't thread-safe. This means that (a) two threads shouldn't modify them at the same time (b) When a thread is modifying them, no other thread should read them.

Right now, there are examples for both issues (a) and (b) in this plugin.

As mysql locks aren't reentrant, this will most likely require some refactoring of AuthLdapConnection

(and I don't think this is fully fixable this without using shared_ptrs, or rethinking how zombie killing works)

plugin/auth_ldap/src/auth_ldap_connection.cc
dutow: Looking at the openldap github sources, there are some asserts about this:

https://github.com/openldap/openldap/blob/b06f5b0493937fc28f2cc86df1d7f464aa4504d8/libraries/libldap/whoami.c#L67

As those are asserts they won't fire in a release build of openldap, but I don't think that means it's supported.

fmbiete:
I refactored the code following your suggestions.

dutow:
@fmbiete Thank you, and apologies for the slow reply! I updated your branch. Please update your local checkout and check my changes.

Here's a summary:

* I modified the simple testcase so it actually contains to an ldap server. For now the testcase uses a simple sample docker container available on github, to make it easily reproducible
* Fixed memory leaks (mostly: objects returned by the ldap functions were leaked. These are all freed now using the correct ldap functions)
* Fixed some minor issues with the ldap api use
* Fixed threading issues (a few calls weren't protected by locks which should have been ; one structure was possibly over indexed, if the pool size was reduced while an user was connecting)

Now I only see one possible (but unlikely) remaining crash: if a user tries to connects the moment the plugin is uninstalled. For now, I've added a todo comment about this in plugin_simple.cc, and trying to figure out a minimal, but safe change for this.

plugin/auth_ldap/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+IF(WITH_LDAP)
+  INCLUDE (CheckLibraryExists)
+  CHECK_LIBRARY_EXISTS(ldap ldap_initialize "" HAVE_LDAP)
+
+  IF(NOT HAVE_LDAP)
+  	MESSAGE(FATAL_ERROR "Required LDAP dev library not found. Please install LDAP development files!")
+  ENDIF(NOT HAVE_LDAP)
+
+  IF(HAVE_LDAP)
```
percona-ysorokin: Probably this `IF` is redundant - `IF(NOT HAVE_LDAP)` already terminates with a fatal error

plugin/auth_ldap/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+      src/plugin_common.cc
+      src/plugin_simple.cc
+      src/auth_ldap_impl.cc
+      src/connection.cc
+      src/pool.cc
+    )
+    ### Configuration ###
+    ADD_DEFINITIONS(-DLOG_COMPONENT_TAG="auth_ldap")
+
+    INCLUDE_DIRECTORIES(SYSTEM /work/openldap/include)
```
percona-ysorokin: Weird system path

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,75 @@
+#ifndef _AUTH_LDAP_IMPL_MPALDAP_H
```
percona-ysorokin, line 1: Identifiers with leading underscores are reserved for c/c++

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include <list>
+#include <string>
+#include <vector>
+
+namespace mysql {
+namespace plugin {
+namespace auth_ldap {
+struct t_group_mapping {
+  std::vector<std::string> ldap_groups;
```
percona-ysorokin, line 28: typedef/using for a custom type

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,75 @@
...
+ public:
+  AuthLDAPImpl(const std::string &user_name, const std::string &auth_string,
+               const std::string &user_search_attr,
+               const std::string &group_search_filter,
+               const std::string &group_search_attr,
+               const std::string &bind_base_dn, Pool *pool);
+  ~AuthLDAPImpl();
+
+  bool bind(const std::string &user_dn, const std::string &password);
+  bool get_ldap_uid(std::string *user_dn);
```
percona-ysorokin, line 42: pass out parameter by reference

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,75 @@
...
+
+  bool get_mysql_uid(std::string *user_mysql, const std::string &user_dn);
+
+ private:
+  const std::string _calc_ldap_uid();
```
percona-ysorokin: `const` return type in c++11 prevents rvo/copy elision.
leading `_`  for identifiers is reserved.

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,75 @@
...
+
+  void _calc_mappings(const std::string &group_str);
```
percona-ysorokin: leading `_`  for identifiers is reserved.

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,75 @@
...
+  const std::string _calc_mysql_user(const std::list<std::string> &groups);
```
percona-ysorokin: const return type in c++11 prevents rvo/copy elision.
leading _ for identifiers is reserved.
typedef/using for custom container `std::list<std::string>`

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,75 @@
...
+
+  bool _matched_map(const t_group_mapping &map,
```
percona-ysorokin: leading _ for identifiers is reserved.
typedef/using for custom container std::list<std::string>

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,75 @@
...
+                    const std::list<std::string> &groups);
+
+  std::list<std::string> _search_ldap_groups(const std::string &user_dn);
+
+  const std::string _search_ldap_uid();
```
percona-ysorokin: const return type in c++11 prevents rvo/copy elision.
leading _ for identifiers is reserved.

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,75 @@
...
+
+ private:
+  Pool *_pool;
```
percona-ysorokin: Here and everywhere
leading _ for identifiers is reserved.

plugin/auth_ldap/include/auth_ldap_impl.h
```diff
@@ -0,0 +1,75 @@
...
+  // Static attributes across all requests
+  std::string _user_search_attr;
+  std::string _group_search_attr;
+  std::string _group_search_filter;
+  std::string _bind_base_dn;
+  // Dynamic attributes per request
+  std::string _user_name;
+  std::string _user_auth_string;
+  std::vector<t_group_mapping> _mappings;
```
percona-ysorokin: typedef/using for custom container

plugin/auth_ldap/include/connection.h
```diff
@@ -0,0 +1,84 @@
+#ifndef _MPAL_CONNECTION_H
```
percona-ysorokin: Leading underscore is reserved

plugin/auth_ldap/include/connection.h
```diff
@@ -0,0 +1,84 @@
...
+#include <string>
+
+#include <ldap.h>
+
+namespace mysql {
+namespace plugin {
+namespace auth_ldap {
+class Connection {
+ public:
+  static const int ZombieAfterSeconds = 120;
```
percona-ysorokin: why signed type?

plugin/auth_ldap/include/connection.h
```diff
@@ -0,0 +1,84 @@
...
+  int get_idx_pool() ;
+  bool is_snipped() ;
+  bool is_zombie() ;
+  void mark_as_busy();
+  void mark_as_free();
+  void mark_as_snipped();
+  std::string search_dn(const std::string &user_name,
+                        const std::string &user_search_attr,
+                        const std::string &base_dn);
+  std::list<std::string> search_groups(const std::string &user_name,
```
percona-ysorokin, line 55: typedef/using for custom container

plugin/auth_ldap/include/log_client.h
```diff
@@ -0,0 +1,112 @@
...
+   GNU General Public License, version 2.0, for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA */
+
+#ifndef MPAL_LOG_H
+#define MPAL_LOG_H
+
+#include <stdio.h>
```
percona-ysorokin:
```suggestion
#include <cstdio>
```

plugin/auth_ldap/include/log_client.h
```diff
@@ -0,0 +1,112 @@
...
+#include <iostream>
+#include <sstream>
+
+namespace mysql {
+namespace plugin {
+namespace auth_ldap {
+struct ldap_log_type {
+  typedef enum {
```
percona-ysorokin: old c-style
typedef enum {} <name> idiom.

plugin/auth_ldap/include/log_client.h
```diff
@@ -0,0 +1,112 @@
...
+  LDAP_LOG_LEVEL_ERROR_WARNING,
+  LDAP_LOG_LEVEL_ERROR_WARNING_INFO,
+  LDAP_LOG_LEVEL_ALL
+};
+
+class Ldap_log_writer_error {
+ public:
+  Ldap_log_writer_error();
+  ~Ldap_log_writer_error();
+  void write(std::string data);
```
percona-ysorokin: pass by const reference

plugin/auth_ldap/include/log_client.h
```diff
@@ -0,0 +1,112 @@
...
+};
+
+class Ldap_logger {
+ public:
+  Ldap_logger();
+  ~Ldap_logger();
+  template <ldap_log_type::ldap_type type>
+  void log(std::string msg);
```
percona-ysorokin: pass by const reference

plugin/auth_ldap/include/log_client.h
```diff
@@ -0,0 +1,112 @@
...
+  void set_log_level(ldap_log_level level);
+
+ private:
+  Ldap_log_writer_error *m_log_writer;
+  ldap_log_level m_log_level;
+};
+
+template <ldap_log_type::ldap_type type>
+void Ldap_logger::log(std::string msg) {
```
percona-ysorokin: pass by const reference

plugin/auth_ldap/include/log_client.h
```diff
@@ -0,0 +1,112 @@
...
+  std::stringstream log_stream;
```
percona-ysorokin:
```suggestion
  std::ostringstream log_stream;
```

plugin/auth_ldap/include/plugin_common.h
```diff
@@ -0,0 +1,50 @@
...
+#include "my_dbug.h"
+#include "my_sys.h"
+#include "mysql/components/my_service.h"
+#include "mysql/components/service_implementation.h"
+#include "mysql/plugin_auth.h"
+#include "mysqld_error.h"
+
+#include "plugin/auth_ldap/include/pool.h"
+
+inline const char *STR_OR_EMPTY(const char *x) { return x == nullptr ? "" : x; }
```
percona-ysorokin: lower case for function names

plugin/auth_ldap/include/plugin_common.h
```diff
@@ -0,0 +1,50 @@
...
+int auth_ldap_common_deinit(mysql::plugin::auth_ldap::Pool *connPool);
+
+int auth_ldap_common_authenticate_user(
+    MYSQL_PLUGIN_VIO *vio, MYSQL_SERVER_AUTH_INFO *info, const char *password,
+    mysql::plugin::auth_ldap::Pool *pool, const char *user_search_attr,
+    const char *group_search_attr, const char *group_search_filter,
+    const char *bind_base_dn);
+
+int auth_ldap_common_generate_auth_string_hash(char *outbuf,
+                                               unsigned int *buflen,
```
percona-ysorokin, line 40: size_t for lengths

plugin/auth_ldap/include/plugin_common.h
```diff
@@ -0,0 +1,50 @@
...
+                                               const char *inbuf,
+                                               unsigned int inbuflen);
```
percona-ysorokin, line 42: size_t for lengths

plugin/auth_ldap/include/plugin_common.h
```diff
@@ -0,0 +1,50 @@
...
+
+int auth_ldap_common_validate_auth_string_hash(char *const buf,
+                                               unsigned int len);
```
percona-ysorokin, line 45: size_t for lengths

plugin/auth_ldap/include/plugin_common.h
```diff
@@ -0,0 +1,50 @@
...
+    mysql::plugin::auth_ldap::Pool *pool, const char *user_search_attr,
+    const char *group_search_attr, const char *group_search_filter,
+    const char *bind_base_dn);
+
+int auth_ldap_common_generate_auth_string_hash(char *outbuf,
+                                               unsigned int *buflen,
+                                               const char *inbuf,
+                                               unsigned int inbuflen);
+
+int auth_ldap_common_validate_auth_string_hash(char *const buf,
```
percona-ysorokin, line 44: is const pointer instead of const data intentional here?

plugin/auth_ldap/include/plugin_common.h
```diff
@@ -0,0 +1,50 @@
...
+                                               unsigned int len);
+
+int auth_ldap_common_set_salt(const char *password, unsigned int password_len,
```
percona-ysorokin, line 47: size_t for lengths

plugin/auth_ldap/include/pool.h
```diff
@@ -0,0 +1,72 @@
...
+#include <vector>
+
+#include "plugin/auth_ldap/include/connection.h"
+
+namespace mysql {
+namespace plugin {
+namespace auth_ldap {
+class Pool {
+ public:
+  Pool(unsigned int pool_initial_size, unsigned int pool_max_size,
```
percona-ysorokin: std::size_t for size types

plugin/auth_ldap/include/pool.h
```diff
@@ -0,0 +1,72 @@
...
+       const std::string &ldap_host, unsigned int ldap_port, bool use_ssl,
+       bool use_tls, const std::string &ca_path, const std::string &bind_dn,
+       const std::string &bind_pwd);
+  ~Pool();
+
+ public:
+  Pool(const Pool &) = delete;             // non construction-copyable
+  Pool &operator=(const Pool &) = delete;  // non copyable
+ public:
+  std::shared_ptr<Connection> borrow_connection(bool default_connect = true);
```
percona-ysorokin: typedef/using for custom smart pointer

plugin/auth_ldap/include/pool.h
```diff
@@ -0,0 +1,72 @@
...
+  void debug_info();
+  void return_connection(std::shared_ptr<Connection> conn);
```
percona-ysorokin: typedef/using for custom smart pointer

plugin/auth_ldap/include/pool.h
```diff
@@ -0,0 +1,72 @@
...
+  void reconfigure(unsigned int new_pool_initial_size,
```
percona-ysorokin: size_t for size types

plugin/auth_ldap/include/pool.h
```diff
@@ -0,0 +1,72 @@
...
+                   unsigned int new_pool_max_size, const std::string &ldap_host,
+                   unsigned int ldap_port, bool use_ssl, bool use_tls,
+                   const std::string &ca_path, const std::string &bind_dn,
+                   const std::string &bind_pwd);
+  void zombie_control();
+
+ private:
+  int _find_first_free();
+  std::shared_ptr<Connection> _get_connection(int idx, bool default_connect);
```
percona-ysorokin: typedef/using for custom smart pointer

plugin/auth_ldap/include/pool.h
```diff
@@ -0,0 +1,72 @@
...
+  void _mark_as_busy(size_t idx);
+  void _mark_as_free(size_t idx);
+
+ private:
+  unsigned int _pool_initial_size;
+  unsigned int _pool_max_size;
+  std::string _ldap_host;
+  unsigned int _ldap_port;
```
percona-ysorokin: Shouldn't port type be 16-bit unsigned like `std::uint16_t`?

plugin/auth_ldap/include/pool.h
```diff
@@ -0,0 +1,72 @@
...
+  bool _use_ssl;
+  bool _use_tls;
+  std::string _ca_path;
+  std::string _bind_dn;
+  std::string _bind_pwd;
+  boost::dynamic_bitset<> _bs_used;
```
percona-ysorokin: typedef/using for custom containers

plugin/auth_ldap/include/pool.h
```diff
@@ -0,0 +1,72 @@
...
+  std::vector<std::shared_ptr<Connection>> _v_connections;
```
percona-ysorokin: typedef/using for custom containers

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -0,0 +1,227 @@
...
+  }
+}
+
+AuthLDAPImpl::~AuthLDAPImpl() {}
+
+bool AuthLDAPImpl::bind(const std::string &user_dn,
+                        const std::string &password) {
+  log_srv_dbg("AuthLDAPImpl::bind()");
+  bool success = false;
+  std::stringstream log_stream;
```
percona-ysorokin: ostringstream

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -0,0 +1,227 @@
...
+  log_srv_dbg("AuthLDAPImpl::get_ldap_uid()");
+
+  if (_user_auth_string.empty()) {
+    *user_dn = _search_ldap_uid();
+  } else {
+    *user_dn = _calc_ldap_uid();
+  }
+
+  if (user_dn->empty()) {
+    std::stringstream log_stream;
```
percona-ysorokin: std::ostringstream

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -0,0 +1,227 @@
...
+
+AuthLDAPImpl::~AuthLDAPImpl() {}
+
+bool AuthLDAPImpl::bind(const std::string &user_dn,
+                        const std::string &password) {
+  log_srv_dbg("AuthLDAPImpl::bind()");
+  bool success = false;
+  std::stringstream log_stream;
+
+  std::shared_ptr<Connection> conn = _pool->borrow_connection(false);
```
percona-ysorokin: auto

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -0,0 +1,227 @@
...
+
+  return !user_dn->empty();
+}
+
+bool AuthLDAPImpl::get_mysql_uid(std::string *user_mysql,
+                                 const std::string &user_dn) {
+  log_srv_dbg("AuthLDAPImpl::get_mysql_uid()");
+  if (user_dn.empty()) return false;
+
+  std::list<std::string> ldap_groups = _search_ldap_groups(user_dn);
```
percona-ysorokin: auto

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -0,0 +1,227 @@
...
+}
+
+/**
+ * All the groups in a map are present in ldap: MATCH!
+ */
+bool AuthLDAPImpl::_matched_map(const t_group_mapping &map,
+                                const std::list<std::string> &groups) {
+  log_srv_dbg("AuthLDAPImpl::_matched_map()");
+  bool matched = true;
+  std::stringstream log_stream;
```
percona-ysorokin: std::ostringstream

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -0,0 +1,227 @@
...
+  log_stream << uid;
+  log_srv_dbg(log_stream.str());
+
+  return uid;
+}
+
+void AuthLDAPImpl::_calc_mappings(const std::string &group_str) {
+  std::vector<std::string> parts2;
+  boost::algorithm::split(parts2, group_str, boost::is_any_of(","));
+  for (std::string const &s : parts2) {
```
percona-ysorokin: be more consistent with `east const` vs `west const`

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -0,0 +1,227 @@
...
+  _pool->return_connection(conn);
+
+  return list;
+}
+
+const std::string AuthLDAPImpl::_search_ldap_uid() {
+  log_srv_dbg("AuthLDAPImpl::_search_ldap_uid()");
+  std::string uid;
+
+  std::shared_ptr<Connection> conn = _pool->borrow_connection();
```
percona-ysorokin: auto

plugin/auth_ldap/src/connection.cc
```diff
@@ -0,0 +1,304 @@
...
+
+    if (_ca_path.size() == 0) {
+      int reqCert = LDAP_OPT_X_TLS_NEVER;
+      err = ldap_set_option(_ldap, LDAP_OPT_X_TLS_REQUIRE_CERT, &reqCert);
+      if (err != LDAP_OPT_SUCCESS) {
+        _log_error("ldap_set_option(LDAP_OPT_X_TLS_REQUIRE_CERT)", err);
+        return false;
+      }
+    } else {
+      char *cca_path = strdup(_ca_path.c_str());
```
percona-ysorokin: Using const_cast to fix incorrect interfaces is OK - no need to duplicate ca_path

plugin/auth_ldap/src/connection.cc
```diff
@@ -0,0 +1,304 @@
...
+      err = ldap_start_tls_s(_ldap, nullptr, nullptr);
+      if (err != LDAP_SUCCESS) {
+        _log_error("ldap_start_tls_s", err);
+        return false;
+      }
+    }
+
+    struct berval *serverCreds;
+    struct berval *userCreds =
+        ber_str2bv(strdup(bind_pwd.c_str()), 0, 0, nullptr);
```
percona-ysorokin, line 113: Is this strdup a memory leak?

plugin/auth_ldap/src/connection.cc
```diff
@@ -0,0 +1,304 @@
...
+      return false;
+    }
+
+    return true;
+  }
+
+  return false;
+}
+
+int Connection::get_idx_pool(){
```
percona-ysorokin: std::size_t for index types
this method should be const

plugin/auth_ldap/src/connection.cc
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  return false;
+}
+
+int Connection::get_idx_pool(){ 
+  return _index; }
+
+bool Connection::is_snipped(){ 
+  std::lock_guard<std::mutex> lock(_conn_mutex);
```
percona-ysorokin: use std::atomic_bool for `snipped` member.
this method should be const

plugin/auth_ldap/src/connection.cc
```diff
@@ -0,0 +1,304 @@
...
+  _snipped = true;
+}
+
+std::string Connection::search_dn(const std::string &user_name,
+                                  const std::string &user_search_attr,
+                                  const std::string &base_dn) {
+  std::lock_guard<std::mutex> lock(_conn_mutex);
+
+  std::string str;
+  std::stringstream log_stream;
```
percona-ysorokin: std::ostringstream

plugin/auth_ldap/src/connection.cc
```diff
@@ -0,0 +1,304 @@
...
+  std::copy(list.begin(), list.end(),
+            std::ostream_iterator<std::string>(log_stream, ","));
+  log_srv_dbg(log_stream.str());
+  log_stream.str("");
+
+  return list;
+}
+
+std::string Connection::_get_ldap_uri() {
+  std::stringstream str_stream;
```
percona-ysorokin: std::ostringstream

plugin/auth_ldap/src/plugin_common.cc
```diff
@@ -0,0 +1,114 @@
...
+
+  return 0;
+}
+
+int auth_ldap_common_authenticate_user(
+    MYSQL_PLUGIN_VIO *vio MY_ATTRIBUTE((unused)), MYSQL_SERVER_AUTH_INFO *info,
+    const char *password, mysql::plugin::auth_ldap::Pool *pool,
+    const char *user_search_attr, const char *group_search_attr,
+    const char *group_search_filter, const char *bind_base_dn) {
+  std::stringstream log_stream;
```
percona-ysorokin, line 40: std::ostringstream

plugin/auth_ldap/src/plugin_common.cc
```diff
@@ -0,0 +1,114 @@
...
+    fail if buffer specified by server cannot be copied to output buffer
+  */
+  if (*buflen < inbuflen) return 1; /* error */
+  strncpy(outbuf, inbuf, inbuflen);
+  *buflen = strnlen(inbuf, inbuflen);
+  return 0; /* success */
+}
+
+int auth_ldap_common_validate_auth_string_hash(char *const buf
+                                               __attribute__((unused)),
```
percona-ysorokin: Use just unnamed parameter instead of `__attribute__((unused))`

plugin/auth_ldap/src/plugin_simple.cc
```diff
@@ -0,0 +1,189 @@
...
+    connPool->reconfigure(init_pool_size, max_pool_size,
+                          STR_OR_EMPTY(server_host), server_port, ssl, tls,
+                          STR_OR_EMPTY(ca_path), STR_OR_EMPTY(bind_root_dn),
+                          STR_OR_EMPTY(bind_root_pwd));
+    connPool->debug_info();
+  }
+}
+
+static int auth_ldap_simple_init(MYSQL_PLUGIN plugin_info) {
+  g_logger_server = new mysql::plugin::auth_ldap::Ldap_logger();
```
percona-ysorokin, line 61: Why do we need custom logger at all? Can existing MySQL logging infrastructure be reused here?

plugin/auth_ldap/src/plugin_simple.cc
```diff
@@ -0,0 +1,189 @@
...
+#define PASSWORD_QUESTION "\5"
+
+mysql::plugin::auth_ldap::Ldap_logger *g_logger_server;
+
+MYSQL_PLUGIN auth_ldap_simple_plugin_info;
+
+namespace {
+std::mutex active_m;
+std::condition_variable active_cv;
+int active_connections = 0;
```
percona-ysorokin, line 35: std::size_t

plugin/auth_ldap/src/pool.cc
```diff
@@ -0,0 +1,229 @@
...
+      _use_ssl(use_ssl),
+      _use_tls(use_tls),
+      _ca_path(ca_path),
+      _bind_dn(bind_dn),
+      _bind_pwd(bind_pwd) {
+  std::lock_guard<std::mutex> lock(_pool_mutex);
+
+  _bs_used.resize(_pool_max_size);
+  _v_connections.resize(_pool_max_size);
+  for (unsigned int i = 0; i < _pool_max_size; i++) {
```
percona-ysorokin: std::size_t

plugin/auth_ldap/tests/mtr/auth_ldap_simple.test
```diff
@@ -0,0 +1,44 @@
+
+# This testcase requires the following ldap structure:
+# https://github.com/rroemhild/docker-test-openldap
+# The easiest way to run it is to use the provided docker image
+
```
percona-ysorokin, line 5: use `count_sessions.inc/wait_until_count_sessions.inc` for MTR test cases establising new connections

plugin/auth_ldap/tests/mtr/auth_ldap_simple.test
```diff
@@ -0,0 +1,44 @@
...
+SELECT PLUGIN_NAME, PLUGIN_STATUS FROM INFORMATION_SCHEMA.PLUGINS WHERE PLUGIN_NAME LIKE 'authentication_ldap_simple%';
+SHOW GLOBAL VARIABLES LIKE 'authentication_ldap_simple%';
+SET GLOBAL authentication_ldap_simple_bind_base_dn = 'ou=people,dc=planetexpress,dc=com';
```
percona-ysorokin, line 8: The following 3 variables should be propagated to this MTR test case via environment variables

plugin/auth_ldap/src/plugin_simple.cc
```diff
@@ -0,0 +1,189 @@
...
+#include "plugin/auth_ldap/include/connection.h"
+#include "plugin/auth_ldap/include/plugin_common.h"
+#include "plugin/auth_ldap/include/plugin_log.h"
+#include "plugin/auth_ldap/include/plugin_variables.h"
+#include "plugin/auth_ldap/include/pool.h"
+
+#include <condition_variable>
+#include <mutex>
+
+#define PASSWORD_QUESTION "\5"
```
percona-ysorokin: static constant instead of a macro

percona-ysorokin (changes requested):
Most common problems:
1. Using leading underscore (`_`) characters for identifiers, member names, macro guards - thes names are reserved by c++ standats and should be avoided.
2. Const correctness - please, review method definitions if const attribute can be applied to them.
3. `noexcept` specifier is very important in c++11 - please mark methods if appropriate.
4. For custom container types that are used more than once (like `std::vector<std::string>`) use typedef/using.
5. const return types in function/method declarations are pessimization in c++11 - they prevent rvo/nrvo/copy elision.
6. Most of the index types should be unsigned (almost always std::size_t).
7. c++ style headers `<cstdint>` instead of `<stdint.h>`
8. no old style `typedef struct` / `typedef enum` idioms.
9. Output parameters should be passed by reference. Also consider returning boost::optional<type> instead.
10. Consistency in using west const vs east const - `const std::string s;` vs `std::string const s;`
11. `std::ostringstream` instead of `std::stringstream` for output-only streams.
12. Consider more aggressive using of the `auto` keyword.
13. Consider using `static constexpr` for defining constants where possible.
14. Consider using atomic variables instead of updates in sections guarded by mutexes.

plugin/auth_ldap/tests/mtr/auth_ldap_sasl.result
```diff
@@ -0,0 +1,44 @@
+INSTALL PLUGIN authentication_ldap_sasl SONAME 'authentication_ldap_sasl.so';
```
tplavcic, line 1: This plugin is not being built by this change from what I see. So either this needs to be renamed together with variables and other stuff or some other change needs to be done.

tplavcic (changes requested):

plugin/auth_ldap/include/plugin_variables.h
```diff
@@ -0,0 +1,160 @@
...
+
+struct SYS_VAR;
+
+// If we define static variables in a header file we will have a copy of that
+// variable in each of the compilation units that include that header. Usually
+// not something we want, but perfect in this case to avoid duplicate variable
+// definitions for the simple and sasl plugins.
+
+// Static plugin variables
+static char *auth_method_name;
```
tplavcic, line 29: It seems that variable `auth_method_name` is not used anywhere so it should probably be removed.

plugin/auth_ldap/src/plugin_simple.cc
dutow: It can be negative.

plugin/auth_ldap/src/plugin_simple.cc
dutow: I also asked this, but looks like the enterprise plugin also has this feature.

plugin/auth_ldap/include/plugin_common.h
dutow: This is a mysql callback, with an unsigned int parameter.

plugin/auth_ldap/include/plugin_common.h
dutow: This is a mysql callback, with a parameter like this.

plugin/auth_ldap/include/plugin_common.h
dutow: This is a mysql callback, with an unsigned int parameter.

plugin/auth_ldap/include/plugin_common.h
dutow: This is a mysql callback, with an unsigned int parameter.

plugin/auth_ldap/include/plugin_common.h
dutow: This is a mysql callback, with an unsigned int parameter.

percona-ysorokin (approved):
LGTM as initial (experimental) implementation.

PR #3357, opened 2019-08-11
GPL UDF function to send emails from MySQL

fmbiete:
Simple curl based UDF function to send emails from MySQL. 
For example; I use this to notify of the results of MySQL scheduled events.

### DIFF
--- a/plugin/mail_udf/CMakeLists.txt
+++ b/plugin/mail_udf/CMakeLists.txt
@@ -1,0 +1,36 @@
+# Copyright (c) 2018, 2019, Francisco Miguel Biete Banon. All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License, version 2.0,
+# as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License, version 2.0, for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA
+IF(WITH_MAIL)
+  SET(MAILUDF_SOURCES
+    src/plugin.cc
+    src/udf_registration.cc
+    src/udf_mail_send.cc
+  )
+
+  ### Configuration ###
+
+  INCLUDE(curl)
+
+  IF(CURL_FOUND)
+    # declare the plugin itself
+    MYSQL_ADD_PLUGIN(udf_mail
+                     ${MAILUDF_SOURCES}
+                     LINK_LIBRARIES curl
+                     MODULE_ONLY
+                     MODULE_OUTPUT_NAME "udf_mail")
+  ELSE()
+    MESSAGE(FATAL_ERROR "Plugin Mail UDF won't be compiled: curl package is missing")
+  ENDIF()
+ENDIF(WITH_MAIL)
--- a/plugin/mail_udf/README.md
+++ b/plugin/mail_udf/README.md
@@ -1,0 +1,19 @@
+# Mail UDF
+
+Sends basic emails from MySQL using curl
+
+## Requirements
+- libcurl
+
+
+## Installation
+_Requires MySQL v8.0_
+```
+mysql> INSTALL PLUGIN mail SONAME 'udf_mail.so';
+```
+
+## Usage
+```
+select mail_send('smtp://smtp.domain:port', 'from@address', 'to@adddress', 'cc@address or empty', 'Mail subject', 'Mail body');
+select mail_send('smtps://smtp.domain:port', 'from@address', 'to@adddress', 'cc@address or empty', 'Mail subject', 'Mail body');
+```
--- a/plugin/mail_udf/include/plugin.h
+++ b/plugin/mail_udf/include/plugin.h
@@ -1,0 +1,35 @@
+#ifndef _PLUGIN_MAIL_PLUGIN_H
+/* Copyright (c) 2019 Francisco Miguel Biete Banon. All rights reserved.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; version 2 of the License.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software Foundation,
+   51 Franklin Street, Suite 500, Boston, MA 02110-1335 USA
+*/
+#define _PLUGIN_MAIL_PLUGIN_H
+
+#include <string>
+
+#include "m_string.h"
+#include "my_dbug.h"
+#include "my_sys.h"
+#include "mysql/plugin.h"
+#include "sql/log.h"
+#include "sql/sql_plugin.h"
+
+#define LOG_COMPONENT_TAG "mail_udf"
+
+#include "mysql/components/services/log_builtins.h"
+#include "mysqld_error.h"
+
+#include <curl/curl.h>
+
+#endif  //_PLUGIN_MAIL_PLUGIN_H
--- a/plugin/mail_udf/include/udf_descriptor.h
+++ b/plugin/mail_udf/include/udf_descriptor.h
@@ -1,0 +1,53 @@
+#ifndef _PLUGIN_MAIL_UDFDESCRIPTOR_H
+/* Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License, version 2.0,
+   as published by the Free Software Foundation.
+
+   This program is also distributed with certain software (including
+   but not limited to OpenSSL) that is licensed under separate terms,
+   as designated in a particular file or component or in included license
+   documentation.  The authors of MySQL hereby grant you an additional
+   permission to link the program and your derivative works with the
+   separately licensed software that they have included with MySQL.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License, version 2.0, for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA
+*/
+
+#define _PLUGIN_MAIL_UDFDESCRIPTOR_H
+
+#include "mysql/udf_registration_types.h"
+
+/**
+ * Contains all the necessary data to register an UDF in MySQL.
+ */
+struct udf_descriptor {
+  char const *name;
+  enum Item_result result_type;
+  Udf_func_any main_function;
+  Udf_func_init init_function;
+  Udf_func_deinit deinit_function;
+
+  udf_descriptor(char const *udf_name, enum Item_result udf_result_type,
+                 Udf_func_any udf_function, Udf_func_init udf_init,
+                 Udf_func_deinit udf_deinit)
+      : name(udf_name),
+        result_type(udf_result_type),
+        main_function(udf_function),
+        init_function(udf_init),
+        deinit_function(udf_deinit) {}
+  udf_descriptor(udf_descriptor const &) = delete;
+  udf_descriptor(udf_descriptor &&other) = default;
+  udf_descriptor &operator=(udf_descriptor const &) = delete;
+  udf_descriptor &operator=(udf_descriptor &&other) = default;
+};
+
+#endif /* _PLUGIN_MAIL_UDFDESCRIPTOR_H */
--- a/plugin/mail_udf/src/plugin.cc
+++ b/plugin/mail_udf/src/plugin.cc
@@ -1,0 +1,63 @@
+/* Copyright (c) 2019 Francisco Miguel Biete Banon. All rights reserved.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; version 2 of the License.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software Foundation,
+   51 Franklin Street, Suite 500, Boston, MA 02110-1335 USA
+*/
+
+#include "plugin/mail_udf/include/plugin.h"
+#include "plugin/mail_udf/include/udf_registration.h"
+
+static int mail_plugin_init(void *) {
+  DBUG_ENTER("mail_plugin_init");
+
+  CURLcode res = curl_global_init(CURL_GLOBAL_ALL);
+  if (res != CURLE_OK) {
+    LogPluginErr(ERROR_LEVEL, ER_LOG_PRINTF_MSG,
+                 "Mail Plugin: ERROR initilizing CURL");
+    LogPluginErr(ERROR_LEVEL, ER_LOG_PRINTF_MSG, curl_easy_strerror(res));
+    DBUG_RETURN(1);
+  }
+
+  register_udfs();
+
+  DBUG_RETURN(0);
+}
+
+static int mail_plugin_deinit(void *) {
+  DBUG_ENTER("mail_plugin_deinit");
+
+  curl_global_cleanup();
+
+  unregister_udfs();
+
+  DBUG_RETURN(0);
+}
+
+struct st_mysql_daemon mail_plugin = {MYSQL_DAEMON_INTERFACE_VERSION};
+
+mysql_declare_plugin(mail){
+    MYSQL_DAEMON_PLUGIN,                     // Type
+    &mail_plugin,                            // info struct for plugin type
+    "mail",                                  // Short name
+    "Oracle. Francisco Miguel Biete Banon",  // Author
+    "Mail plugin",                           // Long name
+    PLUGIN_LICENSE_GPL,                      // License
+    mail_plugin_init,                        // Init function
+    NULL,                                    // Uninstall function
+    mail_plugin_deinit,                      // Deinit function
+    0x0100,                                  // version
+    NULL,                                    // status variables
+    NULL,                                    // system variables
+    NULL,                                    // config options
+    0,                                       // flags
+} mysql_declare_plugin_end;
--- a/plugin/mail_udf/src/udf_mail_send.cc
+++ b/plugin/mail_udf/src/udf_mail_send.cc
@@ -1,0 +1,231 @@
+/* Copyright (c) 2019 Francisco Miguel Biete Banon. All rights reserved.
+
+   This program is free software; you can redistribute it and/or modify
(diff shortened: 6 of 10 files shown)


plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+  std::string result;
+  CURL *curl;
+
+  curl = curl_easy_init();
+  if (curl) {
+    CURLcode res = CURLE_OK;
+    struct curl_slist *recipients = nullptr;
+    time_t rawtime;
+    struct tm *timeinfo;
+    char buffer[80];
```
dutow: Please use a static const variable instead of the magic constant (below too). Please also use a more specific variable name, e.g. `formatted_date`.

plugin/mail_udf/src/plugin.cc
```diff
@@ -0,0 +1,69 @@
...
+  curl_global_cleanup();
+
+  unregister_udfs();
+
+  DBUG_RETURN(0);
+}
+
+struct st_mysql_daemon mail_plugin = {MYSQL_DAEMON_INTERFACE_VERSION};
+
+//#define mysql_declare_plugin_end   ,{0,0,0,0,0,0,0,0,0,0,0,0,0,0}} <- 14 v8.0
```
dutow: Please don't leave commented out code

plugin/mail_udf/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+
+IF(CURL_FOUND)
+  # declare the plugin itself
+  MYSQL_ADD_PLUGIN(udf_mail
+                   ${MAILUDF_SOURCES}
+                   LINK_LIBRARIES curl
+                   MODULE_ONLY
+                   MODULE_OUTPUT_NAME "udf_mail")
+ELSE()
+  message(STATUS "Plugin Mail UDF won't be compiled: curl package is missing")
```
dutow: I would go with `FATAL_ERROR`, as this means we'll skip the plugin. A `WITH_EMAIL` or similar switch could be useful too, like in the LDAP plugin.

Also please use consistent case in CMake too (`MESSAGE`)

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+static std::string _mail_send(const char *smtp_uri, const char *from,
+                              const char *to, const char *cc,
+                              const char *subject, const char *body) {
+  std::string result;
+  CURL *curl;
+
+  curl = curl_easy_init();
+  if (curl) {
+    CURLcode res = CURLE_OK;
+    struct curl_slist *recipients = nullptr;
```
dutow: These `struct` keywords aren't required, as this is a C++ source, please use C++ style.

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+    char buffer[80];
+    struct upload_message msg;
+
+    time(&rawtime);
+    timeinfo = localtime(&rawtime);
+    strftime(buffer, 80, "%a, %e %b %Y %H:%M:%S %z", timeinfo);
+
+    msg.lines_read = 0;
+    msg.lines.push_back(std::string("Date: ").append(buffer).append("\r\n"));
+    msg.lines.push_back(std::string("From: ").append(from).append("\r\n"));
```
dutow: These user specified parameters should be validated before used, otherwise, it's possible for users to inject additional headers - and even to skip later headers specified in this file by injecting an additional line break.

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+    msg.lines.push_back(
+        std::string("Subject: ").append(subject).append("\r\n"));
+    msg.lines.push_back(std::string("\r\n"));
+    msg.lines.push_back(body);
+    msg.lines.push_back(UDF_MAIL_SEND_EOM);
+
+    curl_easy_setopt(curl, CURLOPT_URL, smtp_uri);
+    if (strncmp(smtp_uri, "smtps", 5) == 0) {
+      curl_easy_setopt(curl, CURLOPT_USE_SSL, (long)CURLUSESSL_ALL);
+      curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, 0L);
```
dutow: I don't think disabling these checks without any condition is a good idea - maybe behind an option?

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+    if (cc && strlen(cc) > 0)
+      msg.lines.push_back(std::string("Cc: ").append(cc).append("\r\n"));
+    msg.lines.push_back(
+        std::string("Subject: ").append(subject).append("\r\n"));
+    msg.lines.push_back(std::string("\r\n"));
+    msg.lines.push_back(body);
+    msg.lines.push_back(UDF_MAIL_SEND_EOM);
+
+    curl_easy_setopt(curl, CURLOPT_URL, smtp_uri);
+    if (strncmp(smtp_uri, "smtps", 5) == 0) {
```
dutow, line 151: Please also check the other case - the user should specify an SMTP server, not something else.

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  return result;
+}
+
+static char *mail_send(UDF_INIT *, UDF_ARGS *args, char *result,
+                       unsigned long *length, char *is_null, char *is_error) {
+  DBUG_ENTER("mail_send");
+
+  std::string res = _mail_send(args->args[0], args->args[1], args->args[2],
```
dutow: Please use `const` where possible.

plugin/mail_udf/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+
+SET(MAILUDF_SOURCES
+  src/plugin.cc
+  src/udf_registration.cc
+  src/udf_mail_send.cc
+)
+
+### Configuration ###
+
+INCLUDE(CheckLibraryExists)
```
dutow: Why is this needed?

plugin/mail_udf/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+Sends basic emails from MySQL using curl
+
+## Requirements
+- libcurl
+
+
+## Installation
+
+### MySQL v5.7 or superior
```
dutow: The code as-is only works with MySQL 8.0, as it requires C++11. I would go with only documenting that version here.

plugin/mail_udf/include/plugin.h
```diff
@@ -0,0 +1,30 @@
...
+#include "m_string.h"
+#include "my_dbug.h"
+#include "my_sys.h"
+#include "mysql/plugin.h"
+#include "sql/log.h"
+#include "sql/sql_plugin.h"
+
+#include <curl/curl.h>
+
+#endif  //_PLUGIN_MAIL_PLUGIN_H
```
dutow, line 35: Please use `/* ... */` for comments. (at other places too)

plugin/mail_udf/src/plugin.cc
```diff
@@ -0,0 +1,69 @@
...
+   51 Franklin Street, Suite 500, Boston, MA 02110-1335 USA
+*/
+
+#include "plugin/mail_udf/include/plugin.h"
+#include "plugin/mail_udf/include/udf_registration.h"
+
+static int mail_plugin_init(void *) {
+  DBUG_ENTER("mail_plugin_init");
+
+  sql_print_information("Mail Plugin: Initializing Mail plugin structures");
```
dutow: This seems like debug information, and even then DBUG_ENTER will tell the same in a trace file. Some other similar prints too. Please consider removing them.

plugin/mail_udf/src/plugin.cc
```diff
@@ -0,0 +1,69 @@
...
+
+  CURLcode res = curl_global_init(CURL_GLOBAL_ALL);
+  if (res != CURLE_OK) {
+    sql_print_error("Mail Plugin: ERROR initilizing CURL");
```
dutow: Please use `LogPluginErr`

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+#include <vector>
+
+#include <curl/curl.h>
+
+#define UDF_MAIL_SEND_EOM "UDF_MAIL_SEND_EOM"
+#define UDF_MAIL_SEND_OK "Mail sent"
+
+struct upload_message {
+  int lines_read;
+  std::vector<std::string> lines;
```
dutow: I have two issues with this class/how it is used:

(1) it's misleading, this vector doesn't store lines - for example, the entire message body is in one entry of this vector
(2) most of the time, it underuses the buffer in the payload functions, and in some cause it could cause an overflow

It looks like you copied this from the basic curl smtp example (which also has an author and a copyright message), with the simple difference that there lines were really short lines.

I don't think  `READFUNCTION` is needed for this plugin at all - you could simply build a stringstream, and then pass the data as `stream.str().c_str()` to curl, it should be able to handle it.

And at that point, this code won't be the same as the curl example anymore.

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+#include "plugin/mail_udf/include/plugin.h"
+
+#include <ctime>
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include <curl/curl.h>
+
+#define UDF_MAIL_SEND_EOM "UDF_MAIL_SEND_EOM"
```
dutow: I don't see the need for this define. With the suggested stringstream change, it won't be needed at all - but even with the current code, you could check `lines_read` against the size of the vector.

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+#define UDF_MAIL_SEND_OK "Mail sent"
```
dutow: Please change this to a static constant.

plugin/mail_udf/include/udf_descriptor.h
```diff
@@ -0,0 +1,48 @@
...
+   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA */
+
+#define _PLUGIN_MAIL_UDFDESCRIPTOR_H
+
+#include "mysql/udf_registration_types.h"
+
+/**
+ * Contains all the necessary data to register an UDF in MySQL.
+ */
+struct udf_descriptor {
```
dutow, line 32: This (and later related parts) seems to be copy-pasted from some other (group replication?) plugin, and are present in the data masking pull request too. 

This is bad for multiple reasons:
* duplicating a few hundred lines of code isn't a good idea
* Oracle isn't mentioned as author in any of the plugins
* for this plugin, this complex infrastructure isn't needed at all: registering a single UDF function would only take a few lines, in one file

So in this case, I think writing a simple init functions which register the UDF function without these additional utilities will be easier.

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+      0;  // Non-Deterministic: same arguments will produce different values
+  initid->ptr = NULL;
+
+  DBUG_RETURN(false);
+}
+
+static void mail_send_deinit(UDF_INIT *initid) {
+  DBUG_ENTER("mail_send_deinit");
+
+  if (initid->ptr) free(initid->ptr);
```
dutow: When is this pointer set?

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+
+  return;
+}
+
+static size_t _prepare_message(char *ptr, size_t size, size_t nitems,
```
dutow, line 95: This function isn't necessary. (see other comment about using a stringstream & copying an example with copyright)

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+
+  if (size == 0 || nitems == 0 || (size * nitems) < 1) {
+    return 0;
+  }
+
+  data = msg->lines[msg->lines_read].c_str();
+
+  if (strcmp(data, UDF_MAIL_SEND_EOM) != 0) {
+    size_t len = strlen(data);
+    memcpy(ptr, data, len);
```
dutow: This is where currently a buffer overflow is possible: for a long email, `len > size * nitems` is possible.

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+    curl_easy_setopt(curl, CURLOPT_URL, smtp_uri);
+    if (strncmp(smtp_uri, "smtps", 5) == 0) {
+      curl_easy_setopt(curl, CURLOPT_USE_SSL, (long)CURLUSESSL_ALL);
+      curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, 0L);
+      curl_easy_setopt(curl, CURLOPT_SSL_VERIFYHOST, 0L);
+    }
+
+    curl_easy_setopt(curl, CURLOPT_MAIL_FROM, from);
+    recipients = curl_slist_append(recipients, to);
+    if (strlen(cc) > 0) recipients = curl_slist_append(recipients, cc);
```
dutow: And if it's empty?

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+      !args->args[4] ||  // subject
+      !args->args[5]     // body
+  ) {
+    std::snprintf(message, MYSQL_ERRMSG_SIZE,
+                  "Arguments cannot be null: mail_send(smtp uri, from, to, cc, "
+                  "subject, body)");
+    DBUG_RETURN(true);
+  }
+
+  initid->maybe_null = 1;
```
dutow: I don't see when this function returns a NULL?

plugin/mail_udf/src/udf_mail_send.cc
```diff
@@ -0,0 +1,190 @@
...
+}
+
+static char *mail_send(UDF_INIT *, UDF_ARGS *args, char *result,
+                       unsigned long *length, char *is_null, char *is_error) {
+  DBUG_ENTER("mail_send");
+
+  std::string res = _mail_send(args->args[0], args->args[1], args->args[2],
+                               args->args[3], args->args[4], args->args[5]);
+  *length = res.length();
+  strcpy(result, res.c_str());
```
dutow: Please use `srtncpy`, there is a possible buffer overflow here.

plugin/mail_udf/src/udf_registration.cc
```diff
@@ -0,0 +1,110 @@
+/* Copyright (c) 2019 Francisco Miguel Biete Banon. All rights reserved.
```
dutow: This file looks mostly the same as the one in the GR plugin? If you do keep it, please don't include it under your own copyright and mention Oracle as an author. If you simplify the UDF initialization, it isn't needed.

dutow (changes requested):
Thanks!

The idea is great, sending e-mails directly from MySQL could have its uses.

I added some comments to specific codeparts, where I see some issues.

I also have a more generic concern: security. 

In this form, if this plugin is installed, it allows any user to send any number of messages, to any server, at any rate, without any logging.

* I don't think allowing anybody to send an email is a good idea - somehow this should be limited (list of allowed users, some role, ...)
* I don't think allowing the user of the function to specify any SMTP server is a good idea. The server, or a whilelist of possible servers should be specified as a global variable
* The same could be true for the sender/addressee too
* There should be an option to log sent emails - at least which user sent an email to who, when, so the server administrator could see this without owning the used SMTP server too

We will also need some tests with this, possibly with a dummy smtp server started as part of an MTR testcase. But as even that needs some environment setup, maybe we should implement that after merging.

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Changed

plugin/mail_udf/src/plugin.cc
fmbiete: Removed

plugin/mail_udf/CMakeLists.txt
fmbiete: Changed

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Changed

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Fixed.

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Changed.

plugin/mail_udf/src/udf_registration.cc
fmbiete: Added Oracle.

plugin/mail_udf/CMakeLists.txt
fmbiete: Removed

plugin/mail_udf/README.md
fmbiete: Changed

plugin/mail_udf/src/plugin.cc
fmbiete: Removed

plugin/mail_udf/src/plugin.cc
fmbiete: Changed

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Changed to use stringstream.
But we need to use READFUNCTION. READDATA is a pointer to the argument passed to READFUNCTION (if it's used) or a FILE pointer (if READFUNCTION is not specified). We cannot put the message text as it is.

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Changed

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Removed

plugin/mail_udf/include/udf_descriptor.h
fmbiete: Added Oracle copyright

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Removed

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: We need it

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Added a flag to disable/or not certificate verification.

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Added validation for smtp uri

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: It's already using const

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Fixed; validating that header parameters don't contain line breaks. But they can contain anything else.

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Fixed

plugin/mail_udf/src/udf_mail_send.cc
fmbiete: Fixed

plugin/mail_udf/include/plugin.h
fmbiete: // is also valid and used across MySQL code. In some parts the old C format makes sense, in other I don't think so.

dutow:
Hello!

We decided not to include these patches in Percona Server for MySQL because of security considerations.

PR #3367, opened 2019-08-15
(8.0) PS-5747 : Remove AHI overhead for DDLs when AHI is disabled

satya-bodapati:
Problem:
--------
Even when AHI is disabled, there is overhead because of AHI.
This is because of checks required for AHI to delete hash entries
of pages being dropped.

These checks are not necessary if we know for sure that AHI is already
disabled.

Fix:
----
1. Take extra dict_sys mutex when enabling AHI (btr_search_enable()). This
   is now similar to AHI disable (btr_search_disable())

2. btr_search_enabled variable read under dict_sys mutex is now accurate
   and can be used to know if AHI is disabled for sure and can't be enabled
   currently.

Fix is similar to PS-5576 - Server stalls because ALTER TABLE on partitioned
                             table holds dict mutex

This fix is more generic and should cover DROP TABLE, DROP INDEX also (apart
from TRUNCATE)

(cherry picked from commit 496cde45f0c33d0dd69ed43c3d79936e9b1b13e9)

### DIFF
--- a/mysql-test/suite/innodb/r/percona_ddl_ahi.result
+++ b/mysql-test/suite/innodb/r/percona_ddl_ahi.result
@@ -1,0 +1,40 @@
+CREATE TABLE t1(a INT, b INT, KEY k1(b));
+INSERT INTO t1 VALUES (1,1),(2,2),(3,3),(4,4),(5,5);
+CREATE TABLE t2 AS SELECT * FROM t1;
+FLUSH TABLES;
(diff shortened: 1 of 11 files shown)


satya-bodapati:
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/6056/

laurynas-biveinis (approved):

PR #3370, opened 2019-08-16
PS-5867: Re-enable keyring_vault tests on Jenkins.

robgolebiowski:
After PS-5772 and PS-5790 are completed it is safe now to re-add
keyring_vault to default test suite. They were removed by PS-5102.


robgolebiowski:
self-approved

PR #3372, opened 2019-08-19
 PS-5216: Fix utility user invisibility issues, and document limitations (5.7)

dutow:


storage/perfschema/unittest/pfs_server_stubs.cc
```diff
@@ -44,3 +44,9 @@ void sql_print_warning(const char *format, ...)
 {
   /* Do not pollute the unit test output with annoying messages. */
 }
+
+my_bool
+acl_is_utility_user(const char *user, const char *host, const char *ip)
```
percona-ysorokin: Please, double check this does not trigger any `unused arguments` warnings. This concerns 5.6 as well

percona-ysorokin (approved):
LGTM

percona-ysorokin (approved):
LGTM

sql/auth/sql_auth_cache.cc
```diff
@@ -1606,11 +1606,13 @@ acl_init_utility_user(bool check_no_resolve)
     {
       if (*cur_pos == ',' || *cur_pos == '\0')
       {
-        acl_utility_user_schema_access.push_back(std::string(cur_db,
-                                                             cur_pos
-                                                             - cur_db));
+	if (cur_pos - cur_db > 1) {
```
percona-ysorokin: Shouldn't there be `cur_pos - cur_db >= 1`?

dutow: No, that still would include the empty entries - if the length of this string is one, it is the comma, or the end of string character.

percona-ysorokin: consider the input `a,b`
when `cur_pos` points to the comma `,`,  `cur_db` points to the beginning of the string `a`
in this case `cur_pos - cur_db == 1`, we won't get into this branch

dutow: Sorry, you are right. And I did this correctly in 5.6, it's only incorrect on 5.7...

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/150/

PR #3381, opened 2019-08-20
5.7 PS-5642 - page tracker thread won't exist if startup fails

altmannmarcelo:
If page tracker is enabled and innodb fail to start,
shutdown process won't signal page tracker thread to exit.

This fix ensure those threads will be proper signaled and exit.

### DIFF
--- a/mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread.test
+++ b/mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread.test
@@ -1,0 +1,20 @@
+#
+# PS-5642 - page tracker thread won't exist if startup fails
+#
+
+--source include/have_debug.inc
+--source include/not_embedded.inc
+
+SELECT 1;
+--source include/shutdown_mysqld.inc
+--error 1
+--exec $MYSQLD_CMD --innodb_track_changed_pages=TRUE --debug=d,ib_dic_boot_error --log-error=$MYSQLTEST_VARDIR/tmp/percona_changed_page_bmp_shutdown_thread.err
+
+--echo # Asserting that the error was not thrown.
+--let log_file_full_path=$MYSQLTEST_VARDIR/tmp/percona_changed_page_bmp_shutdown_thread.err
+--let grep_pattern=threads created by InnoDB had not exited at shutdown
+--let log_expected_matches=0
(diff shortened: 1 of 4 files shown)


altmannmarcelo:
https://ps57.cd.percona.com/job/percona-server-5.7-param/145/

mysql-test/suite/innodb/t/percona_changed_page_bmp_shutdown_thread.test
```diff
@@ -0,0 +1,25 @@
...
+--error 1
+--exec $MYSQLD_CMD --innodb_track_changed_pages=TRUE --debug=d,ib_dic_boot_error --log-error=$MYSQLTEST_VARDIR/tmp/percona_changed_page_bmp_shutdown_thread.err
+
+--echo # Asserting that the error was not thrown.
+--let log_file_full_path=$MYSQLTEST_VARDIR/tmp/percona_changed_page_bmp_shutdown_thread.err
+--let grep_pattern=threads created by InnoDB had not exited at shutdown
+--let log_expected_matches=0
+--source include/log_grep.inc
+
+--echo # Asserting that the server aborted.
```
laurynas-biveinis: Same comment as for the 8.0 PR

altmannmarcelo:
ready for next review

dutow (changes requested):

altmannmarcelo:
@dutow ready for next review.

altmannmarcelo:
@dutow ready for next review

storage/innobase/srv/srv0start.cc
```diff
@@ -1451,6 +1451,12 @@ srv_shutdown_all_bg_threads()
 			if (srv_n_fil_crypt_threads_started) {
 				os_event_set(fil_crypt_threads_event);
 			}
+
+			/* Stop srv_redo_log_follow_thread thread */
+			if (srv_redo_log_thread_started) {
```
dutow, line 1456: Is 5.7 affected then? If yes, the jira ticket should also be updated.

altmannmarcelo:
@dutow fixed. Added 5.6 and 5.7 as affected.

dutow (approved):
5.6 isn't affected, you only ported the testcase there. Other than that looks good.

PR #3384, opened 2019-08-21
Merge upstream 8.0.17 into Percona Server

percona-ysorokin:
https://jira.percona.com/browse/PS-5363


share/errmsg-utf8.txt
```diff
@@ -9143,6 +9143,131 @@ ER_DA_GRP_RPL_STARTED_AUTO_REJOIN
 reserved-error-section 5000 5999
 
 
+#
```
percona-ysorokin: @georgelorch @satya-bodapati Please read the commit message and confirm if you are OK with such approach - this is the best way to avoid assertions I come up with so far.

storage/innobase/include/data0type.h
```diff
@@ -223,7 +223,7 @@ temporary table */
 #define DATA_MULTI_VALUE 16384 /* Multi-value Virtual column */
 
 #define DATA_COMPRESSED                          \
-  16384 /* this is ORed to the precise data      \
+  32768 /* this is ORed to the precise data      \
```
percona-ysorokin: @satya-bodapati Do you think there can be potential conflict after upgrading a table with COMPRESSED STORED VIRTUAL columns from 8.0.16 to 8.0.17 in this case. Or is this flag used only in run-time?
May be move this bit flag even further to avoid situations like this in future 65536 or 131072?

breisig:
What else needs to get done to get this merged and have a release soon after?

storage/innobase/include/data0type.h
satya-bodapati: I think this will have upgrade implications. 5.7 datadir with compressed columns cannot be upgraded to 8.0.17. Please verify

In 8.0, this flag value might not be stored on disk. So we can move this to higher value I think. Am debugging a bit. Will confirm.

satya-bodapati: Confirmed that 8.0 can use higher in-mem value. We don't store this DATA_COMPRESSED on disk and we build this from TABLE*->field->format()

satya-bodapati: We cannot shift this to higher value. The prtype is stored as 4 bytes and split as first two btyes (collation) and next two bytes for precision type (pr stands for precision). 
So DATA_COMPRESSED is already at the last available position. We are lucky!

But this is a problem indeed. We should plan to handle this as separate PR
1.  Remove DATA_COMPRESESD altogether. Add in-memory members in dict_col_t and dfield_t. This could cause more code changes.
Or
2. Extend prtype to 8byte. Reserve the last two bytes for collation and remaining 6bytes for collation. Less code changes but we have to careful with casting and bit shifting at many places.

My preference is 1.

breisig:
Any update to get this merged?

breisig:
MySQL 8.0.18 was released 4 days ago.

percona-ysorokin:
Jenkins build link (--big-test)
https://ps80.cd.percona.com/job/percona-server-8.0-param/370/

PR #3385, opened 2019-08-21
ps-5.7-5797 Correction on release note's release date

patrickbirch:
fixed release date to 2019


dutow (changes requested):
I think this pull request is based on a wrong parent commit - the file you add here exists for 3 months now, but in this change, it looks like it's freshly reated.

This is also not what the commit message describes: it should be only a date change, not adding a whole file.

dutow (approved):
Please edit the commit message before merging, and mention that this is specific to 5.7, doesn't need any change in 8.0. This will make the next null merge easier.

PR #3386, opened 2019-08-21
ps-5.7-5907

patrickbirch:
Added a paragraph in the redo log about the experimental key rotation. This rotation is mentioned in the RN for 5.7.27-30.


dutow:
I don't see any new paragraphs, only alpha changed to experimental everywhere?

dutow (approved):

PR #3387, opened 2019-08-22
Fix PS-5881 (innodb_empty_free_list_algorithm incorrect default)

laurynas-biveinis:
Change innodb_empty_free_list_algorithm default to backoff as it's
supposed to have been since the LRU manager threads had been ported to
8.0. Cleanup some option descriptions, adjust MTR: re-record affected
sys_vars.innodb_empty_free_list_algorithm_basic; switch
innodb.log_file_name_1 to legacy, tweak buffer pool resize in
main.persisted_variables to stay compatible with backoff.

https://ps80.cd.percona.com/job/percona-server-8.0-param/302/
& https://ps80.cd.percona.com/job/percona-server-8.0-param/305/

### DIFF
--- a/mysql-test/t/persisted_variables.test
+++ b/mysql-test/t/persisted_variables.test
@@ -384,7 +384,9 @@ SELECT @@global.general_log, @@global.autocommit,
 
 SELECT @@global.innodb_buffer_pool_size;
 --remove_file $MYSQLD_DATADIR/mysqld-auto.cnf
-SET PERSIST innodb_buffer_pool_size=10*1024*1024;
+# In Percona Server, resize to 20MB instead of 10MB to stay compatible
+# with innodb_empty_free_list_algorithm=backoff
+SET PERSIST innodb_buffer_pool_size=20*1024*1024;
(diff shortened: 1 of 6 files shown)


mysql-test/t/persisted_variables.test
```diff
@@ -384,7 +384,9 @@ SELECT @@global.general_log, @@global.autocommit,
 
 SELECT @@global.innodb_buffer_pool_size;
 --remove_file $MYSQLD_DATADIR/mysqld-auto.cnf
-SET PERSIST innodb_buffer_pool_size=10*1024*1024;
+# In Percona Server, resize to 20MB instead of 10MB to stay compatible
```
satya-bodapati, line 387: optional: instead of 'compatible', would be nice if you can mention about the hang with buf pool size < 20MB

satya-bodapati (approved):

mysql-test/t/persisted_variables.test
laurynas-biveinis: not doing this as discussed on Slack

PR #16, opened 2015-02-24
ps_tokudb_admin fixed error reporting and MYSQLD_SAFE my.cnf section (BLD-250)

tplavcic:
In BLD-250 there's a request to show actual mysql client errors when using ps_tokudb_admin script (it was showing only some generic error before which was misleading) and also to respect if in my.cnf there is a [mysqld_safe] section with all caps so [MYSQLD_SAFE] so this is to fix those two things.

Testing packages are available here (centos7 build fails for some patch which should work after next upstream merge, so I've included bzr build for rpm packages also):
git builds:
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-redhat-binary/132/
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-debian-binary/122/
bzr build for centos7:
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-redhat-binary-bzr/4/

Some testing:
#1: centos6: normal install with changed root pass - error is shown

---

[vagrant@t-centos6-64 ~]$ sudo ps_tokudb_admin --enable
Checking if Percona server is running with jemalloc enabled...
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)

> > Error checking pid file location!
#2: centos6: normal install proper password included

---

[vagrant@t-centos6-64 ~]$ sudo ps_tokudb_admin --enable -u root -pxxx
Script output:
Checking if Percona server is running with jemalloc enabled...

> > Percona server is running with jemalloc enabled.

Checking transparent huge pages status on the system...

> > Transparent huge pages are currently disabled on the system.

Checking if thp-setting=never option is already set in config file...

> > Option thp-setting=never is not set in the config file.
> > (needed only if THP is not disabled permanently on the system)

Checking TokuDB plugin status...

> > TokuDB plugin is not installed.

Adding thp-setting=never option into /etc/my.cnf

> > Successfuly added thp-setting=never option into /etc/my.cnf

Installing TokuDB engine...

> > Successfuly installed TokuDB plugin.

my.cnf output:
[mysqld_safe]
thp-setting=never

show plugins output:
| TokuDB                        | ACTIVE   | STORAGE ENGINE     | ha_tokudb.so | GPL     |
| TokuDB_file_map               | ACTIVE   | INFORMATION SCHEMA | ha_tokudb.so | GPL     |
[...]

PR #13, opened 2015-02-24
Fixed maintainer mode for debug build (BLD-238)

tplavcic:
There was a change before in upstream to enable mysql maintainer_mode for linux debug build by default which created problems with our current build flags so this is to fix it.

build_binary and rpm spec file build using the %optflags so it was added to build_binary to remove optimization flags when Debug build is used - that behavior was already in rpm spec but it didn't contain removal of "-Wp,-D_FORTIFY_SOURCE=2" which needs to be removed also if optimization flags get removed or we get error.
In debian packaging for debug build only optimization flag was removed.
And in both debian/rpm we remove the "-DMYSQL_MAINTAINER_MODE=OFF" option which was there only temporary anyway.

ADDITIONAL FOR 5.6: In rpm spec there's a change for TokuDB build to respect %optflags same as standard build. In debian we build tokudb with standard debian flags, and in build_binary.sh we build release binary with tokudb with respecting these flags. I don't see anything in tickets and in mailing list regarding this - it was introduced when building tokudb alpha and it was left like that after - and we should go for simplifying/standardizing our build flags.

Here's some tests (I've included the git and bzr builds because git has some error in centos7 because of the centos7 patch related to bzr/git migration, but from what I see it should be resolved with the next upstream merge):
5.6 git build:
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-RELEASE/120/
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-redhat-binary/131/
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-debian-binary/121/
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-debian-binary-32/107/
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-binaries-release/92/
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-binaries-opt-yassl/263/
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-binaries-debug-yassl/272/
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-binaries-valgrind-yassl/272/
5.6 bzr job:
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.6-RELEASE-bzr/4/

PR #12, opened 2015-02-24
Fixed maintainer mode for debug build (BLD-238)

tplavcic:
There was a change before in upstream to enable mysql maintainer_mode for linux debug build by default which created problems with our current build flags so this is to fix it.

build_binary and rpm spec file build using the %optflags so it was added to build_binary to remove optimization flags when Debug build is used - that behavior was already in rpm spec but it didn't contain removal of "-Wp,-D_FORTIFY_SOURCE=2" which needs to be removed also if optimization flags get removed or we get error.
In debian packaging for debug build only optimization flag was removed.
And in both debian/rpm we remove the "-DMYSQL_MAINTAINER_MODE=OFF" option which was there only temporary anyway.

Here's some tests (I've included the git and bzr builds because git has some error in centos7 because of the centos7 patch related to bzr/git migration, but from what I see it should be resolved with the next upstream merge):
5.5 git build:
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.5-RELEASE/132/
5.5 bzr build:
http://jenkins.percona.com/view/Percona-RELEASES/job/percona-server-5.5-RELEASE-bzr/1/

PR #11, opened 2015-02-23
Merge MySQL 5.6.23

laurynas-biveinis:
http://jenkins.percona.com/job/percona-server-5.6-param/821/

G2

PR #3390, opened 2019-08-23
PS-5874 : Upgrade to 8.0.16-7 breaks on innodb_dynamic_metadata with encrypted

robgolebiowski:
tablespace

Upgrade creates tables in mysql tablespaces in two phases:
- Phase 1 creates tables based on actual DD properties,
  taken from dd_properties table, option SYSTEM_TABLES.
  We add here in-memory encryption flag to CREATE statement
  if the mysql tablespace is encrypted.
- Phase 2 create target tables. We add here in-memory encryption
  flag to CREATE statement if mysql tablespace is encrypted and update
  SYSTEM_TABLES option in dd_properties with encryption flag.

### DIFF
--- a/mysql-test/t/percona_8_0_15_dd_upgrade_encrypted.test
+++ b/mysql-test/t/percona_8_0_15_dd_upgrade_encrypted.test
@@ -1,0 +1,98 @@
+# This test contains keyring file created on 64bit which is not
+# portable
+--source include/have_64bit.inc
+--source include/no_valgrind_without_big.inc
+--source include/have_util_unzip.inc
+
+--disable_query_log
+call mtr.add_suppression("Resizing redo log from");
+call mtr.add_suppression("Upgrading redo log");
+call mtr.add_suppression("Starting to delete and rewrite log files");
+call mtr.add_suppression("New log files created");
+call mtr.add_suppression("Unknown system variable 'show_compatibility_56'");
+call mtr.add_suppression("You need to use --log-bin to make --binlog-format work");
+call mtr.add_suppression("Creating routine without parsing routine body");
+call mtr.add_suppression("Resolving dependency for the view");
+call mtr.add_suppression("references invalid");
+call mtr.add_suppression("doesn't exist");
+call mtr.add_suppression("information_schema");
+call mtr.add_suppression("Storage engine '.*' does not support system tables. \\[mysql.*\\]");
+call mtr.add_suppression("Table 'mysql.component' doesn't exist");
+call mtr.add_suppression("is expected to be transactional");
+call mtr.add_suppression("table is missing or has an incorrect definition");
+call mtr.add_suppression("ACL DDLs will not work unless mysql_upgrade is executed");
+call mtr.add_suppression("Native table .* has the wrong structure");
+call mtr.add_suppression("Column count of mysql.* is wrong");
+call mtr.add_suppression("Cannot open table mysql/version from the internal data dictionary of InnoDB though the .frm file for the table exists");
+call mtr.add_suppression("Column count of performance_schema.events_statements_summary_by_digest is wrong");
+call mtr.add_suppression("The privilege system failed to initialize correctly");
+call mtr.add_suppression("Missing system table mysql.global_grants");
+# InnoDB reports "Lock wait timeout" warnings when it tries to drop persistent
+# statistics while persistent statistics table is altered during upgrade.
+# This issue doesn't seem to cause any further trouble (as there is no persistent
+# stats for persistent stats table anyway), so we ignore these warnings here.
+call mtr.add_suppression("Unable to delete statistics for table mysql.");
+# new fields were added to these tables
+call mtr.add_suppression("Column count of performance_schema.replication_group_members is wrong. Expected 7, found 5");
+call mtr.add_suppression("Column count of performance_schema.replication_group_member_stats is wrong. Expected 13, found 9");
+call mtr.add_suppression("Column count of performance_schema.threads is wrong. Expected 18, found 17");
+call mtr.add_suppression("ACL table mysql.[a-zA-Z_]* missing. Some operations may fail.");
+call mtr.add_suppression("Info table is not ready to be used. Table 'mysql.slave_master_info' cannot be opened");
+call mtr.add_suppression("Error in checking mysql.slave_master_info repository info type of TABLE");
+call mtr.add_suppression("Error creating master info: Error checking repositories.");
+call mtr.add_suppression("Slave: Failed to initialize the master info structure for channel");
+call mtr.add_suppression("Failed to create or recover replication info repositories.");
+call mtr.add_suppression("db.opt file not found for test database. Using default Character set");
+call mtr.add_suppression("Skip re-populating collations and character sets tables in InnoDB read-only mode");
+call mtr.add_suppression("Skipped updating resource group metadata in InnoDB read only mode");
+--enable_query_log
+
+--echo # Set different paths for --datadir
+let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/data_8015_dd_encrypted;
+
+--copy_file $MYSQLTEST_VARDIR/std_data/data_8015_dd_encrypted.zip $MYSQL_TMP_DIR/data_8015_dd_encrypted.zip
+
+--echo # Check that the file exists in the working folder.
+--file_exists $MYSQL_TMP_DIR/data_8015_dd_encrypted.zip
+
+--echo # Unzip the zip file.
+--exec unzip -qo $MYSQL_TMP_DIR/data_8015_dd_encrypted.zip -d $MYSQL_TMP_DIR
+
+--let $MYSQLD_DATADIR=`SELECT @@datadir`
+
+--echo # Stop DB server which was created by MTR default
+--source include/shutdown_mysqld.inc
+
+--echo # Start the 8.0.16 server on 8.0.15 datadir
+--let $restart_parameters = "restart: --loose-skip-log-bin --skip-log-slave-updates --datadir=$MYSQLD_DATADIR1 --keyring_file_data=$MYSQL_TMP_DIR/data_8015_dd_encrypted/mysecret_keyring $KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD"
+--replace_result $MYSQLD_DATADIR1 MYSQLD_DATADIR1 $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD KEYRING_PLUGIN_EARLY_LOAD
+--source include/start_mysqld.inc
+
+--let $assert_text = Encryption of mysql.ibd should be set to Y.
+--let $assert_cond= "[SELECT ENCRYPTION FROM INFORMATION_SCHEMA.INNODB_TABLESPACES WHERE NAME=\"mysql\"]" = "Y"
+--source include/assert.inc
+
+SET SESSION DEBUG = '+d,skip_dd_table_access_check';
+
+--let $assert_text = dd_properties should not contain information about encryption
+--let $assert_cond= "[SELECT COUNT(*) FROM mysql.dd_properties WHERE properties LIKE \'%=ENCRYPTION%\']" = 0
+--source include/assert.inc
+
+--let $assert_text = All tables in mysql tablespace should have encrypt_type=Y set in options.
+--let $assert_cond= "[SELECT COUNT(*) FROM mysql.tables WHERE tablespace_id = 1 AND options NOT LIKE \'%encrypt_type=Y%\']" = 0
+--source include/assert.inc
+
+# Check if server restarts correctly
+--replace_result $MYSQLD_DATADIR1 MYSQLD_DATADIR1 $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD KEYRING_PLUGIN_EARLY_LOAD
+--source include/restart_mysqld.inc
+
+--source include/shutdown_mysqld.inc
+
+--echo # Remove copied files
+--remove_file $MYSQL_TMP_DIR/data_8015_dd_encrypted.zip
+
+--force-rmdir $MYSQL_TMP_DIR/data_8015_dd_encrypted
+
+--echo # Restart the server with default options.
+--let $restart_parameters=
+--source include/start_mysqld.inc
--- a/sql/dd/impl/bootstrap/bootstrapper.cc
+++ b/sql/dd/impl/bootstrap/bootstrapper.cc
@@ -165,6 +165,9 @@ bool update_system_tables(THD *thd) {
       std::unique_ptr<Properties> table_def_properties(
           Properties::parse_properties(def));
       table_def->set_actual_table_definition(*table_def_properties);
+      if (bootstrap::DD_bootstrap_ctx::instance().is_dd_encrypted()) {
+        table_def->set_actual_encrypted();
+      }
     }
   }
 
@@ -1198,7 +1201,7 @@ bool create_dd_schema(THD *thd) {
 bool initialize_dd_properties(THD *thd) {
   // Create the dd_properties table.
   if (bootstrap::DD_bootstrap_ctx::instance().is_dd_encrypted()) {
-    dd::tables::DD_properties::instance().set_encrypted();
+    dd::tables::DD_properties::instance().set_target_encrypted();
   }
   const Object_table_definition *dd_properties_def =
       dd::tables::DD_properties::instance().target_table_definition();
@@ -1673,6 +1676,25 @@ bool sync_meta_data(THD *thd) {
   return false;
 }
 
+// Helper guard used in update_properties, to be sure
+// that encryption will get set back before
+// update_properties exits.
+struct Target_encryption_guard {
+ public:
+  Target_encryption_guard(const Object_table *object_table)
+      : m_object_table(object_table)
+      : set_encryption(object_table->is_target_encrypted()) {}
+  ~Target_encryption_guard() {
+    if (set_encryption) {
+      m_object_table->set_target_encrypted();
+    }
+  }
+
+ private:
+  const Object_table *m_object_table;
+  bool set_encryption;
+};
+
 bool update_properties(THD *thd, const std::set<String_type> *create_set,
                        const std::set<String_type> *remove_set,
                        const String_type &target_table_schema_name) {
@@ -1692,6 +1714,22 @@ bool update_properties(THD *thd, const std::set<String_type> *create_set,
         will have a corresponding Object_table.
       */
       DBUG_ASSERT((*it)->entity() != nullptr);
+
+      /*
+        Percona Server supports mysql.ibd encryption in earlier versions than
+        upstream. Upstream started supporting it since 8.0.16. Upstream when
+        ALTER TABLESPACE mysql ENCRYPTION='Y' is issued does not update
+        dd_properties table that is updated here. To be in sync with upstream we
+        also do not want to update dd_properties. Since dd_properties are
+        updated based on target definition we unset the encryption from target
+        definition for the time of updating dd_properties. The exact field that
+        contains system tables properties in dd_properties is SYSTEM_TABLES.
+      */
+      Target_encryption_guard target_encryption_guard((*it)->entity());
+      if ((*it)->entity()->is_target_encrypted()) {
+        (*it)->entity()->unset_target_encrypted();
+      }
+
       const Object_table_definition *table_def =
           (*it)->entity()->target_table_definition();
 
--- a/sql/dd/impl/types/object_table_impl.cc
+++ b/sql/dd/impl/types/object_table_impl.cc
@@ -57,9 +57,26 @@ Object_table_impl::Object_table_impl()
       String_type("TABLESPACE=") + String_type(MYSQL_TABLESPACE_NAME.str));
 }
 
-void Object_table_impl::set_encrypted() {
-  m_target_def.add_option(static_cast<int>(Common_option::ENCRYPTION),
-                          "ENCRYPTION", "ENCRYPTION='Y'");
+bool Object_table_impl::is_target_encrypted() const {
+  return m_target_def.has_option(static_cast<int>(Common_option::ENCRYPTION));
+}
+
+void Object_table_impl::unset_target_encrypted() const {
+  m_target_def.remove_option("ENCRYPTION");
+}
+
+void Object_table_impl::set_target_encrypted() const {
+  if (!m_target_def.has_option(static_cast<int>(Common_option::ENCRYPTION))) {
+    m_target_def.add_option(static_cast<int>(Common_option::ENCRYPTION),
+                            "ENCRYPTION", "ENCRYPTION='Y'");
+  }
+}
+
+void Object_table_impl::set_actual_encrypted() const {
+  if (!m_actual_def.has_option(static_cast<int>(Common_option::ENCRYPTION))) {
+    m_actual_def.add_option(static_cast<int>(Common_option::ENCRYPTION),
+                            "ENCRYPTION", "ENCRYPTION='Y'");
+  }
 }
 
 bool Object_table_impl::set_actual_table_definition(
--- a/sql/dd/impl/types/object_table_impl.h
+++ b/sql/dd/impl/types/object_table_impl.h
@@ -35,7 +35,7 @@ namespace dd {
 class Object_table_impl : virtual public Object_table {
  protected:
   mutable uint m_last_dd_version;
-  Object_table_definition_impl m_target_def;
+  mutable Object_table_definition_impl m_target_def;
   mutable bool m_actual_present;
   mutable Object_table_definition_impl m_actual_def;
   bool m_hidden;
@@ -123,7 +123,13 @@ class Object_table_impl : virtual public Object_table {
 
   virtual void set_hidden(bool hidden) { m_hidden = hidden; }
 
-  virtual void set_encrypted();
+  virtual bool is_target_encrypted() const;
+
+  virtual void unset_target_encrypted() const;
+
+  virtual void set_target_encrypted() const;
+
+  virtual void set_actual_encrypted() const;
 
   virtual ~Object_table_impl() {}
 };
--- a/mysql-test/r/percona_8_0_15_dd_upgrade_encrypted.result
+++ b/mysql-test/r/percona_8_0_15_dd_upgrade_encrypted.result
@@ -1,0 +1,14 @@
+# Set different paths for --datadir
+# Check that the file exists in the working folder.
+# Unzip the zip file.
+# Stop DB server which was created by MTR default
+# Start the 8.0.16 server on 8.0.15 datadir
+# restart: --loose-skip-log-bin --skip-log-slave-updates --datadir=MYSQLD_DATADIR1 --keyring_file_data=MYSQLD_DATADIR1/mysecret_keyring KEYRING_PLUGIN_OPT KEYRING_PLUGIN_EARLY_LOAD
+include/assert.inc [Encryption of mysql.ibd should be set to Y.]
+SET SESSION DEBUG = '+d,skip_dd_table_access_check';
+include/assert.inc [dd_properties should not contain information about encryption]
+include/assert.inc [All tables in mysql tablespace should have encrypt_type=Y set in options.]
+# restart: --loose-skip-log-bin --skip-log-slave-updates --datadir=MYSQLD_DATADIR1 --keyring_file_data=MYSQLD_DATADIR1/mysecret_keyring KEYRING_PLUGIN_OPT KEYRING_PLUGIN_EARLY_LOAD
+# Remove copied files
+# Restart the server with default options.
+# restart
--- a/mysql-test/t/percona_8_0_15_dd_upgrade_encrypted-master.opt
+++ b/mysql-test/t/percona_8_0_15_dd_upgrade_encrypted-master.opt
@@ -1,0 +1,1 @@
+--initialize --innodb_page_size=16k
--- a/sql/dd/impl/types/object_table_definition_impl.h
+++ b/sql/dd/impl/types/object_table_definition_impl.h
@@ -118,6 +118,18 @@ class Object_table_definition_impl : public Object_table_definition {
     (*element_definitions)[element_number] = element_definition;
   }
 
+  void remove_element(const String_type &element_name,
+                      Element_numbers *element_numbers,
+                      Element_definitions *element_definitions) {
+    DBUG_ASSERT(element_numbers != nullptr &&
+                element_numbers->find(element_name) != element_numbers->end() &&
+                element_definitions->find((*element_numbers)[element_name]) !=
+                    element_definitions->end());
+
+    element_definitions->erase((*element_numbers)[element_name]);
+    element_numbers->erase(element_name);
+  }
+
   int element_number(const String_type &element_name,
                      const Element_numbers &element_numbers) const {
     DBUG_ASSERT(element_numbers.find(element_name) != element_numbers.end());
@@ -282,6 +294,15 @@ class Object_table_definition_impl : public Object_table_definition {
                           const String_type &option_definition) {
     add_element(option_number, option_name, option_definition,
                 &m_option_numbers, &m_option_definitions);
+  }
+
+  virtual void remove_option(const String_type &option_name) {
+    remove_element(option_name, &m_option_numbers, &m_option_definitions);
+  }
+
+  virtual bool has_option(int option_number) const {
+    return m_option_definitions.find(option_number) !=
+           m_option_definitions.end();
   }
 
   virtual void add_populate_statement(const String_type &statement) {
--- a/sql/dd/types/object_table.h
+++ b/sql/dd/types/object_table.h
@@ -175,7 +175,13 @@ class Object_table {
   */
   virtual void set_hidden(bool hidden) = 0;
 
-  virtual void set_encrypted() = 0;
+  virtual bool is_target_encrypted() const = 0;
+
+  virtual void unset_target_encrypted() const = 0;
+
+  virtual void set_target_encrypted() const = 0;
+
+  virtual void set_actual_encrypted() const = 0;
 
  public:
   virtual ~Object_table() {}
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -12812,10 +12812,10 @@ static bool innobase_ddse_dict_init(
   /* Options and tablespace are set at the SQL layer. */
 
   if (is_dd_encrypted) {
-    innodb_dynamic_metadata->set_encrypted();
-    innodb_table_stats->set_encrypted();
-    innodb_index_stats->set_encrypted();
-    innodb_ddl_log->set_encrypted();
+    innodb_dynamic_metadata->set_target_encrypted();
+    innodb_table_stats->set_target_encrypted();
+    innodb_index_stats->set_target_encrypted();
+    innodb_ddl_log->set_target_encrypted();
   }
 
   tables->push_back(innodb_dynamic_metadata);


sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -165,6 +165,7 @@ bool update_system_tables(THD *thd) {
       std::unique_ptr<Properties> table_def_properties(
           Properties::parse_properties(def));
       table_def->set_actual_table_definition(*table_def_properties);
+      table_def->set_actual_encrypted();
```
satya-bodapati: Why is this done unconditionally?

Side note: since robert is on vacation, I am debugging this to find the answer

satya-bodapati: Few more questions:

Why is the property change done in-memory only and not registerd in DD properties?

Why do we do the "CREATE TABLE mysql.innodb_dynamic_metadata" (the actual table vs target table) when the table already existed in SE. Does the mysql_execute_command() or mysql_create_table() really try to create table in SE?

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1352,7 +1353,9 @@ bool initialize_dd_properties(THD *thd) {
   */
   if (!bootstrap::DD_bootstrap_ctx::instance().is_initialize() &&
       bootstrap::DD_bootstrap_ctx::instance().is_dd_upgrade() &&
-      update_system_tables(thd)) {
+      update_system_tables(thd) &&
+      update_properties(thd, nullptr, nullptr,
```
satya-bodapati: By using this update_properties() the DD_properties of the dictionary tables are stored on disk? Next restart shouldn't depend on in-memory flag?

Also why do we need this unconditionally?

sql/dd/impl/types/object_table_impl.cc
```diff
@@ -58,8 +60,19 @@ Object_table_impl::Object_table_impl()
 }
 
 void Object_table_impl::set_encrypted() {
```
satya-bodapati: I observed that this is used only for DD properties table and for the rest the constructor does this.

satya-bodapati: And in the constructor the member variable is not set. This is a bug
I would suggest to call this set_encrypted() method from constructor

sql/dd/impl/types/object_table_impl.cc
```diff
@@ -35,7 +35,9 @@ Object_table_impl::Object_table_impl()
       m_target_def(),
       m_actual_present(false),
       m_actual_def(),
-      m_hidden(true) {
+      m_hidden(true),
+      was_target_encryption_set{false},
```
satya-bodapati: Better names are m_target_encryption_set and m_actual_encryption_set

Why do you always use "was_"?

sql/dd/impl/types/object_table_impl.h
```diff
@@ -125,6 +127,8 @@ class Object_table_impl : virtual public Object_table {
 
   virtual void set_encrypted();
```
satya-bodapati: may be rename this set_target_encrypted() instead of generic set_encrypted(). We have sibling set_actual_encrypted().

satya-bodapati: Although the person who reviews this code should understand what actual vs target table means, it is better if there is atleast 'some' comment on function (along with description in commit message)

sql/dd/impl/bootstrap/bootstrapper.cc
satya-bodapati: Am guessing that restart would keep using this in-memory flag and add double "encryption=y" again? Have to debug and find out

sql/dd/impl/types/object_table_impl.cc
satya-bodapati: And in the constructor the member variable is not set. This is a bug
I would suggest to call this set_encrypted() method from constructor

satya-bodapati (changes requested):

satya-bodapati:
After applying patch:

Crash after upgrade and restart

Y-010116] [Server] /home/satya/WORK/PS-8.0/bld/runtime_output_directory/mysqld (mysqld 8.0.16-6-debug) starting as process 15893
2019-08-28T11:31:17.368595Z 1 [ERROR] [MY-013183] [InnoDB] Assertion failure: os0file.cc:8983:strlen(srv_uuid) != 0 thread 140737087076096
InnoDB: We intentionally generate a memory trap.
InnoDB: Submit a detailed bug report to http://bugs.mysql.com.
InnoDB: If you get repeated assertion failures or crashes, even
InnoDB: immediately after the mysqld startup, there may be
InnoDB: corruption in the InnoDB tablespace. Please refer to
InnoDB: http://dev.mysql.com/doc/refman/8.0/en/forcing-innodb-recovery.html
InnoDB: about forcing recovery.

Thread 2 "mysqld" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffe814f700 (LWP 15899)]
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff5d3c801 in __GI_abort () at abort.c:79
#2  0x0000555559e1168b in ut_dbg_assertion_failed (expr=0x55555b75a742 "strlen(srv_uuid) != 0", file=0x55555b758038 "/home/satya/WORK/PS-8.0/storage/innobase/os/os0file.cc", line=8983)
    at /home/satya/WORK/PS-8.0/storage/innobase/ut/ut0dbg.cc:92
#3  0x0000555559c4716d in Encryption::get_master_key_from_info (encrypt_info=0x7fffe04ce899 "", version=Encryption::ENCRYPTION_VERSION_3, m_key_id=0x7fffe814d85c, srv_uuid=0x7fffe814da20 "", 
    master_key=0x7fffe814d868) at /home/satya/WORK/PS-8.0/storage/innobase/os/os0file.cc:8983
#4  0x0000555559c47350 in Encryption::decode_encryption_info (key=0x7fffe04c9588 "", iv=0x7fffe04c95c8 "", encryption_info=0x7fffe04ce896 "lCC")
    at /home/satya/WORK/PS-8.0/storage/innobase/os/os0file.cc:9043
#5  0x000055555a00ce70 in fsp_header_get_encryption_key (fsp_flags=26624, key=0x7fffe04c9588 "", iv=0x7fffe04c95c8 "", page=0x7fffe04cc000 "\351%\276\257")
    at /home/satya/WORK/PS-8.0/storage/innobase/fsp/fsp0fsp.cc:1193
#6  0x000055555a01e44d in Datafile::validate_first_page (this=0x7fffe814e070, space_id=4294967294, flush_lsn=0x0, for_import=false) at /home/satya/WORK/PS-8.0/storage/innobase/fsp/fsp0file.cc:724
#7  0x000055555a01d284 in Datafile::validate_to_dd (this=0x7fffe814e070, space_id=4294967294, flags=26624, for_import=false) at /home/satya/WORK/PS-8.0/storage/innobase/fsp/fsp0file.cc:405
#8  0x0000555559fc599f in fil_ibd_open (validate=true, purpose=FIL_TYPE_TABLESPACE, space_id=4294967294, flags=26624, space_name=0x55555b82e2e8 "mysql", table_name=0x55555b82e2e8 "mysql", 
    path_in=0x55555b82e2ee "mysql.ibd", strict=true, old_space=false, keyring_encryption_info=...) at /home/satya/WORK/PS-8.0/storage/innobase/fil/fil0fil.cc:5729
#9  0x0000555559ac55f3 in dd_open_hardcoded (space_id=4294967294, filename=0x55555b82e2ee "mysql.ibd", flags=26624) at /home/satya/WORK/PS-8.0/storage/innobase/handler/ha_innodb.cc:5512
#10 0x0000555559ac5c23 in innobase_init_files (dict_init_mode=DICT_INIT_CHECK_FILES, tablespaces=0x7fffe814e910, is_dd_encrypted=@0x7fffe814e7f5: true)
    at /home/satya/WORK/PS-8.0/storage/innobase/handler/ha_innodb.cc:5655
#11 0x0000555559ad6ec5 in innobase_ddse_dict_init (dict_init_mode=DICT_INIT_CHECK_FILES, version=80016, tables=0x7fffe814e8f0, tablespaces=0x7fffe814e910)
    at /home/satya/WORK/PS-8.0/storage/innobase/handler/ha_innodb.cc:12728
#12 0x0000555559605430 in dd::bootstrap::DDSE_dict_init (thd=0x55555cd56850, dict_init_mode=DICT_INIT_CHECK_FILES, version=80016) at /home/satya/WORK/PS-8.0/sql/dd/impl/bootstrap/bootstrapper.cc:748
#13 0x0000555559810486 in dd::upgrade_57::do_pre_checks_and_initialize_dd (thd=0x55555cd56850) at /home/satya/WORK/PS-8.0/sql/dd/upgrade_57/upgrade.cc:912
#14 0x000055555885d4ff in bootstrap::handle_bootstrap (arg=0x55555cd56850) at /home/satya/WORK/PS-8.0/sql/bootstrap.cc:347
#15 0x00005555599cbb74 in pfs_spawn_thread (arg=0x55555cc48d70) at /home/satya/WORK/PS-8.0/storage/perfschema/pfs.cc:2836
#16 0x00007ffff79b66db in start_thread (arg=0x7fffe814f700) at pthread_create.c:463
#17 0x00007ffff5e1d88f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

satya-bodapati:
I tried to reproduce the above crash in 8.0.15 and I couldn't. May be specific to patch or specific to 8.0.16. Will check

sql/dd/impl/types/object_table_impl.h
satya-bodapati: > I tried to reproduce the above crash in 8.0.15 and I couldn't. May be specific to patch or specific to 8.0.16. Will check

There is new bug with 8.0.16. default_table_encryption=ON at bootstrap doesn't encrypt mysql.ibd.  I fixed this (wrote patch and I have to submit PR)

8.0.16 doesn't have the above issue. So this crash is specific to upgrade of encrypted mysql.ibd from 8.0.15 to 8.0.16

satya-bodapati:
the above crash is easy to reproduce. Just add restart in the test

```
--let $assert_cond= "[SELECT ENCRYPTION FROM INFORMATION_SCHEMA.INNODB_TABLESPACES WHERE NAME=\\'mysql\\']" = 'Y'

--source include/restart_mysqld.inc
SHOW TABLES;

--source include/shutdown_mysqld.inc

```

satya-bodapati:
Another thing to find out how the upstream does "ALTER TABLESPACE mysql ENCRYPTION='Y'". Does it add encryption attribute in dd_properties of all tables?

satya-bodapati:
Found the reason for crash. it is not related to this patch. I could reproduce the  same crash in 8.0.15 using same datadir

sql/dd/impl/bootstrap/bootstrapper.cc
satya-bodapati: Answers from debugging

1. (on the unconditionally question) : No answer yet
2. property change (adding encryption='y') to DD properties is not just in-memory change. It is written to disk as well ( dd::update_properties()) is called
3. the CREATE TABLE innodb_dynamic_metadata doesn't result in table creation in SE. (we already have the table in SE). This step is only to create TABLE*/TABLE_SHARE* for core tables.

sql/dd/impl/types/object_table_impl.h
satya-bodapati: Fixed as part of PS-5925 and PS-5926 (the blockers for this PR)

satya-bodapati:
> Another thing to find out how the upstream does "ALTER TABLESPACE mysql ENCRYPTION='Y'". Does it add encryption attribute in dd_properties of all tables?

The answer is NO. Upstream does kind of "cheating". It doesn't treat mysql tablespace as encrypted during the bootstrap phase.

satya-bodapati:
> Found the reason for crash. it is not related to this patch. I could reproduce the same crash in 8.0.15 using same datadir

Fixed as PS-5925 and PS-5926

sql/dd/impl/bootstrap/bootstrapper.cc
satya-bodapati: DD properties after update_properties is like this:
 tbl_props->raw_string()
$12 = "col0=table_id\\=2\\;;col1=table_id\\=2\\;;col2=table_id\\=2\\;;col3=table_id\\=2\\;;col4=table_id\\=2\\;;data=;def=fields\\=elem0\\\\\\=def\\\\\\\\\\\\\\=metadata BLOB NOT NULL\\\\\\\\\\\\\\;lbl\\\
\\\\\\\\\\\=metadata\\\\\\\\\\\\\\;pos\\\\\\\\\\\\\\=2\\\\\\\\\\\\\\;\\\\\\;elem1\\\\\\=def\\\\\\\\\\\\\\=table_id BIGINT UNSIGNED NOT NULL\\\\\\\\\\\\\\;lbl\\\\\\\\\\\\\\=table_id\\\\\\\\\\\\\\;pos\\\\\\
\\\\\\\\=0\\\\\\\\\\\\\\;\\\\\\;elem2\\\\\\=def\\\\\\\\\\\\\\=version BIGINT UNSIGNED NOT NULL\\\\\\\\\\\\\\;lbl\\\\\\\\\\\\\\=version\\\\\\\\\\\\\\;pos\\\\\\\\\\\\\\=1\\\\\\\\\\\\\\;\\\\\\;\\;foreign_key
s\\=\\;indexes\\=elem0\\\\\\=def\\\\\\\\\\\\\\=PRIMARY KEY (table_id)\\\\\\\\\\\\\\;lbl\\\\\\\\\\\\\\=index_pk\\\\\\\\\\\\\\;pos\\\\\\\\\\\\\\=0\\\\\\\\\\\\\\;\\\\\\;\\;name\\=innodb_dynamic_metadata\\;op
tions\\=elem0\\\\\\=def\\\\\\\\\\\\\\=DEFAULT CHARSET", '\\' <repeats 15 times>, "=utf8\\\\\\\\\\\\\\;lbl\\\\\\\\\\\\\\=CHARSET\\\\\\\\\\\\\\;pos\\\\\\\\\\\\\\=1\\\\\\\\\\\\\\;\\\\\\;elem1\\\\\\=def\\\\\\
\\\\\\\\=COLLATE", '\\' <repeats 15 times>, "=utf8_bin\\\\\\\\\\\\\\;lbl\\\\\\\\\\\\\\=COLLATION\\\\\\\\\\\\\\;pos\\\\\\\\\\\\\\=2\\\\\\\\\\\\\\;\\\\\\;elem2\\\\\\=def\\\\\\\\\\\\\\=ENCRYPTION", '\\' <rep
eats 15 times>, "='Y'\\\\\\\\\\\\\\;lbl\\\\\\\\\\\\\\=ENCRYPTION\\\\\\\\\\\\\\;pos\\\\\\\\\\\\\\=6\\\\\\\\\\\\\\;\\\\\\;elem3\\\\\\=def\\\\\\\\\\\\\\=ENGINE", '\\' <repeats 15 times>, "=INNODB\\\\\\\\\\\\
\\;lbl\\\\\\\\\\\\\\=ENGINE\\\\\\\\\\\\\\;pos\\\\\\\\\\\\\\=0\\\\\\\\\\\\\\;\\\\\\;elem4\\\\\\=def\\\\\\\\\\\\\\=ROW_FORMAT", '\\' <repeats 15 times>, "=DYNAMIC\\\\\\\\\\\\\\;lbl\\\\\\\\\\\\\\=ROW_FORMAT\
\\\\\\\\\\\\\;pos\\\\\\\\\\\\\\=3\\\\\\\\\\\\\\;\\\\\\;elem5\\\\\\=def\\\\\\\\\\\\\\=STATS_PERSISTENT", '\\' <repeats 15 times>, "=0\\\\\\\\\\\\\\;lbl\\\\\\\\\\\\\\=STATS_PERSISTENT\\\\\\\\\\\\\\;pos\\\\\
\\\\\\\\\=4\\\\\\\\\\\\\\;\\\\\\;elem6\\\\\\=def\\\\\\\\\\\\\\=TABLESPACE", '\\' <repeats 15 times>, "=mysql\\\\\\\\\\\\\\;lbl\\\\\\\\\\\\\\=TABLESPACE\\\\\\\\\\\\\\;pos\\\\\\\\\\\\\\=5\\\\\\\\\\\\\\;\\\\
\\;\\;;id=2;idx0=id\\=2\\;root\\=5\\;space_id\\=4294967294\\;table_id\\=2\\;trx_id\\=0\\;;space_id=1;"

We can see that ENCRYPTION="Y". So this goes to disk. Next I have to check if next restart also adds the same

satya-bodapati:
My fear on double encryption is indeed true. I just bumped up DD version to simulate upgrade and it crashed.

#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff5d3c801 in __GI_abort () at abort.c:79
#2  0x00007ffff5d2c39a in __assert_fail_base (fmt=0x7ffff5eb37d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=assertion@entry=0x55555afddef8 "element_numbers != nullptr && element_definitions != nullptr && element_numbers->find(element_name) == element_numbers->end() && element_definitions->find(element_number) == element_definitions->end()", file=file@entry=0x55555afddea8 "/home/satya/WORK/PS-8.0/sql/dd/impl/types/object_table_definition_impl.h", line=line@entry=115, 
    function=function@entry=0x55555afde960 <dd::Object_table_definition_impl::add_element(int, std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const&, std::map<std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> >, int, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const, int> > >*, std::map<int, std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> >, std::less<int>, std::allocator<std::pair<int const, std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > > > >*)::__PRETTY_FUNCTION__> "void dd::Object_table_definition_impl::add_element(int, const String_type&, const String_type&, dd::Object_table_definition_impl::Element_numbers*, dd::Object_table_definition_impl::Element_definitions*)") at assert.c:92
#3  0x00007ffff5d2c412 in __GI___assert_fail (
    assertion=0x55555afddef8 "element_numbers != nullptr && element_definitions != nullptr && element_numbers->find(element_name) == element_numbers->end() && element_definitions->find(element_number) == element_definitions->end()", file=0x55555afddea8 "/home/satya/WORK/PS-8.0/sql/dd/impl/types/object_table_definition_impl.h", line=115, 
    function=0x55555afde960 <dd::Object_table_definition_impl::add_element(int, std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const&, std::map<std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> >, int, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const, int> > >*, std::map<int, std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> >, std::less<int>, std::allocator<std::pair<int const, std::__cxx11::basic_string<char, std::char_traits<char>, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > > > >*)::__PRETTY_FUNCTION__> "void dd::Object_table_definition_impl::add_element(int, const String_type&, const String_type&, dd::Object_table_definition_impl::Element_numbers*, dd::Object_table_definition_impl::Element_definitions*)") at assert.c:101
#4  0x00005555595f016b in dd::Object_table_definition_impl::add_element (this=0x7fffe04d0ac8, element_number=6, element_name="ENCRYPTION", element_definition="ENCRYPTION='Y'", 
    element_numbers=0x7fffe04d0c50, element_definitions=0x7fffe04d0c80) at /home/satya/WORK/PS-8.0/sql/dd/impl/types/object_table_definition_impl.h:112
#5  0x00005555595f0f86 in dd::Object_table_definition_impl::add_option (this=0x7fffe04d0ac8, option_number=6, option_name="ENCRYPTION", option_definition="ENCRYPTION='Y'")
    at /home/satya/WORK/PS-8.0/sql/dd/impl/types/object_table_definition_impl.h:283
#6  0x00005555597bbad0 in dd::Object_table_impl::set_actual_encrypted (this=0x7fffe04d08b0) at /home/satya/WORK/PS-8.0/sql/dd/impl/types/object_table_impl.cc:72
#7  0x00005555596027a5 in (anonymous namespace)::update_system_tables (thd=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/dd/impl/bootstrap/bootstrapper.cc:168
#8  0x0000555559608f03 in dd::initialize_dd_properties (thd=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/dd/impl/bootstrap/bootstrapper.cc:1356
#9  0x0000555559606c56 in dd::bootstrap::restart (thd=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/dd/impl/bootstrap/bootstrapper.cc:1061
#10 0x000055555980feb5 in dd::upgrade_57::restart_dictionary (thd=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/dd/upgrade_57/upgrade.cc:795
#11 0x00005555598107f9 in dd::upgrade_57::do_pre_checks_and_initialize_dd (thd=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/dd/upgrade_57/upgrade.cc:951
#12 0x000055555885d5df in bootstrap::handle_bootstrap (arg=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/bootstrap.cc:347
#13 0x00005555599cbc54 in pfs_spawn_thread (arg=0x55555cc48d70) at /home/satya/WORK/PS-8.0/storage/perfschema/pfs.cc:2836
#14 0x00007ffff79b66db in start_thread (arg=0x7fffe814f700) at pthread_create.c:463
#15 0x00007ffff5e1d88f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb)

satya-bodapati:
So on restart, dd::Object_table_impl::Object_table_impl() sets encryption property..

> My fear on double encryption is indeed true. I just bumped up DD version to simulate upgrade and it crashed.
> 
> #0 __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
> #1 0x00007ffff5d3c801 in __GI_abort () at abort.c:79
> #2 0x00007ffff5d2c39a in __assert_fail_base (fmt=0x7ffff5eb37d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n",
> assertion=assertion@entry=0x55555afddef8 "element_numbers != nullptr && element_definitions != nullptr && element_numbers->find(element_name) == element_numbers->end() && element_definitions->find(element_number) == element_definitions->end()", file=file@entry=0x55555afddea8 "/home/satya/WORK/PS-8.0/sql/dd/impl/types/object_table_definition_impl.h", line=line@entry=115,
> function=function@entry=0x55555afde960 <dd::Object_table_definition_impl::add_element(int, std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const&, std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const&, std::map<std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> >, int, std::less<std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const, int> > >_, std::map<int, std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> >, std::less, std::allocator<std::pair<int const, std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > > > >_)::**PRETTY_FUNCTION**> "void dd::Object_table_definition_impl::add_element(int, const String_type&, const String_type&, dd::Object_table_definition_impl::Element_numbers*, dd::Object_table_definition_impl::Element_definitions*)") at assert.c:92
> #3 0x00007ffff5d2c412 in __GI___assert_fail (
> assertion=0x55555afddef8 "element_numbers != nullptr && element_definitions != nullptr && element_numbers->find(element_name) == element_numbers->end() && element_definitions->find(element_number) == element_definitions->end()", file=0x55555afddea8 "/home/satya/WORK/PS-8.0/sql/dd/impl/types/object_table_definition_impl.h", line=115,
> function=0x55555afde960 <dd::Object_table_definition_impl::add_element(int, std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const&, std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const&, std::map<std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> >, int, std::less<std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > const, int> > >_, std::map<int, std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> >, std::less, std::allocator<std::pair<int const, std::__cxx11::basic_string<char, std::char_traits, Stateless_allocator<char, dd::String_type_alloc, My_free_functor> > > > >_)::**PRETTY_FUNCTION**> "void dd::Object_table_definition_impl::add_element(int, const String_type&, const String_type&, dd::Object_table_definition_impl::Element_numbers*, dd::Object_table_definition_impl::Element_definitions*)") at assert.c:101
> #4 0x00005555595f016b in dd::Object_table_definition_impl::add_element (this=0x7fffe04d0ac8, element_number=6, element_name="ENCRYPTION", element_definition="ENCRYPTION='Y'",
> element_numbers=0x7fffe04d0c50, element_definitions=0x7fffe04d0c80) at /home/satya/WORK/PS-8.0/sql/dd/impl/types/object_table_definition_impl.h:112
> #5 0x00005555595f0f86 in dd::Object_table_definition_impl::add_option (this=0x7fffe04d0ac8, option_number=6, option_name="ENCRYPTION", option_definition="ENCRYPTION='Y'")
> at /home/satya/WORK/PS-8.0/sql/dd/impl/types/object_table_definition_impl.h:283
> #6 0x00005555597bbad0 in dd::Object_table_impl::set_actual_encrypted (this=0x7fffe04d08b0) at /home/satya/WORK/PS-8.0/sql/dd/impl/types/object_table_impl.cc:72
> #7 0x00005555596027a5 in (anonymous namespace)::update_system_tables (thd=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/dd/impl/bootstrap/bootstrapper.cc:168
> #8 0x0000555559608f03 in dd::initialize_dd_properties (thd=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/dd/impl/bootstrap/bootstrapper.cc:1356
> #9 0x0000555559606c56 in dd::bootstrap::restart (thd=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/dd/impl/bootstrap/bootstrapper.cc:1061
> #10 0x000055555980feb5 in dd::upgrade_57::restart_dictionary (thd=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/dd/upgrade_57/upgrade.cc:795
> #11 0x00005555598107f9 in dd::upgrade_57::do_pre_checks_and_initialize_dd (thd=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/dd/upgrade_57/upgrade.cc:951
> #12 0x000055555885d5df in bootstrap::handle_bootstrap (arg=0x55555cd58d50) at /home/satya/WORK/PS-8.0/sql/bootstrap.cc:347
> #13 0x00005555599cbc54 in pfs_spawn_thread (arg=0x55555cc48d70) at /home/satya/WORK/PS-8.0/storage/perfschema/pfs.cc:2836
> #14 0x00007ffff79b66db in start_thread (arg=0x7fffe814f700) at pthread_create.c:463
> #15 0x00007ffff5e1d88f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
> (gdb)

To get this crash: Apply this PR, get datadir, and then apply this patch
```
diff --git a/sql/dd/dd_version.h b/sql/dd/dd_version.h
index ee1dd0e1d96..216225b958f 100644
--- a/sql/dd/dd_version.h
+++ b/sql/dd/dd_version.h
@@ -176,7 +176,7 @@
 */
 namespace dd {
 
-static const uint DD_VERSION = 80016;
+static const uint DD_VERSION = 80017;
```
Startup on this datadir by 8017 (upcoming merge too may be) will cause failure

satya-bodapati:
Very happy to find a problem that will occur in future :blush:

satya-bodapati:
Next onto fixing this issue

satya-bodapati:
> Next onto fixing this issue

A temporary patch that solves the double encryption issue.
```
diff --git a/sql/dd/impl/bootstrap/bootstrapper.cc b/sql/dd/impl/bootstrap/bootstrapper.cc
index 9b564821dc9..1a1cadab5ae 100644
--- a/sql/dd/impl/bootstrap/bootstrapper.cc
+++ b/sql/dd/impl/bootstrap/bootstrapper.cc
@@ -72,6 +72,7 @@
 #include "sql/mysqld.h"
 #include "sql/sql_zip_dict.h"
 #include "sql/thd_raii.h"
+#include "sql/dd/impl/types/object_table_definition_impl.h"
 
 using namespace dd;
 
@@ -162,9 +163,39 @@ bool update_system_tables(THD *thd) {
         my_error(ER_DD_METADATA_NOT_FOUND, MYF(0), it->first.c_str());
         return true;
       }
+
       std::unique_ptr<Properties> table_def_properties(
           Properties::parse_properties(def));
+
+      String_type options_str;
+      table_def_properties->get("options", &options_str);
+      std::unique_ptr<Properties> opts(
+          Properties::parse_properties(options_str));
+      bool encryption_exists = false;
+      for (Properties::const_iterator it = opts->begin(); it != opts->end();
+           ++it) {
+        String_type label;
+        int pos = 0;
+        String_type def;
+
+        std::unique_ptr<Properties> element(
+            Properties::parse_properties(it->second));
+        if (element->get("lbl", &label) || element->get("pos", &pos) ||
+            element->get("def", &def)) {
+        }
+        fprintf(stderr, "Label is %s\n", label.c_str());
+        fprintf(stderr, "pos is %d\n", pos);
+        fprintf(stderr, "definition is %s\n", def.c_str());
+
+        if (my_strcasecmp(system_charset_info, label.c_str(), "ENCRYPTION") ==
+            0) {
+          encryption_exists = true;
+          break;
+        }
+      }
+
       table_def->set_actual_table_definition(*table_def_properties);
+      if (!encryption_exists) table_def->set_actual_encrypted();
     }
   }
```

satya-bodapati:
I have to check if we can survive without the on-disk change

satya-bodapati:
> I have to check if we can survive without the on-disk change

No, even if I remove dd::update_properties line from this PR, there is another place where dd::update_properties is called.

So it goes to disk by default on upgrades.

So two options infront of us
1. Accept on-disk changes. DD properties of core DD tables will deviate from Upstream (My major concern) :( We will need some kind of special handling for future versions too (to avoid double encryption attribute additions)

2. Since table creation in mysql is hardcoded (all done by server), we can exempt the encryption flag check for tables to be created in mysql tablespace. This is not as bad as it looks :) because it is restricted to tablespace that is used only server and not users. Also upstream is also doing some kind of cheating here.

satya-bodapati:
> > I have to check if we can survive without the on-disk change
> 
> No, even if I remove dd::update_properties line from this PR, there is another place where dd::update_properties is called.
> 
> So it goes to disk by default on upgrades.
> 
> So two options infront of us
> 
>     1. Accept on-disk changes. DD properties of core DD tables will deviate from Upstream (My major concern) :( We will need some kind of special handling for future versions too (to avoid double encryption attribute additions)
> 
>     2. Since table creation in mysql is hardcoded (all done by server), we can exempt the encryption flag check for tables to be created in mysql tablespace. This is not as bad as it looks :) because it is restricted to tablespace that is used only server and not users. Also upstream is also doing some kind of cheating here.

Its kind of pick the poison scenario
1. Clean but deviate from upstream disk format, do extra handling for future version
2. Not clean (somewhat hackish, same as upstream) but compatible to upstream disk format

satya-bodapati:
Other issues with deviating from upgrade on-disk format is that 8.0 provides ability to downgrade until as specific version (like two minor versions?). So such on-disk deviation will prevent downgrades

satya-bodapati:
> > Another thing to find out how the upstream does "ALTER TABLESPACE mysql ENCRYPTION='Y'". Does it add encryption attribute in dd_properties of all tables?
> 
> The answer is NO. Upstream does kind of "cheating". It doesn't treat mysql tablespace as encrypted during the bootstrap phase.

DD properties are not updated but the table options of DD tables in mysql.tables.options are updated.

satya-bodapati:
> > > Another thing to find out how the upstream does "ALTER TABLESPACE mysql ENCRYPTION='Y'". Does it add encryption attribute in dd_properties of all tables?
> > 
> > 
> > The answer is NO. Upstream does kind of "cheating". It doesn't treat mysql tablespace as encrypted during the bootstrap phase.
> 
> DD properties are not updated but the table options of DD tables in mysql.tables.options are updated.

Not sure if this is useful  or not .. Because when DD is booted, the tables are created from DD properties? This is because mysql.tables is not available during bootstrap

mysql-test/t/percona_8_0_15_dd_upgrade_encrypted.test
```diff
@@ -0,0 +1,82 @@
...
+
+--echo # Stop DB server which was created by MTR default
+--source include/shutdown_mysqld.inc
+
+--echo # Start the 8.0.16 server on 8.0.15 datadir
+--let $restart_parameters = "restart: --loose-skip-log-bin --skip-log-slave-updates --datadir=$MYSQLD_DATADIR1 --keyring_file_data=$MYSQL_TMP_DIR/data_8015_dd_encrypted/mysecret_keyring $KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD"
+--replace_result $MYSQLD_DATADIR1 MYSQLD_DATADIR1 $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD KEYRING_PLUGIN_EARLY_LOAD
+--source include/start_mysqld.inc
+
+--let $assert_cond= "[SELECT ENCRYPTION FROM INFORMATION_SCHEMA.INNODB_TABLESPACES WHERE NAME=\\'mysql\\']" = 'Y'
```
satya-bodapati: We only have assert_cond without assert.inc :)

satya-bodapati: Use SET DEBUG='+d, skip_dd_table_access_check' to access DD and verify if all the tables in mysql tablespace have "encryption=y" in options

satya-bodapati: Ideally I would have loved to have a test that verifies dd_props table definitions doesn't have "encryption". Not sure if this is feasible

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1352,7 +1355,9 @@ bool initialize_dd_properties(THD *thd) {
...
```
satya-bodapati: Remove this. 
1. It is not reached. (update_system returns false on success)
2. dd::update_properties is called at later stage of upgrade

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1673,6 +1678,23 @@ bool sync_meta_data(THD *thd) {
   return false;
 }
 
+// Helper guard used in update_properties, to be sure
+// that encryption will get set back before
+// update_properties exits.
+struct Target_encryption_guard {
+  bool set_encryption{false};
```
satya-bodapati: make this private

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1673,6 +1678,23 @@ bool sync_meta_data(THD *thd) {
   return false;
 }
 
+// Helper guard used in update_properties, to be sure
+// that encryption will get set back before
+// update_properties exits.
+struct Target_encryption_guard {
```
satya-bodapati, line 1682: we can use class instead of struct

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1673,6 +1678,23 @@ bool sync_meta_data(THD *thd) {
...
+  bool set_encryption{false};
+  Target_encryption_guard(const Object_table *object_table)
```
satya-bodapati, line 1684: In constructor initialization list: use m_is_encrypted(object_table->is_target_encrypted())

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1692,6 +1714,24 @@ bool update_properties(THD *thd, const std::set<String_type> *create_set,
...
+        also do not want to update dd_properties. Since dd_properties are
+        updated based on target definition we unset the encryption from target
+        definition for the time of updating dd_properties. The exact field that
+        contains system tables properties in dd_properties is SYSTEM_TABLES.
+      */
+      Target_encryption_guard target_encryption_guard((*it)->entity());
+      bool is_target_encrypted = (*it)->entity()->is_target_encrypted();
+      if (is_target_encrypted) {
+        (*it)->entity()->unset_target_encrypted();
+        target_encryption_guard.set_encryption = true;
```
satya-bodapati: No need to set this explicitly.

With constructor member we don't need this

sql/dd/impl/types/object_table_impl.h
```diff
@@ -35,7 +35,7 @@ namespace dd {
 class Object_table_impl : virtual public Object_table {
  protected:
   mutable uint m_last_dd_version;
-  Object_table_definition_impl m_target_def;
+  mutable Object_table_definition_impl m_target_def;
```
satya-bodapati, line 38: Why mutable is required now?

satya-bodapati (changes requested):

mysql-test/t/percona_8_0_15_dd_upgrade_encrypted.test
```diff
@@ -0,0 +1,82 @@
...
+--echo # Set different paths for --datadir
+let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/data_8015_dd_encrypted;
+
+--copy_file $MYSQLTEST_VARDIR/std_data/data_8015_dd_encrypted.zip $MYSQL_TMP_DIR/data_8015_dd_encrypted.zip
+
+--echo # Check that the file exists in the working folder.
+--file_exists $MYSQL_TMP_DIR/data_8015_dd_encrypted.zip
+
+--echo # Unzip the zip file.
+--exec unzip -qo $MYSQL_TMP_DIR/data_8015_dd_encrypted.zip -d $MYSQL_TMP_DIR
```
satya-bodapati, line 59: Can you please recreate this datadir?

After the assert.inc, please add restart and then do the same test

satya-bodapati:
> tablespace
> 
> Upgrade creates tables in mysql tablespaces in two phases:
> 
>     * Phase 1 creates tables based on actual DD properties,
>       taken from dd_properties table, option SYSTEM_TABLES.
>       We add here in-memory encryption flag to CREATE statement
>       if the mysql tablespace is encrypted.
> 
>     * Phase 2 create target tables. We add here in-memory encryption
>       flag to CREATE statement if mysql tablespace is encrypted and update
>       SYSTEM_TABLES option in dd_properties with encryption flag.

Please elaborate the problem clearly and the document the fix

sql/dd/impl/bootstrap/bootstrapper.cc
robgolebiowski: done

sql/dd/impl/bootstrap/bootstrapper.cc
robgolebiowski: done

sql/dd/impl/bootstrap/bootstrapper.cc
robgolebiowski: done

sql/dd/impl/bootstrap/bootstrapper.cc
robgolebiowski: done

sql/dd/impl/bootstrap/bootstrapper.cc
robgolebiowski: done

sql/dd/impl/types/object_table_impl.h
robgolebiowski: Because we want to remove "encryption" option only for the time of updating dd_properties. Once System_tables container is filled with system tables they are only available as cont objects. We create system tables with encryption='y' to satisfy the requirement for encrypted mysql tablespace.

sql/dd/impl/bootstrap/bootstrapper.cc
robgolebiowski: yes, this is what I wrote I need to remove, removed.

mysql-test/t/percona_8_0_15_dd_upgrade_encrypted.test
robgolebiowski: Recreate the default 8.0.16 datadir and run the asserts on it. What is the point ?

mysql-test/t/percona_8_0_15_dd_upgrade_encrypted.test
robgolebiowski: done

robgolebiowski: done

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1673,6 +1676,25 @@ bool sync_meta_data(THD *thd) {
...
 }
 
+// Helper guard used in update_properties, to be sure
+// that encryption will get set back before
+// update_properties exits.
+struct Target_encryption_guard {
+ public:
+  Target_encryption_guard(const Object_table *object_table)
+      : m_object_table(object_table)
+      : set_encryption(object_table->is_target_encrypted) {}
```
satya-bodapati: is_target_encrypted should be called as function right? is_target_encrypted"()"?

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1673,6 +1676,25 @@ bool sync_meta_data(THD *thd) {
...
+  ~Target_encryption_guard() {
+    if (set_encryption) {
+      m_object_table->set_target_encrypted();
+    }
+  }
+
+ private:
+  const Object_table *m_object_table;
+  bool set_encryption{false};
```
satya-bodapati: we can remove this initialization of false here because now it is initialized at constructor

sql/dd/impl/bootstrap/bootstrapper.cc
satya-bodapati: I wonder how this worked at all :scream:

robgolebiowski: yeah, scary, fixed.

sql/dd/impl/bootstrap/bootstrapper.cc
robgolebiowski: done

satya-bodapati (approved):

PR #3392, opened 2019-08-23
PS-5909 : Rotating Innodb Master key post upgrade crashes the server

satya-bodapati:
Problem:
--------
After upgrade from 5.7 server which has encrypted undo tablespaces,
8.0 crashes on ALTER INSTANCE ROTATE INNODB MASTER KEY.

This is because upgrade recreates undo tablespaces. During this process,
old 5.7 undo tablespaces are deleted. The in-memory tablespace object
for them are not deleted. These objects I refer them as orphan or zombie tablespaces

ALTER INSTANCE finds this orphan tablespaces, tries to access page 0 and
cannot get it.

Note that this problem doesn't happen after restart (because after restart
the orphan in-memory objects for undo tablespaces don't exist anymore)

Fix:
----
During upgrade, remove the in-memory tablespace object (fil_space_t) along with physical file
removal of 5.7 undo tablespace.

### DIFF
--- a/share/errmsg-utf8.txt
+++ b/share/errmsg-utf8.txt
@@ -19271,6 +19271,9 @@ ER_PARALLEL_DOUBLEWRITE_WRITE_ERROR
 ER_XB_MSG_6
   eng "Key rotation of encrypted session temporary tablespace %s failed"
 
+ER_XB_UNDO_DELETE_FAILURE
+  eng "Deletion of 5.7 undo tablespace %s failed"
+
 #
 # End of Percona Server 8.0 server error log messages
 #
--- a/mysql-test/r/percona_dd_upgrade_undo_encrypted.result
+++ b/mysql-test/r/percona_dd_upgrade_undo_encrypted.result
@@ -16,6 +16,7 @@ c1	c2
 200	bbbbb
 300	aaaaa
 400	bbbbb
+ALTER INSTANCE ROTATE INNODB MASTER KEY;
 # Now let's test what happens when encryption key is missing and we try to start 8.0 server on 5.7 directory
 # with encrypted undo tablespaces
 # Recreate the directories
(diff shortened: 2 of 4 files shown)


share/errmsg-utf8.txt
```diff
@@ -19271,6 +19271,9 @@ ER_PARALLEL_DOUBLEWRITE_WRITE_ERROR
 ER_XB_MSG_6
   eng "Key rotation of encrypted session temporary tablespace %s failed"
 
+ER_XB_MSG_7
```
laurynas-biveinis: New error messages can try to use meaningful mnemonics (the ER_IB_MSG_1234 existed because it was a mass change as far as I can tell)

laurynas-biveinis (changes requested):
Fix commit message "This I call has ... "

share/errmsg-utf8.txt
satya-bodapati: I see, let me add better name. I thought, it has to be sequence :)

laurynas-biveinis (approved):

satya-bodapati:
> Fix commit message "This I call has ... "

done. Fixed.. fingers didn't listen to mind.. :)

satya-bodapati:
```
./mtr --mem main.percona_dd_upgrade_undo_encrypted 
Logging: /home/satya/WORK/PS-8.0/mysql-test/mysql-test-run.pl  --mem main.percona_dd_upgrade_undo_encrypted
MySQL Version 8.0.16
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Checking leftover processes
Removing old var directory
Creating var directory '/home/satya/WORK/PS-8.0/bld/mysql-test/var'
 - symlinking 'var' to '/dev/shm/var_auto_NEQ9'
Installing system database
Using parallel: 1

[100%] main.percona_dd_upgrade_undo_encrypted    [ pass ]  38519
The servers were reinitialized 1 times
Spent 38.519 of 61 seconds executing testcases

Completed: All 1 tests were successful.

```

PR #3393, opened 2019-08-26
PS-5910: Support V3 encryption header

dutow:
This fixes some 32-64 bit differences, and the 5.7->8.0
redo log encryption upgrade encryption issue.

### DIFF
--- a/mysql-test/suite/innodb/t/table_encrypt_upgrade-master.opt
+++ b/mysql-test/suite/innodb/t/table_encrypt_upgrade-master.opt
@@ -1,0 +1,4 @@
+$KEYRING_PLUGIN_EARLY_LOAD
+$KEYRING_PLUGIN_OPT
+--loose-keyring-file-data=$MYSQL_TMP_DIR/mydummy_key
+
--- a/mysql-test/suite/innodb/t/table_encrypt_upgrade.test
+++ b/mysql-test/suite/innodb/t/table_encrypt_upgrade.test
@@ -1,0 +1,118 @@
+# This test case will test the V2 => V3 encryption header upgrade
+--source include/have_innodb.inc
+--source include/not_embedded.inc
+--source include/have_debug.inc
+
+--disable_query_log
+call mtr.add_suppression("\\[Error\\] InnoDB: Encryption can't find master key, please check the keyring plugin is loaded.");
+call mtr.add_suppression("\\[ERROR\\] Function 'keyring_file' already exists");
+call mtr.add_suppression("\\[ERROR\\] Couldn't load plugin named 'keyring_file' with soname 'keyring_file.*'.");
+call mtr.add_suppression("Plugin keyring_file reported");
+#call mtr.add_suppression("InnoDB: Failed to find tablespace for table");
+--enable_query_log
+
+let $innodb_file_per_table = `SELECT @@innodb_file_per_table`;
+SET GLOBAL innodb_file_per_table = 1;
+
+# Create a table with encryption
+SET DEBUG='+d,force_v2_encryption';
+CREATE TABLE t1(i INT,  c2 char(20)) ENCRYPTION="Y" ENGINE = InnoDB;
+INSERT INTO t1 VALUES (1, "znmcg");
+INSERT INTO t1 VALUES (2, "prtsd");
+INSERT INTO t1 VALUES (3, "vlwfb");
+CREATE TABLE t2(i INT,  c2 char(20)) ENCRYPTION="Y" ENGINE = InnoDB;
+INSERT INTO t2 VALUES (1, "znmcg");
+INSERT INTO t2 VALUES (2, "prtsd");
+INSERT INTO t2 VALUES (3, "vlwfb");
+SET DEBUG='';
+SELECT * FROM t1;
+SELECT * FROM t2;
+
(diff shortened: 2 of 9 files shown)


storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -978,17 +978,23 @@ fsp_header_fill_encryption_info(
...
 		memcpy(ptr, ENCRYPTION_KEY_MAGIC_V2, ENCRYPTION_MAGIC_SIZE);
+	} else {
+		memcpy(ptr, ENCRYPTION_KEY_MAGIC_V3, ENCRYPTION_MAGIC_SIZE);
 	}
 	ptr += ENCRYPTION_MAGIC_SIZE;
 
 	/* Write master key id. */
 	mach_write_to_4(ptr, master_key_id);
-	ptr += sizeof(ulint);
+	if (version == Encryption::ENCRYPTION_VERSION_3) {
```
satya-bodapati, line 990: While writing, there is no need to write V1 or V2. Just write V3 format (No need check for 'version' during write)

Change the entire patch to
1. Write only V3
2. Handle reading of v1, v2 and v3.

satya-bodapati (changes requested):

storage/innobase/fsp/fsp0fsp.cc
dutow: I have to keep at least V1: some encryption features are already GA, and V1 stored keys in the keyring under a different name.
For V2, we would lose some (not 100% complete) downgrade ability within 5.7, that's probably not an issue - but then it's only a few lines of code.

satya-bodapati: So tablespaces that come from older 5.7 continue to use V2?  I don't think we should care about downgrade ability before GA.

Also ALTER INSTANCE INNODB ROTATE MASTER KEY should write encryption headers in V3. Please verify this.

satya-bodapati: any rebuild of the tables would bring the tables to V3 format anyway

dutow: This PR is for 5.7, it is not about the 5.7->8.0 upgrade, it's about how versions within 5.7 behave. And some encryption features are already GA on 5.7.

satya-bodapati: @dutow I understand that. So my questions are
1. ALTERs which rebuild table would recreate the tablespace encryption header in V3 format right? please check and confirm
2. ALTER INSTANCE .. on upgraded tables from earlier version. Would they continue using V2 format after rotation?

dutow: I've added a testcase about this.

mysql-test/suite/innodb/t/table_encrypt_upgrade-master.opt
```diff
@@ -0,0 +1,4 @@
+--early-plugin-load="keyring_file=$KEYRING_PLUGIN"
```
satya-bodapati: use $KEYRING_PLUGIN_EARLY_LOAD

mysql-test/suite/innodb/t/table_encrypt_upgrade.test
```diff
@@ -0,0 +1,61 @@
...
+let SEARCH_PATTERN=lCB;
+--source include/search_pattern.inc
+LET $MYSQLD_DATADIR = `select @@datadir`;
+let SEARCH_PATTERN=lCC;
+--source include/search_pattern.inc
+let SEARCH_PATTERN=znmcg;
+--source include/search_pattern.inc
+
+# Rebuild the table
+ALTER TABLE t1 ENGINE = InnoDB ENCRYPTION="Y";
```
satya-bodapati, line 57: Add one more V2 TABLE. After restart, do ROTATE INNODB MASTER KEY.

check if the IBD has v2 or v3?

storage/innobase/fsp/fsp0fsp.cc
satya-bodapati: Thanks, looks good. added comment on that

satya-bodapati (changes requested):

mysql-test/suite/innodb/t/table_encrypt_upgrade.test
dutow: Done. Also removed the unnecessary restarts.

storage/innobase/fsp/fsp0fsp.cc
satya-bodapati: Now your tests confirm that DDLs (and ALTER INSTANCE) will change all tables to V3 and downgrade is not possible.

I will make this suggestion optional. So upto you if you want to write only V3.

satya-bodapati (approved):

satya-bodapati:
btw, can you bring  the import export test from 8.0? f565013b7f31f225c349a6484ccab8cf4b5c8f52

I've reviewed this patch too at Oracle :)

storage/innobase/fsp/fsp0fsp.cc
dutow: I'll keep this functionality for now - without it, I wouldn't be able to add this test :)

PR #3399, opened 2019-08-29
PS-5923: Accepting SELECT ... INTO as documented

dutow:
Issue: according to the documentation, INTO should be accepted at:

(a) before the FROM clause
(b) before the (optional) FOR clause

But instead of (b) MySQL currently accepts it after the FOR clause.

Fix: we keep the current behavior, but also accept INTO clauses
where the documentation specifies.


sql/sql_yacc.yy
```diff
@@ -444,7 +444,7 @@ void warn_about_deprecated_national(THD *thd)
   1. We do not accept any reduce/reduce conflicts
   2. We should not introduce new shift/reduce conflicts any more.
 */
-%expect 107
+%expect 109
```
satya-bodapati, line 447: the increase 2 indicates?

mysql-test/r/parser.result
```diff
@@ -770,15 +770,15 @@ FROM t1 WHERE 1 GROUP BY 1 HAVING 1 ORDER BY 1 LIMIT 1
...
+ERROR 42000: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(SELECT 1 FROM t1 INTO @var17727401) a' at line 1
 SELECT EXISTS(SELECT 1 FROM t1 INTO @var17727401);
-ERROR 42000: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'INTO @var17727401)' at line 1
+ERROR 42000: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(SELECT 1 FROM t1 INTO @var17727401)' at line 1
 SELECT 1 FROM t1 INTO @var17727401 UNION SELECT 1 FROM t1 INTO t1;
 ERROR 42000: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'UNION SELECT 1 FROM t1 INTO t1' at line 1
 (SELECT 1 FROM t1 INTO @var17727401) UNION (SELECT 1 FROM t1 INTO t1);
-ERROR 42000: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'UNION (SELECT 1 FROM t1 INTO t1)' at line 1
+ERROR HY000: Incorrect usage of UNION and INTO
```
satya-bodapati, line 781: Where does this message come from? Is the error message used before your patch?

satya-bodapati (commented):
Minor comments. Can you please run the new testcase on upstream 8.0 and paste result in comment

mysql-test/t/bug96677.test
```diff
@@ -0,0 +1,46 @@
+CREATE TABLE t1(id INT);
```
satya-bodapati, line 1: please add percona_ prefix to testcase.

Also shouldn't we use percona bug number?

May be write upstream bug number inside the testcase

satya-bodapati (changes requested):

mysql-test/t/bug96677.test
dutow: For upstream bugs, we are using bug<upstream_id> names, as upstream

mysql-test/r/parser.result
dutow: Yes, this is sql_lex.cc:650. It was already used (see some other errors around this location in the parser), but for slightly different queries.

mysql-test/r/bug96677.result
```diff
@@ -0,0 +1,36 @@
+CREATE TABLE t1(id INT);
+INSERT INTO t1 VALUES (1);
+SELECT id AS foobar FROM t1 GROUP BY id LIMIT 1 INTO @id_var1 FOR UPDATE;
```
dutow, line 3: On upstream, this first select fails with ER_SYNTAX_ERROR, so that's all the test does.

sql/sql_yacc.yy
dutow: I've added into_clause to two new places in query_expression. As the commend above select_stmt_with_into explains, this is ambiguous.

sql/sql_yacc.yy
```diff
@@ -9054,6 +9054,17 @@ query_expression:
           {
             $$= NEW_PTN PT_query_expression($1, $2, $3, $4);
           }
+        | query_expression_body
+          opt_order_clause
+          opt_limit_clause
+          into_clause
+          opt_locking_clause_list
+          {
+            if(!$1->set_into($4)) {
```
percona-ysorokin, line 9063: We check here for `!xxx->set_into(...)` (with negation) to return parse error whereas comment for `set_into()` says `true if couldn't set`.

dutow: Fixed, the comment was wrong.

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/310/

percona-ysorokin (approved):
In general the idea with extending `node` hierarchy with a new method `set_into()` that determines whether `INTO` clause is allowed or not looks very elegant. However, even in original Oracle MySQL 8.0 grammar file I see a lot of mess related to resolving `INTO` clause grammar conflicts and I cannot guarantee that we covered all corner cases and that the grammar is still unambiguous.

The patch is fine only for the custom build - it should not get into the trunk unless Oracle decides otherwise in the upstream.

dutow:
" that the grammar is still unambiguous." - into grammar isn't unambiguous even without my patch, as explained in a comment in sql_lex.yy, it's handled by precedence rules.

satya-bodapati (approved):
From my limited knowledge in this area, patch looks safe enough (not adding extra syntaxes or behaviours)

satya-bodapati:
> Issue: according to the documentation, INTO should be accepted at:
> 
> (a) before the FROM clause
> (b) before the (optional) FOR clause
> 
> But instead of (b) MySQL currently accepts it after the FOR clause.
> 
> Fix: we keep the current behavior, but also accept INTO clauses
> where the documentation specifies.

Would be nice if you can put what is allowed in 5.7 and whats not 8.0

With this patch, "the following  SQL .. " works

dutow:
Closing for now as the custom build is done & I submitted the patch to upstream. Will reopen later if needed.

PR #3403, opened 2019-08-30
PS-5933: Better MTR terminal support (5.6)

dutow:
* Introducing the MTR_TERM environment variable, which defaults to xterm.
  This variable can be configured to basically any visual terminal,
  and setting it in .profile will result in a configured default setting
  for the user.
* Intrudocing the MTR_LLDB environment variable, to deal with versioned
  lldb executable names in ubuntu.

### DIFF
--- a/mysql-test/mysql-test-run.pl
+++ b/mysql-test/mysql-test-run.pl
@@ -157,6 +157,24 @@ END {
 
 sub env_or_val($$) { defined $ENV{$_[0]} ? $ENV{$_[0]} : $_[1] }
 
+# set_term_args(user_specified_string, term_exe_variable, term_args_arr, title)
+sub set_term_args {
+  my $term_cmd = $_[0];
+  if ($term_cmd =~ /^ *$/) {
+    mtr_error("MTR_TERM is defined, but empty");
+  }
+  my @term_args = split / /, $term_cmd;
+  $_[1] = shift @term_args;
+  foreach my $t_arg (@term_args) {
+    if ($t_arg eq "%title%") {
+      mtr_add_arg($_[2], "$_[3]");
+    } else {
+      mtr_add_arg($_[2], $t_arg);
+    }
+  }
+
+}
+
 my $path_config_file;           # The generated config file, var/my.cnf
 
 # Visual Studio produces executables in different sub-directories based on the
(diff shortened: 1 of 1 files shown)


mysql-test/mysql-test-run.pl
```diff
@@ -6279,17 +6287,32 @@ sub lldb_arguments {
...
+  my @term_args = split / /, $term_cmd;
+  my $term_exe = shift @term_args;
 
   $$args= [];
-  mtr_add_arg($$args, "-n");
-  mtr_add_arg($$args, "-W");
-  mtr_add_arg($$args, "-a");
-  mtr_add_arg($$args, "Terminal");
-  mtr_add_arg($$args, $lldb_start_file);
+  foreach my $t_arg (@term_args) {
```
satya-bodapati: THis duplication can be avoided?

mysql-test/mysql-test-run.pl
```diff
@@ -6796,6 +6819,21 @@ ($)
                         $opt_max_test_fail, set to 0 for no limit. Set
                         it's default with MTR_MAX_TEST_FAIL
 
+Environment variables controlling debugging parameters
+
+  MTR_TERM              Configures the terminal command to run the debugger.
```
satya-bodapati, line 6815: Comment alignment

satya-bodapati (changes requested):
When upmerging, remove --gterm option/changes

mysql-test/mysql-test-run.pl
```diff
@@ -6228,10 +6228,18 @@ sub gdb_arguments {
      return;
   }
 
+  my $term_cmd = $ENV{MTR_TERM} || "xterm -title %title% -e";
+  my @term_args = split / /, $term_cmd;
+  my $term_exe = shift @term_args;
```
percona-ysorokin: I would also add some handling for the case when `MTR_TERM` is defined but is an empty string (not sure how `split` and `shift` will work in this case).

mysql-test/mysql-test-run.pl
dutow: What's the issue with it? It's aligned as other parts of the help message.

mysql-test/mysql-test-run.pl
dutow: Fixed, I extracted it to a helper function.

mysql-test/mysql-test-run.pl
dutow: Done, it reports an error correctly now.

mysql-test/mysql-test-run.pl
satya-bodapati: For example, I mean the alignment of lines "Configure..", "Defaults too .." , "can also".. doesn't seem to be aligned.

satya-bodapati (approved):
Looks good. OK to push after checking the alignment of lines in help section

mysql-test/mysql-test-run.pl
dutow: Got it, some lines had tabs

PR #3404, opened 2019-08-30
PS-5925 PS-5926 : ut_dbg_assertion_failed (expr=0x55555b75a742 "strlen(srv_uu…

satya-bodapati:
…id) != 0")

PS-5926 : default_table_encryption=ON at bootstrap doesn't encrypt mysql.ibd

PS-5925 happens in very corner case as described below:

Problem:
--------

I could reproduce only with Robert's zip file (data_8015_dd_encrypted.zip) for PS-5874

The magic in his datadir is that there is crash recovery involved and there redo is only on "mysql.ibd"

So the steps are
1. Only mysql.ibd should be encrypted at initialize
2. Do some change only on mysql.ibd (how? insert into some non core tables in mysql tablespace)
3. Ensure that there will recovery only on mysql.ibd (disable page cleaners, log chekpointers etc)
4. Let the recovery proceed
5. Restart server
6. Boom! You will get strlen(srv_uuid) != 0 assert

There is rotation of tablespaces that is done during startup, just after recovery

And during recovery only mysql.ibd is opened which has master key_id and zero uuid
Rotation happens before uuid is intialized, it writes uses master key_id as 1 with zero uuid (the core issue)
After this all tablespaces are booted (from sys_config, the highest found master key id is 2)
Fix empty tablespaces uuid doesn't do anything as master key id is 2.

Fix:
----
Fix UUID of mysql.ibd in innobase_fix_tablespaces_uuid()

PS-5926 : default_table_encryption=ON at bootstrap doesn't encrypt mysql.ibd

Problem:
--------
This varaible is duplicated in InnoDB by srv_default_table_encryption.
srv_default_table_encryption is set using fix_default_table_encryption() handlerton API
This handlerton API is called only after InnoDB is initialized.

mysql.ibd encryption happens as part of InnoDB initialization and the srv_default_table_encryption doesn't
have the correct value.

Fix:
----
Use the server global structure value to decide if mysql.ibd should be encrypted or not at bootstrap

### DIFF
--- a/mysql-test/suite/innodb/t/percona_encrypt_mysql_ibd.test
+++ b/mysql-test/suite/innodb/t/percona_encrypt_mysql_ibd.test
@@ -1,0 +1,45 @@
+let $MYSQLD_BASEDIR= `select @@basedir`;
+--mkdir $MYSQL_TMP_DIR/datadir1
+
+let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/datadir1/data;
+--source include/shutdown_mysqld.inc
+
+--let BOOTSTRAP_SQL=$MYSQL_TMP_DIR/boot.sql
+--let KEYRING_DATA="--keyring_file_data=$MYSQL_TMP_DIR/mysecret_keyring"
+--let ERROR_LOG=$MYSQL_TMP_DIR/percona_encrypt_mysql_ibd_err.log
+
+--echo # create bootstrap file
+write_file $BOOTSTRAP_SQL;
+CREATE DATABASE test;
+EOF
+
+--let BOOTSTRAP_CMD = $MYSQLD_CMD --initialize-insecure --basedir=$MYSQLD_BASEDIR --datadir=$MYSQLD_DATADIR1 --init-file=$BOOTSTRAP_SQL $KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD $KEYRING_DATA > $ERROR_LOG --default_table_encryption=ON 2>&1
+
+--echo # Bootstrap new instance with encrypted mysql.ibd
+--exec $BOOTSTRAP_CMD
+
+--echo # Check Encryption header of mysql.ibd
+--let IBD_FILE=$MYSQLD_DATADIR1/mysql.ibd
+--source include/ibd_encryption_check.inc
+
+--let $restart_hide_args=1
+--let $restart_parameters="restart: --datadir=$MYSQLD_DATADIR1"
+--source include/start_mysqld.inc
+
+--let $assert_text = Mysql.ibd should be encrypted
+--let $assert_cond = "[SELECT ENCRYPTION FROM INFORMATION_SCHEMA.INNODB_TABLESPACES WHERE NAME = \"mysql\"]" = "Y"
+--source include/assert.inc
+
+--echo # Check Encryption header of mysql.ibd after shutdown
+--source include/shutdown_mysqld.inc
+
+--let IBD_FILE=$MYSQLD_DATADIR1/mysql.ibd
+--source include/ibd_encryption_check.inc
+
+--echo # Start default MTR instance
+--let $restart_parameters=
+--source include/start_mysqld.inc
+
+--remove_file $BOOTSTRAP_SQL
+--force-rmdir $MYSQL_TMP_DIR/datadir1
+--remove_file $ERROR_LOG
--- a/storage/innobase/dict/dict0dict.cc
+++ b/storage/innobase/dict/dict0dict.cc
@@ -160,6 +160,8 @@ const char *dict_sys_t::s_file_per_table_name = "innodb_file_per_table";
 const char *dict_sys_t::s_default_undo_space_name_1 = "innodb_undo_001";
 const char *dict_sys_t::s_default_undo_space_name_2 = "innodb_undo_002";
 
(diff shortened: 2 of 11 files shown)


mysql-test/suite/innodb/t/percona_encrypt_mysql_ibd.test
```diff
@@ -0,0 +1,44 @@
...
+
+--echo # Bootstrap new instance with encrypted mysql.ibd
+--exec $BOOTSTRAP_CMD
+
+--echo # Check Encryption header of mysql.ibd
+--let IBD_FILE=$MYSQLD_DATADIR1/mysql.ibd
+--source include/ibd_encryption_check.inc
+
+--let $restart_parameters="restart: --datadir=$MYSQLD_DATADIR1"
+--source include/start_mysqld.inc
```
dutow, line 27: Disable printing the restart parameters, the absolute path is written to the result file.

mysql-test/suite/innodb/t/percona_encrypt_mysql_ibd.test
```diff
@@ -0,0 +1,44 @@
...
+
+--let $assert_text = Mysql.ibd should be encrypted
+--let $assert_cond = "[SELECT ENCRYPTION FROM INFORMATION_SCHEMA.INNODB_TABLESPACES WHERE NAME = \"mysql\"]" = "Y"
+--source include/assert.inc
+
+--echo # Check Encryption header of mysql.ibd after restart
```
dutow: After shutdown? The server isn't started again.

storage/innobase/dict/dict0dict.cc
```diff
@@ -7486,8 +7488,18 @@ ONLINE_TO_KEYRING*
 @return true if innodb_encrypt_tables is equal to ON or FORCE or
 ONLINE_TO_KEYRING* */
 static bool dict_should_be_keyring_encrypted() {
-  return srv_default_table_encryption == DEFAULT_TABLE_ENC_ON ||
-         srv_default_table_encryption == DEFAULT_TABLE_ENC_ONLINE_TO_KEYRING;
+  /* We cannot use srv_default_encryption here because it is
```
dutow, line 7488: This sounds like it won't work with a persisted value?

storage/innobase/dict/dict0dict.cc
```diff
@@ -7486,8 +7488,18 @@ ONLINE_TO_KEYRING*
 @return true if innodb_encrypt_tables is equal to ON or FORCE or
```
dutow: Please also update this comment, looks outdated.

dutow (commented):
Please also mention both ticket ids in the first line of the commit message - I think jira only associates pull requests to tickets based on the first line.

mysql-test/suite/innodb/t/percona_encrypt_mysql_ibd.test
satya-bodapati: There is start of the default MTR server

--echo # Start default MTR instance
--let $restart_parameters=

dutow: Yes, but that's a different server with a different datadir - the ibd_encryption_check below this line is about DATADIR1, which will never be started after this shutdown.

storage/innobase/dict/dict0dict.cc
satya-bodapati: Good question. Is the persist framework not available before InnoDB initialization?

dutow: Yes, it is initialized later. I've had an issue with this and redo log encryption (persistent variables were set too late for that), that's why I disabled persisting it. I'm not 100% sure, but based on the comment you wrote here, this could be an issue here too.

satya-bodapati: Then it looks like, there will be issue here too. I will test and file a bug or may be ask @mohitj1988 to do ;)

storage/innobase/dict/dict0dict.cc
satya-bodapati: Done. Good catch

mysql-test/suite/innodb/t/percona_encrypt_mysql_ibd.test
satya-bodapati: Hmm, Got you point now. I should have written 'after shutdown' instead of 'after restart'. will do it next round.

storage/innobase/dict/dict0dict.cc
satya-bodapati: I just realized that default_table_encryption to encrypt mysql.ibd is only at bootstrap time. So there shouldn't be issue

satya-bodapati: subsequent restart do not depend on the variable and they depend on mysql.ibd page 0

dutow: Okay, that shouldn't be an issue then.

satya-bodapati: Great! will tick this issue as resolved..

dutow (approved):
The restart comment is still there in the test, other than that looks good.

satya-bodapati:
> The restart comment is still there in the test, other than that looks good.

Yes, that I will fix before push

Thank you @dutow

PR #3406, opened 2019-09-03
PS-5783: New thread variables: innodb_(force_index_)records_in_range (5.7)

dutow:
These variables make it possible to override reconds_in_range, effectively
skipping its computation. This could be useful in some cases where reconds_in_range
takes up most of the query cost.


george-lorch:
missing sys_vars tests?

george-lorch (approved):

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/184/

PR #3409, opened 2019-09-04
PS-5933: Better MTR terminal support (8.0)

dutow:
* Introducing the MTR_TERM environment variable, which defaults to xterm.
  This variable can be configured to basically any visual terminal,
  and setting it in .profile will result in a configured default setting
  for the user.
* Intrudocing the MTR_LLDB environment variable, to deal with versioned
  lldb executable names in ubuntu.


mysql-test/mysql-test-run.pl
```diff
@@ -1505,7 +1523,6 @@ sub command_line_setup {
     'gdb'                  => \$opt_gdb,
     'gdb-secondary-engine' => \$opt_gdb_secondary_engine,
     # For using gnome-terminal when using --gdb option
```
satya-bodapati: Remove this

satya-bodapati (approved):

PR #3413, opened 2019-09-05
PS-5932: Implementing --tokudb-force-recovery=6 to skip reading the logs

dutow:
This could be useful when the server crashed with a corrupted rollback file,
and is unable to start up. Specifying --tokudb--force-recovery=6 --super-read-only
should start it up in a read-only, but usable state. Some data may be lost and
unrecoverable.

Starting the server without the read only option is NOT supported.

#### Additional notes:

* needs https://github.com/percona/PerconaFT/pull/430
* tested with the mtr test case added in this PR
* tested with sysbench: a read_only run after startup with --tokudb-force-recovery=6

### DIFF
--- a/mysql-test/suite/tokudb/r/recovery.result
+++ b/mysql-test/suite/tokudb/r/recovery.result
@@ -1,0 +1,31 @@
+create table t (a int, b int, primary key (a)) engine=tokudb;
+insert into t values (1,2),(2,4),(3,8);
+optimize table t;
+Table	Op	Msg_type	Msg_text
+test.t	optimize	note	Table does not support optimize, doing recreate + analyze instead
+test.t	optimize	status	OK
+# restart
+insert into t values (4,9),(10,11),(12,13);
+select * from t;
+a	b
+1	2
+2	4
+3	8
+4	9
+10	11
+12	13
+# Kill the server
+# restart:--tokudb-force-recovery=6 --read-only --super-read-only
+select * from t;
+a	b
+1	2
+2	4
+3	8
+# restart:--tokudb-force-recovery=6 --read-only --super-read-only
+select * from t;
+a	b
+1	2
+2	4
+3	8
+# restart
+drop table t;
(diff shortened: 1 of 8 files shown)


mysql-test/suite/tokudb/r/recovery.result
```diff
@@ -0,0 +1,31 @@
...
+10	11
+12	13
+# Kill the server
+# restart:--tokudb-force-recovery=6 --read-only --super-read-only
+select * from t;
+a	b
+1	2
+2	4
+3	8
+# restart:--tokudb-force-recovery=6 --read-only --super-read-only
```
george-lorch, line 24: Can we test the behavior if --tokudb-force-recovery=6 and --read-only --super-read-only is not specified?  What happens?  If --read-only --super-read-only are not specified and --tokudb-force-recovery=6 is specified, I think we should fail to initialize the plugin.

george-lorch:
Once https://github.com/percona/PerconaFT/pull/430 lands, you should also amend this commit to update the git submodule commit pointer.

mysql-test/suite/tokudb/r/recovery.result
dutow: Okay, I'll add that. (We open the tables in read only mode anyway with force recovery, so my guess would be that we get errors elsewhere in the code)

george-lorch:
One more minor thing, it would be good to see jenkins results of 5.6 and 8.0 with the new FT changes to ensure compatibility even though 5.6 (and 8.0?) will not break due to sharing the changes.

dutow:
Update:
* no longer WIP
* updated submodule
* added cherry pick reference to 5.6 commit
* disabled the test for now: it fails when ran together with other tests, I'll fix that later (it works on jenkins when only this test is ran)

george-lorch (approved):

george-lorch (approved):

PR #3415, opened 2019-09-05
ps-5.7-5941

patrickbirch:
apt_repo - deleted wheezy, trusty, artful and added debian buster


dutow:
@patrickbirch why did you close this? You can rebase on the GCA point and force-push the merge from 5.6 with the same PR.

PR #3416, opened 2019-09-05
PS-5931: Allow temporary tables without PKs with sql_require_primary_key

dutow:
The original implementation disables craeting temporary tables with
primary keys. However, the reasoning behind the variable (row based/mixed
replication) is invalid for them, as temporary tables are never
replicated in this mode.

This change alows temporary tables without primary keys even when
sql_require_primary_key is ON.


satya-bodapati (commented):
any mtr tests that cover this change?

dutow:
Specifcally, no - I can extend one. But without this change, check-testcase fails now that I reverted that change in it.

satya-bodapati:
> Specifcally, no - I can extend one. But without this change, check-testcase fails now that I reverted that change in it.

Good to have one explicitly because checktestcase is mtr behaviour and can be disabled with option

dutow:
I've updated the create test.

satya-bodapati (approved):

PR #3418, opened 2019-09-06
Ps 8.0 5427 document the correct distribution upgrade process for percona server

patrickbirch:


dutow:
The 8.0 docs needs no change? The jira ticket mentions that 8.0 is affected.

patrickbirch:
Updated with changes to 8.0. Thank you.

dutow:
This is a different change than the 5.6 and 5.7 PRs, and I don't see that part in the 8.0 docs. Is this really the same issue?

doc/source/.res/replace.txt
```diff
@@ -12,6 +12,7 @@
 
 .. |debian| replace:: Debian
 .. |debian-last| replace:: Debian 9.0 (stretch)
+.. |debian-tenth| replace:: Debian 10.0 (buster)
```
hrvojem, line 15: if we're already changing this should we change last -> ninth?

hrvojem (changes requested):
This change doesn't look to be related to distribution upgrade docs

doc/source/.res/replace.txt
patrickbirch: Sure.

patrickbirch:
duplicate pull request - closing

PR #3420, opened 2019-09-11
PS-5629: Fixed handling of persisted bool variables set as integer with persist_only option.

kamil-holubicki:
https://jira.percona.com/browse/PS-5629

Fixed handling of integer passed as persisted variable value with persist_only option. Such value was handled as string literal for bool and rejected because only allowed string literals are ON/OFF.


kamil-holubicki:
https://ps80.cd.percona.com/job/percona-server-8.0-param/316/

kamil-holubicki:
I'm closing this PR as patch has been submitted to the upstream bug.

PR #3422, opened 2019-09-11
PS-5932: Update PerconaFT submodule (5.6)

dutow:
Also fixes the use of permission parameters where it is actually used now.


dutow:
https://ps56.cd.percona.com/job/percona-server-5.6-param/50/

george-lorch (approved):

PR #3424, opened 2019-09-11
ps-5.7-5754-add-limitation-undo_ts-encryption-cannot-be-disabled

patrickbirch:
Added a note below the innodb_undo_log_encrypt variable definition. The note says you cannot disable encryption.

### DIFF
--- a/doc/source/management/data_at_rest_encryption.rst
+++ b/doc/source/management/data_at_rest_encryption.rst
@@ -437,7 +437,16 @@ System variables
    :vartype: Boolean
    :default: ``Off``
 
-Enables the encryption of InnoDB Undo tablespaces
+Enables the encryption of InnoDB Undo tablespaces. You can enable encryption and
+disable encryption while the server is running. 
+
+.. note:: 
+
+    If you enable undo log encryption, the server writes encryption information
+    into the header. That information stays in the header during the life of the
+    undo log. If you restart the server, the server will try to load the
+    encryption key from the keyring during startup. If the keyring is not available, the server
+    cannot start.
 
 Binary log encryption
 ================================================================================


doc/source/management/data_at_rest_encryption.rst
```diff
@@ -437,7 +437,12 @@ System variables
    :vartype: Boolean
    :default: ``Off``
 
-Enables the encryption of InnoDB Undo tablespaces
+Enables the encryption of InnoDB Undo tablespaces.
+
+.. note::
+
+    After you have encrypted the undo log, you cannot disable the undo log encryption.
```
dutow: You can disable it, and if you do, newly written undo pages won't be encrypted. The setting is dynamic, you can turn it on and off while the server is running.

The limitation is that the encryption info stays in the header, and the server will try to load it during startup. If it can't do it (e.g. no keyring plugin is loaded), it can't start.

dutow (changes requested):

dutow:
Please also include the jira ticket id (ps-5754) as a separate word in the commit title (so jira can link the PR to the issue), and use spaces instead of `-`.

doc/source/management/data_at_rest_encryption.rst
```diff
@@ -437,7 +437,16 @@ System variables
...
-Enables the encryption of InnoDB Undo tablespaces
+Enables the encryption of InnoDB Undo tablespaces. You can enable encryption and
+disable encryption while the server is running. 
+
+.. note:: 
+
+    If you enable undo log encryption, the server writes encryption information
+    into the header. That information stays in the header during the life of the
+    undo log. If you restart the server, the server will try to load the keyring
+    encryption during startup. If the keyring is not available, the server
```
dutow: "the server will try to load the keyring encryption during startup" - this doesn't make much sense. Maybe "the server will try to load the encryption key from the keyring during startup"?

dutow (approved):

PR #3425, opened 2019-09-12
PS-5629:Fixed handling of persisted bool and enum variables set as integer with persist_only option.

kamil-holubicki:
https://jira.percona.com/browse/PS-5629

If the user sets bool/enum value providing alias iteger value, it is stored in mysqld-auto.cnf file. Then when value is read on server start, it was considered to be string literal and validated against allowed bool/enum string literals which caused value rejection and server not to start.

### DIFF
--- a/mysql-test/r/percona_persist_only_bool_enum.result
+++ b/mysql-test/r/percona_persist_only_bool_enum.result
@@ -1,0 +1,161 @@
+#
+# Changing values
+#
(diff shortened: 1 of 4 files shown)


kamil-holubicki:
https://ps80.cd.percona.com/job/percona-server-8.0-param/319/

kamil-holubicki:
I'm closing this PR as patch has been submitted to the upstream bug.

PR #3428, opened 2019-09-16
PS-5939: Null merge PS-5.7.26 into 8.0

inikep:
https://jira.percona.com/browse/PS-5939
    
***
    Added 'main.bug93388' MTR test case for the Bug #28986737 / #93388
    "Renaming and replacing mysql.user table can lead to a server crash"
    fixed in 5.5.64, 5.6.44, 5.7.26, 8.0.16
    (commit mysql/mysql-server@e02123c).
    A damaged mysql.user table could cause a server exit.
    
***
    Removed 'mysql-test/suite/encryption/t/.swp' file added by mistake
    in one of the previous commits.    
    
***
    Reverted our fix for PS-4570
    "Flush status; adds twice to global values"
    (https://jira.percona.com/browse/PS-4570)
    (commit 9b66cba)
    in favor of the upstream fix for the bug #28291258 / #91541
    "'Flush status' statement adds twice to global values"
    (https://bugs.mysql.com/bug.php?id=91541)
    (commit mysql/mysql-server@a900a05).
    Only our MTR test case 'perfschema.bug91541' was removed with this patch.
    
***
    Fixed regression in 'main.join_file_handler' MTR test case.
    Reverted removal of the 'reinit_io_cache()' call inside 'Unique::get()' in
    'sql/uniques.cc'. This call was originally removed as unnecessary as part of
    the PS-3947
    "Refactor IO_CACHE layer to allow custom read/write function (Prerequisite for temporary file encryption)"
    (https://jira.percona.com/browse/PS-3947)
    implementation
    (commit f8157f3)
    but became an issue after Oracle fixed Bug #28039829 / #90902
    "Select Query With Complex Joins Leaks File Handles"
    (https://bugs.mysql.com/bug.php?id=90902)
    (commit mysql/mysql-server@d009f1a).


inikep:
https://ps80.cd.percona.com/job/percona-server-8.0-param/325/

percona-ysorokin (approved):
LGTM

PR #3429, opened 2019-09-16
PS-5546

rahulmalik87:
Empty compression dictionaries should be disallowed

CREATE COMPRESSION_DICTIONARY numbers ('');
This should be disallowed because we cannot later update
the compression dictionary to a valid text.

### DIFF
--- a/sql/handler.h
+++ b/sql/handler.h
@@ -712,6 +712,7 @@ enum handler_create_zip_dict_result
                                           exists */
   HA_CREATE_ZIP_DICT_NAME_TOO_LONG,  /*!< zip dict name is too long */
   HA_CREATE_ZIP_DICT_DATA_TOO_LONG,  /*!< zip dict data is too long */
+  HA_CREATE_ZIP_DICT_DATA_ZERO_LEN,  /*!< zip dict data is zero len */
   HA_CREATE_ZIP_DICT_READ_ONLY,      /*!< cannot create in read-only mode */
   HA_CREATE_ZIP_DICT_OUT_OF_MEMORY,  /*!< out of memory */
   HA_CREATE_ZIP_DICT_OUT_OF_FILE_SPACE,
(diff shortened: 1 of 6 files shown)


sql/share/errmsg-utf8.txt
```diff
@@ -7792,6 +7792,9 @@ ER_COMPRESSION_DICTIONARY_NAME_TOO_LONG
 ER_COMPRESSION_DICTIONARY_DATA_TOO_LONG
   eng "Data for compression dictionary '%-.192s' is too long (max length = %lu)"
 
+ER_COMPRESSION_DICTIONARY_DATA_ZERO_LEN
+  eng "Data for compression dictionary can't be of zero lenght"
```
satya-bodapati, line 7796: length typo

satya-bodapati: Also add this one as last error message (else it would error number for other codes)

satya-bodapati: Have a dictionary name passed to it, print the name of dictionary

sql/handler.h
```diff
@@ -712,6 +712,7 @@ enum handler_create_zip_dict_result
                                           exists */
   HA_CREATE_ZIP_DICT_NAME_TOO_LONG,  /*!< zip dict name is too long */
   HA_CREATE_ZIP_DICT_DATA_TOO_LONG,  /*!< zip dict data is too long */
+  HA_CREATE_ZIP_DICT_DATA_ZERO_LEN,  /*!< zip dict data is zero len */
```
satya-bodapati, line 715: We don't need this. We can check at Server layer and return error from there directly

sql/sql_zip_dict.cc
```diff
@@ -77,6 +78,10 @@ int mysql_create_zip_dict(THD* thd, const char* name, ulong name_len,
         error= ER_COMPRESSION_DICTIONARY_NAME_TOO_LONG;
         my_error(error, MYF(0), name, local_name_len);
         break;
+      case HA_CREATE_ZIP_DICT_DATA_ZERO_LEN:
```
satya-bodapati, line 81: Remove this entire block

satya-bodapati: Add the length check at mysql_create_zip_dict()

satya-bodapati (changes requested):

satya-bodapati:
> Empty compression dictionaries should be disallowed
> 
> CREATE COMPRESSION_DICTIONARY numbers ('');
> This should be disallowed because we cannot later update
> the compression dictionary to a valid text.

Your commit msg should have
"PS-5546: Empty compression dictionaries should be disallowed" in first line

PR #3430, opened 2019-09-16
PS-3411: Prevent creation of xb_doublewrite file when innodb_doublewrite is disabled (5.7)

kamil-holubicki:
https://jira.percona.com/browse/PS-3411

Unnecessary xb_doublewrite file is no more created when innodb_doublewrite option is disabled.

### DIFF
--- a/mysql-test/suite/innodb/t/percona_doublewrite.test
+++ b/mysql-test/suite/innodb/t/percona_doublewrite.test
@@ -138,6 +138,62 @@ SET GLOBAL innodb_fast_shutdown=2;
 --let $restart_parameters=restart:--innodb-force-recovery=6
 --source include/start_mysqld.inc
 
+--echo #
+--echo # Bug PS-3411: Parallel doublewrite buffer file created when skip-innodb_doublewrite is set
+--echo #
+
+--let $restart_hide_args= 1
+--let $restart_parameters= restart:--innodb-parallel-doublewrite-path=$DOUBLEWRITE_FILE
+--source include/restart_mysqld.inc
+file_exists $DOUBLEWRITE_FILE;
+
+CREATE TABLE t1 (a INT);
+
+--let $i= 1000
+--disable_query_log
+START TRANSACTION;
+while($i)
+{
+    --eval INSERT INTO t1 VALUES ($i);
+    --dec $i
+}
+COMMIT;
+--enable_query_log
+
+# wait until pages flushed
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-param/163/

mysql-test/suite/innodb/t/percona_doublewrite.test
```diff
@@ -138,6 +138,64 @@ SET GLOBAL innodb_fast_shutdown=2;
...
+--echo # Bug PS-3411: Parallel doublewrite buffer file created when skip-innodb_doublewrite is set
+--echo #
+
+--let $restart_hide_args= 1
+--let $restart_parameters= restart:--innodb-parallel-doublewrite-path=$DOUBLEWRITE_FILE
+--source include/restart_mysqld.inc
+file_exists $DOUBLEWRITE_FILE;
+
+--disable_warnings
+DROP TABLE IF EXISTS t1;
```
dutow: Code like this is present in some old tests, but it's not used anymore: if it exists, drop it (without if exists, without disabling warnings), if it doesn't, delete this block.

mysql-test/suite/innodb/t/percona_doublewrite.test
```diff
@@ -138,6 +138,64 @@ SET GLOBAL innodb_fast_shutdown=2;
...
+while($i)
+{
+    --eval INSERT INTO t1 VALUES ($i);
+    --dec $i
+}
+COMMIT;
+--enable_query_log
+
+# flush is done once a second so wait to be sure that we test what we need
+--real_sleep 3
```
dutow: Sleep (even real_sleep) can be unreliable on jenkins. It's better to use wait_condition / debug sync points, depending on what's the easiest to do for a given situation. For page flushes, you probably want somethning like

```
# wait until pages flushed
--let $wait_condition= SELECT variable_value = 0 FROM performance_schema.global_status WHERE LOWER(variable_name) = 'innodb_buffer_pool_pages_dirty'
--source include/wait_condition.inc
```

dutow (changes requested):

mysql-test/suite/innodb/t/percona_doublewrite.test
```diff
@@ -138,6 +138,64 @@ SET GLOBAL innodb_fast_shutdown=2;
...
+
+SELECT * FROM t1 LIMIT 10;
```
dutow: Either remove this select (is it really part of the test?), or add an order by

dutow (approved):

PR #3431, opened 2019-09-16
PS-3411: Prevent creation of xb_doublewrite file when innodb_doublewrite is disabled (8.0)

kamil-holubicki:
https://jira.percona.com/browse/PS-3411

Unnecessary xb_doublewrite file is no more created when innodb_doublewrite option is disabled.


dutow (changes requested):
Same comments as for 5.7.

kamil-holubicki:
The same fix as for 5.7.
https://ps80.cd.percona.com/job/percona-server-8.0-param/330/

dutow (approved):

PR #3433, opened 2019-09-17
PS-5813: Fixed handling of 'none' value for slow_query_log_use_global_control server variable (5.6)

kamil-holubicki:
https://jira.percona.com/browse/PS-5813

slow_query_log_use_global_control variable was implemented as Sys_var_set class. The purpose of 'none' value is to clear the set, but for Sys_var_set it is done by providing empty value. Specialized class implemented that detects 'none' literal.

### DIFF
--- a/sql/sys_vars.cc
+++ b/sql/sys_vars.cc
@@ -4414,7 +4414,45 @@ void init_log_slow_verbosity()
 {
   update_slow_query_log_use_global_control(0,0,OPT_GLOBAL);
 }
-static Sys_var_set Sys_slow_query_log_use_global_control(
+
+/**
+  Specialized class that handles "none" value of
+  slow_query_log_use_global_control_set variable.
+  When "none" only value is detected, it is rewriten to empty
+  causing set to be cleared.
+*/
+class Sys_var_set_none: public Sys_var_set {
+public:
+  Sys_var_set_none(const char *name_arg,
+        const char *comment, int flag_args, ptrdiff_t off, size_t size,
+        CMD_LINE getopt, const char *values[], ulonglong def_val, PolyLock *lock =
+        0, enum binlog_status_enum binlog_status_arg = VARIABLE_NOT_IN_BINLOG,
+        on_check_function on_check_func = 0,
+        on_update_function on_update_func = 0, const char *substitute = 0)
+    : Sys_var_set(name_arg, comment, flag_args, off, size, getopt, values,
+          def_val, lock, binlog_status_arg, on_check_func, on_update_func,
+          substitute)
+  {
+  }
+
+  virtual bool do_check(THD *thd, set_var *var)
+  {
+    if (var->value->result_type() == STRING_RESULT) {
+      char buff[STRING_BUFFER_USUAL_SIZE];
+      String str(buff, sizeof(buff), system_charset_info);
+
+      String *res = var->value->val_str(&str);
+      if (res
+          && (res->length() > 0)
+          && (0 == my_strcasecmp(system_charset_info, res->ptr(), "none"))) {
+        var->value = new Item_string("", 0, system_charset_info);
+      }
+    }
+    return Sys_var_set::do_check(thd, var);
+  }
+};
+
+static Sys_var_set_none Sys_slow_query_log_use_global_control(
        "slow_query_log_use_global_control",
        "Choose flags, wich always use the global variables. Multiple flags allowed in a comma-separated string. [none, log_slow_filter, log_slow_rate_limit, log_slow_verbosity, long_query_time, min_examined_row_limit, all]",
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://ps56.cd.percona.com/job/percona-server-5.6-param/52/

mysql-test/t/percona_slow_query_log_use_global_control.test
```diff
@@ -25,6 +25,23 @@ SET SESSION query_exec_time=default;
 
 --source include/log_stop.inc
 
+--echo #
+--echo # Bug PS-5813: Setting 'none' value for slow_query_log_use_global_control throwing error.
```
dutow: Can't this go into `mysql-test/suite/sys_vars/t/slow_query_log_use_global_control_basic.test`? That's where basic value validation tests are usually placed.

sql/sys_vars.cc
```diff
@@ -4414,7 +4414,43 @@ void init_log_slow_verbosity()
...
 }
-static Sys_var_set Sys_slow_query_log_use_global_control(
+
+/**
+  Specialized class that handles "none" value of
+  slow_query_log_use_global_control_set variable.
+  When "none" only value is detected, it is rewriten to empty
+  causing set to be cleared.
+*/
+class Sys_var_set_none: public Sys_var_set {
```
dutow, line 4424: Shouldn't this go into the header, where other classes like this are placed?

sql/sys_vars.cc
```diff
@@ -4414,7 +4414,43 @@ void init_log_slow_verbosity()
...
+  {
+  }
+
+  virtual bool do_check(THD *thd, set_var *var)
+  {
+    char buff[STRING_BUFFER_USUAL_SIZE];
+    String str(buff, sizeof(buff), system_charset_info), *res;
+
+    if ((var->value->result_type() == STRING_RESULT)
+        && (res = var->value->val_str(&str))
```
dutow: Please don't do an assignment in the middle of an if, it's easy to miss and confusing.

sql/sys_vars.cc
```diff
@@ -4414,7 +4414,43 @@ void init_log_slow_verbosity()
...
+        && (res->length() > 0)
+        && (0 == my_strcasecmp(system_charset_info, res->ptr(), "none"))) {
+      var->value = new Item_string("", 0, system_charset_info);
```
dutow: This won't result in leaking the old or the new value?

dutow (changes requested):
See my specific comments - but are you sure about the fix and the explanation? To me, it looks like the main issue is that the `"none"` string isn't mentioned in the `slow_query_log_use_global_control_name` array. None should be an allowed value, and it should be the default value, instead of the empty string, and the code using the variable should handle it.

Otherwise, the documentation should be changed instead of the code - as even with your fix, the value shows up as the empty string for selects, not as none.

edit: I missed that this is a set, so allowing none there would require additional validation too (so it can't be mixed with other values), and probably doesn't make much sense - the empty string is simply better here. But I'm still not sure if this is a good idea.

dutow:
According to the current docs, values like `none,log_slow_filter` are valid too, which doesn't make any sense. The documentation definitely needs changes.

sql/sys_vars.cc
kamil-holubicki: No, the old (and new) value of Item_string will be deleted automatically when connection is closed. All 'Items' when created are added to the list of objects allocated in Query Arena object.

sql/sys_vars.cc
kamil-holubicki: I think no. This specialization is only for this particular variable and there is no need to make this type containing fix public.
Of course we can imagine more sophisticated general purpose class like this, that allows us to provide list of possible values that should behave like 'none' like vector<string> = {'none', 'null', 'nothing'}. 
But in this case I would keep it here because there is no need to make it public.

mysql-test/t/percona_slow_query_log_use_global_control.test
kamil-holubicki: Done

sql/sys_vars.cc
kamil-holubicki: Done

kamil-holubicki:
As discussed on Slack:
1. adding 'none' to the set is possible form technical point of view, but there is too much hassle after that. All already existing values will change their int aliases. Probably more adjustments in code that acts according to these values will be needed. As the consequence the whole functionality would need to be fully retested.
2. As per spec, user can set 'none' as parameter value. This fix that introduces customized handling of 'none' value is the less intrusive solution.
3. Another 'fix' would be just adjusting the documentation by removing 'none' from allowed values.

https://ps56.cd.percona.com/job/percona-server-5.6-param/56/

dutow (approved):

kamil-holubicki:
https://ps56.cd.percona.com/job/percona-server-5.6-param/69/

PR #3438, opened 2019-09-19
PS-5761: gcc-9 compilation warnings (5.6)

inikep:
https://jira.percona.com/browse/PS-5761

Backport fixes from 5.7 for gcc-9 compilation warnings

### DIFF
--- a/sql/sp_rcontext.h
+++ b/sql/sp_rcontext.h
@@ -144,7 +144,7 @@ public:
       memcpy(sql_state, _sql_state, SQLSTATE_LENGTH);
       sql_state[SQLSTATE_LENGTH]= '\0';
 
-      strncpy(message, _message, MYSQL_ERRMSG_SIZE - 1);
+      strncpy(message, _message, MYSQL_ERRMSG_SIZE);
       message[MYSQL_ERRMSG_SIZE - 1]= '\0';
     }
   };
--- a/sql/sql_list.h
+++ b/sql/sql_list.h
@@ -52,6 +52,14 @@ public:
     next= elements ? tmp.next : &first;
   }
 
+  SQL_I_List &operator=(const SQL_I_List &tmp)
+  {
+    elements= tmp.elements;
+    first= tmp.first;
+    next= tmp.elements ? tmp.next : &first;
+    return *this;
+  }
+
   inline void empty()
   {
     elements= 0;
@@ -168,11 +176,20 @@ public:
     relies on this behaviour. This logic is quite tricky: please do not use
     it in any new code.
   */
-  inline base_list(const base_list &tmp) :Sql_alloc()
+  base_list(const base_list &tmp)
+    : Sql_alloc(),
+      first(tmp.first),
+      last(tmp.elements ? tmp.last : &first),
+      elements(tmp.elements)
   {
+  }
+  base_list &operator=(const base_list &tmp)
+  {
+    Sql_alloc::operator= (tmp);
     elements= tmp.elements;
     first= tmp.first;
     last= elements ? tmp.last : &first;
+    return *this;
   }
   /**
     Construct a deep copy of the argument in memory root mem_root.
@@ -471,6 +488,10 @@ template <class T> class List :public base_list
 public:
   inline List() :base_list() {}
   inline List(const List<T> &tmp) :base_list(tmp) {}
+  List &operator=(const List &tmp)
+  {
+    return static_cast<List &>(base_list::operator=(tmp));
+  }
   inline List(const List<T> &tmp, MEM_ROOT *mem_root) :
     base_list(tmp, mem_root) {}
   /*
--- a/storage/tokudb/CMakeLists.txt
+++ b/storage/tokudb/CMakeLists.txt
@@ -82,6 +82,13 @@ endmacro(append_cflags_if_supported)
 set_cflags_if_supported(-Wno-missing-field-initializers)
 append_cflags_if_supported(-Wno-vla)
 
+# Disable warnings for gcc-9 or higher
+IF(CMAKE_COMPILER_IS_GNUCXX AND
+    (CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 9.0 OR
+     CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 9.0))
+  append_cflags_if_supported(-Wno-address-of-packed-member)
+ENDIF()
+
 IF (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/PerconaFT/")
     IF (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/ft-index/")
         MESSAGE(FATAL_ERROR "Found both PerconaFT and ft-index sources.  Don't know which to use.")
--- a/unittest/gunit/CMakeLists.txt
+++ b/unittest/gunit/CMakeLists.txt
@@ -191,12 +191,15 @@ SET(GMOCK_INCLUDE_DIRS
   CACHE INTERNAL "")
 
 # Build gmock/gtest libraries.
-INCLUDE_DIRECTORIES(
+INCLUDE_DIRECTORIES(SYSTEM
   ${GMOCK_SOURCE_DIR}
   ${GMOCK_SOURCE_DIR}/include
   ${GTEST_SOURCE_DIR}
   ${GTEST_SOURCE_DIR}/include
 )
+
+MY_CHECK_CXX_COMPILER_FLAG("-Wformat-overflow" HAVE_FORMAT_OVERFLOW)
+
 ADD_LIBRARY(gmock STATIC ${GMOCK_SOURCE_DIR}/src/gmock-all.cc)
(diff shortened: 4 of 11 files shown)


inikep:
https://ps56.cd.percona.com/job/percona-server-5.6-param/55/

storage/tokudb/CMakeLists.txt
```diff
@@ -82,6 +82,13 @@ endmacro(append_cflags_if_supported)
 set_cflags_if_supported(-Wno-missing-field-initializers)
 append_cflags_if_supported(-Wno-vla)
 
+# Disable warnings for gcc-9 or higher
+IF(CMAKE_COMPILER_IS_GNUCXX AND
+    (CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 9.0 OR
+     CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 9.0))
+  append_cflags_if_supported(-Wno-address-of-packed-member)
```
dutow, line 89: Why not `MY_CHECK_CXX_COMPILER_FLAG`?

unittest/gunit/CMakeLists.txt
```diff
@@ -317,6 +322,13 @@ IF (MERGE_UNITTESTS)
   SET_SOURCE_FILES_PROPERTIES(MERGE_SMALL_TESTS MERGE_LARGE_TESTS
     PROPERTIES GENERATED 1)
 
+  # Suppress -Wformat-overflow= for my_safe_snprintf(" %s ", nullptr)
+  IF(HAVE_FORMAT_OVERFLOW)
+    ADD_COMPILE_FLAGS(
+      ${MERGE_LARGE_TESTS}
+      COMPILE_FLAGS "-Wno-format-overflow")
```
dutow, line 329: With `set_source_files_properties(segfault-t.cc PROPERTIES COMPILE_FLAGS -Wno-format-overflow)`, you don't have to duplicate this.

sql/sql_list.h
```diff
@@ -52,6 +52,14 @@ class SQL_I_List :public Sql_alloc
     next= elements ? tmp.next : &first;
   }
 
+  SQL_I_List &operator=(const SQL_I_List &tmp)
```
dutow, line 55: These operators / copy constructors doesn't really work together with other parts of this class... not that we can do anything about this, as looks like they are used already :(

unittest/gunit/CMakeLists.txt
inikep: It's backported from 5.7 so we may keep it just to minimize a diff.

storage/tokudb/CMakeLists.txt
inikep: `append_cflags_if_supported` is shorter than:
```
MY_CHECK_CXX_COMPILER_FLAG("-Wno-address-of-packed-member" HAVE_ADDRESS_OF_PACKED_MEMBER)

IF(HAVE_ADDRESS_OF_PACKED_MEMBER)  
  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-address-of-packed-member")
  SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-address-of-packed-member")
ENDIF()
```

dutow: I missed that `append_cflags_if_supported` is also conditional. Then why is the compiler version check needed?

dutow: (my concern here is that if this currently isn't an issue with clang, it will be in the future)

inikep: Currently we have `-Waddress-of-packed-member` warnings only with `gcc-9`. I didn't want to disable these warnings for `clang` and older `gcc` because these warnings may be useful in the future in another code/places.

sql/sp_rcontext.h
```diff
@@ -144,7 +144,7 @@ class sp_rcontext : public Sql_alloc
       memcpy(sql_state, _sql_state, SQLSTATE_LENGTH);
       sql_state[SQLSTATE_LENGTH]= '\0';
 
-      strncpy(message, _message, MYSQL_ERRMSG_SIZE - 1);
+      strncpy(message, _message, MYSQL_ERRMSG_SIZE);
```
percona-ysorokin, line 147: What was the warning here? I am curious only because this type of warnings we usually fix with decreasing the number of bytes being copied.

sql/sql_alloc.h
```diff
@@ -44,15 +44,6 @@ class Sql_alloc
   static void operator delete[](void *ptr, MEM_ROOT *mem_root)
   { /* never called */ }
   static void operator delete[](void *ptr, size_t size) { TRASH(ptr, size); }
-#ifdef HAVE_purify
```
percona-ysorokin: I believe this fragment was originally meant to overcome some Valgrind limitation (not being able to handle empty base classes properly, I guess). What was the warning here?

sql/sql_list.h
percona-ysorokin: My thinking is that before these changes assignment operator was already there (a default generated one with bitwise copying). So, having it here defined explicitly shouldn't do any harm. Moreover, the initialization of the `next` member is handled better in this case.

sql/sql_list.h
```diff
@@ -168,11 +176,19 @@ class base_list :public Sql_alloc
...
+  base_list(const base_list &tmp)
+    : Sql_alloc(),
+      first(tmp.first),
+      last(tmp.elements ? tmp.last : &first),
+      elements(tmp.elements)
+  {
+  }
+  base_list &operator=(const base_list &tmp)
   {
     elements= tmp.elements;
```
percona-ysorokin, line 189: I would also add base class assignment operator call
`Sql_alloc::operator = (*this);`

unittest/gunit/CMakeLists.txt
```diff
@@ -247,6 +250,8 @@ ADD_DEFINITIONS(-DMYSQL_SERVER)
 ADD_DEFINITIONS(-DERRMSG_DIR="${PROJECT_BINARY_DIR}/sql/share")
 ADD_DEFINITIONS(-DDATA_DIR="${CMAKE_CURRENT_BINARY_DIR}")
 
+INCLUDE(${MYSQL_CMAKE_SCRIPT_DIR}/compile_flags.cmake)
```
percona-ysorokin, line 253: Shouldn't this be included before `MY_CHECK_CXX_COMPILER_FLAG("-Wformat-overflow" HAVE_FORMAT_OVERFLOW)`?

percona-ysorokin (changes requested):

sql/sp_rcontext.h
inikep:
```
In file included from /usr/include/string.h:494,
                 from /data/mysql-server/PS-5761-5.6/include/m_string.h:34,
                 from /data/mysql-server/PS-5761-5.6/sql/sql_plugin.h:34,
                 from /data/mysql-server/PS-5761-5.6/sql/structs.h:23,
                 from /data/mysql-server/PS-5761-5.6/sql/unireg.h:170,
                 from /data/mysql-server/PS-5761-5.6/sql/sp_rcontext.cc:17:
In function ‘char* strncpy(char*, const char*, size_t)’,
    inlined from ‘sp_rcontext::Sql_condition_info::Sql_condition_info(uint, const char*, Sql_condition::enum_warning_level, const char*)’ at /data/mysql-server/PS-5761-5.6/sql/sp_rcontext.h:147:14,
    inlined from ‘sp_rcontext::Handler_call_frame::Handler_call_frame(const sp_handler*, uint, const char*, Sql_condition::enum_warning_level, const char*, uint)’ at /data/mysql-server/PS-5761-5.6/sql/sp_rcontext.h:185:31,
    inlined from ‘bool sp_rcontext::handle_sql_condition(THD*, uint*, const sp_instr*)’ at /data/mysql-server/PS-5761-5.6/sql/sp_rcontext.cc:400:54:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:106:34: error: ‘char* __builtin_strncpy(char*, const char*, long unsigned int)’ output may be truncated copying 511 bytes from a string of length 511 [-Werror=stringop-truncation]
```

inikep: The change is safe as the next line is `message[MYSQL_ERRMSG_SIZE - 1]= '\0';`

sql/sql_alloc.h
inikep: Currently I can't reproduce a warning. 
I may removed it because it was removed in 5.7 by upstream:
https://github.com/mysql/mysql-server/commit/955e335b5729a7d74e88f00e6bd46b88a0682670

unittest/gunit/CMakeLists.txt
inikep: `MY_CHECK_CXX_COMPILER_FLAG` is defined in `./CMakeLists.txt`, not in `compile_flags.cmake`

sql/sql_alloc.h
inikep: Changes reverted.

sql/sql_list.h
inikep: done

percona-ysorokin (approved):
LGTM

PR #3441, opened 2019-09-19
PS-5178 Document upgrade of innodb-encrypt-tables=ON instance

patrickbirch:
Added section in the innodb_encrypt_tables settings about the upgrade from 5.7

### DIFF
--- a/doc/source/management/data_at_rest_encryption.rst
+++ b/doc/source/management/data_at_rest_encryption.rst
@@ -532,6 +532,10 @@ tables.
    |MariaDB| Documentation: ``innodb_encrypt_tables`` Option
       https://mariadb.com/kb/en/library/xtradbinnodb-server-system-variables/#innodb_encrypt_tables
 
+.. note:: 
+
+If the 5.7 version has encrypted MySQL schema tables due to ``--innodb-encrypt-tables`` ="ON" during bootstrap or if ``--innodb_encrypt-tables`` is present at the upgrade time, the create mysql tablespace is encrypted during upgrade.
+
       ..variable:: table_encryption_privilege_check
 
         :cli: table-encryption-privilege-check


doc/source/management/data_at_rest_encryption.rst
```diff
@@ -532,6 +532,10 @@ tables.
    |MariaDB| Documentation: ``innodb_encrypt_tables`` Option
       https://mariadb.com/kb/en/library/xtradbinnodb-server-system-variables/#innodb_encrypt_tables
 
+.. note:: 
+
+If the 5.7 version has encrypted MySQL schema tables due to ``--innodb-encrypt-tables`` ="ON" during bootstrap or if ``--innodb_encrypt-tables`` is present at the upgrade time, the create mysql tablespace is encrypted during upgrade.
```
dutow, line 537: There is no "create mysql tablespace" - the jira ticket means that the mysql tablespace is created as encrypted.

It also should be `innodb-encrypt-tables`, or `innodb_encrypt_tables`, consistently, not mixed.

`innodb-encrpyt-tables` also doesn't exist anymore in 8.0. The first mention (5.7 bootstrap) is still true), but the second (8.0 upgrade) isn't anymore. @satya-bodapati According to upstream docs, the new `default_table_encryption` doesn't encrypt the mysql tablespace, which means now only the first option is possible?

dutow (changes requested):

doc/source/management/data_at_rest_encryption.rst
patrickbirch: Yes, I wasn't sure how 8.0 would handle a 5.7 upgrade.

patrickbirch:
Close this pull request - the information was included in the encryption doc upgrade.

PR #3442, opened 2019-09-20
PS-5738: Decrease size of KEYRING encryption information on page0

robgolebiowski:
Decreased Keyring encryption information stored on page0 of KEYRING
encrypted tables.

Removed:
 - space_id (4 bytes)
 - offset (2 bytes)
 - tablespace_iv (32 bytes)
Shrank the size of encryption_rotation from 4 bytes to 1 byte
(currently this can have only one of two values: NO_ROTATION or
MASTER_KEY_TO_KEYRING)

tablespace_iv was only used when there was MASTER_KEY_TO_KEYRING
rotation taking place. We would have two IVs then - one for KEYRING
encryption and the second one for MK decryption. However, those IVs do
not need to be different as long as the key used is different - which is
the case here. So instead of generating new IV for KEYRING encryption we
reuse the MK encryption IV when re-encrypting from MK to KEYRING.

### DIFF
--- a/storage/innobase/fil/fil0crypt.cc
+++ b/storage/innobase/fil/fil0crypt.cc
@@ -119,15 +119,12 @@ EncryptionKeyId get_global_default_encryption_key_id_value();
 #define DEBUG_KEYROTATION_THROTTLING 0
 
 static constexpr uint KERYING_ENCRYPTION_INFO_MAX_SIZE =
-    ENCRYPTION_MAGIC_SIZE + 2  // length of iv
-    + 4                        // space id
-    + 2                        // offset
-    + 1                        // type
+    ENCRYPTION_MAGIC_SIZE + 1  // type
     + 4                        // min_key_version
     + 4                        // key_id
     + 1                        // encryption
     + CRYPT_SCHEME_1_IV_LEN    // iv (16 bytes)
-    + 4                        // encryption rotation type
+    + 1                        // encryption rotation type
     + ENCRYPTION_KEY_LEN       // tablespace key
     + ENCRYPTION_KEY_LEN;      // tablespace iv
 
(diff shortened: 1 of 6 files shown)


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/328/

storage/innobase/include/os0file.h
```diff
@@ -338,7 +338,7 @@ struct Encryption {
     KEYRING = 2
   };
 
-  enum Encryption_rotation { NO_ROTATION, MASTER_KEY_TO_KEYRING };
+  enum class Encryption_rotation : char { NO_ROTATION, MASTER_KEY_TO_KEYRING };
```
dutow: unsigned char or std::uint8_t would be better

You could also move it outside of Encryption, that would make the value references shorter.

storage/innobase/include/os0file.h
```diff
@@ -360,27 +360,25 @@ struct Encryption {
...
 
   /** Specific constructor
   @param[in]	type		Algorithm type */
   explicit Encryption(Type type)
       : m_type(type),
-        m_key(NULL),
+        m_key(nullptr),
         m_klen(0),
         m_key_allocated(false),
         m_iv(NULL),
```
dutow: nullptr

storage/innobase/include/os0file.h
```diff
@@ -954,8 +943,8 @@ class IORequest {
     m_encryption.set_key(NULL, 0, false);
     m_encryption.m_iv = NULL;
```
dutow: nullptr twice, and below

storage/innobase/fil/fil0crypt.cc
```diff
@@ -575,21 +547,16 @@ void fil_space_crypt_t::write_page0(
   memcpy(encrypt_info_ptr, iv, CRYPT_SCHEME_1_IV_LEN);
   encrypt_info_ptr += CRYPT_SCHEME_1_IV_LEN;
 
-  mach_write_to_4(encrypt_info_ptr, current_encryption_rotation);
-  encrypt_info_ptr += 4;
+  mach_write_to_1(encrypt_info_ptr,
+                  static_cast<char>(current_encryption_rotation));
```
dutow: `static_cast<byte>`

dutow (changes requested):

storage/innobase/include/os0file.h
robgolebiowski: done

storage/innobase/include/os0file.h
robgolebiowski: done

storage/innobase/include/os0file.h
robgolebiowski: done

storage/innobase/fil/fil0crypt.cc

dutow (approved):

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/346/

satya-bodapati:
this has been reverted on 8.0.17 merge branch

PR #3444, opened 2019-09-27
Post-push PS-5772 fix: ose default master key for undo when needed

dutow:
Issue: when undo tablespaces were enrypted during normal startup,
the master key was generated without an UUID. This was a regression
of Bug#29006275, caused by a revert PS-5772.

Fix: during undo log creation, use the default master key if the
server has no uuid yet.


dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/337/

dutow:
closing in favor of a different fix

PR #3448, opened 2019-10-01
PS-5932: Implementing --tokudb-force-recovery=6 to skip reading the logs

dutow:
This could be useful when the server crashed with a corrupted rollback file,
and is unable to start up. Specifying --tokudb--force-recovery=6 --super-read-only
should start it up in a read-only, but usable state. Some data may be lost and
unrecoverable.

Starting the server without the read only option is NOT supported.

(this is a backport from 5.7, including all bugfixes)


george-lorch (approved):

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/463/#showFailuresLink + all_persisted_vars rerecorded

PR #3449, opened 2019-10-02
PS-5819: Information which tablespace is excluded from encryption

robgolebiowski:
threads should be visible via I_S.

Explicit ENCRYPTION='N' is still used to mark tablespace as to be
excluded from encryption threads. i.e. ALTER/CREATE TABLE/TABLESPACE
ENCRYPTION=N will mean that TABLE/TABLESPACE should be excluded from
encryption threads.

The information if TABLE/TABLESPACE is excluded from encryption
threads is visible through I_S table: INNODB_TABLESPACE_ENCRYPTION,
field EXCLUDED.
Due to changes in 8.0.16 we will no longer rely on
SHOW CREATE TABLE to check if ENCRYPTION=’N’ was explicitly
specified.

### DIFF
--- a/mysql-test/suite/encryption/t/encrypt_threads_skip_tablespace.test
+++ b/mysql-test/suite/encryption/t/encrypt_threads_skip_tablespace.test
@@ -1,0 +1,207 @@
+--source include/have_debug.inc
+
+--replace_result $MYSQL_TMP_DIR MYSQL_TMP_DIR
+eval SET @@global.keyring_file_data="$MYSQL_TMP_DIR/mysecret_keyring";
+
+CREATE TABLE t1 (a varchar(255)) ENCRYPTION='KEYRING';
+CREATE TABLE t2 (a varchar(255));
+CREATE TABLE t3 (a varchar(255)) ENCRYPTION='N';
+CREATE TABLE t4 (a varchar(255)) ENCRYPTION='KEYRING' ENCRYPTION_KEY_ID=5;
+
+CREATE TABLESPACE ts_encrypted ADD DATAFILE 'ts_encrypted.ibd' ENCRYPTION="Y";
+CREATE TABLE t5 (a varchar(255)) TABLESPACE ts_encrypted ENCRYPTION="Y";
+
+CREATE TABLESPACE ts_not_encrypted ADD DATAFILE 'ts_not_encrypted.ibd' ENCRYPTION="N";
+CREATE TABLE t6 (a varchar(255)) TABLESPACE ts_not_encrypted;
+
+CREATE TABLESPACE ts_with_encryption_key_id ADD DATAFILE 'ts_with_encryption_key_id.ibd'  ENCRYPTION_KEY_ID=2;
+CREATE TABLE t7 (a varchar(255)) TABLESPACE ts_with_encryption_key_id;
+
+CREATE TABLE t8 (a varchar(255)) ENCRYPTION_KEY_ID=3;
+
+--let $assert_text= TABLE t3 should have EXCLUDED set to Y in INNODB_TABLESPACES_ENCRYPTION
+--let $assert_cond= "[SELECT EXCLUDED FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'test/t3\\']" = "Y"
+--source include/assert.inc
+
+--let $assert_text= TABLE ts_not_encrypted should have EXCLUDED set to Y in INNODB_TABLESPACES_ENCRYPTION
+--let $assert_cond= "[SELECT EXCLUDED FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'ts_not_encrypted\\']" = "Y"
+--source include/assert.inc
+
+--let $assert_text=All the other tablespaces should have EXCLUDED set to N
+--let $assert_cond= "[SELECT COUNT(*) FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name NOT IN(\\'ts_not_encrypted\\',\\'test/t3\\') AND EXCLUDED = \\'Y\\']" = 0
+--source include/assert.inc
+
+# Try EXCLUDE t1 from rotation while encryption threads are running on it - should result in error
+SET GLOBAL debug="+d,hang_on_ts_with_encryption_key_id_rotation";
+
+SET GLOBAL default_table_encryption=ONLINE_TO_KEYRING;
+# CURRENT_KEY_ID=10 is a marker that we have reached the hang_on_ts_with_encryption_key_id_rotation
+# Can be only one encryption threads - in other case artifical CURRENT_KEY_ID=10 will cause asserts to fail
+SET GLOBAL innodb_encryption_threads=1;
+
+--let $wait_timeout= 600
+--let $wait_condition=SELECT COUNT(*)=1 FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE CURRENT_KEY_ID=10
+--source include/wait_condition.inc
+
+# Now let's try to exlude ts_encrypted while there are encryption threads running on it.
+# It should fail.
+--error ER_EXCLUDE_ENCRYPTION_THREADS_RUNNING
+ALTER TABLESPACE ts_with_encryption_key_id ENCRYPTION='N';
+
+SET GLOBAL debug="-d,hang_on_ts_with_encryption_key_id_rotation";
+
+# Wait for CURRENT_KEY_ID to be restored
+--let $wait_timeout= 600
+--let $wait_condition=SELECT COUNT(*)=0 FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE CURRENT_KEY_ID=10
+--source include/wait_condition.inc
+
+SET GLOBAL innodb_encryption_threads=4;
+
+insert t1 values (repeat('foobarsecret', 12));
+insert t2 values (repeat('tempsecret', 12));
+insert t3 values (repeat('dummysecret', 12));
+insert t4 values (repeat('verysecret', 12));
+insert t5 values (repeat('moresecret', 12));
+insert t6 values (repeat('sooosecret', 12));
+insert t7 values (repeat('notsosecret', 12));
+insert t8 values (repeat('uuuuusecret', 12));
+
+--let tables_count=`select count(*) from INFORMATION_SCHEMA.INNODB_TABLESPACES`
+--echo # Wait max 10 min for key encryption threads to encrypt all spaces
+--let $wait_timeout= 600
+# - 3 we do not encrypt t3, ts_not_encrypted and temporary tablespace
+--let $wait_condition=SELECT COUNT(*) = $tables_count - 3 FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION = 1
+--source include/wait_condition.inc
+
+# Make sure are EXLCUDED flags stays the same after encryption
+
+--let $assert_text= TABLE ts3 should have EXCLUDED set to Y in INNODB_TABLESPACES_ENCRYPTION
+--let $assert_cond= "[SELECT EXCLUDED FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'test/t3\\']" = "Y"
+--source include/assert.inc
+
+--let $assert_text= TABLE ts_not_encrypted should have EXCLUDED set to Y in INNODB_TABLESPACES_ENCRYPTION
+--let $assert_cond= "[SELECT EXCLUDED FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'ts_not_encrypted\\']" = "Y"
+--source include/assert.inc
+
+--let $assert_text=All the other tablespaces should have EXCLUDED set to N
+--let $assert_cond= "[SELECT COUNT(*) FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name NOT IN(\\'ts_not_encrypted\\',\\'test/t3\\') AND EXCLUDED=\\'Y\\']" = 0
+--source include/assert.inc
+
+# It should not be possible to EXCLUDE tablespace encrypted by encryption threads
+--error ER_EXPLICIT_DECRYPTION_OF_ONLINE_ENCRYPTED_TABLESPACE
+ALTER TABLESPACE ts_encrypted ENCRYPTION='N';
+--error ER_EXPLICIT_DECRYPTION_OF_ONLINE_ENCRYPTED_TABLESPACE
+ALTER TABLESPACE ts_with_encryption_key_id ENCRYPTION='N';
+
+SET GLOBAL default_table_encryption=ONLINE_FROM_KEYRING_TO_UNENCRYPTED;
+
+# Make sure that EXLCUDED flags stay the same after decryption
+
+--let tables_count=`select count(*) from INFORMATION_SCHEMA.INNODB_TABLESPACES`
+--echo # Wait max 10 min for key encryption threads to decrypt spaces
+--let $wait_timeout= 600
+# 2 - the ones with explicit KEYRING encryption: t1,t4 stay encrypted
+--let $wait_condition=SELECT COUNT(*) = 2 FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION = 1
+--source include/wait_condition.inc
+
+--let $assert_text= TABLE t3 should have EXCLUDED set to Y in INNODB_TABLESPACES_ENCRYPTION
+--let $assert_cond= "[SELECT EXCLUDED FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'test/t3\\']" = "Y"
+--source include/assert.inc
+
+--let $assert_text= TABLE ts_not_encrypted should have EXCLUDED set to Y in INNODB_TABLESPACES_ENCRYPTION
+--let $assert_cond= "[SELECT EXCLUDED FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'ts_not_encrypted\\']" = "Y"
+--source include/assert.inc
+
+--let $assert_text=All the other tablespaces should have EXCLUDED set to N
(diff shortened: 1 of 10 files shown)


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/351/

storage/innobase/fil/fil0crypt.cc
```diff
@@ -836,6 +836,65 @@ static inline void fil_crypt_read_crypt_data(fil_space_t *space) {
...
+          page_id_t(space->id, 0), page_size_t(space->flags), RW_X_LATCH, NULL,
+          Page_fetch::NORMAL, __FILE__, __LINE__, &mtr)) {
+    space->crypt_data->write_page0(
+        space, block->frame, &mtr, space->crypt_data->min_key_version,
+        space->crypt_data->type, space->crypt_data->encryption_rotation);
+  }
+  mtr.commit();
+}
+
+struct MutexGuard {
```
satya-bodapati: Dont we already have IB_mutex_guard ?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -836,6 +836,65 @@ static inline void fil_crypt_read_crypt_data(fil_space_t *space) {
   mtr.commit();
 }
 
+/**
+Write crypt data belonging to space to page0
+@param space tablespace with crypt data to write
```
satya-bodapati, line 843: @param[in]

storage/innobase/fil/fil0crypt.cc
```diff
@@ -836,6 +836,65 @@ static inline void fil_crypt_read_crypt_data(fil_space_t *space) {
...
+*/
+static void fil_crypt_write_crypt_data_to_page0(fil_space_t *space) {
```
satya-bodapati, line 845: const fil_space_t*

storage/innobase/fil/fil0crypt.cc
```diff
@@ -836,6 +836,65 @@ static inline void fil_crypt_read_crypt_data(fil_space_t *space) {
...
+  mtr_t mtr;
+  mtr.start();
+
+  if (buf_block_t *block = buf_page_get_gen(
+          page_id_t(space->id, 0), page_size_t(space->flags), RW_X_LATCH, NULL,
+          Page_fetch::NORMAL, __FILE__, __LINE__, &mtr)) {
+    space->crypt_data->write_page0(
```
satya-bodapati, line 852: this looks weird. why do we need to pass the space->crypt_data parameters again?

storage/innobase/handler/i_s.cc
```diff
@@ -7562,56 +7562,62 @@ static ST_FIELD_INFO innodb_tablespaces_encryption_fields_info[] = {
      STRUCT_FLD(field_flags, MY_I_S_MAYBE_NULL), STRUCT_FLD(old_name, ""),
      STRUCT_FLD(open_method, 0)},
 
-#define TABLESPACES_ENCRYPTION_ENCRYPTION_SCHEME 2
+#define TABLESPACES_ENCRYPTION_EXCLUDED 2
```
satya-bodapati, line 7563: Better to add this as last parameter

storage/innobase/include/fil0crypt.h
```diff
@@ -379,6 +379,15 @@ class redo_log_keys final {
 
 extern redo_log_keys redo_log_key_mgr;
 
+/**
+Exclude tablespace from encryption threads rotation
+@param space tablespace to exclude
```
satya-bodapati: @param[in]

storage/innobase/include/fil0crypt.h
```diff
@@ -379,6 +379,15 @@ class redo_log_keys final {
...
+@return false tablespace cannot be excluded because there are encryption
+              threads currently operating on it.
+        true  success
+*/
+bool fil_crypt_exclude_tablespace_from_rotation(fil_space_t *space);
```
satya-bodapati, line 390: const fil_space_t*

satya-bodapati (changes requested):
Basic review.

Can you do the failure analysis for the jenkins link?
1. List of failures -> that are old, new, random (ideally new should be none :))
2. add outputs of testcases added by this commit (that new tests pass)

storage/innobase/fil/fil0crypt.cc

storage/innobase/fil/fil0crypt.cc
robgolebiowski: done

storage/innobase/fil/fil0crypt.cc
robgolebiowski: done

storage/innobase/fil/fil0crypt.cc
robgolebiowski: Because sometimes we want to first update page0 before we update the in-memory crypt_data struct. Please have a look at fil_crypt_flush_space and the place where it is called in fil_crypt_complete_rotate_space. We update crypt_data structure only when write to page0 was successful.

storage/innobase/handler/i_s.cc
robgolebiowski: what is the reason behind this? My reasoning was - when users check whole content of the INNODB_TABLESPACES_ENCRYPTION he first want to check if the table is excluded - to know if he should be checking any of the other columns. As it makes sense only for the tables that are not excluded.

storage/innobase/include/fil0crypt.h
robgolebiowski: It cannot be const as I modify space->crypt_data in this function.

robgolebiowski:
@satya-bodapati - there are no MTR tests failures.

satya-bodapati:
@robgolebiowski It would be better if excluded column shows 'Y' or 'N' instead of 1 and 0.

satya-bodapati:
> @satya-bodapati - there are no MTR tests failures.

Well, there will be some failures from Jenkins. Please paste list of failing tests here and then say they are failing on trunk too. (Paste the link of trunk job as well)

mysql-test/suite/encryption/t/encrypt_threads_skip_tablespace.test
```diff
@@ -0,0 +1,207 @@
...
+--let $assert_text= TABLE ts_not_encrypted should have EXCLUDED set to 1 in INNODB_TABLESPACES_ENCRYPTION
+--let $assert_cond= "[SELECT EXCLUDED FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'ts_not_encrypted\\']" = 1
+--source include/assert.inc
+
+--let $assert_text=All the other tablespaces should have EXCLUDED set to 0
+--let $assert_cond= "[SELECT MAX(EXCLUDED) FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name NOT IN(\\'ts_not_encrypted\\',\\'test/t3\\')]" = 0
+--source include/assert.inc
+
+# It should not be possible to EXCLUDE tablespace encrypted by encryption threads
+--error ER_EXPLICIT_DECRYPTION_OF_ONLINE_ENCRYPTED_TABLESPACE
```
satya-bodapati, line 91: This looks a limitation imposed. Any reason to not allow?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -836,6 +836,65 @@ static inline void fil_crypt_read_crypt_data(fil_space_t *space) {
...
+
+ private:
+  ib_mutex_t *mutex;
+};
+
+bool fil_crypt_exclude_tablespace_from_rotation(fil_space_t *space) {
+  MutexGuard fil_crypt_threads_mutex_guard(&fil_crypt_threads_mutex);
+
+  if (space->crypt_data == nullptr) {
+    space->crypt_data = fil_space_create_crypt_data(FIL_ENCRYPTION_OFF,
```
satya-bodapati, line 863: You shouldn't assign space->crypt_data until it is written to page 0 and commited and flushed. I don't know how you are ensuring that page 0 is flushed?
So like at other place, first use crypt_data variable (without assigning space->crypt_data)
then call fil_crypt_write_data_to_page0(space)
finally assign space->crypt_data to crypt_data variable

storage/innobase/fil/fil0crypt.cc
satya-bodapati: I read this code, and it is wrong to do this way. See below comment. "You shouldn't assign.."

storage/innobase/fil/fil0crypt.cc
satya-bodapati: why u wrote "done" when this entire piece is same as before? I still this duplicate MutexGuard

satya-bodapati (changes requested):

satya-bodapati:
Please check compilation errors reported by travis ci. Not sure if they are related to your changes.

/home/travis/build/percona/percona-server/storage/perfschema/pfs_global.cc:150:5: error: adding 'int' to a string does not append to the string [-Werror,-Wstring-plus-int]

    log_errlog(WARNING_LEVEL, ER_PFS_MALLOC_ARRAY_OVERFLOW, n, size,

    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

/home/travis/build/percona/percona-server/sql/log.h:749:36: note: expanded from macro 'log_errlog'

              LOG_ITEM_SRV_SUBSYS, LOG_SUBSYSTEM_TAG, LOG_ITEM_SRC_LINE, \

                                   ^~~~~~~~~~~~~~~~~

/home/travis/build/percona/percona-server/include/my_basename.h:55:33: note: expanded from macro 'LOG_SUBSYSTEM_TAG'

  basename_prefix_eval(__FILE__ + \

satya-bodapati:
This one looks related your patch

/home/travis/build/percona/percona-server/storage/innobase/fil/fil0crypt.cc:810:71: error: extra ';' after member function definition [-Werror,-Wextra-semi]

  MutexGuard(ib_mutex_t *mutex) : mutex(mutex) { mutex_enter(mutex); };

storage/innobase/fil/fil0crypt.cc
robgolebiowski: In all places crypt_data is first assigned to space before calling write_page0. There is even an assert in write_page0: ut_ad(this == space->crypt_data);. If we want to change that maybe we should do that as a separate bug and let this PR follow the current behavior ?

mysql-test/suite/encryption/t/encrypt_threads_skip_tablespace.test
robgolebiowski: Yes. Because it is online encrypted tablespace user should use encryption threads to decrypt it. Then he can use ENCRYPTION='N' to exclude the tablespace from encryption (encryption by encryption threads). This is how encryption threads were designed. If we want to allow this "offline" decryption and exclusion, we can do that by rebuilding the tablespace. However, this should be done as a separate FR, as this was outside the encryption threads workload.

robgolebiowski:
Hey @satya-bodapati. Sorry this took so long, was involved in adding kv2 support to keyring_vault and also I had to get PS-6051 merged first - to clean up the encryption threads from random failures. Now I can confirm that all failing tests from the current run : https://ps80.cd.percona.com/job/percona-server-8.0-param/426/ (--suite=encryption,innodb) are also failing on trunk.

storage/innobase/dict/dict0dd.cc
```diff
@@ -6224,17 +6224,19 @@ static bool dd_update_tablespace_dd_flags(
 
 bool dd_set_encryption_flag(THD *thd, const char *space_name,
                             volatile bool *is_space_being_removed) {
-  auto update_func = [](uint32_t &dd_space_flags) {
+  auto update_func = [](uint32_t &dd_space_flags, dd::Tablespace *dd_space) {
     dd_space_flags |= (1U << FSP_FLAGS_POS_ENCRYPTION);
+    dd_space->options().set("encryption", "Y");
```
satya-bodapati, line 6229: why do we update DD now? I_S queries use only the in-memory object(fil_space_t) right?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15904,13 +15904,42 @@ static int innobase_alter_encrypt_tablespace(handlerton *hton, THD *thd,
...
+    if (!fil_crypt_exclude_tablespace_from_rotation(space)) {
+      my_error(ER_EXCLUDE_ENCRYPTION_THREADS_RUNNING, MYF(0), space->name);
+      DBUG_RETURN(convert_error_code_to_mysql(DB_ERROR, 0, NULL));
+    }
+    DBUG_RETURN(error);
+  } else if (!is_old_encrypted && is_new_encrypted) {
+    // TODO: PS-5323 should disallow reaching this code if encryption threads
+    // are enabled. Thus it should be safe to remove the crypt_data here. Here
+    // we encrypt with MK, the tablespace was previously excluded from rotation.
+    if (space->crypt_data) {
```
satya-bodapati: Here and rest of code, use space->crypt_data != nulltpr

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15904,13 +15904,42 @@ static int innobase_alter_encrypt_tablespace(handlerton *hton, THD *thd,
...
+      ut_ad(space->crypt_data->type == CRYPT_SCHEME_UNENCRYPTED);
+      fil_space_destroy_crypt_data(&space->crypt_data);
+      space->crypt_data = nullptr;
```
satya-bodapati, line 15923: This might cause troubles. many places space->crypt_data is used for takin decisions (wihtout any protection) and here too, without any protection, you are setting it to nullptr

Prone to race condition IMO

storage/innobase/handler/ha_innodb.cc
satya-bodapati: is 5323 done?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15904,13 +15904,42 @@ static int innobase_alter_encrypt_tablespace(handlerton *hton, THD *thd,
...
+      rw_lock_x_lock(&space->latch);
+      space->encryption_type = Encryption::NONE;
+      // err = fil_set_encryption(space_id, Encryption::AES, key, iv);
```
satya-bodapati: Remove commented out code

satya-bodapati (changes requested):

satya-bodapati:
Also, afair, you plan to remove DD.explicit_encryption attribute? separately as part of another PR?

I would like to know why. Because I thought that it is best way to rely to DD and not create crypt_data for encryption='n' tablespaces.

@satya-bodapati : as discussed on slack this can be an improvement over the current design, mostly to get rid of unnecessary tablespace rebuild when it is excluded from encryption threads. Although I would prefer to leave the crypt_data on page0 for encryption='n' and only update page0 when doing ALTER ENCRYPTION='N' instead of rebuilding whole tablespace. This is mostly because that changes quite a lot on how the MariaDB port is working. Also we would need to consider updating DD flags when importing/exporting tables. Overall, the amount of work to do can be significant with explicit_encryption. Create a jira ticket: PS-6102: Consider to re-add explicit_encryption flag to DD to speed up excluding table from encryption threads.

storage/innobase/dict/dict0dd.cc
robgolebiowski: This maybe should not be part of this PR. It's been a while, but if I remember correctly I just noticed that encryption threads only update the flags and they should also update the options.

storage/innobase/handler/ha_innodb.cc
robgolebiowski: Not yet. I was already implemented in PR: https://github.com/percona/percona-server/pull/3246, but although it was approved by Laurynas, it was never approved by you. default_table_encryption came along with tablespace encryption and now this PR needs to be rewritten.

storage/innobase/handler/ha_innodb.cc
robgolebiowski: as discussed on slack: crypt_data can be only removed if encryption threads are disabled. Since crypt_data can be only access concurrently by encryption threads - this is race free.

storage/innobase/handler/ha_innodb.cc

storage/innobase/handler/ha_innodb.cc
robgolebiowski: changed to space->crypt_data != nullptr

storage/innobase/fil/fil0crypt.cc
satya-bodapati: IMHO this is simple request. Again separate JIRA ticket, PR, scheduling is un-necessary overhead.

all I ask is to do this
```
fil_space_crypt_t* crypt_data = fil_space_create_crypt_data()
fil_crypt_write_crypt_data_to_page0(crypt_data);
space->crypt_data = crypt_data
```
Also you claim it is race-free, so only doing this at place should be fine right?

In my analysis (below), space->crypt_data is read/written under different mutexes at different times. So it is really hard for me..

1. fil_space_set_crypt_data under  mutex_enter(&crypt_data->mutex);
2. fil_crypt_read_crypt_data() uses some kind of double checking & space->size to avoid race. No latch protection
3. some readers under  mutex_enter(&fil_crypt_threads_mutex);
4. fil_crypt_rotate_page() reads space->crypt_data without any protection
5. fil_space_crypt_close_tablespace(sp); without any protection. Releases shard mutex as well :(
6. fil_space_crypt_get_status() used by I_S , no protection. Uses double check of space->crypt_data poniter
7. may other usages that I am yet to analyze..

satya-bodapati (changes requested):
Mostly looks good except  the crypt_data assignment.. If it is hard for you to address, let me know here

satya-bodapati:
> Also, afair, you plan to remove DD.explicit_encryption attribute? separately as part of another PR?
> 
> I would like to know why. Because I thought that it is best way to rely to DD and not create crypt_data for encryption='n' tablespaces.
> 
> @satya-bodapati : as discussed on slack this can be an improvement over the current design, mostly to get rid of unnecessary tablespace rebuild when it is excluded from encryption threads. Although I would prefer to leave the crypt_data on page0 for encryption='n' and only update page0 when doing ALTER ENCRYPTION='N' instead of rebuilding whole tablespace. This is mostly because that changes quite a lot on how the MariaDB port is working. Also we would need to consider updating DD flags when importing/exporting tables. Overall, the amount of work to do can be significant with explicit_encryption. Create a jira ticket: PS-6102: Consider to re-add explicit_encryption flag to DD to speed up excluding table from encryption threads.


Ok thanks. My concern is 'right foundation(adding explicit_encryption to DD)' laid has been removed

storage/innobase/fil/fil0crypt.cc
robgolebiowski: I do not want to do it as part of this PR as crypt_data is assigned first to space and then written to page0 in many places. It needs correction in all those places, not just here.

satya-bodapati: ok, fine. Please create JIRA ticket. I will look into this

robgolebiowski: PS-6107

satya-bodapati (approved):
LGTM

PR #3456, opened 2019-10-08
PS-5921 - Detect memory leaks on stage of bootstrap if PS was built w…

altmannmarcelo:
…ith '-DWITH_ASAN'

Problem:
As part of boostrap process, DDSE_dict_init reads proprieties from
the first table in the data dictionary to decide whatever or not
the next tables will be created with encryption. The object that
stores the proprieties of the first table is never destroyed.

Solution:
Call delete on dd::Properties *p object.

### DIFF
--- a/sql/dd/impl/bootstrap/bootstrapper.cc
+++ b/sql/dd/impl/bootstrap/bootstrapper.cc
@@ -768,6 +768,7 @@ bool DDSE_dict_init(THD *thd, dict_init_mode_t dict_init_mode, uint version) {
       // From now on all DD tables will be created with encryption='y'
     }
   }
+  delete p;
 
   /*
     Iterate over the table definitions and add them to the System_tables


altmannmarcelo:
https://ps80.cd.percona.com/job/percona-server-8.0-param/355/ - Jenkins tests are failing with:

```
. . . 
Removing old var directory
Creating var directory '/tmp/results/PS/mysql-test/var'
Installing system database
Using parallel: 1

worker[1] 191008 21:31:29 mysql-test-run: WARNING: Process [mysqld.1 - pid: 109, winpid: 109, exit: 10752] died after mysql-test-run waited 0.2 seconds for /tmp/results/PS/mysql-test/var/run/mysqld.1.pid to be created.
191008 18:31:29 [ 50%] main.thread_cache_size                    [ fail ]
        Test ended at 2019-10-08 18:31:29

Server log is:
ERROR: ld.so: object '/usr/lib64/libasan.so.5
/usr/lib64/libasan.so.5' from LD_PRELOAD cannot be preloaded: ignored.
==110==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
safe_process[109]: Child process: 110, exit: 42
``` 

Running manually on my machine works:
```
bash-4.2$ ./mysql-test/mtr --big-test --debug-server --sanitize main.thread_cache_size
Logging: /tmp/ps/mysql-test/mysql-test-run.pl  --big-test --debug-server --sanitize main.thread_cache_size
MySQL Version 8.0.16
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Removing old var directory
Creating var directory '/tmp/ps/bld/mysql-test/var'
Installing system database
Using parallel: 1

[ 50%] main.thread_cache_size                    [ pass ]    163
[100%] sanitize_report                           [ pass ]
Spent 0.163 of 29 seconds executing testcases

Completed: All 2 tests were successful.
```

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -768,6 +768,7 @@ bool DDSE_dict_init(THD *thd, dict_init_mode_t dict_init_mode, uint version) {
       // From now on all DD tables will be created with encryption='y'
     }
   }
+  delete p;
```
dutow, line 771: A better (safer) fix would be to change the type of p to `std::unique_ptr<dd::Properties>` - grep for other uses of parse_properties for examples.

dutow (changes requested):

altmannmarcelo:
Closing since 8.0.17 has @percona-ysorokin 's fix https://github.com/percona-ysorokin/percona-server/commit/9752e38f3089f819a35a616785a518227f397721

PR #3463, opened 2019-10-14
PS-3970: Enable RocksDB in Travis CI for gcc-8 (5.7)

inikep:
After fixing issues with gcc-9, current trunk compiles without warnings with RocksDB also with gcc-8.
This patch enables RocksDB in Travis CI for gcc-8.


percona-ysorokin (approved):
LGTM

PR #3466, opened 2019-10-17
PS-5181: Assertion when executing multi table DROP

kamil-holubicki:
https://jira.percona.com/browse/PS-5181

This is handling of corner case when:
1. gtid_next is set to GTID value
2. DROP statement with temporary and non-temporary tables is executed
3. row-based logging is enabled

Theoretically such transaction should not be logged by master. It would be split into 2 separate transactions (SBR) or drop of temporary table would be skipped from logging, resulting in single table DROP for non-temporary table (RBR)

However there are still at least 2 cases when it can happen:

1. Simulate replicated transaction by invoking DROP with 2 such tables directly on slave.
2. Create t1, t2 as non-temporary tables on master, then replicate to slave. Next drop t1, t2 from slave and create t1, t2 on slave only using stored procedure. This scenario is strange, however rpl_gtid_split_statements.test performs such scenario.

If slave uses RBL, dropping of temporary table will not be logged, so single table drop of non-temporary table will go to slave's binlog.
If slave uses SBL, case 1 above will issue the error (2 tables would be logged to binlog, but as separate transactions using the same GTID, which is not allowed).
If slave uses SBL, case 2 will still work, as temporary tables created by stored procedures are not logged even if SBL is used.

### DIFF
--- a/mysql-test/suite/rpl_gtid/r/rpl_split_statements_bug_ps5181.result
+++ b/mysql-test/suite/rpl_gtid/r/rpl_split_statements_bug_ps5181.result
@@ -1,0 +1,19 @@
+#
+# PS-5181 Assertion when droping temporary and non-temporary table as single DROP statement when Row Based Logging (RBL)
+#
+RESET MASTER;
+CREATE TABLE t1 (a int) engine = MyISAM;
+CREATE TEMPORARY TABLE t2 (a int);
+SET SESSION gtid_next = '11111111-AAAA-2222-BBBB-000000000000:1';
+DROP TABLE t1, t2;
+SET SESSION gtid_next = 'AUTOMATIC';
+SET SESSION binlog_format = 'STATEMENT';
+CREATE TABLE t1 (a int) engine = MyISAM;
+CREATE TEMPORARY TABLE t2 (a int);
+SET SESSION gtid_next = '11111111-AAAA-2222-BBBB-000000000000:2';
+DROP TABLE t1, t2;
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://ps80.cd.percona.com/job/percona-server-8.0-param/366/

kamil-holubicki:
Hi @percona-ysorokin , @dutow .
Could you please review?

mysql-test/suite/rpl/t/rpl_split_statements_bug_ps5181-master.opt
```diff
@@ -0,0 +1 @@
+--gtid_mode=ON --enforce_gtid_consistency=ON
```
dutow: Place the testcase in the rpl_gtid suite, these are defaults there.

mysql-test/suite/rpl/t/rpl_split_statements_bug_ps5181.test
```diff
@@ -0,0 +1,49 @@
...
+
+--source include/have_myisam.inc
+--source include/have_binlog_format_mixed_or_row.inc
+
+--echo #
+--echo # PS-5181 Assertion when droping temporary and non-temporary table as single DROP statement when Row Based Logging (RBL)
+--echo #
+
+RESET MASTER;
+CREATE TABLE t1 (a int) engine = MyISAM;
```
dutow: 8.0 doesn't always has myisam, include have_myisam.inc

dutow (changes requested):
Only two minor test related comments.

kamil-holubicki:
https://ps80.cd.percona.com/job/percona-server-8.0-param/412/

PR #3469, opened 2019-10-17
PS-5865 : Switching SE from InnoDB to MyRocks isn't supported in 8.0.16

george-lorch:
- Removed incomplete check for ENCRYPTION=Y|N in MyRocks as 8.0.16 adds a more
  correct and complete check now before entering the handler ::create

- Added simple test that reproduces the issue before the fix.


george-lorch:
https://ps80.cd.percona.com/job/percona-server-8.0-param/369/

percona-ysorokin (approved):
LGTM

PR #3477, opened 2019-10-22
PS-5842 - Memory leak after 'rocksdb.sys_vars.rocksdb_update_cf_optio…

altmannmarcelo:
…ns_basic' - in 'sys_var_pluginvar::global_update()'

MyRocks uses update_cf_options option to trigger updates of cf options
using functions like SetOptions. While setting the value of var_ptr
we were copying the value of val leaking the old value of the pointer.
update_cf_options variable is declared with PLUGIN_VAR_MEMALLOC flag,
which will call my_free on previously used resources.

The solution is to use a static cast of save pointer and let
plugin_var_memalloc_global_update take care of freeing allocated
resources.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -13111,7 +13111,8 @@ static void rocksdb_set_update_cf_options(
   // Reset the pointers regardless of how much success we had with updating
   // the CF options. This will results in consistent behavior and avoids
   // dealing with cases when only a subset of CF-s was successfully updated.
-  *reinterpret_cast<char **>(var_ptr) = my_strdup(0, val, MYF(0));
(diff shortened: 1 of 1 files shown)


altmannmarcelo:
https://ps57.cd.percona.com/job/percona-server-5.7-pipeline/7490/ - Debug
https://ps57.cd.percona.com/job/percona-server-5.7-pipeline/7488/ - RelWithDebInfo

dutow (changes requested):
Are you sure this is always correct?

Looking at other string sysvar update functions, for example innodb_internal_table_update, the correct approach seem to be:

* save the current value of var_ptr into a temporary variable
* update var_ptr with my_strdup
* free the saved old value of var_ptr with my_free

dutow (approved):

PR #3485, opened 2019-10-29
PS-6023: Percona server crashes after a Kerberos password change

gl-sergei:
response

Problem:

Issue comes when PAM plugin needs to perform dialog consisting of more
than one request/response actions. Response buffer address passed to
auth_pam_talk_perform is invalid for second and subsequent responses.

Fix:

Pass correct response buffer address to auth_pam_talk_perform.


gl-sergei:
PAM plugin is not testable with Jenkins. Best I can do is to build with -DWITH_PAM=1 and see if plugin successfully installs.

Tested manually in vagrant box as described in the issue.

gl-sergei:
https://ps56.cd.percona.com/view/5.6/job/percona-server-5.6-param/68/

george-lorch (approved):

PR #3490, opened 2019-10-30
PS-6034 Misc typos on the Installing Percona Server for MySQL on Red …

patrickbirch:
…Hat Enterprise Linux and CentOS page

fixed the page reference
fixed the first line in Repo Installation
fixed the note for package installation and added the command to disable mysql module
fixed the Important section on curl which needed formatting.


hrvojem (approved):
LGTM

PR #3499, opened 2019-11-04
PS-6052: gcc 9.2.1 reports -Werror=stringop-overflow in os_file_get_parent_dir (8.0)

inikep:


storage/innobase/os/os0file.cc
```diff
@@ -1990,6 +1990,10 @@ static char *os_file_get_parent_dir(const char *path) {
     return (NULL);
   }
 
+  if (last_slash - path < 0) {
```
satya-bodapati, line 1994: what does this code mean/ does? both are char pointers

satya-bodapati: Please add comment (like it was done above) :)

satya-bodapati (changes requested):

inikep:
Tested at https://ps80.cd.percona.com/job/percona-server-8.0-param/399/

storage/innobase/os/os0file.cc
inikep: done

percona-ysorokin (approved):
LGTM

PR #3500, opened 2019-11-04
PS-6054: clang 5.0.1 reports "missing-braces" warnings (8.0)

inikep:


cmake/maintainer.cmake
```diff
@@ -140,6 +140,7 @@ ENDIF()
 IF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   # Disable a few default Clang++ warnings
   STRING_APPEND(MY_CXX_WARNING_FLAGS " -Wno-null-conversion -Wno-unused-private-field")
+  STRING_APPEND(MY_CXX_WARNING_FLAGS " -Wno-missing-braces")
```
satya-bodapati: Isn't it better to add braces to reported pieces of code instead of suppressing on whole code base?

satya-bodapati (changes requested):

cmake/maintainer.cmake
inikep: It's at least in 4 separate modules so it should be in `cmake/maintainer.cmake`, but now it will work only for clang less than 6.0.

inikep: This is a `clang` bug therefore I expect additional warnings from Oracle in the future so `maintainer.cmake` is better.

percona-ysorokin (approved):
LGTM

PR #3505, opened 2019-11-05
8.0 PS-5325 Conditional jump or move depends on uninitialised value

altmannmarcelo:
Problem:
As part of DISCARD TABLESPACE, MySQL deinitialize table statistics.
Later if we need to access the table (MySQL allows DDL)
those variables won't have been initialized again. This is a regression
introduced by PS-3829 at [1].

Solution:
Call dict_stats_init even if the table has been discarded.
This will result in dict_stats_report_error been called, which as
part of the error handling will init empty table statistics.
As part of the fix, we need to suppress warnings that will be generated
by dict_stats_report_error on related tests.

[1]:

commit c7f44ee
Author: Robert Golebiowski <robert.golebiowski@percona.com>
Date:   Thu Nov 16 17:53:11 2017 +0100

    PS-3829 : Innodb key rotation. ALPHA

    This PR implements Encryption threads for
    encryption/re-encryption/decryption of
    innodb tablespaces. It is based on MariaDB implementation ported from
    Google's
    patch. With this WL user can:

    1) Change default KEYRING encryption key
    2) Create a new table with KEYRING
    3) Encrypt offline already existing tables with KEYRING (with alter)
    4) Encrypt/re-encrypt online already existing tables with KEYRING (with
    innodb_online_encryption, innodb_online_encryption_rotate_key_age
    variables and innodb_online_encryption_threads) – including tables
    already encrypted with Master Key encryption.
    5) Re-encrypt online already encrypted tables with newer version of
    encryption key (with variables innodb_online_encryption variable,
    innodb_online_encryption_threads,
    innodb_online_encryption_rotate_key_age).

    This WL also fixed the following bugs reported to MariaDB:

    (MDEV-17231) Encryption threads keep re-encrypting/encrypting corrupted
    pages
    (MDEV-17235) Data dictonary is not updated with encryption flags
    (MDEV-17234) In memory crypt_data is updated before page0 is updated
    (MDEV-17233) Page 0 is updated more than once when encryption completes
    (MDEV-17230) encryption_key_id from alter is ignored by encryption
    threads
    (MDEV-17229) Encryption threads ignore innodb_default_encryption_key_id

### DIFF
--- a/mysql-test/suite/encryption/r/innodb-bad-key-change3.result
+++ b/mysql-test/suite/encryption/r/innodb-bad-key-change3.result
@@ -1,6 +1,8 @@
 call mtr.add_suppression("cannot be decrypted. Are you using correct keyring that contain the");
 call mtr.add_suppression("InnoDB: Tablespace for table .* is set as discarded.");
 call mtr.add_suppression("InnoDB: Cannot calculate statistics for table .* because the .ibd file is missing. Please refer to .* for how to resolve the issue.");
+call mtr.add_suppression("InnoDB: Trying to access missing tablespace [0-9]+");
+call mtr.add_suppression("InnoDB: Cannot save statistics for table .* because the .ibd file is missing.");
 CREATE TABLE t1 (pk INT PRIMARY KEY, f VARCHAR(255)) ENGINE=InnoDB ROW_FORMAT=COMPRESSED ENCRYPTION='KEYRING' ENCRYPTION_KEY_ID=4;
 SHOW WARNINGS;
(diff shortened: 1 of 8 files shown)


altmannmarcelo:
Valdrind - https://ps80.cd.percona.com/job/percona-server-8.0-param/409/ (tests are failing on debian buster, not related to the issue itself).

full test - https://ps80.cd.percona.com/job/percona-server-8.0-param/408/

dutow (approved):

PR #3507, opened 2019-11-06
PS-6062: Azure Pipelines: task CacheBeta@0 doesn't accept "." (8.0)

inikep:
Use only a major number of compiler version (e.g. "6" instead of "6.0")

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,6 @@
 jobs:
 - job:
-  timeoutInMinutes: 180
+  timeoutInMinutes: 240
   displayName: Ubuntu Bionic
   pool:
     vmImage: 'ubuntu-18.04'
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -162,6 +162,8 @@ jobs:
       ccache --print-config
       ccache --zero-stats
 
+      COMPILER_VER=$(CompilerVer)
+      echo '##vso[task.setvariable variable=CompilerMajorVer]'${COMPILER_VER:0:1}
```
percona-ysorokin: This will make `gcc-1` from `gcc-10` when it is released. Not a problem now but will be in future.

inikep: Changed to `${COMPILER_VER//[^0-9]/}`. Clang 4.0 will be `40` now.

inikep: Changed to `${COMPILER_VER%.*}`:
```
$ ver=10.0
$ echo ${ver%.*}
10
```

percona-ysorokin (approved):
LGTM

PR #3511, opened 2019-11-06
PS-6056 Document MySQL Queries per second is decreasing with higher n…

patrickbirch:
…umber of table partitions

in the performance section, added Limiting the Estimation of Records in a Query

### DIFF
--- a/doc/source/performance/query_limit_records.rst
+++ b/doc/source/performance/query_limit_records.rst
@@ -1,0 +1,50 @@
+.. _query-limit-esitmates:
+
+=========================================================
+Limiting the Estimatation of Records in a Query
+=========================================================
+
+:Availability:  This feature is **Experimental** quality.
+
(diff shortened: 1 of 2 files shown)


doc/source/performance/query_limit_records.rst
```diff
@@ -0,0 +1,50 @@
...
+
+.. variable:: innodb_force_index_records_in_range
+
+    :cli: ``--innodb-force-index-records-in-range``
+    :dyn: Yes
+    :scope: Global
+    :vartype: Numeric
+    :default: 0
+
+This variable provides a method to force the query estimation process to use indexes.
```
george-lorch: This is actually to override the records_in_range result when a FORCE INDEX is used in a query.

george-lorch (changes requested):

doc/source/performance/query_limit_records.rst
patrickbirch: Fixed.

george-lorch (approved):

dutow (approved):

PR #3517, opened 2019-11-11
PS-6063: Azure Pipelines: No space left on device

inikep:
Disable unit tests for gcc-8 and gcc-9 because of error `No space left on device`.


percona-ysorokin (approved):
LGTM

PR #3518, opened 2019-11-11
PS-6073 Document that TokuDB and MyRocks tables do not support the ne…

patrickbirch:
…w clone plugin and API

Updated tokudb/troubleshooting
Updated myrocks/limitations


george-lorch (approved):
Looks like a fair amount of spurious reformatting, but I am fine with this.

PR #3520, opened 2019-11-11
PS-6075 PXB reference in Encryption docs

patrickbirch:
Added the PXB note from a discussion with Sergei on 5.7 and 2.4


doc/source/management/data_at_rest_encryption.rst
```diff
@@ -222,8 +222,7 @@ Trying to add unencrypted table to this table space will result in an error:
 
 .. note::
 
-   |Percona XtraBackup| currently doesn't support backup of encrypted general
-   tablespaces.
+   |Percona XtraBackup| version 2.4 does support the backup of encrypted general tablespaces. Version 2.4 does not support the backup of encrypted system tablespace at this time.
```
dutow: It would be a good idea to mention starting with which 2.4.x version

patrickbirch: Added reference to 2.4.11

dutow (approved):

PR #3521, opened 2019-11-12
PS-6051: After 8.0.17 has been merged encryption mtr tests started to

robgolebiowski:
fail.

All transaction related objects should be freed before thread is marked
as inactive. It means calling destroy_thd (where transaction object
associated with THD is freed) before setting srv_threads.m_encryption_threads_active
to false (marking thread as inactive).

### DIFF
--- a/storage/innobase/fil/fil0crypt.cc
+++ b/storage/innobase/fil/fil0crypt.cc
@@ -73,10 +73,7 @@ static ib_mutex_t fil_crypt_key_mutex;
 static bool fil_crypt_threads_inited = false;
 
 /** No of key rotation threads requested */
-uint srv_n_fil_crypt_threads = 0;
-
-/** No of key rotation threads started */
-uint srv_n_fil_crypt_threads_started = 0;
+uint srv_n_fil_crypt_threads_requested = 0;
 
 /** At this age or older a space/page will be rotated */
 uint srv_fil_crypt_rotate_key_age;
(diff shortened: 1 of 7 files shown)


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/419/

storage/innobase/fil/fil0crypt.cc
```diff
@@ -2525,8 +2535,8 @@ void fil_crypt_thread() {
 
   mutex_enter(&fil_crypt_threads_mutex);
   uint thread_no = srv_n_fil_crypt_threads_started;
-  srv_n_fil_crypt_threads_started++;
   srv_threads.m_encryption_threads_active = true;
+  srv_n_fil_crypt_threads_started++;
```
percona-ysorokin: Do we need `m_encryption_threads_active` at all. It looks like it is always `true` when `srv_n_fil_crypt_threads_started > 0`?

robgolebiowski: yes, but this is part of srv_threads structure which was a common place to hold information on active threads and was consulted by the server to check whether all background threads are done executing.

percona-ysorokin: Can `srv_n_fil_crypt_threads_started` be moved into `srv_threads` then?

percona-ysorokin: @robgolebiowski The only reason I started this was to avoid unsync states between `m_encryption_threads_active` and `srv_n_fil_crypt_threads_started`. Moving `srv_n_fil_crypt_threads_started` into `srv_threads` is fine but my primary comment was about getting rid of `m_encryption_threads_active`.

robgolebiowski: sorry, I just run sed without thinking - right, I will remove the active - working on it

percona-ysorokin (approved):
LGTM

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/421/

PR #3522, opened 2019-11-12
PS-6078 [DOC] verify encryption page has invalid query info

patrickbirch:
verified the first example was not needed.


tplavcic (approved):
LGTM

PR #3524, opened 2019-11-13
PS-5738: Decrease size of KEYRING encryption information on page0

robgolebiowski:
Decreased Keyring encryption information stored on page0 of KEYRING
encrypted tables.

Removed:
 - space_id (4 bytes)
 - offset (2 bytes)
 - tablespace_iv (32 bytes)
Shrank the size of encryption_rotation from 4 bytes to 1 byte
(currently this can have only one of two values: NO_ROTATION or
MASTER_KEY_TO_KEYRING)

tablespace_iv was only used when there was MASTER_KEY_TO_KEYRING
rotation taking place. We would have two IVs then - one for KEYRING
encryption and the second one for MK decryption. However, those IVs do
not need to be different as long as the key used is different - which is
the case here. So instead of generating new IV for KEYRING encryption we
reuse the MK encryption IV when re-encrypting from MK to KEYRING.

This change introduces upgrade problem. When table is in the middle of
MK to KEYRING rotation it cannot be upgraded to this new schema as we do
not have place to store tablespace_iv. We could introduce this field in
the crypt_data in-memory structure to support this rotation. However,
since this feature is experimental we chose to ask user to finish off
the rotation before the upgrade and abort the upgrade if there is table
for which this rotation has not been finished. It also helps to clean up
the code, because we have to care only about one iv, instead of
two (and tablespace_iv being used in very rare cases).

The table which is being re-encrypted from MK to KEYRING has it first
page always validated (since encrypted flag is set for this table). We
chose to check if it is in the middle of MK to KEYRING re-encryption
there.

### DIFF
--- a/mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
+++ b/mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
@@ -1,0 +1,114 @@
+# This test contains keyring file created on 64bit which is not
+# portable
+--source include/have_64bit.inc
+--source include/no_valgrind_without_big.inc
+--source include/have_util_unzip.inc
+--source include/have_debug.inc
+
+--disable_query_log
+call mtr.add_suppression("Resizing redo log from");
+call mtr.add_suppression("Upgrading redo log");
+call mtr.add_suppression("Starting to delete and rewrite log files");
+call mtr.add_suppression("New log files created");
+call mtr.add_suppression("Unknown system variable 'show_compatibility_56'");
+call mtr.add_suppression("You need to use --log-bin to make --binlog-format work");
+call mtr.add_suppression("Creating routine without parsing routine body");
+call mtr.add_suppression("Resolving dependency for the view");
+call mtr.add_suppression("references invalid");
+call mtr.add_suppression("doesn't exist");
+call mtr.add_suppression("information_schema");
+call mtr.add_suppression("Storage engine '.*' does not support system tables. \\[mysql.*\\]");
+call mtr.add_suppression("Table 'mysql.component' doesn't exist");
+call mtr.add_suppression("is expected to be transactional");
+call mtr.add_suppression("table is missing or has an incorrect definition");
+call mtr.add_suppression("ACL DDLs will not work unless mysql_upgrade is executed");
+call mtr.add_suppression("Native table .* has the wrong structure");
+call mtr.add_suppression("Column count of mysql.* is wrong");
+call mtr.add_suppression("Cannot open table mysql/version from the internal data dictionary of InnoDB though the .frm file for the table exists");
+call mtr.add_suppression("Column count of performance_schema.events_statements_summary_by_digest is wrong");
+call mtr.add_suppression("The privilege system failed to initialize correctly");
+call mtr.add_suppression("Missing system table mysql.global_grants");
+# InnoDB reports "Lock wait timeout" warnings when it tries to drop persistent
+# statistics while persistent statistics table is altered during upgrade.
+# This issue doesn't seem to cause any further trouble (as there is no persistent
+# stats for persistent stats table anyway), so we ignore these warnings here.
+call mtr.add_suppression("Unable to delete statistics for table mysql.");
+# new fields were added to these tables
+call mtr.add_suppression("Column count of performance_schema.replication_group_members is wrong. Expected 7, found 5");
+call mtr.add_suppression("Column count of performance_schema.replication_group_member_stats is wrong. Expected 13, found 9");
+call mtr.add_suppression("Column count of performance_schema.threads is wrong. Expected 18, found 17");
+call mtr.add_suppression("ACL table mysql.[a-zA-Z_]* missing. Some operations may fail.");
+call mtr.add_suppression("Info table is not ready to be used. Table 'mysql.slave_master_info' cannot be opened");
+call mtr.add_suppression("Error in checking mysql.slave_master_info repository info type of TABLE");
+call mtr.add_suppression("Error creating master info: Error checking repositories.");
+call mtr.add_suppression("Slave: Failed to initialize the master info structure for channel");
+call mtr.add_suppression("Failed to create or recover replication info repositories.");
+call mtr.add_suppression("db.opt file not found for test database. Using default Character set");
+call mtr.add_suppression("Skip re-populating collations and character sets tables in InnoDB read-only mode");
+call mtr.add_suppression("Skipped updating resource group metadata in InnoDB read only mode");
+--enable_query_log
+
+--echo # Set different paths for --datadir
+let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/data_57_crypt_v1/data;
+
+--copy_file $MYSQLTEST_VARDIR/std_data/data_57_crypt_v1.zip $MYSQL_TMP_DIR/data_57_crypt_v1.zip
+
+--echo # Check that the file exists in the working folder.
+--file_exists $MYSQL_TMP_DIR/data_57_crypt_v1.zip
+
+--echo # Unzip the zip file.
+--exec unzip -qo $MYSQL_TMP_DIR/data_57_crypt_v1.zip -d $MYSQL_TMP_DIR
+
+--let $MYSQLD_DATADIR=`SELECT @@datadir`
+
+--echo # Stop DB server which was created by MTR default
+--source include/shutdown_mysqld.inc
+
+--echo # Start the 8.0.18 server on 5.7 datadir that contains tables with crypt_data v1
+--let $restart_parameters = "restart: --loose-skip-log-bin --skip-log-slave-updates --datadir=$MYSQLD_DATADIR1 --keyring_file_data=$MYSQL_TMP_DIR/data_57_crypt_v1/mysecret_keyring $KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD"
+--replace_result $MYSQLD_DATADIR1 MYSQLD_DATADIR1 $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD KEYRING_PLUGIN_EARLY_LOAD
+--source include/start_mysqld.inc
+
+--let tables_count=`select count(*) from INFORMATION_SCHEMA.INNODB_TABLESPACES`
+
+--echo # Now we will check if the encryption will finish off correctly on the tables with crypt_data v1
+--echo # and if the tables are readable. The tables were created in 5.7 as:
+--echo #CREATE TABLE t1 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION='Y';
+--echo #CREATE TABLE t2 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION='KEYRING';
+--echo #CREATE TABLE t3 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION_KEY_ID=4;
+--echo #CREATE TABLE t4 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION='N';
+
+--let $assert_text= table t4 should have excluded set to Y
+--let $assert_cond= "[SELECT EXCLUDED FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'test/t4\\']" = "Y"
+--source include/assert.inc
+
+SET GLOBAL innodb_encryption_threads = 4;
+SET GLOBAL default_table_encryption=ONLINE_TO_KEYRING;
+
+--echo # Wait max 10 min for key encryption threads to encrypt all spaces
+--let $wait_timeout= 600
+# All tables should get encrypted. tables_count - 2 because temporary tablespace, t4 are not encrypted and.
+--let $wait_condition=SELECT COUNT(*) = $tables_count - 2 FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION = 1
+--source include/wait_condition.inc
+
+SELECT COUNT(*) FROM t1;
+SELECT COUNT(*) FROM t2;
+SELECT COUNT(*) FROM t3;
+SELECT COUNT(*) FROM t4;
+
+# Check if server restarts correctly
+--replace_result $MYSQLD_DATADIR1 MYSQLD_DATADIR1 $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD KEYRING_PLUGIN_EARLY_LOAD
+--source include/restart_mysqld.inc
+
+--source include/shutdown_mysqld.inc
+
+--echo # Remove copied files
+--remove_file $MYSQL_TMP_DIR/data_57_crypt_v1.zip
+
+--force-rmdir $MYSQL_TMP_DIR/data_57_crypt_v1
+
+--echo # Restart the server with default options.
+--let $restart_parameters=
+--source include/start_mysqld.inc
+
+SELECT 1;
--- a/mysql-test/suite/encryption/t/upgrade_crypt_data_v1.test
+++ b/mysql-test/suite/encryption/t/upgrade_crypt_data_v1.test
@@ -1,0 +1,128 @@
+# This test contains keyring file created on 64bit which is not
+# portable
+--source include/have_64bit.inc
+--source include/no_valgrind_without_big.inc
+--source include/have_util_unzip.inc
+--source include/have_debug.inc
+
+--disable_query_log
+call mtr.add_suppression("Resizing redo log from");
+call mtr.add_suppression("Upgrading redo log");
+call mtr.add_suppression("Starting to delete and rewrite log files");
+call mtr.add_suppression("New log files created");
+call mtr.add_suppression("Unknown system variable 'show_compatibility_56'");
+call mtr.add_suppression("You need to use --log-bin to make --binlog-format work");
+call mtr.add_suppression("Creating routine without parsing routine body");
+call mtr.add_suppression("Resolving dependency for the view");
+call mtr.add_suppression("references invalid");
+call mtr.add_suppression("doesn't exist");
+call mtr.add_suppression("information_schema");
+call mtr.add_suppression("Storage engine '.*' does not support system tables. \\[mysql.*\\]");
+call mtr.add_suppression("Table 'mysql.component' doesn't exist");
+call mtr.add_suppression("is expected to be transactional");
+call mtr.add_suppression("table is missing or has an incorrect definition");
+call mtr.add_suppression("ACL DDLs will not work unless mysql_upgrade is executed");
+call mtr.add_suppression("Native table .* has the wrong structure");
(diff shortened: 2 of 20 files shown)


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/427

storage/innobase/fil/fil0crypt.cc
```diff
@@ -635,36 +694,112 @@ static fil_space_crypt_t *fil_space_set_crypt_data(
...
+@param[in]  ptr  Log entry start
+@param[in]  end_ptr  Log entry end
+@param[in]  len  Log entry length
+@return position on log buffer */
+byte *fil_parse_write_crypt_data_v1(space_id_t space_id, byte *ptr,
+                                    const byte *end_ptr, ulint len) {
+  ptr += 4;  // skip offset and len
+  ptr += 2;  // skip iv_length
+
+  if (len != KERYING_ENCRYPTION_INFO_MAX_SIZE) {
```
satya-bodapati: How can this work?

In v1 KERYING_ENCRYPTION_INFO_MAX_SIZE should be different and
v2 KERYING_ENCRYPTION_INFO_MAX_SIZE size is something else.

Ideally, we should have KERYING_ENCRYPTION_INFO_MAX_SIZE_57 with old data layout and old fields(struct).

sql/dd/impl/upgrade/server.cc
```diff
@@ -788,6 +788,23 @@ bool do_server_upgrade_checks(THD *thd) {
...
+      Get hold of the InnoDB handlerton. The check for partitioned tables
+      using shared tablespaces is only relevant for InnoDB.
+    */
+    plugin_ref pr =
+        ha_resolve_by_name_raw(thd, LEX_CSTRING{STRING_WITH_LEN("InnoDB")});
+    handlerton *hton =
+        (pr != nullptr ? plugin_data<handlerton *>(pr) : nullptr);
+    DBUG_ASSERT(hton != nullptr && hton->is_there_mk_to_keyring_rotation);
+
+    if (hton->is_there_mk_to_keyring_rotation()) {
```
satya-bodapati, line 700: Why do we need this new handlerton API? Because when tablespace are loaded, InnoDB initialization would fail?
(for 5.7->8.0 and 8.0.X->8.0.Y upgrades)

The handlerton API that uses for tablespace upgrade should be reused for this purpose.

satya-bodapati (changes requested):
some minor comments left.. Review in progress

storage/innobase/fil/fil0crypt.cc
robgolebiowski: right , added KEYRING_ENCRYPTION_INFO_MAX_SIZE_V1 - to keep the naming inline with crypt_data_v1 and v2.

sql/dd/impl/upgrade/server.cc
robgolebiowski: As we have discussed on slack. When server is initialized it validates page0 of each tablespace. This is were we can check if tablespace is in rotation from MK encryption to keyring encryption for free. This handleron API indicates whether such table was discovered during the startup. It is checked in do_server_upgrade_check - which is natural place where to check for situations that would cause the upgrade to fail.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -411,22 +421,16 @@ static ulint fsp_header_get_keyring_encryption_offset(
...
-  }
+  ut_ad(memcmp(page + offset, ENCRYPTION_KEY_MAGIC_PS_V1,
+               ENCRYPTION_MAGIC_SIZE) == 0);
 
-  bytes_read += ENCRYPTION_MAGIC_SIZE;
+  ulint bytes_read{ENCRYPTION_MAGIC_SIZE};
 
-  uint8_t iv_length = mach_read_from_2(page + offset + bytes_read);
-  bytes_read += 2;
+  bytes_read += 2;  // skip iv_length
```
satya-bodapati: For v1, why do we now skip iv_length check ?

robgolebiowski: it is always 16 bytes, so no need to read this ... do you want me to ?

satya-bodapati: better to read and assert that it is 16 bytes. It can help in catching corruption

storage/innobase/fil/fil0crypt.cc
```diff
@@ -486,24 +561,34 @@ fil_space_crypt_t *fil_space_read_crypt_data(const page_size_t &page_size,
...
-                                                     // tablespace_iv
-    uchar tablespace_iv[ENCRYPTION_KEY_LEN];
-    ut_ad(tablespace_iv != NULL);
-    memcpy(tablespace_iv, page + offset + bytes_read, ENCRYPTION_KEY_LEN);
-    bytes_read += ENCRYPTION_KEY_LEN;
-    crypt_data->set_tablespace_iv(tablespace_iv);
+    crypt_data->set_tablespace_key(
+        tablespace_key);  // We are using the same iv for both
+                          // MK encryption and KEYRING encryption
   }
```
satya-bodapati, line 577: This has been removed because it was redundant?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -467,16 +470,88 @@ fil_space_crypt_t *fil_space_read_crypt_data(const page_size_t &page_size,
...
+  if (memcmp(page + offset, ENCRYPTION_KEY_MAGIC_PS_V2,
+             ENCRYPTION_MAGIC_SIZE) != 0) {
+    /* Crypt data is not stored. */
+    return nullptr;
+  }
+
+  ulint bytes_read{ENCRYPTION_MAGIC_SIZE};
+
+  uint8_t type = mach_read_from_1(page + offset + bytes_read);
+  bytes_read += 1;
```
satya-bodapati, line 521: Can you please add this assert (like you add in parse)


  ut_a(type == CRYPT_SCHEME_UNENCRYPTED ||
       type == CRYPT_SCHEME_1);  // only supported

storage/innobase/fil/fil0crypt.cc
```diff
@@ -635,36 +707,112 @@ static fil_space_crypt_t *fil_space_set_crypt_data(
...
-    return NULL;
+    return nullptr;
   }
 
   if (ptr + KERYING_ENCRYPTION_INFO_MAX_SIZE > end_ptr) {
-    return NULL;
+    return nullptr;
   }
 
   // We should only enter this function if ENCRYPTION_KEY_MAGIC_PS_V1 is set
```
satya-bodapati: Only if  V2 is set

storage/innobase/fil/fil0crypt.cc
```diff
@@ -635,36 +707,112 @@ static fil_space_crypt_t *fil_space_set_crypt_data(
...
+  }
+
+  /* Check is used key found from encryption plugin */
+  if (crypt_data->should_encrypt() && !crypt_data->is_key_found()) {
+    ib::error() << "Key cannot be read for space id = " << space_id;
+    recv_sys->set_corrupt_log();
+  }
+
+  /* update fil_space memory cache with crypt_data */
+  if (fil_space_t *space = fil_space_acquire_silent(space_id)) {
```
satya-bodapati: would be nice if you can avoid assignments inside if expression..

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
```diff
@@ -0,0 +1,108 @@
...
+--source include/start_mysqld.inc
+
+--let tables_count=`select count(*) from INFORMATION_SCHEMA.INNODB_TABLESPACES`
+
+--echo # Now we will check if the encryption will finish off correctly on the tables with crypt_data v1
+--echo # and if the tables are readable. The tables were created in 5.7 as:
+--echo #CREATE TABLE t1 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION='Y';
+--echo #CREATE TABLE t2 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION='KEYRING';
+--echo #CREATE TABLE t3 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION_KEY_ID=4;
+--echo #CREATE TABLE t4 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION='N';
```
satya-bodapati, line 79: can you do show create table for these tables after upgrade?

satya-bodapati: Also assert from I_S.INNODB_TABLESPACES that t4 ENCRYPTION is "N"

mysql-test/suite/encryption/t/upgrade_crypt_data_v1.test
```diff
@@ -0,0 +1,128 @@
...
+--error 1
+--exec $MYSQLD_CMD --log-error=$MYSQLD_LOG --loose-skip-log-bin --skip-log-slave-updates --datadir=$MYSQLD_DATADIR1 --keyring_file_data=$MYSQL_TMP_DIR/data_8016_crypt_v1/mysecret_keyring $KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD
+
+--let ABORT_ON=NOT_FOUND
+--let SEARCH_FILE=$MYSQLD_LOG
+--let SEARCH_PATTERN=Upgrade is not possible as there are tables for which re-encryption from master key encryption to online keyring encryption was not finished\. Please either finish off the re-encryption or decrypt the tables\. This feature is experimental in the version you are upgrading from\.
+--source include/search_pattern_in_file.inc
+
+--echo # Remove the problematic tablespace and re-try the upgrade. It should pass.
+--remove_file $MYSQLD_DATADIR1/test/t1.ibd
```
satya-bodapati, line 91: wouldn't there be stale DD entry, and InnoDB should abort startup complaining that IBD file is missing?

mysql-test/suite/encryption/t/upgrade_crypt_data_v1.test
```diff
@@ -0,0 +1,128 @@
...
+--source include/start_mysqld.inc
+
+--let tables_count=`select count(*) from INFORMATION_SCHEMA.INNODB_TABLESPACES`
+
+SET GLOBAL innodb_encryption_threads = 4;
+SET GLOBAL default_table_encryption=ONLINE_TO_KEYRING;
+
+--echo # Wait max 10 min for key encryption threads to encrypt all spaces
+--let $wait_timeout= 600
+# All tables should get encrypted. tables_count - 3 because temporary tablespace, t5 are not encrypted and t2 is missing, because we have
```
satya-bodapati: t2 is missing or t1?

share/errmsg-utf8.txt
```diff
@@ -19802,6 +19802,9 @@ ER_XB_MSG_6
 ER_XB_UNDO_DELETE_FAILURE
   eng "Deletion of 5.7 undo tablespace %s failed"
 
+ER_XB_MSG_7
```
satya-bodapati: You should use proper name here.

satya-bodapati (changes requested):

storage/innobase/fil/fil0crypt.cc
robgolebiowski: done

storage/innobase/fil/fil0crypt.cc

storage/innobase/fil/fil0crypt.cc

storage/innobase/fil/fil0crypt.cc
robgolebiowski: Yes. This used to work like this - we have MK encrypted table with its iv (crypt_data->tablespace_iv). We start to re-encrypt this table with keyring encryption. We used to generate a new iv and put it into crypt_data->iv for keyring encryption. We used tablespace_iv when decrypting MK encrypted pages and crypt_data->iv when encrypting pages with Keyring. However, this crypt_data->iv can be the same as tablespace_iv. So I put tablespace_iv in crypt_data->iv and removed  tablespace_iv.

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
robgolebiowski: done

mysql-test/suite/encryption/t/upgrade_crypt_data_v1.test
robgolebiowski: It will complain t1 cannot be find (warning) but will not abort the startup.

mysql-test/suite/encryption/t/upgrade_crypt_data_v1.test
robgolebiowski: right, corrected.

share/errmsg-utf8.txt

storage/innobase/fil/fil0crypt.cc

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
```diff
@@ -0,0 +1,114 @@
...
+
+#SELECT EXCLUDED FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'table/t4\\'
+
+--let $assert_text= table ts4 should have excluded set to Y
```
satya-bodapati: table t4

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
```diff
@@ -0,0 +1,114 @@
...
+--let tables_count=`select count(*) from INFORMATION_SCHEMA.INNODB_TABLESPACES`
+
+--echo # Now we will check if the encryption will finish off correctly on the tables with crypt_data v1
+--echo # and if the tables are readable. The tables were created in 5.7 as:
+--echo #CREATE TABLE t1 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION='Y';
+--echo #CREATE TABLE t2 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION='KEYRING';
+--echo #CREATE TABLE t3 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION_KEY_ID=4;
+--echo #CREATE TABLE t4 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENCRYPTION='N';
+
+#SELECT EXCLUDED FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'table/t4\\'
```
satya-bodapati: remove commented out SQL

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
```diff
@@ -0,0 +1,114 @@
...
+
+SELECT COUNT(*) FROM t1;
+SELECT COUNT(*) FROM t2;
+SELECT COUNT(*) FROM t3;
+SELECT COUNT(*) FROM t4;
+
+# Check if server restarts correctly
+--replace_result $MYSQLD_DATADIR1 MYSQLD_DATADIR1 $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD KEYRING_PLUGIN_EARLY_LOAD
+--source include/restart_mysqld.inc
+
```
satya-bodapati, line 102: Better to add some basic tests/SQLs after restart

storage/innobase/fil/fil0crypt.cc
satya-bodapati: This reusuing of tablespace_iv happened with v1 as well?

satya-bodapati: my concern is that, can we reuse tablespace ivs for v1 format on v2 server?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -636,36 +716,113 @@ static fil_space_crypt_t *fil_space_set_crypt_data(
 
 /******************************************************************
 Parse a MLOG_FILE_WRITE_CRYPT_DATA log entry
```
satya-bodapati: ignore the decorations. Just use ```/** Parse ... ```

storage/innobase/fil/fil0crypt.cc
```diff
@@ -636,36 +716,113 @@ static fil_space_crypt_t *fil_space_set_crypt_data(
...
+    crypt_data = fil_space_set_crypt_data(space, crypt_data);
+    fil_space_release(space);
+  } else {
+    fil_space_destroy_crypt_data(&crypt_data);
+  }
+
+  return ptr;
+}
+/******************************************************************
+Parse a MLOG_FILE_WRITE_CRYPT_DATA log entry
```
satya-bodapati: ignore the decorations. Just use ```/** Parse ... ```

storage/innobase/include/fil0crypt.h
```diff
@@ -436,13 +431,24 @@ void fil_space_destroy_crypt_data(fil_space_crypt_t **crypt_data);
...
+@param[in]  ptr  Log entry start
+@param[in]  end_ptr  Log entry end
+@param[in]  len  Log entry length
+@return position on log buffer */
+byte *fil_parse_write_crypt_data_v1(space_id_t space_id, byte *ptr,
+                                    const byte *end_ptr, ulint len)
+    MY_ATTRIBUTE((warn_unused_result));
+
+/******************************************************************
+Parse a MLOG_FILE_WRITE_CRYPT_DATA log entry
```
satya-bodapati: same here

storage/innobase/include/fil0crypt.h
```diff
@@ -436,13 +431,24 @@ void fil_space_destroy_crypt_data(fil_space_crypt_t **crypt_data);
 
 /******************************************************************
 Parse a MLOG_FILE_WRITE_CRYPT_DATA log entry
-@param[in]	ptr		Log entry start
-@param[in]	end_ptr		Log entry end
-@param[in]	block		buffer block
-@param[out]	err		DB_SUCCESS or DB_IO_DECRYPT_FAIL
+@param[in]  space_id  id of space that this log entry refers to
```
satya-bodapati, line 433: same here (remove decoration)

satya-bodapati (changes requested):

satya-bodapati (approved):
Ok to push after addressing above comments

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
robgolebiowski: done

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
robgolebiowski: done

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
robgolebiowski: done

storage/innobase/fil/fil0crypt.cc

storage/innobase/fil/fil0crypt.cc

storage/innobase/include/fil0crypt.h
robgolebiowski: done

storage/innobase/include/fil0crypt.h
robgolebiowski: done

storage/innobase/fil/fil0crypt.cc
robgolebiowski: no, on v2 server we abort the upgrade if there is a table in v1 format that is in a middle of MK to KEYRING rotation - i.e. it has tablespace_iv set.

PR #3530, opened 2019-11-17
INFORMATION_SCHEMA.PROCFS plugin

ihanick:
Allows to read files from /proc and /sys with select queries for monitoring purposes

https://ps57.cd.percona.com/job/percona-server-5.7-param/187/


plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+ Clients could fetch individual files:
+ SELECT * FROM INFORMATION_SCHEMA.PROCFS WHERE FILE = '/proc/version';
+
+ To install this plugin, copy the .so file to the plugin directory and do
+ INSTALL PLUGIN procfs SONAME 'procfs.so';
+*/
+
+#include <m_ctype.h> /* my_charset_bin */
+#include <mysql_version.h>
+#if MYSQL_VERSION_ID >= 80002
```
dutow, line 35: Since we are adding the plugin code to the main repository, don't include 8.0 specific code in the 5.7 codebase (or 5.7 code in the 8.0 version).

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+#define IS_PROCFS_FILES_PER_PATTERN 10000
+#define IS_PROCFS_PATTERN_DEPTH 10
+
+namespace procfs_plugin {
+std::vector<std::string> files;
+time_t                   files_last_updated_at;
+}  // namespace procfs_plugin
+
+static struct st_mysql_information_schema procfs_view= {
+    MYSQL_INFORMATION_SCHEMA_INTERFACE_VERSION};
```
dutow, line 66: Please also use clang-format for the 5.7 version

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+{
+  const char *category= "PROCFS";
+  int         count;
+
+  count= array_elements(all_procfs_rwlocks);
+  mysql_rwlock_register(category, all_procfs_rwlocks, count);
+}
+#endif
+
+static bool get_equal_condition_argument(Item *cond, std::string *eq_arg,
```
dutow, line 98: cond (and most variables in the function) could be const too. Use const where possible.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+        }
+      }
+    }
+  }
+  return true;
+}
+
+
+static bool get_in_condition_argument(Item *                       cond,
+                                      std::map<std::string, bool> &in_args,
```
dutow, line 158: The value isn't really used in this map - could be an std::set instead. Or even a vector, as it's better for a few items than a set.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+{
+  std::vector<std::string> files_found;
+
+  if (max_results <= 0)
+    return files_found;
+
+  DIR *          dir= opendir(path.c_str());
+  struct dirent *dir_entry;
+
+  if (!dir)
```
dutow, line 214: Don't rely on implicit casts, use dir == NULL instead

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+    return files_found;
+
+  std::vector<char> real_file_path(PATH_MAX);
+
+  while ((dir_entry= readdir(dir)) != 0)
```
dutow, line 219: NULL

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+            files_found.push_back(fname);
+        }
+        if (files_found.size() >= static_cast<size_t>(max_results))
+          break;
+      }
+    }
+  }
+
+  closedir(dir);
+  return files_found;
```
dutow, line 242: This is good for the 8.0 (c++11) code, but problematic in 5.7/c++03: there's no guarantee that the compiler will optimize out the return value, and we don't have move constructors there.

Either the plugin should use C++11 (and then at other parts too, e.g. NULL->nullptr) even for 5.7, or the 5.7 version should use a reference out parameter instead.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+}
+
+static std::vector<std::string> limited_glob(std::string query,
```
dutow, line 245: Same comment about the return value.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+                 FNM_FILE_NAME | FNM_CASEFOLD | FNM_NOESCAPE | FNM_PERIOD))
+    {
+      if (dir_entry->d_type != DT_DIR)
+      {
+        std::string fname=
+            std::string(path).append("/").append(dir_entry->d_name);
+
+        if (realpath(fname.c_str(), real_file_path.data()))
+        {
+          if (strncmp("/sys/", real_file_path.data(), strlen("/sys/")) == 0 ||
```
dutow, line 231: Either create a simple named constans for the lengths, or you could keep a `static const sys_len = strlen("/sys/")` outside the loop, to avoid repeated strlen calls. Same for the next line.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+  struct dirent *dir_entry;
+
+  if (!dir)
+    return files_found;
+
+  std::vector<char> real_file_path(PATH_MAX);
+
+  while ((dir_entry= readdir(dir)) != 0)
+  {
+    if (!fnmatch(pattern.c_str(), dir_entry->d_name,
```
dutow, line 221: 0 == fnmatch(...)

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+  if (max_results <= 0)
+    return files_found;
+
+  DIR *          dir= opendir(path.c_str());
+  struct dirent *dir_entry;
+
+  if (!dir)
+    return files_found;
+
+  std::vector<char> real_file_path(PATH_MAX);
```
dutow, line 217: No need to allocate this dynamically. char real_file_path[PATH_MAX] is fine, or std::array if switching to C++11.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+        std::string fname=
+            std::string(path).append("/").append(dir_entry->d_name);
+
+        if (realpath(fname.c_str(), real_file_path.data()))
+        {
+          if (strncmp("/sys/", real_file_path.data(), strlen("/sys/")) == 0 ||
+              strncmp("/proc/", real_file_path.data(), strlen("/proc/")) == 0)
+            files_found.push_back(fname);
+        }
+        if (files_found.size() >= static_cast<size_t>(max_results))
```
dutow, line 235: Why isn't max_results a std::size_t?

plugin/procfs/procfs.cc
dutow, line 245: Please document what syntax this method supports.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+      query.find_last_of("/", first_star);
+  std::string::size_type first_sep_after_star=
+      query.find_first_of("/", first_star);
+
+  if (last_sep_before_star == std::string::npos)
+  {
+    path= std::string(".");
+    pattern= query;
+  }
+  else if (first_sep_after_star == std::string::npos)
```
dutow, line 268: How is this case possible? If <is not npos, then we have a separator, which means first_sep_after_star will be smaller or equal to last_sep_before_star, it can't be npos.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+          dir_entry->d_type == DT_UNKNOWN)
+      {
+        std::string new_pattern= std::string(path)
+                                     .append("/")
+                                     .append(dir_entry->d_name)
+                                     .append("/")
+                                     .append(rest);
+        std::vector<std::string> v= limited_glob(
+            new_pattern, max_results - files_found.size(), max_depth - 1);
+        files_found.insert(files_found.end(), v.begin(), v.end());
```
dutow, line 308: Considering this use, it would be better for limited_glob to add the results into an out parameter, it would avoid some unnecessary copies.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+
+    if (path.find_first_of("*?[") == std::string::npos)
+    {
+      procfs_plugin::files.push_back(path);
+      continue;
+    }
+
+    std::vector<std::string> v1= limited_glob(
+        path, IS_PROCFS_FILES_PER_PATTERN, IS_PROCFS_PATTERN_DEPTH);
+    procfs_plugin::files.insert(procfs_plugin::files.end(), v1.begin(),
```
dutow, line 357: Same comment for this use too: an out parameter would avoid this copy.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+  mysql_rwlock_unlock(&LOCK_procfs_files);
+
+  mysql_rwlock_wrlock(&LOCK_procfs_files);
+  if (ts < procfs_plugin::files_last_updated_at + IS_PROCFS_REFRESH_SEC)
+  {
+    mysql_rwlock_unlock(&LOCK_procfs_files);
+    return;
+  }
+
+  procfs_plugin::files_last_updated_at= ts;
```
dutow, line 337: `= time(NULL)`, to avoid possible repeated refreshes when multiple sessions call this method around the same time.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+  }
+  mysql_rwlock_unlock(&LOCK_procfs_files);
+
+  procfs_cnf.close();
+}
+
+static int fill_procfs_view(THD *thd, TABLE_LIST *tables, Item *cond)
+{
+  TABLE *table= tables->table;
+  char * buf= static_cast<char *>(
```
dutow, line 368: Why malloc? This could be a local array.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+#include <mysql/plugin.h>
+
+#include <fstream>
+#define IS_PROCFS_CONTENTS_SIZE 60000
+#define IS_PROCFS_REFRESH_SEC 60
+#define IS_PROCFS_FILES_PER_PATTERN 10000
+#define IS_PROCFS_PATTERN_DEPTH 10
+
+namespace procfs_plugin {
+std::vector<std::string> files;
```
dutow, line 61: Using an `std::shared_ptr<std::vector<std::string>>` instead, and creating new vectors instead of clearing the existing vector during a file refresh would allow the fill method to aquire a read lock for filling only for a short time. This could increase the response time when multiple sessions use this table.

Another improvement after this could be moving the filelist refresh to a background thread, which would mean that the client threads only have to read the actual files. (assuming that globbing proc and sys is slow because of the huge amount of files)

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+    return files_found;
+
+  std::vector<char> real_file_path(PATH_MAX);
+
+  while ((dir_entry= readdir(dir)) != 0)
+  {
+    if (!fnmatch(pattern.c_str(), dir_entry->d_name,
+                 FNM_FILE_NAME | FNM_CASEFOLD | FNM_NOESCAPE | FNM_PERIOD))
+    {
+      if (dir_entry->d_type != DT_DIR)
```
dutow, line 224: Are you sure about this condition? This allows sockets, named pipes, block devices, ...

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,475 @@
...
+
+  procfs_plugin::files.clear();
+  while (procfs_cnf)
+  {
+    std::string path;
+    std::getline(procfs_cnf, path);
+    if (path.rfind("/proc", 0) != 0 && path.rfind("/sys", 0) != 0)
+      continue;
+
+    if (path.find_first_of("*?[") == std::string::npos)
```
dutow, line 349: And what if I add "/proc/../some/sensitive/data.txt"? Realpath check seems to be missing in this case.

dutow (changes requested):
I've added some comments about specific parts of the fill methods.

However, because of the complex nature of these methods, I would consider a generic refactoring into a more structured format: smaller, better-separated methods, doing only a simple part of the logic.

E.g.:
* there's some longer code duplication / similar code repetition
* short simple methods with good names would make the logic much easier to understand
* some checks are present in some cases, not in others
* it would be great if the custom glob logic part could be unit tested without actual files, by mocking the few used directory iterator methods, to see that it correctly handles every edge case

dutow:
Closing this as it was already merged in 1f2b64cdb9786d6 to 8.0

PR #3532, opened 2019-11-18
PS-4775 [MyRocks] Improve documentation about rocksdb_wal_recovery mode

patrickbirch:
Added myrocks/recover_from_crash.rst and updated index.rst

### DIFF
--- a/doc/source/myrocks/variables.rst
+++ b/doc/source/myrocks/variables.rst
@@ -619,7 +619,7 @@ Disabled by default.
   :default: ``OFF``
 
 Specifies whether to allow server to restart once MyRocks reported data
-corruption. Disabled by default. 
+corruption. Disabled by default.
 
 Once corruption is detected server writes marker file (named
 ROCKSDB_CORRUPTED) in the data directory and aborts. If marker file exists,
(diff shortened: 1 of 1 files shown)


doc/source/myrocks/recover_from_crash.rst
```diff
@@ -0,0 +1,56 @@
+.. _myrocks_recovery:
+
+======================================
+Recovering from a MyRocks Server Crash
+======================================
+
+MyRocks provides multiple crash recovery configurations. Crash recovery is
+accomplished by replaying the transactions from the WAL (Write Ahead Log) files.
+
+You can set the following crash recovery modes in my.cnf to define how WAL files are applied:
```
george-lorch: These are for the rocksdb variable wal_recovery_mode.  This variable is set from the MyRocks variable rocksdb_wal_recovery_mode, so, we do not need to discuss the wal_recovery_mode for rocksdb as it is redundant with the description of rocksdb_wal_recovery_mode MyRocks variable.  The original bug report was that our documentation of rocksdb_wal_recovery_mode just states that the value can be from 0 to 3 with no description of what 0, 1, 2, or 3 means.

george-lorch (changes requested):

doc/source/myrocks/recover_from_crash.rst
patrickbirch: Fixed. I moved some information to the variables page. Removed the Recover page.

george-lorch (approved):
Some spurious reformatting, but it's OK I guess.

patrickbirch:
Yes. I must investigate why vim is reformatting other areas of the doc.

PR #3534, opened 2019-11-20
PS-1469: MEMORY table "is full" flag stuck when should not be (5.6)

kamil-holubicki:

### DIFF
--- a/mysql-test/t/percona_heap_bug_ps1469.test
+++ b/mysql-test/t/percona_heap_bug_ps1469.test
@@ -1,0 +1,34 @@
+#
+# PS-1469: MEMORY table "is full" flag stuck when should not be
+#
+
+CALL mtr.add_suppression("The table 't1' is full");
+
+SET @@session.max_heap_table_size = 10485760;
+
+CREATE TABLE t1 (inmem VARCHAR(8192)) ENGINE=MEMORY;
+
+DELIMITER //;
+
+CREATE PROCEDURE load_test() 
+  BEGIN 
+    DECLARE v_iterations int default 0; 
+    TRUNCATE TABLE t1; 
+    WHILE v_iterations < 8192 DO 
+      INSERT INTO t1 (inmem) VALUES (REPEAT("a", 8192)); 
+      SET v_iterations=v_iterations+1; 
+    END WHILE;
+  END//
+
+DELIMITER ;//
+
+--error ER_RECORD_FILE_FULL
+CALL load_test;
+
+DELETE FROM t1 LIMIT 100;
+
+INSERT INTO t1 VALUES (REPEAT("a", 8192));
+
+# cleanup
+DROP PROCEDURE load_test;
+DROP TABLE t1;
--- a/storage/heap/hp_write.c
+++ b/storage/heap/hp_write.c
@@ -44,15 +44,21 @@ int heap_write(HP_INFO *info, const uchar *record)
   }
 #endif
 
-  if ((share->records >= share->max_records && share->max_records) ||
-      (share->recordspace.total_data_length + share->index_length >=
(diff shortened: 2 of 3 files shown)


kamil-holubicki:
https://ps56.cd.percona.com/job/percona-server-5.6-param/72/

storage/heap/hp_write.c
```diff
@@ -45,8 +45,8 @@ int heap_write(HP_INFO *info, const uchar *record)
 #endif
 
   if ((share->records >= share->max_records && share->max_records) ||
-      (share->recordspace.total_data_length + share->index_length >=
-       share->max_table_size))
+      ((share->recordspace.total_data_length + share->index_length >=
+       share->max_table_size) && share->recordspace.del_chunk_count == 0)  )
```
dutow: Formatting: `)  )`

mysql-test/t/percona_heap_bug_ps1469.test
```diff
@@ -0,0 +1,38 @@
+#
+# PS-1469: MEMORY table "is full" flag stuck when should not be
+#
+
+#save initial state
+set @old_max_heap_table_size = @@session.max_heap_table_size;
```
percona-ysorokin: No need to save/restore SESSION variables in MTR test cases.

mysql-test/t/percona_heap_bug_ps1469.test
```diff
@@ -0,0 +1,38 @@
...
+CALL load_test;
+
+DELETE FROM t1 LIMIT 100;
+
+INSERT INTO t1 VALUES (REPEAT("a", 8192));
+
+# cleanup
+DROP PROCEDURE load_test;
+DROP TABLE t1;
+SET @@session.max_heap_table_size = @old_max_heap_table_size;
```
percona-ysorokin: No need to save/restore SESSION variables in MTR test cases.

storage/heap/hp_write.c
percona-ysorokin: It looks like `share->recordspace.del_chunk_count == 0` condition is not 100% correct. In general we may be requesting more than 1 chunk, so in my opinion, the correct one is `share->recordspace.del_chunk_count < chunk_count` but it should be moved after `chunk_count` is calculated in `hp_get_encoded_data_length(share, record, &chunk_count);`
So, something like this
```c++
  if ((share->records >= share->max_records && share->max_records))
  {
    my_errno= HA_ERR_RECORD_FILE_FULL;
    DBUG_RETURN(my_errno);
  }

  hp_get_encoded_data_length(share, record, &chunk_count);

  if ((share->recordspace.total_data_length + share->index_length >=
       share->max_table_size) && share->recordspace.del_chunk_count < chunk_count)  )
  {
    my_errno= HA_ERR_RECORD_FILE_FULL;
    DBUG_RETURN(my_errno);
  }
  ...
```

percona-ysorokin (changes requested):

storage/heap/hp_write.c
kamil-holubicki: @percona-ysorokin , Current implementation handles this case properly. Allocation of more than 1 chunk is possible if the record is not fixed size. Even if we have only 1 chunk free, but record needs 2 chunks the situation will be handled in hp_allocate_chunkset() and we will return with error on next 'if'.
However I see that hp_allocate_chunkset() does not set my_errno, so we will get some random error => your proposal improves this, so let's go this way.

kamil-holubicki:
Jenkins build after applying post review adjustments:
https://ps56.cd.percona.com/job/percona-server-5.6-param/73/

percona-ysorokin (approved):
LGTM

PR #3535, opened 2019-11-20
PS-1469: MEMORY table "is full" flag stuck when should not be (5.7)

kamil-holubicki:

### DIFF
--- a/storage/heap/hp_write.c
+++ b/storage/heap/hp_write.c
@@ -44,15 +44,21 @@ int heap_write(HP_INFO *info, const uchar *record)
   }
 #endif
 
-  if ((share->records >= share->max_records && share->max_records) ||
-      (share->recordspace.total_data_length + share->index_length >=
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-param/188/

storage/heap/hp_write.c
```diff
@@ -45,8 +45,8 @@ int heap_write(HP_INFO *info, const uchar *record)
 #endif
 
   if ((share->records >= share->max_records && share->max_records) ||
-      (share->recordspace.total_data_length + share->index_length >=
-       share->max_table_size))
+      ((share->recordspace.total_data_length + share->index_length >=
+       share->max_table_size) && share->recordspace.del_chunk_count == 0)  )
```
dutow: Formatting

percona-ysorokin (approved):
LGTM

PR #3536, opened 2019-11-20
PS-1469: MEMORY table "is full" flag stuck when should not be (8.0)

kamil-holubicki:


kamil-holubicki:
https://ps80.cd.percona.com/job/percona-server-8.0-param/439/

percona-ysorokin (approved):
LGTM

PR #3537, opened 2019-11-20
PS-6065 : PS fails to recover with smaller buffer pool size with impr…

satya-bodapati:
…oved innodb_empty_free_list_algorithm=backoff

Probem:
-------
With small buffer pool size (such as 24M) and innodb_empty_free_list_algorithm=backoff,
recovery hangs looking for free blocks

With backoff, we rely on page cleaners to give free blocks. At recovery stage, page cleaners
wait for recovery flush events and not yet start flushing. The LRU flushers thread are also
idle and not doing anything because the minimum size they need to be active is 256.

At this hanging state, the LRU list length is less than 256 and so it remains idle.

Fix:
----
We remove the minimum length of LRU restriction for LRU flusher threads during recovery

### DIFF
--- a/mysql-test/suite/innodb/r/percona_recovery_backoff_algo.result
+++ b/mysql-test/suite/innodb/r/percona_recovery_backoff_algo.result
@@ -1,0 +1,11 @@
+SELECT @@global.innodb_empty_free_list_algorithm;
+@@global.innodb_empty_free_list_algorithm
+backoff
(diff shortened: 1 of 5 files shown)


satya-bodapati:
patch brewing at https://ps80.cd.percona.com/job/percona-server-8.0-param/441/

percona-ysorokin (approved):
LGTM

PR #3539, opened 2019-11-22
PS-6013: Prevented ICP for keys that cannot be decoded (backported PS-5024 fix) (5.7)

kamil-holubicki:
https://jira.percona.com/browse/PS-6013

Note that PS-5024 problem was not visible in 5.7. For detailed explanation see
https://jira.percona.com/browse/PS-5024


kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-param/189/

george-lorch (approved):
LGTM, also remember that you will need to null-merge this to 8.0

PR #3543, opened 2019-11-27
PS-6106: ALTER TABLE without ENCRYPTION clause causes tables to get

robgolebiowski:
encrypted.
    
    When writting tests for this bug, lost of smaller issues where unveiled.
    In order to have this bug-fix properly tested those smaller issues were
    also fixed. Here is a list of things that this commit addresses:
    
    1.Original problem:
    
    Bug description:
    Given: --default-table-encryption=ONLINE_TO_KEYRING and TABLE t created
    to be always unencrypted:
    CREATE TABLE t (..) ENCRYPTION='N';
    Doing any alter without ENCRYPTION='N' clause, for instance:
    ALTER TABLE t ENGINE InnoDB;
    Will cause t to get encrypted.
    
    Solution description:
    The decision if table should never be encrypted was done based on
    whether create_info->used_fields contained HA_USED_ENCRYPT and
    ENCRYPTION was equal 'N'. The problem here is that used_fields for ALTER
    statement are filled only based on the current CREATE statement.
    Thus making this condition (create_info->used_fields & HA_USED_ENCRYPT) always false
    when ENCRYPTION was not explicitly specified in ALTER statement. create_info
    is generated for ALTER statement based on TABLE_SHARE, which in turn is
    generated based on DD. The fix was to add explicit_encryption to DD so
    to remember whether table was originally created with
    explicit_encryption.
    
    2. CREATE TABLE LIKE does not retain ENCRYPTION='N'
    
    When test was written for this bug it became apparent that CREATE TABLE
    LIKE statement was dropping the explicit ENCRYPTION clause when table was
    unencrypted. We need to retain this explicit ENCRYPTION clause.
    
    3. For any value of default_table_encryption other than 'N' tablespaces
    would be created as encrypted. Even for ONLINE_FROM_KEYRING_TO_UNENCRYPTED.
    
    This was fixed and now tablespaces are created as encrypted when
    default_table_encryption is equal either to Y or ONLINE_TO_KEYRING. When
    ONLINE_TO_KEYRING is set tablespaces are created as encrypted, but can
    be still decrypted by encryption threads.
    During fixing this issue it also become apparent that synchronization
    between global and session value of default_table_encryption is needed.
    For this purpose I have added PS-6759: Sync global value of
    default_table_encryption with session value of this variable.
    
    5. Allowing table creation (without explicit ENCRYPTION clause) when
    default_table_encryption is ONLINE_TO_KEYRING inside encrypted
    tablespaces.
    
    Also I have allowed tables that do not have ENCRYPTION clause specified
    to be created inside encrypted tablespace when d_t_e is equal to
    ONLINE_TO_KEYRING. This follows the rule that when d_t_e is ON such
    tables are also allowed to be created in encrypted tablespaces.
    This might be bit controversial for scenarios when table without
    ENCRYPTION clause is created inside Master Key encrypted tablespace.
    Such table will be allowed with d_t_e=ONLINE_TO_KEYRING. The reasoning
    behind is that when user sets d_t_e=ONLINE_TO_KEYRING the Master Key
    encrypted tablespace should be encrypted with KEYRING anyways.
    
    6. ENCRYPTION_KEY_ID should not appear in SHOW CREATE TABLE if it was
    not explicitly specifed.
    
    When table was encrypted by encryption threads ENCRYPTION_KEY_ID was
    added to SHOW CREATE TABLE statement. This was inconsistent as SHOW
    CREATE TABLE should only show values that table was CREATE/ALTERed
    with.


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/455/

sql/dd_table_share.cc
```diff
@@ -759,6 +759,11 @@ static bool fill_share_from_dd(THD *thd, TABLE_SHARE *share,
     table_options->get("encryption_key_id", &share->encryption_key_id);
   }
 
+  if (table_options->exists("explicit_encryption")) {
+    table_options->get("explicit_encryption", &bool_opt);
+    share->explicit_encryption = bool_opt;
```
dutow: Why not `table_options->get("explicit_encryption", &share->explicit_encryption);`

dutow (commented):
Looks good

sql/dd_table_share.cc
robgolebiowski: fixed :)

sql/handler.cc
```diff
@@ -2705,6 +2705,9 @@ void HA_CREATE_INFO::init_create_options_from_share(const TABLE_SHARE *share,
     // Assert to check that used_fields flag and encrypt_type are in sync
     DBUG_ASSERT(!encrypt_type.str);
     encrypt_type = share->encrypt_type;
+    explicit_encryption = share->explicit_encryption;
+  } else {
+    explicit_encryption = true;
```
satya-bodapati, line 2709: Regarding the else part:

Does this mean ENCRYPTION='Y' used ?

will  you set  explicit_encryption=true for ALTER TABLE t1 ENGINE=InnoDB ENCRYPTION='Y' as well?

mysql-test/suite/encryption/t/bug_ps_6106.test
```diff
@@ -0,0 +1,111 @@
...
+--let $t3_ibd = $MYSQLD_DATADIR/$DB_NAME/t3.ibd
+--let $t4_ibd = $MYSQLD_DATADIR/$DB_NAME/t4.ibd
+--let $t4_dup_ibd = $MYSQLD_DATADIR/$DB_NAME/t4_dup.ibd
+
+CREATE TABLE t1 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB ENCRYPTION='KEYRING';
+CREATE TABLE t2 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB COMPRESSION="zlib";
+CREATE TABLE t3 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB;
+CREATE TABLE t4 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255)) ENGINE=InnoDB ENCRYPTION='N';
+CREATE TABLE t4_dup LIKE t4;
+
```
satya-bodapati: can you add another table t5 that inherits encryption from schema with 'default encryption=n' ?

mysql-test/suite/encryption/t/bug_ps_6106.test
```diff
@@ -0,0 +1,111 @@
...
+--let $assert_text = Make sure t4 is not encrypted
+--let $assert_cond = [SELECT COUNT(*) FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION = 0 AND NAME = "$DB_NAME/t4"] = 1
+--source include/assert.inc
+
+--let $assert_text = Make sure t4_dup is not encrypted
+--let $assert_cond = [SELECT COUNT(*) FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION = 0 AND NAME = "$DB_NAME/t4_dup"] = 1
+--source include/assert.inc
+
+ALTER TABLE t1 ENGINE InnoDB;
+SHOW CREATE TABLE t1;
```
satya-bodapati: can you also test with ALTER TABLE t1 with explicit encryption on the alters, both ALTER TABLE ENCRYPTION='Y' and ENCRYPTION='N'.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -2688,9 +2688,9 @@ bool Encryption::is_master_key_encryption(const char *algorithm) {
 /** Check if the NO algorithm was explicitly specified.
 @param[in]      algorithm       Encryption algorithm to check
```
satya-bodapati, line 2711: add doxygen comment

satya-bodapati (changes requested):

sql/handler.cc
satya-bodapati: may be we should rename DD attribute. instead of explicit_encryption, it is better to indicate th type in name. `explicit_no_encryption`

robgolebiowski: I think it is better this way. It is more generic and thus may be reused in the future. explicit_encryption tells us that ENCRYPTION clause what used. To check whether it was N or Y we can check DD encrypt_type.

mysql-test/suite/encryption/t/bug_ps_6106.test
robgolebiowski: With what we discussed on slack - we will make default encryption = n not inheritable and user needs to set ENCRYPTION='N' explicitly to exclude table from rotation. I will add this to the test.

mysql-test/suite/encryption/t/bug_ps_6106.test
robgolebiowski: TABLE_ENCRYPTION=Y is immediately re-encrypted with encryption threads since they are on. I have added ENCRYPTION='N' though.

storage/innobase/handler/ha_innodb.cc
robgolebiowski: done

mysql-test/suite/encryption/t/bug_ps_6106.test
```diff
@@ -0,0 +1,145 @@
...
+
+SHOW CREATE TABLE t4_dup;
+
+# TODO: Setting DEFAULT ENCRYPTION with D_T_E=ONLINE_TO_KEYRING should not be possible
+# once PS-5323 is implemented
+CREATE DATABASE test_enc_n DEFAULT ENCRYPTION='N';
+USE test_enc_n;
+CREATE TABLE t5 (id INT NOT NULL PRIMARY KEY, a VARCHAR(255));
+USE test;
+
```
satya-bodapati: can you add the other inheritance scenario.

From WL:
C) Behavior of CREATE|ALTER TABLESPACE:

 1. CREATE TABLESPACE without explicit ENCRYPTION clause
    always succeeds and uses 'default-table-encryption' as
    tablespace encryption property. The ENCRYPTION clause would
    always be stored with tablespace metadata hence forth.


 2. CREATE/ALTER TABLESPACE with explicit ENCRYPTION clause
    succeeds and uses value specified in this clause as
    tablespace encryption property unless following condition is
    met,

    a) User misses TABLE_ENCRYPTION_ADMIN privilege
    b) AND 'table-encryption-privilege-check' is 'Y'
    c) AND this value doesn't match 'default-table-encryption' value,

    In the latter case we fail with an appropriate error message.

 3. ALTER TABLESPACE without explicit ENCRYPTION clause will retain
    the original ENCRYPTION type of tablespace

 4. ALTER TABLESPACE which encrypts tablespace would succeed,
    unless following condition is met,

mysql-test/suite/encryption/r/bug_ps_6106.result
```diff
@@ -0,0 +1,101 @@
...
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci ENCRYPTION='KEYRING' ENCRYPTION_KEY_ID=0
+ALTER TABLE t2 ENGINE InnoDB;
+SHOW CREATE TABLE t2;
+Table	Create Table
+t2	CREATE TABLE `t2` (
+  `id` int(11) NOT NULL,
+  `a` varchar(255) DEFAULT NULL,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci COMPRESSION='zlib' ENCRYPTION_KEY_ID=0
```
satya-bodapati: why are we seeing ENCRYPTION_KEY_ID=0 but not ENCRYPTION=KEYRING (after background threads finished encryption)?

mysql-test/suite/encryption/r/bug_ps_6106.result
```diff
@@ -0,0 +1,101 @@
...
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
+ALTER TABLE t4_dup ENGINE InnoDB;
+SHOW CREATE TABLE t4_dup;
+Table	Create Table
+t4_dup	CREATE TABLE `t4_dup` (
+  `id` int(11) NOT NULL,
+  `a` varchar(255) DEFAULT NULL,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
```
satya-bodapati: can you alter default table encryption of database and ensure ENCRYPTION='N' is shown for t4 and t4_dup?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -13084,9 +13098,26 @@ void ha_innobase::adjust_encryption_options(HA_CREATE_INFO *create_info,
...
+  if (!create_info->explicit_encryption &&
+      create_info->used_fields & HA_CREATE_USED_ENCRYPT) {
+#ifdef UNIV_DEBUG
+    ut_ad(table_def->options().exists("explicit_encryption"));
+    bool explicit_encryption{false};
+    table_def->options().get("explicit_encryption", &explicit_encryption);
+    ut_ad(explicit_encryption);
+#endif /* UNIV_DEBUG */
+    create_info->explicit_encryption = true;
+  }
```
satya-bodapati: Looks like DD already has this? can't we do this at server layer?

satya-bodapati (changes requested):
Almost there

Can you rename to test to use name with words? old name was fine: explicit_encryption_keyring may be?

and use the below inside test file
``` 
--echo #
--echo # PS-BUG
--echo  #
```

mysql-test/suite/encryption/r/bug_ps_6106.result
satya-bodapati: I mean for SHOW CREATE TABLE

storage/innobase/handler/ha_innodb.cc
robgolebiowski: Here we use DD in debug mode to check whether explicit_encryption was set when ENCRYPTION clause was used). On server level were CREATE_INFO is amended like in function: mysql_create_table_no_lock I do not have access to DD. I can move the create_info->explicit_encryption correction to this function and leave the assert here. Or I can leave both here. Which way to go ?

satya-bodapati: Not sure, if you are saying it is not possible to do in server.

How are other create_info attributes filled from DD? We can think of this as attribute not related to storage engines.

If we do at server, other SEs can utilize too.

Extra asserts at SE layer is fine

mysql-test/suite/encryption/t/bug_ps_6106.test
robgolebiowski: The ALTER TABLESPACE ENCRYPTION='N' for tablespace encrypted by encryption threads is not possible. For this user needs to first decrypt the tablespace with encryption threads. This was added as part of PS-5819: Information which tablespace is excluded from encryption threads should be visible via I_S.

mysql-test/suite/encryption/r/bug_ps_6106.result
robgolebiowski: ENCRYPTION=KEYRING is for tables created with explicit ENCRYPTION='KEYRING' clause. This is for foreground encryption like with MK : ENCRYPTION='Y'. Those tables cannot be decrypted by background threads. We see ENCRYPTION_KEY_ID - which shows us the id of the key encryption threads used to encrypt table.

mysql-test/suite/encryption/r/bug_ps_6106.result
robgolebiowski: done

storage/innobase/handler/ha_innodb.cc

mysql-test/suite/encryption/t/bug_ps_6106.test
satya-bodapati: ok, I understand that part. My question is how encryption threads skip general tablespace which are created with CREATE TABLESPCE ENCRYPTION='N' ? if it relies on crypt_data, the eventual goal is to rely on DD attribute. Can we have this attribute in tablespace (dd::Tablespace) as well?

mysql-test/suite/encryption/r/bug_ps_6106.result
satya-bodapati: Ok, this looks like separate issue to discuss. My question is ideally we should hide ENCRYPTION_KEY_ID from SHOW CREATE TABLE.

There should be two approached IMHO.
1) After background threads encrypt table, ENCRYPTION='KEYRING' should be shown along with ENCRYPTION_KEY_ID. Whats the benefit in hiding them? 
or
2) Hide everything related to background keyring encryption. (Although I wonder whats the benefit)

satya-bodapati:
> Almost there
> 
> Can you rename to test to use name with words? old name was fine: explicit_encryption_keyring may be?
> 
> and use the below inside test file
> 
> ```
> --echo #
> --echo # PS-BUG
> --echo  #
> ```

can you please address this comment as well

satya-bodapati:
1)
Its better to add explicit_encryption for all ENCRYPTION='N' tablespaces. i.e. both file_per_table tablespaces and General tablespaces.

Whether we use it or not for GT (right now) is different question

2) And for explicit_encryption to be useful (and help us in get rid of crypt_data for encryptio='n') tablespaces, we need the attribute in dd::Tablespace as well (apart from dd::Table)

Another PR would use dd::Tablespace attribute or transform the mapping to fil_space_t and use it.

mysql-test/suite/encryption/r/bug_ps_6106.result
robgolebiowski: ENCRYPTION='KEYRING' is for tables that were either CREATEd or ALTERed with encryption clause = 'KEYRING'. Those tables are always encrypted with KEYRING and never decrypted by encryption threads. In other words they are always skipped by encryption threads.
ENCRYPTION_KEY_ID on the other hand is used by encryption threads - this is the key_id that is going to be used by encryption threads when encrypting tables - thus it can be shown in SHOW CREATE TABLE.
If we shown ENCRYPTION='KEYRING' for tables that were encrypted by encryption threads we would be saying that those tables where CREATEd or ALTERed to ENCRYPTION='KEYRING' which is not true and has different meaning.

mysql-test/suite/encryption/t/bug_ps_6106.test
robgolebiowski: done

mysql-test/suite/encryption/r/bug_ps_6106.result
robgolebiowski: I will remove ENCRYPTION_KEY_ID if it was not set during CREATE or ATLER table. SHOW CREATE TABLE should reflect the statement with which a table was created with.

sql/sql_tablespace.cc
```diff
@@ -510,6 +515,8 @@ bool Sql_cmd_create_tablespace::execute(THD *thd) {
   // - Disallow encryption='y', if SE does not support it.
   if (hton->flags & HTON_SUPPORTS_TABLE_ENCRYPTION) {
     tablespace->options().set("encryption", encrypt_type);
+    tablespace->options().set("explicit_encryption",
+                              m_options->encryption.str ? true : false);
```
satya-bodapati, line 517: Only create path is handled. ALTER TABLESPACE should also set this? :thinking:

sql/sql_tablespace.cc
```diff
@@ -470,7 +470,12 @@ bool Sql_cmd_create_tablespace::execute(THD *thd) {
     encrypt_tablespace = dd::is_encrypted(m_options->encryption);
     encrypt_type = dd::make_string_type(m_options->encryption);
   } else {
-    encrypt_tablespace = thd->variables.default_table_encryption;
+    encrypt_tablespace =
+        thd->variables.default_table_encryption == DEFAULT_TABLE_ENC_ON ||
+        thd->variables.default_table_encryption ==
+            DEFAULT_TABLE_ENC_KEYRING_ON ||
+        thd->variables.default_table_encryption ==
+            DEFAULT_TABLE_ENC_ONLINE_TO_KEYRING;
```
satya-bodapati: Is this a different bug? handled in this PR?

mysql-test/suite/encryption/t/explicit_encryption.test
```diff
@@ -24,6 +26,31 @@ CREATE TABLESPACE ts1;
 CREATE TABLESPACE ts_encryption_n ADD DATAFILE 'ts_encryption_n.ibd' ENCRYPTION='N';
```
satya-bodapati, line 36: also handle ALTER TABLESPACE

satya-bodapati (changes requested):

mysql-test/suite/encryption/t/bug_ps_6106.test
satya-bodapati: thanks

mysql-test/suite/encryption/r/bug_ps_6106.result
satya-bodapati: Okay. I see them as design issues inherited from upstream(MariaDB). There is too much detail (what is explicit and implicit, which tables are encrypted by background) left to user to possibly reduce scope.

You need not fix this in this PR

mysql-test/suite/encryption/t/explicit_encryption.test
robgolebiowski: done

sql/sql_tablespace.cc
robgolebiowski: yes. Basically the sync between global and session scope for d_t_e still needs some work. I have added PS-6759: Sync global value of default_table_encryption with session value of this variable to address this issue.

sql/sql_tablespace.cc
robgolebiowski: I have added handling the ALTER path.

sql/sql_tablespace.cc
satya-bodapati: Ok, will you rever this change here( and handle in PS-6759) or it is necessary for this PR?

mysql-test/suite/encryption/r/bug_ps_6106.result
robgolebiowski: I have already done it in this PR :)

sql/sql_tablespace.cc
robgolebiowski: I have added the minimal change here - just to instruct the encryption of tablespaces based on d_t_e. Previously encrypt_tablespace = thd->variables.default_table_encryption always evaluated to true for values different than 'N'. So for ONLINE_FROM_KEYRING_TO_UNENCRYPTED the old value of encrypt_tablespace was true - and should be false.

sql/sql_tablespace.cc
```diff
@@ -934,11 +941,15 @@ bool Sql_cmd_alter_tablespace::execute(THD *thd) {
...
+      tsmp.second->options().set("explicit_encryption", true);
       if (set_table_encryption_type(thd, *tsmp.first, &trefs,
                                     &table_object_pairs, m_options->encryption,
                                     &table_mdl_reqs))
         return true;
     }
+
+  } else if (hton->flags & HTON_SUPPORTS_TABLE_ENCRYPTION) {
+    tsmp.second->options().set("explicit_encryption", false);
   }
```
satya-bodapati: although this is correct, can you rewrite this a bit.

if (hton->flags & HTON_SUPPORTS_ENCRYPTION) {

  if ( m_options->encryption.str) {
    other stuff
   set explicit_encryption =true;
  } else {
  set explicit_encryption = false
  }
}

ALTER TABELSPACE ts1 ENGINE=InnoDB--> this valid SQL? Does it translate to ENCRYPTION='N'?

mysql-test/suite/encryption/r/bug_ps_6106.result
satya-bodapati: Ok, hopefully it is small change.. otherwise, its better if it is done separately

mysql-test/suite/encryption/t/explicit_encryption.test
```diff
@@ -0,0 +1,249 @@
...
+ALTER TABLE t4_dup ENGINE InnoDB;
+SHOW CREATE TABLE t4_dup;
+ALTER TABLE t5 ENGINE InnoDB;
+SHOW CREATE TABLE t5;
+ALTER TABLE test_enc_n.t6 ENGINE InnoDB;
+SHOW CREATE TABLE test_enc_n.t6;
+ALTER TABLESPACE ts1 ENGINE InnoDB;
+ALTER TABLESPACE ts_encryption_n ENGINE InnoDB;
+ALTER TABLESPACE ts_d_t_e_off ENGINE InnoDB;
+ALTER TABLESPACE ts_altered_to_n ENGINE InnoDB;
```
satya-bodapati, line 155: Hmm.. so part of my question is resolved. Is this translated to ENCRYPTION='N' ?

sql/sql_table.cc
```diff
@@ -8743,10 +8742,17 @@ static bool validate_table_encryption(THD *thd, HA_CREATE_INFO *create_info) {
...
   */
   bool requested_type = dd::is_encrypted(create_info->encrypt_type);
+
   if ((uses_general_tablespace || uses_system_tablespace) &&
-      requested_type != uses_encrypted_tablespace) {
+      ((requested_type != uses_encrypted_tablespace) &&
+       (!uses_encrypted_tablespace ||
+        global_system_variables.default_table_encryption !=
+            DEFAULT_TABLE_ENC_ONLINE_TO_KEYRING))) {
     my_error(ER_INVALID_ENCRYPTION_REQUEST, MYF(0),
```
satya-bodapati, line 8756: this is related to other PS-6759. Can you please back out all related changes from this PR and move it to PS-6759?

sql/sql_tablespace.cc
```diff
@@ -470,7 +470,12 @@ bool Sql_cmd_create_tablespace::execute(THD *thd) {
     encrypt_tablespace = dd::is_encrypted(m_options->encryption);
     encrypt_type = dd::make_string_type(m_options->encryption);
   } else {
-    encrypt_tablespace = thd->variables.default_table_encryption;
+    encrypt_tablespace =
+        thd->variables.default_table_encryption == DEFAULT_TABLE_ENC_ON ||
+        thd->variables.default_table_encryption ==
+            DEFAULT_TABLE_ENC_KEYRING_ON ||
+        global_system_variables.default_table_encryption ==
+            DEFAULT_TABLE_ENC_ONLINE_TO_KEYRING;
```
satya-bodapati: PS-6759?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -11693,7 +11696,12 @@ inline MY_ATTRIBUTE((warn_unused_result)) int create_table_info_t::
   space_id_t space_id = 0;
   dd::Object_id dd_space_id = dd::INVALID_OBJECT_ID;
   ulint actual_n_cols;
-  fil_encryption_t keyring_encryption_option = FIL_ENCRYPTION_DEFAULT;
+
+  fil_encryption_t keyring_encryption_option =
+      Encryption::none_explicitly_specified(m_create_info->explicit_encryption,
+                                            m_create_info->encrypt_type.str)
+          ? FIL_ENCRYPTION_OFF
+          : FIL_ENCRYPTION_DEFAULT;
```
satya-bodapati, line 11704: Is this related to ENCRYPTION_KEY_ID change?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -12591,7 +12602,8 @@ bool create_table_info_t::create_option_tablespace_is_valid() {
 
   if (!m_use_shared_space) {
     if (!m_use_file_per_table) {
-      if (m_create_info->encrypt_type.str != nullptr && is_temp) {
+      if (m_create_info->encrypt_type.str != nullptr &&
+          m_create_info->explicit_encryption && is_temp) {
```
satya-bodapati, line 12606: How did this work before?
This change is to error out "CREATE TEMPORARY TABLE t1(a INT) ENCRYPTION=N" ?
Do we really need the explicit_encryption check here?

satya-bodapati (changes requested):

sql/sql_tablespace.cc
robgolebiowski: It is a valid one, although it gives a warning that it is a deprecated sql statement. The root cause of this bug is the same ALTER TABLE without ENCRYPTION clause would translate to ENCYRYPTION='N' - which was incorrect. It should not result in ENCRYPTION='N' I have added check if ts1 remains encrypted.

mysql-test/suite/encryption/t/explicit_encryption.test
robgolebiowski: no

sql/sql_table.cc
robgolebiowski: PS-6759 is more about sync between global and session. Here we should allow creating tables inside encrypted tablespace when d_t_e == ONLINE_TO_KEYRING.

sql/sql_tablespace.cc
robgolebiowski: PS-6759 is about syncing global and session. Here it should not be that d_t_e != OFF means encrypt_tablespace.

storage/innobase/handler/ha_innodb.cc
robgolebiowski: no, in this PR I need to retain explicit ENCRYPTION='N' when ALTERing a table - as I explained in commit message.

storage/innobase/handler/ha_innodb.cc
robgolebiowski: agreed, not needed, removed.

sql/sql_tablespace.cc
robgolebiowski: I had a test failing because of that.

sql/sql_table.cc
satya-bodapati: what is the SQL failing without this change (when dte= ONLINE_TO_KEYRING)?

storage/innobase/include/os0file.h
```diff
@@ -470,9 +470,12 @@ struct Encryption {
   @param[in]      algorithm       Encryption algorithm to check
```
satya-bodapati, line 472: missing doxygen comment for explicit_encryption parameter

satya-bodapati (changes requested):
This PR handles several small things. Can you please rewrite commit message with 'all the fixes' using numbering scheme.

AFAIK, the issues are:

1. Original ALTER issue
2. CREATE TABLE LIKE  (retain encryption='n' from original table) 
3. Tablespace encryption not following global value of default_table_encryption (Whats this?)
4. Allowing table creation (without explicit ENCRYPTION clause) when dte is ONLINE_to_Keyring
5. encryption_key_id changes in show create table output

Not a big fan of mixing things in one PR. But I will let it go for this time. Please do not add things unless there is agreement(Just wanted to highlight encryption_key_id change in SHOW CREATE TABLE and get agreement that is an issue or not first).

is this zombie file " mysql-test/suite/encryption/r/bug_ps_6106.result"  now ? the test is now explicit_encryption.test.

So few minor requests
1. fix Doxygen comment (highlighted in PR)
2. Rewrite commit to indicate it fixes 5 things (or more). Use numbering scheme
3. Remove zombie result file

sql/sql_table.cc
robgolebiowski: in encryption/tablespace_encryption.test :
CREATE TABLESPACE ts5 ENCRYPTION_KEY_ID=5;
CREATE TABLE t1_in_ts5 (a varchar(255)) TABLESPACE=ts5;
and default_table_encryption is set to ONLINE_TO_KEYRING
without this change the CREATE TABLE t1_in_ts5 will fail.

storage/innobase/include/os0file.h
robgolebiowski: done

satya-bodapati (approved):
I think I found one more issue.  explicit_encryption is not present in dd::Tablespace for file per table tablespaces. It has to be in sync with dd::Table explicit_encryption. (similar to encrypt_type which is present in both dd::Tablespace and dd::Table)

since this is absolutely not necessary for current PR and there will be one more PR which will translate dd::Tablespace to fil_space_t attribute, we can do this later.

Please file new JIRA issue for this

Hence 'ap'proved ;)

storage/innobase/handler/ha_innodb.cc
robgolebiowski: Correction - this is needed. CREATE TEMPORARY LIKE needs to retain encryption clause - even if it is N. Upstream did not care about the ENCRYPTION clause. We retain it. Thus here it will be not empty for implicit N (encrypt_type.str != nullptr ). We should fail only when ENCRYPTION was explicitly specified. Thus I need this check for it here.

PR #3552, opened 2019-12-05
PS-6027: Null merge PS-5.7.28 into 8.0

inikep:
Revert ssl.cmake to upstream:
https://github.com/mysql/mysql-server/blob/8.0/cmake/ssl.cmake#L223


percona-ysorokin (approved):
LGTM

PR #3557, opened 2019-12-11
PS-6093 Unhandled mutex in Aggregate StatsProblem:

rahulmalik87:
Problem
----
This is a regression from PS-5639.
Aggregate type mutexes were assumed to be used only by two mutexes
BUF_BLOCK_MUTEX and AUTO INC.
BUF_POOL_ZIP mutex was not considered as aggregate mutex
and Latch ID to name conversion didn't handle BUF_POOL_ZIP mutex.
Fix:
----
Handle LATCH_ID_BUF_POOL_ZIP in AggregateMutexStatsPolicy ID to name conversion

### DIFF
--- a/storage/innobase/include/sync0policy.ic
+++ b/storage/innobase/include/sync0policy.ic
@@ -45,6 +45,7 @@ std::string AggregateMutexStatsPolicy<Mutex>::to_string() const
 	switch (m_id) {
 
 	case LATCH_ID_BUF_BLOCK_MUTEX:
+	case LATCH_ID_BUF_POOL_ZIP:
 		/* I don't think it makes sense to keep track of the file name
 		and line number for each block mutex. Too much of overhead.
 		Use the latch id to figure out the location from the source. */


storage/innobase/include/sync0policy.ic
```diff
@@ -45,6 +45,7 @@ std::string AggregateMutexStatsPolicy<Mutex>::to_string() const
 	switch (m_id) {
 
 	case LATCH_ID_BUF_BLOCK_MUTEX:
+	case LATCH_ID_BUF_POOL_ZIP:
```
satya-bodapati, line 48: I wonder why upstream chose to use ib_bpmutex_t(AggreateStat Mutex) instead of regular ib_mutex_t.

We could use ib_mutex_t for zip mutex. There are not many of them.

At max 64 buffer pools possible. So max 64 zip mutexes.
So please try to iuse b_mutex_t for BUF_POOL_ZIP and see if that solves the problem

satya-bodapati (changes requested):

kamil-holubicki:
Fix for this issue will be delivered in PR's related to PS-6150.
5.7: https://github.com/percona/percona-server/pull/3594
8.0: https://github.com/percona/percona-server/pull/3601

PR #3564, opened 2019-12-12
PS-4693 PS - Explain what is meant by innodb_buffered_aio_submitted o…

patrickbirch:
…n in aio_page_requests doc

updated opening paragraphs
added example

### DIFF
--- a/doc/source/performance/aio_page_requests.rst
+++ b/doc/source/performance/aio_page_requests.rst
@@ -4,13 +4,13 @@
 Multiple page asynchronous I/O requests
 =======================================
 
-I/O unit size in |InnoDB| is only one page, even if doing read ahead. 16KB
-I/O unit size is too small for sequential reads, and much less efficient than
-larger I/O unit size.
+The I/O unit size in |InnoDB| is only one page, even if doing a read ahead. 
+The 16KB I/O unit is too small for sequential reads, and much less efficient
+than a larger I/O unit size.
 
-|InnoDB| uses Linux asynchronous I/O (``aio``) by default. By submitting multiple
-consecutive 16KB read requests at once, Linux internally can merge requests and
-reads can be done more efficiently.
+|InnoDB| uses Linux asynchronous I/O (``aio``) by default. By submitting
+multiple consecutive 16KB read requests at once, Linux internally can merge
+requests and reads can be done more efficiently.
 
 `On a HDD RAID 1+0 environment
 <http://yoshinorimatsunobu.blogspot.hr/2013/10/making-full-table-scan-10x-faster-in.html>`_,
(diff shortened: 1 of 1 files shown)


doc/source/performance/aio_page_requests.rst
```diff
@@ -34,7 +35,16 @@ Status Variables
   :vartype: Numeric
   :scope: Global
 
-This variable shows the number of submitted buffered asynchronous I/O requests.
+This variable shows the number of asynchronous I/O requests buffered in the
+Linux Native subsystem before submission.
```
dutow: I don't understand this change?

Looking at the status variable, it only increases at https://github.com/percona/percona-server/blob/5.7/storage/innobase/os/os0file.cc#L3103, and never decreases.

So this variable is updated after the request was submitted to the kernel, not before.
Also, it is an always increasing total submitted counter, not the number of entries we currently have in the buffer, which is suggested by this change.

doc/source/performance/aio_page_requests.rst
```diff
@@ -4,22 +4,23 @@
 Multiple page asynchronous I/O requests
```
dutow, line 4: Again, I don't understand the change in this section - why it needs to be changed at all?

The current text describes the multiple page readahead feature, which is an improvement in PS. The new change doesn't really explain this at all, and has some misinformation about how the (upstream) aio implementation works.

dutow (changes requested):
I don't think this is what PS-4693 is about.

The ticket says that the `buffered` part of the variable name is confusing, and should be explained. The current documentation is not wrong at all, what is already written is correct.

The only issue is that according to the ticket, users can misunderstand the meaning of this status variable, because it doesn't describe what's in the innodb buffers, but what's already submitted to the kernel buffers.

(I'm not sure if I agree with the ticket, I think the variable name is clear. But adding an additional sentence to clarify it wouldn't hurt)

doc/source/performance/aio_page_requests.rst
patrickbirch: Got it. Thanks.

doc/source/performance/aio_page_requests.rst
patrickbirch: OK.  Will fix.

doc/source/performance/aio_page_requests.rst
```diff
@@ -4,22 +4,20 @@
...
+multiple consecutive 16KB read requests at once, Linux internally can merge
+requests and reads can be done more efficiently.
 
 `On a HDD RAID 1+0 environment
 <http://yoshinorimatsunobu.blogspot.hr/2013/10/making-full-table-scan-10x-faster-in.html>`_,
 more than 1000MB/s disk reads can be achieved by submitting 64 consecutive pages
 requests at once, while only
 160MB/s disk reads is shown by submitting single page request.
 
-With this feature |InnoDB| submits multiple page I/O requests.
```
dutow: Why removed?

dutow (approved):
Looks good, but I'm not sure if we should remove that line (I added a comment there)

doc/source/performance/aio_page_requests.rst
patrickbirch: Added

patrickbirch:
waiting for approval or comments.

hrvojem (approved):
LGTM

PR #3565, opened 2019-12-13
PS-6125: Server crashes when querying ROCKSDB_GLOBAL_INFO table (5.7)

kamil-holubicki:
https://jira.percona.com/browse/PS-6125

When we set 'rocksdb_update_cf_options' and provided CF that does not exist, its in memory descriptor was created + entries were added to .rocksdb/OPTIONS-ABC and Manifest file, however its CF_DEFINITION_VERSION was not stored in DB.
After that, when we query information schema for CFs, we validated all in memory CF descriptors against what we have in DB and abort if information is not found in DB.

As the 'rocksdb_update_cf_options' is intended to change parameters of CFs (not create new ones), it was added detection of nonexisting CFs passed as the value of 'rocksdb_update_cf_options'. In case of any CF does not exist, error is returned and no parameters are changed.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -13081,13 +13081,28 @@ static int rocksdb_validate_update_cf_options(
     my_error(ER_WRONG_VALUE_FOR_VAR, MYF(0), "rocksdb_update_cf_options", str);
     return HA_EXIT_FAILURE;
   }
-  // Loop through option_map and create missing column families
-  for (Rdb_cf_options::Name_to_config_t::iterator it = option_map.begin();
-       it != option_map.end(); ++it) {
-    if (!cf_manager.get_or_create_cf(rdb, it->first,
-                                     !rocksdb_no_create_column_family)) {
-      return HA_EXIT_FAILURE;
+  // Loop through option_map and check if all specified CFs exist.
+  std::vector<const std::string *> unknown_cfs;
+  for (const auto &option : option_map) {
+    if (!cf_manager.get_cf(option.first)) {
+      unknown_cfs.push_back(&(option.first));
     }
+  }
+
+  if (!unknown_cfs.empty()) {
+    std::string err(str);
+    err.append(" Unknown CF: ");
+    bool first = true;
+    for (const auto cf : unknown_cfs) {
+      if (first)
+        first = false;
+      else
+        err.append(", ");
+      err.append(*cf);
+    }
+    my_error(ER_WRONG_VALUE_FOR_VAR, MYF(0), "rocksdb_update_cf_options",
+             err.c_str());
+    return HA_EXIT_FAILURE;
   }
   return HA_EXIT_SUCCESS;
 }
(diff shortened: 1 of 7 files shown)


kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-param/194/

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -13081,13 +13081,29 @@ static int rocksdb_validate_update_cf_options(
...
-  // Loop through option_map and create missing column families
+  // Loop through option_map and check if all specified CFs exist.
+  std::vector<std::string> unknown_cfs;
   for (Rdb_cf_options::Name_to_config_t::iterator it = option_map.begin();
        it != option_map.end(); ++it) {
-    if (!cf_manager.get_or_create_cf(rdb, it->first,
-                                     !rocksdb_no_create_column_family)) {
-      return HA_EXIT_FAILURE;
+    if (!cf_manager.get_cf(it->first)) {
+      unknown_cfs.push_back(it->first);
```
dutow: Do we have to copy the string here? Looks like the lifetime of option_map is longer than unknown_cfs

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -13081,13 +13081,29 @@ static int rocksdb_validate_update_cf_options(
...
+    }
+  }
+
+  if (unknown_cfs.size() > 0) {
+    std::string err(str);
+    err.append(" Unknown CF: ");
+    bool first = true;
+    for (const auto cf : unknown_cfs) {
```
dutow, line 13096: Wouldn't it be better to use stringstream + std::copy + ostream_iterator instead of the for loop?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -13081,13 +13081,29 @@ static int rocksdb_validate_update_cf_options(
...
        it != option_map.end(); ++it) {
-    if (!cf_manager.get_or_create_cf(rdb, it->first,
-                                     !rocksdb_no_create_column_family)) {
-      return HA_EXIT_FAILURE;
+    if (!cf_manager.get_cf(it->first)) {
+      unknown_cfs.push_back(it->first);
+    }
+  }
+
+  if (unknown_cfs.size() > 0) {
```
dutow: `!unknown_cft.empty()`

storage/rocksdb/ha_rocksdb.cc
kamil-holubicki: Not a mission critical part :-) , but I agree, we can collect pointers only.
DONE.

storage/rocksdb/ha_rocksdb.cc
kamil-holubicki: done

storage/rocksdb/ha_rocksdb.cc
kamil-holubicki: I'm not sure it it will make generation of error message easier to read. We need to deal with trailing coma.
ostringstream can be also used to construct the final string, but again, I'm not sure if it is worth introducing dependency to ostringstream (other parts in this file use string concatenation by string::append())

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -13081,13 +13081,29 @@ static int rocksdb_validate_update_cf_options(
...
   }
-  // Loop through option_map and create missing column families
+  // Loop through option_map and check if all specified CFs exist.
+  std::vector<const std::string *> unknown_cfs;
   for (Rdb_cf_options::Name_to_config_t::iterator it = option_map.begin();
        it != option_map.end(); ++it) {
-    if (!cf_manager.get_or_create_cf(rdb, it->first,
-                                     !rocksdb_no_create_column_family)) {
-      return HA_EXIT_FAILURE;
+    if (!cf_manager.get_cf(it->first)) {
```
george-lorch: Since we're talking 'modern C++' a bit more these days, why not take this opportunity to change this to `for (auto& it : option_map)` paradigm?

george-lorch (approved):

storage/rocksdb/ha_rocksdb.cc
kamil-holubicki: done

kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-param/195

dutow (approved):

PR #3566, opened 2019-12-13
PS-1761 LP #1640062: Document that QRT accounts query execution time …

patrickbirch:
…after the locks are taken

Added sentence to opening paragraph based on the description in the Jira ticket.

### DIFF
--- a/doc/source/diagnostics/response_time_distribution.rst
+++ b/doc/source/diagnostics/response_time_distribution.rst
@@ -4,7 +4,7 @@
  Response Time Distribution
 ============================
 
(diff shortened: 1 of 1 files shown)


doc/source/diagnostics/response_time_distribution.rst
```diff
@@ -4,7 +4,7 @@
...
+The slow query log provides exact information about queries that take a long time to execute. However, sometimes there are a large number of queries that each take a very short amount of time to execute. This feature provides a tool for analyzing that information by counting and displaying the number of queries according to the length of time they took to execute. The query execution time begins when the initial locks are acquired. The user can define time intervals that divide the range 0 to positive infinity into smaller intervals and then collect the number of commands whose execution times fall into each of those intervals.
```
dutow: I think "after" would be clearer here than "when".

patrickbirch: fixed to use "after".

dutow (approved):

PR #3574, opened 2019-12-17
PS-6125: Server crashes when querying ROCKSDB_GLOBAL_INFO table (8.0)

kamil-holubicki:
https://jira.percona.com/browse/PS-6125

When we set 'rocksdb_update_cf_options' and provided CF that does not exist, its in memory descriptor was created + entries were added to .rocksdb/OPTIONS-ABC and Manifest file, however its CF_DEFINITION_VERSION was not stored in DB.
After that, when we query information schema for CFs, we validated all in memory CF descriptors against what we have in DB and abort if information is not found in DB.

As the 'rocksdb_update_cf_options' is intended to change parameters of CFs (not create new ones), it was added detection of nonexisting CFs passed as the value of 'rocksdb_update_cf_options'. In case of any CF does not exist, error is returned and no parameters are changed.


george-lorch (approved):

kamil-holubicki:
https://ps80.cd.percona.com/job/percona-server-8.0-param/472/

dutow (approved):

PR #3575, opened 2019-12-17
PS-1466 LP #1264306: Per Session innodb_flush_log_at_trx_commit poorl…

patrickbirch:
…y documented

The variable documented is innodb_use_global_flush_log_at_trx_commit
Rewrote the description based on investigation and comments in the Jira ticket

### DIFF
--- a/doc/source/scalability/innodb_io.rst
+++ b/doc/source/scalability/innodb_io.rst
@@ -31,13 +31,27 @@ System Variables
    :default: True
    :range: True/False
 
-This variable is used to control the ability of the user to set the value of the global |MySQL| variable ``innodb_flush_log_at_trx_commit``.
+This variable enables or disables the effect of the per-session value of 
+the `innodb_flush_log_at_trx_commit` variable. 
 
-If ``innodb_use_global_flush_log_at_trx_commit=0`` (False), the client can set the global |MySQL| variable, using: ::
+If the global variable  `innodb_use_global_flush_log_at_trx_commit` is
+set to ``1``, the session will always use the current
+global value of `innodb_flush_log_at_trx_commit`. This is the
+upstream compatible mode. If the user attempts to change the
+`innodb_flush_log_at_trx_commit` value for a
+session, the session value is ignored.
 
-  SET innodb_use_global_flush_log_at_trx_commit=N
+If the global variable `innodb_use_global_flush_log_at_trx_commit` is set to
+``0``, a user can modify the
+``innodb_flush_log_at_trx_commit`` per-session using the following command:
 
-If ``innodb_use_global_flush_log_at_trx_commit=1`` (True), the user session will use the current value of ``innodb_flush_log_at_trx_commit``, and the user cannot reset the value of the global variable using a ``SET`` command.
+.. code-block:: MySQL
+
+    SET SESSION innodb_flush_log_at_trx_commit=0
+
+This modification only affects the transactions in that session. Other sessions,
+if they have not been individually modified, continue to use the
+global `innodb_use_flush_log_at_trx_commit` value.
 
 .. variable:: innodb_log_block_size
 


doc/source/scalability/innodb_io.rst
```diff
@@ -31,13 +31,32 @@ System Variables
...
 
-This variable is used to control the ability of the user to set the value of the global |MySQL| variable ``innodb_flush_log_at_trx_commit``.
+This variable enables or disables the effect of the per-session value of 
+the `innodb_flush_log_at_trx_commit` variable. 
 
-If ``innodb_use_global_flush_log_at_trx_commit=0`` (False), the client can set the global |MySQL| variable, using: ::
+The following are the available values:
 
-  SET innodb_use_global_flush_log_at_trx_commit=N
+    * ``1`` - Evaluated as ``TRUE`` and is the default value
```
dutow: It's not evaluated as - this is a boolean variable, the possible values are true and false. (which is represented as 0/1, but that's not really important)

I don't think this and the next line adds any information - the possible values and the default is already mentioned a few paragraphs above.

doc/source/scalability/innodb_io.rst
```diff
@@ -31,13 +31,32 @@ System Variables
...
+    * ``0`` - Evaluated as ``FALSE``
 
-If ``innodb_use_global_flush_log_at_trx_commit=1`` (True), the user session will use the current value of ``innodb_flush_log_at_trx_commit``, and the user cannot reset the value of the global variable using a ``SET`` command.
+If the global variable  `innodb_use_global_flush_log_at_trx_commit` is
+set to ``1``, the session will always use the current
+global value of `innodb_flush_log_at_trx_commit`. A user cannot reset the value
+with a ``SET`` command for a separate session. Under this mode, changing the
```
dutow: What does "for a separate session" mean? Or reset? I don't understand this sentence.

doc/source/scalability/innodb_io.rst
```diff
@@ -31,13 +31,32 @@ System Variables
...
+session variable for ``innodb_flush_log_at_trx_commit`` does not affect
```
dutow: session value of

doc/source/scalability/innodb_io.rst
```diff
@@ -31,13 +31,32 @@ System Variables
...
+upstream compatibility.
```
dutow: maybe "does not affect the value used by the session"? Or simply "this is the upstream compatible mode".

doc/source/scalability/innodb_io.rst
```diff
@@ -31,13 +31,32 @@ System Variables
...
+
+If the global variable `innodb_use_global_flush_log_at_trx_commit` is set to
+``0``, a user can modify the
```
dutow, line 45: Users can modify it anyway, even if the variable is true - but in that case, the session value is ignored.

dutow (changes requested):

doc/source/scalability/innodb_io.rst
patrickbirch: Removed.

doc/source/scalability/innodb_io.rst
patrickbirch: Removed.

doc/source/scalability/innodb_io.rst
patrickbirch: Updated.

doc/source/scalability/innodb_io.rst
patrickbirch: Rewritten to "This is the upstream compatible mode."

doc/source/scalability/innodb_io.rst
patrickbirch: Added to the `1` paragraph.

dutow (approved):

PR #3583, opened 2019-12-18
PS-6733: Use binary search in data masking dictionaries

dutow:


satya-bodapati (approved):
Please add decent commit msg.

Highlight the fact that loading the list is one time operation. So sort is done only one time
and gen_blacklist() is used multiple times and binary search  saves the time for multiple queries.

plugin/data_masking/src/udf/udf_gen_dictionary_load.cc
```diff
@@ -111,6 +111,7 @@ static std::string _gen_dictionary_load(const char *dictionary_path,
         if (s.length() > 0) list.push_back(s);
       }
       if (list.size() > 0) {
+        std::sort(list.begin(), list.end());
```
percona-ysorokin, line 114: Can we make this list a `std::unordered_set<std::string>`?

dutow: No, there is a function that selects a randomly chosen item from a dictionary. We could make it an unordered_set and a vector, but that would increase the memory usage a bit...

PR #3587, opened 2019-12-23
PS-6112: Inconsistent Binlog_snapshot_gtid when mysqldump used with --single-transaction

kamil-holubicki:

The fix is to lock binglog, then start transaction, capture gtid_executed, then unlock binglog to allow parallel transactions to run when dumping db in repeatable read transaction.

If server does not support backup locks instead of locking binlog we lock all tables for the time of capturing gtid_executed (the same behavior as upstream).

Two options were added to mysqldump: start-sql-file and finish-sql-file. This allows better integration of mysqldump with mtr tests as debug sync facility asserver sync points are activated per thread.


kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-param/202/
Failed tests not related to the change.

client/mysqldump.c
```diff
@@ -7107,12 +7128,64 @@ static my_bool server_supports_backup_locks(void)
...
+    return 1;
+  }
+
+  while (fgets(buf, SQL_STATEMENT_MAX_LEN, file))
+  {
+    // simple validation
+    size_t query_len= strlen(buf);
+    if (*buf == '\n')
+      continue;
+    if (buf[query_len - 2] != ';' && buf[query_len - 1] != '\n' &&
```
dutow: Also check query_len >= 2. (and: consider supporting windows line endings)

client/mysqldump.c
```diff
@@ -570,6 +572,16 @@ static struct my_option my_long_options[] =
   {"secure-auth", OPT_SECURE_AUTH, "Refuse client connecting to server if it"
     " uses old (pre-4.1.1) protocol. Deprecated. Always TRUE",
     &opt_secure_auth, &opt_secure_auth, 0, GET_BOOL, NO_ARG, 1, 0, 0, 0, 0, 0},
+  {"start-sql-file", OPT_START_SQL_FILE, "Execute SQL statements from the file"
```
dutow: Shouldn't we limit these options to debug builds? As I can't imagine a production use case for them.

kamil-holubicki: It is not a problem to do this, but this will lead to the series of #ifdefs which are code smells. Current implementation is generic, so it says nothing about Debug Sync. It is just executing sql file on start and on the end. How to use it in production? I don't know. i.e. you can drop all tables after doing dump ;-). I would say that adding it permanently does not cause problems. Additionally it allows us to test mysqldump in release builds as well. 
@percona-ysorokin what is your opinion?

client/mysqldump.c
kamil-holubicki: done

client/mysqldump.c
percona-ysorokin: I only have security concerns. Imagine there is a web backup tool that allows to specify additional options to `mysqldump`. Potentially I can do something like `mysqldump --start-sql-file=/etc/shadow ...` and analyze error log to extract data.

client/mysqldump.c
```diff
@@ -570,6 +572,16 @@ static struct my_option my_long_options[] =
...
+    " at the mysqldump start. Each line has to contain one statement finished"
```
percona-ysorokin:
```suggestion
    " at the mysqldump start. Each line has to contain one statement terminated"
```

client/mysqldump.c
```diff
@@ -570,6 +572,16 @@ static struct my_option my_long_options[] =
...
+    " with the semicolon. Line length limit is 1023 characters.",
```
percona-ysorokin:
```suggestion
    " with a semicolon. Line length limit is 1023 characters.",
```

client/mysqldump.c
```diff
@@ -570,6 +572,16 @@ static struct my_option my_long_options[] =
...
+    &start_sql_file, &start_sql_file, 0, GET_STR,
+   REQUIRED_ARG, 0, 0, 0, 0, 0, 0},
+  {"finish-sql-file", OPT_FINISH_SQL_FILE, "Execute SQL statements from the file"
+    " at the mysqldump finish. Each line has to contain one statement finished"
```
percona-ysorokin: likewise

client/mysqldump.c
```diff
@@ -570,6 +572,16 @@ static struct my_option my_long_options[] =
...
+    " with the semicolon. Line length limit is 1023 characters.",
```
percona-ysorokin: likewise

mysql-test/std_data/mysqldump_gtid_finish.sql
```diff
@@ -0,0 +1,4 @@
+# This file will be executed at the end of mysqldump
+
+SET DEBUG_SYNC='mysqldump SIGNAL mysqldump.finished';
```
percona-ysorokin: I might be missing something but I don't see `DEBUG_SYNC` point named `mysqldump` in the code

mysql-test/t/mysqldump_gtid.inc
```diff
@@ -0,0 +1,69 @@
+--source include/have_log_bin.inc
+--source include/have_innodb.inc
+--source include/not_embedded.inc
+--source include/have_debug_sync.inc
+
+--echo #################################################
+--echo # Testing mysqldump with options: $MYSQLDUMP_OPTS
+--echo #################################################
+
+--let sleep_sec = 5
```
percona-ysorokin:
```suggestion
--let $sleep_sec = 5
```

mysql-test/t/mysqldump_gtid.inc
```diff
@@ -0,0 +1,69 @@
...
+
+--connect(con1, localhost, root)
```
percona-ysorokin, line 22: Add `count_sessions.inc` / `wait_count_sessions.inc`

mysql-test/t/mysqldump_gtid.inc
```diff
@@ -0,0 +1,69 @@
...
+INSERT INTO t1 VALUES (0);
+
+# Capture current gtid
+--let $gtid_start = `SELECT @@GLOBAL.GTID_EXECUTED`
+
+# Start mysqldump
+--let $file = $MYSQLTEST_VARDIR/tmp/dump.sql
+--let $start_sql_file = $MYSQL_TEST_DIR/std_data/mysqldump_gtid_start.sql
+--let $finish_sql_file = $MYSQL_TEST_DIR/std_data/mysqldump_gtid_finish.sql
+--exec_in_background $MYSQL_DUMP $MYSQLDUMP_OPTS test t1 --start-sql-file=$start_sql_file --finish-sql-file=$finish_sql_file >$file
```
percona-ysorokin: Is it possible to somehow get process id here to be able to wait for its termination at the end?

mysql-test/t/mysqldump_gtid.inc
```diff
@@ -0,0 +1,69 @@
...
+
+# We expect this gtid not to be present
+--let $grep_pattern = $gtid_finish
+--let $include_filename = grep_pattern.inc
+--source include/grep_pattern.inc
+
+# cleanup
+--disconnect con1
+DROP TABLE t1;
+SET DEBUG_SYNC = 'RESET';
```
percona-ysorokin, line 78: Remove `$file` at the end

client/mysqldump.c
percona-ysorokin: Not sure if this method is 100% suitable here but please consider
https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_init_connect
I think you can create a non-privileged user for `mysqldump` for thich this init script will be executed whereas for `root` connections (default one in MTR) it will be not.

percona-ysorokin (changes requested):

client/mysqldump.c
kamil-holubicki: OK, I have removed these options from release build as initially suggested by @dutow .
The idea wiht sysvar_init_connect would work in this test case if not the one small problem:
1. For users that have the SUPER privilege, the content of init_connect is not executed.
2. Setting DEBUG_SYNC value requires the SUPER privilege.
So when connecting as non-SUPER, server starts execution of SET DEBUG_SYNC=... but fails because of privileges check.

client/mysqldump.c
kamil-holubicki: done

client/mysqldump.c
kamil-holubicki: done

client/mysqldump.c
kamil-holubicki: done

mysql-test/t/mysqldump_gtid.inc
kamil-holubicki: done

mysql-test/t/mysqldump_gtid.inc
kamil-holubicki: I have introduced 2 generic inc files for 
1. starting process in background + collect its pid and 
2. waiting for pid to finish

Please check comments in start_proc_in_background.inc why it cannot be achieved with --exec_in_background command.
I think in general escaping characters when passing them via Perl to mysqltest layer could be fixed/improved, but if we decide to invest time there my suggestion is to handle it as the separate ticket/PR as it will need bigger rework/testing (and impact other tests). In general escape characters should be removed one by one by every layer, so 
1. having \\$abc in *.test file should cause passing \$abc to mysqltest
2. having \$abc in *.test file should cause passing $abc to mysqltest
3. having $abc in *.test file should substitue abc variable by its value and pass the value to mysqltest

mysql-test/t/mysqldump_gtid.inc
kamil-holubicki: done

kamil-holubicki:
Jenkins job after rework:
https://ps57.cd.percona.com/job/percona-server-5.7-param/208/

client/mysqldump.c
percona-ysorokin: Yeah, you are right. I haven't thought about this.

client/client_priv.h
```diff
@@ -120,6 +120,8 @@ enum options_client
   OPT_ENABLE_COMPRESSED_COLUMNS_WITH_DICTIONARIES,
   OPT_DROP_COMPRESSION_DICTIONARY,
   OPT_ORDER_BY_PRIMARY_DESC,
+  OPT_START_SQL_FILE,
```
percona-ysorokin, line 123: Should this be wrapped into `#ifndef DBUG_OFF` as well?

mysql-test/t/mysqldump_gtid.inc
percona-ysorokin: Not addressed

mysql-test/include/wait_proc_to_finish.inc
```diff
@@ -0,0 +1,37 @@
...
+}
+
+--let _PID_FILE = $pid_file
+
+perl;
+  use Errno;
+  my $pid_file = $ENV{'_PID_FILE'};
+  open(my $f, '<', $pid_file) or die "OPENING $pid_file: $!\n";
+  my $pid = do { local($/); <$f> };
+  close($f);  use strict;
```
percona-ysorokin:
```suggestion
  close($f);
  use strict;
```

percona-ysorokin (approved):
LGTM if `count_sessions` issue addressed

client/client_priv.h
kamil-holubicki: This file is common for several tools from 'client' directory. I didn't investigate if they share some common parts which would cause enum values mismatch/shift when some of them built in debug and other in release. 
Oh the other size mysqldump does not use all these enums anyway (they are used by other client tools), enum acts as common place for collecting cmd line options identifiers.

It will be just unused enum value in release build. We can add #ifdef here but in my opinion there will be no advantage.

mysql-test/t/mysqldump_gtid.inc
kamil-holubicki: done

kamil-holubicki: done

PR #3589, opened 2019-12-31
PS-6148: If ANALYZE TABLE with transient statistics runs more than 600 seconds diagnostic query may crash server

kamil-holubicki:
https://jira.percona.com/browse/PS-6148

Locking of table statistics limited to the time when statistics related table members are changed. While calculation is done, statistics are not locked, so can be queried.

Index analysis protected by analyze_index_mutex (the same which protects index when analysis is done for persistent statistics)

Fix applied on top of 5.7 trunk (not GCA) because GCA does not contain changes needed for this fix.

### DIFF
--- a/mysql-test/suite/innodb/r/analyze_index.result
+++ b/mysql-test/suite/innodb/r/analyze_index.result
@@ -10,4 +10,15 @@ SET DEBUG_SYNC='now SIGNAL analyze.finish';
 Table	Op	Msg_type	Msg_text
 test.t1	analyze	status	OK
 SET DEBUG_SYNC= 'RESET';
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-param/203

satya-bodapati:
@kamil-holubicki  Patch looks good. Did you run the new test with --innodb-sync-debug=1 ?

kamil-holubicki:
Hi @satya-bodapati , yes I followed the same testing procedure as with PR 3562. Results are OK.

satya-bodapati (approved):

PR #3591, opened 2020-01-03
PS-6753: Travis CI: Limit number of tests for 8.0

inikep:
1. Fix issue described at https://bugs.mysql.com/bug.php?id=97246
2. Disable gcc-6, gcc-7, gcc-8, clang-5, clang-6, clang-7, clang-8 tests for pull requests
3. Compile with `-g1` to decrease output size

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,7 +87,7 @@ matrix:
 
     # Configurations for a pull request and after merging for percona/percona-server
     # 1
-    - if: branch =~ fullci OR repo = percona/percona-server AND type IN (push, pull_request, cron)
+    - if: branch =~ fullci OR repo = percona/percona-server AND type IN (push, cron)
       env:              BUILD=RelWithDebInfo
       compiler: clang
       os: osx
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -266,6 +268,12 @@ script:
     ccache --zero-stats;
 
   - echo --- Set cmake parameters;
+    COMPILE_OPT+=(
+      -DCMAKE_C_FLAGS_DEBUG=-g1
+      -DCMAKE_CXX_FLAGS_DEBUG=-g1
+      '-DCMAKE_C_FLAGS_RELWITHDEBINFO=-O2 -g1 -DNDEBUG'
+      '-DCMAKE_CXX_FLAGS_RELWITHDEBINFO=-O2 -g1 -DNDEBUG'
```
dutow, line 275: Why do you add NDEBUG and -O2 here? https://github.com/percona/percona-server/blob/8.0/CMakeLists.txt#L937 seems to prepend NDEBUG even if you don't speficy this variable, and the O2 isn't mentioned in the commit message at all. You changed it to reduce the running time of the optimizer? If yes, then it should probably be also mentioned in the commit message.

inikep: Look at https://github.com/percona/percona-server/blob/8.0/CMakeLists.txt#L1571
`-O2 -g -DNDEBUG` are defaults for `CMAKE_CXX_FLAGS_RELWITHDEBINFO`.
We only switch from `-g` (which translates to `-g2`) to `-g1` to decrease output size (since we have 500 MB limit for ccache for Travis).

dutow: You are right, I checked an 5.7 release build and assumed it's the same... (I wonder why did we switch to -O2 with 8.0?)

dutow: But I also verified the other part of my comment: `-DCMAKE_CXX_FLAGS_RELWITHDEBINFO=-g1 -O2'` is enough,  `-NDEBUG` will be added to the build commands even with this.

Strangely `-O2` is required if you don't specify it, it won't be present for all files (but will be there for some)... something is strange with this CMake script.

inikep: We have Release with `-O3` i.e. `-DCMAKE_CXX_FLAGS_RELEASE=-O3 -DNDEBUG`.

percona-ysorokin (approved):
LGTM

PR #3594, opened 2020-01-04
PS-6150: MySQL crash - sync_mutex_to_string

kamil-holubicki:
https://jira.percona.com/browse/PS-6150
https://jira.percona.com/browse/PS-6093

Two related issues fixed:
1. When printing out locked mutexes (SHOW ENGINE INNODB STATUS), LATCH_ID_AUTOINC object was cast to wrong type in sync0arr.cc sync_array_cell_print(), because of wrong request_type stored in cell. That caused calling GenericPolicy<Mutex>::to_string() instead of AggregateMutexStatsPolicy<Mutex>::to_string(). As the result of wrong cast, m_id member of policy used inside to_string() contained random value. This value was used to index the array in sync_latch_get_meta() which caused memory access violation.

2. When locking of the mutex reaches OS level, sync_array cell is created in ut0mutex.ic TTASEventMutex<Policy>::wait() and LATCH_ID_BUF_POOL_ZIP is mapped to SYNC_BUF_BLOCK cell type. This causes proper cast of mutex in sync0arr.cc sync_array_cell_print() to AggregateMutexStatsPolicy<Mutex>::to_string(). However to_string() method did not service LATCH_ID_BUF_POOL_ZIP causing assertion.
MTR test covering this case was not implemented, because the problem shows up when innodb internal threads lock zip_mutex (buf_flush_page_cleaner_coordinator thread and buf_lru_manager). Locking of zip_mutex on OS level can be reached by heavy insertion into the compressed table (and setting innodb_lru_scan_depth to high value) which will cause InnoDB to aggressively flush buffer pool pages. 
However, as locking threads are InnoDB internal threads, and DEBUG_SYNC facility is thread aware, there is no way to setup debug sync points from MTR test level.

### DIFF
--- a/mysql-test/suite/innodb/t/percona_extended_innodb_status.test
+++ b/mysql-test/suite/innodb/t/percona_extended_innodb_status.test
@@ -42,9 +42,6 @@ SET GLOBAL innodb_show_locks_held = 0;
 --let expected_locks= 4
 --source suite/innodb/include/percona_extended_innodb_status.inc
 
-# wait until all additional connections close
---source include/wait_until_count_sessions.inc
-
 # restoring affected global variables
 SET GLOBAL innodb_show_verbose_locks = @innodb_show_verbose_locks_save;
 SET GLOBAL innodb_show_locks_held = @innodb_show_locks_held_save;
@@ -66,3 +63,60 @@ DROP TABLE t;
 --echo Grepping InnoDB status for $SEARCH_PATTERN
 --source include/search_pattern_in_file.inc
 --remove_file $innodb_status
+
+
+--echo #
+--echo # PS-6150: MySQL crash - sync_mutex_to_string
+--echo #
+
+connect (con1,localhost,root,,);
+connect (con2,localhost,root,,);
+
+--connection default
+CREATE TABLE t1(id INT(15) NOT NULL AUTO_INCREMENT, a INT, PRIMARY KEY(id)) ENGINE=InnoDB;
+
+--connection con1
+SET DEBUG_SYNC='innobase_lock_autoinc SIGNAL autoinc_lock.locked WAIT_FOR autoinc_lock.continue'; 
+--send INSERT INTO test.t1(a) VALUES(1)
+
+--connection con2
+# Be sure that con1 locked autoinc_lock
+SET DEBUG_SYNC='now WAIT_FOR autoinc_lock.locked';
+
+# The following will cause autoinc_lock to reach OS level locking and registering the lock in sync_array used by 'SHOW ENGINE INNODB STATUS' to list locked mutexes
+SET SESSION debug="+d,catch_autoinc_mutex_os_lock";
+SET DEBUG_SYNC='autoinc_mutex_wait SIGNAL autoinc_mutex_wait.locked'; 
+--send INSERT INTO test.t1(a) VALUES(1)
+
+--connection default
+# wait up to the point when lock is registered in sync_array (in fact OS lock is not acquired yet, but this is enough)
+SET DEBUG_SYNC='now WAIT_FOR autoinc_mutex_wait.locked';
+
+# Send the query that caused the crash
+--disable_result_log
+SHOW ENGINE INNODB STATUS;
+--enable_result_log
+
+
+# cleanup
+# Unblock connections.
+--connection default
+SET DEBUG_SYNC='now SIGNAL autoinc_lock.continue';
+SET SESSION debug="-d,catch_autoinc_mutex_os_lock";
+SET DEBUG_SYNC = 'RESET';
+
+--connection con1
+--reap
+
+--connection con2
+--reap
+
+--disconnect con1
+--disconnect con2
+
+--connection default
+DROP TABLE t1;
+
+# wait until all additional connections close
+--source include/wait_until_count_sessions.inc
+
--- a/storage/innobase/include/ut0mutex.ic
+++ b/storage/innobase/include/ut0mutex.ic
@@ -54,13 +54,19 @@ TTASEventMutex<Policy>::wait(
 {
 	sync_cell_t*	cell;
 	sync_array_t*	sync_arr;
-
+	latch_id_t	latch_id = m_policy.get_id();
+	ulint		type = SYNC_MUTEX;
+	
+	if (latch_id == LATCH_ID_BUF_BLOCK_MUTEX
+	    || latch_id == LATCH_ID_BUF_POOL_ZIP) {
+	    type = SYNC_BUF_BLOCK;
+	} else if (latch_id == LATCH_ID_AUTOINC) {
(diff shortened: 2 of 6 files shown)


kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-param/205/

satya-bodapati:
Can you accommodate another change from @rahulmalik87  PR? (PS-6093)

satya-bodapati:
Looks good and Thanks for addressing the review comment (posted on slack)

kamil-holubicki:
Hi @satya-bodapati 
I'm just looking into PS-6093 and related PR from @rahulmalik87 .
As this is separate issue (the same are of the code, but different problem) I would prefer providing it as separate Jira ticket/PR. It will need its own fix and its own MTR test.

satya-bodapati:
> Hi @satya-bodapati
> I'm just looking into PS-6093 and related PR from @rahulmalik87 .
> As this is separate issue (the same are of the code, but different problem) I would prefer providing it as separate Jira ticket/PR. It will need its own fix and its own MTR test.

well, it is one line change and related to autoinc(PS-5639) and SHOW ENGINE INNODB STATUS. If we nitpick, we can say it is 'diffierent'. 

IMHO, it can be clubbed in this PR

satya-bodapati:
Also I don't like commit messages saying "Please look there.. here.. etc"
In future JIRA might be closed or migrated to some other system.

Please provide enough/sufficient detail in commit message (and preferably the description of this PR should have same message as commit message)

kamil-holubicki:
Hi @satya-bodapati,
PS-6093 and PS-6150 are regressions caused by PS-5639. 
Saying that PS-6093 is different issue I meant that it is related to LATCH_ID_BUF_POOL_ZIP. Fix is one liner (as in Rahul's PR), but additionally it needs MTR test which is totally different test than the one related to PS-6150. This is the only reason I see it as two separate chunks of work :) 

Anyway, if your suggestions is to include both fixes/tests in the same (this) PR I will go for it.

satya-bodapati:
@kamil-holubicki if you plan to add testcase, separate PR is fine.
If you want to go ahead with only the code change, then please combine the PRs.

satya-bodapati:
But both the PRs should be in same release. Otherwise it doesn't make sense

kamil-holubicki:
HI @satya-bodapati ,
After spending the whole day on figuring out how to design MTR test for the case of PS-6093 I decided to include PS-6093 fix here as well without MTR. Commit message contains all reasons.

storage/innobase/include/ut0mutex.ic
```diff
@@ -54,13 +54,19 @@ TTASEventMutex<Policy>::wait(
 {
 	sync_cell_t*	cell;
 	sync_array_t*	sync_arr;
-
+	latch_id_t		latch_id = m_policy.get_id();
+	ulint			type = SYNC_MUTEX;
+
```
satya-bodapati: can you align latch_id and type at same level as "cell and sync_arr variables"

storage/innobase/include/ut0mutex.ic
```diff
@@ -54,13 +54,19 @@ TTASEventMutex<Policy>::wait(
 {
 	sync_cell_t*	cell;
 	sync_array_t*	sync_arr;
-
+	latch_id_t		latch_id = m_policy.get_id();
+	ulint			type = SYNC_MUTEX;
+	
+	if (latch_id == LATCH_ID_BUF_BLOCK_MUTEX
+		|| latch_id == LATCH_ID_BUF_POOL_ZIP) {
```
satya-bodapati: || latch_id should be aligned to the latch_id above line

storage/innobase/include/ut0mutex.ic
```diff
@@ -71,6 +77,12 @@ TTASEventMutex<Policy>::wait(
 	and then the event is set to the signaled state. */
 
 	set_waiters();
+	
+	DBUG_EXECUTE_IF("catch_autoinc_mutex_os_lock", {
```
satya-bodapati: remove the trailing "{". It is not necessary

storage/innobase/include/ut0mutex.ic
```diff
@@ -71,6 +77,12 @@ TTASEventMutex<Policy>::wait(
...
+			if (m_policy.get_id() == LATCH_ID_AUTOINC) {
+				DEBUG_SYNC_C("autoinc_mutex_wait");
+			}
```
satya-bodapati: one less tab indentation for these lines

storage/innobase/include/ut0mutex.ic
```diff
@@ -71,6 +77,12 @@ TTASEventMutex<Policy>::wait(
...
+		});
```
satya-bodapati: remove the } and align ); to the D in DBUG_EXECUTE_IF

mysql-test/suite/innodb/t/percona_extended_innodb_status.test
```diff
@@ -66,3 +66,56 @@ DROP TABLE t;
...
 --source include/search_pattern_in_file.inc
 --remove_file $innodb_status
+
+
+--echo #
+--echo # PS-6150: MySQL crash - sync_mutex_to_string
+--echo #
+
+connect (con1,localhost,root,,);
+connect (con2,localhost,root,,);
```
satya-bodapati, line 73: count_session.inc and wait_until_count_session.inc

satya-bodapati (changes requested):
Looks good. Style issues..

storage/innobase/include/ut0mutex.ic
kamil-holubicki: done. This file uses tabs instead of spaces and that causes that formatting depends on tab width. I.e. in my editor where tab is set to 2 spaces now it looks like this:
image

storage/innobase/include/ut0mutex.ic
kamil-holubicki: done

storage/innobase/include/ut0mutex.ic
kamil-holubicki: done

mysql-test/suite/innodb/t/percona_extended_innodb_status.test
kamil-holubicki: done

storage/innobase/include/ut0mutex.ic
kamil-holubicki: done

storage/innobase/include/ut0mutex.ic
satya-bodapati: I think you should set it to 4. Check github page. It looks correct to me now

satya-bodapati (approved):
LGTM! Thank you!

PR #3596, opened 2020-01-07
PS-5729 Add server's UUID to percona system keys.

robgolebiowski:
Added server’s uuid to percona system key name. With this addition the
per server separation of keys inside Vault server (or any other remote
key store) is no longer needed.

System keys names used in the server correspond to the system keys names
in keyring, however key names in keyrings have UUID added.
For instance system key percona_innodb-0 (version 1) relates to system
key percona_innodb-0-UUID:1.
When you rotate a system key with rotate_system_key - a new version of
key containing UUID is created in keyring.
For instance SELECT rotate_system_key(“percona_innodb-0”) (rotation from
version 1) will create a key percona_innodb-0-UUID:2 (version 2) in
keyring.

The UUID addition is not backward compatible. The new version has
private_version field in fil_space_crypt_t structure set to 2 and the
earlier version has it set to 1.
During upgrade - if there is any table that is encrypted (even
partially) with KEYRING encryption version 1 – it will cause the upgrade
to fail.

The tables that were created with ENCRYPTION=’N’ in earlier versions of
the server will not be encrypted by encryption threads (provide that
‘the earlier’ version had KEYRING encryption feature in).

Normally during upgrade we go over all tablespaces and check if page0 of
any tablespace contains crypt_data (KEYRING encryption) and if the
tablespace is fully or partially encrypted. If it is - we fail the
upgrade. There are special cases for system tablespace and undo
tablespace. In case of those two tablespaces their page0 is always read
on server startup. Thus we can fail the server startup earlier in case
system tablespace or undo tablespace is KEYRING encrypted.

For tablespaces that have crypt_data in their page0, but are not
encrypted (at all, even partially) we can have the following cases :
1) tablespace was encrypted and then decrypted – and thus has crypt_data
in page0. We read crypt_data v1 of this tablespace and when we encrypt
it - we write crypt_data v2. Also we generate a key for it (with UUID in
its name) in keyring, if the associated key does not exist.
2) tablespace has encryption key id assigned – and thus has crypt_data
in page0. We follow the same scenario as in 1).
3) tablespace had ENCRYPTION=’N’ assigned and thus has crypt_data in
page0. We do not encrypt such tablespace.


robgolebiowski:
JR=> https://ps80.cd.percona.com/job/percona-server-8.0-param/482/

mysql-test/suite/encryption/t/encrypt_and_grep.test
```diff
@@ -172,9 +172,11 @@ SET GLOBAL default_table_encryption = ONLINE_FROM_KEYRING_TO_UNENCRYPTED;
...
 SET GLOBAL default_table_encryption = ONLINE_TO_KEYRING;
 
---echo # Wait max 10 min for key encryption threads to encrypt all spaces
+--let tables_count=`select count(*) from INFORMATION_SCHEMA.INNODB_TABLESPACES`
+
+--echo # Wait max 10 min for key encryption threads to encrypt all spaces, apart from temporary and t3
 --let $wait_timeout= 600
---let $wait_condition=SELECT COUNT(*) = 1 FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION = 0
+--let $wait_condition=SELECT COUNT(*) = $tables_count - 2 FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION = 1
 --source include/wait_condition.inc
```
satya-bodapati, line 180: How did this work before? after encryption by bg threads, MIN_KEY_VERSION was supposed to be 1 right?
With 0, how did this wait condition work?

robgolebiowski: I was waiting for 1 table to remain unencrypted (MIN_KEY_VERSION == 0). Now I am waiting for all tables, but two, to be encrypted (MIN_KEY_VERSION == 1).

mysql-test/suite/encryption/t/innodb-bad-key-change.test
```diff
@@ -4,6 +4,20 @@
 # present in error log
 --let $test_error_log=$MYSQL_TMP_DIR/error_log.err
 
+# Keys in std_data/keys2 and std_data/keys3 belong to the server with uuid b3665c10-e6c0-11e9-8e47-9cb6d0d5dc99. In order for MTR server
+# to be able to read those keys we need to assign this uuid to server_uuid via auto.conf file.
```
satya-bodapati, line 8: How did you find UUID? because this test exist before and there was no auot.cnf before?

mysql-test/suite/encryption/t/innodb-bad-key-change.test
```diff
@@ -28,7 +42,7 @@ SELECT * FROM t1;
 --echo #### Missing key errors ####
 --echo # Make sure that there is no error in error log related to missing key for t1
 
---let SEARCH_PATTERN=Tablespace id 5 in file t1\.ibd is encrypted but keyring or used key_id 2 is not available\. Can't continue reading table\. Please provide the correct keyring
+--let SEARCH_PATTERN=Tablespace id . in file t1\.ibd is encrypted but keyring or used key_id 2 is not available\. Can't continue reading table\. Please provide the correct keyring
```
satya-bodapati, line 45: Are these changes general improvement? Or tablespace ids changed?

mysql-test/suite/encryption/t/percona_system_key_rotation.test
```diff
@@ -38,15 +41,17 @@
 --source include/assert.inc
 
 --let $assert_text= It should be possible to rotate key percona_innodb
---let $assert_cond= "[SELECT rotate_system_key\\(\\"percona_innodb\\"\\) = 1]" = 1
+--let $assert_cond= "[SELECT rotate_system_key\\(\\"percona_innodb-0\\"\\) = 1]" = 1
```
satya-bodapati, line 44: Is the key name convention changed in this PR?

mysql-test/suite/encryption/t/percona_system_key_with_srv_guuid.test
```diff
@@ -0,0 +1,116 @@
+# PS-5729: Add server's UUID to percona system keys.
+
+CREATE TABLE t1 (a varchar(255)) ENCRYPTION='KEYRING'; # default encryption_key_id=0
+CREATE TABLE t2 (a varchar(255)) ENCRYPTION='KEYRING' ENCRYPTION_KEY_ID=1;
+CREATE TABLE t3 (a varchar(255)) ENCRYPTION_KEY_ID=2;
+CREATE TABLE t4 (a varchar(255)) ENCRYPTION_KEY_ID=3;
```
satya-bodapati, line 6: t4 looks duplicate scenario of t3

satya-bodapati, line 6: testing t4 has no extra value. Same as t3.

mysql-test/suite/encryption/t/percona_system_key_rotation.test
satya-bodapati, line 44: Did the naming scheme change in this PR?  The extra "-0"

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
```diff
@@ -64,48 +64,107 @@ let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/data_57_crypt_v1/data;
...
+--let $assert_text= test/t1 key version 1 assigned
+--let $assert_cond= "[SELECT MIN_KEY_VERSION FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'test/t1\\']" = 1
+--source include/assert.inc
+
+--let $assert_text= test/t2 should still have EXCLUDED=Y set.
+--let $assert_cond= "[SELECT name FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE EXCLUDED=\\'Y\\']" = "test/t2"
+--source include/assert.inc
+
+--let $assert_text= test/t3 should have key version 1 assigned
+--let $assert_cond= "[SELECT MIN_KEY_VERSION FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE name=\\'test/t1\\']" = 1
```
satya-bodapati: checking the wrong table here? shouldn't it be test/t3 here
and t3 is unencrypted right? (or is it encrypted by encryption threads in 5.7) ?

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
```diff
@@ -64,48 +64,107 @@ let $MYSQLD_DATADIR1 = $MYSQL_TMP_DIR/data_57_crypt_v1/data;
...
+--source include/search_pattern_in_file.inc
+--let SEARCH_PATTERN=foobar
+--let ABORT_ON=FOUND
+--let SEARCH_FILE=$t3_IBD
+--source include/search_pattern_in_file.inc
 
 # Check if server restarts correctly
 --replace_result $MYSQLD_DATADIR1 MYSQLD_DATADIR1 $MYSQL_TMP_DIR MYSQL_TMP_DIR $KEYRING_PLUGIN_OPT KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD KEYRING_PLUGIN_EARLY_LOAD
---source include/restart_mysqld.inc
+--source include/start_mysqld.inc
```
satya-bodapati, line 161: after restart. please add basic selects on the encrypted tables. Count(*) should be enough too.

storage/innobase/dict/dict0load.cc
```diff
@@ -1349,11 +1352,21 @@ space_id_t dict_check_sys_tablespaces(bool validate) {
...
     /* Ignore system, undo and file-per-table tablespaces,
-    and tablespaces that already are in the tablespace cache. */
+    and tablespaces that already are in the tablespace cache.
+    */
     if (fsp_is_system_or_temp_tablespace(space_id) ||
         fsp_is_undo_tablespace(space_id) ||
-        !fsp_is_shared_tablespace(fsp_flags) ||
-        fil_space_exists_in_mem(space_id, space_name, false, true, NULL, 0)) {
+        !fsp_is_shared_tablespace(fsp_flags)) {
+      continue;
```
satya-bodapati, line 1360: The check for system tablespace is skipped. Where do you check system tablespace is keyring encrypted?

storage/innobase/dict/dict0load.cc
```diff
@@ -1373,8 +1386,15 @@ space_id_t dict_check_sys_tablespaces(bool validate) {
     if (err != DB_SUCCESS) {
       ib::warn(ER_IB_MSG_191) << "Ignoring tablespace " << id_name_t(space_name)
                               << " because it could not be opened.";
+    } else {
+      fil_space_t *space = fil_space_get(space_id);
+      ut_ad(space != nullptr);
+      if (is_space_keyring_v1_encrypted(space)) {
```
satya-bodapati: its better if space_id is pased to is_space_keyring_v1_encrypted() and get the tablespace inside that.
Many callers (atleast in this patch) are using space_id.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -116,14 +117,14 @@ EncryptionKeyId get_global_default_encryption_key_id_value();
...
-    + ENCRYPTION_KEY_LEN       // tablespace key
-    + ENCRYPTION_KEY_LEN;      // tablespace iv
+    ENCRYPTION_MAGIC_SIZE + 1      // type
+    + 4                            // min_key_version
+    + 4                            // key_id
+    + 1                            // encryption
+    + CRYPT_SCHEME_1_IV_LEN        // iv (16 bytes)
+    + 1                            // encryption rotation type
+    + ENCRYPTION_KEY_LEN           // tablespace key
+    + ENCRYPTION_SERVER_UUID_LEN;  // server's UUID
```
satya-bodapati, line 127: Just for my knowledge, what is total size now?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -221,8 +222,8 @@ void fil_space_crypt_cleanup() {
   mutex_free(&crypt_stat_mutex);
 }
 
-fil_space_crypt_t::fil_space_crypt_t(uint new_type, uint new_min_key_version,
-                                     uint new_key_id,
+fil_space_crypt_t::fil_space_crypt_t(uint new_min_key_version, uint new_key_id,
+                                     const char *new_uuid,
                                      fil_encryption_t new_encryption,
```
satya-bodapati, line 227: Why do we need to store uuid on every crypt_data? This seems to waste 36 bytes per crypt_data.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -232,40 +233,58 @@ fil_space_crypt_t::fil_space_crypt_t(uint new_type, uint new_min_key_version,
...
+    key_found = false;
     min_key_version = ENCRYPTION_KEY_VERSION_NOT_ENCRYPTED;
-    key_found = true;
+  } else if (Encryption::is_online_encryption_on() == false &&
+             new_encryption == FIL_ENCRYPTION_DEFAULT) {
+    type = CRYPT_SCHEME_UNENCRYPTED;
+    min_key_version = ENCRYPTION_KEY_VERSION_NOT_ENCRYPTED;
+    key_found =
+        key_operation == FETCH_OR_GENERATE_KEY
+            ? Encryption::
```
satya-bodapati, line 269: This part is for unencrypted with encryption_key_id CREATE TABLE t1(a INT) ENCRYPTION_KEY_ID=4 ?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -232,40 +233,58 @@ fil_space_crypt_t::fil_space_crypt_t(uint new_type, uint new_min_key_version,
...
 
-  if (new_encryption == FIL_ENCRYPTION_OFF ||
-      (Encryption::is_online_encryption_on() == false &&
-       new_encryption == FIL_ENCRYPTION_DEFAULT)) {
+  if (strlen(new_uuid) == 0) {
+    ut_ad(strlen(server_uuid) == 0);
+    key_found = false;
+    min_key_version = ENCRYPTION_KEY_VERSION_INVALID;
+    // This was read - because when creating new crypt data - it means that
+    // uuid is never empty. type will be overwritten by read function
```
satya-bodapati, line 258: Didn't understand the comment. It conflicts with assert and condition above.
uuid is never empty but the above check is for empty uuid.

Is this for tablespaces that are encrypted during startup before uuid is ready?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -232,40 +233,58 @@ fil_space_crypt_t::fil_space_crypt_t(uint new_type, uint new_min_key_version,
...
-  mutex_create(LATCH_ID_FIL_CRYPT_DATA_MUTEX, &mutex);
-  // locker = crypt_data_scheme_locker; // TODO:Robert: Co to za locker, nie
-  // mogę znaleść jego definicji nawet w mariadb
-  type = new_type;
+  if (new_uuid != nullptr && strlen(new_uuid) > 0) {
+    memcpy(uuid, new_uuid, ENCRYPTION_SERVER_UUID_LEN);
+    uuid[ENCRYPTION_SERVER_UUID_LEN] = '\0';
+  } else {
+    uuid[0] = '\0';
+  }
```
satya-bodapati, line 251: Why is Encryption::s_uuid not used?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -232,40 +233,58 @@ fil_space_crypt_t::fil_space_crypt_t(uint new_type, uint new_min_key_version,
...
   } else {
     type = CRYPT_SCHEME_1;
     // key_found = true; // cheat key_get_latest_version that the key exists -
     // if it does not it will return ENCRYPTION_KEY_VERSION_INVALID
-    uchar *key = NULL;
-    uint key_version = 0;
+    uchar *key = nullptr;
+    uint key_version = ENCRYPTION_KEY_VERSION_NOT_ENCRYPTED;
     if (key_operation == FETCH_OR_GENERATE_KEY) {
       Encryption::get_latest_tablespace_key_or_create_new_one(
```
satya-bodapati: Since the get variant exists, I think it is sufficient to just use create() function and get rid of this function.
Name too is also very long.  Encryption::get_latest_key_or_create() is enough?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -793,6 +833,8 @@ byte *fil_parse_write_crypt_data_v2(space_id_t space_id, byte *ptr,
     fil_space_destroy_crypt_data(&crypt_data);
   }
 
+  ut_ad((ulint)(ptr - start_ptr) == len);
```
satya-bodapati, line 844: what does this assert mean? Can you please add comment?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -837,8 +879,12 @@ byte *fil_parse_write_crypt_data_v1(space_id_t space_id, byte *ptr,
...
   ptr += 1;
 
-  fil_space_crypt_t *crypt_data = fil_space_create_crypt_data(
-      encryption, key_id, Crypt_key_operation::FETCH_OR_GENERATE_KEY);
+  // since we are parsing v1 crypt_data - it means that encryption of this table
+  // has not yet started and thus we might need to create a key with uuid in
+  // keyring for this table now
+  fil_space_crypt_t *crypt_data =
+      fil_space_create_crypt_data(encryption, key_id, server_uuid,
+                                  Crypt_key_operation::FETCH_OR_GENERATE_KEY);
```
satya-bodapati, line 895: I dont understand this. When do we need to create key_with_uuid for  non-encrypted table (which is v1 & key without uuid)?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1305,16 +1351,48 @@ static bool fil_crypt_space_needs_rotation(rotate_thread_t *state,
...
   continue to rotate the tablespace */
+
   if (!crypt_data->is_key_found()) {
-    return false;
+    // We can end up here in case we try to encrypt tablespace but the key used
+    // by this tablespace is no longer in keyring. This can happen when keyring
+    // was changed or crypt_data is in version 1 and key's uuid is empty.
+    if (crypt_data->rotate_state.active_threads == 0 &&
+        crypt_data->encryption == FIL_ENCRYPTION_DEFAULT) {
+      ut_ad(
```
satya-bodapati, line 1373: 1) When keyring changed and key is missing, why are we handling this situation instead of throwing error?

2) If crypt_data is v1 (which is encryption='n' table), how will we end up here (rotation)?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17896,6 +17894,35 @@ static bool innobase_get_index_column_cardinality(
   return (failure);
 }
 
+static bool innobase_is_tablespace_keyring_v1_encrypted(
+    const dd::Tablespace &tablespace, int &error) {
+  error = 0;
+  space_id_t id = 0;
```
satya-bodapati, line 17929: is this used only during upgrade?
If so, can you add assert ut_ad(srv_is_upgrade_mode);

satya-bodapati (changes requested):

mysql-test/suite/encryption/t/innodb-bad-key-change.test
robgolebiowski: I have modified by hand all the std_data/keys[2-3] and added the uuid b3665c10-e6c0-11e9-8e47-9cb6d0d5dc99 to all the keys.

mysql-test/suite/encryption/t/innodb-bad-key-change.test
robgolebiowski: tablespace ids were changed, but after merging with latest trunk, not because of this PR.

mysql-test/suite/encryption/t/percona_system_key_rotation.test
robgolebiowski: No. Previously was possible to issue percona_innodb without the number specified - which was incorrect. That was fixed as part of this PR. Now it is only possible to rotate keys that already exist in the keyring.

robgolebiowski: I will add this to commit message.

mysql-test/suite/encryption/t/percona_system_key_with_srv_guuid.test
robgolebiowski: It seems it is. I probably just added it to be extra sure, that rotation with two different ENCRYPTION_KEY_IDs work. If it is confusing I can remove it - or maybe just added a comment ?

robgolebiowski: The same as previously - checking for two different keys, just to be sure.

mysql-test/suite/encryption/t/percona_system_key_rotation.test
robgolebiowski: No, the same answer as to the same question above. I have added description to the commit message.

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
robgolebiowski: Yes, it should be test/t3. Fixed. It is encrypted. It was created as unencrypted in 5.7, but we have encryption threads turned on and ONLINE_TO_KEYRING set. So t3 will be encrypted by encryption threads.

mysql-test/suite/encryption/t/upgrade_crypt_data_57_v1.test
robgolebiowski: Added.

storage/innobase/dict/dict0load.cc
robgolebiowski: in srv0start.cc

robgolebiowski: +  // Fail startup if sys space is encrypted with crypt_data v1
+  // This should only happen on upgrade
+  if (srv_sys_space.keyring_encryption_info.page0_has_crypt_data &&
+      srv_sys_space.keyring_encryption_info.type != CRYPT_SCHEME_UNENCRYPTED &&
+      srv_sys_space.keyring_encryption_info.private_version == 1) {
+    ib::error(ER_UPGRADE_KEYRING_V1_ENCRYPTION);
+    return (DB_ERROR);
+  }

robgolebiowski: `+  // Fail startup if sys space is encrypted with crypt_data v1
+  // This should only happen on upgrade
+  if (srv_sys_space.keyring_encryption_info.page0_has_crypt_data &&
+      srv_sys_space.keyring_encryption_info.type != CRYPT_SCHEME_UNENCRYPTED &&
+      srv_sys_space.keyring_encryption_info.private_version == 1) {
+    ib::error(ER_UPGRADE_KEYRING_V1_ENCRYPTION);
+    return (DB_ERROR);
+  }
`

robgolebiowski: // Fail startup if sys space is encrypted with crypt_data v1
// This should only happen on upgrade
if (srv_sys_space.keyring_encryption_info.page0_has_crypt_data &&
     srv_sys_space.keyring_encryption_info.type != CRYPT_SCHEME_UNENCRYPTED &&
     srv_sys_space.keyring_encryption_info.private_version == 1) {
  ib::error(ER_UPGRADE_KEYRING_V1_ENCRYPTION);
  return (DB_ERROR);
}

storage/innobase/dict/dict0load.cc
robgolebiowski: I have added overload version of this function which takes space_id. Whenever I had to explicitly retrieve this space I use this function. However, when space was already retrieved I use the old function taking fil_space_t*.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: 98 bytes

storage/innobase/fil/fil0crypt.cc
robgolebiowski: To be able to identify the encryption key id for every table without any additional information. While copying tables this can be very important information (backups). This is the same as with MK encryption. You can identified MK by looking at encryption header. We better follow the same approach here.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: This is for all tables that have crypt_data written in page0 and online encryption is disabled. This can be due to ENCRYPTION_KEY_ID being specified, but also for tables that were once encrypted by encryption threads and then were decrypted.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: yes, sorry - this commit is a mess. I have moved this comment up and changed it.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: Encryption::s_uuid is not always equal to server_uuid. It is a bit of weird thing. When a table encrypted with Master Key is read and this table was copied over from some different location - for instance backup, and its Master Key Id (which does not include uuid) is higher than currently discovered Id - the Encryption::s_uuid will be replaced with UUID read from this table header. This can be seen in Encryption::decode_encryption_information. It is better to keep keyring encryption away from it.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: I am using this function in more places and I would have to repeat the code with just create() function in those places. I prefer this. I can change the name of the function though :).

storage/innobase/fil/fil0crypt.cc
robgolebiowski: done

storage/innobase/fil/fil0crypt.cc
robgolebiowski: Well, true. My thinking was that we always try to be in a state that all necessary keys are already in keyring when we decide to encrypt a tablespace. I guess it seems safer to me than creating key just before encryption starts. Anyways, this is more of a "safer" path and I can remove it if you do not like this approach ?

storage/innobase/fil/fil0crypt.cc
robgolebiowski: 1. Because tablespace could be first encrypted then decrypted - i.e. we have unencrypted tablespace with crypt_data in page0 and key_id assigned in page0. Assume the scenario that all tablespaces in the server were encrypted, then decrypted and user removed the keyring. After some time the user decides to encrypt the tablespace again - so we need to create the necessary keys again. Throwing an error here would not be good option - why user would need to retain old keyring if he decrypted all the tablespaces?
2. crypt_data can be v1 without encryption='n'. Remember that encryption threads works in earlier versions of 8.0.

storage/innobase/handler/ha_innodb.cc
robgolebiowski: srv_is_upgrade_mode is only true for 5.7 upgrade. Check innobase_init_files : srv_is_upgrade_mode = (dict_init_mode == DICT_INIT_UPGRADE_57_FILES);

mysql-test/suite/encryption/t/innodb-bad-key-change.test
satya-bodapati: but they are in non-readble/binary format right?

mysql-test/suite/encryption/t/innodb-bad-key-change.test
satya-bodapati: How tablespace id changed? This can happen only when datadir is upgraded

mysql-test/suite/encryption/t/percona_system_key_rotation.test
satya-bodapati: Previously, percona_innodb and percona_innodb-0 were same?

mysql-test/suite/encryption/t/percona_system_key_with_srv_guuid.test
satya-bodapati: ok to keep it as it is

storage/innobase/fil/fil0crypt.cc
satya-bodapati: thanks.. still below the MK header limit.. :)

storage/innobase/fil/fil0crypt.cc
satya-bodapati: well, how does it work with MK, fil_space_t doesn't have uuid member?

storage/innobase/fil/fil0crypt.cc
satya-bodapati: Why do we care about crypt_data of the tables that are once encrypted and decrypted later. Can't we treat them as garbage and write the new crypt_data with V3 format.

Also am thinking, since ENCRYPTION_KEY_ID is also KEYRING feature, we can disallow upgrade from previous versions. What do you say? Does it simplify?

storage/innobase/fil/fil0crypt.cc
satya-bodapati: Ok. I am thinking if we can avoid uuid in every fil_space_t/fil_crypt_t and simplify things.. (but also wondering how it works with MK.. if it only has one "global" uuid).

storage/innobase/fil/fil0crypt.cc
satya-bodapati: ok, please change the function name then.. thank you

storage/innobase/fil/fil0crypt.cc
satya-bodapati: will this be relevant if we stop supporting ENCRYPTION_KEY_ID tables from previous versions?

Also, if this 'safety' path indicates something wrong happenned (like keyring changed or key missing?) it is not got good to silently recreate keys.

Or may be warn user and then recreate?

storage/innobase/fil/fil0crypt.cc
satya-bodapati: Agree with the intention but not with the approach or design. Once a tablespace is decrypted, why do we care about crypt_data? Why do we even load crypt_data? 

Can't we just reset crypt_data when it was decrypted? Looks like we retaining some 'state' even after it is decrypted.

On 2)

Didn't we agree that we will not handle v1 format at all (except that somehow handle encryption='n' tables somehow)?

storage/innobase/handler/ha_innodb.cc
satya-bodapati: ok

mysql-test/suite/encryption/t/innodb-bad-key-change.test
robgolebiowski: "...there's way too much information to decode the Matrix. You get used to it, though. Your brain does the translating. I don't even see the code. All I see is blonde, brunette, redhead. Hey uh, you want a drink?"  -- Cypher

mysql-test/suite/encryption/t/innodb-bad-key-change.test
robgolebiowski: When I wrote this test some time ago t1.ibd had Tablespace id 5. After merging newer version of trunk the id has changed. t1.ibd is newly created here, it is not upgraded. So the server assigns id here.

mysql-test/suite/encryption/t/percona_system_key_rotation.test
robgolebiowski: No, they were different.

storage/innobase/fil/fil0crypt.cc

storage/innobase/fil/fil0crypt.cc
robgolebiowski: Maybe I understood your first question incorrectly. I will try to answer the first question once again. We have a table that has crypt_data v1 assigned. This table is unencrypted - otherwise the upgrade would have failed. We have turned on encryption threads. The encryption key that is needed to encrypt this table will not be in keyring. We need to create this key now.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: Overall the design came from MariaDB. In details:
1) crypt_data can also contains encryption key id - we retain this id.
2) This is how we handle encryption='n' - by having crypt_data with information that this table should be excluded by encryption threads.

storage/innobase/fil/fil0crypt.cc
satya-bodapati: I can't see how.. Can you list the steps.

1. Execute these SQLs (8.0.18 or below)
2. upgrade to mysqld (which has this patch)

Can you list the SQLs?

satya-bodapati: As per slack discussion, understood that  Keys without UUID are not created/read/written.

Request to add assert that all keys are V2 format and with valid UUID.

storage/innobase/fil/fil0crypt.cc
satya-bodapati: ok, this is similar to my above concern to get rid of handling of v1 keys (the goal).
As per IM discussion, we only need to handle v2 keys. Then it is fine.

mysql-test/suite/encryption/t/innodb-bad-key-change.test
george-lorch: ^^ Definitely in the top 10 of PC comments...

storage/innobase/fil/fil0crypt.cc
robgolebiowski: I have added assert to fill_key_name functions. All KEYRING InnoDB keys need to go through this function. I check there that uuid is never empty. This means that all keys created/read/written in the server need to have uuid set.

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/506/

storage/innobase/os/os0file.cc
```diff
@@ -8447,32 +8454,50 @@ void Encryption::random_value(byte *value) {
...
   int ret;
 
-  fill_key_name(key_name, key_id);
+  // Newly created tablespace keys should always have uuid equal to server_uuid.
+  // There are situations when server_uuid is not available - like when parsing
+  // redo logs. Then we read uuid from crypto's redo log.
+  ut_ad(strlen(server_uuid) == 0 ||
+        memcmp(server_uuid, uuid, ENCRYPTION_SERVER_UUID_LEN) == 0);
+
+  fill_key_name(key_name, key_id, uuid);
```
satya-bodapati, line 8500: can we also assert that uuid (not the server_uuid) is not empty ?

satya-bodapati (changes requested):
Still trying to understand how keys are created and handled when uuid is not available? (For example during InnoDB startup, for redo, temporary, system, undo etc)

storage/innobase/os/os0file.cc
satya-bodapati: Ignore, inside fill_key_name(), you already have.. :)

robgolebiowski:
system tablespace, temporary, undo, redo etc. are not encrypted during startup. ONLINE_TO_KEYRING will work for those tablespaces only after the startup is completed and encryption threads are on. This in line with what we have for Master Key encryption. I.e. encryption for those tablespaces have to be enabled outside default_table_encryption. Here is even better. After startup is completed user turns on encryption threads and waits for those tablespaces to get encrypted.

storage/innobase/fil/fil0crypt.cc
satya-bodapati: as per slack discussion, the answer is  uuid and master_key_id from page0 are used to decrypt the tablespace key and the unenecrpyted tablespace key is stored in fil_space_t. So there is no nned for UUID member for MK encryption.

In case of KEYRING, theere is no tablespace key and so UUID is necessary to be form the key and decrypt pages. So it has to be stored in every fil_space_t (currently fil_sapce_t->crypt_data)

storage/innobase/fil/fil0crypt.cc
```diff
@@ -541,11 +569,16 @@ static fil_space_crypt_t *fil_space_read_crypt_data_v2(
 
   ut_ad(key_id != (uint)(~0));
 
+  char uuid[ENCRYPTION_SERVER_UUID_LEN];
+  memset(uuid, 0, ENCRYPTION_SERVER_UUID_LEN);
+  memcpy(uuid, page + offset + bytes_read, ENCRYPTION_SERVER_UUID_LEN);
+  bytes_read += ENCRYPTION_SERVER_UUID_LEN;
+
```
satya-bodapati, line 576: Here can we assert that if crypt_schem is encrypted, uuid shouldn't be empty. (This is the reader part)
I verfied that we assert the same during write to page 0 (writer part)

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1305,16 +1357,48 @@ static bool fil_crypt_space_needs_rotation(rotate_thread_t *state,
...
+        return false;  // failed to fetch or create the key - skip the
+                       // tablespace
+      }
+
+      key_state->key_version = 1;
+      // We assing here uuid to crypt_data key's uuid. If crypt_data
+      // does not make it to page0 (due to crash) we will do the same
+      // after the restart, because we will end up here again -
+      // encryption key will not be found.
+      memcpy(crypt_data->uuid, server_uuid, ENCRYPTION_SERVER_UUID_LEN);
```
satya-bodapati, line 1393: can this rotation thread become active at startup (when uuid is not ready).? Is it possible to assert that server_uuid is not empty here?
If uuid can be empty, how to do we fix it?

storage/innobase/os/os0file.cc
```diff
@@ -8447,32 +8454,50 @@ void Encryption::random_value(byte *value) {
   my_rand_buffer(value, ENCRYPTION_KEY_LEN);
 }
 
-void Encryption::fill_key_name(char *key_name, uint key_id) {
+void Encryption::fill_key_name(char *key_name, uint key_id, const char *uuid) {
```
satya-bodapati, line 8457: General comment: it looks like we have bunch of function that can name "keyring' in the function name.  This one and get_tablespace_key() and other friends.

It is optional and upto you :)

satya-bodapati (changes requested):

storage/innobase/fil/fil0crypt.cc
robgolebiowski: They cannot be active at the startup. The encryption by encryption threads is on when srv_default_table_encryption == DEFAULT_TABLE_ENC_ONLINE_TO_KEYRING. srv_default_table_encryption is InnoDB's copy of default_table_encryption. It gets set in mysqld.cc by calling innodb_hton->fix_default_table_encryption(global_system_variables.default_table_encryption);
This is done immediately after innodb_hton->fix_tablespaces_empty_uuid.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: I have added the assert

storage/innobase/fil/fil0crypt.cc
robgolebiowski: I have added a comment for this.

mysql-test/suite/encryption/t/percona_redo_default_key.test
```diff
@@ -0,0 +1,158 @@
...
+--source include/start_mysqld.inc
+
+--source include/shutdown_mysqld.inc
+
+--remove_file $MYSQLD_DATADIR1/ib_logfile0
+--remove_file $MYSQLD_DATADIR1/ib_logfile1
+
+# We have removed redo log files. The server is no longer able to read percona_redo key's UUID from the redo log.
+# It should reuse the percona_redo:0 i.e. the default percona redo key and next generate and switch to using
+# new key - percona_redo-UUID:1 - when UUID becomes available.
```
dutow: Is this a new key at this point? (uuid is unchanged, encryption was already turned on during bootstrap)

mysql-test/suite/encryption/t/upgrade_redo_keyring_enc_v1.test
```diff
@@ -0,0 +1,99 @@
...
+call mtr.add_suppression("Info table is not ready to be used. Table 'mysql.slave_master_info' cannot be opened");
+call mtr.add_suppression("Error in checking mysql.slave_master_info repository info type of TABLE");
+call mtr.add_suppression("Error creating master info: Error checking repositories.");
+call mtr.add_suppression("Slave: Failed to initialize the master info structure for channel");
+call mtr.add_suppression("Failed to create or recover replication info repositories.");
+call mtr.add_suppression("db.opt file not found for test database. Using default Character set");
+call mtr.add_suppression("Skip re-populating collations and character sets tables in InnoDB read-only mode");
+call mtr.add_suppression("Skipped updating resource group metadata in InnoDB read only mode");
+
+# PS-5729 (Add server's UUID to percona system keys) introduces new version of redo log encryption header.
```
dutow, line 50: Since keyring redo encryption isn't GA, do we want to support this upgrade path? Previously for similar changes, we didn't do it.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3450,26 +3473,49 @@ redo_log_key *redo_log_keys::generate_and_store_new_key(THD *thd) {
...
+    return nullptr;
   }
 
-  return true;
+  if (default_rkey != nullptr) {
+    redo_log_key *rk = &m_keys[0];
+    rk->version = 0;
+    memcpy(rk->key, default_rkey, ENCRYPTION_KEY_LEN);
+    rk->present = true;
+    return rk;
```
dutow, line 3505: leaking memory (default_redo, default_rkey)

mysql-test/suite/encryption/t/percona_redo_default_key.test
robgolebiowski: You are right, this key exists because of the bootstrap. I have removed this part of the test, because it is redundant in this case.

mysql-test/suite/encryption/t/upgrade_redo_keyring_enc_v1.test
robgolebiowski: We have done it for innodb tablespaces, so makes sense to do it here too. It is also quite easy failure path.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: SHAME ! SHAME !

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3450,26 +3473,49 @@ redo_log_key *redo_log_keys::generate_and_store_new_key(THD *thd) {
...
-  /* This is a for loop, but it really only should store a key with current
-  version 0 */
-  for (const auto &item : m_keys) {
-    if (!item.second.persisted()) {
-      ut_ad(item.first == 0);
-      if (my_key_store(PERCONA_REDO_KEY_NAME, "AES", nullptr, item.second.key,
-                       ENCRYPTION_KEY_LEN)) {
-        return false;
-      }
+  std::string default_key_name{get_key_name("", 0)};
```
satya-bodapati: IMHO, there is no need to use get_key_name here. Just use default_key{PERCONA_REDO_KEY_NAME} When we do this, we can assert in get_key_name() variants that uuid must not be nullptr and strlen(uuid) > 0.

storage/innobase/fil/fil0crypt.cc
satya-bodapati: Optional note: according to effective c++, the API can enforce std::unique_ptr when it returns allocated memory to caller. But doing that means, we should change API signature here.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3450,26 +3473,49 @@ redo_log_key *redo_log_keys::generate_and_store_new_key(THD *thd) {
...
+    memcpy(rk->key, default_rkey, ENCRYPTION_KEY_LEN);
+    rk->present = true;
+    return rk;
+  }
+
+  // we use store instead of generate because we want to store system key
+  // with illegal version - percona_redo:0.
+  Encryption::random_value(reinterpret_cast<byte *>(&m_keys[0].key));
+
+  if (my_key_store(default_key_name.c_str(), "AES", nullptr, m_keys[0].key,
```
satya-bodapati, line 3512: Double check:  ;-)

what would be the name here just "percona_redo"? and  the key_store will save it as percona_redo:0 ?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3450,26 +3473,49 @@ redo_log_key *redo_log_keys::generate_and_store_new_key(THD *thd) {
...
+  size_t default_klen = 0;
+
+  if (my_key_fetch(default_key_name.c_str(), &default_redo_key_type, nullptr,
+                   reinterpret_cast<void **>(&default_rkey), &default_klen)) {
+    ib::error(ER_REDO_ENCRYPTION_CANT_FETCH_DEFAULT_KEY);
+    if (thd != nullptr) {
+      ib_senderrf(thd, IB_LOG_LEVEL_WARN, ER_REDO_ENCRYPTION_CANT_FETCH_DEFAULT_KEY);
     }
+    my_free(default_redo_key_type);
+    my_free(default_rkey);
```
satya-bodapati, line 3494: Can we assume memory is allocated on failure?
And can my_free() handle free(nullptr)?

storage/innobase/include/fil0crypt.h
```diff
@@ -355,22 +353,39 @@ class redo_log_keys final {
...
+  @return KEYRING encryption redo key name */
+  std::string get_key_name(const char *uuid, uint key_version);
+  /**
+  Get KEYRING encryption redo key name
+  @param[in] uuid - key's UUID
+  @return KEYRING encryption redo key name */
+  std::string get_key_name(const char *uuid);
+  /**
+  Get KEYRING encryption redo key name
+  @param[in]  oss - output string stream
```
satya-bodapati: in,out

storage/innobase/include/fil0crypt.h
```diff
@@ -355,22 +353,39 @@ class redo_log_keys final {
...
+  @param[out] uuid - key's UUID */
```
satya-bodapati: this should be of type 'in' ?

storage/innobase/include/fil0fil.h
```diff
@@ -318,6 +318,9 @@ struct fil_space_t {
   /** Only used for redo log encryption: the currently active key handle */
   redo_log_key *encryption_redo_key;
 
+  /** Only used for redo log encryption: the redo encryption's key uuid */
+  char encryption_redo_key_uuid[ENCRYPTION_SERVER_UUID_LEN+1];
+
```
satya-bodapati: small concern. This will be unused for almost all other fil_space_ts. So for 1 million fil_space_ts, the wastage is around 34~35MB.

Is it possible to make it nullptr and use space->heap when allocation is required?

storage/innobase/include/log0log.h
```diff
@@ -812,9 +812,12 @@ It just need to flush the file header block with current master key.
 @param[in]	iv	encryption iv
 @param[in]	is_boot	if it is for bootstrap
 @param[in]	redo_log_encrypt	encryption type
+@param[in]	version	                key's version (used for KEYRING
+                                        encryption)
 @return true if success. */
 bool log_write_encryption(byte *key, byte *iv, bool is_boot,
-                          redo_log_encrypt_enum redo_log_encrypt);
+                          redo_log_encrypt_enum redo_log_encrypt,
+                          uint version = 0);
```
satya-bodapati, line 820: Please add assert in the implementation that  if (redo_log_encrypt_enum == MK) (ut_ad(version == 0))

storage/innobase/include/log0log.h
```diff
@@ -812,9 +812,12 @@ It just need to flush the file header block with current master key.
 @param[in]	iv	encryption iv
 @param[in]	is_boot	if it is for bootstrap
 @param[in]	redo_log_encrypt	encryption type
+@param[in]	version	                key's version (used for KEYRING
+                                        encryption)
```
satya-bodapati: align the 'encryption' to the above line word "key's"

storage/innobase/include/os0file.h
```diff
@@ -274,7 +274,9 @@ static const char ENCRYPTION_KEY_MAGIC_V2[] = "lCB";
 information version. */
 static const char ENCRYPTION_KEY_MAGIC_V3[] = "lCC";
 
-static const char ENCRYPTION_KEY_MAGIC_RK[] = "lRK";
+static const char ENCRYPTION_KEY_MAGIC_RK_V1[] = "lRK";
+
+static const char ENCRYPTION_KEY_MAGIC_RK_V2[] = "RKB";
```
satya-bodapati: what are these? Can you comment what they stand for?

storage/innobase/log/log0write.cc
```diff
@@ -2713,7 +2715,16 @@ bool log_read_encryption() {
   redo_log_key *mkey = nullptr;
   Encryption::Type encryption_type = Encryption::NONE;
   uint version = 0;
-  if (memcmp(log_block_buf + LOG_HEADER_CREATOR_END, ENCRYPTION_KEY_MAGIC_RK,
+
+  if (memcmp(log_block_buf + LOG_HEADER_CREATOR_END, ENCRYPTION_KEY_MAGIC_RK_V1,
+               ENCRYPTION_MAGIC_SIZE) == 0) {
+    // can only happen during the upgrade
+    ib::error() << "Upgrade cannot continue as redo is encrypted with old, experimental version of KEYRING encryption. "
```
satya-bodapati: this is old usage. ideal usage is ib::error(ER_BLAH);

satya-bodapati (changes requested):
will check testcases now (for redo uuid)

mysql-test/suite/encryption/t/upgrade_redo_keyring_enc_v1.test
satya-bodapati: We are not supporting upgrade of keyring encrypted redo. Upgrade aborts. The only reason it is allowed because redo files are removed(assumed slow shutdown).

mysql-test/suite/encryption/t/percona_redo_default_key.test
```diff
@@ -0,0 +1,158 @@
...
+--remove_file $BOOTSTRAP_LOG
+--remove_file $MYSQLD_DATADIR1/ib_logfile0
+--remove_file $MYSQLD_DATADIR1/ib_logfile1
+
+--let $test_error_log=$MYSQL_TMP_DIR/error_log.err
+
+# We crash before percona_redo:0 gets rotated to percona_redo-UUID:1 (it gets rotated when UUID becomes available)
+
+--error 137
+--exec $MYSQLD_CMD --loose-skip-log-bin --skip-log-slave-updates --datadir=$MYSQLD_DATADIR1 $KEYRING_DATA $KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD --log-error=$test_error_log --innodb_redo_log_encrypt=KEYRING_KEY --debug=+d,crash_before_redo_key_is_rotated
```
satya-bodapati: just use d,crash_before..
It might create problems on other operating systems

mysql-test/suite/encryption/t/percona_redo_default_key.test
```diff
@@ -0,0 +1,158 @@
...
+--let $assert_text= rotate percona_redo-UUID key to version 2
+--let $assert_cond= "[SELECT rotate_system_key\\(\\"percona_redo\\"\\) = 1]" = 1
+--source include/assert.inc
+
+--source include/shutdown_mysqld.inc
+
+--remove_file $MYSQLD_DATADIR1/ib_logfile0
+--remove_file $MYSQLD_DATADIR1/ib_logfile1
+
+--let $restart_parameters = "restart: --loose-skip-log-bin --skip-log-slave-updates --datadir=$MYSQLD_DATADIR1 $KEYRING_DATA $KEYRING_PLUGIN_OPT $KEYRING_PLUGIN_EARLY_LOAD --log-error=$test_error_log --innodb_redo_log_encrypt=KEYRING_KEY --debug=+d,assert_default_to_ver2_rotation"
```
satya-bodapati, line 103: isn't the startup supposed to crash/abort?

Try with --debug=d,assert..
if it is supposed to crash, you my have to use --exec

mysql-test/suite/innodb/r/log_encrypt_2_mk.result
```diff
@@ -63,9 +63,6 @@ c1	c2
 7	hhhhh
 8	iiiii
 9	jjjjj
-SELECT rotate_system_key("percona_redo");
-rotate_system_key("percona_redo")
-1
 ALTER INSTANCE ROTATE INNODB MASTER KEY;
```
satya-bodapati, line 66: why is this missing?

mysql-test/suite/innodb/r/log_encrypt_2_rk.result
```diff
@@ -66,6 +66,9 @@ c1	c2
 SELECT rotate_system_key("percona_redo");
 rotate_system_key("percona_redo")
 1
+SELECT rotate_system_key("percona_redo");
+rotate_system_key("percona_redo")
+1
 ALTER INSTANCE ROTATE INNODB MASTER KEY;
```
satya-bodapati, line 72: is the duplication intentional?

share/errmsg-utf8.txt
```diff
@@ -19930,6 +19930,12 @@ ER_REDO_ENCRYPTION_KEYRING
 ER_UNDO_NO_KEYRING
   eng "Undo log can't be encrypted if the keyring plugin is not loaded."
 
+ER_REDO_ENCRYPTION_CANT_FETCH_DEFAULT_KEY
+  eng "Couldn't fetch default percona_redo key. Please check if keyring is operational."
+
+ER_REDO_ENCRYPTION_CANT_FETCH_NOR_GENERATE_DEFAULT_KEY
+  eng "Couldn't fetch nor generate default percona_redo key. Please check if keyring is operational."
+
```
satya-bodapati: may be simple ER_REDO_KEYRING_NO_KEY or CANT_FETCH? 

I dont see how the second message is useful to user. they are similar w.r.t to action taken by user.
Just one is sufficient IMO

storage/innobase/fil/fil0crypt.cc
robgolebiowski: not really, you need to also append a version to it - i.e. 0

storage/innobase/fil/fil0crypt.cc
robgolebiowski: the name would be the same as the key stored in keyring. I pass here "percona_redo:0". The discussion from slack was about my_key_generate - there when you provide a system key (for instance my_key_generate("percona_redo"); - it will either create a new key with version 1 - percona_redo-server_uuid:1 or rotate existing key to a new version.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: all keyrings I have written do not allocate memory on error. However, it might change if others develop new keyrings and will allocate memory even in case of error. Yes, my_free can handler nullptr :
void my_free(void *ptr) {
  my_memory_header *mh;

  if (ptr == NULL) return;

and NULL == nullptr evaluates to true.

storage/innobase/include/fil0crypt.h
robgolebiowski: done

storage/innobase/include/fil0crypt.h
robgolebiowski: done

storage/innobase/include/fil0fil.h
robgolebiowski: done

storage/innobase/include/log0log.h
robgolebiowski: done

storage/innobase/include/log0log.h
robgolebiowski: done

storage/innobase/include/os0file.h
robgolebiowski: done

storage/innobase/log/log0write.cc
robgolebiowski: done

mysql-test/suite/encryption/t/percona_redo_default_key.test
robgolebiowski: done

mysql-test/suite/encryption/t/percona_redo_default_key.test
robgolebiowski: actually it shouldn't. This just activates the assert that default key was rotated to ver.2 of percona_redo-UUID key.

mysql-test/suite/innodb/r/log_encrypt_2_mk.result
robgolebiowski: Previously it was possible to rotate percona_redo key, even if it didn't exist. Now it is impossible - so performing percona_redo key rotation in case of Master Key encryption does not make sense.

mysql-test/suite/innodb/r/log_encrypt_2_rk.result
robgolebiowski: yes, I am checking if key was rotated correctly from version 2 to 3.

share/errmsg-utf8.txt

storage/innobase/log/log0write.cc
```diff
@@ -2800,7 +2797,15 @@ bool log_read_encryption() {
     space->encryption_redo_key = mkey;
     dberr_t err = fil_set_encryption(space->id, encryption_type, key, iv);
     space->encryption_key_version = version;
-    memcpy(space->encryption_redo_key_uuid, uuid,
+    space->encryption_redo_key_uuid.reset(
+        new (std::nothrow) char[ENCRYPTION_SERVER_UUID_LEN + 1]);
+    if (space->encryption_redo_key_uuid.get() == nullptr) {
```
satya-bodapati, line 2802: clever :+1:

storage/innobase/include/os0file.h
```diff
@@ -274,8 +274,12 @@ static const char ENCRYPTION_KEY_MAGIC_V2[] = "lCB";
 information version. */
 static const char ENCRYPTION_KEY_MAGIC_V3[] = "lCC";
 
+/** Encryption magic bytes before 8.0.19, it's for checking KEYRING
+redo log information version. */
 static const char ENCRYPTION_KEY_MAGIC_RK_V1[] = "lRK";
 
+/** Encryption magic bytes for 8.0.19+, it's for checking KEYRING
+redo log information version. */
```
satya-bodapati: Also just wanted to know what these letters meant "lRK" and "RKB" ?

satya-bodapati (approved):
Please run on jenkins again. Verify the failures.
Also one valgrind/ASAN run.

storage/innobase/log/log0write.cc
robgolebiowski: thanks :)

storage/innobase/srv/srv0srv.cc
```diff
@@ -2870,7 +2870,18 @@ bool srv_enable_redo_encryption_rk(THD *thd) {
   space->encryption_redo_key = mkey;
   space->flags |= FSP_FLAGS_MASK_ENCRYPTION;
   space->encryption_key_version = version;
-  memcpy(space->encryption_redo_key_uuid, server_uuid,
+  space->encryption_redo_key_uuid.reset(
+      new (std::nothrow) char[ENCRYPTION_SERVER_UUID_LEN + 1]);
+  if (space->encryption_redo_key_uuid.get() == nullptr) {
```
satya-bodapati, line 2875: btw, should you do reset() on fil_space_t destruction.. Although this is done on shutdown only.. for redo log tablespaces..

storage/innobase/include/os0file.h
robgolebiowski: I guess lRK was for Rotated Keys. The other versions names like PSA was easy to upgrade PSB. Here lRB would be confusing, so I changed the version name to RKB - next will be RKC etc.

storage/innobase/srv/srv0srv.cc
robgolebiowski: that's the point of smart_pointer  - when it gets destructed - it will deallocate the memory it points to.

robgolebiowski: and it will get destructed when fil_space_t gets destructed.

dutow (approved):

storage/innobase/srv/srv0srv.cc
satya-bodapati: Right.. reset() is only required f you want to deallocate explicity..

PR #3601, opened 2020-01-10
PS-6150: MySQL crash - sync_mutex_to_string 8.0

kamil-holubicki:
https://jira.percona.com/browse/PS-6150
https://jira.percona.com/browse/PS-6093

Two related issues fixed:

When printing out locked mutexes (SHOW ENGINE INNODB STATUS), LATCH_ID_AUTOINC object was cast to wrong type in sync0arr.cc sync_array_cell_print(), because of wrong request_type stored in cell. That caused calling GenericPolicy::to_string() instead of AggregateMutexStatsPolicy::to_string(). As the result of wrong cast, m_id member of policy used inside to_string() contained random value. This value was used to index the array in sync_latch_get_meta() which caused memory access violation.

When locking of the mutex reaches OS level, sync_array cell is created in ut0mutex.ic TTASEventMutex::wait() and LATCH_ID_BUF_POOL_ZIP is mapped to SYNC_BUF_BLOCK cell type. This causes proper cast of mutex in sync0arr.cc sync_array_cell_print() to AggregateMutexStatsPolicy::to_string(). However to_string() method did not service LATCH_ID_BUF_POOL_ZIP causing assertion.
MTR test covering this case was not implemented, because the problem shows up when innodb internal threads lock zip_mutex (buf_flush_page_cleaner_coordinator thread and buf_lru_manager). Locking of zip_mutex on OS level can be reached by heavy insertion into the compressed table (and setting innodb_lru_scan_depth to high value) which will cause InnoDB to aggressively flush buffer pool pages.
However, as locking threads are InnoDB internal threads, and DEBUG_SYNC facility is thread aware, there is no way to setup debug sync points from MTR test level.

### DIFF
--- a/mysql-test/suite/innodb/r/percona_extended_innodb_status.result
+++ b/mysql-test/suite/innodb/r/percona_extended_innodb_status.result
@@ -67,3 +67,19 @@ DROP TABLE t;
 mysql -e "SHOW ENGINE INNODB STATUS"
 Grepping InnoDB status for Buffer pool size, bytes 25149440
 Pattern "Buffer pool size, bytes 25149440" found
+#
+# PS-6150: MySQL crash - sync_mutex_to_string
+#
+CREATE TABLE t1(id INT NOT NULL AUTO_INCREMENT, a INT, PRIMARY KEY(id)) ENGINE=InnoDB;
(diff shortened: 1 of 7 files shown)


kamil-holubicki:
https://ps80.cd.percona.com/job/percona-server-8.0-param/489/

satya-bodapati (approved):
LGTM!

PR #3602, opened 2020-01-10
PS-6761: macOS: threadpool_unix.cc error: no matching member function…

inikep:
… for call to 'compare_exchange_weak'

### DIFF
--- a/sql/threadpool_unix.cc
+++ b/sql/threadpool_unix.cc
@@ -1304,7 +1304,7 @@ void tp_wait_end(THD *thd) {
 static void set_next_timeout_check(ulonglong abstime) {
   DBUG_ENTER("set_next_timeout_check");
   while (abstime < pool_timer.next_timeout_check.load()) {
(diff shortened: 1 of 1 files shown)


sql/threadpool_unix.cc
```diff
@@ -1304,7 +1304,7 @@ void tp_wait_end(THD *thd) {
 static void set_next_timeout_check(ulonglong abstime) {
   DBUG_ENTER("set_next_timeout_check");
   while (abstime < pool_timer.next_timeout_check.load()) {
-    ulong old = pool_timer.next_timeout_check.load();
+    ulonglong old = pool_timer.next_timeout_check.load();
```
percona-ysorokin: `next_timeout_check` is declared as
```
  std::atomic<uint64> next_timeout_check;
```
So, may be 
```suggestion
    uint64 old = pool_timer.next_timeout_check.load();
```
or even `auto`

percona-ysorokin (approved):
LGTM with minor comments

PR #3608, opened 2020-01-15
PS-6775: Remove KEYRING_ON option from default_table_encryption.

robgolebiowski:
Value of default_table_encryption is inherited by:
 - CREATE TABLESPACE's ENCRYPTION clause,
 - CREATE DATABASE's ENCRYPTION clause

Since we are not supporting KEYRING encryption for tablespaces we
decided to remove KEYRING option from default_table_encryption. The
implication of that is that DEFAULT ENCRYPTION will not inherit KEYRING
value from d_t_e. We have not yet implemented that (DEFAULT
ENCRYPTION=KEYRING) so this is OK - we double check it was not
implemented in encryption.database test.


robgolebiowski:
JR => https://ps80.cd.percona.com/job/percona-server-8.0-param/493/

satya-bodapati:
> Value of default_table_encryption is inherited by:
> 
>     * CREATE TABLESPACE's ENCRYPTION clause,
> 
>     * CREATE DATABASE's ENCRYPTION clause
> 
> 
> Since we are not supporting KEYRING encryption for tablespaces we
> decided to remove KEYRING option from default_table_encryption. The
> implication of that is that DEFAULT ENCRYPTION will not inherit KEYRING
> value from d_t_e. We have not yet implemented that (DEFAULT
> ENCRYPTION=KEYRING) so this is OK - we double check it was not
> implemented in encryption.database test.

" Since we are not supporting KEYRING encryption for tablespaces" --> This seems like we dont support KEYRING encryption at all. Can you clarify here that explicit creation of KEYRING tables are not supported (for general tablespaces) and background threads can still convert tables to KEYRING encrypted.

mysql-test/collections/disabled.def
```diff
@@ -168,7 +168,6 @@ tokudb_rpl.rpl_deadlock_tokudb : BUG#0 temporarily disabled until properly debug
 tokudb_rpl.rpl_partition_tokudb : BUG#0 exchange partition is not supported
 
 # encryption suite tests
-encryption.encryption_force : BUG#5817 Align Keyring encryption and default_table_encryption
 encryption.innodb_lotoftables : BUG#5817 Align Keyring encryption and default_table_encryption
 encryption.innodb-bad-key-change2 : BUG#5322 Keyring encrypted tablespace import/export is broken
 encryption.innodb-bad-key-change3 : BUG#5322 Keyring encrypted tablespace import/export is broken
```
satya-bodapati, line 173: Just observation(need not fix in this PR), import export is fixed and these tests can be enabled?

satya-bodapati (approved):
LGTM! Please fix the commit message a bit
(Ensure no new test failures. Leaving the responsibility to you ;))

satya-bodapati:
Btw, not related to this PR, in general I feel variables are abused here. Done by upstream as well. This is fastest way to control a system without touching lot of server code.

IMHO, the allowed values of d_t_e should be ON(MK), OFF and KEYRING (when allowed)

And actions to convert from MK_TO_KEYRING, KEYRING_TO_UNENCRYPTED etc should have been conveyed via SQL.
In my dream, I wish it would be like "ALTER INSTANCE CHANGE ENCRYPTION_MODE TO KEYRING" or "ALTER INSTANCE CHANGE ENCRYPTION_MODE To UNENCRYPTED" etc.. (sample commands)..

Definitely not talking about implementing this in PR ;) Just saying because it seems related :)

mysql-test/collections/disabled.def
robgolebiowski: when import export is fixed then yes, but not yet.

PR #3616, opened 2020-01-17
PS-5552: Assertion `m_idx >= 0' failed in plan_idx QEP_share d::idx() const

kamil-holubicki:
https://jira.percona.com/browse/PS-5552

Note: This issue affected only debug build as it was caught by assert, without having side effects in release build.

When subselect_hash_sj_engine is created it creates its internal temporary table where IN subquery is materialized. Temporary table related objects are initialized only partially.
(introduced by commit e0a309bd92f12b344c9995511d47fd7aee07fd70)

Then if the result does not fit into heap tmp table, the tmp table switches to on-disk table and trace for new table is added. This uses idx member from QEP_TAB which is not initialized.
(introduced by commit 68151294e444692fb2d639e8b557cbb3f0730b47)

The fix is to initialize idx member to value 0. It is not used anyway, but assertion does not complain.


kamil-holubicki:
https://ps80.cd.percona.com/job/percona-server-8.0-param/500
failed tests not related to the fix.

percona-ysorokin (approved):
LGTM

PR #3631, opened 2020-01-28
PS-4649 [TokuDB] Add section about PerconaFT (fractal tree indexing) to PS docs

patrickbirch:
Merged with 5.7 version
modified file
Index
added file
tokudb_fractal_tree_indexing


george-lorch (approved):
+1

PR #3632, opened 2020-01-28
PS-5425 MyRocks - Write Ahead Log (WAL) is not archived unless rocksd…

patrickbirch:
…b_wal_ttl_seconds is set to non-zero value

Added a note to the variable

### DIFF
--- a/doc/source/myrocks/variables.rst
+++ b/doc/source/myrocks/variables.rst
@@ -2385,7 +2385,7 @@ Allowed range is up to ``9223372036854775807``.
   :default: ``0``
 
 Specifies the timeout in seconds before deleting archived WAL files.
-Default is ``0`` (archived WAL files are never deleted).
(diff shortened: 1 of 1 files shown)


doc/source/myrocks/variables.rst
```diff
@@ -2388,6 +2388,11 @@ Specifies the timeout in seconds before deleting archived WAL files.
 Default is ``0`` (archived WAL files are never deleted).
```
george-lorch: This should probably say something like "(WAL files are NOT archived)" instead.

george-lorch (changes requested):

doc/source/myrocks/variables.rst
patrickbirch: Good point. Fixed.

george-lorch (approved):

PR #3640, opened 2020-01-31
PS-5955: Must not be possible to grant or revoke a role to/from utili…

dutow:
…ty user

### DIFF
--- a/storage/perfschema/table_ees_by_thread_by_error.cc
+++ b/storage/perfschema/table_ees_by_thread_by_error.cc
@@ -46,6 +46,9 @@ bool PFS_index_ees_by_thread_by_error::match(PFS_thread *pfs) {
     if (!m_key_1.match(pfs)) {
       return false;
     }
+    if (pfs->m_disable_instrumentation) {
+      return false;
+    }
   }
 
   return true;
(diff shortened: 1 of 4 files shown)


storage/perfschema/table_ees_by_thread_by_error.cc
```diff
@@ -46,6 +46,9 @@ bool PFS_index_ees_by_thread_by_error::match(PFS_thread *pfs) {
     if (!m_key_1.match(pfs)) {
       return false;
     }
+    if (pfs->m_disable_instrumentation) {
```
percona-ysorokin, line 49: Does this fix a related problem?

dutow: Looks like this is an improvement for PS-5952. In a comment there Mohit found several performance schema tables there which expose utility user threads, and some of those can't be fixed. When I fixed this grant bug, I also discovered that I can fix one more of those visibility issue. Probably I added it here because PS-5952 was already closed, and I didn't want to create a separate ticket for this one line fix.

I edited the commit message to at least mention this.

percona-ysorokin (approved):
LGTM

PR #3643, opened 2020-02-03
PS-6824 Debian/Ubuntu installation - add gnupg2

patrickbirch:
modified file:
added the step to install gnupg2 to the Installing Percona Server for MySQL from Percona apt repository section


tplavcic (approved):
LGTM

PR #3647, opened 2020-02-04
PS-5844 Memory leak after 'innodb.alter_crash' - in 'prepare_inplace_…

altmannmarcelo:
…alter_table_dict()'

The problem
If online alter fails to allocate undo log, heap memory allocated to
index object won't be freed. This can be simulated by using debug point
'ib_create_table_fail_too_many_trx' prior to an alter table.

Solution
If we fail to allocate undo log, prior to set index pointer to null,
call dict_mem_index_free to correctly free memory objects allocated
by the index.

### DIFF
--- a/storage/innobase/row/row0merge.cc
+++ b/storage/innobase/row/row0merge.cc
@@ -4338,6 +4338,11 @@ row_merge_create_index(
 		this index, to ensure read consistency. */
 		ut_ad(index->trx_id == trx->id);
 	} else {
+		/* In case we were unable to assign an undo record for this index
+		we won't free index memory object */
+		if (err == DB_TOO_MANY_CONCURRENT_TRXS) {
+			dict_mem_index_free(index);
+		}
 		index = NULL;
 	}
 


altmannmarcelo:
https://ps57.cd.percona.com/job/percona-server-5.7-param/217/

storage/innobase/row/row0merge.cc
```diff
@@ -4338,6 +4338,12 @@ row_merge_create_index(
 		this index, to ensure read consistency. */
 		ut_ad(index->trx_id == trx->id);
 	} else {
+		/* In case we were unable to assign an undo record for this index
+		we won't free index memory object */
+		if(err == DB_TOO_MANY_CONCURRENT_TRXS)
```
satya-bodapati: where is it freed for other err codes?

Alos leave space after if

satya-bodapati (changes requested):

storage/innobase/row/row0merge.cc
altmannmarcelo: It happens on different places depending on the error. 
https://github.com/percona/percona-server/blob/5.7/storage/innobase/dict/dict0dict.cc#L2699 
https://github.com/percona/percona-server/blob/5.7/storage/innobase/dict/dict0dict.cc#L2728

For `DB_TOO_MANY_CONCURRENT_TRXS` it's missing. It seems to be a special case that needs to be handled manually, since there are other places in the code that also filter this error in particular for example at [row_merge_drop_indexes_dict](https://github.com/percona/percona-server/blob/5.7/storage/innobase/row/row0merge.cc#L3607)

storage/innobase/row/row0merge.cc
```diff
@@ -4338,6 +4338,11 @@ row_merge_create_index(
 		this index, to ensure read consistency. */
 		ut_ad(index->trx_id == trx->id);
 	} else {
+		/* In case we were unable to assign an undo record for this index
+		we won't free index memory object */
+		if(err == DB_TOO_MANY_CONCURRENT_TRXS) {
```
satya-bodapati:
```suggestion
		if (err == DB_TOO_MANY_CONCURRENT_TRXS) {
```

storage/innobase/row/row0merge.cc
satya-bodapati: Ideally there should be one place to free this object for all error paths. But I guess it is more complicated. The responsibility is left to callers.

So ok with your patch. Please check the style suggestion (space issue)

PR #3651, opened 2020-02-05
PS-4675 Documented a sample bash script for Tokubackup to create time…

patrickbirch:
…stamped backup directories

modified file: toku_backup
added section on timestamp backup directories


george-lorch (approved):
LGTM

PR #3652, opened 2020-02-06
PS-4675 Documented a sample bash script for TokuBackup to create timestamped backup directories

patrickbirch:
Merged with 5.6 version.


george-lorch (approved):
+1

PR #3653, opened 2020-02-10
PS-4464: Expose GTID_EXECUTED for CONSISTENT SNAPSHOT

kamil-holubicki:
This patch adds a status variable `Binlog_snapshot_gtid_executed` with the following changes:
1. Add tracking of GTID_EXECUTED in `Gtid_state::gtid_executed`
2. Add function to snapshot GTID_EXECUTED with `Gtid_state::get_snapshot_gtid_executed()`
3. Add a status variable `Binlog_snapshot_gtid_executed`
4. Fix/re-record tests where new status is exposed in results.


kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-param/221/
MTR errors not related to the change.

sql/binlog.cc
```diff
@@ -112,13 +112,18 @@ static ulonglong binlog_global_snapshot_position;
...
+
+static const char *snapshot_gtid_executed_var_name = "snapshot_gtid_executed";
 
 static SHOW_VAR binlog_status_vars_detail[]=
 {
   {"snapshot_file",
    (char *)&binlog_snapshot_file, SHOW_CHAR, SHOW_SCOPE_GLOBAL},
   {"snapshot_position",
    (char *)&binlog_snapshot_position, SHOW_LONGLONG, SHOW_SCOPE_GLOBAL},
+  {snapshot_gtid_executed_var_name, &binlog_snapshot_gtid_executed[0], SHOW_CHAR,
```
percona-ysorokin: This `&binlog_snapshot_gtid_executed[0]` looks extremely dangerous - relying on a pointer from std::string implementation than can be changed on reallocation, even without taking into account small string optimization(SSO), is scary.
I would go with conventional MySQL approach here - static buffer / my_malloc()+my_free()

sql/rpl_gtid_state.cc
```diff
@@ -148,6 +148,21 @@ void Gtid_state::broadcast_owned_sidnos(const THD *thd)
   }
 }
 
+void Gtid_state::get_snapshot_gtid_executed(
+    std::string &snapshot_gtid_executed)
+{
+  global_sid_lock->wrlock();
+  snapshot_gtid_executed.clear();
+  size_t size= executed_gtids.get_string_length() + 1;
+  char * buf= (char *)malloc(size);
```
percona-ysorokin: I would go with creating `std::vector<char> buf(size);` here as a temp buffer, then
`executed_gtids.to_string(buf.data());`
and
`snapshot_gtid_executed.assign(buf.begin(), buf.end())`

sql/rpl_gtid.h
```diff
@@ -2986,6 +2986,9 @@ class Gtid_state
   int warn_or_err_on_modify_gtid_table(THD *thd, TABLE_LIST *table);
 #endif
 
+  /* Allow to snapshot GTID_EXECUTED */
```
percona-ysorokin:
```suggestion
  /* Allows to extract snapshot GTID_EXECUTED */
```

mysql-test/mysql-test-run.pl
```diff
@@ -194,7 +194,7 @@ sub set_term_args {
 
 # If you add a new suite, please check TEST_DIRS in Makefile.am.
 #
-my $DEFAULT_SUITES= "main,sys_vars,binlog,binlog_encryption,rpl_encryption,encryption,"
+my $DEFAULT_SUITES= "main,sys_vars,binlog,binlog_encryption,binlog_gtid,rpl_encryption,encryption,"
```
percona-ysorokin: Not sure I like introducing new `binlog_gtid` suite in 5.7 - it's more like 8.0 refactoring feature.
I am not saying it does not make sense, but in order to stick closer to the upstream code I wouldn't do this

mysql-test/include/validate_binlog_snapshot_gtid.inc
```diff
@@ -0,0 +1,33 @@
+--let $snapshot_gtid_executed = query_get_value(SHOW STATUS LIKE "Binlog_snapshot_gtid_executed", Value, 1)
```
percona-ysorokin, line 1: Not sure it this helps but 5.7 and 8.0 have GTID set functions
https://dev.mysql.com/doc/refman/5.7/en/gtid-functions.html

percona-ysorokin (changes requested):

sql/binlog.cc
kamil-holubicki: Hi @percona-ysorokin, std::string binlog_snapshot_gtid_executed was acting as the object managing allocation and deallocation of dynamic buffer holding gtid_executed string. Here I just initialized the structure with some value, which does not matter anyway (I could initialize it with nullptr), because it was always updated when variable was read by upper layer (show_binlog_vars()).
Anyway, I have changed the implementation as requested.

sql/rpl_gtid_state.cc
kamil-holubicki: done.
I didn't use std::vector initially, because if I remember well, some time ago std::vector was not necessarily continuous according to the standard. After checking again, it seems that it was in ancient ages :)

sql/rpl_gtid.h
kamil-holubicki: done

mysql-test/mysql-test-run.pl
kamil-holubicki: I decided to introduce new suite, because this tests ideally fits binlog_gtid suite that is present in 8.0. Of course I can move it somewhere else. Please let me know if you want me to do this.

mysql-test/include/validate_binlog_snapshot_gtid.inc
kamil-holubicki: Good idea to use GTID_SUBSET() however in this case it will not work because:
1. (Most important) binlog_snapshot_gitd_executed can return 'not-in-consistent-snapshot' value if we are not in transaction with consistent snapshot. This will not work with GTID_SUBSET()
2. We need to check if what we expect is exactly what we get (not the subset)

mysql-test/mysql-test-run.pl
kamil-holubicki: Moved to suite/binlog as all binlog_gtid related tests are there as well in 5.7. For 8.0 the test will go to suite/binlog_gtid.

sql/binlog.cc
```diff
@@ -112,13 +112,24 @@ static ulonglong binlog_global_snapshot_position;
...
 static SHOW_VAR binlog_status_vars_detail[]=
 {
   {"snapshot_file",
    (char *)&binlog_snapshot_file, SHOW_CHAR, SHOW_SCOPE_GLOBAL},
   {"snapshot_position",
    (char *)&binlog_snapshot_position, SHOW_LONGLONG, SHOW_SCOPE_GLOBAL},
+   /* Structure needs to be initialized with some value pointer (null here).
+   It will be updated to the current, dynamically allocated one
+   in set_binlog_snapshot_gtid_executed() */
+  {snapshot_gtid_executed_var_name, binlog_snapshot_gtid_executed, SHOW_CHAR,
```
percona-ysorokin: @kamil-holubicki Digging a bit deeper I encountered that there is also SHOW_FUNC kind of SHOW_VAR, that in my opinion is more suitable here.
Check `caching_sha2_password_status_variables` in `sha2_password.cc` as an example.
In this case, you will be able to have global std::string object (as in the previous iteration) and all manual memory management should go away.
Moreover, you won't need `get_var_index()` in this case

percona-ysorokin (changes requested):

sql/binlog.cc
kamil-holubicki: Done.

sql/binlog.cc
```diff
@@ -1027,10 +1028,38 @@ class binlog_cache_mngr {
...
+    Check if manager contains consistent snapshot of log coordinates
+    and gtid_executed.
+
+    @return true  Consistent snapshot available
+    @return false Otherwise
+   */
+  bool has_consistent_snapshot() const
+  {
+    bool has_log_file_name= binlog_info.log_file_name[0] != '\0';
+    bool has_gtid_executed= snapshot_gtid_executed.length() > 0;
```
percona-ysorokin: This may trigger `unused variable` in release mode

percona-ysorokin (approved):
LGTM with minor comment

kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-param/228/
MTR failures not related to the changes.

PR #3665, opened 2020-02-19
PS-6820: Add compile time option to enable MyRocks in Group Replication

inikep:
This patch adds a compile time option `GROUP_REPLICATION_WITH_ROCKSDB` for the Group Replication plugin that enables MyRocks storage engine.
Usage: `cmake -DGROUP_REPLICATION_WITH_ROCKSDB=ON`


inikep:
https://ps80.cd.percona.com/job/percona-server-8.0-param/518/

plugin/group_replication/CMakeLists.txt
```diff
@@ -113,6 +113,14 @@ ADD_DEFINITIONS(-DMYSQL_SERVER)
 ADD_DEFINITIONS(-DLOG_SUBSYSTEM_TAG="Repl")
 ADD_DEFINITIONS(-DLOG_COMPONENT_TAG="group_replication")
 
+IF(GROUP_REPLICATION_WITH_ROCKSDB)
```
percona-ysorokin, line 116: Why do we need introduce a new cmake option `GROUP_REPLICATION_WITH_ROCKSDB`?
Shouldn't
```
      if (trans_param->tables_info[table].db_type != DB_TYPE_INNODB &&
          trans_param->tables_info[table].db_type != DB_TYPE_ROCKSDB) {
```
always be enabled when we compile with RocksDB?

A bit rephrasing, is having RocksDB enabled and deliberately not wanting group replication plagin to have support for it a valid use case?

george-lorch: Yes @percona-ysorokin  MyRocks DOES NOT WORK with GR and we do not want anyone in the general public to accidentally use it this way.  This option is to allow us to build it in a way that one can enable it for the purposes of testing.

percona-ysorokin: Oh. Now it makes sense. Thanks @georgelorchpercona

percona-ysorokin (approved):
LGTM

PR #3666, opened 2020-02-20
PS-6112: Inconsistent Binlog_snapshot_gtid when mysqldump used with --single-transaction (5.7)

kamil-holubicki:

The fix is to use 'Binlog_snapshot_gtid_executed' variable introduced by PS-4464. If the variable is not supported by server, behavior falls back to upstream's mysqldump behavior.

Additionally fixed wrong error code returned by check_consistent_binlog_pos() in case of query execution failure.

Two options were added to mysqldump: start-sql-file and finish-sql-file. This allows better integration of mysqldump with mtr tests as debug sync facility asserver sync points are activated per thread.

This is not GCA commit because it needs changes done in PS-4464


kamil-holubicki:
This is heavily reworked #3587. This time the solution is based on new functionality provided by PS-4464.
https://ps57.cd.percona.com/job/percona-server-5.7-param/231/
MTR test failures not related to changes.

client/mysqldump.c
```diff
@@ -6673,43 +6715,90 @@ static void set_session_binlog(my_bool flag)
...
+    {
+      fprintf(stderr,
+              "\nWARNING: Server does not support consistent snapshot of "
+              "GTID_EXECUTED."
+              "\nThe value provided for GTID_PURGED may be inaccurate!"
+              "\nTo overcome this issue, start mysqldump with "
+              "--lock-all-tables.\n");
+
+      fprintf(md_result_file,
+              "\n\n--"
```
percona-ysorokin, line 6754: Probably define a string literal with this message

mysql-test/include/wait_proc_to_finish.inc
```diff
@@ -0,0 +1,37 @@
...
+}
+
+--let _PID_FILE = $pid_file
+
+perl;
+  use Errno;
+  my $pid_file = $ENV{'_PID_FILE'};
+  open(my $f, '<', $pid_file) or die "OPENING $pid_file: $!\n";
+  my $pid = do { local($/); <$f> };
+  close($f);  use strict;
```
percona-ysorokin: formatting

mysql-test/t/mysqldump_gtid_purged.inc
```diff
@@ -0,0 +1,91 @@
...
+
+# Start mysqldump in background collecting its PID
+--let $command = $MYSQL_DUMP
+--let $command_opt = $MYSQLDUMP_OPTS test t1 --start-sql-file=$start_sql_file --finish-sql-file=$finish_sql_file
+--let $output_file = $MYSQLTEST_VARDIR/tmp/dump.sql
+--let $pid_file = $MYSQLTEST_VARDIR/tmp/mysqldump.pid
+--source include/start_proc_in_background.inc
+
+# Wait for mysqldump to reach the point just before getting gtid_executed
+SET DEBUG_SYNC = 'now WAIT_FOR before_show_gtid_executed.reached';
```
percona-ysorokin, line 39: Is it OK if `before_show_gtid_executed.reached` is signaled before (by mysqldump) we start waiting on it here?

mysql-test/t/mysqldump_gtid_purged.inc
```diff
@@ -0,0 +1,91 @@
...
+
+
+# Now mysqldump started its job.
+# Following insert will change @@global.gtid_executed, but it should not change the snapshot
+# It may block if mysqldump was started with --lock-all-tables, so we need to do it in async way
+--send INSERT INTO t1 VALUES (1)
+--real_sleep $sleep_sec
```
percona-ysorokin, line 46: What is the  reason for this `real_sleep`?

mysql-test/t/mysqldump_gtid_purged.inc
```diff
@@ -0,0 +1,91 @@
...
+--connection con1
+SET DEBUG_SYNC = 'now SIGNAL before_show_gtid_executed.continue';
+
+# Back to default connection. Get result of insert as it should be unblocked when gtid_executed is queried
+# and wait for mysqldump to finish
+--connection default
+--reap
+SET DEBUG_SYNC = 'now WAIT_FOR mysqldump.finished';
+
+# gtid_executed changed, but mysqldump should have grab the one that was present before 2nd insert.
```
percona-ysorokin:
```suggestion
# gtid_executed changed, but mysqldump should have grabbed the one that was present before 2nd insert.
```

mysql-test/t/mysqldump_gtid_purged.inc
```diff
@@ -0,0 +1,91 @@
...
+# mysqldump signaled its finish, however process may not exited yet and dump file probably is not flushed yet.
```
percona-ysorokin:
```suggestion
# mysqldump signaled its finish, however process may have not exited yet and dump file probably is not flushed yet.
```

mysql-test/t/mysqldump_gtid_purged.inc
```diff
@@ -0,0 +1,91 @@
...
+--source include/grep_pattern.inc
+--let $grep_pattern = origin: @@global.gtid_executed
+--echo grep_pattern: $grep_pattern
+--source include/grep_pattern.inc
+
+# cleanup
+--disconnect con1
+DROP TABLE t1;
+SET DEBUG_SYNC = 'RESET';
+#--remove_file $output_file
```
percona-ysorokin: Should this be uncommented?

client/mysqldump.c
kamil-holubicki: Please note that these 2 strings are not identical. The first one is intended to be print out to the stderr and the second one is to be included in generated sql file (message is put in comment).
I can extract them to static const char * msg = ... static const char *msg2 = ... but we will still have 2 strings.
Moreover the convention of mysqldump file is like it is implemented right now.

mysql-test/t/mysqldump_gtid_purged.inc
kamil-holubicki: The point here is that mysqldump is started in async way. We need to stop it just before gtid_executed is queried, but after all potential locks and transaction is started (P1). So mysqldump signals that that it reached P1 and waits for signal to continue.
MTR test starts mysqldump and 
Case 1:
if mysqldump gets to P1 as the first one it will emit 'reached' singnal.  When MTR get to the wait_for 'reached' signal, it will just go ahead as singlal was already emitted.
Case 2:
MTR gets to wait_for 'reached' point before mysqldump reaches P1 and signals 'reached'. In this case MTR will wait for 'reached' singal emitted by mysqldump when it gets to P1.

Answering the question: Yes, it is OK.

mysql-test/t/mysqldump_gtid_purged.inc
kamil-holubicki: This 'inc' file is used several times during the test to test mysqldump behavior when it is started with different parameters. Depending on parameters, mysqldump may block all tables with read lock (FLUSH TABLES WITH READ LOCK) which will cause all inserts to block, or it may issue LOCK TABLES FOR BACKUP which will NOT cause inserts to block. (BTW this is also the reason why INSET is done from parallel connection in 'async' way. For cases when inserts are not blocked, it could have been done from default connection) 
During the test we need to be sure if INSERT was fully executed, or it was blocked. So the reason for this sleep is to give 5 seconds time for it. Probably it could be achieved with some debug sync points, but there is no single point for these 2 cases that would signal that what we needed was done (blocked or inserted). 
That's why I decided to keep things simple and put sleep here.

mysql-test/include/wait_proc_to_finish.inc
kamil-holubicki: done

mysql-test/t/mysqldump_gtid_purged.inc
kamil-holubicki: done

mysql-test/t/mysqldump_gtid_purged.inc
kamil-holubicki: done

mysql-test/t/mysqldump_gtid_purged.inc
kamil-holubicki: Sure, side effect of the battle with MTR test.

mysql-test/t/mysqldump_gtid_purged.inc
percona-ysorokin: Yes, My primary concern was `Case 1` - for some reason I thought that if you emit a signal and nobody is waiting for it, then it just gets lost. If you know it works, then no problem here.

client/mysqldump.c
percona-ysorokin: My mistake, it looked to me that they were the same

mysql-test/t/mysqldump_gtid_purged.inc
percona-ysorokin: OK

percona-ysorokin (approved):
LGTM

PR #3667, opened 2020-02-20
PS-6112: Inconsistent Binlog_snapshot_gtid when mysqldump used with --single-transaction (8.0)

kamil-holubicki:

The fix is to use 'Binlog_snapshot_gtid_executed' variable introduced by PS-4464. If the variable is not supported by server, behavior falls back to upstream's mysqldump behavior.

Two options were added to mysqldump: start-sql-file and finish-sql-file. This allows better integration of mysqldump with MTR tests as debug sync facility as server sync points are activated per thread.

This is not GCA commit because it needs changes done in PS-4464


kamil-holubicki:
https://ps80.cd.percona.com/job/percona-server-8.0-param/520/
Failed MTR tests not related to the change.

client/mysqldump.cc
```diff
@@ -1252,6 +1269,30 @@ static int fetch_db_collation(const char *db_name, char *db_cl_name,
...
 }
 
+/*
+  Check if server supports binlog_snapshot_gtid_executed.
+  Returns 1 supported, 0 if not.
+ */
+static int consistent_gtid_executed_supported(MYSQL *mysql_con) {
+  MYSQL_RES *res;
+  MYSQL_ROW row;
+  int found = 0;
```
percona-ysorokin:
```suggestion
  bool found = 0;
```

percona-ysorokin (approved):
LGTM with minor comment (can be ignored)

client/mysqldump.cc
kamil-holubicki: done

PR #3671, opened 2020-02-24
PS-5323: Align Keyring encryption with Master Key encryption.

robgolebiowski:
Tablespaces, system tablespace, undo tablespaces can be either KEYRING
encrypted or Master Key encrypted, but cannot be both.

Tablespaces cannot be created with Master Key encryption when online to
KEYRING encryption is turned on. Otherwise the tablespaces would be
immediately re-encrypted with KEYRING encryption, so allowing creating
Master Key encrypted tablespaces in such case does not make sense.

When tablespace's ENCRYPTION is altered (Master Key encryption), the
tablespace itself is excluded from encryption threads.
I.e. encryption threads cannot encrypt/decrypt tablespace while
it is being encrypted/decrypted to/from Master Key encryption.
When tablespace cannot be excluded from encryption threads - due to
it being actively encrypted/decrypted by encryption threads -
appropriate error message is generated.

It is not possible to encrypt with Master Key encryption
(ENCRYPTION='N') tablespace that is partially/fully encrypted by
encryption threads. In order to encrypt such tablespace with
encryption threads, user has to first decrypt it using encryption
threads and then encrypt it with Master Key encryption.

System tablespace Master Key is bootstrap only option. We disallow
enabling online to KEYRING re-encryption (when system tablespace is
Master Key encrypted) unless innodb-sys-tablespace-encrypt is set to
RE_ENCRYPTING_TO_KEYRING.

innodb-sys-tablespace-encrypt was changed from boolean readonly variable
to enum dynamic variable. This is to allow setting
innodb-sys-tablespace-encrypt variable to RE_ENCRYPTING_TO_KEYRING.
It is not possible to change this variable from ON to OFF and the
other way around. However, when user tries to do so, appropriate
warning is generated, with a hint, that encryption threads can be used
for dynamic re-encryption of system tablespace.

Undo tablespace Master Key encryption cannot be turned ON when online to
KEYRING encryption is turned ON, as it would immediately be re-encrypted
with KEYRING encryption. Undo tablespace encryption can be dynamically
changed from Master Key encryption to KEYRING encryption and in the
opposite direction. However, to change from KEYRING encryption to Master
Key encryption the undo tablespaces must first be decrypted by
encryption threads. Otherwise KEYRING encryption information stored in
page0 of undo tablespace would be overwritten by Master Key encryption
information. These would make logs already encrypted with KEYRING
encryption unreadable.

When:
1) DEFAULT_TABLE_ENCRYPTION == ONLINE_TO_KEYRING
- no MK encrypted tables/tablespace can be created, as they would be
immediately re-encrypted with encryption threads.
- no DATABASE can be created with ENCRYPTION=Y/N :
  in both cases tables created in such database would be immediately
re-encrypted with keyring
- when a table is created without encryption clause - it is encrypted
with "ONLINE_TO_KEYRING" encryption from
  the beginning (i.e. it can be decrypted by encryption threads).

If database was created with ENCRYPTION='Y' and DEFAULT_TABLE_ENCRYPTION
was changed to ONLINE_TO_KEYRING
no new tables can be created in this database as long as
DEFAULT_TABLE_ENCRYPTION == ONLINE_TO_KEYRING.

2) DEFAULT_TABLE_ENCRYPTION == ONLINE_FROM_KEYRING_TO_UNENCRYPTED
- MK encrypted tables/tablespaces can be created
- DATABASEs can be created with ENCRYPTION='Y/N'
- when tables are created without encryption clause - they are created
unencrypted.

When a user wants to change ONLINE_TO_KEYRING encryption to Master Key
encryption s/he needs to:
- decrypt the tablespace with threads
- ALTER ENCRYPTION to 'Y'


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/522/

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
```diff
@@ -0,0 +1,561 @@
+--source include/have_debug.inc
+
+let $MYSQLD_BASEDIR= `select @@basedir`;
```
dutow: SELECT

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
```diff
@@ -0,0 +1,561 @@
...
+--echo # Wait max 10 min for key encryption threads to decrypt all spaces
+--let $wait_timeout= 600
+--let $wait_condition=SELECT COUNT(*) = 0 FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION = 1
+--source include/wait_condition.inc
+
+--let $assert_text= Make sure ENCRYPTION is set to N in INFORMATION_SCHEMA.INNODB_TABLESPACES
+--let $assert_cond= "[SELECT COUNT(*) FROM INFORMATION_SCHEMA.INNODB_TABLESPACES WHERE NAME=\\'ts0\\' AND ENCRYPTION=\\'N\\']" = 1
+--source include/assert.inc
+
+# Now it should be possible to encrypt ts0 with MK encryption,
```
dutow: `--echo`?

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
```diff
@@ -0,0 +1,561 @@
...
+SET GLOBAL default_table_encryption=OFF;
+SET GLOBAL innodb_encryption_threads=0;
+
+CREATE TABLESPACE ts1 ADD DATAFILE 'ts1.ibd';
+
+SET DEBUG_SYNC = 'alter_encryption_to_y_tablespace_excluded SIGNAL t1_excluded WAIT_FOR t1_was_considered_for_rotation';
+--send ALTER TABLESPACE ts1 ENCRYPTION='Y'
+
+--connection con1
+SET DEBUG_SYNC = 'now WAIT_FOR t1_excluded';
```
dutow, line 208: Please use consistent spacing around `=` in this testcase

sql/sys_vars.cc
```diff
@@ -7111,6 +7111,28 @@ static Sys_var_uint Sys_immediate_server_version(
     BLOCK_SIZE(1), NO_MUTEX_GUARD, IN_BINLOG,
     ON_CHECK(check_session_admin_or_replication_applier));
 
+static bool check_set_default_table_encryption_exclusions(THD *thd,
+                                                          set_var *var) {
+  longlong val = (longlong)var->save_result.ulonglong_value;
```
dutow: `static_cast<longlong>`

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1074,7 +1074,52 @@ static void fil_crypt_write_crypt_data_to_page0(fil_space_t *space) {
   mtr.commit();
 }
 
-bool fil_crypt_exclude_tablespace_from_rotation(fil_space_t *space) {
+bool fil_crypt_exclude_tablespace_from_rotation_temporarily(
+    fil_space_t *space) {
+  // We acquire fil_crypt_threads_mutex to stop encryption threads from
+  // generating crypt_data for tablespaces that they are about to encrypt.
+  // Generating crypt_data is the fist step in encryption process.
+  mutex_enter(&fil_crypt_threads_mutex);
```
dutow, line 1094: can't we use `MUTEX_LOCK` instead?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1074,7 +1074,52 @@ static void fil_crypt_write_crypt_data_to_page0(fil_space_t *space) {
...
+  if (space->crypt_data == nullptr) {
+    space->exclude_from_rotation = true;
+    mutex_exit(&fil_crypt_threads_mutex);
+    return true;
+  }
+
+  mutex_exit(&fil_crypt_threads_mutex);
+
+  // crypt_data already existed.
+  fil_space_crypt_t *crypt_data = space->crypt_data;
```
dutow, line 1105: Decryption won't remove the crypt_data? Or encryption threads never create a new one?

Because at this point we don't hold any mutexes, I'm wondering if this pointer is guaranteed to be valid / current.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1074,7 +1074,52 @@ static void fil_crypt_write_crypt_data_to_page0(fil_space_t *space) {
...
+  // take a lock on crypt_data to block rotation from starting
+  IB_mutex_guard crypt_data_mutex_guard(&crypt_data->mutex);
+
+  if (space->exclude_from_rotation) {
```
dutow: Does this check need the crypt_data mutex?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1474,6 +1534,40 @@ static bool fil_crypt_space_needs_rotation(rotate_thread_t *state,
   return false;
 }
 
+bool fil_crypt_exclude_from_rotation(fil_space_t *space) {
```
dutow: This is a mainly duplicated function, please refactor.

dutow (commented):
(still reviewing, these are just small comments / a few questions)

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
robgolebiowski: done

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
robgolebiowski: done

sql/sys_vars.cc
robgolebiowski: done

storage/innobase/fil/fil0crypt.cc
robgolebiowski: MUTEX_LOCK is available in server code, but not in InnoDB.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: encryption threads never remove crypt_data. If it is created once it stays. That's why it is safe to acquire this lock here.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: no, that's why I have made the exclude_from_rotation atomic operation - so it can be accessed safely with and without crypt_data mutex.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: It was a leftover. Removed it.Thanks for spotting this.

storage/innobase/fil/fil0crypt.cc
dutow: In this case, can't we perform this check at the start of this function, before acquiring any mutexes?

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
robgolebiowski: done

storage/innobase/fil/fil0crypt.cc
robgolebiowski: right, very much so, changing. Thanks :+1:

satya-bodapati:
Its looks a bit inconsistent that undo_log_encrypt is not used to convert existing undo to keyring. Similar to how innodb_redo_encrypt is used.
@satya-bodapati encryption threads encrypt tablespaces. Since undo is a tablespace it is encrypted by encryption threads. It would also looked inconsistent if all tablespaces could be rotated by encryption threads, but not undo tablespace.

satya-bodapati:
"If database was created with ENCRYPTION='Y' and DEFAULT_TABLE_ENCRYPTION
was changed to ONLINE_TO_KEYRING
no new tables can be created in this database as long as
DEFAULT_TABLE_ENCRYPTION == ONLINE_TO_KEYRING."

In this case, What options user has to create keyring encrypted tables? Do we support ALTER DATABASE db DEFAULT_ENCRYPTION='KEYRING'?

@satya-bodapati: No, DEFAULT ENCRYPTION='KEYRING' is not supported on DB level. We can create an improvement to add it in. I am not quite sure if there are any problems with that, that we might have discussed earlier. Apart from that for d_t_e = "ONLINE_TO_KEYRING" we would allow D_E = 'KEYRING' which cannot be decrypted by encryption threads. I do not see this as a big problem though...

@satya-bodapati according to our latest discussions, I will change this behavior - so the ONLINE_KEYRING and ENCRYPTION='KEYRING' tables could be created in databases with D_T_E='Y'

satya-bodapati:
if a MK General tablespace is created with dte=ON, now change dte to ONLINE_TO_KEYRING, tables can be still created in MK General tablespace. Will you prevent this behaviour in this PR?

@satya-bodapati This will be a part of a separate jira ticket and PR. (The ticket: PS-7026: Constrains for keyring encryptions.)

satya-bodapati:
there was no need to change dte to ONLINE_TO_KEYRING. Below SQLs show the problem that we can create keyring tables in MK general tablespace.

```
SELECT @@default_table_encryption
    -> ;
+----------------------------+
| @@default_table_encryption |
+----------------------------+
| ON                         |
+----------------------------+
1 row in set (0.00 sec)

mysql> CREATE TABLESPACE ts2 ADD DATAFILE 'ts2.ibd' ENCRYPTION='Y';
Query OK, 0 rows affected (0.02 sec)

mysql> CREATE TABLE t2 (a INT) TABLESPACE=ts2 ENCRYPTION='KEYRING';
Query OK, 0 rows affected (0.05 sec)

mysql> SHOW CREATE TABLE t2;
+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                    |
+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t2    | CREATE TABLE `t2` (
  `a` int(11) DEFAULT NULL
) /*!50100 TABLESPACE `ts2` */ ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci /*!80016 ENCRYPTION='KEYRING' */ |
+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.02 sec)

mysql> 
```

satya-bodapati:
If dte=ONLINE_TO_KEYRING, it is wise to disallow ALTER DATABSE test DEFAULT ENCRYPTION='Y'. 

@satya-bodapati done

satya-bodapati:
I found  a case where the below contract is broken.

ER_ONLINE_KEYRING_TO_MK_RE_ENCRYPTION
  eng "Tablespace %s is encrypted by encryption threads and cannot be ALTERed to Master Key encryption. In order to encrypt this tablespace with Master Key encryption - first decrypt this tablespace with encryption threads."

create table that is encrypted by background threads.
SET GLOBAL default_table_encryption=ON;
SET GLOBAL innodb_encryption_threads =0;
ALTER TABLE t3 ENCRYPTION='Y';

t3 should be disallowed to converted to MK. But the above ER_ error is NOT thrown. Please check

@satya-bodapati the error is for tablespaces, not tables. Those can be re-encrypted, as they by re-encrypted by foreground thread and copy algorithm. if t3 was a tablespace you would have gotten the error.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -482,6 +482,12 @@ ibool meb_get_checksum_algorithm_enum(const char *algo_name,
 }
 #endif /* !UNIV_HOTBACKUP */
 
+static const char *sys_tablespace_encrypt_names[] = {
+    "OFF", "ON", "RE_ENCRYPTING_TO_KEYRING", NullS};
```
satya-bodapati, line 496: IMHO, the name should be simple "TO_KEYRING" or even  "KEYRING" is even better

satya-bodapati (changes requested):

storage/innobase/handler/ha_innodb.cc
satya-bodapati: if you are open for suggestions,  please consider these

dte variable values could be shortened to  FROM ONLINE_TO_KEYRING to 'TO_KEYRING' and (ONLINE_FROM_KEYRING_TO_UNENCRYPTED' to "TO_UNENCRYPTED" or "TO_MASTER_KEY"

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
```diff
@@ -0,0 +1,561 @@
+--source include/have_debug.inc
+
+let $MYSQLD_BASEDIR= `select @@basedir`;
+
+SET GLOBAL default_table_encryption=ONLINE_TO_KEYRING;
+
+--echo # Online encryption to keyring is ON. It should not be possible to create
+--echo # MK encrypted table or MK encrypted general tablespace.
+--error ER_ILLEGAL_HA_CREATE_OPTION
+CREATE TABLE t1 (a varchar(255)) ENGINE=INNODB ENCRYPTION='Y';
```
satya-bodapati: IMO, ILLEGAL_OPTION is overused, is it possible to use new error code for keyring related errors?

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
```diff
@@ -0,0 +1,561 @@
...
+
+--error ER_ILLEGAL_HA_CREATE_OPTION
+CREATE TABLESPACE `ts1` ADD DATAFILE 'ts1.ibd' ENCRYPTION = 'Y' Engine=InnoDB;
+
+--echo # It should not be possible to turn on undo-tablespace-encrypt
+--error ER_WRONG_VALUE_FOR_VAR
+SET GLOBAL innodb_undo_log_encrypt=ON;
+SHOW WARNINGS;
+
```
satya-bodapati: add SELECT @@innodb_undo_log_encrypt;

share/errmsg-utf8.txt
```diff
@@ -9237,11 +9237,23 @@ ER_DA_UNDO_NO_KEYRING
...
-  eng "Tablespace %s cannot be excluded from encryption threads as it is being currently encrypted/decrypted by encryption threads. Please try again later."
+ER_EXCLUDE_ENCRYPTION_TABLE_ENCRYPTED
+  eng "Tablespace %s is fully or partially encrypted by encryption threads and thus cannot be excluded from encryption threads. Please first finish off encryption/decryption and try again."
 
 ER_EXPLICIT_DECRYPTION_OF_ONLINE_ENCRYPTED_TABLESPACE
-  eng "Tablespace %s is encrypted by encryption threads and cannot be explicitly decrypted. Please use default-table-encryption=ONLINE_FROM_KEYRING_TO_UNENECRYPTED to decypt the tablespace"
+  eng "Tablespace %s is encrypted by encryption threads and cannot be explicitly decrypted. Please use default-table-encryption=ONLINE_FROM_KEYRING_TO_UNENECRYPTED to decypt the tablespace."
+
+ER_ONLINE_KEYRING_TO_MK_RE_ENCRYPTION
+  eng "Tablespace %s is encrypted by encryption threads and cannot be ALTERed to Master Key encryption. In order to encrypt this tablespace with Master Key encryption - first decrypt this tablespace with encryption threads."
```
satya-bodapati:
```suggestion
ER_ONLINE_KEYRING_TO_MK_RE_ENCRYPTION
  eng "Tablespace %s is encrypted by encryption threads and cannot be ALTERed to Master Key encryption. In order to encrypt this tablespace with Master Key encryption, first decrypt this tablespace with encryption threads."
```

sql/sql_db.cc
```diff
@@ -258,8 +258,19 @@ bool mysql_create_db(THD *thd, const char *db, HA_CREATE_INFO *create_info) {
   bool encrypt_schema = false;
   if (create_info->encrypt_type.str) {
     encrypt_schema = dd::is_encrypted(create_info->encrypt_type);
+    if (global_system_variables.default_table_encryption ==
+        DEFAULT_TABLE_ENC_ONLINE_TO_KEYRING) {
+      if (encrypt_schema) {
+        my_error(ER_DATABASE_ENCRYPTION_MK_KEYRING_MISMATCH, MYF(0));
+        return true;
```
satya-bodapati: merge the two return true to outside condition

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1031,7 +1031,7 @@ static bool fil_crypt_needs_rotation(fil_encryption_t encrypt_mode,
 /** Read page 0 and possible crypt data from there.
 @param[in,out]	space		Tablespace */
 static inline void fil_crypt_read_crypt_data(fil_space_t *space) {
-  if (space->crypt_data && space->size) {
+  if (space->size) {
     /* The encryption metadata has already been read, or
     the tablespace is not encrypted and the file has been
     opened already. */
```
satya-bodapati: why is this check removed? Is it nullptr when space->size !=0 ?

satya-bodapati: Also make it (space->size != 0)

satya-bodapati (changes requested):

sql/sql_db.cc
robgolebiowski: done

storage/innobase/fil/fil0crypt.cc
robgolebiowski: I have added a comment in code to make it clear:
   /* When space->size != 0 it means that page0 of the tablespace has
    already been read in order to read the size of the space (i.e. number of the
    pages). When we read page0 - we also read crypt_data information from it - thus
    if space->size != 0 - it means that crypt_data was already read (if existed).
    Also we need to read space->size in order to know how many pages we are going
    to rotate in space. buf_page_get will read space's size if it is not yet read */

storage/innobase/handler/ha_innodb.cc
robgolebiowski: About changing RE_ENCRYPTING_TO_KEYRING - maybe I will try to explain why this name in the first place. We have a system tablespace encrypted with MK encryption, then we decide to encrypt it with encryption threads. You enable encryption threads for it, but I did not want a situation that - you enable encryption threads while having sys_tablespace_encrypt set to ON. You would end up with KEYRING encrypted system tablespace, although sys_tablespace_encrypt would indicated that it is encrypted with MK encryption. Changing it to "TO_KEYRING" or "KEYRING" could indicate that this is enough for system tablespace to be re-encrypted with keyring, although it is not. It is just a marker that encryption threads can re-encrypt sys tablespace.

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
robgolebiowski: done

share/errmsg-utf8.txt

mysql-test/suite/encryption/t/keyring_and_mk_encryption_exclusions.test
robgolebiowski: Changed it to ER_KEYRING_ILLEGAL_ENCRYPTION_OPTION.

satya-bodapati:
> I found a case where the below contract is broken.
> 
> ER_ONLINE_KEYRING_TO_MK_RE_ENCRYPTION
> eng "Tablespace %s is encrypted by encryption threads and cannot be ALTERed to Master Key encryption. In order to encrypt this tablespace with Master Key encryption - first decrypt this tablespace with encryption threads."
> 
> create table that is encrypted by background threads.
> SET GLOBAL default_table_encryption=ON;
> SET GLOBAL innodb_encryption_threads =0;
> ALTER TABLE t3 ENCRYPTION='Y';
> 
> t3 should be disallowed to converted to MK. But the above ER_ error is NOT thrown. Please check
> 
> @satya-bodapati the error is for tablespaces, not tables. Those can be re-encrypted, as they by re-encrypted by foreground thread and copy algorithm. if t3 was a tablespace you would have gotten the error.

@robgolebiowski You missed the point that this tablespace is encrypted by background encryption threads.. Then it shouldn't be allowed to  convert to MK

@satya this is not a tablespace, but a table and tables can converted to MK encryption directly from ONLINE_KEYRING encryption.

storage/innobase/handler/ha_innodb.cc
satya-bodapati: I think this is problem specific to "Forcing MariaDB encryption threads with Oracle/PS".  that is, we have two variables to control encryption of same subsystem.

MariaDB says, just innodb_encrypt_tables=ON with encryption threads is enough to encrypt any tablespace. They dont have other options/variables to control encryption specifically for undo and redo(I assume).  But when it is brought here,

we have two options (dte and innodb_undo_log_encrypt/redo_log_encrypt/ssys_tablespace_encrypt) to control subsystems.

With the fine tuning of encryption options available for these subsystems, may be it is better to control them via their respective options. What do you think?

storage/innobase/fil/fil0crypt.cc
satya-bodapati: This seems to be more of optimization? Is there functional issue without this change.

sql/sql_tablespace.cc
```diff
@@ -474,11 +474,20 @@ bool Sql_cmd_create_tablespace::execute(THD *thd) {
...
 
-  if (opt_table_encryption_privilege_check &&
-      encrypt_tablespace != thd->variables.default_table_encryption &&
+  // check if default has been overwrriten. Note that ENCRYPTION='KEYRING' will
+  // be blocked by Innodb for tablespaces.
+  if (opt_table_encryption_privilege_check && m_options->encryption.str &&
+      (((encrypt_type == "Y" || encrypt_type == "y") &&
+        thd->variables.default_table_encryption != DEFAULT_TABLE_ENC_ON) ||
+       ((encrypt_type == "N" || encrypt_type == "n") &&
+        thd->variables.default_table_encryption == DEFAULT_TABLE_ENC_ON)) &&
```
satya-bodapati, line 499: May I know what is this blocking? what SQLs?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15950,42 +16016,67 @@ static int innobase_alter_encrypt_tablespace(handlerton *hton, THD *thd,
       !(oldenc.empty() || Encryption::is_none(oldenc.data()));
   /* If new tablespace definition says it's encrypted */
   bool is_new_encrypted = !Encryption::is_none(newenc.data());
+  fil_space_crypt_t *crypt_data = space->crypt_data;
 
   // It is only possible to mark tablespace to be skipped by encryption threads
   // if it is not already encrypted.
   if (!is_old_encrypted && !is_new_encrypted &&
       alter_info->explicit_encryption) {
```
satya-bodapati, line 16164: old is not encrypted and new is not encrypted but expicit_encryption is set? Can you tell me how? ALTER TABLESPACE ENGINE=INNODB?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15950,42 +16016,67 @@ static int innobase_alter_encrypt_tablespace(handlerton *hton, THD *thd,
...
-    if (!fil_crypt_exclude_tablespace_from_rotation(space)) {
-      my_error(ER_EXCLUDE_ENCRYPTION_THREADS_RUNNING, MYF(0), space->name);
+    if (!fil_crypt_exclude_tablespace_from_rotation_permanently(space)) {
```
satya-bodapati, line 16165: Can you please add testcase to reach this code?

satya-bodapati (changes requested):

storage/innobase/fil/fil0crypt.cc
satya-bodapati: This seems to be used only on general tabelspaces. Can you add assertion that space is General Tablespace

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1116,6 +1166,15 @@ bool fil_crypt_exclude_tablespace_from_rotation(fil_space_t *space) {
   return true;
 }
 
+void fil_crypt_readd_space_to_rotation(space_id_t space_id) {
```
satya-bodapati, line 1176: s/readd/read

suggestion: fil_crypt_read_rotation_status()

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1428,8 +1487,9 @@ static bool fil_crypt_space_needs_rotation(rotate_thread_t *state,
       break;
     }
 
-    /* No need to rotate space if encryption is disabled */
-    if (crypt_data->is_encryption_disabled()) {
+    /* No need to rotate space if encryption is disabled
+     * permanently or temporarily */
+    if (crypt_data->is_encryption_disabled() || space->exclude_from_rotation) {
```
satya-bodapati, line 1499: Like space->crypt_data, this variable (space->exclude_from_rotation) is also inconsistent w.r.t mutexes.

Sometimes, it is read/write under fil_crypt_threads_mutex, sometimes no mutex.. :(

satya-bodapati (changes requested):

storage/innobase/handler/ha_innodb.cc
robgolebiowski: I do not think that fine tuning for all subsystem is needed. Let leave those variables for foreground encryption.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: yes, the functional issue is that we need to know the size of the space in order to know how many pages there are to rotate in a space.

sql/sql_tablespace.cc
robgolebiowski: it is blocking ENCRYPTION clause when it is different than default_table_encryption and table_encryption_privilege_check is true and user do not have TABLE_ENCRYPTION_ADMIN privileges. For instance (user without TABLE_ENCRYPTION_ADMIN):
SET GLOBAL default_table_encryption='Y';
SET GLOBAL  table_encryption_privilege_check=ON;
CREATE  TABLESPACE (...) ENCRYPTION='N'

The same behavior is for d_t_e=N and ENCRYPTION='Y' mismatch.

storage/innobase/handler/ha_innodb.cc
robgolebiowski: For instance tablespace was created without any ENCRYPTION clause - CREATE TABLESPACE ... - then the user decided to explicitly remove this tablespace from encryption threads by specifying ALTER TABLESPACE ENCRYPTION='N'. So before and after is not encrypted, but explicit_encryption is set. Probably the name is misleading explicit_encryption - as it was explicitly encrypted, whilst it means that ENCRYPTION clause was explicitly specified. Do you think we should change the name ? Although I do not know if this would not break the compatibility and we would need to rename it at the upgrade, bit painful ...

storage/innobase/handler/ha_innodb.cc
robgolebiowski: There are couple of ENCRYPTION='N' sets for unencrypted tables/tablespaces. Please have a look at encrypt_threads_skip_tablespace.test, line 121.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: This variable is std::atomic<bool> so no mutex would be required.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: actually, it is readd_space_to_rotation - the space->exclude_from_rotation is assigned to false in this function. I can change the name of it to something else if it is to weird ...

storage/innobase/fil/fil0crypt.cc

sql/sql_tablespace.cc
satya-bodapati: Without this PR change, I get the same error on mismatch.. So am still confused what exactly this is for. 
Btw, I assume you default_table_encryption="ON" when you wrote "Y"

storage/innobase/handler/ha_innodb.cc
satya-bodapati: Ok, lets keep the name. Is there test that covers this change? Also will DD attribute 'explicit_encrytion' be present ?

sql/sql_tablespace.cc
robgolebiowski: The old thd->variables.default_table_encryption was boolean variable, now it is an enum - thus the change. It is not change in behavior, just code was amended to work with enum.

storage/innobase/handler/ha_innodb.cc
robgolebiowski: yes, explicit_encryption in DD should be set when you use explicit clause.

sql/sql_tablespace.cc
satya-bodapati: Thanks for explanation!

satya-bodapati: old code was enum too. "Sys_var_enum_default_table_encryption". Did I miss something?

sql/sys_vars.cc
```diff
@@ -7111,6 +7111,28 @@ static Sys_var_uint Sys_immediate_server_version(
...
+    static const LEX_CSTRING innodb_engine{STRING_WITH_LEN("innodb")};
+
+    bool is_online_enc_disallowed = false;
+
+    plugin_ref plugin;
+    if ((plugin = ha_resolve_by_name(nullptr, &innodb_engine, false))) {
+      handlerton *hton = plugin_data<handlerton *>(plugin);
+      is_online_enc_disallowed =
+          hton->check_mk_and_keyring_encrypt_exclusion_for_online_enc(thd);
+      plugin_unlock(nullptr, plugin);
```
satya-bodapati, line 7176: optional: check_mk_keyring_exclusions()

sql/sql_tablespace.cc
robgolebiowski: When I was doing initial reconciliation I changed it to enum:
-static Sys_var_bool Sys_default_table_encryption(
+static const char *default_table_encryption_names[] = {
+    "OFF", "ON", "KEYRING_ON", "ONLINE_TO_KEYRING",
+    "ONLINE_FROM_KEYRING_TO_UNENCRYPTED"};
Commit#: (4c0689594de6970102036deabd9f4b4beb837630) PS-5772: 8.0.16 encryption reconciliation.

sql/sys_vars.cc
robgolebiowski: done

storage/innobase/include/fil0fil.h
```diff
@@ -288,7 +288,11 @@ struct fil_space_t {
 
   bool is_encrypted;
 
-  bool exclude_from_rotation;
+  // Used by encryption threads to check whether a space was
+  // excluded from encryption/decryption. We use atomic because
+  // this variable can be check by encryption threads without
+  // them acquiring any locks.
+  std::atomic<bool> exclude_from_rotation;
```
satya-bodapati, line 318: I feel encryption threads in 8.0 should move to MDLs and get rid of variables/ custom mutexes..

robgolebiowski: That would be 6 months more work ? ...

satya-bodapati (changes requested):
Please submit the PR to jenkins. (Close to approval)

storage/innobase/include/fil0fil.h
satya-bodapati: May be :).. what I observed is Two threads are trying to synchronize on tablespace/table operation. Using variables imho is not the proper way.

User threads already synchronize by using MDLs. InnoDB background threads in 8.0 also moved to MDLs. Purge uses MDLs now. Encryption threads should follow the suite.. Not now.. sometime in future..

PR #3673, opened 2020-02-27
PS-6884 data masking page needed updates

patrickbirch:
fixed examples to correct errors
added installation command


tplavcic (approved):
LGTM

PR #3675, opened 2020-02-27
PS-6860 : Innodb_buffer_pool_pages_LRU_flushed doesn't work

george-lorch:
- Simple fix to merge buf_lru_flush_page_count from individual pool instances
  into total stats.

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -495,6 +495,9 @@ buf_get_total_stat(
 
 		tot_stat->n_pages_not_made_young +=
 			buf_stat->n_pages_not_made_young;
+
+		tot_stat->buf_lru_flush_page_count +=
+		    buf_stat->buf_lru_flush_page_count;
 	}
 }
 


george-lorch:
https://ps57.cd.percona.com/job/percona-server-5.7-param/235/

george-lorch:
@percona-ysorokin I know you are busy with a great many things these days, but, when you get a minute, can you take a quick look at this?  Shouldn't take too long.

storage/innobase/buf/buf0buf.cc
```diff
@@ -495,6 +495,9 @@ buf_get_total_stat(
 
 		tot_stat->n_pages_not_made_young +=
 			buf_stat->n_pages_not_made_young;
+
+		tot_stat->buf_lru_flush_page_count +=
+		    buf_stat->buf_lru_flush_page_count;
```
percona-ysorokin, line 500: What about other 2 remaining data members
```
	ulint	LRU_bytes;	/*!< LRU size in bytes. Protected by
				LRU_list_mutex. */
	ulint	flush_list_bytes;/*!< flush_list size in bytes.
				Protected by flush_list_mutex */
```
https://github.com/percona/percona-server/blob/eac6473879d484f6df72240928d964dc562c8842/storage/innobase/include/buf0buf.h#L2089
and
https://github.com/percona/percona-server/blob/eac6473879d484f6df72240928d964dc562c8842/storage/innobase/include/buf0buf.h#L2091

storage/innobase/buf/buf0buf.cc
```diff
@@ -495,6 +495,9 @@ buf_get_total_stat(
 
 		tot_stat->n_pages_not_made_young +=
 			buf_stat->n_pages_not_made_young;
+
```
percona-ysorokin, line 498: Should this change also be propagated to 8.0?

storage/innobase/buf/buf0buf.cc
george-lorch: Thanks or looking @percona-ysorokin.  I thought the same thing, but it seems that these are not actually exposed in SHOW STATUS, so I left them as they are.

storage/innobase/buf/buf0buf.cc
george-lorch: Yes, but I wanted 5.7 ready and reviewed first before making and testing 8.0 branch.  I also needed a branch prepared for a custom build.

percona-ysorokin (approved):

PR #3676, opened 2020-02-27
PS-6849 Add JUnit reporting support to MTR (5.6)

i-rinat:
* A timeout for the unit tests as a whole.


i-rinat:
Found that final Valgrind reports which describe memory leaks were only printed to the console. Added a patch that collects that data and puts it to the JUnit reports.

i-rinat:
Patchset was reworked.

* Sample testcases in junit suite now include variations that start master-slave pair. Were needed to make sure memory leaks were tracked in all processes.
* If MTR is started with `--valgrind` or `--valgrind-mysqld` options, CTest is invoked with `-T memcheck` which uses Valgrind's memcheck tool for checking.
* JUnit reports now include test's text not in `<failure>` tag but in a separate tag named `<system-out>`. That makes it possible to have output of a non-failing test case, whic can't have `<failure>` tag.
* CTest output is always included into a JUnit report. Turns out, Valgrind warnings are not fatal errors for CTest, so the only way to pass the warning existence information is to always include output.
* If one or more test cases cause `mysqld` to hang or crash during shutdown procedure, an artificial test named `clean_shutdown` is added to the report. A list of test cases that were executed on a `mysqld` instance, and excerpts of error logs are printed to the standard output and added to the JUnit report.

i-rinat:
`SET(BUILD_TESTING OFF)` in the root `CMakeLists.txt` is now called unconditionally to keep `BUILD_TESTING` disabled.

Previous patchset version tried to detect whenever `BUILD_TESTING` was set in `cmake` parameters to mimic existing behavior. But that may cause subsequent `CMake` runs to fail which is extremely inconvenient.

PR #3677, opened 2020-02-28
PS-3390 LP #1553705: document the options available for non-interactive installation in Debian

patrickbirch:
added section on non-interactive installation


tplavcic (approved):
LGTM

PR #3684, opened 2020-03-06
Fixed PS-6899 (main.events_bugs and main.events_1 fail because 2020-01-01 is considered future) (5.6)

percona-ysorokin:
https://jira.percona.com/browse/PS-6899

Fixed and re-recorded 'main.events_1' and 'main.events_bugs' MTR test cases
because 2020 is not a timestamp in future any more.


percona-ysorokin:
Trivial fix - self-approving.

PR #3686, opened 2020-03-06
Fixed PS-6899 (main.events_bugs and main.events_1 fail because 2020-01-01 is considered future) (8.0)

percona-ysorokin:


percona-ysorokin:
Trivial fix - self-approving.

PR #3697, opened 2020-03-11
PS-6811: Crash in ACL_PROXY_USER::check_validity with --skip-name-resolve=1 (8.0) 

inikep:
This patch fixes the issue when `host.get_host()` returns `nullptr`.
It also expands `main.skip_name_resolve` with additional tests.

The bug was introduced with https://github.com/mysql/mysql-server/commit/93b813a3ca44


percona-ysorokin (approved):
LGTM

inikep:
https://ps80.cd.percona.com/job/percona-server-8.0-param/535/

PR #3701, opened 2020-03-12
PS-6913 Update data scrubbing description

patrickbirch:
modified file: data-scrubbing
rewrote description.


patrickbirch:
Data scrubbing is changed in 8.0.20-11.

PR #3705, opened 2020-03-13
PS-5938 8.0.19 post-merge fix: re-recorded rocksdb.add_index_inplace_sstfilewriter

percona-ysorokin:
https://jira.percona.com/browse/PS-5938

Re-recorded 'rocksdb.add_index_inplace_sstfilewriter' MTR test case because of the
removed integer width specifiers.


percona-ysorokin:
MTR changes only - self-approving.

PR #3706, opened 2020-03-13
PS-6907: Create temporary table like crashes with expand fast index c…

dutow:
…reation

Issue: There is no src table set in this scenario, which resulted in
dereferencing a null pointer

Fix: check for null pointers


percona-ysorokin (approved):
LGTM

PR #3714, opened 2020-03-18
PS-6922: Running ALTER TABLE with with expand_fast_index_creation

robgolebiowski:
enabled can lead to a crash

Statement like:
SET expand_fast_index_creation=1;
alter table mysql.db order by db desc;
Would cause server to crash. This is because ALTER TABLE would rebuild
mysql tablespace. This ALTER would rebuild mysql TABLESPACE because
table db was created with statement CREATE TABLE db TABLESPACE=mysql
ENCRYPTION='N'. So it has assigned explicit_encryption=true.
We always rebuild tablespace for explicit_encryption=true and ENCRYPTION
clause = 'N' - in order to add information to page0 - so the encryption
threads would skip the space. However, we did this also for tables in
general tablespace and we should rebuild only in case of file_per_table
tablespaces. This is now fixed.

### DIFF
--- a/storage/innobase/handler/handler0alter.cc
+++ b/storage/innobase/handler/handler0alter.cc
@@ -654,9 +654,19 @@ static MY_ATTRIBUTE((warn_unused_result)) bool innobase_need_rebuild(
   Alter_inplace_info::HA_ALTER_FLAGS alter_inplace_flags =
       ha_alter_info->handler_flags & ~(INNOBASE_INPLACE_IGNORE);
 
-  if (Encryption::none_explicitly_specified(
-          ha_alter_info->create_info->explicit_encryption,
-          ha_alter_info->create_info->encrypt_type.str) ||
+  const bool was_none_explicitly_specified{
+      Encryption::none_explicitly_specified(old_table->s->explicit_encryption,
+                                            old_table->s->encrypt_type.str)};
+
+  const bool is_none_explicitly_specified{Encryption::none_explicitly_specified(
+      ha_alter_info->create_info->explicit_encryption,
+      ha_alter_info->create_info->encrypt_type.str)};
+
+  const bool is_file_per_table{
+      !tablespace_is_shared_space(ha_alter_info->create_info)};
+
+  if ((!was_none_explicitly_specified && is_none_explicitly_specified &&
+       is_file_per_table) ||
       (Encryption::is_keyring(ha_alter_info->create_info->encrypt_type.str) &&
        !Encryption::is_keyring(old_table->s->encrypt_type.str)) ||
       ha_alter_info->create_info->encryption_key_id !=
--- a/mysql-test/r/percona_expand_fast_index_creation.result
+++ b/mysql-test/r/percona_expand_fast_index_creation.result
@@ -153,4 +153,36 @@ DROP TABLE t1;
 # PS-6903: CREATE TEMPORARY TABLE crashes with expand_fast index creation
 #
 CREATE TEMPORARY TABLE t1 LIKE information_schema.processlist;
+#
+# PS-6922: Running ALTER TABLE with with expand_fast_index_creation enabled can lead to a crash
+#
+# Below queries would crash the server. Check that they work OK now.
+ALTER TABLE mysql.db ORDER BY db DESC;
+Warnings:
+Warning	1105	ORDER BY ignored as there is a user-defined clustered index in the table 'db'
+1
+ALTER TABLE mysql.tables_priv MODIFY User CHAR DEFAULT'',MODIFY gRANTOR CHAR DEFAULT'';
+ERROR 01000: Data truncated for column 'User' at row 1
+CREATE TABLE t_part
+(a int,
+b int,
+c varchar(64),
+KEY (a),
+KEY (b),
+KEY (c,b))
+ENGINE = InnoDB
+ENCRYPTION='N'
+PARTITION BY HASH (a) PARTITIONS 5;
+ALTER TABLE t_part ORDER BY a DESC;
+CREATE TABLE t2_part
+(a int,
+b int,
+c varchar(64),
+KEY (a),
+KEY (b),
+KEY (c,b))
+ENGINE = InnoDB
+PARTITION BY HASH (a) PARTITIONS 5;
+ALTER TABLE t2_part ENCRYPTION='N', ORDER BY a DESC;
+DROP TABLE t_part,t2_part;
 SET SESSION expand_fast_index_creation=OFF;
--- a/mysql-test/t/percona_expand_fast_index_creation.test
+++ b/mysql-test/t/percona_expand_fast_index_creation.test
@@ -156,6 +156,43 @@ DROP TABLE t1;
 --echo #
 CREATE TEMPORARY TABLE t1 LIKE information_schema.processlist;
 
+--echo #
+--echo # PS-6922: Running ALTER TABLE with with expand_fast_index_creation enabled can lead to a crash
+--echo #
+--echo # Below queries would crash the server. Check that they work OK now.
+
+ALTER TABLE mysql.db ORDER BY db DESC;
+--echo 1
+--error 1265 #WARN_DATA_TRUNCATED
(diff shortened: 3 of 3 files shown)


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/543/

mysql-test/suite/encryption/r/bug_ps_6922.result
```diff
@@ -0,0 +1,8 @@
+# Below queries would crash the server. Check that they work OK now.
```
dutow: Please add this to `mysql-test/t/percona_expand_fast_index_creation.test` instead.

storage/innobase/handler/handler0alter.cc
```diff
@@ -654,9 +654,13 @@ static MY_ATTRIBUTE((warn_unused_result)) bool innobase_need_rebuild(
   Alter_inplace_info::HA_ALTER_FLAGS alter_inplace_flags =
       ha_alter_info->handler_flags & ~(INNOBASE_INPLACE_IGNORE);
 
-  if (Encryption::none_explicitly_specified(
-          ha_alter_info->create_info->explicit_encryption,
-          ha_alter_info->create_info->encrypt_type.str) ||
+  if ((Encryption::none_explicitly_specified(
```
dutow: Please also consider using two helper `const bool` variables to make this if more readable / understandable.

storage/innobase/handler/handler0alter.cc
```diff
@@ -654,9 +654,13 @@ static MY_ATTRIBUTE((warn_unused_result)) bool innobase_need_rebuild(
...
+           ha_alter_info->create_info->explicit_encryption,
+           ha_alter_info->create_info->encrypt_type.str) &&
+       ((old_table->s->tablespace == nullptr && !old_table->s->system) ||
+        (old_table->s->tablespace != nullptr &&
+         strcmp(old_table->s->tablespace, dict_sys_t::s_file_per_table_name) ==
```
satya-bodapati: Please find a simpler way. This is error prone. Also please add a comment why this code says redbuild required when encryption='n'.

mysql-test/suite/encryption/r/bug_ps_6922.result
```diff
@@ -0,0 +1,8 @@
...
+SET expand_fast_index_creation=1;
+ALTER TABLE mysql.db ORDER BY db DESC;
```
satya-bodapati: Do we really need expand_fast_index_creation for this bug to appear?
And the table must be in mysql tablespace? Wouldn't a rebuild of table in General tablespace throw the same assert/error?

satya-bodapati (changes requested):

mysql-test/suite/encryption/r/bug_ps_6922.result
dutow: Yes - the crash happens when expand fast index creation readds the indexes.
And yes - we spent a lot of time trying to reproduce this with custom tables, but without luck. Yura found a possible explanation for this, see on slack.

storage/innobase/handler/handler0alter.cc
satya-bodapati: Try this : dd::is_tablespace_encrypted() 
THis is available everywhere.. if not just include relevant header.

satya-bodapati: The name is misleading. It has out parameter which tells if tablespace is general or not.

mysql-test/suite/encryption/r/bug_ps_6922.result
satya-bodapati: Good suggestion to reuse existing feature based test file :)

mysql-test/suite/encryption/r/bug_ps_6922.result
satya-bodapati: but a normal table with General tablesapce is not tried afaiu..

dutow: I tried it, it didn't crash.

satya-bodapati: Please add a test for partition tables too.

storage/innobase/handler/handler0alter.cc
robgolebiowski: done

mysql-test/suite/encryption/r/bug_ps_6922.result
robgolebiowski: done

storage/innobase/handler/handler0alter.cc
robgolebiowski: dd::is_tablespace_encrypted works with dd::Table, I have Table here. I have reused tablespace_is_shared_space.

mysql-test/suite/encryption/r/bug_ps_6922.result
robgolebiowski: added

storage/innobase/handler/handler0alter.cc
```diff
@@ -654,9 +654,20 @@ static MY_ATTRIBUTE((warn_unused_result)) bool innobase_need_rebuild(
...
+      Encryption::none_explicitly_specified(old_table->s->explicit_encryption,
+                                            old_table->s->encrypt_type.str)};
+
+  const bool is_none_explicitly_specified{Encryption::none_explicitly_specified(
+      ha_alter_info->create_info->explicit_encryption,
+      ha_alter_info->create_info->encrypt_type.str)};
+
+  const bool is_file_per_table{
+      !tablespace_is_shared_space(ha_alter_info->create_info) &&
+      !old_table->s->system};
```
satya-bodapati: Looks much better. Why do we need old_table->s->system check?

satya-bodapati (changes requested):

storage/innobase/handler/handler0alter.cc
robgolebiowski: I have removed it, as instructed on slack:
@robert.golebiowski patch looks good. Just one doubt on old_table->s->system. AFAIU this is TABLE_SHARE.. why do we need it?
6:35
I dont see why it is relavant here.. it is not related to real system tabelspace or anything
6:35
afaiu, it is meant for optimizer
6:35
for one-row tables.. which are also called 'system'..
6:35
// Mark 'system' tables (tables with one row) to help the Optimizer.
  share->system =
      ((share->max_rows == 1) && (share->min_rows == 1) && (share->keys == 0));

Thanks for checking this.

storage/innobase/handler/handler0alter.cc
```diff
@@ -654,9 +654,19 @@ static MY_ATTRIBUTE((warn_unused_result)) bool innobase_need_rebuild(
...
+                                            old_table->s->encrypt_type.str)};
+
+  const bool is_none_explicitly_specified{Encryption::none_explicitly_specified(
+      ha_alter_info->create_info->explicit_encryption,
+      ha_alter_info->create_info->encrypt_type.str)};
+
+  const bool is_file_per_table{
+      !tablespace_is_shared_space(ha_alter_info->create_info)};
+
+  if ((!was_none_explicitly_specified && is_none_explicitly_specified &&
```
satya-bodapati, line 668: Why do you need the was_none check?

satya-bodapati (changes requested):

storage/innobase/handler/handler0alter.cc
satya-bodapati: question 2: is there similar problem with ENCRYPTION=KEYRING? The OR part doesn't check the file_per_table?

robgolebiowski: q1. Apparently when I do a rebuild without any change it can also cause a crash (discovered it when testing partition table) - so I make sure rebuild is needed.
q2. I do not see a problem with ENCRYPTION = KEYRING, apart from the fact that table with ENCRYPTION='KEYRING' can be created in MK encrypted tablespace. It is possible that Align KEYRING and MK encryption PR is already fixing this. I will create a bug for it, so it would not be forgotten.

satya-bodapati (approved):

PR #3715, opened 2020-03-18
PS-5737 fix mtr tests from Percona-Server-test_* pkg

vorsel:


build-ps/percona-server.spec
```diff
@@ -574,6 +580,8 @@ fi
   fi
 %endif
 
+ln -s %{_datadir}/percona-server %{_datadir}/mysql
```
EvgeniyPatlan: this should be done only in case if %{_datadir}/mysql directory doesn't exist

EvgeniyPatlan (changes requested):

EvgeniyPatlan (approved):

PR #3720, opened 2020-03-20
PS-5252 Add compression differences to MyRocks vs Facebook Difference…

patrickbirch:
…s doc.

Modified file: Differences
Added compression section


inikep (approved):
LGTM

PR #3730, opened 2020-03-24
PS-6945 : tokubackup is broken since it no longer exports the correct…

george-lorch:
… large file API

PS-6946 : Sanitizers find memory use after free and other issues in tokubackup

PS-6945 fixes:
Tokubackup is broken because the large file contract expected by applications that
use tokubackup is no longer exported by tokubackup. The contract was broken by commit
0ab06b5 made in early 2018.

The large file contract is supported by compiling with -D_LARGEFILE64_SOURCE and
-D_FILE_OFFSET_BITS=64 AND including the appropriate include files. When an
application is compiled with large file support, it has link time dependencies
to the large file API. For example, the compiler causes calls to pwrite
to refer to pwrite64.

The hot backup library MUST export pwrite64. The broken commit removes the include
of unistd.h, which removes the large file link time trickery. The result is that
the hot backup library exports pwrite instead of pwrite64.

The effect is that pwrite calls made by the application are NOT intercepted by tokubackup.
Therefore, the backup files are no longer consistent with each other when the backup
completes.

This pull request:
Fixes the compilation of the tests.
Adds tests for lseek and pwrite during the backup.
Reverts the commit that broke the large file API contract.

PS-6946 fixes:
The address and thread sanitizers found some issues in the tokubackup software
when running the tokubackup tests.

This pull request:

Fixes trivial data races in the tests found by the thread sanitizer. This
allows one to focus on important issues.
Fixes a heap use after free problem detected by the thread sanitizer. The
solution replaces the use of std vector with std deque to match the producer
consumer pattern used by tokubackup (and eliminate the problem).
Fixes a concurrent open memory and file descriptor leak caused by missing
locking in the manager open function.


george-lorch:
https://ps56.cd.percona.com/job/percona-server-5.6-param/89/

george-lorch:
@percona-ysorokin may I have a quick superficial review (or as deep as you would like:) )

percona-ysorokin (approved):
LGTM

PR #3732, opened 2020-03-25
PS-6951 Documented new RocksDB variables in fb-prod201907.

patrickbirch:
modified file: myrocks/variables
added disabled to rocksdb_read_free_rpl_tables
added rocksdb_delete_cf
added rocksdb_enable_iterate_bounds
added rocksdb_enable_remove_orphaned_dropped_cfs


doc/source/myrocks/variables.rst
```diff
@@ -1964,6 +2020,9 @@ that should use read-free replication on the slave
 (that is, replication without row lookups).
 Empty by default.
 
+This variable is disabled in |Percona Server| 5.7.30-33. We recommend
```
inikep, line 2151: It seems that this variable was disabled from the beginning. Please ask George about it.

patrickbirch: checking with George.

george-lorch: I believe it was in use, perhaps not correctly, but code paths that use it exist, so let's leave this as-is.

inikep (approved):

PR #3733, opened 2020-03-25
PS-5737 fix mtr tests from Percona-Server-test_* pkg

vorsel:


build-ps/percona-server.spec
```diff
@@ -986,6 +986,12 @@ if [ $? -eq 1 -a "$SERVER_TO_START" = "true" ] ; then
 fi
 %endif
 
+if [ ! -d %{_datadir}/mysql ]; then
+    pushd %{_datadir}
+    ln -s percona-server mysql
+    popd
+fi
```
EvgeniyPatlan, line 993: please add condition on preun to remove link if we are removing package.

EvgeniyPatlan (changes requested):

build-ps/percona-server.spec
vorsel: done

EvgeniyPatlan (changes requested):
remove link only on package removal

EvgeniyPatlan (approved):

PR #3742, opened 2020-03-27
Ps 5.7 6945 6946

george-lorch:


george-lorch:
Due to PS-6954 ( collision of -std=c++11 and -std=gnu++03) this will fail jenkins as-is, so, here is a jenkins job simulating this change merged to trunk on top of the fix for PS-6954 https://ps57.cd.percona.com/job/percona-server-5.7-param/256/

george-lorch:
5.6 was reviewed and tested, 5.7 tested, self approving.

PR #3745, opened 2020-03-28
PS-6948 : Advance PerconaFT git submodule commit pointer

george-lorch:
Ppick up escalation work and changes to allow building on Ubuntu 20.04

PS-5932: disable cleaner thread and assert on dirty pages

PS-5932: Fix double initialization in ft tools

PS-5674: Fixed compilation error caused by #warning in sys/sysctrl.h


george-lorch:
https://ps56.cd.percona.com/job/percona-server-5.6-param/92/
self approving

PR #3760, opened 2020-04-03
Implement PS-5764 (Introduce SEQUENCE_TABLE() table-level SQL function)

percona-ysorokin:
https://jira.percona.com/browse/PS-5764

Implemented 'SEQUENCE_TABLE(<expr>)' table function that is supposed to be
used for generating sequences of integers. For instance

'SELECT * FROM SEQUENCE_TABLE(3) AS tt'
should return
| value |
|-----|
| 0     |
| 1     |
| 2     |

Please notice that as for other table functions (like 'JSON_TABLE()') table
alias ('AS tt' in the example) is not optional.

The result set will include a single column named 'value' that will be defined
as 'BIGINT UNSIGNED NOT NULL'. In other words, the following two statements
'CREATE TABLE t1 AS SELECT * FROM SEQUENCE_TABLE(...) AS tt;'
'SHOW CREATE TABLE t1;'
will result in
"Table Create Table"
"t1 CREATE TABLE `t1` ("
"  `value` bigint unsigned NOT NULL DEFAULT '0'"
") ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci".

Introduced new global dynamic variable 'tf_sequence_table_max_upper_bound'
that determines max value of how many records 'SEQUENCE_TABLE()' table
function is allowed to generate. Default - 1048576, min - 1024,
max - 2^64 - 1.

The max value of the expression passed to 'SEQUENCE_TABLE()' should not
exceed '@@tf_sequence_table_max_upper_bound'. Otherwise,
'ER_SEQUENCE_TABLE_SIZE_LIMIT' error will be reported.

Similarly to 'Table_function_json' class that is used for 'JSON_TABLE()',
introduced a new 'Table_function_sequence' class derived from 'Table_function'
that is supposed to fill virtual result table with a zero-based sequence of
integers.

Depending on the type of the '<expr>' passed to 'SEQUENCE_TABLE(<expr>)' there
are three possible cases of its const-ness:

1. Always const (e.g. a numeric literal or an expression involving only
   numeric literals), '<expr>' will be evaluated only once inside
   'Table_function_sequence::do_init_args()'.
2. Non-const during init(), but becomes const after it (e.g a field from a
   const table), '<expr>' will be evaluated only the first time
   'Table_function_sequence::fill_result_table()' is called. In every
   subsequent call of this function a precalculated value from the very first
   call will be re-used.
3. Non-const (e.g. a table field) '<expr>' will be evaluated every time
   'Table_function_sequence::fill_result_table()' is called.

Similarly to 'PT_table_factor_function' class that is used for 'JSON_TABLE()'
introduced a new 'PT_table_sequence_function' class derived from
'PT_table_reference' that is supposed to add a new virtual joined table into
'Parse_context' whose functionality implemented in 'Table_function_json'.

MySQL grammar extended with a alternative for 'table_function' - now there are
two 'JSON_TABLE_SYM' ... and 'SEQUENCE_TABLE_SYM' ...

Added 'main.percona_sequence_table' MTR test case that checks
'SEQUENCE_TABLE()' functionality in various SQL contexts:
- Checking invalid SEQUENCE_TABLE() arguments
- Checking SEQUENCE_TABLE() without table alias
- Simple usage scenarios
- Checking SEQUENCE_TABLE() max records limit
- Checking SEQUENCE_TABLE() upper bound value conversions (NULLs, negatives, floating points, strings)
- Checking SEQUENCE_TABLE() upper bound expressions (session variables, simple arithmetic)
- Checking SEQUENCE_TABLE() with WHERE clause (basic point selection, ranges, every second)
- Selecting custom value expressions from SEQUENCE_TABLE()
- Checking VIEWs based on SEQUENCE_TABLE()
- Checking stored procedures / functions
- Checking prepared statements
- Checking SHOW CREATE TABLE
- Checking table joins
- Checking for various forbidden upper_bound constructs

Added 'sys_vars.tf_sequence_table_max_upper_bound_basic' MTR test case that
checks basic variable functionality.


percona-ysorokin:
Jenkins build link
https://ps80.cd.percona.com/job/percona-server-8.0-param/555/

inikep (approved):
LGTM

percona-ysorokin:
Updated Jenkins build link
https://ps80.cd.percona.com/job/percona-server-8.0-param/574/

inikep (approved):
LGTM

PR #3761, opened 2020-04-06
PS-6953: Fail to load specific utf8 data (5.6)

inikep:
Reference patch: https://github.com/mysql/mysql-server/commit/ac64a647e95cc87cb8228e6086d2c0fa8c7589f9

Description:
LOAD DATA fails to accept the multibyte character which follows escape mark,
but INSERT can.

Fix:
Change READ_INFO::read_field  to ignore the escape mark if it is followed by
mulitibyte character.


inikep:
Tested at https://ps56.cd.percona.com/job/percona-server-5.6-param/93/

george-lorch (approved):

PR #3766, opened 2020-04-08
5.7 PS-6979 - SHOW PROCESSLIST shows CREATE USER all the time till next w…

altmannmarcelo:
…riteset

Problem
If a query had been rewritten by the parser because it contains
sensitive information ( mysql_rewrite_query ), it wont be cleaned up
when sql thread applies it, making it get stuck on SHOW PROCESSLIST and
performance_schema.threads.

Solution
Call proper cleanup functions as part of
Query_log_event::do_apply_event end block.

### DIFF
--- a/mysql-test/suite/rpl/t/percona_bug_ps6979_showprocesslist.test
+++ b/mysql-test/suite/rpl/t/percona_bug_ps6979_showprocesslist.test
@@ -1,0 +1,22 @@
+--source include/master-slave.inc
+
+CREATE USER `bugddl1`@`localhost` IDENTIFIED BY 'S0meRand0mPWD!';
+
+# performance_schema.threads requires 2 DDL's to manifest the issue.
+CREATE USER `bugddl2`@`localhost` IDENTIFIED BY 'S0meRand0mPWD!';
+
+--source include/sync_slave_sql_with_master.inc
+
+let $wait_timeout= 10;
+let $wait_condition= SELECT COUNT(*)=0 FROM performance_schema.threads WHERE PROCESSLIST_INFO LIKE "CREATE USER%";
+--source include/wait_condition.inc
+
+--let $assert_text= Check that CREATE USER statement has been removed from IS.PROCESSLIST.
+--let $assert_cond= [SELECT COUNT(*) FROM INFORMATION_SCHEMA.PROCESSLIST WHERE INFO LIKE "CREATE USER%"] = 0
+--source include/assert.inc
+
+# Cleanup
+--source include/rpl_connection_master.inc
+DROP USER `bugddl1`@`localhost`;
+DROP USER `bugddl2`@`localhost`;
+--source include/rpl_end.inc
--- a/sql/log_event.cc
+++ b/sql/log_event.cc
@@ -5217,6 +5217,13 @@ end:
   thd->m_digest= NULL;
 
   /*
+    Prevent rewritten query from getting "stuck" in SHOW PROCESSLIST,
+    and performance_schema.threads.
+  */
+  thd->rewritten_query.mem_free();
+  thd->reset_query_for_display();
+
+  /*
     As a disk space optimization, future masters will not log an event for
     LAST_INSERT_ID() if that function returned 0 (and thus they will be able
     to replace the THD::stmt_depends_on_first_successful_insert_id_in_prev_stmt
--- a/mysql-test/suite/rpl/r/percona_bug_ps6979_showprocesslist.result
+++ b/mysql-test/suite/rpl/r/percona_bug_ps6979_showprocesslist.result
@@ -1,0 +1,13 @@
+include/master-slave.inc
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
+Note	####	Storing MySQL user name or password information in the master info repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START SLAVE; see the 'START SLAVE Syntax' in the MySQL Manual for more information.
+[connection master]
+CREATE USER `bugddl1`@`localhost` IDENTIFIED BY 'S0meRand0mPWD!';
+CREATE USER `bugddl2`@`localhost` IDENTIFIED BY 'S0meRand0mPWD!';
+include/sync_slave_sql_with_master.inc
+include/assert.inc [Check that CREATE USER statement has been removed from IS.PROCESSLIST.]
+[connection master]
+DROP USER `bugddl1`@`localhost`;
+DROP USER `bugddl2`@`localhost`;
+include/rpl_end.inc


altmannmarcelo:
https://ps57.cd.percona.com/job/percona-server-5.7-param/261/

mysql-test/suite/rpl/t/percona_bug_ps6979_showprocesslist.test
```diff
@@ -0,0 +1,16 @@
+--source include/master-slave.inc
+
+connection master;
+create database bug111;
```
satya-bodapati: is this statement required?

satya-bodapati (commented):
Patch looks good

mysql-test/suite/rpl/t/percona_bug_ps6979_showprocesslist.test
altmannmarcelo: The issue on ps.threads only happens from the second DDL forward.

altmannmarcelo:
@satya-bodapati I've adjusted the test case to test both situations 1) SHOW PROCESSLIST and 2) ps.threads.

Ran another jenkis build to execute only this particular test - https://ps57.cd.percona.com/job/percona-server-5.7-param/263/

sql/log_event.cc
```diff
@@ -5216,6 +5216,13 @@ int Query_log_event::do_apply_event(Relay_log_info const *rli,
   thd->m_statement_psi= NULL;
   thd->m_digest= NULL;
 
+  /*
+    Prevent rewritten query from getting "stuck" in SHOW PROCESSLIST,
+    and performance_schema.threads.
+  */
+  thd->rewritten_query.mem_free();
```
satya-bodapati, line 5223: should you chcek if rewritten_query is used first?
```suggestion
 if (thd->rewritten_query.length())
   thd->rewritten_query.mem_free();
 ```

File: sql/log_event.cc [hunk 3]
```diff
@@ -5216,6 +5216,13 @@ int Query_log_event::do_apply_event(Relay_log_info const *rli,
...
+  thd->reset_query_for_display();
```
satya-bodapati, line 5224: On master, when create user is executed, how is it freed? and is reset_query_for_display() is called?

satya-bodapati (changes requested):

mysql-test/suite/rpl/t/percona_bug_ps6979_showprocesslist.test
satya-bodapati: Okay

sql/log_event.cc
altmannmarcelo: On master size, we always free it - https://github.com/percona/percona-server/blob/8.0/sql/sql_parse.cc#L2262

sql/log_event.cc
altmannmarcelo: it's free unconditionally via `thd->rewritten_query.mem_free();` .
Regarding `thd->reset_query_for_display();` this is currently not been called, I've raised a bug which has been confirmed - https://bugs.mysql.com/bug.php?id=99204

mysql-test/suite/rpl/t/percona_bug_ps6979_showprocesslist.test
```diff
@@ -0,0 +1,21 @@
+--source include/master-slave.inc
+
+connection master;
```
venkatesh-prasad-v: Even though I'm not an official reviewer, here are my two cents.

1. You may remove line no 3 as `master-slave.inc` by default leaves the connection to master.
2. Consider using `rpl_connection_master.inc` and `rpl_connection_slave.inc` instead of just `connection` mysqltest command. Using `rpl_connection_master.inc` will print `[connection master]` into the result file and makes the result file more readable.

satya-bodapati (changes requested):
With afcaac4d6f8a, thd->reset_query_for_display() is already called in log applier threads. See
Query_log_event::do_apply_event
Rows_query_log_event::do_apply_event
functions in the commit.

Can you check why this extra reset() is necessary?

satya-bodapati:
Also please address comment from  venkatesh

sql/log_event.cc
satya-bodapati: Thanks for confirming

altmannmarcelo:
> With [afcaac4](https://github.com/percona/percona-server/commit/afcaac4d6f8aecc70c418138f1d8dd4e4965e629), thd->reset_query_for_display() is already called in log applier threads. See
> Query_log_event::do_apply_event
> Rows_query_log_event::do_apply_event
> functions in the commit.
> 
> Can you check why this extra reset() is necessary?

@satya-bodapati  The parser will set the query again https://github.com/altmannmarcelo/percona-server/blob/PS-5.7-6979/sql/sql_parse.cc#L5849 and `mysql_parse` is called after that `reset_query_for_display()` https://github.com/altmannmarcelo/percona-server/blob/PS-5.7-6979/sql/log_event.cc#L4923

@venkatesh-prasad-v - Adjusted the test cases as per your comments

satya-bodapati:
> > With [afcaac4](https://github.com/percona/percona-server/commit/afcaac4d6f8aecc70c418138f1d8dd4e4965e629), thd->reset_query_for_display() is already called in log applier threads. See
> > Query_log_event::do_apply_event
> > Rows_query_log_event::do_apply_event
> > functions in the commit.
> > Can you check why this extra reset() is necessary?
> 
> @satya-bodapati The parser will set the query again https://github.com/altmannmarcelo/percona-server/blob/PS-5.7-6979/sql/sql_parse.cc#L5849 and `mysql_parse` is called after that `reset_query_for_display()` https://github.com/altmannmarcelo/percona-server/blob/PS-5.7-6979/sql/log_event.cc#L4923

Ok, after the above mysql_parse() and mysql_command_execute(), wouldn't it reach thd->reset_for_display() added in afcaac4 ?

altmannmarcelo:
afcaac4  adds `thd->reset_for_display` in two places (apart from creating the function itself):
https://github.com/percona/percona-server/commit/afcaac4d6f8aecc70c418138f1d8dd4e4965e629#diff-2943dbb03780b0d55c6216ee7227bd62R4993 - In case a PREPARE failed.
https://github.com/percona/percona-server/commit/afcaac4d6f8aecc70c418138f1d8dd4e4965e629#diff-8c78bfc758be72b7df2306f1d9984c47R4693 - At Query_log_event::do_apply_event when it has to rewrite the query in case the charset of the connection has changed. This happens when the statement is been setup to be executed. However, we are missing to clean it up once it has completed the execution. 

Note that a few lines before my fix, we are resetting the `query_arg` and `query_length_arg` variables of the thd ( https://github.com/percona/percona-server/pull/3766/files#diff-8c78bfc758be72b7df2306f1d9984c47R5210 ). But we also need to reset query for display. That is what I'm doing a few lines after, once PS instrumentation has ended https://github.com/percona/percona-server/pull/3766/files#diff-8c78bfc758be72b7df2306f1d9984c47R5224 .

satya-bodapati (approved):

satya-bodapati:
> [afcaac4](https://github.com/percona/percona-server/commit/afcaac4d6f8aecc70c418138f1d8dd4e4965e629) adds `thd->reset_for_display` in two places (apart from creating the function itself):
> [afcaac4#diff-2943dbb03780b0d55c6216ee7227bd62R4993](https://github.com/percona/percona-server/commit/afcaac4d6f8aecc70c418138f1d8dd4e4965e629#diff-2943dbb03780b0d55c6216ee7227bd62R4993) - In case a PREPARE failed.
> [afcaac4#diff-8c78bfc758be72b7df2306f1d9984c47R4693](https://github.com/percona/percona-server/commit/afcaac4d6f8aecc70c418138f1d8dd4e4965e629#diff-8c78bfc758be72b7df2306f1d9984c47R4693) - At Query_log_event::do_apply_event when it has to rewrite the query in case the charset of the connection has changed. This happens when the statement is been setup to be executed. However, we are missing to clean it up once it has completed the execution.
> 
> Note that a few lines before my fix, we are resetting the `query_arg` and `query_length_arg` variables of the thd ( https://github.com/percona/percona-server/pull/3766/files#diff-8c78bfc758be72b7df2306f1d9984c47R5210 ). But we also need to reset query for display. That is what I'm doing a few lines after, once PS instrumentation has ended https://github.com/percona/percona-server/pull/3766/files#diff-8c78bfc758be72b7df2306f1d9984c47R5224 .

Thanks for the explanation!

PR #3770, opened 2020-04-10
PS-5454. Test percona_changed_page_bmp_log_resize is not stable

avbelov23:
https://jira.percona.com/browse/PS-5454

The tests is unstable and sometimes fails with this assertion:
2019-03-04T17:06:21.909284Z 0 [ERROR] [MY-013183] [InnoDB] Assertion failure: log0online.cc:348:copy_start_lsn <= current_lsn + unparsed_size() + LOG_BLOCK_BOUNDARY_LSN_PAD || current_lsn == 0

The current last block of the redo log may contain less data than LOG_BLOCK_DATA_SIZE (maximum data size per block). We need to skip this block and move to the next block. If the parsing buffer does not know about this, then the calculated parsing buffer lsn will continue to point to the current last block. And in further work (for example, after restart mysqld) after the next iteration of reading, the parsing will continue after the end of the data in this block.
We reset the parsing buffer lsn so that it synchronizes with the read buffer lsn and the parsing moves to the next block.

### DIFF
--- a/storage/innobase/log/log0online.cc
+++ b/storage/innobase/log/log0online.cc
@@ -295,6 +295,8 @@ class log_parse_buffer final : public log_buffer<RECV_PARSING_BUF_SIZE> {
     return true;
   }
 #endif
+
+  void reset_lsn() { current_lsn = 0; }
 };
 
 bool log_parse_buffer::advance(ulint delta) noexcept {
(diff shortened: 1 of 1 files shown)


satya-bodapati:
I think skipping the redo bytes on incomplete block is wrong. I will paste the comment from @laurynas-biveinis when he debugged this:

"
I spent time to review all the code and I believe there is a
fundamental issue with the plan.

The last tracked LSN will be equal to checkpoint LSN, hence
potentially falling in a middle of a log record. The way log tracking
works, bitmap data will be written for the last full record <= that
checkpoint LSN (that is, potentially leaving untracked the last few
bytes before checkpoint).

And crash recovery works differently: the checkpoint LSN in a middle
of a log record means that this record change has been flushed
completely, that is, the actual recovery should start after the
checkpoint LSN. So for bitmap, we need to use "before" LSN, and
recovery uses "after" LSN, not good. Thus it seems we cannot copy
crash recovery code or it will leave that one log record on the top of
checkpoint LSN unhandled. Also theoretically anything before the
checkpoint LSN could be rewritten already (the 8.0 log work removed
all the capacity safety margins as unnecessary).

So the right fix is to change bitmap code that it writes bitmap data
for the records after checkpoint LSN during normal operation, which
could be a relatively large effort - if you want, try taking a quick
look at that (grep for PAST_END in log0online.cc) or we can find
another task.
"

So the fix should be to find a way to include the records from incomplete block (at checkpoint_lsn?)

satya-bodapati (changes requested):
see my comment above

PR #3775, opened 2020-04-15
PS-6989: "Cannot write: No space left on device" error with Azure Piplines (8.0)

inikep:
1. Limit ccache cache size to 2 GB
2. Change `Git.ShallowFetchDepth` to 256
3. Remove compiled files before uploading caches


percona-ysorokin (approved):
LGTM

PR #3776, opened 2020-04-15
PS-6925: Mismatch in default socket values for mysqld and mysqld_safe

Sudokamikaze:
As *RPM distributions have different path for socket and mysql-client expects it in `/var/lib/mysql/mysql.sock` -- `[mysqld_safe]` should have same one as well.

### DIFF
--- a/build-ps/rpm/mysqld_safe.cnf
+++ b/build-ps/rpm/mysqld_safe.cnf
@@ -10,5 +10,5 @@
 
 [mysqld_safe]
 pid-file = /var/run/mysqld/mysqld.pid
-socket   = /var/run/mysqld/mysqld.sock
+socket   = /var/lib/mysql/mysql.sock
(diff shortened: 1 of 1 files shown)


EvgeniyPatlan:
As this review requires changes in *.cnf files we need to wait for review from @vadimtk

Sudokamikaze:
Also, there's a different approach if we want to make it have "standard" look: Make a patch which will correct path expectation of socket in mysql-client  and another one for re-locating them to *DEB like locations in .cnf files

EvgeniyPatlan (approved):
:+1:

vorsel (approved):

EvgeniyPatlan:
@vadimtk could you please take a look as  it is change in cnf file

vadimtk (approved):

PR #3777, opened 2020-04-15
PS-5635: Introduce crypt_schema 2 for better error checking in encryption

robgolebiowski:
threads.

This WL implements MySQL way of validating if all needed encryption keys
are present and valid -  for tables encrypted by encryption threads.
crypt_data, stored on page0 of a space was extended with encryption
validation tag. For not encrypted space this tag is just a string:
ENC_VAL_TAG_V1_1. Also crypt_data was extended with max_key_version.
When we rotate from not-encrypted to encrypted space (first encryption
of a space) we execute the following steps (at the start of rotation):
1) max_key_version is set to the latest version of encryption key that
we use to encrypt the space.
2) validation tag is encrypted with max_key_version of encryption key.
3) DD space’s flag online_encryption is set to true.

online_encryption is a new flag added by this WL. We set it to true so
not fully encrypted tables would be also validated on server startup.
After a successful rotation online_encryption flag will be replaced with
encrypted flag, also min_key_version will be set to max_key_version.

On server startup we validate if all needed keys are available (and are
the correct ones) for such table by executing the steps:
1) if encrypted flag is set or online_encryption DD flag is set goto 2;
else return OK
2) fetch max_key_version version of a key_id encryption key from
keyring; else return FAIL
3) decrypt validation tag with key fetched in 2).
4) compare decrypted validation tag with ENC_VAL_TAG_V1_1 – if it
matches return OK; else return FAIL.

When rotation not encrypted => encrypted finishes we remove DD
online_encryption flag and replace it with encryption flag.

The situation is similar when we re-encrypt space. The difference is
that validation tag is encrypted with a range of key versions
[min_key_version; max_key_version]; given min_key_version > 0, i.e. we
are re-encrypting a fully encrypted space. We now will re-encrypt
validation tag with range of key versions [max_key_version + 1;
new_max_key_version] We execute the steps (at the beginning of
re-encryption):
1) we fetch latest version of encryption key from keyring. The latest
version is saved as new_max_key_version.
2) we fetch keys with versions [max_key_version + 1;
new_max_key_version] from keyring.
3) we re-encrypt already encrypted validation tag with key_versions from
max_key_version + 1 till new_max_key_version. i.e. tag is already
encrypted with keys from min_key_version to max_key_version, now we
encrypt again this tag with max_key_version+1, then this re-encrypted
tag gets encrypted with max_key_version+2 and so on, till we reach
version new_max_key_version (tag is also encrypted with
new_max_key_version).
4) we update max_key_version to new_max_key_version and we store the
max_key_version and encrypted validation tag on page0.

When re-encryption is finished we encrypt validation tag plaintext
(ENC_VAL_TAG_V1_1) with only max_key_version and save it on page0. We
encrypt with only max_key_version, because this is the only key that we
are currently using to encrypt this table (all pages have been
re-encrypted with this key when rotation finishes).

On server startup we validate if all needed keys are available (and are
the correct ones) for such table by executing the steps:

1) if encrypted flag is set we goto step 2 ; else OK;
2) we fetch all keys versions from [min_key_version; max_key_version] if
any of the keys is missing – then FAIL.
3) we decrypt encrypted validation_tag with fetched keys version –
starting from max_key_version and going down to min_key_version.
4) we compare if final decrypted validation tag is equal to
ENC_VAL_TAG_V1_1.

If server crashes or server gets shutdown before encryption threads
finish off rotating the tables we may be in a situation that we need to
validated multiple key versions from min_key_version to max_key_version.
However, normally the encryption threads will have a chance to fully
rotate the spaces and there will be only one key to validate, i.e.
max_key_version.

In case a space is fully rotated we encrypt validation tag plaintext
(i.e. ENC_VAL_TAG_V1_1) with max_key_version and save it to page0. Also
we set min_key_version to max_key_version.

There is also a corner case. In case we are validating a space for which
only a subset of pages is
encrypted, we may be in a situation that there are muliple encryption
keys and they do not fully fill the range [min_key_version, max_key_version],
since min_key_version == 0 is a marker that there are some unencrypted pages,
without any version. The encryption keys might be in some range [n, max_key_version],
where n > min_key_version and min_key_version = 0.
Thus for this situation (min_key_version = 0) we validate the tag after
each decryption. If the tag matches after any decryption it means we
have all the valid keys we need to decrypt space.

This WL also changes the way uuid is stored on page0. Instead of using
uuid as a string we write it as hex on page0.

This WL also adds a new field to crypt_data, and thus to page0 –
max_key_version.

Since the page0 crypt_data information was changed (uuid, max_key_version) we
also added a new private version (3) of crypt_data.
Since we are not backward compatible yet with this feature, we fail the
upgrade if table is either fully or partially encrypted by encryption
threads.

This WL also removes MariaDB’s error handling for missing encryption
keys or when incorrect key is used to decrypt a table. Decryption with
incorrect key would result in returning a corrupted page. Since MySQL
does not cope with corrupted pages and in most cases crashes when comes
across a corrupted page – the error handling was a big diff and it was
hard to maintain. Also, we found out that not all cases where covered.


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/569/

satya-bodapati:
[Design Review]
1. By not having encrypted flag until end, please check how default_table_encryption and validations work. For example CREATE TABLE t1(a INT) ENCRYPTION='N' TABLESPACE=ts1 where ts1 is tablespace with keyring encryption in progress

2. Also check how InnoDB would treat such a tablespace? encrypted or not? Specially due to lack of encryption bit in fsp_flags.

3. Rename the DD_ONLINE_ENCRYPTION flag to something that suggests it is temporary. Encryption is very generic word. Could be MK or keyring. DD_KERYING_ENC_PROGRESS may be?

4. Please check/verify if encryption validation happens on recovery? ie. how it happens on ibd_open_for_recovery()? (and instead of fil_ibd_open()). During recovery, DD is not available.
Its OK if validation is done post recovery. But please verify this.

5. With this crypt_data v3 format, will we support upgrade of encryption='n' tables with older v1 and v2 crypt_data? (We should)

@satya-bodapati re 4). When ibd is opened for recovery its first page is always validated, without checking DD flags.
re 1) For this I need to have PS-5323 merged first and possibly PS-7026. I have made this part of PS-7026.
re 2) InnoDB is already dealing with such tables by checking for present of crypt_data in page0, min_key_version and type (CRYPT_SCHEME_UNENCRYPTED and CRYPT_SCHEME_1) in crypt_data.
re 5) Yes, we support this. For v1 it is checked in upgrade_crypt_data_v1 and I have extended upgrade_crypt_data_v2 with check for ENCRYPTION='N'.
re 3) Renamed DD_SPACE_ONLINE_ENCRYPTION to DD_SPACE_ONLINE_ENC_PROGRESS.

storage/innobase/dict/dict0dd.cc
```diff
@@ -6220,22 +6224,17 @@ static bool dd_update_tablespace_dd_flags(
...
+    return true;
+  }
 
-  if (dd_space == nullptr ||
-      dd_space->se_private_data().get(dd_space_key_strings[DD_SPACE_FLAGS],
-                                      &dd_space_flags)) {
+  if (update(dd_space)) {
     dd::commit_or_rollback_tablespace_change(thd, dd_space, true);
     return (true);
   }
```
satya-bodapati: [optional] Can we combine these two?

storage/innobase/dict/dict0dd.cc
```diff
@@ -6249,24 +6248,59 @@ static bool dd_update_tablespace_dd_flags(
   return (false);
 }
 
+bool dd_set_online_encryption(THD *thd, const char *space_name,
+                              volatile bool *is_space_being_removed) {
```
satya-bodapati, line 6263: Please avoid volatile.

If a barrier is needed, please use it explicitly.

storage/innobase/fil/fil0fil.cc
```diff
@@ -6283,10 +6287,14 @@ fil_load_status Fil_shard::ibd_open_for_recovery(space_id_t space_id,
...
-      first_page
-          ? fil_space_read_crypt_data(page_size_t(df.flags()), first_page)
-          : NULL;
+  fil_space_crypt_t *crypt_data{nullptr};
+  if (df.m_crypt_data) {
+    // crypt data was already read by validate_first_page
+    crypt_data = df.m_crypt_data;
+    df.m_crypt_data = nullptr;
+  } else if (first_page) {
+    crypt_data = fil_space_read_crypt_data(page_size_t(df.flags()), first_page);
```
satya-bodapati: What if crypt_data is only in redo? Where do we read it?

storage/innobase/fil/fil0fil.cc
```diff
@@ -5927,9 +5923,14 @@ dberr_t fil_ibd_open(bool validate, fil_type_t purpose, space_id_t space_id,
...
 
   const byte *first_page = df.is_open() ? df.get_first_page() : nullptr;
-  fil_space_crypt_t *crypt_data =
-      first_page ? fil_space_read_crypt_data(page_size_t(flags), first_page)
-                 : nullptr;
+  fil_space_crypt_t *crypt_data{nullptr};
+  if (df.m_crypt_data) {
+    // crypt data was already read by validate_first_page
+    crypt_data = df.m_crypt_data;
+    df.m_crypt_data = nullptr;
```
satya-bodapati: How does this work? This is a can of worms.

df is destroyed after the initial startup phase. crypt_data is destroyed along with it.  So fil_space_t will point to dangling pointer (df.m_crypt_data)

Another issue is when freeing this dangling pointer.

satya-bodapati: Can we avoid reading crypt_data at this stage from df? Remember, we are reading from non-redo applied page. Better to read page 0 from buffer pool and get crypt-data.

satya-bodapati (changes requested):
I am not sure if relying on dd::Tablespace for this 'in-progress' state is good idea. DD is not consistent at fil_ibd_open() stage. It is really consistent only after DDL recovery. ie. after  innobase_post_recover().

What if DD trx that updates dd::Tablespace aborts or killed before it is committed?

[thinking on the alternative: may be stored in DDL_Log table?]
And analyzing if this is safe enough? :thinking:

storage/innobase/dict/dict0dd.cc
robgolebiowski: On recovery all the tables that have some redo logs to apply - have theirs page0 validated. So if there is in progress online encryption and there is no DD flag for it - the page0 will be read anyways. After it is read - crypt_data will be read and all the keys that are used by the tablespace will be checked.

storage/innobase/dict/dict0dd.cc
robgolebiowski: Barrier will have no use here. I do not want to order any reads/writes. The volatile pointer is_space_being_removed here reaches out to bool stop_new_ops. I want CPU to read this memory - in case it was changed. I only use it for handling case when space has been dropped or timeout on mdl rock was reached. To make sure I release the mdl in case of timeout or tablespace drop. If you do not like this volatile - I can break the wait on mdl only in case of timeout - but volatile here is a good option imho - the stop_new_ops can only go from false to true - So if the volatile bool becomes true - I am sure I can break waiting on the mdl. It has been working really well for quite some time and I would like to leave it as is.

storage/innobase/fil/fil0fil.cc
robgolebiowski: There was no dangling pointers - please note that after assigning m_crypt_data to space - I have made df.m_crypt_data = nullptr. So df destructor would not remove the, now space's, crypt_data. However, you are right about non-redo applied page. I have missed that. I have removed the m_crypt_data from df and went back to reading the crypt_data from page0.

storage/innobase/fil/fil0fil.cc
robgolebiowski: log0recv.cc (fil_parse_write_crypt_data_v3)

storage/innobase/dict/dict0dd.cc
satya-bodapati: Ok, this is an existing change. Ideally it should keep trying until it is successful. The MDL timeouts are regular uses case. for case of DD, I dont see a reason why there would be infinte lock time..

SInce it is an existing change, ok to keep it.
Otherwise, honestly, we need a better implementation to deal with MDL timeout on DD objects.

storage/innobase/fil/fil0fil.cc
satya-bodapati: Datafile objects lifetime is very less and to transfer data to the proper life time objects should be one in proper way. THis is risky and error prone. We avoided the pitfall for now..

Thanks for moving to page 0 reading.

storage/innobase/dict/dict0dd.cc
satya-bodapati: I mean, combining these steps would still be the logical equivalent before combining.

satya-bodapati (commented):
@robgolebiowski is it possible to do the min-max key validation a bit later, ie after DDL rollback is done?
@satya-bodapati - this is where it is validated for Master Key - so I think this should be the safest, validated place to put it in - i.e. validate_first_page.

storage/innobase/fil/fil0fil.cc
satya-bodapati: Thanks

satya-bodapati:
@robgolebiowski another question, can there be concurrent DDLs when there is key rotation in progress? Assuming the answer is no. Just double-checking.
@satya-bodapati you are right, the answer is no :) I am implementing it now.

storage/innobase/dict/dict0dd.cc
robgolebiowski: yes sure, I do not know what I was referring to in my previous comment.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1742,24 +2183,66 @@ static bool fil_crypt_start_rotate_space(const key_state_t *key_state,
...
+          if (dd_set_online_encryption(state->thd, state->space->name,
+                                       &state->space->stop_new_ops)) {
+            // should not happen
+            ib::error() << "Could not update DD for tablespace "
+                        << state->space->name
+                        << " with information on online encryption."
+                        << " Removing space from online encryption.";
+            state->space->exclude_from_rotation = true;
+            // crypt_data->max_key_version = org_max_key_version;
+            // mutex_exit(&crypt_data->mutex);
```
satya-bodapati: remove this

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1742,24 +2183,66 @@ static bool fil_crypt_start_rotate_space(const key_state_t *key_state,
...
+        if (crypt_data->min_key_version == 0) {
+          mutex_exit(&crypt_data->mutex);
+
+          if (dd_set_online_encryption(state->thd, state->space->name,
+                                       &state->space->stop_new_ops)) {
+            // should not happen
+            ib::error() << "Could not update DD for tablespace "
+                        << state->space->name
+                        << " with information on online encryption."
+                        << " Removing space from online encryption.";
```
satya-bodapati: just "online encryption" is ambiguous IMHO. Can you add the word "keyring" to this?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1742,24 +2183,66 @@ static bool fil_crypt_start_rotate_space(const key_state_t *key_state,
...
+    bool load_key_failure{false};
+    uint org_max_key_version = crypt_data->max_key_version;
+
+    if (crypt_data->load_keys_to_local_cache() == false) {
+      load_key_failure = true;
+    } else if (key_state->key_version != ENCRYPTION_KEY_VERSION_NOT_ENCRYPTED) {
+      if (crypt_data->re_encrypt_validation_tag(crypt_data->max_key_version + 1,
+                                                key_state->key_version) ==
+          false) {
+        validation_tag_re_encryption_failure = true;
```
satya-bodapati, line 2308: What do we do in this scenario? Ignore rotation for this space? Retry later?

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1711,6 +2111,47 @@ static bool fil_crypt_find_space_to_rotate(key_state_t *key_state,
...
+  }
+
+  byte re_encrypted_validation_tag[ENCRYPTION_KEYRING_VALIDATION_TAG_SIZE] = {
+      0};
+  byte copy_encrypted_validation_tag[ENCRYPTION_KEYRING_VALIDATION_TAG_SIZE];
+  memcpy(copy_encrypted_validation_tag, encrypted_validation_tag,
+         ENCRYPTION_KEYRING_VALIDATION_TAG_SIZE);
+
+  for (uint key_version = from_key_version; key_version <= to_key_version;
+       ++key_version) {
```
satya-bodapati, line 2234: Do we write intermediate versions of the tag to page 0? Also, if the write to page 0 is still in page 0, we need to validate using the version found from redo [we discussed on the channel]

satya-bodapati (changes requested):
Looks good mostly. Few questions/requests

Do you plan to add the code to prevent concurrent DDLs when key rotation is in progress as part of this PR?

@satya-bodapati no, this is a part of PS-5730 - implementation is almost ready.

Please add jenkins link and specially paste the result of newly added testcases here.

satya-bodapati:
@robgolebiowski one more request:  please update the JIRA ticket description with the updated design.

storage/innobase/fil/fil0crypt.cc

storage/innobase/fil/fil0crypt.cc

storage/innobase/fil/fil0crypt.cc
robgolebiowski: remove this space from rotation

storage/innobase/fil/fil0crypt.cc
robgolebiowski: We don't write intermediate versions of the tag to page0. Added validation if correct encryption keys are available when processing redo log entry.

satya-bodapati:
Looks good. Few things before approval

1. Update JIRA ticket with updated design
2. Test results of newly added tests in this PR as github comment
3. Jenkins link

@satya-bodapati done

robgolebiowski:
JR link  => https://ps80.cd.percona.com/job/percona-server-8.0-param/604/

tests added/changed in this PR:

200610 13:55:46 [  6%] encryption.innodb-redo-wrongkey          w2  [ pass ]  16121
200610 13:57:43 [  7%] encryption.innodb-bad-key-change4        w6  [ pass ]  23077
200610 13:58:33 [  8%] encryption.innodb-redo-nokeys            w5  [ pass ]  21504
200610 13:58:19 [  8%] encryption.innodb-redo-page0-wrongkey    w6  [ pass ]  21103
200610 13:59:32 [  8%] encryption.innodb_encryption_aborted_key_rotation w6  [ pass ]  65817
200610 14:00:47 [  9%] encryption.innodb_encryption_aborted_key_rotation_mk w6  [ pass ]  68155
200610 14:00:13 [  9%] encryption.encrypt_and_grep              w4  [ pass ]  96524
200610 13:59:45 [  9%] encryption.innodb-bad-key-change         w1  [ pass ]  28191
200610 14:01:59 [ 10%] encryption.innodb_encryption_aborted_key_rotation_mk_page_compressed w6  [ pass ]  66035
200610 14:03:19 [ 11%] encryption.innodb_encryption_aborted_key_rotation_mk_row_compressed w6  [ pass ]  72406
200610 14:04:32 [ 12%] encryption.innodb_encryption_aborted_key_rotation_page_compressed w6  [ pass ]  66068
200610 14:05:38 [ 13%] encryption.innodb_encryption_aborted_rotation w8  [ pass ]  228688
200610 14:05:50 [ 14%] encryption.innodb_encryption_aborted_key_rotation_row_compressed w6  [ pass ]  70485
200610 14:09:39 [ 18%] encryption.innodb_encryption_aborted_rotation_mk w8  [ pass ]  233981
200610 14:13:36 [ 24%] encryption.innodb_encryption_aborted_rotation_mk_page_compressed w8  [ pass ]  230049
200610 14:17:50 [ 27%] encryption.innodb_encryption_aborted_rotation_mk_row_compressed w8  [ pass ]  247268
200610 14:21:47 [ 31%] encryption.innodb_encryption_aborted_rotation_page_compressed w8  [ pass ]  229464
200610 14:26:04 [ 38%] encryption.innodb_encryption_aborted_rotation_row_compressed w8  [ pass ]  249844
200610 14:13:34 [ 24%] encryption.upgrade_crypt_data_v2         w6  [ pass ]  55836
200610 14:31:04 [ 41%] encryption.validation                    w8  [ pass ]  291283

satya-bodapati (approved):
LGTM! Please squash before merging

PR #3779, opened 2020-04-17
PS-6915 Documented support of CentOS 8. 

patrickbirch:
Modified installation/yum_repo
updated text


patrickbirch:
This pull request is outdated - PS 7054 supersedes it.

PR #3781, opened 2020-04-20
PS-6991: Remove obsolete my_config.h header and clean-up spec file

Sudokamikaze:
This was removed from upstream in version 8.0.2. Change log says:
```
Client programs should only need to #include the <mysql.h> header file. In particular, <my_config.h> should not be needed, and is no longer installed.
```

https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-2.html


EvgeniyPatlan (approved):
LGTM let's wait for @vorsel

Sudokamikaze:
@vorsel 

Upstream has old includes as well. But in our case, `my_config.h` was header with includes to `my_config_%{_arch}.h` which is missing as upstream stopped building it in 8.0.2.
So, our `my_config.h` is obsolete and may be removed

vorsel (approved):

PR #3783, opened 2020-04-20
8.0 PS-6979 - SHOW PROCESSLIST shows CREATE USER all the time till next w…

altmannmarcelo:
…riteset

Problem
If a query had been rewritten by the parser because it contains
sensitive information ( mysql_rewrite_query ), it wont be cleaned up
when sql thread applies it, making it get stuck on SHOW PROCESSLIST and
performance_schema.threads.

Solution
Call proper cleanup functions as part of
Query_log_event::do_apply_event end block.


altmannmarcelo:
Full suite - https://ps80.cd.percona.com/job/percona-server-8.0-param/570/
only new test case (it was failing under load) - https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/16617/

satya-bodapati (approved):

PR #3784, opened 2020-04-21
PS-5735: Handle linking of %{_datadir}/mysql directory properly

Sudokamikaze:
By default, CentOS 7 comes with mariadb-libs preinstalled as dependency for postfix. Re-worked scheme of linking /usr/share/percona-server to /usr/share/mysql as `if` check will always fail by default because of mariadb-libs existance and because order of processing package operations in RPM.

### DIFF
--- a/build-ps/percona-server.spec
+++ b/build-ps/percona-server.spec
@@ -552,7 +552,7 @@ rm -rf %{buildroot}%{_bindir}/mysql_embedded
 
 %pretrans -n Percona-Server-server%{product_suffix}
 if [ -d %{_datadir}/mysql ] && [ ! -L %{_datadir}/mysql ]; then
-  MYCNF_PACKAGE=$(rpm -qi `rpm -qf /usr/share/mysql` | grep -m 1 Name | awk '{print $3}')
+  MYCNF_PACKAGE=$(rpm -qf /usr/share/mysql --queryformat "%{NAME}")
 fi
 
 if [ "$MYCNF_PACKAGE" == "mariadb-libs" -o "$MYCNF_PACKAGE" == "mysql-libs" ]; then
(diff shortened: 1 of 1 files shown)


Sudokamikaze:
I think we should include another if check if it's a symlink on upgrade procedure, as %pretrans will re-create it all the time. Like `if [ -L ]` in %pretrans

What do you think? @EvgeniyPatlan @vorsel

build-ps/percona-server.spec
```diff
@@ -550,6 +550,18 @@ rm -rf %{buildroot}%{_bindir}/mysql_embedded
...
 %endif
 
+%pretrans -n Percona-Server-server%{product_suffix}
+if [ -d %{_datadir}/mysql ]; then
+  MYCNF_PACKAGE=$(rpm -qi `rpm -qf /usr/share/mysql` | grep -m 1 Name | awk '{print $3}')
+fi
+
+if [ "$MYCNF_PACKAGE" == "mariadb-libs" -o "$MYCNF_PACKAGE" == "mysql-libs" ]; then
+  MODIFIED=$(rpm -Va "$MYCNF_PACKAGE" | grep '/usr/share/mysql' | awk '{print $1}' | grep -c 5)
+  if [ "$MODIFIED" == 1 ]; then
```
EvgeniyPatlan, line 560: it is better to add check that '/usr/share/mysql' is not link

EvgeniyPatlan (changes requested):

build-ps/percona-server.spec
Sudokamikaze: Gotcha, will do

build-ps/percona-server.spec
```diff
@@ -680,6 +686,11 @@ fi
   fi
 %endif
 
+%posttrans -n Percona-Server-server%{product_suffix}
+if [ ! -d %{_datadir}/mysql ]; then
```
EvgeniyPatlan: and if it is link?

Sudokamikaze: Gotcha, sorry for missing this point. Fixed

vorsel (approved):

EvgeniyPatlan (approved):

Sudokamikaze:
Tested with new changes, works as expected and moves directory to old in %posttrans section

vorsel (approved):

EvgeniyPatlan (approved):

PR #3790, opened 2020-04-28
PS-5391: Support including sysconfig file for init.d

Sudokamikaze:


support-files/mysql.server.sh
```diff
@@ -53,6 +53,8 @@ datadir=
 # Negative numbers mean to wait indefinitely
 service_startup_timeout=900
 
+[ -e /etc/sysconfig/mysql ] && . /etc/sysconfig/mysql
```
EvgeniyPatlan, line 56: It is for centos only? what about debian/ubuntu?

EvgeniyPatlan (changes requested):

support-files/mysql.server.sh
Sudokamikaze: This change is only for centos. 
DEB has `[ -e /etc/default/mysql ] && . /etc/default/mysql || true` in init.d file

vorsel (approved):

EvgeniyPatlan (approved):

PR #3795, opened 2020-04-29
PS-6849 Add JUnit reporting support to MTR (5.7)

i-rinat:
* A timeout for the unit tests as a whole.


i-rinat:
This is almost the same as #3676, but with changes appropriate for the 5.7 branch:
* Bootstrap failures were moved from `mysqld.cc` to `bootstrap.cc`, as bootstrap code itself was moved in 5.7.
* Shutdown timeout was increased from 10 seconds to 20 seconds. In master-slave configurations, 10 seconds wasn't enough, `mysqld` was forcefully killed too often. It was OK before changes in this patches, but since shutdown timeouts are now tracked, insufficient time limits generated a lot of false positives.

PR #3796, opened 2020-04-29
PS-6849 Add JUnit reporting support to MTR (8.0)

i-rinat:


i-rinat:
This is almost the same as #3676 (and #3795), but with changes appropriate for the 8.0 branch:
* Patch with unit test timeout support was dropped, as 8.0 already has a setting for that.
* Patch that increased shutdown timeout to 20 seconds was dropped, as 8.0 already has shutdown timeout increased to 20 seconds.
* Memory writes that generate buffer overruns were changed from `my_atomic_*()` to `volatile int` writes, since `my_atomic_*()` family of functions were removed in 8.0.

PR #3797, opened 2020-04-30
PS-7004 add update possibility from 5.7 to 8.0

vorsel:
8.0 packages should obsolete 5.7 packages in order to peform an update
from ps-57 to ps-80


Sudokamikaze (approved):
LGTM

EvgeniyPatlan:
this is not needed. we shouldn't allow such updates.

PR #3799, opened 2020-05-01
PS-5008 Memory leak in sync_latch_meta_init() after mysqld shutdown d…

avbelov23:
…etected by ASan

Free all the resources acquired by InnoDB (mutexes, events, memory) in
case of a failed start (errors in
storage/innobase/srv/srv0start.cc::srv_start())

Run test with `--sanitize`
 `mtr --debug-server --sanitize innodb.log_file_name`

### DIFF
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -391,10 +391,14 @@ void buf_flush_init_flush_rbt(void) {
 void buf_flush_free_flush_rbt(void) {
   ulint i;
 
+  if (!buf_pool_ptr) return;
+
   for (i = 0; i < srv_buf_pool_instances; i++) {
     buf_pool_t *buf_pool;
 
     buf_pool = buf_pool_from_array(i);
+
+    if (!buf_pool || !buf_pool->flush_rbt) continue;
 
     buf_flush_list_mutex_enter(buf_pool);
 
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -4047,7 +4047,11 @@ void Fil_system::close_all_files() {
 
 /** Closes all open files. There must not be any pending i/o's or not flushed
 modifications in the files. */
-void fil_close_all_files() { fil_system->close_all_files(); }
+void fil_close_all_files() {
+  if (!fil_system) return;
+
+  fil_system->close_all_files();
+}
 
 /** Close log files.
 @param[in]	free_all	If set then free all instances */
@@ -4250,6 +4254,7 @@ for concurrency control.
 @param[in]	space_id	Tablespace ID
 @return the tablespace, or nullptr if missing or being deleted */
 fil_space_t *fil_space_acquire(space_id_t space_id) {
+  if (!fil_system) return nullptr;
   return fil_system->space_acquire(space_id, false);
 }
 
@@ -9278,6 +9283,8 @@ possibly cached by the OS.
 @param[in]	purpose		FIL_TYPE_TABLESPACE or FIL_TYPE_LOG, can be
 ORred. */
 void fil_flush_file_spaces(uint8_t purpose) {
+  if (!fil_system) return;
+
   fil_system->flush_file_spaces(purpose);
 }
 
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -1776,29 +1776,28 @@ static void innodb_space_shutdown() {
 static int innodb_shutdown(handlerton *, ha_panic_function) {
   DBUG_TRACE;
 
+  hash_table_free(innobase_open_tables);
+  innobase_open_tables = nullptr;
+
+  for (auto file : innobase_sys_files) {
+    ut::delete_(file);
+  }
+  innobase_sys_files.clear();
+  innobase_sys_files.shrink_to_fit();
+
+  if (mutex_monitor) mutex_free(&master_key_id_mutex);
+  srv_shutdown();
+  innodb_space_shutdown();
+
   if (innodb_inited) {
-    innodb_inited = false;
-    hash_table_free(innobase_open_tables);
-    innobase_open_tables = nullptr;
-
-    for (auto file : innobase_sys_files) {
-      ut::delete_(file);
-    }
-    innobase_sys_files.clear();
-    innobase_sys_files.shrink_to_fit();
-
-    mutex_free(&master_key_id_mutex);
-    srv_shutdown();
-    innodb_space_shutdown();
-
     mysql_mutex_destroy(&innobase_share_mutex);
     mysql_mutex_destroy(&commit_cond_m);
     mysql_cond_destroy(&commit_cond);
     mysql_mutex_destroy(&resume_encryption_cond_m);
     mysql_cond_destroy(&resume_encryption_cond);
-
-    os_event_global_destroy();
   }
+
+  os_event_global_destroy();
 
   innobase::component_services::deinitialize_service_handles();
 
@@ -5846,8 +5845,8 @@ static int innodb_init(void *p) {
 
   os_event_global_init();
 
-  if (int error = innodb_init_params()) {
-    return error;
+  if (innodb_init_params()) {
+    return innodb_init_abort();
   }
 
   /* After this point, error handling has to use
--- a/storage/innobase/log/log0online.cc
+++ b/storage/innobase/log/log0online.cc
@@ -125,6 +125,9 @@ static const constexpr auto LOG_BLOCK_SIZE_NO_TRL =
 static const constexpr auto LOG_BLOCK_BOUNDARY_LSN_PAD =
     LOG_BLOCK_HDR_SIZE + LOG_BLOCK_TRL_SIZE;
 
+/** Database log parsing inited */
+static bool log_online_inited = false;
+
 /** The buffer for reading log data, filled in by recv_read_log_recs and moved
 in chunks to the parse buffer while skipping log block headers and trailers. */
 class log_read_buffer final : public log_buffer<FOLLOW_SCAN_SIZE> {
@@ -875,6 +878,7 @@ static bool log_online_is_bitmap_file(
 
 void log_online_init(void) noexcept {
   mutex_create(LATCH_ID_LOG_ONLINE, &log_bmp_sys_mutex);
+  log_online_inited = true;
 }
 
 /** Initialize the dynamic part of the log tracking subsystem */
@@ -1050,7 +1054,12 @@ void log_online_read_shutdown(void) noexcept {
 }
 
 /** Shut down the constant part of the log tracking subsystem */
-void log_online_shutdown(void) noexcept { mutex_free(&log_bmp_sys_mutex); }
+void log_online_shutdown(void) noexcept {
+  if (!log_online_inited) return;
+
+  mutex_free(&log_bmp_sys_mutex);
+  log_online_inited = false;
+}
 
 /** Parse the log data in the parse buffer for the (space, page) pairs and add
 them to the modified page set as necessary.  Removes the fully-parsed records
--- a/storage/innobase/pars/pars0pars.cc
+++ b/storage/innobase/pars/pars0pars.cc
@@ -93,11 +93,23 @@ ulint pars_star_denoter = 12345678;
 /** Mutex to protect the sql parser */
 ib_mutex_t pars_mutex;
 
+/** Parsing inited */
+static bool pars_inited = false;
+
 /** Initialize for the internal parser */
-void pars_init() { mutex_create(LATCH_ID_PARSER, &pars_mutex); }
+void pars_init() {
+  mutex_create(LATCH_ID_PARSER, &pars_mutex);
+  pars_inited = true;
+}
 
 /** Clean up the internal parser */
-void pars_close() { mutex_free(&pars_mutex); }
+void pars_close() {
+  if (!pars_inited) return;
+
+  mutex_free(&pars_mutex);
+
+  pars_inited = false;
+}
 
 /********************************************************************
 Get user function with the given name.*/
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -108,6 +108,7 @@ this program; if not, write to the Free Software Foundation, Inc.,
 #include "dict0load.h"
 #include "dict0stats_bg.h"
 #include "lock0lock.h"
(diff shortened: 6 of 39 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -1642,8 +1641,12 @@ static int innodb_shutdown(handlerton *, ha_panic_function) {
     mysql_cond_destroy(&resume_encryption_cond);
 
     os_event_global_destroy();
+  } else {
+      srv_fast_shutdown = 2;
```
satya-bodapati: why do we need to set fast_shutdown as 2 if innodb is not inited?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1631,7 +1631,6 @@ static int innodb_shutdown(handlerton *, ha_panic_function) {
     innobase_sys_files.clear();
     innobase_sys_files.shrink_to_fit();
 
-    mutex_free(&master_key_id_mutex);
```
satya-bodapati, line 1790: why this mutex alone is special?

storage/innobase/log/log0online.cc
```diff
@@ -1051,7 +1051,13 @@ void log_online_read_shutdown(void) noexcept {
 }
 
 /** Shut down the constant part of the log tracking subsystem */
-void log_online_shutdown(void) noexcept { mutex_free(&log_bmp_sys_mutex); }
+void log_online_shutdown(void) noexcept {
+  /* log_online_init() is called after buf_pool_init(), and if buf_pool_ptr !=
+   * nullptr, then log_bmp_sys_mutex was created */
+  if (!buf_pool_ptr) return;
+
```
satya-bodapati: Its better to use a log online state (or a corresponding global variable) instead of buf_pool

storage/innobase/pars/pars0pars.cc
```diff
@@ -97,7 +97,13 @@ ib_mutex_t pars_mutex;
 void pars_init() { mutex_create(LATCH_ID_PARSER, &pars_mutex); }
 
 /** Clean up the internal parser */
-void pars_close() { mutex_free(&pars_mutex); }
+void pars_close() {
+  /* pars_init() is called after buf_pool_init(), and if buf_pool_ptr !=
+   * nullptr, then pars_mutex was created */
+  if (!buf_pool_ptr) return;
```
satya-bodapati: Same as above, Its better to use a internal parser state (or a corresponding global variable) instead of buf_pool

storage/innobase/srv/srv0start.cc
```diff
@@ -3935,7 +3947,10 @@ void srv_fatal_error() {
 
   ut_d(innodb_calling_exit = true);
 
+  srv_fast_shutdown = 2;
+
```
satya-bodapati: why 2 on fatal error?, what if we don't set this to 2?

storage/innobase/srv/srv0tmp.cc
```diff
@@ -81,7 +81,6 @@ Tablespace::~Tablespace() {
 
   close();
 
-  ut_ad(srv_shutdown_state.load() == SRV_SHUTDOWN_EXIT_THREADS);
```
satya-bodapati: why this assertion removed? what is the state on the error condition?

storage/innobase/trx/trx0sys.cc
```diff
@@ -577,8 +577,6 @@ void trx_sys_create_sys_pages(void) {
 /*********************************************************************
 Shutdown/Close the transaction system. */
 void trx_sys_close(void) {
-  ut_ad(srv_shutdown_state.load() == SRV_SHUTDOWN_EXIT_THREADS);
```
satya-bodapati: same as above, why is it removed? can we fix the assert instead? what is the state on the error condition?

satya-bodapati (changes requested):

storage/innobase/srv/srv0tmp.cc
avbelov23: As far as I remember we want to use these functions, but in cases where `srv_shutdown_state != SRV_SHUTDOWN_EXIT_THREADS` (from `srv_fatal_error()`)

storage/innobase/handler/ha_innodb.cc
avbelov23: No. You can see below in `srv_free_resources()` its freeing

storage/innobase/trx/trx0sys.cc
avbelov23: As far as I remember we want to use these functions, but in cases where srv_shutdown_state != SRV_SHUTDOWN_EXIT_THREADS (from srv_fatal_error())

avbelov23:
@satya-bodapati, do you have enough explanation to remove `ut_ad()`?
We now clean up resources and not during `SRV_SHUTDOWN_EXIT_THREADS` state - e.g. on innodb initialization failure (see description for PR) 
If not difficult restart CI. There is a failure due to 403 Forbidden on wget ...

satya-bodapati:
> @satya-bodapati, do you have enough explanation to remove `ut_ad()`? We now clean up resources and not during `SRV_SHUTDOWN_EXIT_THREADS` state - e.g. on innodb initialization failure (see description for PR) If not difficult restart CI. There is a failure due to 403 Forbidden on wget ...
Hi @avbelov23 
what will be the SRV_ state on error path? Can we extend the assert instead?

mysql-test/suite/innodb/t/log_file_name.test
```diff
@@ -0,0 +1,246 @@
...
+RENAME TABLE t1 TO t2;
+UPDATE t2 SET a=347 WHERE a=42;
+COMMIT;
+
+--let CLEANUP_IF_CHECKPOINT=DROP TABLE t2,t3;
+--source include/no_checkpoint_end.inc
+
+--echo # Fault 0 (no real fault): Orphan file with duplicate space_id.
+--copy_file $MYSQLD_DATADIR/test/t2.ibd $MYSQLD_DATADIR/test/t0.ibd
+
```
satya-bodapati: is this an existing test duplicated?

satya-bodapati (changes requested):

avbelov23:
>is this an existing test duplicated?

@satya-bodapati, maybe I didn't understand the question.
This is a test from 5.7

storage/innobase/trx/trx0sys.cc
satya-bodapati: @avbelov23 did you print srv_shutdown_state and check?

avbelov23:
@satya-bodapati, I expanded asserts
DONE

mysql-test/suite/innodb/t/log_file_name.test
satya-bodapati: This test is removed in the 8.0 series because this WL is reverted. I don't think we should bring back the entire test as is.  Can you please identify the scenarios that fail starupt and are not from this test file?

storage/innobase/buf/buf0flu.cc
```diff
@@ -365,11 +365,15 @@ void buf_flush_init_flush_rbt(void) {
...
 
+  if (!buf_pool_ptr) return;
+
   for (i = 0; i < srv_buf_pool_instances; i++) {
     buf_pool_t *buf_pool;
 
     buf_pool = buf_pool_from_array(i);
 
+    if (!buf_pool || !buf_pool->flush_rbt) continue;
+
```
satya-bodapati, line 402: why do we need this change if we are returning from above? (368)

satya-bodapati (changes requested):

storage/innobase/buf/buf0flu.cc
avbelov23: Above `buf_pool_ptr`. This is `buf_pool`

avbelov23: These are initialized `buf_pool_ptr` in `buf_flush_init_flush_rbt()`, here we correctly handle the situation when `buf_flush_init_flush_rbt()` was not called (we clear only those resources that were initialized during an error at the initialization stage).
In the condition above, we handle the situation when `buf_pool_init()` was not called in the same way, so as not to crash here

avbelov23:
@satya-bodapati, if it's ok, then I'll merge the commits and do GCA

storage/innobase/buf/buf0flu.cc
satya-bodapati: @avbelov23 can you please reply to this question?

avbelov23:
```
avb@avb-percona:~/Projects/build-percona-server-Desktop-Debug$ MTR_SUITE_TIMEOUT=1440 MTR_TESTCASE_TIMEOUT=30 time mysql-test/mtr --debug-server --sanitize innodb.fail_start_san
Logging: /home/avb/Projects/percona-server/mysql-test/mysql-test-run.pl  --debug-server --sanitize innodb.fail_start_san
MySQL Version 8.0.19
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Checking leftover processes
Removing old var directory
Creating var directory '/home/avb/Projects/build-percona-server-Desktop-Debug/mysql-test/var'
Installing system database
Using parallel: 1

[ 50%] innodb.fail_start_san                     [ pass ]   7093
[100%] sanitize_report                           [ pass ]       
Spent 7.093 of 72 seconds executing testcases

Completed: All 2 tests were successful.

55.61user 13.25system 1:11.99elapsed 95%CPU (0avgtext+0avgdata 781600maxresident)k
1126808inputs+2410544outputs (1641major+849419minor)pagefaults 0swaps
```

satya-bodapati:
> ```
> avb@avb-percona:~/Projects/build-percona-server-Desktop-Debug$ MTR_SUITE_TIMEOUT=1440 MTR_TESTCASE_TIMEOUT=30 time mysql-test/mtr --debug-server --sanitize innodb.fail_start_san
> Logging: /home/avb/Projects/percona-server/mysql-test/mysql-test-run.pl  --debug-server --sanitize innodb.fail_start_san
> MySQL Version 8.0.19
> Checking supported features
>  - Binaries are debug compiled
> Using 'all' suites
> Collecting tests
> Checking leftover processes
> Removing old var directory
> Creating var directory '/home/avb/Projects/build-percona-server-Desktop-Debug/mysql-test/var'
> Installing system database
> Using parallel: 1
> 
> ==============================================================================
>                   TEST NAME                       RESULT  TIME (ms) COMMENT
> ------------------------------------------------------------------------------
> [ 50%] innodb.fail_start_san                     [ pass ]   7093
> [100%] sanitize_report                           [ pass ]       
> ------------------------------------------------------------------------------
> The servers were restarted 0 times
> The servers were reinitialized 0 times
> Spent 7.093 of 72 seconds executing testcases
> 
> Completed: All 2 tests were successful.
> 
> 55.61user 13.25system 1:11.99elapsed 95%CPU (0avgtext+0avgdata 781600maxresident)k
> 1126808inputs+2410544outputs (1641major+849419minor)pagefaults 0swaps
> ```
Looks good. Can you please rename the test to 'startup_failure'?

satya-bodapati (approved):
I think we can skip this PR for 5.7 because this impacts only shutdowns caused by errors.

avbelov23:
@satya-bodapati, need to revert a previous commit w/o GCA?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1776,30 +1776,29 @@ static void innodb_space_shutdown() {
...
+  for (auto file : innobase_sys_files) {
+    ut::delete_(file);
+  }
+  innobase_sys_files.clear();
+  innobase_sys_files.shrink_to_fit();
 
-    mutex_free(&master_key_id_mutex);
-    srv_shutdown();
-    innodb_space_shutdown();
+  if (mutex_monitor) mutex_free(&master_key_id_mutex);
```
satya-bodapati, line 1788: This is create at the very start. IMHO, we shouldn't need the mutex_monitor check

storage/innobase/fil/fil0fil.cc
```diff
@@ -9278,6 +9283,8 @@ possibly cached by the OS.
 @param[in]	purpose		FIL_TYPE_TABLESPACE or FIL_TYPE_LOG, can be
 ORred. */
 void fil_flush_file_spaces(uint8_t purpose) {
+  if (!fil_system) return;
```
satya-bodapati, line 9286: when will we reach here without the fil_system initialization?

satya-bodapati (commented):
Looks much better with the testcase adjustment and minimal code.

storage/innobase/fil/fil0fil.cc
avbelov23: We call `fil_flush_file_spaces()` from `srv_shutdown()`. Now `srv_shutdown()` work and w/ innodb inititialization error, but innodb inititialization error may be w/o the `fil_system` initialization

storage/innobase/handler/ha_innodb.cc
avbelov23: Returns from `srv_start()` exist above `master_key_id_mutex` creating. `if` need

storage/innobase/fil/fil0fil.cc
satya-bodapati: @avbelov23 I see that fil_init() is called at very early stage of srv_start(). This means fil_system should be initialized. Is it possible that we are erroring out without srv_start() being called?

Can you please make it as an assert ut_a(fil_system != nullptr) and get stack trace?

storage/innobase/handler/ha_innodb.cc
satya-bodapati: You mean, we can reach here without srv_start()?

storage/innobase/fil/fil0fil.cc
avbelov23: See https://github.com/tempesta-tech/percona-server/blob/avb-log_file_name/storage/innobase/srv/srv0start.cc#L2196
This return before fil_init()

storage/innobase/handler/ha_innodb.cc
avbelov23: See https://github.com/tempesta-tech/percona-server/blob/avb-log_file_name/storage/innobase/srv/srv0start.cc#L2196
This return before `master_key_id_mutex` creating

satya-bodapati:
Looks good. tested on Jenkins: https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-asan-param/31/

satya-bodapati (approved):

PR #3803, opened 2020-05-06
PS-5008 Memory leak in sync_latch_meta_init() after mysqld shutdown d…

avbelov23:
…etected by ASan

Free all the resources acquired by InnoDB (mutexes, events, memory) in
case of a failed start (errors in
storage/innobase/srv/srv0start.cc::innobase_start_or_create_for_mysql())

Also fixed the launch of tests under asan


satya-bodapati:
@avbelov23 Let's skip this bug fix for 5.7 and do 8.0 only.

PR #3807, opened 2020-05-07
PS-5620: Add OpenSSL to dependencies of packages

Sudokamikaze:
All OS **exept** Docker images have OpenSSL installed by default

Because `mysql_ssl_rsa_setup` uses OpenSSL CLI utils and invokes them as `command()`


EvgeniyPatlan (changes requested):
please detect what exect dependency is needed

EvgeniyPatlan (approved):

vorsel (approved):

PR #3808, opened 2020-05-07
PS-6989: "Cannot write: No space left on device" error with Azure Pipelines

inikep:
Use `-DMERGE_UNITTESTS=ON` to save about 900 MB with `gcc-9` in Debug compilation.


percona-ysorokin (approved):
LGTM

PR #3811, opened 2020-05-08
PS-7041: Compilation doesn't work -DWITH_EDITLINE=bundled

inikep:
1. Restore upstream solution from https://github.com/mysql/mysql-server/commit/62f319d3a620
2. Set `-DWITH_EDITLINE=bundled` as default for macOS


percona-ysorokin (approved):
LGTM

PR #3825, opened 2020-05-14
PS-6886 Documented a new variable rocksdb_cache_dump which includes RocksDB block cache content in a core dump.

patrickbirch:
Merged 5.7 version
modified variable description.


inikep (approved):
LGTM

PR #3826, opened 2020-05-14
PS-6885 Documented the new variable rocksdb_rollback_on_timeout which allows the rollback of an entire transaction on timeout.

patrickbirch:
Merged 5.7 version
Modified variable description.


inikep (approved):
LGTM

PR #3827, opened 2020-05-15
PS-7067: Cannot build PS 5.7 for centos6, centos7 and debian:jessie

inikep:
Replace `STRING(APPEND ...)` that requires cmake 3.4 with `SET`.


inikep:
Tested at https://ps57.cd.percona.com/job/percona-server-5.7-param/274/

hors (approved):

PR #3832, opened 2020-05-18
PS-7054 Link to Percona Software Lifecycle

patrickbirch:
Merged 5.6 version
fixed conflicts in apt_repo 
fixed conflicts in yum_repo


kathy-williamson (changes requested):
I believe the following should remain in the documentation - it is not version related

 *CentOS* 6 offers an outdated version of the ``curl`` library
    required by the :ref:`keyring Vault plugin <keyring_vault_plugin>`
    of |Percona Server|. The version of the ``curl`` library in
    *CentOS* 6, which depends on the ``nss`` library, is known to
    create memory corruption issues. This bug is `registered in Red
    Hat Bugzilla
    <https://bugzilla.redhat.com/show_bug.cgi?id=1057388>`_. Its
    current status is `CLOSED WONTFIX`.

    If you intend to use the keyring Vault plugin of |Percona Server|
    make sure that you use the latest version of the ``curl`` library.
    We recommend that you `build it from source
    <https://curl.haxx.se/docs/install.html>`_ configuring with
    ``ssl`` but without ``nss``:

kathy-williamson (approved):

PR #3833, opened 2020-05-19
PS-7077: Azure Pipelines: clang-5.0 fails if gcc-9+ is installed

inikep:
This is workaround for clang-5 bug with incorrect order of using system include directories that collide with gcc-9/gcc-10 include files.


percona-ysorokin (approved):
LGTM

PR #3839, opened 2020-05-22
5.6 PS-7043 - Incorrect result when constant equality expression is used …

altmannmarcelo:
…in LEFT JOIN condition

Problem
Optimizer is converting ON condition into a multiple equality. When
transforming the condition (table1.id=table2.id AND table2.id=1), it
transformed the query as only (multiple equal(1,'1', table2.id)).

Solution
This has been fixed on 8.0 series already. Partially cherry picked
commit [1] and its test case. Adjusted tests that were failing as result
of imported test case.

[1]:

commit 035fc0f6600253294834482fcad00050db55864c
Author: Ole John Aske <ole.john.aske@oracle.com>
Date:   Wed Mar 20 13:55:26 2019 +0100

    Bug#29493830 CONST'IFIED OUTER JOIN RETURN INCORRECT RESULTS

    When a table is const optimized, any predicate terms on the joins ON-condition
    is also evaluated on the const'ified tables to make sure that the row
    qualifies on the ON-condition.

    The ON-conditions might have been transformed to a 'multiple equality Item',
    represented as a Item_equal object at this point of the optimization process.
    Effectively the ON-condition is evaluated in Item_equal::val_int(), which
    will only check the equality for any Item_field where the fields table
    has been marked as a 'const_table' - comparison of other field values
    are intentionaly skipped.

    It turned out that the optimizer failed to mark a table as a 'const_table'
    when it was handled as a 'impossible ON condition'. In this case the resulting
    row from this table will always be a NULL-extended row.

    This patch set the 'const_table' flag together with setting the
    row as a NULL-row.

    Patch also change some explain output from other test cases.
    These changes are exclusively related to table being
    'Impossible ON-condition' optimized, where the resulting NULL values from
    this row is now const propagated into the rows Field values.
    (Just like any other const-row)

### DIFF
--- a/mysql-test/r/join_outer.result
+++ b/mysql-test/r/join_outer.result
@@ -2106,3 +2106,65 @@ SELECT straight_join t1.vc, t1.ik
 FROM t1 JOIN t1 AS t2 ON t1.vc=t2.vc LEFT JOIN t1 AS t3 ON t1.vc=t3.vc;
 vc	ik
 DROP TABLE t1;
+#
+# Bug#29493830 CONST'IFIED OUTER JOIN RETURN INCORRECT RESULTS
+#
+CREATE TABLE t1 (
+col_int_unique INT DEFAULT NULL,
+col_int_key INT DEFAULT NULL,
+UNIQUE KEY col_int_unique (col_int_unique)
+) ENGINE=InnoDB;
+INSERT INTO t1 VALUES (5,0);
+CREATE TABLE t2 (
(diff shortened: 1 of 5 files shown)


altmannmarcelo:
https://ps56.cd.percona.com/job/percona-server-5.6-param/99/

altmannmarcelo:
New param link after jenkins fixes - https://ps56.cd.percona.com/job/percona-server-5.6-param/100/

percona-ysorokin (approved):
LGTM

PR #3843, opened 2020-05-26
PS-7001 Document new SEQUENCE_TABLE() table function.

patrickbirch:
modified index
added sequence_table

### DIFF
--- a/doc/source/flexibility/sequence_table.rst
+++ b/doc/source/flexibility/sequence_table.rst
@@ -1,0 +1,124 @@
+.. _sequence_table:
+
+===============================================================================
+SEQUENCE_TABLE(n) Function
+===============================================================================
+
+|Percona Server| 8.0.20-11 adds the SEQUENCE_TABLE() function.
+
+A sequence of numbers can be defined as an arithmetic progression when the common difference between two consecutive terms is always the same.
+
+The function is an inline table-valued function. A single SELECT statement generates a multi-row result set. In contrast, a scalar function (like `EXP(x) <https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_exp>`_ or `LOWER(str) <https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_lower>`_ always  returns a single value of a specific data type.
+
+The `JSON_TABLE() <https://dev.mysql.com/doc/refman/8.0/en/json-table-functions.html>`_ is the only table function available in Oracle MySQL Server. ``JSON_TABLE`` and ``SEQUENCE_TABLE()`` are the only table functions available in Percona Server.
+
+The basic syntax is the following:
+
+* SEQUENCE_TABLE(n) [AS] alias
+ 
+  *Usage*:
+  SELECT ... FROM SEQUENCE_TABLE(n) [AS] alias
+
+    .. code-block:: mysql
+
+        SEQUENCE_TABLE(
+            n
+        ) [AS] alias
+    
+``n:`` The number of generated values.
+
+As with any `derived tables <https://dev.mysql.com/doc/refman/8.0/en/derived-tables.html>`_, a table function requires an `alias <https://dev.mysql.com/doc/refman/8.0/en/identifiers.html>`_ in the ``SELECT`` statement. 
+
+The result set is a single column with the predefined column name ``value`` of type ``BIGINT UNSIGNED``. You can reference the ``value`` column in ``SELECT`` statements. The following statements are valid. ::
+
+    SELECT * FROM SEQUENCE_TABLE(n) AS tt;
+    SELECT <expr(value)> FROM SEQUENCE_TABLE(n) AS tt;
+
+The first number in the series, the initial term, is defined as ``0`` and the series ends with a value less then ``n``. In this example, enter the following statement to generate a sequence:
+
+..  code-block:: mysql
+
+    mysql> SELECT * FROM SEQUENCE_TABLE(3) AS tt;
+    +-------+
+    | value |
+    +-------+
+    |     0 |
+    |     1 |
+    |     2 |
+    +-------+
+    
+You can define  the initial term using the ``WHERE`` clause. The following example starts the sequence with ``4``.
+
+.. code-block:: mysql
+
+    SELECT value AS result FROM SEQUENCE_TABLE(8) AS tt WHERE value >= 4;
+    +--------+
+    | result |
+    +--------+
+    |      4 |
+    |      5 |
+    |      6 |
+    |      7 |
+    +--------+
+    
+Consecutive terms increase or decrease by a common difference. The default common difference value is ``1``. However, it is possible to filter the results using the WHERE clause to simulate common differences greater than 1.
+
+The following example prints only even numbers from the 0..7 range:
+
+.. code-block:: mysql
+
+    SELECT value AS result FROM SEQUENCE_TABLE(8) AS tt WHERE value % 2 = 0;
+    +--------+
+    | result |
+    +--------+
+    |      0 |
+    |      2 |
+    |      4 |
+    |      6 |
+    +--------+
+
+The following is an example of using the function to populate a table with a set of random numbers:
+
+.. code-block:: mysql
+
+    mysql> SELECT FLOOR(RAND()) * 100) AS result FROM SEQUENCE_TABLE(4) AS tt;
+    +--------+
+    | result |
+    +--------+
+    |     24 |
+    |     56 |
+    |     70 |
+    |     25 |
+    +--------+
+
+You can populate a table with a set of pseudo-random strings with the following statement:
+
+.. code-block:: mysql
+
+    mysql> SELECT MD5(value) AS result FROM SEQUENCE_TABLE(4) AS tt;
+    +----------------------------------+
+    | result                           |
+    +----------------------------------+
+    | f17d9c990f40f8ac215f2ecdfd7d0451 |
+    | 2e5751b7cfd7f053cd29e946fb2649a4 |
+    | b026324c6904b2a9cb4b88d6d61c81d1 |
+    | 26ab0db90d72e28ad0ba1e22ee510510 |
+    +----------------------------------+
+
+You can add the sequence as a column to a new table or an existing table, as shown in this example:
+
+.. code-block:: mysql
+
+    mysql> CREATE TABLE t1 AS SELECT * FROM SEQUENCE_TABLE(4) AS tt;
+    
+    mysql> SELECT * FROM t1;
+    +-------+
+    | value |
+    +-------+
+    |     0 |
+    |     1 |
+    |     2 |
+    |     3 |
+    +-------+
+
+There are many uses for a sequence when populating tables.
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -92,6 +92,7 @@ Flexibility Improvements
    flexibility/innodb_fts_improvements
    flexibility/binlogging_replication_improvements
    flexibility/extended_set_var
+   flexibility/sequence_table
 
 Reliability Improvements
 ================================================================================


kathy-williamson (approved):

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,66 @@
+.. _sequence_table:
+
+===============================================================================
+Sequence_Table() Function
```
percona-ysorokin:
```suggestion
SEQUENCE_TABLE(n) Function
```
Also explain what `n` means

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,66 @@
...
+===============================================================================
+
+A sequence is a set of unique, user-defined values produced on demand. |Percona Server| 8.0.20-11 adds the SEQUENCE_TABLE() function which simplifies creating the sequence.
```
percona-ysorokin: Mathematically speaking, `SEQUENCE_TABLE(n)` generates an arithmetic progression (https://en.wikipedia.org/wiki/Arithmetic_progression) with the first member (a0) equal to 0 and common difference of 1 limited to values less than `n`
I would also highlight that this is a TABLE function (not a scalar one). In other words, this is a functions that generates a multirow result.
At the moment (in 8.0.20) Oracle MySQL Server has only one existing TABLE function - `JSON_TABLE()` (https://dev.mysql.com/doc/refman/8.0/en/json-table-functions.html).
Percona server in 8.0.20-11 introduces the second one - `SEQUENCE_TABLE()`.
The primary difference in usage scenarios between scalar and table functions is the following.
Scalar - `SELECT scalar_function(...) FROM ...`
Table - `SELECT... FROM table_function() AS table_alias`

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,66 @@
...
+
+To generate a sequence, use the following statement:
+
+..  code-block:: mysql
+
+    mysql> SELECT * FROM SEQUENCE_TABLE(3) as tt;
```
percona-ysorokin:
```suggestion
   mysql> SELECT * FROM SEQUENCE_TABLE(3) AS tt;
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,66 @@
...
+    |     0 |
+    |     1 |
+    |     2 |
+    +-------+
+    
+The following is an example of using the function to populate a table with a set of random numbers:
+
+.. code-block:: mysql
+
+    mysql> SELECT FLOOR(RAND()) * 100) AS result FROM SEQUENCE_TABLE(4) as tt;
```
percona-ysorokin:
```suggestion
    mysql> SELECT FLOOR(RAND()) * 100) AS result FROM SEQUENCE_TABLE(4) AS tt;
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,66 @@
...
+    |     56 |
+    |     70 |
+    |     25 |
+    +--------+
+
+You can populate a table with a set of pseudo-random strings with the following statement:
+
+.. code-block:: mysql
+
+    mysql> SELECT MD5(value) AS result FROM SEQUENCE_TABLE(4) as tt;
```
percona-ysorokin:
```suggestion
    mysql> SELECT MD5(value) AS result FROM SEQUENCE_TABLE(4) AS tt;
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,66 @@
...
+    | 2e5751b7cfd7f053cd29e946fb2649a4 |
+    | b026324c6904b2a9cb4b88d6d61c81d1 |
+    | 26ab0db90d72e28ad0ba1e22ee510510 |
+    +----------------------------------+
+
+You can add the sequence as a column to a new table or an existing table, as shown in this example:
+
+.. code-block:: mysql
+
+    mysql> CREATE TABLE t1 AS SELECT * FROM SEQUENCE_TABLE(4) as tt;
```
percona-ysorokin:
```suggestion
    mysql> CREATE TABLE t1 AS SELECT * FROM SEQUENCE_TABLE(4) AS tt;
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,66 @@
+.. _sequence_table:
+
+===============================================================================
+Sequence_Table() Function
+===============================================================================
+
+A sequence is a set of unique, user-defined values produced on demand. |Percona Server| 8.0.20-11 adds the SEQUENCE_TABLE() function which simplifies creating the sequence.
+
```
percona-ysorokin, line 79: Please also add that MySQL syntax requires that table functions must always be used with table aliases (`AS <table_alias>`).
For instance,
`SELECT * FROM SEQUENCE_TABLE(4)` - not OK
`SELECT * FROM SEQUENCE_TABLE(4) AS tt` - OK

percona-ysorokin (changes requested):

doc/source/flexibility/sequence_table.rst
patrickbirch: Redid the title.

doc/source/flexibility/sequence_table.rst
patrickbirch: Updated the text to arithmetic progression instead of sequence.

doc/source/flexibility/sequence_table.rst
patrickbirch: Fixed.

doc/source/flexibility/sequence_table.rst
patrickbirch: Fixed.

doc/source/flexibility/sequence_table.rst
patrickbirch: Fixed.

doc/source/flexibility/sequence_table.rst
patrickbirch: Fixed.

doc/source/flexibility/sequence_table.rst
patrickbirch: Updated.

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
...
+===============================================================================
+SEQUENCE_TABLE(n) Function
+===============================================================================
+
+|Percona Server| 8.0.20-11 adds the SEQUENCE_TABLE() function.
+
+A sequence of numbers can be defined as an arithmetic progression when the common difference between two consecutive terms is always the same. If the common difference is positive, the sequence grows in positive terms; if the common difference is negative, the sequence grows in negative terms.
+
+The function is an inline table-valued function. A single SELECT statement generates a multi-row result set. A scalar function returns a single value of a specific data type, which is defined in a ``RETURNS`` clause.
```
percona-ysorokin:
```suggestion
The function is an inline table-valued function. A single SELECT statement generates a multi-row result set. In contrast, a scalar function (like `EXP(x) <https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_exp>`_ or `LOWER(str) <https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_lower>`_ always returns a single value of a specific data type.
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
+.. _sequence_table:
+
+===============================================================================
+SEQUENCE_TABLE(n) Function
+===============================================================================
+
+|Percona Server| 8.0.20-11 adds the SEQUENCE_TABLE() function.
+
+A sequence of numbers can be defined as an arithmetic progression when the common difference between two consecutive terms is always the same. If the common difference is positive, the sequence grows in positive terms; if the common difference is negative, the sequence grows in negative terms.
```
percona-ysorokin: The sentence about negative terms is definitely true but as our implementation of `sequence_table(n)` generates a sequence that always starts from 0 and always grow with common term of 1, I would remove it to avoid confusion. 
```suggestion
A sequence of numbers can be defined as an arithmetic progression when the common difference between two consecutive terms is always the same.
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
...
+
+The function is an inline table-valued function. A single SELECT statement generates a multi-row result set. A scalar function returns a single value of a specific data type, which is defined in a ``RETURNS`` clause.
+
+The SEQUENCE_TABLE() and the `JSON_TABLE() <https://dev.mysql.com/doc/refman/8.0/en/json-table-functions.html>`_ functions are the only TABLE functions available in MySQL.
```
percona-ysorokin: `JSON_TABLE()` is the only table function available in Oracle MySQL Server.
`JSON_TABLE()` and `SEQUENCE_TABLE()` are the only table functions available in Percona Server

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
...
+
+The basic syntax is the following:
+
+* SEQUENCE_TABLE(n) [AS] alias
```
percona-ysorokin, line 17: I would put additional context here indication where it can be used
```suggestion
* SELECT ... FROM SEQUENCE_TABLE(n) [AS] alias
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
...
+
+    .. code-block:: mysql
+
+        SEQUENCE_TABLE(
```
percona-ysorokin, line 24:
```suggestion
        SELECT ... FROM SEQUENCE_TABLE(
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
...
+            n
+        ) [AS] alias
+    
+``n:`` The number of generated values.
+
+The MySQL syntax requires the ``alias`` in the SELECT statement. The rules to define an alias are defined `here <https://dev.mysql.com/doc/refman/8.0/en/identifiers.html>`_.
```
percona-ysorokin:
```suggestion
As for any other `Derived Table <https://dev.mysql.com/doc/refman/8.0/en/derived-tables.html>`_ (including table functions) the MySQL syntax requires the ``alias`` in the SELECT statement. The rules to define an alias are defined `here <https://dev.mysql.com/doc/refman/8.0/en/identifiers.html>`_.

File: doc/source/flexibility/sequence_table.rst [hunk 14]
```diff
@@ -0,0 +1,118 @@
...
+
+The first number in the series, the initial term, is defined as ``0`` unless the first number is defined in the ``WHERE`` clause.
```
percona-ysorokin: Please, add the simplest example `SELECT * FROM SEQUENCE_TABLE(n)` first and highlight that the returned results will always start from 0 and will always be less then `n`

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
...
+
+The following example defines the initial term as ``4``.
+
+.. code-block:: mysql
+
+    SELECT value as result FROM SEQUENCE_TABLE(8) AS tt WHERE value >= 4;
```
percona-ysorokin:
```suggestion
    SELECT value AS result FROM SEQUENCE_TABLE(8) AS tt WHERE value >= 4;
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
...
+    +--------+
+    | result |
+    +--------+
+    |      4 |
+    |      5 |
+    |      6 |
+    |      7 |
+    +--------+
+    
+Consecutive terms increase or decrease by a common difference. The default common difference value is ``1`` unless a different value is defined in the ``WHERE`` clause.
```
percona-ysorokin:
```suggestion
Consecutive terms increase or decrease by a common difference. The default common difference value is ``1``. However, it is always possible to filter the results using WHERE clause to simulate common differences greater than 1. 
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
...
+
+The following example defines the common difference as ``2``.
```
percona-ysorokin:
```suggestion
The following example will print only even numbers from the 0..7 range.
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
...
+
+.. code-block:: mysql
+
+    SELECT value as result FROM SEQUENCE_TABLE(8) AS tt WHERE value % 2 = 0;
```
percona-ysorokin:
```suggestion
    SELECT value AS result FROM SEQUENCE_TABLE(8) AS tt WHERE value % 2 = 0;
```

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
...
+    +--------+
+    | result |
+    +--------+
+    |      2 |
```
percona-ysorokin, line 75: The output sequence should be 0,2,4,6

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,118 @@
...
+    .. code-block:: mysql
+
+        SEQUENCE_TABLE(
+            n
+        ) [AS] alias
+    
+``n:`` The number of generated values.
+
+The MySQL syntax requires the ``alias`` in the SELECT statement. The rules to define an alias are defined `here <https://dev.mysql.com/doc/refman/8.0/en/identifiers.html>`_.
+
```
percona-ysorokin, line 36: Also, it needs to be mentioned before the examples section that `SEQUENCE_TABLE()` result set will have a single column with predefined name `value` of type `BIGINT UNSIGNED` that can be referenced in the `SELECT` expressions.
Not only can you do `SELECT * FROM SEQUENCE_TABLE(n) AS tt` but `SELECT <expr(value)> FROM SEQUENCE_TABLE(n) AS tt` as well.

percona-ysorokin (changes requested):

doc/source/flexibility/sequence_table.rst
patrickbirch: updated

doc/source/flexibility/sequence_table.rst
patrickbirch: updated.

doc/source/flexibility/sequence_table.rst
patrickbirch: Updated.

doc/source/flexibility/sequence_table.rst
patrickbirch: Updated.

doc/source/flexibility/sequence_table.rst
patrickbirch: removed.

doc/source/flexibility/sequence_table.rst
patrickbirch: added.

doc/source/flexibility/sequence_table.rst
patrickbirch: added.

doc/source/flexibility/sequence_table.rst
patrickbirch: This block of test is explaining the function. The "SELECT .. FROM statement does not fit in this block. I have added it to the Usage section.

doc/source/flexibility/sequence_table.rst
patrickbirch: Updated.

doc/source/flexibility/sequence_table.rst
patrickbirch: Added.

doc/source/flexibility/sequence_table.rst
patrickbirch: moved and reworded.

doc/source/flexibility/sequence_table.rst
```diff
@@ -0,0 +1,124 @@
...
+    .. code-block:: mysql
+
+        SEQUENCE_TABLE(
+            n
+        ) [AS] alias
+    
+``n:`` The number of generated values.
+
+The MySQL syntax requires the ``alias`` in the ``SELECT`` statement. The rules to define an alias are defined `here <https://dev.mysql.com/doc/refman/8.0/en/identifiers.html>`_.
```
percona-ysorokin:
```suggestion
As any other `derived tables <https://dev.mysql.com/doc/refman/8.0/en/derived-tables.html>`_, all table functions (including ``SEQUENCE_TABLE()``) must be used  with the ``alias`` in the ``SELECT`` statement. The rules to define an alias are defined `here <https://dev.mysql.com/doc/refman/8.0/en/identifiers.html>`_.
```

percona-ysorokin (approved):
LGTM
if the comment regarding `derived tables` addressed

doc/source/flexibility/sequence_table.rst
patrickbirch: Fixed.

PR #3846, opened 2020-05-28
Implement PS-6789 (Merge MySQL 8.0.20)

percona-ysorokin:
https://jira.percona.com/browse/PS-6789


sql/sql_yacc.yy
```diff
@@ -1263,21 +1257,21 @@ void warn_about_deprecated_binary(THD *thd)
...
+%token<lexer.keyword> THREAD_STATS_SYM 1306
+%token<lexer.keyword> USER_STATS_SYM 1307
+%token<lexer.keyword> ENCRYPTION_KEY_ID_SYM 1308
 
 /*
    Tokens from Percona Server 8.0
 */
-%token<lexer.keyword> EFFECTIVE_SYM
-%token  SEQUENCE_TABLE_SYM
+%token<lexer.keyword> EFFECTIVE_SYM 1350
```
kamil-holubicki: This will result with different digest as before it was 1309

kamil-holubicki (commented):
gen_lex_token.cc line 242 range_for_sql_yacc2 object is initialized with title ""sql/sql_yacc.yy (before NOT_A_TOKEN_SYM)". This is actually wrong because it is not before, but after.

Similar situation for line 252 range_for_sql_yacc1 object. The title is "sql/sql_yacc.yy (after NOT_A_TOKEN_SYM)", but symbols are before, not after.

The above is inherited from upstream, but makes a confusion.

sql/sql_yacc.yy
percona-ysorokin: @kamil-holubicki Yes, I agree.
However, this is a sacrifice I am willing to make deliberately to introduce a free range for future 5.7 extensions

percona-ysorokin:
@kamil-holubicki Re `before/after` comments. Yes, I confirm.
Here is a fragment from the generated `lex_token.h`
```
/* 996 */  { "REQUIRE_TABLE_PRIMARY_KEY_CHECK", 31, true, false},
/* 997 */  { "STREAM", 6, true, false},
/* 998 */  { "OFF", 3, true, false},
/* PART 3: padding reserved for sql/sql_yacc.yy extensions. */
/* reserved 999 for sql/sql_yacc.yy (after NOT_A_TOKEN_SYM) */  { "", 0, false, false},
/* PART 4: named tokens from sql/sql_hints.yy. */
/* 1000 */  { "MAX_EXECUTION_TIME", 18, true, false},
/* 1001 */  { "RESOURCE_GROUP", 14, true, false},
/* 1002 */  { "BKA", 3, true, false},
```
Although this is a bit confusing, it has no effect on the binaries - just a comment in a intermediary generated file. I will most probably ignore this.

breisig:
Any update when this will be merged?

percona-ysorokin:
Jenkins Build link (no --big-test)
https://ps80.cd.percona.com/job/percona-server-8.0-param/640/

The following MTR test cases
```
group_replication.gr_ssl_options2
main.ssl_crl
auth_sec.server_withssl_client_withssl
main.ssl_crl_crlpath
main.ssl_ca
main.ssl_crl_clients_valid
```
fail on all platforms because of the Oracle SSL Certificates expired on June 20, 2020 
```
crl-client-cert.cer
crl-client-revoked-cert.cer
crl-server-cert.cer
```

percona-ysorokin:
Jenkins build link (--only-big-test)
https://ps80.cd.percona.com/job/percona-server-8.0-param/641/

```
main.log_buffered-big
```
is proven to be very unstable

PR #3847, opened 2020-05-29
ps-5.7-7061 Percona Server Install Process Just Ends

patrickbirch:
modified apt-repo


kathy-williamson (changes requested):
This does not satisfy my request.  We need to document that the process just ends and when they receive that hanging message the process is complete or the install scripts need to be modified to present a message saying the install is complete (this is the preferred solution).

patrickbirch:
Thanks. Will fix.

patrickbirch:
Install scripts should be modified to add a successful install message.

PR #3850, opened 2020-06-02
PS-7100 Change approach to rows counting in rocksdb_read_free_rpl test

oleksandr-kachan:
Using "select count(column_name)" doesn't reveal an issue with
broken index entries when there is an inconsistency between
tables on primary and secondary servers. Use "select column_name"
instead.


george-lorch (approved):
LGTM

PR #3852, opened 2020-06-02
PS-5666: 100% CPU with innodb_encrypt_threads=1

robgolebiowski:
The issue was that when encryption threads did not find any space to
rotated they keep signalling the crypt_threads_event, which in turn
caused other threads that there might be some work to do. That caused
infinitive loop. The fix is to not set when a thread did not find any
space to rotated.

Apart from running test from ticket PS-5666 I have run encryption test
suite, with and without the fix and I have measured the times:

Without fix:
real    13m48,692s
user    34m39,848s
sys     2m30,826s

With fix:
real    13m57,575s
user    20m34,633s
sys     1m55,404s

### DIFF
--- a/storage/innobase/fil/fil0crypt.cc
+++ b/storage/innobase/fil/fil0crypt.cc
@@ -2101,8 +2101,12 @@ static void fil_crypt_realloc_iops(rotate_thread_t *state) {
 
 /***********************************************************************
 Return allocated iops to global
-@param[in,out]		state		Rotation state */
-static void fil_crypt_return_iops(rotate_thread_t *state) {
+@param[in,out] state                       rotation state
+@param[in]     set_fil_crypt_threads_event should fil_crypt_threads_event be set
+                                           so to notify other events that there
+                                           might be some work to do. */
+static void fil_crypt_return_iops(rotate_thread_t *state,
+                                  bool set_fil_crypt_threads_event = true) {
   if (state->allocated_iops > 0) {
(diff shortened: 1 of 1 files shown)


robgolebiowski:
JR => https://ps80.cd.percona.com/job/percona-server-8.0-param/595 (the failure of test encryption.innodb_default_key is being addressed as PS-7101

satya-bodapati (changes requested):
Can you change "kept setting" to a more meaningful word? May be "they keep signalling the crypt_threads_event, which in turn caused other threads to..."

review in progress...

@satya-bodapati done. I have also checked if there is increase in CPU usage when innodb-encryption-rotate-key-age > 0. There isn't.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -1613,8 +1613,11 @@ static void fil_crypt_realloc_iops(rotate_thread_t *state) {
 
 /***********************************************************************
 Return allocated iops to global
-@param[in,out]		state		Rotation state */
-static void fil_crypt_return_iops(rotate_thread_t *state) {
+@param[in,out]		state		Rotation state
+@param[in]          set_fil_crypt_threads_event should fil_crypt_threads_event be set
+                                                so to notify other events that there
+                                                might be some work to do. */
```
satya-bodapati: Please try to align

satya-bodapati (approved):
LGTM! Merge after addressing align comment

PR #3853, opened 2020-06-02
5.6 Fixed PS-6995 (Optimizer not picking best execution plan for Primary …

altmannmarcelo:
…Key lookup)

https://jira.percona.com/browse/PS-6995

In some cases optimizer prefers to do a Full Table Scan instead of a
PK range scan.
Introduced a new optimizer switch to allow the user to reduce the cost
of a range scan. If favor_range_scan switch is on, the cost for range
plan will be reduced by 90%. Patch has been provided by Yves Trudeau.

### DIFF
--- a/mysql-test/r/index_merge_myisam.result
+++ b/mysql-test/r/index_merge_myisam.result
@@ -1556,19 +1556,19 @@ DROP TABLE t1,t2;
 #
 select @@optimizer_switch;
 @@optimizer_switch
-index_merge=on,index_merge_union=on,index_merge_sort_union=on,index_merge_intersection=on,engine_condition_pushdown=on,index_condition_pushdown=on,mrr=on,mrr_cost_based=on,block_nested_loop=on,batched_key_access=off,materialization=on,semijoin=on,loosescan=on,firstmatch=on,subquery_materialization_cost_based=on,use_index_extensions=on
+index_merge=on,index_merge_union=on,index_merge_sort_union=on,index_merge_intersection=on,engine_condition_pushdown=on,index_condition_pushdown=on,mrr=on,mrr_cost_based=on,block_nested_loop=on,batched_key_access=off,materialization=on,semijoin=on,loosescan=on,firstmatch=on,subquery_materialization_cost_based=on,use_index_extensions=on,favor_range_scan=off
 set optimizer_switch='index_merge=off,index_merge_union=off';
 select @@optimizer_switch;
 @@optimizer_switch
-index_merge=off,index_merge_union=off,index_merge_sort_union=on,index_merge_intersection=on,engine_condition_pushdown=on,index_condition_pushdown=on,mrr=on,mrr_cost_based=on,block_nested_loop=on,batched_key_access=off,materialization=on,semijoin=on,loosescan=on,firstmatch=on,subquery_materialization_cost_based=on,use_index_extensions=on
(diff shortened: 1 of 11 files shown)


altmannmarcelo:
https://ps56.cd.percona.com/job/percona-server-5.6-param/101/

satya-bodapati (commented):
Few doubts/questions
1. Why the FORCE INDEX(PRIMARY) isn't working?
2. Why is the range scan cost produced very high? Can we work to get accurate value instead of manipulating and forcing an artificial lower cost for range scan?
3. For the schema in the bug page, "best_covering_index_scan" strategy wins and that means a FULL TABLE SCAN." is this because the covering index testdata_key_v3` (`testtype`,`gid`,`revision`) has only testtype and gid but the query does filtering on additional field "gid2"?

y-trudeau:
Responses:
   1.  The primary key is used for both plans but the strategy is different.  The wrong plan scans from the beginning, the good plan dives to correct place in the b-tree and read only one record
  2.  If you look at the code, the range scan costs are ridiculously high because the optimizer doesn't consider the "ORDER BY" and LIMIT clause correctly (if at all).  It also considers the 3 ranges separatly and sum the results.  There is obviously a considerable overlap between the ranges.
  3. The best_covering_index_scan is on the PRIMARY KEY, not on testdata_key_v3.  There is a force index (PRIMARY) in the pt-table-checksum query.

satya-bodapati:
Thanks for the answers.

I guess there is a good reason why the optimizer doesn't consider the ORDER BY and LIMIT clause. I don't know the optimizer code well though :( 

1.
So ideal fix would be to change optimizer to consider ORDER BY LIMIT and start with correct costs for range scan? @y-trudeau? 

2.
When and how to enable this optimizer flag? How will pt-checksum enable this? (Will it turn on by default)?

3. With this flag ON, and when there are no overlaps between the ranges, will it degrade the query performance?

Since we don't do intrusive optimizer changes (do the correct and proper fix), selectively turning on this change with a flag (and non-default) is OK

satya-bodapati:
> …Key lookup)
> 
> https://jira.percona.com/browse/PS-6995
> 
> In some cases optimizer prefers do do a FTS instead of a PK range scan.
> Introduced a new optimizer switch to allow the user to reduce the cost
> of a range scan. If favor_range_scan switch is on, the cost for range
> plan will be reduced by 90%. Patch has been provided by Yves Trudeau.

@altmannmarcelo 
FTS also means Full Text Search, Please change it to Full Table Scan :)

altmannmarcelo:
> > …Key lookup)
> > https://jira.percona.com/browse/PS-6995
> > In some cases optimizer prefers do do a FTS instead of a PK range scan.
> > Introduced a new optimizer switch to allow the user to reduce the cost
> > of a range scan. If favor_range_scan switch is on, the cost for range
> > plan will be reduced by 90%. Patch has been provided by Yves Trudeau.
> 
> @altmannmarcelo
> FTS also means Full Text Search, Please change it to Full Table Scan :)

Done

altmannmarcelo:
> > …Key lookup)
> > https://jira.percona.com/browse/PS-6995
> > In some cases optimizer prefers do do a FTS instead of a PK range scan.
> > Introduced a new optimizer switch to allow the user to reduce the cost
> > of a range scan. If favor_range_scan switch is on, the cost for range
> > plan will be reduced by 90%. Patch has been provided by Yves Trudeau.
> 
> @altmannmarcelo
> FTS also means Full Text Search, Please change it to Full Table Scan :)

@satya-bodapati done

y-trudeau:
@satya-bodapati  

I guess the reason for the missing logic in the optimizer is just... never been done or completed.

1. Yes, that would be the real fix

2. pt-tc has a "--set-vars" option that can be used for that.  The option sets session variables in the MySQL connection used by pt-tc.

3. I checked all the pt-tc queries and found nothing wrong or worse with the flag enabled.

satya-bodapati:
thanks @y-trudeau

satya-bodapati (approved):
Ok to merge after clean Jenkins run. Also please test the patch separately on 5.7 & 8.0. There could be more testcases that will need adjustments/fixing.

PR #3294, opened 2019-07-04
8.0 ps 5736

gl-sergei:
No functional changes. Testcase changes:
- CREATE TEMPORARY TABLE t06 (a INT) ENGINE=InnoDB ENCRYPTION='n' fails with different error code
- grep_pattern replaced with search_pattern_in_file

MTR updated to look for plugin.defs in the plugin's mtr directory. It seems that upstream reorganized their test suites layout for commercial plugins and we may want to eventually follow them.

PR #3862, opened 2020-06-05
[PS-7010] Fix crash in BTR_CUR_LATCH_LEAVES

aasosaa:
This is a fix for PS-7010, which is a similar bug to https://bugs.mysql.com/bug.php?id=74596. Like 74596, the problem comes from doing the check before 'page' is latched. This time
in the BTR_MODIFY_PREV/SEARCH_PREV case.  Like the bugfix for 74596 in 777bf2aab05fb58fad7838aed4ef307b50722786, the fix is to move the check below after 'page' is latched.

### DIFF
--- a/storage/innobase/btr/btr0cur.cc
+++ b/storage/innobase/btr/btr0cur.cc
@@ -397,12 +397,6 @@ btr_cur_latch_leaves(
 
 			SRV_CORRUPT_TABLE_CHECK(get_block, return latch_leaves;);
 
-#ifdef UNIV_BTR_DEBUG
-			ut_a(page_is_comp(get_block->frame)
-			     == page_is_comp(page));
-			ut_a(btr_page_get_next(get_block->frame, mtr)
-			     == page_get_page_no(page));
-#endif /* UNIV_BTR_DEBUG */
 		}
 
 		latch_leaves.savepoints[1] = mtr_set_savepoint(mtr);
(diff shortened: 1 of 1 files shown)


aasosaa:
cc @fbordenave

storage/innobase/btr/btr0cur.cc
```diff
@@ -413,6 +407,13 @@ btr_cur_latch_leaves(
 
 		latch_leaves.blocks[1] = get_block;
 #ifdef UNIV_BTR_DEBUG
+		/* Sanity check only after both the blocks are latched. */
+		if (latch_leaves.blocks[0] != NULL) {
+			ut_a(page_is_comp(latch_leaves.blocks[0]->frame)
+			     == page_get_page_no(page));
```
satya-bodapati, line 413: This is handled by @kamil-holubicki via another pull requests #3293 and #3294 

This assertion is wrong btw,  page_is_comp()  == page_is_comp (instead of page_get_page_no())

satya-bodapati (changes requested):
This is handled

satya-bodapati:
Since it is handled internally. Closing this PR

aasosaa:
For future reference it's https://github.com/percona/percona-server/pull/3924 and https://github.com/percona/percona-server/pull/3923 (not #3293 and #3294)

PR #3867, opened 2020-06-09
PS-7063: Vault Server - Setup

borys-belinsky:


doc/source/security/vault.rst
```diff
@@ -5,29 +5,210 @@ Information about HashiCorp Vault
...
+   $ vault server -dev
+
+Run this way, |vault| features automatic authentication and in-memory
+storage. |vault-server| is not suitable for production use in this mode.
+
+Alternatively, you can *deploy* |vault-server| in your environment. For
+deployment, you use *hcl* configuration files (custom format similar to JSON
+developed by |hashicorp-vault|) where you can set values of multiple
+|vault-server| options.
+
```
patrickbirch, line 101: line 97, we mention HCL. Should we point out that the following is an example of HCL?

doc/source/security/vault.rst
```diff
@@ -5,29 +5,210 @@ Information about HashiCorp Vault
...
+.. code-block:: text
```
patrickbirch, line 102: consul is listed in the example, but the explanation is at line 121. So I read the example and then read what consul is and then go back to re-read the example.

doc/source/security/vault.rst
```diff
@@ -5,29 +5,210 @@ Information about HashiCorp Vault
...
+   Unseal Key 2: Tnd+kJEeaO1yOIYTjFPgqSd8/dyUkprZw1zAYa62ISAs
+   Unseal Key 3: I/kdckz2ELaWSeQS/iL+EffsBnbmMVyoI+T1DrFX5ATM
+   Unseal Key 4: iQBHfQPtRqzPQYlBYpCtKeVWrE5C/OxB5O5jZDGYXEgz
+   Unseal Key 5: E5j8VdttAxC+b/6tARAJRybYfyS1AiDqq5D/wJ4QcEbo
+
+   Initial Root Token: s.oM1h6Inq6WOdppd5sML1gmpc
+   ...
+
+The storage is initialized but it is sealed. To be able to store encryption
+keys, the storage must be unsealed. The output of `vault operator init` command
```
patrickbirch, line 173: Explanation after the example.

george-lorch:
This process must be reviewed by @Oneiroi before merging.

doc/source/security/vault.rst
Oneiroi: we should instead link to the hashicorp documentation on configuration HCL for vault,

doc/source/security/vault.rst
Oneiroi: Unsure why we are not also including links to the Hashicorp documentation ?

Oneiroi (changes requested):
Documentation should link to the relevant Hashicorp pages in addition to laying out said documentation to ensure presence (and should the URLs changed).

No functional code observed in this commit, as such this is fine to go once documentation is linked to Hashicorp URL's

@georgelorchpercona over to you

Oneiroi:
RST formatting issues in full files, please check use of backticks as they are not formatting correctly.

Oneiroi (changes requested):
Consul should be configured with encryption on the gossip procotol to ensure that the replication data is encrypted in transit, in addition TLS configuration should also be noted as this gains an additional protection in combination with the encryption on the GOSSIP protocol itself.

Also note that running vault and consul in dev modes, is never acceptable in production, we should draw attention to this being for experimentation purposes *_ONLY_*

Oneiroi:
> Consul should be configured with encryption on the gossip procotol to ensure that the replication data is encrypted in transit, in addition TLS configuration should also be noted as this gains an additional protection in combination with the encryption on the GOSSIP protocol itself.
> 
> Also note that running vault and consul in dev modes, is never acceptable in production, we should draw attention to this being for experimentation purposes __ONLY__

https://www.consul.io/docs/agent/encryption.html
config.hcl add: 

```
...
"key_file": "/etc/consul/ssl/server.key",
  "cert_file": "/etc/consul/ssl/server.crt"
...
```

To ensure TLS configuration.

In Addtion `vault` should also be configured to use TLS: https://www.vaultproject.io/docs/configuration/listener/tcp

e.g. 
```
...
tls_disable = 0
    tls_min_version = "tls12"
    tls_cert_file = "/etc/vault/ssl/server.crt"
    tls_key_file = "/etc/vault/ssl/server.key"
    tls_cipher_suites = "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_256_GCM_SHA384"
    tls_prefer_server_cipher_suites = 1
...
```

This sets the cipherset to hardened standard and the TLS minimum version to TLSv1.2 users can of course roll their own.

doc/source/security/vault.rst
borys-belinsky: The paragraph above the example mentions that it HCL is similar to  JSON. The example illustrates that. We are safe to assume that the reader is aware of JSON (taking into account the context.)

doc/source/security/vault.rst
borys-belinsky: at this point, it is important to focus the attention on the output after vault init. Note the lead-in text for the example: "For each new storage backend, you need to initialize |vault-server|. The
initialized storage is sealed, however." So, the example follows the explanation. after the user runs the initialization command they get some output which is actually a link to the next step.

doc/source/security/vault.rst
borys-belinsky: @Oneiroi Note the seealso block after the example:
  |hashicorp-vault| Documentation:
      - `Deploying Vault server <https://learn.hashicorp.com/vault/getting-started/deploy>`_
      - `Vault server configuration options <https://www.vaultproject.io/docs/configuration>`_
It provides references to the deployment procedure itself and the configuration file and available options

doc/source/security/vault.rst
borys-belinsky: @Oneiroi there links above this context and after it. Introduced by the "seealso" directive, such as the following:
.. seealso::
  |hashicorp-vault| Documentation:
      - `Deploying Vault server <https://learn.hashicorp.com/vault/getting-started/deploy>`_
      - `Vault server configuration options <https://www.vaultproject.io/docs/configuration>`_

doc/source/security/vault.rst
borys-belinsky: The example is not supposed to be functional: it gives an idea of what a configuration file looks like and what it's made of. There are links to relevant Hashicorp documentation that readers should read very carefully when deploying the server.

PR #3870, opened 2020-06-09
PS-7117 Remove list of Supported OS from MyRocks Installation

patrickbirch:
modified myrocks/install
removed list


kathy-williamson (approved):
Is there a reason why the lifecycle list is not linked?

PR #3873, opened 2020-06-10
PS-5730: Change SELECT rotate_system_key to ALTER INSTANCE for percona

robgolebiowski:
system key rotation.

Problem:
--------
SELECT rotate_system_key() is not properly replicated with the default
binlog format ROW. Also, ENCRYPTION_KEY_ADMIN privlege is not checked
for the key rotation.

Fix:
----

Replace rotate_system_key function with statements
ALTER INSTANCE ROTATE * SYSTEM KEY. We introduce two new statements to
rotate percona system key:

1) ALTER INSTANCE ROTATE INNODB SYSTEM KEY key_id

This statement is used in order to rotate percona innodb system key.
key_id - is the id of the tablespace encryption key that
the user wishes to rotate. ALTER INSTANCE ROTATE INNODB SYSTEM KEY statement
is replicated in all binlog formats.

Apart from that, the ALTER INSTANCE checks for sufficient privileges 
(SUPER or ENCRYPTION_KEY_ADMIN) before allowing rotation. 
Also ALTER INSTANCE takes backup logs to block concurrent backup
and concurrent DDL.

2) ALTER INSTANCE ROTATE REDO SYSTEM KEY

Either SUPER or ENCRYPTION_KEY_ADMIN privileges are required to issue
this statement. The same as binlog key rotation, redo log rotation is
not replicated. The redo log rotation does not need any extra
synchronization, as it will be synchronized by the keyring framework.


robgolebiowski:
JR=>https://ps80.cd.percona.com/job/percona-server-8.0-param/603/

satya-bodapati:
Looks good. Few requests

1. Update commit message with the description from GitHub(I've reformatted a bit)
2. Introduce a test case which shows rotation fails without encryption_admin_privlege and passes with the admin privilege (non-root user)
3. Is it possible to add a test case that shows a concurrent DDL is blocked during the process of key rotation and resumed once the key rotation is over

Question:
Isn't there bin log key rotation that is keyring key specific?

satya-bodapati (changes requested):
changes noted as separate comment

robgolebiowski:
Hey @satya-bodapati:
Re 2) this is already checked by mysql-test/t/percona_rotate_system_key_check_privileges.test, right ?

robgolebiowski:
Re Question:
There isn't. There was in 5.7, but it was never GAed. Now binlog encryption is upstream implementation.

robgolebiowski:
3) As discussed on slack, skipping this. Backup's MDL locks are locked as TRANSACTION and as such are released on commit/rollback.

robgolebiowski:
I have updated the commit message.

satya-bodapati (approved):
LGTM! Please get approval from Yura for the SYMBOL/parser changes.

robgolebiowski:
hey @percona-ysorokin - could you please check for the SYMBOL/parser changes?

sql/sql_yacc.yy
```diff
@@ -14673,6 +14674,7 @@ ident_keywords_unambiguous:
         | INDEX_STATS_SYM
         | INDEXES
         | INITIAL_SIZE_SYM
+        | INNODB_SYM
```
percona-ysorokin: Why do we need to register `INNODB_SYM` as an `ident_keywords_unambiguous` here.
In original MySQL 8.0 syntax
https://dev.mysql.com/doc/refman/8.0/en/alter-instance.html
`ALTER INSTANCE ROTATE INNODB MASTER KEY` was working without this

sql/sql_yacc.yy
```diff
@@ -1259,6 +1259,7 @@ void warn_about_deprecated_binary(THD *thd)
    Tokens from Percona Server 8.0
 */
 %token<lexer.keyword> EFFECTIVE_SYM
+%token<lexer.keyword> INNODB_SYM                    /* MYSQL */
```
percona-ysorokin: Probably, remove, see below

sql/sql_yacc.yy
robgolebiowski: ALTER INSTANCE ROTATE INNODB MASTER KEY is working without this because INNODB is a variable here - it can be either INNODB or BINLOG. Rest of the words in this command are defined as symbols. In our case we only allow ALTER INSTANCE ROTATE INNODB SYSTEM KEY - so there is no reason why not make an INNODB defined as a symbol.

sql/sql_yacc.yy
```diff
@@ -15445,19 +15447,52 @@ alter_instance_stmt:
           ALTER INSTANCE_SYM alter_instance_action
           {
             Lex->sql_command= SQLCOM_ALTER_INSTANCE;
-            $$= NEW_PTN PT_alter_instance($3);
+            $$= NEW_PTN PT_alter_instance($3.alter_instance_action, $3.key_id);
           }
 
 alter_instance_action:
```
percona-ysorokin, line 15451: Personally, I would reorganize the following grammar rule
```
alter_instance_action:
          ROTATE_SYM ident_or_text MASTER_SYM KEY_SYM
          | ROTATE_SYM ident_or_text SYSTEM_SYM KEY_SYM ulong_num
          | ROTATE_SYM ident_or_text SYSTEM_SYM KEY_SYM
          | RELOAD ident
          | RELOAD ident NO_SYM ROLLBACK_SYM ON_SYM ERROR_SYM
          ;
```
to something more unambiguous
```
alter_instance_action:
          ROTATE_SYM ident_or_text master_or_system_key_spec
          | RELOAD ident
          | RELOAD ident NO_SYM ROLLBACK_SYM ON_SYM ERROR_SYM
          ;
master_or_system_key_spec:
          MASTER_SYM KEY_SYM
          | SYSTEM_SYM KEY_SYM opt_key_id_spec
          ;
opt_key_id_spec:
          ulong_num
          |
          ;
```

robgolebiowski: To be honest, I do not like it. You have to really read in - to know when there is system key rotation and when is master key rotation. You really do not see from this that system key can be only rotated from InnoDB. I think it just complicates simple think without strong reason to.

percona-ysorokin: OK, @robgolebiowski 
Your original solution is probably better from the "merging with upstream" point of view - it just adds a few more alternatives to an existing rule. So I am fine with it as long as it does not produce any new `bison` warnings.

sql/sql_yacc.yy
percona-ysorokin: @robgolebiowski I am quiet confused here - I do not see `INNODB_SYM` appearing in the new alternatives you added.
Did you want to add
```
          | ROTATE_SYM INNODB_SYM SYSTEM_SYM KEY_SYM ulong_num
          | ROTATE_SYM INNODB_SYM SYSTEM_SYM KEY_SYM
```
instead of
```
          | ROTATE_SYM ident_or_text SYSTEM_SYM KEY_SYM ulong_num
          | ROTATE_SYM ident_or_text SYSTEM_SYM KEY_SYM
```
?
Moreover, are we potentially considering adding support for `ALTER INSTANCE ROTATE BINLOG SYSTEM KEY` construct in future?

percona-ysorokin (approved):
LGTM

PR #3874, opened 2020-06-10
PS-5978: Handle --numa-interleave flag properly in mysqld_safe script

Sudokamikaze:
Removed unneeded check of variable and moved to top into case() section, in order to prevent flag exposure as mysqld flag(which is incorrect)


EvgeniyPatlan (approved):

EvgeniyPatlan:
@vorsel please review.

vorsel (approved):

PR #3876, opened 2020-06-11
PS-1133 Update Multiple page asynchronous I/O requests doc

patrickbirch:
Merged with 5.6 into 5.7
performance/aio_page_requests updated with approved 5.6 text.


kathy-williamson (changes requested):
change if the server doing read ahead. to if the server is doing read ahead.

patrickbirch:
Updated.

nastena1606 (approved):
LGTM

PR #3877, opened 2020-06-12
PS-6987: encryption threads report space unencrypted while there are

robgolebiowski:
some pages still encrypted in it.

Encryption threads can process one space in parallel. The space is divided
between threads, for instance thread 1 processes pages 1-100, thread 2
pages 101 - 200 and thread 3 processes pages 201 - 300. Let's say
that threads are encrypting the space, threads 2 and 3 had finished, but
thread 1 is still encrypting. Now user changes the
default_table_encryption to ONLINE_FROM_KEYRING_TO_UNENCRYPTED. Threads
2 and 3 find that pages 101 - 300 are not yet being processed and start
unencrypting space. In meantime thread 1 finishes off encrypting pages
1 - 100. When threads 2 and 3 finish they will find out that there are
no more pages to process and mark the space as unencrypted.

The ability to change the value of default_table_encryption while
encryption threads are on opens way to many potential bugs of this kind.
Thus as a fix - it is now possible to change the value of
default_table_encryption only when encryption_threads are off.


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/605
The failing test is due to PS-7101, for which I will merge the fix as soon as trunk is open for merges.

satya-bodapati (changes requested):
Can you add check in default_table_encryption update function() that encrypt_threads should be 0 ?

Also, when threads 2 and 3 finish, wouldn't they see the there is one more thread in progress? If they could, they can avoid start un-encrypting until there are no active threads?

I assume this approach is taken for simplicity purposes.

@satya-bodapati innodb_encryption_threads is an innodb variable. I do not have access to it from server side, unless I implement a handler, which seems like bit too much for assertion. 
About the threads - thread 2 and 3 do not know what the other thread is doing - currently they assume (in this scenario) that thread 1 is decrypting, while it is encrypting. Some extra logic would need to be added - to check that thread 1 is doing what thread 2 and 3 intend to do. Also there is scenario of re-encryption when encryption key gets rotated. Waiting may also open some way to thread starvation in this tangled design. So, yes, I went with the simpler solution.

share/messages_to_clients.txt
```diff
@@ -9361,6 +9361,9 @@ ER_EXPLICIT_DECRYPTION_OF_ONLINE_ENCRYPTED_TABLESPACE
 ER_SEQUENCE_TABLE_SIZE_LIMIT
   eng "The number of records generated by SEQUENCE_TABLE() cannot exceed %llu (%llu requested). Try increasing @@tf_sequence_table_max_upper_bound to a larger value."
 
+ER_D_T_E_CHANGE_WHILE_ENCRYPTION_THREADS_NOT_ZERO
```
satya-bodapati: IMHO, this message tries to explain the entire situation. Can we be precise and say ER_DTE_ENCRYPTION_THREADS_ACTIVE

satya-bodapati: Do we have test case that shows this error?

share/messages_to_clients.txt
```diff
@@ -9361,6 +9361,9 @@ ER_EXPLICIT_DECRYPTION_OF_ONLINE_ENCRYPTED_TABLESPACE
...
+  eng "In order to change default_table_encryption first disable encryption threads. To do so, please set number of innodb_encryption_threads to 0."
```
satya-bodapati:
```suggestion
  eng "In order to change default_table_encryption, disable encryption threads. Please set number of innodb_encryption_threads to 0."
```

storage/innobase/fil/fil0crypt.cc
```diff
@@ -3051,9 +3052,16 @@ void fil_crypt_set_rotation_iops(uint val) {
 Adjust encrypt tables
 @param[in]	val		New setting for innodb-encrypt-tables */
 
-void fil_crypt_set_encrypt_tables(enum_default_table_encryption val) {
+bool fil_crypt_set_encrypt_tables(enum_default_table_encryption val,
+                                  bool is_server_starting) {
+  // It is always OK to set default_table_encryption on server
+  // startup
+  if (!is_server_starting && (srv_n_fil_crypt_threads_requested != 0)) {
```
satya-bodapati, line 3727: do we need is_server_starting? Is this function called on startup? Wouldn't the check in update() function of the variable be sufficient?

satya-bodapati (changes requested):

satya-bodapati:
> https://ps80.cd.percona.com/job/percona-server-8.0-param/605
> The failing test is due to PS-7101, for which I will merge the fix as soon as trunk is open for merges.

Do you mean this test failure?  encryption.innodb_default_key ?

@satya-bodapati yes

satya-bodapati:
> some pages still encrypted in it.
> 
> Encryption threads can process one space in parallel. The space is divieded
> between threads, for instance thread 1 processes pages 1-100, thread 2
> pages 101 - 200 and thread 3 processes pages 201 - 300. Let's say
> that threads are encrypting the space, threads 2 and 3 had finished, but
> thread 1 is still encrypting. Now user changes the
> default_table_encryption to ONLINE_FROM_KEYRING_TO_UNENCRYPTED. Threads
> 2 and 3 find that pages 101 - 300 are not yet being processed and start
> unencrypting space. In meantime thread 1 finishes off encrypting pages
> 1 - 100. When threads 2 and 3 finish they will find out that there are
> no more pages to process and mark the space as unencrypted.
> 
> The ability to change the value of default_table_encryption while
> encryption threads are on opens way to many potential bugs of this kind.
> Thus as a fix - it is now possible to change the value of
> default_table_encryption only when encryption_threads are off.

typo: "divieded" -> "divided"

@satya-bodapati done

share/messages_to_clients.txt
robgolebiowski: done

robgolebiowski: I have extended the encryption.default_table_encryption_var test.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: Yes, it is called on startup via handler - fix_default_table_encryption. We set srv_default_table_encryption only when srv_uuid became available. In this function we check that we can change default_table_encryption (i.e. number of encryption threads is equal to 0).

share/messages_to_clients.txt
robgolebiowski: done

satya-bodapati (approved):

robgolebiowski:
JR after rebase => https://ps80.cd.percona.com/job/percona-server-8.0-param/729

PR #3879, opened 2020-06-16
PS-7119: Tests encryption.innodb_encryption_aborted_rotation* sometime

robgolebiowski:
fail.

This set of tests enforce crash of the server between updating DD with
encryption flag and write to page0 (by encryption threads). Sometimes
the redo record for DD update stays in the redo log and sometimes it
doesn't. In case it doesn't and update to DD is written to disk before
the crash - after the restart mismatch between DD flag and flag in page0
will be visible in the error log. To get rid of this error, the server
is restarted after the crash with innodb-encryption-threads=1.

### DIFF
--- a/mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation.result
+++ b/mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation.result
@@ -82,8 +82,10 @@ SET GLOBAL default_table_encryption=ONLINE_FROM_KEYRING_TO_UNENCRYPTED;
 SET GLOBAL innodb_encryption_threads = 4;
 # Make sure that t1 has been decrypted
 # First restart the server after crash so any redo logs for t1 were proceed. If there are any logs available
-# t1 will be opened without validating idb file to DD.
-# restart:--innodb-encryption-rotate-key-age=15 --innodb-encryption-threads=0
+# t1 will be opened without validating idb file to DD. We set here innodb-encryption-threads to value > 0 - so
+# in case there is no redo for t1 and validation to DD would occur we would not get an error in the log
+# to set innodb-encryption-threads to value > 0 to finish rotation. We grep for this error later.
+# restart:--innodb-encryption-rotate-key-age=15 --innodb-encryption-threads=1
 SET GLOBAL debug="-d,crash_on_t1_flush_after_dd_update";
 # Now restart the server and select from t1. Validation of t1's flags should proceed.
 # It should be discovered that there is mismatch of flags in t1. However it shoud be possible to use
@@ -104,8 +106,10 @@ COUNT(1)
 SET GLOBAL debug="+d,crash_on_t1_flush_after_dd_update";
 SET GLOBAL innodb_encryption_threads=4;
 # First restart the server after crash so any redo logs for t1 were proceed. If there are any logs available
-# t1 will be opened without validating idb file to DD.
-# restart:--innodb-encryption-rotate-key-age=15 --innodb-encryption-threads=0
+# t1 will be opened without validating idb file to DD. We set here innodb-encryption-threads to value > 0 - so
+# in case there is no redo for t1 and validation to DD would occur we would not get an error in the log
+# to set innodb-encryption-threads to value > 0 to finish rotation. We grep for this error later.
+# restart:--innodb-encryption-rotate-key-age=15 --innodb-encryption-threads=1
 SET GLOBAL debug="-d,crash_on_t1_flush_after_dd_update";
 # Make sure that t1 has been encrypted
 # Now start the server and select from t1. Validation of t1's flags should proceed.
--- a/mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation_mk.result
+++ b/mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation_mk.result
@@ -82,8 +82,10 @@ SET GLOBAL default_table_encryption=ONLINE_FROM_KEYRING_TO_UNENCRYPTED;
 SET GLOBAL innodb_encryption_threads = 4;
 # Make sure that t1 has been decrypted
 # First restart the server after crash so any redo logs for t1 were proceed. If there are any logs available
-# t1 will be opened without validating idb file to DD.
-# restart:--innodb-encryption-rotate-key-age=15 --innodb-encryption-threads=0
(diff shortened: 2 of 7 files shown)


robgolebiowski:
JR=>https://ps80.cd.percona.com/job/percona-server-8.0-param/609/

mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation.result
```diff
@@ -76,8 +76,10 @@ SET GLOBAL debug="+d,crash_on_t1_flush_after_dd_update";
 SET GLOBAL default_table_encryption=ONLINE_FROM_KEYRING_TO_UNENCRYPTED;
 # Make sure that t1 has been decrypted
 # First restart the server after crash so any redo logs for t1 were proceed. If there are any logs available
-# t1 will be opened without validating idb file to DD.
-# restart:--innodb-encryption-rotate-key-age=15 --innodb-encryption-threads=0
+# t1 will be opened without validating idb file to DD. We set here innodb-encryption-threads to value > 0 - so
+# in case there is no redo for t1 and validation to DD would occur we would not get an error in the log
+# to set innodb-encryption-threads to value > 0 to finish rotation. We grep for this error later.
+# restart:--innodb-encryption-rotate-key-age=15 --innodb-encryption-threads=1
```
satya-bodapati, line 88: Can you help with the code for below
1. the place where the error message came from "In file './test/t1.ibd' (tablespace id = 4) encryption flag is ON. However the encryption flag in the data dictionary is OFF. This indicates that the rotation of the table was interrupted before space's flags were updated. Please have encryption_thread variable (innodb-encryption-threads) set to value > 0. So the encryption could finish up the rotation."

2. How enabling encryption threads suppress this message?

3. Where is recheck done?

satya-bodapati (changes requested):

mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation.result
```diff
@@ -76,8 +76,10 @@ SET GLOBAL debug="+d,crash_on_t1_flush_after_dd_update";
 SET GLOBAL default_table_encryption=ONLINE_FROM_KEYRING_TO_UNENCRYPTED;
 # Make sure that t1 has been decrypted
 # First restart the server after crash so any redo logs for t1 were proceed. If there are any logs available
-# t1 will be opened without validating idb file to DD.
-# restart:--innodb-encryption-rotate-key-age=15 --innodb-encryption-threads=0
+# t1 will be opened without validating idb file to DD. We set here innodb-encryption-threads to value > 0 - so
+# in case there is no redo for t1 and validation to DD would occur we would not get an error in the log
+# to set innodb-encryption-threads to value > 0 to finish rotation. We grep for this error later.
```
satya-bodapati, line 87: Also see if you want to use use log_buffer_flush_to_disk(); just before the crash. This will ensure the redo is always on disk for the test scenario

satya-bodapati (changes requested):

mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation.result
satya-bodapati: Ok, it looks like we want the update to DD reach the actual data page on mysql.ibd, not just redo? correct?

mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation.result
robgolebiowski: 2. Enabling encryption threads on restart indicate that they will still have a chance to finish up the encryption rotation. Since the page0 has not yet been updated - encryption threads will take this tablespace and finish up the encryption (updating page0 at the end).
3. The recheck if the encryption flag from page0 matches the encryption flag in DD is done after the restart of the server (after the crash). The recheck if tablespace needs encryption is done by encryption threads (when they are enabled).

mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation.result
robgolebiowski: Yes

mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation.result
satya-bodapati: Ok, I meant the exact location in source code. with some minor effort, it is in Fil_shard::get_file_size().

mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation.result
satya-bodapati: Ok, then  buf_LRU_flush_or_remove_pages(space_id, BUF_REMOVE_FLUSH_WRITE, nullptr,
                                      false);  before crash could work.
What do you think?

satya-bodapati (changes requested):
check if one suggestion is possible to implement (as part of the comment)

mysql-test/suite/encryption/r/innodb_encryption_aborted_rotation.result
robgolebiowski: Yes, I think it should work. I just think that if it possible to obtain the same in the MTR test, it is better not to add extra code with some debug statement. If you do not mind, I would like to keep it this way.

satya-bodapati: For consistency reasons, it is perfectly fine to add code before a debug crash. See code like this

/* Test what happens on crash here.
      The data dictionary transaction should be
      rolled back, restoring the old table. */
      DBUG_EXECUTE_IF("innodb_alter_commit_crash_before_commit",
                      log_buffer_flush_to_disk();
                      DBUG_SUICIDE(););
      ut_ad(!trx->fts_trx);
  
      DBUG_EXECUTE_IF("innodb_alter_commit_crash_after_commit",
                      log_make_latest_checkpoint();
                      log_buffer_flush_to_disk(); DBUG_SUICIDE(););

 DBUG_EXECUTE_IF("row_mysql_crash_if_error", {
          log_buffer_flush_to_disk();
          DBUG_SUICIDE();
        });

For this time, ok to skip..

satya-bodapati (approved):

robgolebiowski:
JR => https://ps80.cd.percona.com/job/percona-server-8.0-param/731/

PR #3883, opened 2020-06-18
PS-5741: Incorrect use of memset_s in keyring_vault.

robgolebiowski:
Fixed the usage of memset_s. The arguments should be:
void memset_s(void *dest, size_t dest_max, int c, size_t n)
where the 2nd argument is size of buffer and the 3rd is
argument is character to fill.


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/615/

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/730/ self approved

PR #3885, opened 2020-06-18
PS-7131: Fix unstable mysql_ts_alter_encrypt_X and tablespace_encrypt_X tests

oleksandr-kachan:
A situation may arise when a signal from (un)encryption process in
background thread can arrive earlier than we start waiting on a
conditional variable and mysql freezes


percona-ysorokin (approved):
LGTM

PR #3888, opened 2020-06-22
PS-7133: keyring_vault unit tests fail on Jenkins.

robgolebiowski:
Keyring Vault unittests started to fail when ctest started to run tests
in parallel. Each vault test suite created a mount point called <<server_uuid>> in
Vault server and created/removed keys in this mount point. When the
vault server test suites started to be executed in parallel all of
sudden they all started to use the same mount point, causing the tests
to fail. I have modified the test suites so each would create a
separate mount point called <<server_uuid_test_suite_name>>. Also the
configuration file generated to connect to Vault server is different per
each test suite.

### DIFF
--- a/unittest/gunit/keyring_vault/vault_io-t.cc
+++ b/unittest/gunit/keyring_vault/vault_io-t.cc
@@ -26,7 +26,7 @@ using ::testing::Return;
 using ::testing::SetArgPointee;
 using ::testing::StrEq;
 
-std::string credential_file_url = "./keyring_vault.conf";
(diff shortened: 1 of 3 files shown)


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/619/

percona-ysorokin:
@robgolebiowski Looks like a wrong Jenkins link

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/648/

PR #3890, opened 2020-06-22
5.7 Implemented PS-7114 (Enhance crash artifacts for operator)

altmannmarcelo:
https://jira.percona.com/browse/PS-7114

Added a new submodule for coredumper library
Added a new build flag (WITH_COREDUMPER) that allows disabling
compiling coredumper library. Its ON by default.
Adjusted signal handler code to switch between kernel coredump and
libcoredumper.
Added Server version as part of stack trace.
Added buildID as part of stack trace.


altmannmarcelo:
https://ps57.cd.percona.com/job/percona-server-5.7-param/288/

cmake/coredumper.cmake
```diff
@@ -0,0 +1,14 @@
+MACRO(MYSQL_CHECK_COREDUMPER)
+  # We build coredumper by default
+  IF (NOT DEFINED WITH_COREDUMPER)
+    SET (HAVE_LIBCOREDUMPER 1)
```
dutow: This will still result in the "else" clause of the below if, building ps without coredumper? I think this should set WITH_COREDUMPER to 1.

config.h.cmake
```diff
@@ -462,4 +462,6 @@
 /* For default value of --early_plugin_load */
 #cmakedefine DEFAULT_EARLY_PLUGIN_LOAD @DEFAULT_EARLY_PLUGIN_LOAD@
 
+/* coredumper library */
+#cmakedefine HAVE_LIBCOREDUMPER 1
```
dutow: This will define HAVE_LIBCOREDUMPER to 1 if HAVE_LIBCOREDUMPER is defined, regardless of its value. As it is always defined, it will always be defined.

A better approach is to use `cmakedefine01`, which defines it to 0 or 1 based on the cmake bool values. And then the later code should use `#if HAVE_LIBCOREDUMPER` instead of ifdef.

mysys/stacktrace.c
```diff
@@ -230,6 +243,63 @@ static void my_demangle_symbols(char **addrs, int n)
...
+  char        mysqld_path[PATH_MAX + 1]= {0};
+  FILE *      mysqld;
+  struct stat statbuf;
+  ElfW(Ehdr) *ehdr= 0;
+  ElfW(Phdr) *phdr= 0;
+  ElfW(Nhdr) *nhdr= 0;
+  unsigned char *build_id;
+  char           buff[3];
+  uint           i;
+  sprintf(proc_path, "/proc/%d/exe", getpid());
```
dutow: snprintf

mysys/stacktrace.c
```diff
@@ -230,6 +243,63 @@ static void my_demangle_symbols(char **addrs, int n)
...
+  readlink_bytes= readlink(proc_path, mysqld_path, sizeof(mysqld_path));
+  if (readlink_bytes < 1)
+  {
+    my_safe_printf_stderr("Error reading process path\n");
+    return;
+  }
+  if (!(mysqld= fopen(mysqld_path, "r")))
```
dutow, line 269: I don't see a matching fclose?

sql/mysqld.cc
```diff
@@ -4421,6 +4424,83 @@ a file name for --log-bin-index option", opt_binlog_index_name);
...
+        unireg_abort(MYSQLD_ABORT_EXIT);
+      }
+      if (libcoredumper_file_length > FN_LEN)
+      {  // filename is too long
+        sql_print_error("Variable --coredumper set to a too long filename");
+        unireg_abort(MYSQLD_ABORT_EXIT);
+      }
+      if (my_access(libcoredumper_dir, F_OK))
+      {
+        sql_print_error(
```
dutow: What if it gets removed while the server is running?

sql/mysqld.cc
```diff
@@ -4421,6 +4424,83 @@ a file name for --log-bin-index option", opt_binlog_index_name);
...
+            "Directory specified at --coredumper: %s does not exist",
+            libcoredumper_dir);
+        unireg_abort(MYSQLD_ABORT_EXIT);
+      }
+      if (my_access(libcoredumper_dir, (F_OK | W_OK)))
+      {
+        sql_print_error(
+            "Directory specified at --coredumper: %s is not writable",
```
dutow: Same question. The disk could also become full in the meantime.

sql/mysqld.cc
```diff
@@ -4421,6 +4424,83 @@ a file name for --log-bin-index option", opt_binlog_index_name);
...
+      sql_print_warning(
+          "Started with --core-file and --coredumper. "
+          "--coredumper will take precedence.");
+    }
+    if (opt_libcoredumper_path != NULL)
+    {
+      /* validate path */
+      if (!is_valid_log_name(opt_libcoredumper_path,
+                             strlen(opt_libcoredumper_path)))
+      {  //filename contain .cnf or .ini on it
```
dutow: I'm not sure if I understand this condition - why sepcifically ini file?

sql/mysqld.cc
```diff
@@ -4421,6 +4424,83 @@ a file name for --log-bin-index option", opt_binlog_index_name);
...
+      }
+      size_t libcoredumper_file_length=
+          opt_libcoredumper_path_length - libcoredumper_dir_length;
+      if (libcoredumper_file_length == 0)  //path is a directory
+      {
+        libcoredumper_file_length= 19;  //file is set to core.yyyymmddhhmmss
+      }
+      else
+      {
+        libcoredumper_file_length+= 17;  //file gets .yyyymmddhhmmss appended
```
dutow: Why 17?

sql/mysqld.cc
```diff
@@ -4421,6 +4424,83 @@ a file name for --log-bin-index option", opt_binlog_index_name);
...
+                             strlen(opt_libcoredumper_path)))
+      {  //filename contain .cnf or .ini on it
+        sql_print_error("Variable --coredumper cannot be set to value %s",
+                        opt_libcoredumper_path);
+        unireg_abort(MYSQLD_ABORT_EXIT);
+      }
+      char   libcoredumper_dir[FN_REFLEN];
+      size_t libcoredumper_dir_length;
+      size_t opt_libcoredumper_path_length= strlen(opt_libcoredumper_path);
+      (void)dirname_part(libcoredumper_dir, opt_libcoredumper_path,
```
dutow: This works based on the filename, not based on what's actually on the filesystem. I'm not sure if this will be correct if the user specifies a directory without a "/" at the end - I tihink we'll create core files outside of that directory with this logic?

cmake/coredumper.cmake
altmannmarcelo: It will, but the way you suggested make it easier to understand that if you do not set, the default will be WITH_COREDUMPER

config.h.cmake
altmannmarcelo: Done

sql/mysqld.cc
altmannmarcelo: This feature tries to mimic binlog, you can ommit the path, pass a directory or a full path with file name. Here if for example the user sets `--coredumper=/var/lib/mysql` mysql will try to write a file on `var/lib` as the last part is recognized as file name(in this case it will be `/var/lib/mysql.yyyymmddhhmmss` ), however if they specify `--coredumper=/var/lib/mysql/` it will write a file under `/var/lib/mysql/` in this case,  `/var/lib/mysql/core.yyyymmddhhmmss`

sql/mysqld.cc
altmannmarcelo: adjusted, it should be 15

sql/mysqld.cc
altmannmarcelo: This is just to make sure we are not using something that can be interpreted as config file. is_valid_log_name function test it for linux(.cnf) and windows(.ini)

mysys/stacktrace.c
altmannmarcelo: :facepalm:  fixed

mysys/stacktrace.c
altmannmarcelo: fixed

sql/mysqld.cc
altmannmarcelo: in that case there will an error been reported saying it couldn't write the file:

```
bin/mysqld(my_print_stacktrace+0x4b)[0x56068b4cd6a1]
bin/mysqld(handle_fatal_signal+0x44a)[0x56068aa0d177]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0)[0x7f7dff02c3c0]
/lib/x86_64-linux-gnu/libc.so.6(__poll+0x4f)[0x7f7dfef3896f]
bin/mysqld(_ZN22Mysqld_socket_listener27listen_for_connection_eventEv+0x5b)[0x56068b2a5cbb]
bin/mysqld(_ZN19Connection_acceptorI22Mysqld_socket_listenerE21connection_event_loopEv+0x39)[0x560ca3]
bin/mysqld(_Z11mysqld_mainiPPc+0x156d)[0x56068a9fc4ca]
bin/mysqld(main+0x24)[0x56068a9f220d]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7f7dfee4a0b3]
bin/mysqld(_start+0x2e)[0x56068a9f212e]
You may download the Percona Server operations manual by visiting
http://www.percona.com/software/percona-server/. You may find information
in the manual which will help you identify the cause of the crash.
Writing a core file using lib coredumper
PATH: /mnt/ramdisk/core

Error writting coredump: -1 Signal: 11
```

sql/mysqld.cc
altmannmarcelo: Done, I now validate the path again, in case of any error we revert back to have coredump written on datadir

dutow (approved):

PR #3891, opened 2020-06-23
PS-7127: Utility user cannot update password of the user having SYSTE…

kamil-holubicki:
…M_USER privilege.

https://jira.percona.com/browse/PS-7127

1. Added handling of dynamic privileges for utility user
2. Added missing allowed static privileges: CREATE_ROLE, DROP_ROLE

### DIFF
--- a/sql/auth/sql_auth_cache.cc
+++ b/sql/auth/sql_auth_cache.cc
@@ -88,6 +88,7 @@
 
 #include <algorithm>
 #include <functional>
+#include <regex>
 #include <unordered_map>
 #include <utility>
 #include <vector>
@@ -170,6 +171,7 @@ ACL_USER acl_utility_user;
 static LEX_STRING acl_utility_user_name, acl_utility_user_host_name;
 static bool acl_utility_user_initialized = false;
 static std::vector<std::string> acl_utility_user_schema_access;
+static std::vector<std::string> uu_dynamic_privileges;
 
 static void acl_free_utility_user();
 
@@ -1313,6 +1315,42 @@ exit:
   return db_access & host_access;
 }
 
+/*
+ Parse utility-user-dynamic-privileges parameter.
+*/
+static bool setup_utility_user_dynamic_privileges() {
+  if (!utility_user_dynamic_privileges) {
+    // no dynamic privileges provided
+    return false;
+  }
+
+  std::string privileges(utility_user_dynamic_privileges);
+
+  // check if the string is valid
+  std::regex validate_regex("(\\s*[A-Za-z0-9_]+\\s*,*)*\\s*[A-Za-z0-9_]+\\s*",
+                            std::regex::nosubs);
+  if (!std::regex_match(privileges, validate_regex)) {
+    sql_print_error(
+        "Wrong format of --utility-user-dynamic-privileges parameter value: %s",
+        utility_user_dynamic_privileges);
+    return true;
+  }
+
+  // split it into tokens and trim
+  std::regex split_regex("[^,\\s][^,]*[^,\\s]");
+  for (auto it = std::sregex_iterator(privileges.begin(), privileges.end(),
+                                      split_regex);
+       it != std::sregex_iterator(); ++it) {
+    // Unfortunately there is a bug related to std::regex_constants::icase
+    // https://stackoverflow.com/questions/37757863/c11-regexicase-inconsistent-behavior
+    auto s = it->str();
+    std::for_each(s.begin(), s.end(), [](char &c) { c = ::toupper(c); });
+    uu_dynamic_privileges.push_back(s);
+  }
+
+  return false;
+}
+
 int caching_sha2_password_generate(char *outbuf, unsigned int *buflen,
                                    const char *inbuf, unsigned int inbuflen);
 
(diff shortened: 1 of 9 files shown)


george-lorch (approved):

sql/auth/sql_auth_cache.cc
```diff
@@ -1501,6 +1557,24 @@ static void acl_free_utility_user() {
...
+  }
+
+  // std::regex (..., std::regex_constants::icase) would fit here perfectly,
+  // but unfortunately there is a bug related to std::regex_constants::icase
+  // https://stackoverflow.com/questions/37757863/c11-regexicase-inconsistent-behavior
+  auto tmp_str = privilege;
+  std::for_each(tmp_str.begin(), tmp_str.end(),
+                [](char &c) { c = ::toupper(c); });
+
+  auto it = std::find_if(
```
percona-ysorokin: Why not just `std::find(uu_dynamic_privileges.begin(), uu_dynamic_privileges.end(), tmp_str)`

sql/auth/sql_security_ctx.cc
```diff
@@ -624,6 +624,15 @@ std::pair<bool, bool> Security_context::has_global_grant(const char *priv,
 
   std::string privilege(priv, priv_len);
 
+  if (acl_utility_user.user) {
+    std::string pu(m_priv_user, m_priv_user_length);
+    std::string ph(m_priv_host, m_priv_host_length);
+    if (acl_is_utility_user(pu.c_str(), ph.c_str(), nullptr)) {
```
percona-ysorokin: Here, converting to `std::string` just to use `c_str()` then
```suggestion
    if (acl_is_utility_user(m_priv_user, m_priv_host, nullptr)) {
```

sql/auth/sql_auth_cache.cc
```diff
@@ -1309,6 +1311,42 @@ ulong acl_get(THD *thd, const char *host, const char *ip, const char *user,
...
+  if (!utility_user_dynamic_privileges) {
+    // no dynamic privileges provided
+    return false;
+  }
+
+  std::string privileges(utility_user_dynamic_privileges);
+
+  // check if the string is valid
+  std::regex validate_regex(
+      "^(\\s*[A-Za-z0-9_]+\\s*,*)*\\s*[A-Za-z0-9_]+\\s*$");
```
percona-ysorokin: Use either `nosubs` regex option or `(?:<expr>)` instead of `(expr)` as you do not need to store results of subexpressions.
`^` and `$` are also redundant as you are using `std::regex_match()` here

percona-ysorokin: Also, is it allowed for a privilege to start with a digit?

percona-ysorokin: Moreover, the regex seems to be overcomplicated and I am not sure it covers all the edge cases.
I suggest to do a 2 step checking:
1. Eliminate all whitespace characters from the input string.
2. Create a simpler regex for the list (without `\\s*`)

sql/auth/sql_auth_cache.cc
kamil-holubicki: Are you sure that what we get as the parameter will always be uppercase? This is dynamic privilege, potentially unknown.

sql/auth/sql_security_ctx.cc
kamil-holubicki: You are right. I didn't notice that m_priv_user and m_priv_host are assigned only by assign_priv_user/host functions and are null-terminated strings.
Fixed

sql/auth/sql_auth_cache.cc
kamil-holubicki: Yup, at the beginning I just wanted to use case insensitive regex matching... Now std::find is enough.
fixed.

sql/auth/sql_security_ctx.cc
kamil-holubicki: You are right. I didn't notice that m_priv_user and m_priv_host are assigned only by assign_priv_user/host functions and are null-terminated strings.
Fixed

sql/auth/sql_auth_cache.cc
kamil-holubicki: Comment 1: Agree, fixed
Comment 2: I didn't find anywhere that it is not allowed :) Another question is if leading/trailing underscores are allowed. I think they are. Anyway having starting digit, underscore will not do any harm to this.
Comment 3: regex is long, but I think it is not complicated
(any whitespaces followed by at least 1 alphanum or underscore followed by any whitespaces followed by any comma) repeated any time, followed by any whitespaces followed by at leas 1 alphanum or underscore followed by any whitespaces

That elimitates leading/trailing commas, forbidden characters, forbidden separators.

Just like
'      PRIVILEGE_1    ,,, PRIVILEGE_2,PRIVILEGE_3     '

The idea here is to validate the whole string "as it is" in the 1st step and then split in 2nd step as we know that we have valid string, with potential whitespaces and multiple inner commas. We can use regexp for both steps.

If we want to remove whitespaces from the original string as the first step, I think there is no point in regex usage at all.

percona-ysorokin (approved):
LGTM

PR #3892, opened 2020-06-23
PS-7076 : Cardinality is equal for all indexes when last partition empty

george-lorch:
This is a regression created by the change for PS-4315 : "Index cardinality
stats updated for handler::info(HA_STATUS_CONST), not often enough for
tokudb_cardinality_scale_percent".

The storage/partition/ha_partition.cc::info call for HA_STATUS_VARIABLE iterates
the partitions and calls their individual info(HA_STATUS_VARIABLE) methods.  Due
to the change from PS-4315, this causes each partition to re-push the
HA_STATUS_CONST (cardinality) from the individual partitions to the TABLE_SHARE
instance.  This means that the last partition in the list will always be the one
that updates the TABLE_SHARE cardinality.  This happens because it seems that
info(HA_STATUS_VARIABLE) is always being called on the partition handler
immediately after info(HA_STATUS_CONST) and overwriting the correct cardinality
that was set during the info(HA_STATUS_CONST) call (from the largest partition).

I believe a more appropriate fix for PS-4315 is not to push cardinality on
HA_STATUS_VARIABLE but to accept that cardinality will not be updated after
setting tokudb_cardinality_scale_percent as this is not something that should be
set much at all and document that a new ANALYZE TABLE is required after setting
tokudb_cardinality_scale_percent.

This commit reverts the HA_STATUS_VARIABLE portion of the changes for PS-4315
and introduces a test that explicitly exposes and tests the issue that was
introduced by PS-4315.


george-lorch:
https://ps57.cd.percona.com/job/percona-server-5.7-param/293/

inikep (approved):
LGTM (I compared this patch with https://github.com/percona/percona-server/pull/1978/files)

percona-ysorokin (approved):
LGTM

PR #3899, opened 2020-06-26
PS-6849 Add JUnit reporting support to MTR (5.7)

oleksandr-kachan:
- Add JUnit reporting support to the MySQL test suite runner
- Add "junit" test suite with deliberately failing tests
- Add a deliberately failing unittests suite
- Add unittests timeout, include CTest output into JUnit report
- Attach mysqld shutdown and valgrind reports to JUnit report
- Run unit tests under Valgrind too if it was enabled
- Mark self-skipping tests as skipped in JUnit reports
- Convey tests output to JUnit reports regardless of test failure
- Exclude tests skipped by framework from results xml

There are the following differences compared with 5.6 branch:
- Bootstrap failures were moved from mysqld.cc to bootstrap.cc, as bootstrap code itself was moved in 5.7.
- Shutdown timeout was increased from 10 seconds to 20 seconds. In master-slave configurations, 10 seconds wasn't enough, mysqld was forcefully killed too often. It was OK before changes in this patches, but since shutdown timeouts are now tracked, insufficient time limits generated a lot of false positives.


percona-ysorokin (approved):
LGTM

PR #3901, opened 2020-06-28
PS-6849 Add JUnit reporting support to MTR (8.0)

oleksandr-kachan:
- Add JUnit reporting support to the MySQL test suite runner
- Add "junit" test suite with deliberately failing tests
- Add a deliberately failing unittests suite
- Attach mysqld shutdown and valgrind reports to JUnit report
- Run unit tests under Valgrind too if it was enabled
- Mark self-skipping tests as skipped in JUnit reports
- Convey tests output to JUnit reports regardless of test failure
- Exclude tests skipped by framework from results xml

There are the following changes compared to the fix for 5.7 branch:
- Patch with unit test timeout support was dropped, as 8.0 already has a setting for that.
- Patch that increased shutdown timeout to 20 seconds was dropped, as 8.0 already has shutdown timeout increased to 20 seconds.
- Memory writes that generate buffer overruns were changed from my_atomic_*() to volatile int writes, since my_atomic_*() family of functions were removed in 8.0.


percona-ysorokin (approved):
LGTM

PR #3902, opened 2020-06-29
PS-6990: Server doesn't restart after crash when there are too many gaps in the mysql.gtid_executed table.

venkatesh-prasad-v:
PS-6990: Server doesn't restart after crash when there are too many gaps in the mysql.gtid_executed table.

Problem & Analysis
------------------
Since the introduction of a dedicated thread for persisting GTIDs of
InnoDB transactions by WL#9211, new behavior for updating the
mysql.gtid_executed table is as follows,

- If binary log is enabled, the mysql.gtid_executed table is updated on
  next binlog rotation (either by FLUSH LOGS or server restart).
- If binary log is disabled or log_slave_updates is disabled (for slave
  threads), then
  - If it is an InnoDB transaction, it leaves it to the gtid persister
    thread to update.
  - Otherwise, it writes its GTID into mysql.gtid_executed table.

On a binlogless slave having both transactional and non-transactional
workloads, as per the current design, when the GTID persister thread
reaches the threshold (either once per 1k transactions or every 1
second), it flushes its list (updates the table) to the
mysql.gtid_executed table and tries to compress the gtid_executed table.
It succeeds compressing only the first few rows (because it filled the
gaps present in the beginning of the table by merging transactional
updates) and fails to compress further rows of the table (because of the
gaps introduced by non-transactional updates).

In every attempt it tries to merge the table, it merges only a few
consecutive rows and leaves the other rows as is, thereby taking more
time for scanning the full table. By the time it finishes the table
scan, the slave applier threads would insert few more entries to the end
of the table. As a result, when there is a high load on the server, if
the GTID persister thread once starts scanning the table, it is more
likely that the scan never stops and thus causes the flush list to grow
causing more gaps in the mysql.executed table.

Fix
---
Don't allow GTID persister thread to compress the mysql.gtid_executed
table and instead signal the GTID compressor thread unless there is an
explicit request.

Testing
--------
Jenkins: https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/636/testReport/

Summary:
1. The newly added test `clone.local_basic_gtid_compression` is failing because of missing re-recording. It is fixed now.
2. For those tests which are even reproducibale without this fix, have created JIRA tickets [PS-7147](https://jira.percona.com/browse/PS-7147), [PS-7148](https://jira.percona.com/browse/PS-7147) and [PS-7149](https://jira.percona.com/browse/PS-7147) for tracking them.
3. All other failing tests are sporadic in nature and are unrelated to this fix.

### DIFF
--- a/mysql-test/suite/clone/t/local_basic_gtid_compression.test
+++ b/mysql-test/suite/clone/t/local_basic_gtid_compression.test
@@ -1,0 +1,123 @@
+# ==== Purpose ====
+#
+# This test verifies that compression of mysql.gtid_executed table shall be
+# done only by the gtid persister thread for clone operation.
+#
+# ==== Implementation ====
+#
+# 0. This test requires only one server.
+# 1. Create InnoDB and MyISAM tables and insert 2 rows.
+# 2. Assert that mysql.gtid_executed table is not compressed and has gaps.
+# 3. Install Clone Plugin, enable debug point to print compression info
+#    and initiate Clone operation.
+# 4. Assert that table is compressed after Clone operation.
+# 5. Assert that the compression was done by the gtid persister thread.
+# 6. Cleanup
+#
+# ==== References ====
+#
+# PS-6990: Gaps in mysql.gtid_executed but not in @@GLOBAL.gtid_executed
+
+# This test requires debug binaries
+--source include/have_debug.inc
+# This test requires binlogging.
+--source include/have_log_bin.inc
+--source include/count_sessions.inc
+--let $server_uuid = `SELECT @@GLOBAL.SERVER_UUID`
+
+# Clean the table before starting the test.
+RESET MASTER;
+
+# Test setup for clone test.
+--source ../include/clone_connection_begin.inc
+--let $CLONE_DATADIR = $MYSQL_TMP_DIR/data_new
+
+--echo #
+--echo # 1. Create InnoDB and MyISAM tables and insert 2 rows.
+CREATE TABLE t1(col1 INT PRIMARY KEY, col2 char(64)) ENGINE = INNODB;
+CREATE TABLE t2(col1 INT PRIMARY KEY, col2 char(64)) ENGINE = MyISAM;
+
+INSERT INTO t2 VALUES(10, 'myisam row 1');
+INSERT INTO t1 VALUES(10, 'innodb row 1');
+INSERT INTO t2 VALUES(20, 'myisam row 2');
+INSERT INTO t1 VALUES(20, 'innodb row 2');
+
+--echo #
+--echo # 2. Assert that mysql.gtid_executed table is not compressed and has gaps.
+
+# Assert that 6 transactions have been committed in total.
+--let $assert_text = COMMITTED_GTIDS SERVER_UUID:1-6
+--let $assert_cond = "[SELECT @@GLOBAL.GTID_EXECUTED]" = "$server_uuid:1-6"
+--source include/assert.inc
+
+# When binlog is enabled, GTIDs of InnoDB transactions are updated by gtid
+# persister thread and GTIDs of other transactions are updated on binlog
+# rotation. So, wait till the GTIDs of InnoDB transactions are updated by the
+# gtid persister thread.
+#
+# i.e, wait till the below state is reached.
+#
+# SELECT * FROM mysql.gtid_executed;
+# source_uuid	interval_start	interval_end
+# ------------------------------------------
+# server_uuid	             1	           1
+# server_uuid	             4	           4
+# server_uuid	             6	           6
+
+--let $wait_condition = SELECT count(*) = 3 FROM mysql.gtid_executed
+--source include/wait_condition.inc
+
+--let $assert_cond = COUNT(*) = 3 FROM mysql.gtid_executed
+--let $assert_text = The mysql gtid table is not compressed and has gaps.
+--source include/assert.inc
+
+--echo #
+--echo # 3. Install Clone Plugin, enable debug point to print compression info
+--echo #    and initiate Clone operation.
+
+# Install Clone Plugin
+--replace_result $CLONE_PLUGIN CLONE_PLUGIN
+--eval INSTALL PLUGIN clone SONAME '$CLONE_PLUGIN'
+
+# Enable debug point to print the compression info for validation purpose.
+--let $debug_point = print_gtid_compression_info
+--source include/add_debug_point.inc
+
+# Clone data
+--connection clone_conn_1
+--source ../include/clone_command.inc
+
+# During clone operation, persister thread, along with writing InnoDB GTIDs, it
+# also writes non-InnoDB GTIDs and even compresses it by creating an explicit
+# request.
+
+--echo #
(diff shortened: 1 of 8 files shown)


storage/innobase/clone/clone0repl.cc
```diff
@@ -459,6 +455,10 @@ void Clone_persist_gtid::flush_gtids(THD *thd) {
     clone operation. */
     clone_update_gtid_status(all_gtids);
     my_free(gtid_buffer);
+
+    /* Compress always if in recovery. */
+    compress_recovery = true;
```
satya-bodapati: This is now independent of recovery mode. Is this intentional?

Before your patch, "!m_thread_active.load()" (which is also stored in is_recovery) is used to set compress_recovery only on recovery. Now it is unconditional

If this is a mistake, I would sugged to move this into the above if (is_recovery) code.

satya-bodapati (changes requested):

storage/innobase/clone/clone0repl.cc
venkatesh-prasad-v: Hi Satya,

Two things here.
1. We are setting compress_recovery from inside the `if (is_recovery)` block only. 
2. This change is intentional and necessary to make the compression of the gtid_executed table even when InnoDB recovered zero GTIDs from the undo log post recovery. Right now it is compressed post recovery only when `m_num_gtid_mem.load() != 0` meaning only when InnoDB read at least one GTID from the undolog.

With this patch, we ensure that compression is done either by the persister or compressor threads on server restart.

satya-bodapati: 1) my bad, yeah.. it is already inside is_recovery block

2) why do we need compression of gtids from InnoDB if InnoDB doesn't find any gtids from its undo header. Wouldn't it be better if gtid compressor thread do it on restart?

venkatesh-prasad-v: Referring to https://dev.mysql.com/doc/refman/8.0/en/replication-gtids-concepts.html.

```When the server instance is started and the thread/sql/compress_gtid_table thread is launched, in most server configurations, compression is performed for the mysql.gtid_executed table. In releases before MySQL 8.0.17 when binary logging is enabled, compression is triggered by the fact of the binary log being rotated at startup. In releases from MySQL 8.0.20, compression is triggered by the thread launch. In the intervening releases, compression does not take place at startup.```

I probably didn't read it properly and was under the assumption that compression is not taking place even now when binary logs are rotated at startup. So I went with this approach to perform compression during server restart.

But this issue has been fixed by Oracle in 8.0.20 in a different way by setting the `should_compress` flag before entering the wait loop(https://github.com/mysql/mysql-server/blob/8.0/sql/rpl_gtid_persist.cc#L759).

So, my change is not required and will revert it as we are anyway merging the Oracle bugfix in our 8.0.20 merge.

Comment (satya-bodapati) [reply storage/innobase/clone/clone0repl.cc]: Good find!

File: mysql-test/suite/clone/t/local_basic_gtid_compression.test [hunk 2]
```diff
@@ -0,0 +1,122 @@
...
+--let $debug_point = print_gtid_compression_info
+--source include/add_debug_point.inc
+
+# Clone data
+--connection clone_conn_1
+--source ../include/clone_command.inc
+
+# During clone operation, persister thread, along with writing InnoDB GTIDs, it
+# also writes non-InnoDB GTIDs and even compresses it by creating an explicit
+# request.
```
satya-bodapati, line 92: How will gtid persistor thread know about the gtids commited outside InnoDB?

mysql-test/suite/clone/t/local_basic_gtid_compression.test
```diff
@@ -0,0 +1,122 @@
...
+--echo #
+--echo # 6. Cleanup
+--let $debug_point = print_gtid_compression_info
+--source include/remove_debug_point.inc
+DROP TABLE t1;
+DROP TABLE t2;
+--force-rmdir $CLONE_DATADIR
+--source ../include/clone_connection_end.inc
+
+UNINSTALL PLUGIN clone;
```
satya-bodapati, line 122: Use wait_until_count_sessions.inc and count_session.inc combo

storage/innobase/clone/clone0repl.cc
```diff
@@ -471,7 +472,14 @@ void Clone_persist_gtid::flush_gtids(THD *thd) {
     m_compression_gtid_counter = 0;
     /* Write non-innodb GTIDs before compression. */
     write_other_gtids();
-    err = gtid_table_persistor->compress(thd);
+
+    /* Compress the gtid_executed table if there is an explicit compression
+       request. Otherwise signal the gtid persister thread to perform
+       compression. */
```
satya-bodapati:
```suggestion
    /* Compress the gtid_executed table if there is an explicit compression
    request. Otherwise signal the gtid persister thread to perform
    compression. */
```

satya-bodapati (changes requested):
Looks good mostly..  Only minor comments.

Please also fix commit msg and github description about this part

"
Additionally, this patch also makes the compression thread to perform
first run of compression when the server starts."

mysql-test/suite/clone/t/local_basic_gtid_compression.test
venkatesh-prasad-v: * When clone is [synchronising](https://github.com/percona/percona-server/blob/8.0/storage/innobase/clone/clone0copy.cc#L269) with the binary log, it creates an explicit request for a flush and enters wait. 
* When persister thread receives an explicit request, it updates [non-innodb gtids](https://github.com/percona/percona-server/blob/8.0/storage/innobase/clone/clone0repl.cc#L473) and also compresses it.

mysql-test/suite/clone/t/local_basic_gtid_compression.test
venkatesh-prasad-v: Done.

storage/innobase/clone/clone0repl.cc
venkatesh-prasad-v: Done.

mysql-test/suite/clone/t/local_basic_gtid_compression.test
satya-bodapati: ah.. that way.. but the gtids are are not inserted into mysql.gtid_executed table by InnoDB or persistor thread right? Who will insert into mysql.gtid_executed table for non InnoDB transactions?

venkatesh-prasad-v: Persister thread itself writes that.
```
  307 int Clone_persist_gtid::write_other_gtids() {
  308   int err = 0;
  309   if (opt_bin_log) {
  310     err = gtid_state->save_gtids_of_last_binlog_into_table();
  311   }
  312   return (err);
  313 }
```

satya-bodapati: Thanks! its clear now

satya-bodapati (approved):
LGTM! Good work on chasing this! :+1:  :clap:

PR #3904, opened 2020-06-30
PS-7140: crypt redo logs are not applied correctly

robgolebiowski:
In 5.7 redo log applier (log0recv.cc) would go though all the redo logs
twice. First with apply variable set to false and the second time with
apply variable set to true. The first pass was to parse MLOG_FILE_*
family of redo log records (and MLOG_WRITE_STRING). On this first pass
necessary updates to in-memory structures were made, but no actual
changes were made to the underlying table's pages. In 8.0 upstream
decided to abandon apply variable and now goes twice though the redo
logs. In case of redoing Master Key encryption, in 5.7 log0recv would go
twice through the redo logs (encryption info is written as
MLOG_WRITE_STRING) - first time setting all necessary keys in memory and
the second time applying changes to page0. It was similar with
crypt_data - the first pass would create the crypt_data in-memory and
associate it with space (in case it would exists) and the second pass
would apply changes to page0. The problem was that apply was never set
properly - it was always true and for true it would create in memory
crypt_data only. The fix was to create in memory crypt_data only when
apply is equal to false and also to go through redo logs first with
apply == false and the second time with apply == true. The crypt_data
applied to page0 by redo will be read during opening a tablespace for
recovery (after it was created) and thus crypt_data will get assigned to
tablespace.

### DIFF
--- a/storage/innobase/fil/fil0crypt.cc
+++ b/storage/innobase/fil/fil0crypt.cc
@@ -275,14 +275,18 @@ void fil_space_crypt_cleanup() {
 fil_space_crypt_t::fil_space_crypt_t(uint new_min_key_version, uint new_key_id,
                                      const char *new_uuid,
                                      fil_encryption_t new_encryption,
-                                     Crypt_key_operation key_operation,
-                                     Encryption_rotation encryption_rotation)
+                                     Crypt_key_operation key_operation)
     : min_key_version(new_min_key_version),
       encryption(new_encryption),
       key_found(false),
       rotate_state(),
-      encryption_rotation(encryption_rotation),
       tablespace_key(NULL) {
+  encryption_rotation =
+      (new_encryption == FIL_ENCRYPTION_DEFAULT &&
+       srv_default_table_encryption == DEFAULT_TABLE_ENC_ONLINE_TO_KEYRING)
+          ? Encryption_rotation::ENCRYPTING
+          : Encryption_rotation::NO_ROTATION;
+
   mutex_create(LATCH_ID_FIL_CRYPT_START_ROTATE_MUTEX, &start_rotate_mutex);
   mutex_create(LATCH_ID_FIL_CRYPT_DATA_MUTEX, &mutex);
 
@@ -865,10 +869,9 @@ Write crypt data to a page (0)
 @param[in,out]	mtr	mini-transaction */
 
 // TODO: Should be marked as const when PS-5738 is implemented
-void fil_space_crypt_t::write_page0(
-    const fil_space_t *space, byte *page, mtr_t *mtr, uint a_min_key_version,
-    uint a_max_key_version, uint a_type,
-    Encryption_rotation current_encryption_rotation) {
+void fil_space_crypt_t::write_page0(const fil_space_t *space, byte *page,
+                                    mtr_t *mtr, uint a_min_key_version,
+                                    uint a_max_key_version, uint a_type) {
   ut_ad(this == space->crypt_data);
   const ulint offset =
       fsp_header_get_keyring_encryption_offset(page_size_t(space->flags));
@@ -908,8 +911,7 @@ void fil_space_crypt_t::write_page0(
   memcpy(encrypt_info_ptr, iv, CRYPT_SCHEME_1_IV_LEN);
   encrypt_info_ptr += CRYPT_SCHEME_1_IV_LEN;
 
-  mach_write_to_1(encrypt_info_ptr,
-                  static_cast<byte>(current_encryption_rotation));
+  mach_write_to_1(encrypt_info_ptr, static_cast<byte>(encryption_rotation));
   encrypt_info_ptr += 1;
 
   if (tablespace_key == nullptr) {
@@ -1027,6 +1029,7 @@ byte *fil_parse_write_crypt_data_v3(space_id_t space_id, byte *ptr,
   fil_space_crypt_t *crypt_data =
       fil_space_create_crypt_data(encryption, key_id, uuid, key_operation);
   /* Need to overwrite these as above will initialize fields. */
+  crypt_data->type = type;
   DBUG_ASSERT(min_key_version != ENCRYPTION_KEY_VERSION_INVALID);
   crypt_data->min_key_version = min_key_version;
   crypt_data->max_key_version = max_key_version;
@@ -1087,7 +1090,11 @@ byte *fil_parse_write_crypt_data_v3(space_id_t space_id, byte *ptr,
     crypt_data = fil_space_set_crypt_data(space, crypt_data);
     fil_space_release(space);
   } else {
-    fil_space_destroy_crypt_data(&crypt_data);
+    // crypt_data was created as part of creating a new tablespace
+    if (recv_sys->crypt_datas->count(space_id) > 0) {
+      fil_space_destroy_crypt_data(&(*recv_sys->crypt_datas)[space_id]);
+    }
+    (*recv_sys->crypt_datas)[space_id] = crypt_data;
   }
 
   // We are advancing the ptr pointer while reading crypt_data - make
@@ -1183,8 +1190,7 @@ byte *fil_parse_write_crypt_data_v2(space_id_t space_id, byte *ptr,
   }
 
   /* update fil_space memory cache with crypt_data */
-  fil_space_t *space = fil_space_acquire_silent(space_id);
-  if (space != nullptr) {
+  if (fil_space_t *space = fil_space_acquire_silent(space_id)) {
     crypt_data = fil_space_set_crypt_data(space, crypt_data);
     fil_space_release(space);
   } else {
@@ -1426,8 +1432,7 @@ static void fil_crypt_write_crypt_data_to_page0(fil_space_t *space) {
           Page_fetch::NORMAL, __FILE__, __LINE__, &mtr)) {
     space->crypt_data->write_page0(
         space, block->frame, &mtr, space->crypt_data->min_key_version,
-        space->crypt_data->max_key_version, space->crypt_data->type,
-        space->crypt_data->encryption_rotation);
+        space->crypt_data->max_key_version, space->crypt_data->type);
   }
   mtr.commit();
 }
@@ -1709,8 +1714,7 @@ static bool fil_crypt_start_encrypting_space(fil_space_t *space) {
     byte *frame = buf_block_get_frame(block);
     crypt_data->type = CRYPT_SCHEME_1;
     crypt_data->write_page0(space, frame, &mtr, crypt_data->min_key_version,
-                            crypt_data->max_key_version, crypt_data->type,
-                            crypt_data->encryption_rotation);
+                            crypt_data->max_key_version, crypt_data->type);
 
     mtr.commit();
     /* 4 - sync tablespace before publishing crypt data */
@@ -2372,6 +2376,12 @@ static bool fil_crypt_start_rotate_space(const key_state_t *key_state,
       mutex_exit(&crypt_data->start_rotate_mutex);
       return false;
     }
+
+    if (key_state->key_version == ENCRYPTION_KEY_VERSION_NOT_ENCRYPTED)
+      crypt_data->encryption_rotation = Encryption_rotation::DECRYPTING;
+    else if (crypt_data->encryption_rotation !=
+             Encryption_rotation::MASTER_KEY_TO_KEYRING)
+      crypt_data->encryption_rotation = Encryption_rotation::ENCRYPTING;
 
     /* only first thread needs to init */
     crypt_data->rotate_state.next_offset = 1;  // skip page 0
@@ -3330,6 +3340,12 @@ static dberr_t fil_crypt_flush_space(rotate_thread_t *state) {
         crypt_data->encrypted_validation_tag);
   }
 
+  if (crypt_data->encryption_rotation ==
+      Encryption_rotation::MASTER_KEY_TO_KEYRING) {
+    crypt_data->set_tablespace_key(nullptr);
+    crypt_data->encryption_rotation = Encryption_rotation::ENCRYPTING;
+  }
+
   /* update page 0 */
   mtr_t mtr;
   mtr.start();
@@ -3343,7 +3359,7 @@ static dberr_t fil_crypt_flush_space(rotate_thread_t *state) {
                             current_type == CRYPT_SCHEME_UNENCRYPTED
                                 ? ENCRYPTION_KEY_VERSION_NOT_ENCRYPTED
                                 : crypt_data->max_key_version,
-                            current_type, Encryption_rotation::NO_ROTATION);
+                            current_type);
   }
 
   mtr.commit();
@@ -3409,8 +3425,6 @@ static void fil_crypt_complete_rotate_space(const key_state_t *key_state,
       /* we're the last active thread */
       ut_ad(crypt_data->rotate_state.flushing == false);
       crypt_data->rotate_state.flushing = true;
-      crypt_data->set_tablespace_key(NULL);
-      crypt_data->encryption_rotation = Encryption_rotation::NO_ROTATION;
     }
 
     DBUG_EXECUTE_IF(
--- a/storage/innobase/fil/fil0crypt.cc
+++ b/storage/innobase/fil/fil0crypt.cc
@@ -967,9 +967,11 @@ static fil_space_crypt_t *fil_space_set_crypt_data(
 @param[in]  ptr  Log entry start
 @param[in]  end_ptr  Log entry end
 @param[in]  len  Log entry length
+@param[in]  recv_needed_recovery  missing keys will report error
 @return position on log buffer */
 byte *fil_parse_write_crypt_data_v3(space_id_t space_id, byte *ptr,
-                                    const byte *end_ptr, ulint len) {
+                                    const byte *end_ptr, ulint len,
+                                    bool recv_needed_recovery) {
   ptr += 4;  // skip offset and len
 
 #ifdef UNIV_DEBUG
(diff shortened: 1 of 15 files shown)


storage/innobase/log/log0recv.cc
```diff
@@ -1640,11 +1639,11 @@ static byte *recv_parse_or_apply_log_rec_body(mlog_id_t type, byte *ptr,
...
           } else if (memcmp(ptr_copy, ENCRYPTION_KEY_MAGIC_PS_V1,
                             ENCRYPTION_MAGIC_SIZE) == 0 &&
-                     apply) {
+                     !apply) {
             return (fil_parse_write_crypt_data_v1(space_id, ptr, end_ptr, len));
           } else if (memcmp(ptr_copy, ENCRYPTION_KEY_MAGIC_PS_V2,
                             ENCRYPTION_MAGIC_SIZE) == 0 &&
-                     apply) {
+                     !apply) {
             return (fil_parse_write_crypt_data_v2(space_id, ptr, end_ptr, len));
```
satya-bodapati: So, the apply is used to tell whether to apply a record to the page or not. and it should be false for the 'parse' phase which you have done below.. 

For MK, after a long study, I found out that this fil_tablespace_encryption_redo() is called three times :open_mouth: 
2 times during parse and 1 time from the actual apply.

1st iteration (parse): space is nullptr, so it stores in recv_sys -> keys
2 iteration (parse): space is  nullptr again, wasted work, updates the same key again for key which is already stored in recv_sys keys
3rd iteration (apply from recover_page_func()):  sets space->encryption_key and iv from recv_sys->keys.

There is also encryption decoding when we parse the encryption_data from redo.

WIth your fix, you seem to parse (ie. do fil_parse_crypt_data_v2) during the parse stage and write to page 0 on apply.. 

This looks correct but I think there is issue here
Would there be space->crypt_data during the apply stage? During apply stage, when the space is **not** nullptr, we do not call fil_parse(). So space->crypt_data should be null ptr. This would be necessary for decryption of other pages. isn't it?

storage/innobase/log/log0recv.cc
```diff
@@ -2668,9 +2667,8 @@ ulint recv_parse_log_rec(mlog_id_t *type, byte *ptr, byte *end_ptr,
           mlog_parse_initial_dict_log_record(ptr, end_ptr, type, &id, &version);
 
       if (new_ptr != nullptr) {
-        new_ptr =
-            (apply ? recv_sys->metadata_recover : log_online_metadata_recover)
-                ->parseMetadataLog(id, version, new_ptr, end_ptr, apply);
+        new_ptr = recv_sys->metadata_recover->parseMetadataLog(
```
satya-bodapati: This is dangerous. When we do log online parsing, recv_sys is nullptr. Recipe for  :boom:

storage/innobase/log/log0recv.cc
```diff
@@ -2878,7 +2876,7 @@ static bool recv_multi_rec(byte *ptr, byte *end_ptr) {
     space_id_t space_id = 0;
 
     ulint len = recv_parse_log_rec(&type, ptr, end_ptr, &space_id, &page_no,
-                                   true, &body);
+                                   false, &body);
```
satya-bodapati, line 2887: am thinking if we should get rid of this 'apply' altogether. It can be inferred from block pointer. If block is nullptr, it is scan/parse phase. During apply, we always have block as non null ptr

satya-bodapati (changes requested):
Please format commit message into more readable way
Problem
------------
Fix:
-----

storage/innobase/log/log0recv.cc
robgolebiowski: It is not called during apply stage, but is called during parse stage. So when tablespace already exists in the cache - it will get crypt_data assigned. In case it does not exist (create table is only in redo) only changes to page0 will get applied (crypt_data will not be assigned in log0recv). And all other changes to other pages will get applied (redo is not encrypted, and all pages associated with the table are still in redo). After apply the table will be opened by open_for_recovery (IIRC), space will be loaded into cache and crypt_data will get read from page0.

storage/innobase/log/log0recv.cc
robgolebiowski: But apply before the change was always true, so nothing changes here in execution. Also in upstream they do not check for apply (because they do not have it) and go straight to recv_sys->metadata_recover->parseMetadataLog.

storage/innobase/log/log0recv.cc
robgolebiowski: Yes, but I like apply better - it shows some intentions. If I was to reverse engineer this without apply variable, it would have been much harder.

satya-bodapati: There is already existing variable "    bool applying_redo = (block != nullptr);" in recv_parse_or_apply_log_rec_body(). Please reuse this.

storage/innobase/log/log0recv.cc
satya-bodapati: Ok, this is an existing issue. ie. During log online parsing, which happens during the server, we should use log_online_metadata_recover. Because recv objects are freed after the recovery. So we definitely need to use log_online_metadata_recover.

Please create a new bug to handle this. I know that log online parsing is broken in 8.0 at least just enabling it, shouldn't the crash the server.

storage/innobase/log/log0recv.cc
satya-bodapati: I understood the fix, During parse (apply is false), we just parse the crypt_data in the redo. (if space exists, crypt_data gets assigned), During apply (apply is true), we make the modification to page0.

In the scenario where page 0 changes are only in redo, and changes on other pages need decryption or encryption.
In this case, since space->crypt_data is nullptr, all the pages written during the redo stage will be written either MK encrypted or unencrypted.

Consider calling the parse + page 0 write during the apply (apply is true).

satya-bodapati: For redo parsing tools like PXB, which create file, based on the redo records, the MLOG_FILE_CREATE should the first record. So please change the order of redo creation for crypt_data and MLOG_FILE_CREATE*

satya-bodapati (changes requested):

storage/innobase/log/log0recv.cc
robgolebiowski: I have made a mistake when I was trying to understand how this work. Please check the new commit message. I have changed the fix. Now crypt_data is assigned during opening tablespace for recovery. All changes from redo will follow the encryption in crypt_data.

storage/innobase/log/log0recv.cc
robgolebiowski: I was wrong here too. This is indeed needed. However I changed the apply variable to online_log - to indicate that we are parsing online log and not applying anything. This confused me.

storage/innobase/log/log0recv.cc
robgolebiowski: done

storage/innobase/log/log0recv.cc
robgolebiowski: Space create during parsing MLOG_FILE_CREATE will be opened with open_for_recovery - where we assign harversted (during parsing) crypt_data.

robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/667/

storage/innobase/fil/fil0crypt.cc
```diff
@@ -836,7 +836,11 @@ byte *fil_parse_write_crypt_data_v2(space_id_t space_id, byte *ptr,
     crypt_data = fil_space_set_crypt_data(space, crypt_data);
     fil_space_release(space);
   } else {
-    fil_space_destroy_crypt_data(&crypt_data);
+    // crypt_data was created as part of creating a new tablespace
+    if (recv_sys->crypt_datas->count(space_id) > 0) {
+      fil_space_destroy_crypt_data(&(*recv_sys->crypt_datas)[space_id]);
+    }
+    (*recv_sys->crypt_datas)[space_id] = crypt_data;
```
satya-bodapati, line 1107: Can you please check how this parsing will be handled for the "online_log" parsing mode (i.e parsing redo log for page tracking)?  recv_sys will be nullptr? :thinking:

storage/innobase/include/log0recv.h
```diff
@@ -80,12 +81,12 @@ void recv_read_log_seg(log_t &log, byte *buf, lsn_t start_lsn, lsn_t end_lsn,
 @param[in]	end_ptr		end of the buffer
 @param[out]	space_id	tablespace identifier
 @param[out]	page_no		page number
-@param[in]	apply		whether to apply the record
+@param[in]      online_log      do we process online DDL log
```
satya-bodapati: align the formatting and instead of "do we process online DDL log",  please use "true if we parse redo log for page tracking"

storage/innobase/log/log0online.cc
```diff
@@ -639,8 +639,8 @@ static bool log_online_read_bitmap_page(
   IORequest io_request(IORequest::LOG | IORequest::READ |
                        IORequest::NO_ENCRYPTION);
   const bool success =
-      os_file_read(io_request, bitmap_file->name, bitmap_file->file, page, bitmap_file->offset,
-                   MODIFIED_PAGE_BLOCK_SIZE);
+      os_file_read(io_request, bitmap_file->name, bitmap_file->file, page,
+                   bitmap_file->offset, MODIFIED_PAGE_BLOCK_SIZE);
```
satya-bodapati, line 643: is the clang-formatting broken in this file? Is this change because you modified this file now?

storage/innobase/include/log0recv.h
```diff
@@ -642,6 +640,11 @@ struct recv_sys_t {
   /** Encryption Key information per tablespace ID */
   Encryption_Keys *keys;
 
+  using CryptDatas =
+      std::unordered_map<space_id_t, fil_space_crypt_t *, std::hash<space_id_t>,
+                         std::equal_to<space_id_t>>;
+  CryptDatas *crypt_datas;
```
satya-bodapati, line 585: Can you please add comment for this member? how it is useful/purpose?

Also it looks like upstream is following this style for class/struct names. "Hello_world". So may be rename to Crypt_data?

storage/innobase/fil/fil0fil.cc
```diff
@@ -10453,6 +10453,30 @@ bool Fil_system::open_for_recovery(space_id_t space_id) {
       fil_tablespace_encryption_init(space);
     }
 
+    if (recv_sys->crypt_datas != nullptr &&
+        recv_sys->crypt_datas->count(space_id) > 0) {
+      if (space->crypt_data != nullptr) {
+        fil_space_destroy_crypt_data(&space->crypt_data);
+      }
+      space->crypt_data = (*recv_sys->crypt_datas)[space_id];
+      recv_sys->crypt_datas->erase(space_id);
```
satya-bodapati, line 10450: This is a bit tricky area. I think we should trust crypt_data found in page 0? Looks like, here you destroy the crypt_data found from page 0 read and use the one from recovery. IMHO, we should use crypt_data from recovery only if there is no crypt_data found in page 0.

Am also bit unsure on when to use which crypt_data (from the one recovery or from page 0)... really depends on ALTERs.. Can you please give some more thought? see if it covers the ALTER scenarios as well.

satya-bodapati (changes requested):
Looks good mostly

storage/innobase/include/log0recv.h
robgolebiowski: done

storage/innobase/include/log0recv.h
robgolebiowski: Added comment. About the naming style, from this file from upstream:
using PersistentTables = std::map<
      table_id_t, PersistentTableMetadata *, std::less<table_id_t>,
      ut_allocator<std::pair<const table_id_t, PersistentTableMetadata *>>>;

storage/innobase/log/log0online.cc
robgolebiowski: Yes. To check this, I reverted this change, run clang-format-8 -> the change reappeared.

storage/innobase/fil/fil0fil.cc
robgolebiowski: The write of crypt_data to page0 is always creating a full redo log record of crypt_data. Based on this record we recreate the whole crypt_data structure. Whatever is there in a redo - this is the newest (full) version of crypt_data and should replace existing crypt_data.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: Parsing for online_log will never reach this place. When we start parsing in online_log mode, we have already applied all the log records. Also, from this function there is call to fil_tablespace_encryption, which in turn also uses recv_sys->keys.

storage/innobase/include/log0recv.h
satya-bodapati: this is old code before the convention kicked in..

storage/innobase/fil/fil0fil.cc
satya-bodapati: Not always. what if there is cryp_data from redo and the current table is unencrypted? i.e sequence of ALTERs with final once as unencrypted. I guess you are in the safe zone because you need table rebuilding ALTER to make it unencrypted? :thinking: 

So what would happen is, redo parsing will use crypt_data for tablespace, but in reality, the table should be unencrypted?

robgolebiowski: I have added Encryption_rotation::ENCRYPTING to address this issue. Please check the commit message.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -908,9 +911,8 @@ void fil_space_crypt_t::write_page0(
   memcpy(encrypt_info_ptr, iv, CRYPT_SCHEME_1_IV_LEN);
   encrypt_info_ptr += CRYPT_SCHEME_1_IV_LEN;
 
-  mach_write_to_1(encrypt_info_ptr,
-                  static_cast<byte>(current_encryption_rotation));
-  encrypt_info_ptr += 1;
+  mach_write_to_2(encrypt_info_ptr, static_cast<byte>(encryption_rotation));
+  encrypt_info_ptr += 2;
```
satya-bodapati: Do we need two bytes? And since crypt_data format is chaning, do we need to bump up its version number?

storage/innobase/include/os0enc.h
```diff
@@ -50,7 +50,9 @@ extern ulint os_io_ptr_align;
 
 enum class Encryption_rotation : std::uint8_t {
   NO_ROTATION,
-  MASTER_KEY_TO_KEYRING
+  MASTER_KEY_TO_KEYRING,
+  ENCRYPTING,
+  DECRYPTING
```
satya-bodapati: can you add a brief comment about these new states? How these are useful? in which scenarios?

satya-bodapati (changes requested):
Looks mostly good.  Will check the marker flags in next round.

Minor changes requested.

storage/innobase/fil/fil0crypt.cc
robgolebiowski: uh, no, 1 byte is enough, I will revert this change

storage/innobase/include/os0enc.h
robgolebiowski: added

storage/innobase/fil/fil0fil.cc
satya-bodapati: Looks like you are solving a different problem. my question is how do we decide which crypt_data to use? The one from redo or the one from page 0? that is why I said it is tricky.. Can we always trust and use the crypt_data from redo? (May be :thinking: )

@satya-bodapati - yes we can trust the crypt_data from redo. We could not before, because there was no ENCRYPTING marker. But now the crypt_data in redo is the newest one and all the pages in redo should follow it.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -908,8 +911,7 @@ void fil_space_crypt_t::write_page0(
   memcpy(encrypt_info_ptr, iv, CRYPT_SCHEME_1_IV_LEN);
   encrypt_info_ptr += CRYPT_SCHEME_1_IV_LEN;
 
-  mach_write_to_1(encrypt_info_ptr,
-                  static_cast<byte>(current_encryption_rotation));
+  mach_write_to_1(encrypt_info_ptr, static_cast<byte>(encryption_rotation));
```
satya-bodapati, line 914: if I understand correctly, we now write ENCRYPTING marker to crypt_data. What did we write before?

mysql-test/suite/encryption/t/innodb-redo.test
```diff
@@ -0,0 +1,117 @@
+--source include/have_debug.inc
+
+# PS-7140: crypt redo logs are not applied correctly
+
+--let $MYSQLD_DATADIR=`select @@datadir`
```
satya-bodapati: can you please rename the test to crypt_data_redo.test

satya-bodapati (changes requested):

storage/innobase/fil/fil0crypt.cc
robgolebiowski: NO_ROTATION

mysql-test/suite/encryption/t/innodb-redo.test
robgolebiowski: done

satya-bodapati (approved):
LGTM!
Few checks before merge. Resubmit on jenkins.
1.  Analayze failures and paste failures to github
2. Update github descriptin as well (commit is updated)
3. Happy Merging! :)

storage/innobase/fil/fil0crypt.cc
```diff
@@ -967,9 +967,11 @@ static fil_space_crypt_t *fil_space_set_crypt_data(
 @param[in]  ptr  Log entry start
 @param[in]  end_ptr  Log entry end
 @param[in]  len  Log entry length
+@param[in]  recv_needed_recovery  missing keys will report error
 @return position on log buffer */
 byte *fil_parse_write_crypt_data_v3(space_id_t space_id, byte *ptr,
-                                    const byte *end_ptr, ulint len) {
+                                    const byte *end_ptr, ulint len,
+                                    bool recv_needed_recovery) {
   ptr += 4;  // skip offset and len
```
satya-bodapati, line 975: but upstream also calls fil_tablespace_redo_encryption() always. How does it handle this ? not complain about old keys?

satya-bodapati (changes requested):

storage/innobase/fil/fil0crypt.cc
robgolebiowski: Discussion from Slack:
Robert Golebiowski  1 hour ago
the question is why parse do not always complain about missing keys ?

Satya Bodapati  1 hour ago
yes, how does it choose not to complain or error out?

Robert Golebiowski  1 hour ago
I think it does error out - it just wasn't tested in upstream

Robert Golebiowski  1 hour ago
when "encryption redo" is in redo, but the first page was already flushed

Robert Golebiowski  1 hour ago
@satya - do you want me to test it ?

Satya Bodapati  1 hour ago
how do we make a key invalid? remove after rotation?

Satya Bodapati  1 hour ago
if (!Encryption::decode_encryption_info(key, iv, ptr, true)) {
      recv_sys->found_corrupt_log = true;
      ib::warn(ER_IB_MSG_364)
          << "Encryption information"
          << " in the redo log of space " << space_id << " is invalid";
      return (nullptr);
    }

Satya Bodapati  1 hour ago
in fil_tablespace_redo_encryption()

Satya Bodapati  1 hour ago
so it looks it can fail... :thinking_face:

Satya Bodapati  1 hour ago
yes, then we can create upstream bug

Satya Bodapati  1 hour ago
breaking for lunch.. bbl

Satya Bodapati  1 hour ago
back

Robert Golebiowski  43 minutes ago
kk

Robert Golebiowski  20 minutes ago
yeah, it errors out during redo processing

I have create bug report in upstream : 100491.

PR #3906, opened 2020-07-03
Fixed PS-7019 (Wrong query results for LEFT JOIN with GROUP BY since …

altmannmarcelo:
…8.0.20)

https://jira.percona.com/browse/PS-7019

From https://bugs.mysql.com/bug.php?id=99398:
Refactoring work done in MySQL 8.0.20 caused single-row
buffering for GROUP BY of non-nullable columns not to function
correctly, not taking into account that such a column could be
the inner table for an outer join, and thus would have a NULL
flag that would need to be copied. In a GROUP BY without a
temporary table, this would cause the NULL flag to come from the
next output row instead of the previous one, and data returned
to be inconsistent.

### DIFF
--- a/mysql-test/t/join.test
+++ b/mysql-test/t/join.test
@@ -1411,3 +1411,12 @@ drop table t;
 set optimizer_switch=default;
 
 set global default_storage_engine=myisam;
+--echo #
+--echo BUG #99398 / PS-7019: Wrong query results for LEFT JOIN with GROUP BY since 8.0.20
+--echo #
+CREATE TABLE t1 (t1_id INT PRIMARY KEY AUTO_INCREMENT, t2_id INT DEFAULT NULL);
+CREATE TABLE t2 (t2_id INT PRIMARY KEY AUTO_INCREMENT, is_active TINYINT NOT NULL DEFAULT '1');
+INSERT INTO t2 VALUES (2,1),(3,0),(1000,1);
+INSERT INTO t1 VALUES (1,1000),(2,5);
+SELECT t1.t1_id,t1.t2_id,t2.t2_id FROM t1 LEFT JOIN t2 ON (t1.t2_id = t2.t2_id AND t2.is_active=1) GROUP BY t1_id ORDER BY t1_id LIMIT 2;
+DROP TABLE t1,t2;
--- a/mysql-test/r/join.result
+++ b/mysql-test/r/join.result
@@ -2032,3 +2032,15 @@ a
 drop table t;
 set optimizer_switch=default;
 set global default_storage_engine=myisam;
+#
+BUG #99398 / PS-7019: Wrong query results for LEFT JOIN with GROUP BY since 8.0.20
+#
+CREATE TABLE t1 (t1_id INT PRIMARY KEY AUTO_INCREMENT, t2_id INT DEFAULT NULL);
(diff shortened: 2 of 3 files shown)


altmannmarcelo:
```
marcelo@marce-bld:/work/ps/ins/8.0/mysql-test$ ./mtr join
Logging: ./mtr  join
MySQL Version 8.0.20
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Removing old var directory
Creating var directory '/work/ps/ins/8.0/mysql-test/var'
Installing system database
Using parallel: 1

[100%] main.join                                 [ pass ]   2833
Spent 2.833 of 16 seconds executing testcases

Completed: All 1 tests were successful.
```

altmannmarcelo:
https://ps80.cd.percona.com/job/percona-server-8.0-param/642/

mysql-test/t/join.test
```diff
@@ -1411,3 +1411,11 @@ drop table t;
 set optimizer_switch=default;
 
 set global default_storage_engine=myisam;
+
+# PS-7019 - Wrong query results for LEFT JOIN with GROUP BY since 8.0.20
```
percona-ysorokin: Add here upstream bug number as well and convert comments to `--echo`

mysql-test/t/join.test
```diff
@@ -1411,3 +1411,11 @@ drop table t;
...
+CREATE TABLE `t1` (t1_id int PRIMARY KEY AUTO_INCREMENT, `t2_id` int DEFAULT NULL);
```
percona-ysorokin:
```suggestion
CREATE TABLE t1 (t1_id INT PRIMARY KEY AUTO_INCREMENT, t2_id INT DEFAULT NULL);
```

mysql-test/t/join.test
```diff
@@ -1411,3 +1411,11 @@ drop table t;
...
+CREATE TABLE `t2` (t2_id int PRIMARY KEY AUTO_INCREMENT, `is_active` tinyint NOT NULL DEFAULT '1');
```
percona-ysorokin:
```suggestion
CREATE TABLE t2 (t2_id INT PRIMARY KEY AUTO_INCREMENT, is_active TINYINT NOT NULL DEFAULT 1);
```

mysql-test/t/join.test
```diff
@@ -1411,3 +1411,11 @@ drop table t;
...
+INSERT INTO t2 VALUES (2,1),(3,0),(1000,1);
+INSERT INTO t1 VALUES (1,1000),(2,5);
+SELECT t1.t1_id,t1.t2_id,t2.t2_id FROM t1 LEFT JOIN t2 ON (t1.t2_id = t2.t2_id and t2.is_active=1) group by t1_id order by t1_id LIMIT 2;
```
percona-ysorokin:
```suggestion
SELECT t1.t1_id,t1.t2_id,t2.t2_id FROM t1 LEFT JOIN t2 ON (t1.t2_id = t2.t2_id AND t2.is_active=1) GROUP BY t1_id ORDER BY t1_id LIMIT 2;
```

percona-ysorokin (approved):
LGRM with minor formatting consistency-related comments

altmannmarcelo:
Done, re-recorded join mtr and double-checked it's passing

PR #3923, opened 2020-07-13
PS-7010: Lock buffer blocks before sanity check in btr_cur_latch_leaves (5.7)

kamil-holubicki:
https://jira.percona.com/browse/PS-7010

Problem:
Sanity/integrity check in btr_cur_latch_leaves() for latch_mode == BTR_SEARCH_PREV is done before both buffers are latched. In parallel PAGE_N_HEAP field can be modified with page latched when the row is inserted.

Solution:
Move sanity/integrity check to the point when both validated buffers are latched. This fix is similar to the one done in commit 777bf2aab05fb58fad7838aed4ef307b50722786 for latch_mode == BTR_MODIFY_TREE.

### DIFF
--- a/storage/innobase/btr/btr0cur.cc
+++ b/storage/innobase/btr/btr0cur.cc
@@ -396,13 +396,6 @@ btr_cur_latch_leaves(
 			cursor->left_block = get_block;
 
 			SRV_CORRUPT_TABLE_CHECK(get_block, return latch_leaves;);
-
-#ifdef UNIV_BTR_DEBUG
-			ut_a(page_is_comp(get_block->frame)
-			     == page_is_comp(page));
-			ut_a(btr_page_get_next(get_block->frame, mtr)
-			     == page_get_page_no(page));
-#endif /* UNIV_BTR_DEBUG */
 		}
 
 		latch_leaves.savepoints[1] = mtr_set_savepoint(mtr);
@@ -413,6 +406,13 @@ btr_cur_latch_leaves(
 
 		latch_leaves.blocks[1] = get_block;
 #ifdef UNIV_BTR_DEBUG
+		/* Sanity check only after both the blocks are latched. */
+		if (latch_leaves.blocks[0] != NULL) {
+			ut_a(page_is_comp(latch_leaves.blocks[0]->frame)
+			     == page_is_comp(page));;
+			ut_a(btr_page_get_next(latch_leaves.blocks[0]->frame, mtr)
+			     == page_get_page_no(page));
+		}
 		ut_a(page_is_comp(get_block->frame) == page_is_comp(page));
 #endif /* UNIV_BTR_DEBUG */
 		return(latch_leaves);


storage/innobase/btr/btr0cur.cc
```diff
@@ -413,6 +406,13 @@ btr_cur_latch_leaves(
 
 		latch_leaves.blocks[1] = get_block;
 #ifdef UNIV_BTR_DEBUG
+		/* Sanity check only after both the blocks are latched. */
+		if (latch_leaves.blocks[0] != NULL) {
+			ut_a(page_is_comp(latch_leaves.blocks[0]->frame)
+			     == page_get_page_no(page));
+			ut_a(btr_page_get_next(latch_leaves.blocks[0]->frame, mtr)
+			     == page_get_page_no(page));
+		}
```
satya-bodapati: 1. How is this found? any testcase? because this should happen only special builds when compiled with UNIV_BTR_DEBUG (I think debug builds dont enable this flag by default)

2. How do we know the page is locked here? Can we add the assert that page is locked? 
 mtr_memo_contains_flagged(
              mtr, dict_index_get_lock(index),
              MTR_MEMO_S_LOCK | MTR_MEMO_X_LOCK | MTR_MEMO_SX_LOCK)

Use appropriate lock mode here

satya-bodapati (changes requested):

storage/innobase/btr/btr0cur.cc
kamil-holubicki: HI @satya-bodapati 
1. How is this found? - This is found by customer, reproducible only by customer. Fix was proposed in PR #3862 but without respect to GCA process. As explained the fix was based on identical fix from Oracle done in commit 777bf2a but for another case in the same switch (a few lines above)
UNIV_BTR_DEBUG is defined in univ.i unconditionally.
2. How do we know the page is locked here? - btr_block_get() locks the page

The change was validated by customer (and still is under validation). Please see PS-7010 for details.

satya-bodapati: @kamil-holubicki 

UNIV_BTR_DEBUG is undefined. So enable this, either we modify the univ.i file or pass to cmake cxx/c flags .

Ok, is it possible to add the assertion?

Can you also compile with UNIV_BTR_DEBUG and see if this code is reached via some testcase?

satya-bodapati: may be add ut_a(0), enable UNIV_BTR_DEBUG and run innodb mtr suite?

kamil-holubicki:
Hi @satya-bodapati , This part of code is covered by innodb.innodb_wl6326_big. this is --big-test.
Additionally I discovered a problem inherited from originally proposed patch (which I didn't spot before because I didn't run big-test).
1st assert was wrong. It should be 
`ut_a(page_is_comp(latch_leaves.blocks[0]->frame) == page_is_comp(page));`
instead of
`ut_a(page_is_comp(latch_leaves.blocks[0]->frame) == page_get_page_no(page));`

kamil-holubicki:
big-test run https://ps57.cd.percona.com/job/percona-server-5.7-param/316/

satya-bodapati:
@kamil-holubicki very good thing you spotted.. so me being pain in the __ helped  :stuck_out_tongue_winking_eye: 

Rechecking the PR

satya-bodapati (approved):

PR #3924, opened 2020-07-13
PS-7010: Lock buffer blocks before sanity check in btr_cur_latch_leaves (8.0)

kamil-holubicki:
https://jira.percona.com/browse/PS-7010

Problem:
Sanity/integrity check in btr_cur_latch_leaves() for latch_mode == BTR_SEARCH_PREV is done before both buffers are latched. In parallel PAGE_N_HEAP field can be modified with page latched when the row is inserted.

Solution:
Move sanity/integrity check to the point when both validated buffers are latched. This fix is similar to the one done in commit 777bf2a for latch_mode == BTR_MODIFY_TREE.


kamil-holubicki:
big-test run https://ps80.cd.percona.com/job/percona-server-8.0-param/709/

satya-bodapati (approved):

PR #3927, opened 2020-07-13
5.7 Fixed PS-6995 (Optimizer not picking best execution plan for Primary …

altmannmarcelo:
…Key lookup)

https://jira.percona.com/browse/PS-6995

In some cases optimizer prefers to do a Full Table Scan instead of a
PK range scan.
Introduced a new optimizer switch to allow the user to reduce the cost
of a range scan. If favor_range_scan switch is on, the cost for range
plan will be reduced by 90%. Patch has been provided by Yves Trudeau.


altmannmarcelo:
https://ps57.cd.percona.com/job/percona-server-5.7-param/307/

satya-bodapati (approved):

PR #3929, opened 2020-07-14
PS-7168: Assertion "ctx->trx"

robgolebiowski:
DESCRIPTION:

The bug appeared, because the fix for PS-6922 was incomplete. The
explicit ENCRYPTION='N' should trigger rebuild of a table only in case
when table is file_per_table. PS-6922 was trying to address this issue,
but it failed to recognise that table is not file_per_table when it was
created while global innodb_file_per_table was set to OFF.

FIX:

Whether table is file_per_table is now deduced based on table's flags,
instead of tablespace name.

### DIFF
--- a/mysql-test/suite/encryption/t/explicit_encryption.test
+++ b/mysql-test/suite/encryption/t/explicit_encryption.test
@@ -250,4 +250,42 @@ DROP PROCEDURE innodb_insert_proc;
 DROP DATABASE test_enc_n;
 
 SET GLOBAL innodb_encryption_threads=0;
+SET GLOBAL default_table_encryption=ONLINE_FROM_KEYRING_TO_UNENCRYPTED;
+SET GLOBAL innodb_encryption_threads=4;
+
+--echo # Wait for key encryption threads to decrypt all spaces
+--let $wait_timeout = 600
+--let $wait_condition = SELECT COUNT(*) = 0 FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION = 1
+--source include/wait_condition.inc
+
+SET GLOBAL innodb_encryption_threads=0;
 SET GLOBAL default_table_encryption=OFF;
+
+--echo #
+--echo # PS-7168: Assertion "ctx->trx"
+--echo #
+
+SET GLOBAL innodb_file_per_table=OFF;
+CREATE TABLE t1(ol_o_id int,ol_d_id int,ol_w_id int,ol_number int,ol_tmp int,KEY (ol_w_id,ol_d_id,ol_o_id,ol_number)) PARTITION BY HASH (ol_w_id);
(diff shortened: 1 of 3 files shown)


mysql-test/suite/encryption/t/explicit_encryption.test
```diff
@@ -249,5 +249,41 @@ DROP TABLESPACE ts_altered_to_n;
...
+
+SET GLOBAL innodb_file_per_table=OFF;
+CREATE TABLE t1(ol_o_id int,ol_d_id int,ol_w_id int,ol_number int,ol_tmp int,KEY (ol_w_id,ol_d_id,ol_o_id,ol_number)) PARTITION BY HASH (ol_w_id);
+ALTER TABLE t1 ADD COLUMN v2 CHAR(1),ALGORITHM=INSTANT;
+--echo # Below statement used to crash the server, due to PS-7168
+ALTER TABLE t1 ENCRYPTION='N';
+
+CREATE TABLE t2(ol_o_id int,ol_d_id int,ol_w_id int,ol_number int,ol_tmp int,KEY (ol_w_id,ol_d_id,ol_o_id,ol_number)) PARTITION BY HASH (ol_w_id);
+
+ALTER TABLE t2 TABLESPACE innodb_file_per_table ENCRYPTION='N';
```
satya-bodapati, line 276: can you also add TABLESPACE innodb_system with file_per_table=ON?

storage/innobase/handler/handler0alter.cc
```diff
@@ -662,9 +665,6 @@ static MY_ATTRIBUTE((warn_unused_result)) bool innobase_need_rebuild(
       ha_alter_info->create_info->explicit_encryption,
       ha_alter_info->create_info->encrypt_type.str)};
 
-  const bool is_file_per_table{
-      !tablespace_is_shared_space(ha_alter_info->create_info)};
```
satya-bodapati, line 666: ha_alter_info is created from old table(src) or the new table(dest)?
Also am bit skeptic about m_prebuilt->table. Sometimes it can be nullptr :thinking:

satya-bodapati: Also there is ctx->prebuilt..

satya-bodapati (changes requested):

storage/innobase/handler/handler0alter.cc
robgolebiowski: ha_alter_info->create_info is created based on old table and then can be modify by alter code. ha_alter_info consist of modification to the old table.

satya-bodapati (approved):
Looks harmless. Ok to merge after adding new testcase scenario and pstress approval from @mohitj1988 (DDLs +DMLs + ENCRYPTION=Y/N + no kill only shutdown)

storage/innobase/handler/handler0alter.cc
satya-bodapati: ok thanks

PR #3930, opened 2020-07-14
PS-7169: RocksDB initialization is failing while upgrading the server from 5.7 to 8.0.20

inikep:
1. Don't validate RocksDB DD tables when upgrade from 5.7 is in progress.
2. Add `rocksdb.upgrade_rocksdb_from_5730` test.


inikep:
Tested at https://ps80.cd.percona.com/job/percona-server-8.0-param/672/

george-lorch (approved):
LGTM

PR #3944, opened 2020-07-20
PS-7075 Provide binary tarball with shared libs and glibc suffix

patrickbirch:
modified source/installation
added a note in the "Installing from Binary Tarball" section


kathy-williamson (approved):

hrvojem (approved):
LGTM

PR #3948, opened 2020-07-21
Post merge fix, after merging PS-5730: Change SELECT rotate_system_key

robgolebiowski:
to ALTER INSTANCE for percona system key rotation.


robgolebiowski:
self-approved

PR #3950, opened 2020-07-21
mysqld_safe fails to respawn mysql if the --thp-setting option is set

tonodor:
The code that deals with the --thp-setting option disables and then leaves enabled error checking, causing mysqld_safe to stop if mysql terminates unexpectedly.
As error checking is not enabled in the script there is no need to temporarily disable it, so I've both instructions.

### DIFF
--- a/scripts/mysqld_safe.sh
+++ b/scripts/mysqld_safe.sh
@@ -978,9 +978,7 @@ then
     if [ -f /sys/kernel/mm/transparent_hugepage/enabled ]; then
       CONTENT_THP=$(cat /sys/kernel/mm/transparent_hugepage/enabled)
       STATUS_THP=0
-      set +e
       STATUS_THP=$(echo $CONTENT_THP | grep -cv "\[${thp_setting}\]")
-      set -e
     fi
     if [ $STATUS_THP -eq 0 ]; then
       log_notice "Transparent huge pages are already set to: ${thp_setting}."


scripts/mysqld_safe.sh
```diff
@@ -978,9 +978,7 @@ then
     if [ -f /sys/kernel/mm/transparent_hugepage/enabled ]; then
       CONTENT_THP=$(cat /sys/kernel/mm/transparent_hugepage/enabled)
       STATUS_THP=0
-      set +e
       STATUS_THP=$(echo $CONTENT_THP | grep -cv "\[${thp_setting}\]")
```
EvgeniyPatlan, line 981: It is better to leave it in case errexit is enabled on the system

EvgeniyPatlan (changes requested):
Thank you for your contribution! This PR looks good, there are a couple of edits needed for this to be successfully merged.

Let's imaging that a user have enabled "errexit on" by default on the system. So there are a few cases when $(echo $CONTENT_THP | grep -cv "\[${thp_setting}\]") can return exit code 1.
So with enabled errexit the execution will fail. So it is needed to disable it before command execution.
And I agree that we could skip enabling errexit.

scripts/mysqld_safe.sh
tonodor: Hi @EvgeniyPatlan, yes that makes sense. However, that raises another issue: if errexit is enabled by default on the system and the --thp-setting option is not set, the script will again fail to restart mysql. How about disabling errexit straight away at the beginning of the script?

PR #3953, opened 2020-07-22
PS-7195: After merging PS-5635 - test keyring_and_mk_ecryption_exclus…

robgolebiowski:
…ion started

to fail

PS-5635 introduced updating DD with online_enc_progress flag when
encryption threads are starting to rotate a space.
keyring_and_mk_encryption introduced a hang during ALTER operation on
ts_hang tablespace. ALTER operation locks DD and thus blocked encryption
threads from updating DD with online_enc_progress. Test is watiting for
encryption threads to start hanging also. Since ALTER is waiting for a
signal from a test a deadlock was introduced. In order to break this
deadlock, hang in encryption thread was moved before updating DD with
online_enc_progress flag, which in turns allows test to signal the ALTER
statement that it can continue running. DD is updated by encryption
thread after ALTER statement stops blocking DD for updates.


robgolebiowski:
self-approved
https://ps80.cd.percona.com/job/percona-server-8.0-param/691/

PR #3958, opened 2020-07-22
PS-7076 : Cardinality is equal for all indexes when last partition empty

george-lorch:
This is a regression created by the change for PS-4315 : "Index cardinality
stats updated for handler::info(HA_STATUS_CONST), not often enough for
tokudb_cardinality_scale_percent".  In 5.7, this was fixed in commit
f37b17910038108d14ab2f9071b92aa10c58d43f.  This commit as null merged to 8.0
as the full online analyze patch was not properly and fully ported to 8.0,
it was missing the changes required within the native partitioning code.

This commit is a combination of the 5.7 fix referenced in the commit above
and carrying over the changes needed to the native partitioning code.


george-lorch:
https://ps80.cd.percona.com/job/percona-server-8.0-param/693/

percona-ysorokin (approved):
LGTM

PR #3964, opened 2020-07-27
Fixed PS-7212 (Datatype inside JSON column changes unexpectedly)

altmannmarcelo:
https://jira.percona.com/browse/PS-7212

JSON fields are using Field_blob::cmp_binary to validate if data
has changed. Cherry picked changes from [1] to implement
Field_json::cmp_binary in order to do native JSON comparison.
Added a test case to validate the fix.

[1]:

commit 372d9611a2d47f625530ffdb30dc3b60bcdf6a3f
Author: Knut Anders Hatlen <knut.hatlen@oracle.com>
Date:   Thu Jun 1 13:41:04 2017 +0200

    Bug#26177130: SLAVE CAN'T FIND ROW USING HASH_SCAN
                  IF JSON HAS DIFFERENT BINARY FORMAT

    The same JSON document may have different binary representation after
    WL#8963. Row-based replication uses binary equality to find the
    matching row on the slave, and this may now fail if the binary
    representation of the document is different on the slave than it is on
    the master.

    This fix overrides cmp_binary() in Field_json and makes it use
    Json_wrapper::compare() to compare the JSON values.

    cmp_binary() is also used in UPDATE statements in order to see if a
    record has changed before it is updated in the table. The update is
    skipped if the new record is found to be equal to the original record.
    With these changes, it is now able to skip updates in more cases
    (where the binary representation has changed, but not the contents of
    the document). Some test cases in json_functions_innodb and
    json_functions_ndb, which inspected the binary representation of the
    JSON documents, got changed results because of this.

    Change-Id: If00c4f35d32d2435a148eafc7e0d6a3183bc1b00

### DIFF
--- a/mysql-test/suite/json/r/json_innodb.result
+++ b/mysql-test/suite/json/r/json_innodb.result
@@ -258,3 +258,21 @@ SUM(col_int)	col_int
 4572643328	NULL
 DROP VIEW view1;
 DROP TABLE D;
+#
+# BUG: PS-7212 - Datatype inside JSON column changes unexpectedly
+#
+CREATE TABLE json (
+id int NOT NULL,
+j json DEFAULT NULL,
+s tinyint NOT NULL,
+PRIMARY KEY (id)
+) ENGINE=InnoDB;
+INSERT INTO json VALUES (2, '{"amount": 70.0}', 1);
+SELECT json_type(j->"$.amount") FROM json;
+json_type(j->"$.amount")
+DOUBLE
+INSERT INTO json VALUES (2, '{}', 1) ON DUPLICATE KEY UPDATE j=if(s=0, '{}', j);
+SELECT json_type(j->"$.amount") FROM json;
+json_type(j->"$.amount")
+DOUBLE
+DROP TABLE json;
--- a/mysql-test/suite/json/t/json_innodb.test
+++ b/mysql-test/suite/json/t/json_innodb.test
@@ -249,3 +249,18 @@ DROP TABLE D;
 
 --source include/restore_sql_mode_after_turn_off_only_full_group_by.inc
 
+--echo #
+--echo # BUG: PS-7212 - Datatype inside JSON column changes unexpectedly
+--echo #
+
+CREATE TABLE json (
+id int NOT NULL,
+j json DEFAULT NULL,
+s tinyint NOT NULL,
+PRIMARY KEY (id)
+) ENGINE=InnoDB;
+INSERT INTO json VALUES (2, '{"amount": 70.0}', 1);
+SELECT json_type(j->"$.amount") FROM json;
+INSERT INTO json VALUES (2, '{}', 1) ON DUPLICATE KEY UPDATE j=if(s=0, '{}', j);
+SELECT json_type(j->"$.amount") FROM json;
+DROP TABLE json;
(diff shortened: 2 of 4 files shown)


altmannmarcelo:
https://ps57.cd.percona.com/job/percona-server-5.7-param/313/

altmannmarcelo:
Note for reviewers: Test case from original 8.0 commit (372d9611a2d47f625530ffdb30dc3b60bcdf6a3f - rpl_json) didn't have any effect on 5.7, it was passing with and without the fix, so I decided to no include it as part of cherry-pick.

satya-bodapati (commented):
rpl_json.test from the 8.0 commit can also be ported back to 5.7? 

Also, what is the replication impact(only with ROW format I believe)? ie. Server without fix <-> Server with fix

altmannmarcelo:
@satya-bodapati 
> rpl_json.test from the 8.0 commit can also be ported back to 5.7?

`rpl_json.test` does not apply here on 5.7 because it requires partial JSON updates to trigger the issue. Partial JSON updates were done in 8.0 via https://dev.mysql.com/worklog/task/?id=8963

`rpl_json.test` test the following condition.
It goes to the master, disables binary log, adds a json with values `INSERT INTO t VALUES ('["abc", "abcdefghij"]');` then it updates the JSON so the first element in the document gets updates from `abc` to `a` (using `UPDATE t SET j = JSON_SET(j, '$[0]', 'a');`) and here is where the diff from 5.7 and 8.0 comes from. On 8.0 this will replace the first element inline leaving two blank spaces / null on it (`abc` -> `a`), this causes the binary version of the document to compute those blank spaces. While in 5.7 it replaces the whole document. 
```
# 8.0 after partial JSON update
mysql> SELECT j, JSON_STORAGE_SIZE(j) FROM t;
+---------------------+----------------------+
| j                   | JSON_STORAGE_SIZE(j) |
+---------------------+----------------------+
| ["a", "abcdefghij"] |                   26 |
+---------------------+----------------------+
1 row in set (0.01 sec)

# 5.7 after JSON update
mysql> SELECT j, JSON_STORAGE_SIZE(j) FROM t;
+---------------------+----------------------+
| j                   | JSON_STORAGE_SIZE(j) |
+---------------------+----------------------+
| ["a", "abcdefghij"] |                   24 |
+---------------------+----------------------+
1 row in set (0.00 sec)
```
Test case continues by manually inserting the same document on the slave (`INSERT INTO t VALUES ('["a", "abcdefghij"]');`), since it did not have the document already because we were operating on the master with binlog disabled. Then it moves back to master connection, enable binlog and delete the document from the table, at this point, on 8.0 the slave cannot find the row because it is using `Field_blob::cmp_binary` to compare the rows and they have different binary representation, while on 5.7 the binary version is equal on master and slave.

> Also, what is the replication impact(only with ROW format I believe)? ie. Server without fix <-> Server with fix

I cannot see any impact on 5.7 even for ROW format. The only difference I can see (on the positive side and it's what this PR is trying to fix) is that for a query that should not change the JSON document (test case I added on this PR) it will no longer generate a ROW event since the data does not need to change.

satya-bodapati (approved):
Thanks for the detailed explanation.
Also verified test results from Jenkins link.. looks like they are existing failures (certs expired issue) and other ones are known unstable ones

LGTM!

percona-ysorokin (approved):
LGTM

PR #3971, opened 2020-07-30
PS-7220: Fix activity counter update in purge coordinator and workers. [5.7]

oleksandr-kachan:
The upstream fix for the bug #30875956 "PURGE ACTIVITY IS TOO MUCH AT
CPU% BOUND AND CAUSES IDLE OVERDRIVE"
https://github.com/mysql/mysql-server/commit/e52529d2adecadf1fde929517bbc7a73da1ed086
reveals an issue with an improvement "Server activity check fixes for page cleaner thread"
https://github.com/percona/percona-server/commit/05255ba87e77c46cc84b1611af2ab8dc0d32d51f.

With the upstream patch purge coordinator will be woken up once a second
from srv_master_do_active_tasks and as a side effect of our improvement
it will increase activity counter each time. The combination of these two makes
page cleaner thread do adpative flushing instead of background flushing (which PCT(100) and uses
full io_capacity). With adaptive flushing, the amount of pages to be flushed is dependent
on checkpoint age, free page list length and dirty page ratio. So this makes flusher not
to flush all pages and causes the timeout in mtr testcase.

Fix:
To fix the issue srv_inc_activity_count inside of srv_purge_coordinator_thread
will be invoked only in case there were actual page purges. In addition,
remove the redundant srv_inc_activity_count increment in srv_task_execute().
Purge coordinator increases it anyway if it has purged something.

### DIFF
--- a/storage/innobase/srv/srv0srv.cc
+++ b/storage/innobase/srv/srv0srv.cc
@@ -2967,8 +2967,6 @@ srv_task_execute(void)
 
 		os_atomic_inc_ulint(
 			&purge_sys->pq_mutex, &purge_sys->n_completed, 1);
-
-		srv_inc_activity_count();
 	}
 
 	return(thr != NULL);
@@ -3316,7 +3314,9 @@ DECLARE_THREAD(srv_purge_coordinator_thread)(
 		rseg_history_len = srv_do_purge(
 			srv_n_purge_threads, &n_total_purged);
 
-		srv_inc_activity_count();
+		if (n_total_purged != 0) {
+			srv_inc_activity_count();
+		}
 
 	} while (!srv_purge_should_exit(n_total_purged));
 


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/314/

percona-ysorokin (approved):
LGTM

looks good from my side but this definitely needs @satya-bodapati opinion

storage/innobase/srv/srv0srv.cc
```diff
@@ -3316,7 +3314,9 @@ DECLARE_THREAD(srv_purge_coordinator_thread)(
 		rseg_history_len = srv_do_purge(
 			srv_n_purge_threads, &n_total_purged);
 
-		srv_inc_activity_count();
+		if (n_total_purged != 0) {
+			srv_inc_activity_count();
+		}
```
satya-bodapati, line 3319: How upstream patch affects the activity count? Because upstream only adds srv_wake_purge_thread_if_not_active(). There is no activity counter bump inside this. This only wakes up purge thread.

storage/innobase/srv/srv0srv.cc
```diff
@@ -2967,8 +2967,6 @@ srv_task_execute(void)
 
 		os_atomic_inc_ulint(
 			&purge_sys->pq_mutex, &purge_sys->n_completed, 1);
-
-		srv_inc_activity_count();
```
satya-bodapati, line 2971: so the old logic bumped the activity twice for the purge.
1) when the workers completed. ie.  number_of_pages from batch/total_worker times.. 
2) in coordinator when the total pages handled by a batch is not zero.

So interestingly, the not so accurate activity count by purge didn't influence the flusher that much..

satya-bodapati (changes requested):

storage/innobase/srv/srv0srv.cc
oleksandr-kachan: @satya-bodapati, as I understand it wakes srv_purge_coordinator_thread up which in turn increases activity counter. Am I missing something here?

storage/innobase/srv/srv0srv.cc
oleksandr-kachan: Looks like the main reason for this to become visible is time points when the purge thread becomes active. With upstream patch purge thread will be woken up once a second and at the same time flusher is supposed to clean cache after one second if there is not any activity. As a result we get some kind of a conflict at this point.
Do you think we could get unexpected effects after this fix? Is there anything you could suggest to recheck?

storage/innobase/srv/srv0srv.cc
satya-bodapati: Ok, fair enough.

Can you please tell me how this change in activity impacted flusher decision? ie to not flush?

oleksandr-kachan: In buf_flush_page_cleaner_coordinator we have "next_loop_time = ut_time_monotonic_ms() + 1000". There may be different conditions for this next_loop_time update but in most cases it just does "next_loop_time = curr_time + 1000" in each worker cycle where server activity is detected. The page cleaner is supposed to flush all pages when there is not any server activity detected during the time interval determined by next_loop_time (if (ut_time_monotonic_ms() > next_loop_time)).
Now with upstream changes purge thread increases activity counter once a second. Once buf_flush_page_cleaner_coordinator is about to meet conditions for full pages flush, the purge thread gets woken up and increases activity counter which in turn leads to next_loop_time update and also makes buf_flush_page_cleaner_coordinator to choose a different "if" brunch (else if (srv_check_activity(last_activity))) for buffer processing.

storage/innobase/srv/srv0srv.cc
satya-bodapati: AFAIU, the main reason for this counters for purge is not make server think it is idle.
If server thinks, it is idle, it will consider flushing at full io cpacity. and this is bad if there are some other real flushing happening.

So activity counters help server to flush in more adaptive manner (adaptive flushing/ flush_list flushing)

percona-ysorokin: @satya-bodapati Both of these calls to `srv_inc_activity_count()` were added by Laurynas as an IMPROVEMENT in this commit
https://github.com/percona/percona-server/commit/05255ba87e77c46cc84b1611af2ab8dc0d32d51f
Worst case scenario, even if we remove this call here, we will be just as good as upstream.
Nevertheless, we can potentially do something similar to what we did in the `srv_purge_coordinator_thread` - call `srv_inc_activity_count()` only if there was actual activity. However, in this case we will have to extend `QUE_NODE_PURGE` queue task and probably its dependent data structure to hold the result of task execution and somehow extract it here from the `thr`

satya-bodapati: @percona-ysorokin it looks to me that the activity counters in purge are totally un-necessary with dedicated LRU threads. At the time of commit 05255ba, there was coupling in page cleaner thread that it has to do LRU flushes from same thread.

Since PS has dedicated LRU flushing, IMHO we can get rid of these.

Also, from the minimal mtr testcase you gave, it does only INSERTs which doesn't add to purge history length. So ideally purge shouldn't have a role here. 

Neverthless, am trying to understand what caused purge to interfere.

satya-bodapati (changes requested):
It looks safe to me. 

So insert was just used to dirty pages and has no relation to purge. The purge now wakes up every second and our unconditional activity counter bump made the flusher go to regular adaptive flushing mode. And in this mode, the amount of pages to be flushed is determind by checkpoint age, free list length and dirty page ratio.  So the testcase never does the background flushing PCT(100) ( uses the full IO capacity).

So the fix to make purge activity counter unconditional is correct.

Like I said somewhere in the comment, since we have dedicated LRU threads, we could rid of our acitivity counters in purge thread.. but this needs a bit more evaluation..

So even if purge is active (really purging somethig frequently) quite a lot, the difference now is that we flusher does more of adpative flushing than the background flushing (PCT(100)).

The conclusion is the patch is safe!

Some formalities before I can approve

1. Jenkins link
2. Paste the test case output before and after the fix (or just mention the test name to verify in Jenkins)

storage/innobase/srv/srv0srv.cc
satya-bodapati: thanks

satya-bodapati:
@oleksandr-kachan suggested commit message (slightly modifed your version):

The upstream fix for the bug #30875956 "PURGE ACTIVITY IS TOO MUCH AT
CPU% BOUND AND CAUSES IDLE OVERDRIVE"
mysql/mysql-server@e52529d
reveals an issue with an improvement "Server activity check fixes for page cleaner thread"
05255ba.

With the upstream patch purge coordinator will be woken up once a second
from srv_master_do_active_tasks and as a side effect of our improvement
it will increase activity counter each time. The combination of these two makes
page cleaner thread do adpative flushing instead of background flushing (which PCT(100) and uses
full io_capacity). With adaptive flushing, the amount of pages to be flushed is dependent
on checkpoint age, free page list length and dirty page ratio. So this makes flusher not
to flush all pages and causes the timeout in mtr testcase.

Fix:
----

To fix the issue srv_inc_activity_count inside of srv_purge_coordinator_thread
will be invoked only in case there were actual page purges. In addition,
remove the redundant srv_inc_activity_count increment in srv_task_execute().
Purge coordinator increases it anyway if it has purged something.

satya-bodapati (approved):
Please update the commit message with the new suggested message.
Also, update Github description

PR #3987, opened 2020-08-11
[5.7] PS-7231 Assertion failure `m_status == DA_ERROR` in `Diagnostics_area::mysql_errno()`

venkatesh-prasad-v:
https://jira.percona.com/browse/PS-7231

This is a subtask of PS-7197: Multi-threaded slave hangs when slave_trans_retires gets exhausted

Background
----------
In general, if a replication applier thread fails to execute a
transaction because of an InnoDB deadlock or because the transaction's
execution time exceeded InnoDB's innodb_lock_wait_timeout, it
automatically retries slave_transaction_retries times before stopping
with an error.

And when --slave_preserve_commit_order is enabled, the replica server
ensures that transactions are externalized on the replica in the same
order as they appear in the replica's relay log, and prevents gaps in
the sequence of transactions that have been executed from the relay log.
If a thread's execution is completed before its preceding thread, then
the executing thread waits until all previous transactions are committed
before committing.

Problem & Analysis
------------------
When --slave_preserve_commit_order is enabled on slave and if the
waiting thread has locked the rows which are needed by the thread
executing the previous transaction(as per relay log), then the innodb
deadlock detection algorithm detects the deadlock between workers and
will ask the waiting thread to rollback (only if its sequence number is
lesser than that of the waiting thread).

When this happens, the waiting thread wakes up from the cond_wait(SPCO)
and it gets to know that it was asked to rollback by its preceding
transaction as it was holding a lock that is needed by the other
transaction to progress. It then rolls back its transaction so that the
the preceding transaction can be committed and retries the transaction.

But just before the transaction is retried, the worker checks if it
encountered any errors during its execution. If there is no error, it
simulates ER_LOCK_DEADLOCK error in order for it to be considered as a
temporary error so that the worker thread retries the transaction.

However, when the retries are exhausted, the worker thread logs an error
into the error log by accessing the thread's diagnostic_area by calling
`thd->get_stmt_da()->mysql_errno()`. If the error had been simulate (not
called through `my_error` function call), the diagnostic_area would
still be empty and thus making the assertion `DBUG_ASSERT(m_status ==
DA_ERROR);` to fail.

NOTE: This assertion is observed only when both
slave_transaction_retries and slave_preserve_commit_order are enabled on
the replica server and is more likely to happen when
slave_transaction_retries is set to a lower value.

Fix
---
`Slave_transaction:::retry_transaction()` has been modified to call
`thd->get_stmt_da()->mysql_errno()` only when `thd->is_error()` is
evaluated to true.

Testing
-----
v1 - Jenkins: https://ps57.cd.percona.com/job/percona-server-5.7-param/321/testReport/
Summary: Failing tests are unrelated to the current patch.

### DIFF
--- a/include/violite.h
+++ b/include/violite.h
@@ -146,6 +146,7 @@ Vio* vio_new_win32shared_memory(HANDLE handle_file_map,
 
 void vio_proxy_protocol_add(const struct st_vio_network *net);
 void vio_proxy_cleanup();
+void vio_force_skip_proxy(Vio *vio);
 void    vio_delete(Vio* vio);
 int vio_shutdown(Vio* vio, int how);
 int vio_cancel(Vio* vio, int how);
@@ -306,6 +307,7 @@ struct st_vio
   size_t addrLen;                       /* Length of remote address */
   enum enum_vio_type    type;           /* Type of connection */
   my_bool               inactive; /* Connection inactive (has been shutdown) */
+  my_bool               force_skip_proxy;
   char                  desc[VIO_DESCRIPTION_SIZE]; /* Description string. This
                                                       member MUST NOT be
                                                       used directly, but only
--- a/mysql-test/r/proxy_protocol_extra_port.result
+++ b/mysql-test/r/proxy_protocol_extra_port.result
@@ -1,0 +1,5 @@
+include/assert.inc ['Assert we are binding on all networks']
+# An unproxied client connection should be accepted when connection to an extra_port if proxying enabled on the server
+SHOW GLOBAL STATUS LIKE 'Threads_connected';
+Variable_name	Value
+Threads_connected	1
--- a/mysql-test/t/proxy_protocol_extra_port.cnf
+++ b/mysql-test/t/proxy_protocol_extra_port.cnf
@@ -1,0 +1,13 @@
+!include include/default_my.cnf
+
+[mysqld.1]
+max-connections= 2
+extra-port= @OPT.port
+extra-max-connections= 1
+proxy_protocol_networks=*
+
+[client]
+connect-timeout= 2
+
+[ENV]
+MASTER_EXTRA_PORT= @OPT.port
--- a/mysql-test/t/proxy_protocol_extra_port.test
+++ b/mysql-test/t/proxy_protocol_extra_port.test
@@ -1,0 +1,16 @@
+--source include/not_embedded.inc
+--let $assert_text= 'Assert we are binding on all networks'
+--let $assert_cond= "[SELECT @@GLOBAL.proxy_protocol_networks]" = "*"
+--source include/assert.inc
+
+--source include/count_sessions.inc
+
+
+--echo # An unproxied client connection should be accepted when connection to an extra_port if proxying enabled on the server
+connect (conn1,"127.0.0.1",root,,,$MASTER_EXTRA_PORT,);
+--connection conn1
+SHOW GLOBAL STATUS LIKE 'Threads_connected';
+
+--disconnect conn1
+--connection default
+--source include/wait_until_count_sessions.inc
--- a/sql/sql_connect.cc
+++ b/sql/sql_connect.cc
@@ -1130,6 +1130,10 @@ static int check_connection(THD *thd, bool extra_port_connection)
     char ip[NI_MAXHOST];
     LEX_CSTRING main_sctx_ip;
 
+    if (extra_port_connection) {
+      vio_force_skip_proxy(net->vio);
+    }
+
     peer_rc= vio_peer_addr(net->vio, ip, &thd->peer_port, NI_MAXHOST);
 
     /*
--- a/vio/vio.c
+++ b/vio/vio.c
@@ -113,6 +113,7 @@ static void vio_init(Vio *vio, enum enum_vio_type type,
   memset(vio, 0, sizeof(*vio));
   vio->type= type;
   vio->mysql_socket= MYSQL_INVALID_SOCKET;
+  vio->force_skip_proxy= FALSE;
   mysql_socket_setfd(&vio->mysql_socket, sd);
   vio->localhost= flags & VIO_LOCALHOST;
   vio->read_timeout= vio->write_timeout= -1;
--- a/vio/viosocket.c
+++ b/vio/viosocket.c
@@ -717,6 +717,10 @@ void vio_proxy_cleanup()
   my_free(vio_pp_networks);
 }
 
+void vio_force_skip_proxy(Vio *vio) {
+  vio->force_skip_proxy = TRUE;
+}
+
 /* Check whether a connection from this source address must provide the proxy
 protocol header */
 static my_bool vio_client_must_be_proxied(const struct sockaddr *addr)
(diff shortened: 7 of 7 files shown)


mysql-test/suite/rpl/t/rpl_mts_spco_deadlock_slave_trans_retries_assertion.test
```diff
@@ -0,0 +1,159 @@
...
+# 7. Cleanup.
+#
+# ==== References ====
+#
+# PS-7231 Assertion failure `m_status == DA_ERROR` in `Diagnostics_area::mysql_errno()`
+# PS-7197 Multi-threaded slave randomly stacks
+
+# The test needs debug binaries
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
```
venkatesh-prasad-v: Self Review: Sourcing `have_debug_sync.inc` is not required.

mysql-test/suite/rpl/t/rpl_mts_spco_deadlock_slave_trans_retries_assertion.test
```diff
@@ -0,0 +1,159 @@
...
+SET @saved_slave_parallel_workers = @@GLOBAL.slave_parallel_workers;
+SET @saved_slave_preserve_commit_order = @@GLOBAL.slave_preserve_commit_order;
+SET @saved_innodb_lock_wait_timeout = @@GLOBAL.innodb_lock_wait_timeout;
+SET @saved_slave_transaction_retries = @@GLOBAL.slave_transaction_retries;
+
+SET GLOBAL slave_transaction_retries = 2;
+SET GLOBAL slave_parallel_type = "LOGICAL_CLOCK";
+SET GLOBAL slave_parallel_workers = 2;
+SET GLOBAL slave_preserve_commit_order = ON;
+# Set it a long time to guarantee it doens't report an timeout error
```
venkatesh-prasad-v: Self Review:
1. s/it a/it to a/
2. s/doens't/doesn't/
3. /an/a/

mysql-test/suite/rpl/t/rpl_mts_spco_deadlock_slave_trans_retries_assertion.test
```diff
@@ -0,0 +1,159 @@
...
+#
+#    On replica server, they are applied parallel by w1 and w2. And the deadlock is:
+#      w1                                        w2
+#    -------------------------                 --------------------------
+#     applying INSERT                          applying DELETE
+#                                              get the row locks first.
+#     waiting for w1 to release
+#     the row lock.
+#                                              waiting for w2 to commit
+#
```
satya-bodapati: in this example, why w1 would wait for itself? is it mistake  ? s/w1/w2?

similary on the otherside w2 is waiitng for itself to commit? is it w1? "waitiing for w1 to commit"?

mysql-test/suite/rpl/t/rpl_mts_spco_deadlock_slave_trans_retries_assertion.test
```diff
@@ -0,0 +1,159 @@
...
+#    log a message in the error log.
+# 5. Wait till the co-ordinator thread to error out with ER_LOCK_DEADLOCK.
+# 6. Restart replication threads, sync the replica and verify that table data is
+#    consistent.
+# 7. Cleanup.
+#
+# ==== References ====
+#
+# PS-7231 Assertion failure `m_status == DA_ERROR` in `Diagnostics_area::mysql_errno()`
+# PS-7197 Multi-threaded slave randomly stacks
```
satya-bodapati: 7197 has any role here?
also I guess the title should have randomly hangs instead of "stacks".. may be they meant stuck.

satya-bodapati (changes requested):
Looks good. few questions..

mysql-test/suite/rpl/t/rpl_mts_spco_deadlock_slave_trans_retries_assertion.test
venkatesh-prasad-v: This is a type error. It should have been  
```
#    On replica server, they are applied parallel by w1 and w2. And the deadlock is:
#      w1                                        w2
#    -------------------------                 --------------------------
#                                              get the row locks first.
#     waiting for w2 to release
#     the row lock.
#                                              waiting for w1 to commit
#
```

mysql-test/suite/rpl/t/rpl_mts_spco_deadlock_slave_trans_retries_assertion.test
venkatesh-prasad-v: I mentioned PS-7197 because PS-7231 is marked as a subtask of PS-7197 and I was not sure about if I should mention only the base bug or the subtask.

And regarding the title, it should have been `stuck` and not `stack`. Will update the JIRA title now.

mysql-test/suite/rpl/t/rpl_mts_spco_deadlock_slave_trans_retries_assertion.test
venkatesh-prasad-v: This is a type error. It should have been  
```
#    On replica server, they are applied parallel by w1 and w2. And the deadlock is:
#      w1                                        w2
#    -------------------------                 --------------------------
#                                              get the row locks first.
#     waiting for w2 to release
#     the row lock.
#                                              waiting for w1 to commit
#
```

mysql-test/suite/rpl/t/rpl_mts_spco_deadlock_slave_trans_retries_assertion.test
venkatesh-prasad-v: I mentioned PS-7197 because PS-7231 is marked as a subtask of PS-7197 and I was not sure about if I should mention only the base bug or the subtask.

And regarding the title, it should have been `stuck` and not `stack`. I will update the JIRA title now.

venkatesh-prasad-v: Changed it to `PS-7197 Multi-threaded slave hangs when slave_trans_retires gets exhausted`

mysql-test/suite/rpl/t/rpl_mts_spco_deadlock_slave_trans_retries_assertion.test
```diff
@@ -0,0 +1,158 @@
...
+# 5. Wait till the co-ordinator thread to error out with ER_LOCK_DEADLOCK.
+# 6. Restart replication threads, sync the replica and verify that table data is
+#    consistent.
+# 7. Cleanup.
+#
+# ==== References ====
+#
+# PS-7231 Assertion failure `m_status == DA_ERROR` in `Diagnostics_area::mysql_errno()`
+# PS-7197 Multi-threaded slave hangs when slave_trans_retires gets exhausted
+
```
satya-bodapati: remove this.. not necessary

satya-bodapati (approved):
LGTM! 

Good work on the testcase

mysql-test/suite/rpl/t/rpl_mts_spco_deadlock_slave_trans_retries_assertion.test
venkatesh-prasad-v: Removed.

PR #3991, opened 2020-08-17
PS-7125 feature: Reloading X ssl certificates on ALTER INSTANCE RELOAD TLS

dutow:
https://jira.percona.com/browse/PS-7125

Issue: MySQL lets users change certificates while the server is running,
but this change isn't propagated to the X plugin. X connections still
keep using the old certificate.

Fix: this commit adds a callback to the plugin interface, notifying
plugins if the user executed ALTER INSTANCE RELOAD TLS. The X plugin
implements this callback, and reloads certificates correctly. Existing
connections also keep using the previous certificate, similarly to
normal connections.

### DIFF
--- a/include/mysql/plugin.h
+++ b/include/mysql/plugin.h
@@ -994,6 +994,10 @@ void thd_kill(unsigned long id);
 */
 int thd_get_ft_query_extra_word_chars(void);
 
+typedef bool (*ssl_reload_callback_t)(void *);
+bool register_ssl_reload_callback(ssl_reload_callback_t);
+bool deregister_ssl_reload_callback(ssl_reload_callback_t);
+
 #ifdef __cplusplus
 }
 #endif
--- a/mysql-test/suite/x/t/ssl_dynamic.test
+++ b/mysql-test/suite/x/t/ssl_dynamic.test
@@ -1,0 +1,106 @@
+
+--source include/xplugin_preamble.inc
+--source include/xplugin_create_user.inc
+
+## Test starts here
+
+--let $xtest_file= $MYSQL_TMP_DIR/connection_auth_plugin.tmp
+--write_file $xtest_file
+-->quiet
+-->sql
+SELECT user, plugin FROM mysql.user WHERE user = REPLACE(USER(), '@localhost', '');
+SHOW STATUS LIKE 'Mysqlx_ssl_active';
+-->endsql
+EOF
+
+CREATE USER xuser_sha256@localhost
+            IDENTIFIED WITH 'sha256_password' BY 'sha256';
+GRANT ALL ON *.* TO xuser_sha256@localhost;
+
+--let $XBASIC= --file=$xtest_file 2>&1
+--let $XSSLPARAM= --ssl-cipher='DHE-RSA-AES256-SHA' $XBASIC
+
+# The SSL library may fail initializing during this one
+call mtr.add_suppression("Failed to set up SSL because of the following SSL library error");
+call mtr.add_suppression("mysqlx reported: 'Failed at SSL configuration");
+
+--echo Ssl connection by sha256 plugin  -> success
+--exec $MYSQLXTEST -uxuser_sha256 -psha256 $XSSLPARAM
+
+--echo ################## Test reload
+ALTER INSTANCE RELOAD TLS;
+
+--echo Ssl connection by sha256 plugin  -> success
+--exec $MYSQLXTEST -uxuser_sha256 -psha256 $XSSLPARAM
+
+--echo ################## Test wrong value
+
+--echo # Save the defaults
+SET @orig_ssl_ca= @@global.ssl_ca;
+
+--echo # Set CA to invalid value
+SET GLOBAL ssl_ca = 'gizmo';
+
+--echo Ssl connection by sha256 plugin  -> success
+--exec $MYSQLXTEST -uxuser_sha256 -psha256 $XSSLPARAM
+
+--echo # Must fail and not change the SSL params
+--error ER_DA_SSL_LIBRARY_ERROR
+ALTER INSTANCE RELOAD TLS;
+
+--echo # Must be 1
+SELECT COUNT(*) FROM performance_schema.session_status
+WHERE VARIABLE_NAME = 'Current_tls_ca' AND VARIABLE_VALUE = @orig_ssl_ca;
+
+--echo Ssl connection by sha256 plugin  -> success
+--exec $MYSQLXTEST -uxuser_sha256 -psha256 $XSSLPARAM
+
+#
+--echo # Must pass with a warning and disable SSL
+ALTER INSTANCE RELOAD TLS NO ROLLBACK ON ERROR;
+
+--echo Ssl connection by sha256 plugin  -> failure
+--error 1
+--exec $MYSQLXTEST -uxuser_sha256 -psha256 $XSSLPARAM
+
+--echo # cleanup
+SET GLOBAL ssl_ca = @orig_ssl_ca;
+ALTER INSTANCE RELOAD TLS;
+
+# Must work
+
+
+--echo ################## Change certificate
+
+--echo # Save the defaults
+SET @orig_ssl_ca= @@global.ssl_ca;
+SET @orig_ssl_cert= @@global.ssl_cert;
+SET @orig_ssl_key= @@global.ssl_key;
+SET @orig_mysqlx_ssl_ca= @@global.mysqlx_ssl_ca;
+SET @orig_mysqlx_ssl_cert= @@global.mysqlx_ssl_cert;
+SET @orig_mysqlx_ssl_key= @@global.mysqlx_ssl_key;
+
+--echo # setting new values for ssl_cert, ssl_key and ssl_ca
+--replace_result "$MYSQL_TEST_DIR" MYSQL_TEST_DIR
+eval SET GLOBAL ssl_cert = "$MYSQL_TEST_DIR/std_data/server-cert-sha512.pem";
+--replace_result "$MYSQL_TEST_DIR" MYSQL_TEST_DIR
+eval SET GLOBAL ssl_key = "$MYSQL_TEST_DIR/std_data/server-key-sha512.pem";
+--replace_result "$MYSQL_TEST_DIR" MYSQL_TEST_DIR
+eval SET GLOBAL ssl_ca = "$MYSQL_TEST_DIR/std_data/ca-sha512.pem";
+ALTER INSTANCE RELOAD TLS;
+
+--echo Ssl connection by sha256 plugin  -> success
+--exec $MYSQLXTEST -uxuser_sha256 -psha256 $XSSLPARAM
+#
+--echo # cleanup
+SET GLOBAL ssl_cert = @orig_ssl_cert;
+SET GLOBAL ssl_key = @orig_ssl_key;
+SET GLOBAL ssl_ca = @orig_ssl_ca;
+ALTER INSTANCE RELOAD TLS;
+
+
+--echo ################## End of dynamic SSL tests
+
+DROP USER xuser_sha256@localhost;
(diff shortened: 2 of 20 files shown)


mysql-test/suite/x/t/ssl_dynamic.test
```diff
@@ -0,0 +1,106 @@
...
+
+--echo Ssl connection by sha256 plugin  -> success
+--exec $MYSQLXTEST -uxuser_sha256 -psha256 $XSSLPARAM
+
+--echo ################## Test wrong value
+
+--echo # Save the defaults
+SET @orig_ssl_ca= @@global.ssl_ca;
+
+--echo # Seet CA to invalid value
```
percona-ysorokin:
```suggestion
--echo # Set CA to invalid value
```

include/mysql/plugin.h
```diff
@@ -995,6 +995,10 @@ void thd_kill(unsigned long id);
 */
 int thd_get_ft_query_extra_word_chars(void);
 
+using ssl_reload_callback_t = void(*)();
+int register_ssl_reload_callback(ssl_reload_callback_t);
```
percona-ysorokin: You do not use `register_ssl_reload_callback()` / `deregister_ssl_reload_callback()` return value anywhere. So, probably, `void` return type here?
If still want to check the return value, then `bool` will suite better here.

plugin/x/src/server/server.cc
```diff
@@ -141,6 +141,11 @@ void Server::delayed_start_tasks() {
   });
 }
 
+void Server::reload_ssl_context() {
+  fprintf(stderr, "Reloading ssl stuff!\n");
```
percona-ysorokin: Is this intentional?

plugin/x/src/server/server.cc
```diff
@@ -186,6 +191,7 @@ bool Server::prepare() {
 /** Stop the network acceptor loop */
 void Server::stop(const bool is_called_from_timeout_handler) {
   m_stop_called = true;
+
```
percona-ysorokin: Spurious change

plugin/x/ngs/include/ngs/client.h
```diff
@@ -151,6 +151,7 @@ class Client : public xpl::iface::Client {
   uint16_t m_client_port;
   std::atomic<Client::State> m_state;
   std::atomic<bool> m_removed;
+  std::shared_ptr<xpl::iface::Ssl_context> m_ssl;
```
percona-ysorokin, line 154: `m_ssl` is not used anywhere after assignment in `Client::activate_tls()`. Is this a leftover from previous experiments?

include/mysql/plugin.h
```diff
@@ -995,6 +995,10 @@ void thd_kill(unsigned long id);
 */
 int thd_get_ft_query_extra_word_chars(void);
 
+using ssl_reload_callback_t = void(*)();
```
percona-ysorokin: Just in case, you may want to make the signature `void(*)(void*)` for future possible extensions without changing the interface, so that the callback could be invoked with some `void *cb_data`;

plugin/x/src/interface/server.h
```diff
@@ -67,7 +68,7 @@ class Server {
 
   virtual xpl::Mutex &get_client_exit_mutex() = 0;
 
-  virtual Ssl_context *ssl_context() const = 0;
+  virtual std::shared_ptr<Ssl_context> ssl_context() const = 0;
```
percona-ysorokin, line 71: It looks like changing the signature of this method is not needed if you are not going to save the context in `xpl::iface::Client::m_ssl`

plugin/x/src/server/server.h
```diff
@@ -149,7 +150,7 @@ class Server : public xpl::iface::Server {
   std::shared_ptr<Protocol_global_config> m_config;
   std::unique_ptr<xpl::iface::Document_id_generator> m_id_generator;
 
-  std::unique_ptr<xpl::iface::Ssl_context> m_ssl_context;
+  std::shared_ptr<xpl::iface::Ssl_context> m_ssl_context;
```
percona-ysorokin, line 153: Likewise, no need to change this

sql/sql_admin.cc
```diff
@@ -1651,6 +1651,8 @@ bool Sql_cmd_shutdown::execute(THD *thd) {
   return res;
 }
 
+using ssl_reload_callback_t = void (*)();
```
percona-ysorokin: Likewise
```suggestion
using ssl_reload_callback_t = void (*)(void*);
```

sql/sql_admin.cc
```diff
@@ -1728,8 +1751,24 @@ class Alter_instance_reload_tls : public Alter_instance {
   LEX_CSTRING channel_name_;
   bool force_;
   Ssl_acceptor_context_type context_type_;
+  static std::vector<ssl_reload_callback_t> callbacks_;
```
percona-ysorokin, line 1763: Do we really need to keep more than one active `ssl_reload_callback_t` here?

percona-ysorokin (changes requested):
Please also follow the naming conventions described in
https://confluence.percona.com/pages/viewpage.action?pageId=68393728
in both commit's and PR's title/message.

`keyword` in the title and Jira ticket URL
```
PS-7125 feature: Reloading X ssl certificates on ALTER INSTANCE RELOAD TLS

https://jira.percona.com/browse/PS-7125

...
```

plugin/x/ngs/include/ngs/client.h
dutow: No, it is a shared_ptr. By being there, it ensures that the ssl context exists until the client using it is connected.

plugin/x/src/interface/server.h
dutow: Same answer: the shared_ptr ensures that 2 connections can use 2 different ssl contexts (old and new) at the same time.

sql/sql_admin.cc
dutow: This change extends the plugin interface. Currently, only the X plugin uses it, but in theory, others could also implement this. 

For example, we already got a ticket for PXC to support alter instance reload tls for the wsrep port, which could use this extension point.

include/mysql/plugin.h
dutow: This sounds like similar to why I added an int return type to the registration function, even if currently it always returns 0 - I'm not sure if I remove one futureproofing, adding a different one makes sense.

plugin/x/src/server/server.cc
dutow: No, a leftover from some debugging, I'll remove it.

plugin/x/ngs/include/ngs/client.h
percona-ysorokin: Oh, I got you. By using `shared_ptr` here you extend the lifetime of the old `Ssl_context` which may still be used by existing connections when `ALTER INSTANCE RELOAD TLS` is invoked.

plugin/x/ngs/src/client.cc
```diff
@@ -124,7 +124,9 @@ void Client::activate_tls() {
   const auto real_connect_timeout =
       std::min<uint32_t>(connect_timeout, m_read_timeout);
 
-  if (m_server.ssl_context()->activate_tls(&connection(),
+  m_ssl = m_server.ssl_context();
+
+  if (m_ssl->activate_tls(&connection(),
```
percona-ysorokin: Weird formatting here

sql/sql_admin.cc
percona-ysorokin: Agree. Good point.

sql/sql_admin.cc
percona-ysorokin: Now here and in `mysql/plugin.h` callbacks have different signatures.
May be just `#include "mysql/plugin.h"` here instead of duplicate `using` declaration?

sql/sql_admin.cc
```diff
@@ -1699,11 +1701,32 @@ class Alter_instance_reload_tls : public Alter_instance {
...
+      }
+
+      my_ok(m_thd);
+    }
     return res;
   }
   ~Alter_instance_reload_tls() {}
 
+  static void register_callback(ssl_reload_callback_t c) {
+    callbacks_.push_back(c);
```
percona-ysorokin, line 1728: May be add the same `auto it = std::find(callbacks_.begin(), callbacks_.end(), c);` logic here to ensure that the same callback is not added twice. You can use the return value here (`bool`) to indicate the success of the operation.

sql/sql_admin.cc
```diff
@@ -1728,8 +1751,24 @@ class Alter_instance_reload_tls : public Alter_instance {
...
+  static std::vector<ssl_reload_callback_t> callbacks_;
 };
 
+std::vector<ssl_reload_callback_t> Alter_instance_reload_tls::callbacks_;
+
+extern "C" {
+
+bool register_ssl_reload_callback(ssl_reload_callback_t c) {
+  Alter_instance_reload_tls::register_callback(c);
+  return 0;
```
percona-ysorokin:
```suggestion
  return Alter_instance_reload_tls::register_callback(c);
```

sql/sql_admin.cc
```diff
@@ -1728,8 +1751,24 @@ class Alter_instance_reload_tls : public Alter_instance {
...
+}
+
+bool deregister_ssl_reload_callback(ssl_reload_callback_t c) {
+  Alter_instance_reload_tls::deregister_callback(c);
+  return 0;
```
percona-ysorokin:
```suggestion
  return Alter_instance_reload_tls::deregister_callback(c);
```

sql/sql_admin.cc
```diff
@@ -1699,11 +1701,32 @@ class Alter_instance_reload_tls : public Alter_instance {
...
+        };
+      }
+
+      my_ok(m_thd);
+    }
     return res;
   }
   ~Alter_instance_reload_tls() {}
 
+  static void register_callback(ssl_reload_callback_t c) {
```
percona-ysorokin:
```suggestion
  static bool register_callback(ssl_reload_callback_t c) {
```

sql/sql_admin.cc
```diff
@@ -1699,11 +1701,32 @@ class Alter_instance_reload_tls : public Alter_instance {
...
+    callbacks_.push_back(c);
+  }
+
+  static void deregister_callback(ssl_reload_callback_t c) {
```
percona-ysorokin:
```suggestion
  static bool deregister_callback(ssl_reload_callback_t c) {
```

plugin/x/src/module_mysqlx.cc
```diff
@@ -122,6 +122,11 @@ void Module_mysqlx::unregister_udfs() {
   m_udf_register = nullptr;
 }
 
+static void x_ssl_reload_cb(void*) {
```
percona-ysorokin:
```suggestion
static bool x_ssl_reload_cb(void*) {
```

sql/sql_admin.cc
```diff
@@ -1699,11 +1701,32 @@ class Alter_instance_reload_tls : public Alter_instance {
...
     }
 
-    if (!res) my_ok(m_thd);
+    if (!res) {
+      String specified_channel(channel_name_.str, channel_name_.length,
+                               system_charset_info);
+      if (!my_strcasecmp(system_charset_info, mysql_main_channel.c_str(),
+                         specified_channel.ptr())) {
+        for (auto const &cb : callbacks_) {
+          cb();
```
percona-ysorokin: May be issue an error/warning here if one of the callbacks returns `false`.

percona-ysorokin (changes requested):

percona-ysorokin (approved):
LGTM

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/877/#showFailuresLink

PR #4000, opened 2020-08-24
PS-5454: Fix unstable test percona_changed_page_bmp_log_resize

oleksandr-kachan:
After recent changes to redo log design checkpoint behavior
changed. Now checkpoint LSN may point into a middle of a log
record. The online log tracker wasn't aware of such checkpoint
behavior.

To fix this online log tracker was improved in order to correctly
process log records around checkpoint. As checkpoint may point into
a middle of a log record it needs to properly find a point in log
around checkpoint LSN where to stop log parsing. This will be a
beginning of the next record group which starts at or after
checkpoint LSN.

The same is correct to a point where to start parsing data during
an iteration. This again will be a beginning of a record group
but this time it's searched around previous checkpoint LSN.

### DIFF
--- a/storage/innobase/log/log0online.cc
+++ b/storage/innobase/log/log0online.cc
@@ -81,7 +81,7 @@ class log_buffer {
   lsn_t start_lsn{0};
   lsn_t current_lsn{0};
   const_iterator current_ptr{buffer.cbegin()};
-  lsn_t limit_lsn{0};
+  lsn_t checkpoint_lsn{0};
 
   log_buffer() noexcept {}
 
@@ -89,7 +89,6 @@ class log_buffer {
   MY_NODISCARD bool invariants() const noexcept {
     ut_ad(mutex_own(&log_bmp_sys_mutex));
     ut_ad(start_lsn <= current_lsn);
-    ut_ad(current_lsn <= limit_lsn);
     ut_ad(static_cast<decltype(start_lsn)>(current_ptr - buffer.cbegin()) <=
           current_size);
     return true;
@@ -100,12 +99,6 @@ class log_buffer {
   MY_NODISCARD const_iterator ccurrent() const noexcept {
     ut_ad(mutex_own(&log_bmp_sys_mutex));
     return current_ptr;
-  }
-
-  void set_limit(lsn_t limit) {
-    ut_ad(mutex_own(&log_bmp_sys_mutex));
-    ut_ad(limit >= limit_lsn);
-    limit_lsn = limit;
   }
 
   MY_NODISCARD lsn_t get_current_lsn() const noexcept {
@@ -133,7 +126,7 @@ class log_read_buffer final : public log_buffer<FOLLOW_SCAN_SIZE> {
 
   MY_NODISCARD bool is_data_available() const noexcept {
     ut_ad(mutex_own(&log_bmp_sys_mutex));
-    const auto result = ccurrent() < cend() && get_current_lsn() < limit_lsn;
+    const auto result = ccurrent() < cend();
     if (result) {
       ut_ad(invariants());
     }
@@ -161,6 +154,12 @@ class log_read_buffer final : public log_buffer<FOLLOW_SCAN_SIZE> {
     current_ptr += OS_FILE_LOG_BLOCK_SIZE;
   }
 
+  void set_checkpoint_lsn(lsn_t checkpoint_lsn_) {
+    ut_ad(mutex_own(&log_bmp_sys_mutex));
+    ut_ad(checkpoint_lsn_ >= checkpoint_lsn);
+    checkpoint_lsn = checkpoint_lsn_;
+  }
+
 #ifdef UNIV_DEBUG
  private:
   MY_NODISCARD bool invariants() const noexcept {
@@ -178,7 +177,6 @@ class log_read_buffer final : public log_buffer<FOLLOW_SCAN_SIZE> {
 
 void log_read_buffer::read(lsn_t read_start, lsn_t read_end) noexcept {
   ut_ad(mutex_own(&log_bmp_sys_mutex));
-  ut_ad(read_start < limit_lsn);
   ut_ad(get_current_lsn() >= read_start || get_current_lsn() == 0);
   current_size = read_end - read_start;
   ut_ad(current_size <= capacity);
@@ -224,6 +222,9 @@ class log_parse_buffer final : public log_buffer<RECV_PARSING_BUF_SIZE> {
   parse_result parse_status{parse_result::OK};
   iterator end_ptr{buffer.begin()};
   lsn_t end_lsn{0};
+  lsn_t parse_start_lsn{0};
+  lsn_t parse_stop_lsn{0};
+  bool is_checkpoint_block_reached{false};
 
 #ifdef UNIV_DEBUG
   MY_NODISCARD size_type size() const noexcept {
@@ -235,7 +236,10 @@ class log_parse_buffer final : public log_buffer<RECV_PARSING_BUF_SIZE> {
     ut_ad(current_ptr <= end_ptr);
     ut_ad(buffer.begin() + current_size == end_ptr);
     ut_ad(current_lsn <= end_lsn);
-    if (current_lsn == limit_lsn) ut_ad(parse_status == parse_result::OK);
+    ut_ad(parse_stop_lsn == 0 || current_lsn <= parse_stop_lsn);
+    if (current_lsn == parse_stop_lsn) {
+      ut_ad(parse_status == parse_result::PAST_END);
+    }
     return true;
   }
 #endif
@@ -259,11 +263,12 @@ class log_parse_buffer final : public log_buffer<RECV_PARSING_BUF_SIZE> {
 
   MY_NODISCARD bool can_parse_current_data() const noexcept {
     return parse_status == parse_result::OK && ccurrent() != cend() &&
-           current_lsn < limit_lsn;
+           current_lsn < end_lsn &&
+           (parse_stop_lsn == 0 || current_lsn < parse_stop_lsn);
   }
 
   MY_NODISCARD bool parsed_past_checkpoint() const noexcept {
-    return parse_status == parse_result::PAST_END;
+    return parse_stop_lsn != 0 && current_lsn >= parse_stop_lsn;
   }
 
   void reset_parse_status() { parse_status = parse_result::OK; }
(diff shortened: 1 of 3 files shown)


storage/innobase/log/log0online.cc
```diff
@@ -1370,13 +1372,16 @@ is tracked.
 bool log_online_follow_redo_log(void) {
   ut_ad(!srv_read_only_mode);
 
-  const auto last_checkpoint_lsn = log_get_checkpoint_lsn(*log_sys);
+  const auto last_checkpoint_aligned_lsn = ut_uint64_align_up(
+      log_get_checkpoint_lsn(*log_sys), OS_FILE_LOG_BLOCK_SIZE);
   bool result = true;
```
satya-bodapati: So this is the core change. We are covering bytes after checkpoint_lsn.

But we are using the entire block which may or may not have redo records.

Is it possible to determine the LSN at which the current redo record ends? (I think recovery also tries to find the end of redo record from checkpoint LSN).

satya-bodapati (changes requested):

storage/innobase/log/log0online.cc
satya-bodapati: Btw, Sorry for the late response..  :(

oleksandr-kachan: To my understanding we can not predict where log data will end without parsing that data. What I can see in recovery we still read some number of full blocks from log from `start_lsn` to `start_lsn + 4 * UNIV_PAGE_SIZE` and then stop parsing that data once we find partially filled in block.

satya-bodapati: yes and no. There is no prediction as such. Like you said, we have to parse redo records and find the start of first record after the checkpoint_lsn.

Look for "parse_start_lsn" in recovery code. So essentially we need parse_start_lsn which will include bytes after the checkpoint_lsn and before the start of next record.

oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/863/

storage/innobase/log/log0online.cc
```diff
@@ -273,6 +278,83 @@ class log_parse_buffer final : public log_buffer<RECV_PARSING_BUF_SIZE> {
...
+    parse_stop_lsn = 0;
+    is_checkpoint_block_reached = false;
+  }
+
+  void check_update_parse_stop_lsn(log_read_buffer &from,
+                                   log_buffer::size_type data_len) noexcept {
+    auto read_buf_current_lsn = from.get_current_lsn();
+
+    if (!is_checkpoint_block_reached &&
+        checkpoint_lsn <= read_buf_current_lsn + OS_FILE_LOG_BLOCK_SIZE) {
```
satya-bodapati, line 330: should we also add checkpoint_lsn > read_buf_current_lsn to ensure the checkpoint_lsn is truly in this block?

storage/innobase/log/log0online.cc
```diff
@@ -273,6 +278,83 @@ class log_parse_buffer final : public log_buffer<RECV_PARSING_BUF_SIZE> {
...
+      // are looking for a beginning of the next log record group which starts
+      // after checkpoint_lsn. Parsing should stop at that point.
+      is_checkpoint_block_reached = true;
+    }
+
+    if (is_checkpoint_block_reached && parse_stop_lsn == 0) {
+      if (data_len < OS_FILE_LOG_BLOCK_SIZE) {
+        // Block is filled in partially. Parsing should stop at the end of the
+        // data block.
+        parse_stop_lsn = read_buf_current_lsn + data_len;
```
satya-bodapati, line 341: when the block is filled partially, the checkpoint_lsn can be still in between the block right? Not necessary to the end?

storage/innobase/log/log0online.cc
```diff
@@ -1061,17 +1149,50 @@ static void log_online_parse_redo_log(void) {
...
+          log_bmp_sys->parse_buf.set_parse_start_lsn(last_parsed_lsn);
+        }
+        continue;
+      }
+
+      if (log_bmp_sys->parse_buf.checkpoint_reached() &&
+          log_bmp_sys->parse_buf.get_parse_stop_lsn() == 0 &&
+          last_parsed_lsn >= log_bmp_sys->checkpoint_lsn &&
+          (is_single_rec || type == MLOG_DUMMY_RECORD ||
+           type == MLOG_MULTI_REC_END)) {
```
satya-bodapati, line 1187: is the case to handle that checkpoint_lsn and the parse_stop_lsn (the lsn after checkpoint_lsn first_rec_group) are across blocks?

storage/innobase/log/log0online.cc
```diff
@@ -273,6 +278,83 @@ class log_parse_buffer final : public log_buffer<RECV_PARSING_BUF_SIZE> {
...
+
+    if (is_checkpoint_block_reached && parse_stop_lsn == 0) {
+      if (data_len < OS_FILE_LOG_BLOCK_SIZE) {
+        // Block is filled in partially. Parsing should stop at the end of the
+        // data block.
+        parse_stop_lsn = read_buf_current_lsn + data_len;
+      } else {
+        auto rec_group_offset = log_block_get_first_rec_group(from.ccurrent());
+        if (rec_group_offset > 0 &&
+            read_buf_current_lsn + rec_group_offset >= checkpoint_lsn) {
```
satya-bodapati, line 345: what about the case u mentioned in the comment?  if rec_group_offset is 0, what will be the parse_stop_lsn? 

I also read a comment that sometimes the parsing_stop_len is < checkpoint_lsn, in that case parse_stop_lsn is set as checkpoint_lsn. Do u also do this?

storage/innobase/log/log0online.cc
```diff
@@ -1132,23 +1268,70 @@ static bool log_online_follow_log_seg(
...
           log_bmp_sys->has_parse_data_to());
 
-    /* How many bytes of log data should we skip in the current log block. */
-    const auto skip_already_in_parse_buf_len =
-        static_cast<ulint>(log_bmp_sys->has_parse_data_to() -
-                           log_bmp_sys->read_buf.get_current_lsn());
-    log_online_parse_redo_log_block(log_bmp_sys->read_buf,
-                                    skip_already_in_parse_buf_len);
+    skip_len = static_cast<ulint>(log_bmp_sys->has_parse_data_to() -
+                                  log_bmp_sys->read_buf.get_current_lsn());
```
satya-bodapati, line 1331: where is the code to skip the empty blocks on redo log resize? is this the one?

storage/innobase/log/log0online.cc
```diff
@@ -224,6 +222,9 @@ class log_parse_buffer final : public log_buffer<RECV_PARSING_BUF_SIZE> {
   parse_result parse_status{parse_result::OK};
   iterator end_ptr{buffer.begin()};
   lsn_t end_lsn{0};
+  lsn_t parse_start_lsn{0};
+  lsn_t parse_stop_lsn{0};
```
satya-bodapati, line 226: can you explain briefly how we use these two LSNs to determine the exact stop_LSN (at or around checkpoint_LSN?)

satya-bodapati (changes requested):
Still trying to understand your code changes (understood at the design level though)

storage/innobase/log/log0online.cc
oleksandr-kachan: With the recent changes to redo log in 8.0 checkpoint LSN may point into a middle of log record. So now for both parse stop and start points we need to find first record group which starts at or after checkpoint LSN.
The parse_start_lsn and parse_stop_lsn are set to LSNs corresponding to those record groups starting points.

Once we read a block containing checkpoint LSN for parse start point, we check first record group offset for that block (log_block_get_first_rec_group()). In case this offset points after starting checkpoint within this block - we found parse_start_lsn. When parsing data everything with LSN less than parse_start_lsn will be skipped.

A bit different scenario is when first record group offset for the block points before starting checkpoint - we can not determine actual parse_start_lsn only from reading offset from block header, this can be done only while parsing data. In parse buffer we see that parse_start_lsn isn't set so we parse records one after another and based on parsed record length and their type we determine first record group located after starting checkpoint LSN. That is our parse_start_lsn in this case. Data from those parsed blocks located before parse_start_lsn is skipped.

The similar approach is used for parse_stop_lsn. We start searching for it once we reach a block containing ending  checkpoint LSN. We continue parsing records after checkpoint LSN until parse_stop_lsn is reached.

In addition, parse_start_lsn and parse_stop_lsn are not necessarily located in the same log block as corresponding checkpoint LSN. It may be in the next block.

storage/innobase/log/log0online.cc
oleksandr-kachan: That will be True for all blocks located before one with checkpoint_lsn. For me it looked like this condition isn't really useful here. That is why I ignored it.

storage/innobase/log/log0online.cc
oleksandr-kachan: Not 100% sure about this. From what I've seen it usually points to the actual record group after checkpoint and parse_start_lsn calculated during the next iteration pointed here as well.
I actually took this from recovery code and I could misunderstand its meaning.

storage/innobase/log/log0online.cc
oleksandr-kachan: If rec_group_offset is 0, next record group starts in one of the next blocks. It will parse current block. Read next one and most probably parse_stop_lsn will be found there while actually parsing data.

> the parsing_stop_len is < checkpoint_lsn

Is this one from recovery code?

storage/innobase/log/log0online.cc
oleksandr-kachan: This is for cases when first record group offset points before checkpoint_lsn. It may stop after checkpoint_lsn in the same block as well as continue searching in the next block.

storage/innobase/log/log0online.cc
oleksandr-kachan: Empty blocks are skipped in log_online_parse_redo_log_block(). And here it skips already parsed data.

storage/innobase/log/log0online.cc
satya-bodapati: Yes, in recovery code recv_scan_log_recs().

      if (recv_sys->parse_start_lsn < recv_sys->checkpoint_lsn) {
        /* We start to parse log records even before
        checkpoint_lsn, from the beginning of the log
        block which contains the checkpoint_lsn.

        That's because the first group of log records
        in the log block, starts before checkpoint_lsn,
        and checkpoint_lsn could potentially point to
        the middle of some log record. We need to find
        the first group of log records that starts at
        or after checkpoint_lsn. This could be only
        achieved by traversing all groups of log records
        that start within the log block since the first
        one (to discover their beginnings we need to
        parse them). However, we don't want to report
        missing tablespaces for space_id in log records
        before checkpoint_lsn. Hence we need to ignore
        those records and that's why we need a counter
        of bytes to ignore. */

oleksandr-kachan: As I understand they start parsing at `recv_sys->parse_start_lsn = scanned_lsn + log_block_get_first_rec_group(log_block)` but still skip data until first record group starting after checkpoint_lsn is found.
In my change I do similar thing but in a bit different way. In `log_online_follow_log_seg()` I calculate LSN for first record group in block

`const auto parse_start_lsn = log_bmp_sys->read_buf.get_current_lsn() + first_rec_group_offset;`

In case it points before checkpoint, log parsing still begins from this point (data from block start till this offset is skipped) but parse_start_lsn isn't stored in parse buffer. It will be calculated while parsing the log.

storage/innobase/log/log0online.cc
satya-bodapati: in such case, can't we simply use the checkpoint_lsn itself?

storage/innobase/log/log0online.cc
satya-bodapati: Thanks for a detailed explanation, I think it deserves to be in code (and maybe parts of it in the commit message) as well.

storage/innobase/log/log0online.cc
satya-bodapati: Ok, lets ignore this assert

storage/innobase/log/log0online.cc
satya-bodapati: Can you please point me to the recovery source?

satya-bodapati (changes requested):

storage/innobase/log/log0online.cc
oleksandr-kachan: Not sure I get your idea correctly. Can you please describe it in more detail.

storage/innobase/log/log0online.cc
oleksandr-kachan: There is similar condition in recv_scan_log_recs()

`
if (data_len < OS_FILE_LOG_BLOCK_SIZE) {
      /* Log data for this group ends here */
      finished = true;
      break;
    } else {
      log_block += OS_FILE_LOG_BLOCK_SIZE;
    }
`

storage/innobase/log/log0online.cc
satya-bodapati: why not use log_block_get_first_rec_group(block) ? what is difference between log_block_get_first_rec_group(block) vs first_rec_group_offset?

storage/innobase/log/log0online.cc
```diff
@@ -1132,23 +1268,70 @@ static bool log_online_follow_log_seg(
...
+    // record group which starts at or after checkpoint LSN and ignore
+    // everything before this point.
+    auto first_rec_group_offset{0};
+
+    // Looking for the first complete record to start parsing at.
+    while (log_bmp_sys->read_buf.is_data_available() &&
+           (first_rec_group_offset = log_block_get_first_rec_group(
+                log_bmp_sys->read_buf.ccurrent())) == 0) {
+      log_bmp_sys->read_buf.advance();
+    }
```
satya-bodapati, line 1299: So if I understand correctly, upstream will not use a loop but you use the loop to find the first_rec_group(). Is this is the difference?

storage/innobase/log/log0online.cc
```diff
@@ -1132,23 +1268,70 @@ static bool log_online_follow_log_seg(
...
+      // Go to the next log segment.
+      return true;
+    }
+
+    ut_ad(first_rec_group_offset > 0);
+
+    const auto parse_start_lsn =
+        log_bmp_sys->read_buf.get_current_lsn() + first_rec_group_offset;
+
+    if (parse_start_lsn >= log_bmp_sys->start_lsn) {
```
satya-bodapati, line 1311: how do we get log_bmp_sys->start_lsn ?

storage/innobase/log/log0online.cc
```diff
@@ -1335,24 +1518,29 @@ bool log_online_follow_redo_log_one_pass(void) {
...
+  // Grab the LSN of the last checkpoint. It is possible that checkpoint LSN
+  // points to the middle of a log record. We'd need to find next record group
+  // beginning (which is possibly located in one of the following log blocks)
+  // and parse data up to that point.
+  log_bmp_sys->set_checkpoint_lsn(log_get_checkpoint_lsn(*log_sys));
+
+  ut_ad(log_bmp_sys->parse_buf.get_parse_start_lsn() == 0);
 
-  if (log_bmp_sys->end_lsn == log_bmp_sys->start_lsn) {
+  if (log_bmp_sys->checkpoint_lsn <= log_bmp_sys->has_parse_data_to()) {
```
satya-bodapati, line 1536: what is the difference  bmp_sys->start_lsn and has_parse_data_to()?

storage/innobase/log/log0online.cc
```diff
@@ -1897,14 +2084,14 @@ bool log_online_purge_changed_page_bitmaps(
...
-    if (lsn > log_bmp_sys->end_lsn) {
+    if (log_bmp_sys->checkpoint_lsn != 0 && lsn > log_bmp_sys->checkpoint_lsn) {
       lsn_t new_file_lsn;
       if (lsn == LSN_MAX) {
         /* RESET restarts the sequence */
         log_bmp_sys->out_seq_num = 0;
         new_file_lsn = 0;
       } else {
-        new_file_lsn = log_bmp_sys->end_lsn;
+        new_file_lsn = log_bmp_sys->checkpoint_lsn;
```
satya-bodapati: shouldn't the end_lsn be the parse_stop_lsn?

satya-bodapati (changes requested):
slowly getting there..

storage/innobase/log/log0online.cc
oleksandr-kachan: In this method (log_block_get_first_rec_group) the log_block_get_first_rec_group() is actually used. Maybe you refer to some different place in code?

storage/innobase/log/log0online.cc
oleksandr-kachan: There is a loop in recovery code as well to iterate all blocks, code is just designed in a different way.

storage/innobase/log/log0online.cc
oleksandr-kachan: It is set by invoking log_bmp_sys->start_at() before actually starting parsing log. For example in log_online_read_init() or log_online_track_missing_on_startup().

storage/innobase/log/log0online.cc
oleksandr-kachan: The has_parse_data_to() checks last known LSN in parse buffer at first place. It may return bmp_sys->start_lsn as a fallback in case there were not any data parsed yet.

storage/innobase/log/log0online.cc
oleksandr-kachan: Good point, will recheck this.

oleksandr-kachan: We can use log_bmp_sys->start_lsn to get last parsed LSN from previous parser iteration. We store parse_buf.parse_stop_lsn there to keep state between iterations. I've updated this part a bit.

PR #4001, opened 2020-08-24
PS-7026 Constrains for keyring encryptions.

robgolebiowski:
Added DD flag ONLINE_KEYRING - to mark tablespaces - that are
ONLINE_KEYRING encrypted
This is needed for checks that only keyring (or online keyring)
encrypted tables are moved into online keyring encrypted tablespaces.

Also online or keyring encrypted tables follow the upstream's rules of
moving tables into general tablespaces.
Online/Keyring encrypted tables cannot be moved into Master Key
encrypted or unencrypted general tablespaces and the other way around.

Moving between Master Key encrypted general tablespace and
file_per_table online keyring encrypted tablespace
is allowed (given ENCRYPTION clause is provided).

Moving between online keyring encrypted general tablespace and
file_per_table Master Key encrypted tablespace
is allowed (given ENCRYPTION clause is provided).

Without sufficient privileges online/keyring encrypted tables cannot be
moved to database with default encryption set to N, however they can be
moved into database with default encryption set to Y.
Without privileges check (table_encryption_privilege_check set to OFF)
moving tables between all databases is allowed.

Since online keyring encryption provides a way of encrypting system
tablespace. This PR tests different combination of moving tables into
online encrypted system tablespace. Also checks if online/keyring
encrypted
table cannot be moved to bootstrap Master Key encrypted system.


robgolebiowski:
https://ps80.cd.percona.com/job/percona-server-8.0-param/744/

sql/dd/dd_table.cc
```diff
@@ -2814,17 +2816,19 @@ Encrypt_result is_tablespace_encrypted(THD *thd, const Table &t,
...
+  String_type encryption_option;
   bool encrypted =
       std::any_of(tspids.begin(), unique_end, [&](const Object_id id) {
         bool is_encrypted = false;
         if (is_general_tablespace_and_encrypted(id, thd, &is_encrypted,
-                                                is_general_tablespace)) {
+                                                is_general_tablespace,
+                                                encryption_option)) {
           error = true;
           return false;
```
satya-bodapati, line 2827: How did the casting work here? Shouldn't it be like above code you wrote?

return {false, {false, ""}}; ?

Do we need the bool? can we infer from the string_type returned?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -12867,11 +12867,6 @@ bool create_table_info_t::create_option_encryption_is_valid() const {
   bool table_is_keyring =
       Encryption::is_keyring(m_create_info->encrypt_type.str);
 
-  if (table_is_keyring && !m_allow_file_per_table) {
-    my_printf_error(ER_KEYRING_ILLEGAL_ENCRYPTION_OPTION,
-                    "InnoDB: KEYRING requires innodb_file_per_table.", MYF(0));
-    return (false);
-  }
```
satya-bodapati, line 12874: After this PR, do we allow creation of KEYRING general tablespaces? ie CREATE TABLESPACE ts1 ADD DATAFILE ts1.ibd ENCRYPTION=KEYRING?

Also, does this change mean, we now allow keyring tables in general tablespaces? Previously it wasn't allowed? :thinking:

storage/innobase/dict/dict0dd.cc
```diff
@@ -6282,7 +6283,7 @@ bool dd_set_encryption_flag(THD *thd, const char *space_name,
     dd_space_flags |= (1U << FSP_FLAGS_POS_ENCRYPTION);
     dd_space->se_private_data().set(dd_space_key_strings[DD_SPACE_FLAGS],
                                     dd_space_flags);
-    dd_space->options().set("encryption", "Y");
+    dd_space->options().set("encryption", "ONLINE_KEYRING");
```
satya-bodapati, line 6286: I haven't seen an unset/remove equivalent anywhere?

sql/sql_tablespace.cc
```diff
@@ -483,11 +483,14 @@ bool Sql_cmd_create_tablespace::execute(THD *thd) {
...
             DEFAULT_TABLE_ENC_ONLINE_TO_KEYRING;
-    // We set encrypt_type, which is later assigned to tablespace's DD
-    // encryption option to Y for online KEYRING encryption. This field is
-    // designed so the user could check if given table is encrypted or not. The
-    // details on how it is encrypted (KEYRING in this case) are left in SE.
-    encrypt_type = encrypt_tablespace ? "Y" : "N";
+    if (encrypt_tablespace) {
+      encrypt_type = (global_system_variables.default_table_encryption ==
+                      DEFAULT_TABLE_ENC_ONLINE_TO_KEYRING)
+                         ? "ONLINE_KEYRING"
```
satya-bodapati, line 489: I really hope we now allow foreground KEYRING encrypted general tablespaces. Because this code, does it automatically based on global option.

satya-bodapati (changes requested):
Looks good mostly. A lot of testcase scenarios! Good work.
Have to check though.

sql/sql_tablespace.cc
robgolebiowski: This assigns encrypt_type to ONLINE_TO_KEYRING, not KEYRING and encrypt_type means something else for tablespaces and file_per_table tables.

satya-bodapati: Ok, Generally speaking, is it only a "scope" limitation to support KEYRING for general tablespaces? CREATE TABLESPACE ts1 ADD DATAFILE ts1.ibd ENCRYPTION=KEYRING?

robgolebiowski: Yeah, we didn't want to add additional work. The KEYRING encryption for tablespaces should be done, imho in the same way upstream is doing ENCRYPTION='Y' for tablespaces. It should not be done by encryption threads.

PR #4004, opened 2020-08-25
PS-7143 : Server freezes and reaches max connections due to ACL Cache…

satya-bodapati:
… Lock waits

Problem:
--------
ACL (Access Control Lists) caches since 8.0 are handled by MDL lock.
Previously(5.7) it was handled by a mix of mutex(acl_cache->lock)
+ Paritioned_RW_lock (LOCK_grant)

From 8.0, the cache (of user, hosts, other ACL structures) is handled by MDL
lock. Readers(ex acl_check_host) take MDL_SHARED and Writers (DDLs: GRANT, FLUSH
PRIVILEGES) take MDL_EXCLUSIVE. So ideally we should see improvement when
compared to mutex.

But MDL Lock suffers a performance issue when there are huge number of readers
and atleast (or with just one) writer. MDL lock suffers from contention on its
internal rwlock (lock->m_rwlock). Lets see how

~~~~~
Keywords used:

ACL1 -> any grant/revoke/FLUSH PRIVILEGE (takes MDL EXCLUSIVE)
ACL2 -> is from one of the connection login with default_db
(acl_authenticate()->mysql_change_db()->acl_get()->insert_entry_in_db_cache()->
requires MDL EXCLUSIVE lock onACL cache)
ACL3 -> connection login requests (acl_check_host() etc) that check
cache. These are readers that only need MDL_SHARED
 batch1 size : 200
 batch2 size : 5000
 batch3 size : 200
~~~~~

1.  ACL1 command takes MDL EXCLUSIVE

2.  allow hord of readers (ACL3s) batch1 to come and wait for MDL_SARED lock
(they cannot take it yet as MDL_EXCLUSIVE is already acquired. So these requests
go into something called slow_path and are registered in m_waiting_list and
later in m_granted list after MDL_EXCLUSIVE is released)

3.  let ACL1 command release MDL_EXCLUSIVE lock

4.  All ACL3 Batch1 readers acquire MDL_SHARED lock. (they are in m_granted list
of ACL Cache lock)

5.  Before all the ACL3 Batch1 readers finish, Let there be a new ACL2 command
that requests for MDL_EXCLUSIVE lock

6.  ACL2 goes into m_waitng list, cannot take MDL_EXCLUSIVE lock yet as ACL3
Batch1 readers hold MDL_SHARED lock

7.  Allow new set of readers to come. ALC3 batch2 (size of around 5000). Now,
even though there is no MDL_EXCLUSIVE lock taken, since there is waiting
MDL_EXCLUISVE lock, all new readers (ACL3 batch2) go into m_waiting list.

    At this stage, ACL Cache MDL lock looks like this
    m_granted list: size 200: All MDL_SHARED
    m_waiting list: one MDL_EXCLUSIVE, 5000 MDL_SHARED requests

8.  Allow another set of readers. ALC3 batch3 (size of 200) to come. These
readers, do deadlock checks before going to waiting list. As the
granted list and waiting list are of huge size, the deadlock check now takes lot
of time. These deadlock checks happen after taking the internal RWlock (of ACL
MDL lock) in SHARED MODE.

9.  the ACL1 batch1 readers who are granted, cannot release the MDL LOCK. This
is because for releasing the lock/ticket they need EXCLUSIVE mode of internal
RWlock to remove ticket from m_granted list. (The internal RWlock is taken by
readers who are doing deadlock checks for long long time)

10. Any more new readers (ACL3), they also get stuck at internal RWlock (at much
early stage).

11. Due to contention on internal RW lock of MDL lock, every connection
(requires ACL CACHE Lock SHARED See acl_check_host()) also get stuck

12. Eventually, we will soon run of max_connections.

13. Freeze lasts until ACL3 batch3 readers complete deadlock checks.

Fix:
----
Ideally, we only need a simple RW-lock. The MDL lock seems to suffer when used a
pure RW lock.

The only benefit of ACL Cache lock as MDL is that it can detect lock ordering
issues. (they can arise only due to buggy code) For example opening ACL TABLES
take MDL lock on grant table (TABLE lock. Lets say A). And ACL Cache lock MDL
(Lets say B). Threads should acquire in same order (first A and then B)

If we don't, then

thread1: can acquire lock A wait for B

thread2: acquire B and wait for lock A

With proper lock ordering,  we always acquire in same order for all threads.
Lets say first A and then B..

MDL deadlock detector found this issue (bug in code) and it is solved in 8.0.18.
See cda465b69d5381a739c4669be55e5c500536c7dd

Another minor benefit is ACL cache lock is acquired at higher layers and lower
layers can check if it has already acquired lock or not (using thd->mdl_context)

[Final Fix chosen]
------------------
We simply suppress the expensive deadlock check for ACL Cache MDL lock. This
removes the contention on internal RWlock of MDL lock.

If ever there is really deadlock caused by buggy code, we rely on
ACL_CACHE_LOCK_TIMEOUT value. The value of ACL_CACHE_LOCK_TIMEOUT is reduced to
15mins (from upstream's default of 60 mins). 60 mins is too much. Even InnoDB
Semaphore latch wait timeout is around 15mins

### DIFF
--- a/client/mysqltest.cc
+++ b/client/mysqltest.cc
@@ -112,7 +112,7 @@
 #define MAX_VAR_NAME_LENGTH 256
 #define MAX_COLUMNS 256
 #define MAX_DELIMITER_LENGTH 16
-#define DEFAULT_MAX_CONN 128
+#define DEFAULT_MAX_CONN 1024
 #define REPLACE_ROUND_MAX 16
 
 /* Flags controlling send and reap */
@@ -7400,8 +7400,8 @@ static struct my_option my_long_options[] = {
      REQUIRED_ARG, 500, 1, 10000, nullptr, 0, nullptr},
     {"max-connections", OPT_MAX_CONNECTIONS,
      "Max number of open connections to server", &opt_max_connections,
-     &opt_max_connections, nullptr, GET_INT, REQUIRED_ARG, 128, 8, 5120,
-     nullptr, 0, nullptr},
+     &opt_max_connections, nullptr, GET_INT, REQUIRED_ARG, DEFAULT_MAX_CONN, 8,
+     9120, nullptr, 0, nullptr},
     {"no-skip", OPT_NO_SKIP, "Force the test to run without skip.", &no_skip,
      &no_skip, nullptr, GET_BOOL, NO_ARG, 0, 0, 0, nullptr, 0, nullptr},
     {"no-skip-exclude-list", 'n',
--- a/mysql-test/t/acl_mdl_scalability.test
+++ b/mysql-test/t/acl_mdl_scalability.test
@@ -1,0 +1,268 @@
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+# We rely on g++ to compile c-program
+--source include/not_windows.inc
+# Because of 6K connections, mysqld consumes close to
+# 14G virtual memory
+--source include/big_test.inc
+--source include/not_valgrind.inc
+
+--echo #
+--echo # PS-7143 : Server freezes and reaches max connections due to ACL Cache Lock waits
+--echo #
+
+--echo # Check if system can handle 12k concurrent threads
+--echo # 12K because server creates 6K threads and clients
+--echo # another 6k threads/processes
+
+--echo # There is no easy way to check if the system can
+--echo # handle 12K threads. We compile C-program (need g++)
+--echo # and execute it and check return status
+
+--write_file $MYSQL_TMP_DIR/thread_limit_test.cc
+#include <pthread.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <atomic>
+#include <iostream>
+#include <thread>
+#include <chrono>
+
+#define NUM_THREADS     12000
+std::atomic<int> current_threads{0};
+using namespace std;
+
+void* ThreadTask(void *threadid)
+{
+   long tid;
+   tid = (long)threadid;
+   std::cout << "Hello World! It's me, thread# " << tid << std::endl;
+   ++current_threads;
+   while (current_threads.load() < NUM_THREADS) {
+          std::this_thread::sleep_for(std::chrono::seconds(1));
+   }
+   std::cout << "Thread exiting # " << tid << std::endl;
+   pthread_exit(NULL);
+}
+
+int main ()
+{
+   pthread_t threads[NUM_THREADS];
+   int rc;
+   for(long t=0; t<NUM_THREADS; t++){
+      std::cout << "In main: creating thread: " << t << std::endl;
+      rc = pthread_create(&threads[t], NULL, ThreadTask, (void *)t);
+      if (rc){
+         std::cout << "ERROR: return code from pthread_create() is " << strerror(rc) << std::endl;
+         exit(-1);
+      }
+   }
+
+   for(long t=0; t<NUM_THREADS; t++){
+	pthread_join(threads[t], NULL);
+   }
+
+   std::cout << "finish executing main " << std::endl;
+   return(0);
+}
+EOF
+--echo # Require g++ to compile this program. If the exec fails "sh: 1: g++: not found". Please install g++ and retry
+--exec g++ $MYSQL_TMP_DIR/thread_limit_test.cc -std=c++11 -lpthread -o $MYSQL_TMP_DIR/thread_limit 2>&1 > $MYSQL_TMP_DIR/thread_limit_gcc_error.txt
+
+--echo # if the below exec fails, it means the operating system has limit of threads less than 12K
+--echo # This testcase needs OS that can create 12k threads.
+--echo # Try these settings (only if they are less 12,000)
+--echo # ulimit -i  12000
+--echo # echo 12000 > /proc/sys/kernel/threads-max
+--echo # echo 60000 > /proc/sys/vm/max_map_count
+--echo # echo 12000 > /proc/sys/kernel/pid_max
+--echo # echo 12000 > /sys/fs/cgroup/pids/user.slice/user-$UID.slice/pids.max
+--exec $MYSQL_TMP_DIR/thread_limit 2>&1 > $MYSQL_TMP_DIR/thread_limit_gcc_error.txt
+--remove_file $MYSQL_TMP_DIR/thread_limit_test.cc
+--remove_file $MYSQL_TMP_DIR/thread_limit_gcc_error.txt
+--remove_file $MYSQL_TMP_DIR/thread_limit
+
+--echo # Precondition check of the system is completed. The system can handle 12K concurrent threads
+
+--source include/count_sessions.inc
+
+--let $debug_point = skip_session_admin_check
+--source include/add_debug_point.inc
+
+CREATE USER satya@'%' IDENTIFIED BY 'hello';
+GRANT ALL on test.* to satya@'%';
+GRANT ALL on mysql.* to satya@'%';
+
+--let $old_max_conn=`SELECT @@global.max_connections`
+SET GLOBAL max_connections = 6000;
+
+--connect (con_flush,localhost,root,,)
+--connect (con0,localhost,root,,)
+
+--let $i = 1
+while ($i <= 200) {
+ --connect (con$i, localhost, root)
+ --inc $i
+}
+
+--let $i = 201
+while ($i <= 700) {
+ --connect (con$i, 127.0.0.1, satya,hello,test)
+ --inc $i
+}
+
+--connection default
+--let $initial_con_count=`SELECT COUNT(*) FROM information_schema.processlist`
+
+--echo # Sleeping 15 seconds for allowing user to connect client for observability
+--echo # we cannot establish connection later as database freezes. Do now if you wish
+--echo # be in mysql-source/bld/mysql-test directory and execute the below command
+--echo # ../runtime_output_directory/mysql -uroot -S ./var/tmp/mysqld.1.sock
+#--sleep 15
+#--echo # sleep over
+
+--echo # connection con0: Hold ACL x lock and let readers wait
+--connection con0
+SET DEBUG_SYNC= 'acl_x_lock SIGNAL x_locked WAIT_FOR all_wait_s_locks_done';
+--send FLUSH PRIVILEGES
+
+--echo # 200 connections try for SHARED ACL lock and go into waiting list
+--echo # because FLUSH PRIVILEGES holds EXCLUSIVE lock
+--disable_query_log
+--let $i = 1
+while ($i <= 200) {
+ --connection con$i
+ SET DEBUG_SYNC = 'now WAIT_FOR x_locked NO_CLEAR_EVENT';
+ SET DEBUG_SYNC = 'acl_s_lock SIGNAL s_locked WAIT_FOR new_xlock NO_CLEAR_EVENT';
+ --send SHOW CREATE USER root@localhost
+ --inc $i
+}
+--enable_query_log
+
+--echo # Connection default. Waiting for 200 connections to reach blocking state
+--connection default
+--let $wait_timeout = 600
+let $wait_condition=
+  SELECT COUNT(*) = 200 FROM information_schema.processlist
+    WHERE state = 'Waiting for acl cache lock' AND
+          info = 'SHOW CREATE USER root@localhost';
+--source include/wait_condition.inc
+
+--echo # Now release X-lock by FLUSH PRIVILEGES. Now All 200 readers get the
+--echo # SHARED lock and they wait on debug sync point after acquiring SHARED
+--echo # ACL MDL lock
+SET DEBUG_SYNC = 'now SIGNAL all_wait_s_locks_done';
+
+--echo # Wait for 200 connection to reach acquired SHARED ACL MDL lock state
+--let $wait_timeout = 600
+let $wait_condition=
+ SELECT COUNT(*) = 200 FROM information_schema.processlist
+   WHERE state = 'debug sync point: acl_s_lock' AND
+         info = 'SHOW CREATE USER root@localhost';
+--source include/wait_condition.inc
+
+--echo # Start 5000 connection will go into ACL cache lock waiting state
+--let $i = 1
+while ($i <= 5000) {
+ --exec_in_background $MYSQL -usatya -phello -h 127.0.0.1 -P $MASTER_MYPORT test -e "SELECT 1" 2>&1 > /dev/null
+ if ($i == 1500) {
+   --connection con_flush
+   --send FLUSH PRIVILEGES;
+ }
+ --inc $i
+}
+
+--echo # Connection default: Wait for those 5000 new connections to reach ACL cache lock waiting state
+--connection default
+--let $wait_timeout = 600
+let $wait_condition=
+  SELECT COUNT(*) = 5000 FROM information_schema.processlist
+    WHERE USER = 'unauthenticated user' AND STATE = 'Waiting for acl cache lock';
+--source include/wait_condition.inc
+
+--echo # Start another 500 connections and stop them in just before deadlock check iteration
+--disable_query_log
+--let $i = 201
+while ($i <= 700) {
+  --connection con$i
+  SET DEBUG_SYNC = 'now WAIT_FOR s_locked NO_CLEAR_EVENT';
+  SET DEBUG_SYNC = 'mdl_acquire_lock_wait SIGNAL before_dead_lock WAIT_FOR start_dead_lock_check NO_CLEAR_EVENT';
+  SET DEBUG_SYNC = 'acl_mdl_dead_lock SIGNAL dead_locked WAIT_FOR resume_wait_graph NO_CLEAR_EVENT';
+  --send SHOW CREATE USER root@localhost
+  --inc $i
+}
+--enable_query_log
+
+--echo # Connection default: Make 500 threads stop just before deadlock check
+--connection default
+SET DEBUG_SYNC = 'now WAIT_FOR before_dead_lock NO_CLEAR_EVENT';
+
+--echo # Now signal waiting threads to start deadlock checks
+SET DEBUG_SYNC = 'now SIGNAL start_dead_lock_check';
+
+--echo # Waiting for 500 threads to reach before deadlock check state
+--let $wait_timeout = 60
+let $wait_condition=
+  SELECT COUNT(*) = 500 FROM information_schema.processlist
+    WHERE STATE = 'debug sync point: mdl_acquire_lock_wait'
+    AND info = 'SHOW CREATE USER root@localhost';
+
+--echo # Signal 500 threads to start deadlocks checks
+SET DEBUG_SYNC = 'now SIGNAL resume_wait_graph';
+--echo # Now let the initial 200 readers who took the SHARED lock to resume
+--echo # They will not be able to release the lock/ticket to internal rwlock
+--echo # taken as SHARED by Deadlock check threads. These threads need exclusive
+--echo # mode to remove ticket from granted list
+SET DEBUG_SYNC = 'now SIGNAL new_xlock';
+
+--echo # database freezes now and new connections are blocked
+--echo # and finally we hit max_connections limit
+--echo # Freeze start..  Also observe high cpu usage.. mysql server is doing deadlock checks
+
+SELECT NOW() INTO @start;
+
+--connection default
+--echo # Waiting for connections to finish
+--echo # initial_con_count is $initial_con_count
+--let $wait_timeout = 600
+let $wait_condition=
+  SELECT COUNT(*) = $initial_con_count FROM information_schema.processlist;
+--source include/wait_condition.inc
+
+SELECT NOW() INTO @end;
+#--echo # Freeze over..
+#--echo # total time mysqld freezed:
+#--echo #########################################################################
+#SELECT TIMEDIFF(@end, @start) AS freeze_time;
+#--echo #########################################################################
+
+--echo # Cleanup
+--eval SET GLOBAL max_connections = $old_max_conn
+--disable_result_log
+--echo # closing connections
+--let $i = 0
+while ($i <= 700) {
+ --connection con$i
+ --reap
+ --disconnect con$i
+ --inc $i
+}
+--connection con_flush
+--reap
+--disconnect con_flush
+
+--enable_result_log
+
+--connection default
+--echo # all new connections disconnected.
+--source include/wait_until_count_sessions.inc
+DROP USER satya@'%';
+--source include/remove_debug_point.inc
+SET DEBUG_SYNC = RESET;
+--let $assert_text = All the connections should finish within 60 seconds if there is no hang due to ACL
+--let $assert_cond= "[SELECT TIMEDIFF(@end, @start) < 60]" = 1
+--source include/assert.inc
+# Restart to release virtual memory
+--source include/restart_mysqld.inc
--- a/sql/mdl.cc
+++ b/sql/mdl.cc
@@ -3444,7 +3444,14 @@ bool MDL_context::acquire_lock(MDL_request *mdl_request,
   /* There is a shared or exclusive lock on the object. */
   DEBUG_SYNC(get_thd(), "mdl_acquire_lock_wait");
 
-  find_deadlock();
+  /* We skip deadlock checks for ACL cache lock as there is severe contention
+  on internal RW lock of MDL lock. Since ACL cache lock is used during
+  connection process, all connections get stuck and will soon run out of
+  max_connections. All we need is good RW lock implementation. By removing this
+  deadlock checks, we remove contention on internal RW lock of MDL lock. This
+  means if ever there is deadlock due to code bug, we rely on
+  ACL_CACHE_LOCK_TIMEOUT value. */
+  if (mdl_request->key.mdl_namespace() != MDL_key::ACL_CACHE) find_deadlock();
 
   if (lock->needs_notification(ticket) || lock->needs_connection_check()) {
     struct timespec abs_shortwait;
@@ -3820,6 +3827,8 @@ bool MDL_lock::visit_subgraph(MDL_ticket *waiting_ticket,
   bool result = true;
 
   mysql_prlock_rdlock(&m_rwlock);
+
+  DEBUG_SYNC(src_ctx->get_thd(), "acl_mdl_dead_lock");
 
   /*
     Iterators must be initialized after taking a read lock.
--- a/sql/sys_vars.cc
+++ b/sql/sys_vars.cc
@@ -278,6 +278,10 @@ static bool check_session_admin(sys_var *self MY_ATTRIBUTE((unused)), THD *thd,
   DBUG_ASSERT(self->scope() !=
               sys_var::GLOBAL);  // don't abuse check_session_admin()
   Security_context *sctx = thd->security_context();
+
+  /* Skip ACL checks for SET commands */
+  DBUG_EXECUTE_IF("skip_session_admin_check", return false;);
+
   if ((setv->type == OPT_SESSION || setv->type == OPT_DEFAULT) &&
       !sctx->has_global_grant(STRING_WITH_LEN("SESSION_VARIABLES_ADMIN"))
            .first &&
--- a/mysql-test/r/acl_mdl_scalability.result
+++ b/mysql-test/r/acl_mdl_scalability.result
@@ -1,0 +1,71 @@
+#
+# PS-7143 : Server freezes and reaches max connections due to ACL Cache Lock waits
+#
+# Check if system can handle 12k concurrent threads
+# 12K because server creates 6K threads and clients
+# another 6k threads/processes
+# There is no easy way to check if the system can
+# handle 12K threads. We compile C-program (need g++)
+# and execute it and check return status
+# Require g++ to compile this program. If the exec fails "sh: 1: g++: not found". Please install g++ and retry
+# if the below exec fails, it means the operating system has limit of threads less than 12K
+# This testcase needs OS that can create 12k threads.
+# Try these settings (only if they are less 12,000)
+# ulimit -i  12000
+# echo 12000 > /proc/sys/kernel/threads-max
+# echo 60000 > /proc/sys/vm/max_map_count
+# echo 12000 > /proc/sys/kernel/pid_max
+# echo 12000 > /sys/fs/cgroup/pids/user.slice/user-.slice/pids.max
+# Precondition check of the system is completed. The system can handle 12K concurrent threads
+# Adding debug point 'skip_session_admin_check' to @@GLOBAL.debug
+CREATE USER satya@'%' IDENTIFIED BY 'hello';
+GRANT ALL on test.* to satya@'%';
+GRANT ALL on mysql.* to satya@'%';
+SET GLOBAL max_connections = 6000;
+# Sleeping 15 seconds for allowing user to connect client for observability
+# we cannot establish connection later as database freezes. Do now if you wish
+# be in mysql-source/bld/mysql-test directory and execute the below command
+# ../runtime_output_directory/mysql -uroot -S ./var/tmp/mysqld.1.sock
+# connection con0: Hold ACL x lock and let readers wait
+SET DEBUG_SYNC= 'acl_x_lock SIGNAL x_locked WAIT_FOR all_wait_s_locks_done';
+FLUSH PRIVILEGES;
+# 200 connections try for SHARED ACL lock and go into waiting list
+# because FLUSH PRIVILEGES holds EXCLUSIVE lock
+# Connection default. Waiting for 200 connections to reach blocking state
+# Now release X-lock by FLUSH PRIVILEGES. Now All 200 readers get the
+# SHARED lock and they wait on debug sync point after acquiring SHARED
+# ACL MDL lock
+SET DEBUG_SYNC = 'now SIGNAL all_wait_s_locks_done';
+# Wait for 200 connection to reach acquired SHARED ACL MDL lock state
+# Start 5000 connection will go into ACL cache lock waiting state
+FLUSH PRIVILEGES;;
+# Connection default: Wait for those 5000 new connections to reach ACL cache lock waiting state
+# Start another 500 connections and stop them in just before deadlock check iteration
+# Connection default: Make 500 threads stop just before deadlock check
+SET DEBUG_SYNC = 'now WAIT_FOR before_dead_lock NO_CLEAR_EVENT';
+# Now signal waiting threads to start deadlock checks
+SET DEBUG_SYNC = 'now SIGNAL start_dead_lock_check';
+# Waiting for 500 threads to reach before deadlock check state
+# Signal 500 threads to start deadlocks checks
+SET DEBUG_SYNC = 'now SIGNAL resume_wait_graph';
+# Now let the initial 200 readers who took the SHARED lock to resume
+# They will not be able to release the lock/ticket to internal rwlock
+# taken as SHARED by Deadlock check threads. These threads need exclusive
+# mode to remove ticket from granted list
+SET DEBUG_SYNC = 'now SIGNAL new_xlock';
+# database freezes now and new connections are blocked
+# and finally we hit max_connections limit
+# Freeze start..  Also observe high cpu usage.. mysql server is doing deadlock checks
+SELECT NOW() INTO @start;
+# Waiting for connections to finish
+# initial_con_count is 704
+SELECT NOW() INTO @end;
+# Cleanup
+SET GLOBAL max_connections = 151;
+# closing connections
+# all new connections disconnected.
+DROP USER satya@'%';
+# Removing debug point 'skip_session_admin_check' from @@GLOBAL.debug
+SET DEBUG_SYNC = RESET;
+include/assert.inc [All the connections should finish within 60 seconds if there is no hang due to ACL]
+# restart
--- a/sql/auth/sql_auth_cache.cc
+++ b/sql/auth/sql_auth_cache.cc
@@ -2438,6 +2438,8 @@ bool acl_reload(THD *thd, bool mdl_locked) {
 
   if (!acl_cache_lock.lock()) goto end;
 
+  DEBUG_SYNC(thd, "acl_x_lock");
+
   old_acl_users = acl_users;
   old_acl_dbs = acl_dbs;
   old_acl_proxy_users = acl_proxy_users;
@@ -3653,8 +3655,11 @@ void shutdown_acl_cache() {
   delete g_mandatory_roles;
 }
 
-/* Constants used by Acl_cache_lock_guard */
-static const ulong ACL_CACHE_LOCK_TIMEOUT = 3600UL;
+/* Constants used by Acl_cache_lock_guard. Upstream value of 3600UL
+(60mins) is too high. InnoDB max latch wait timeout
+(srv_fatal_semaphore_wait_threshold is 600 sec). 15mins (900 seconds)
+is good enough value for timeout */
+static const ulong ACL_CACHE_LOCK_TIMEOUT = 900UL;
 static const MDL_key ACL_CACHE_KEY(MDL_key::ACL_CACHE, "", "");
 
 /**
--- a/sql/auth/sql_user.cc
+++ b/sql/auth/sql_user.cc
@@ -226,6 +226,8 @@ bool mysql_show_create_user(THD *thd, LEX_USER *user_name,
   Acl_cache_lock_guard acl_cache_lock(thd, Acl_cache_lock_mode::READ_MODE);
   if (!acl_cache_lock.lock()) return true;
 
+  DEBUG_SYNC(thd, "acl_s_lock");
+
   if (!(acl_user =
             find_acl_user(user_name->host.str, user_name->user.str, true))) {
     String wrong_users;


mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
+--source include/count_sessions.inc
```
percona-ysorokin: Put this after all checking `have_xxx.inc` include files (after `--source include/big_test.inc` in this case)

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+
+--echo # Check if system can handle 12k concurrent threads
+--echo # 12K because server creates 6K threads and clients
+--echo # another 6k threads/processes
+
+--echo # There is no easy way to check if the system can
+--echo # handle 12K threads. We compile C-program (need g++)
+--echo # and execute it and check return status
+
+--write_file $MYSQL_TMP_DIR/thread_limit_test.cc
```
percona-ysorokin, line 22: Nice :) This is probably the first time I see a c++ program generated and compiled from within MTR :)

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+
+#define NUM_THREADS     12000
+std::atomic<int> current_threads{0};
+
+void* ThreadTask(void *threadid)
+{
+   long tid;
+   tid = (long)threadid;
+   printf("Hello World! It's me, thread #%ld!\n", tid);
+   std::atomic_fetch_add(&current_threads, 1);
```
percona-ysorokin: The following should work exactly the same
```suggestion
   ++current_threads;
```

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+   while (current_threads.load() < NUM_THREADS) {
+           //printf("sleeping in thread id #%ld!\n", tid);
```
percona-ysorokin: Probably remove?

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+int main ()
+{
+   pthread_t threads[NUM_THREADS];
+   int rc;
+   for(long t=0; t<NUM_THREADS; t++){
+      printf("In main: creating thread %ld\n", t);
+      rc = pthread_create(&threads[t], NULL, ThreadTask, (void *)t);
+      if (rc){
+         printf("ERROR; return code from pthread_create() is %s\n", strerror(rc));
+         exit(-1);
```
percona-ysorokin, line 59: Under ASan, because of resource leaks here, you will get a different exit code (`42`). Please, double check if this is OK.

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+}
+
+--connection default
+--let $initial_con_count=`SELECT COUNT(*) FROM information_schema.processlist`
+
+--echo # Sleeping 15 seconds for allowing user to connect client for observability
+--echo # we cannot establish connection later as database freezes. Do now if you wish
+--echo # be in mysql-source/bld/mysql-test directory and execute the below command
+--echo # ../runtime_output_directory/mysql -uroot -S ./var/tmp/mysqld.1.sock
+#--sleep 15
```
percona-ysorokin, line 123: Commented out code

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+--echo # Now signal waiting threads to start deadlock checks
+SET DEBUG_SYNC = 'now SIGNAL start_dead_lock_check';
+
+--echo # Waiting for 500 threads to reach before deadlock check state
+--let $wait_timeout = 60
+LET $wait_condition=
+  SELECT COUNT(*) = 500 FROM information_schema.processlist
+    WHERE STATE = 'debug sync point: mdl_acquire_lock_wait'
+    AND info = 'SHOW CREATE USER root@localhost';
+#--sleep 30
```
percona-ysorokin: Commented out code

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+ SELECT COUNT(*) = 200 FROM information_schema.processlist
+   WHERE state = 'debug sync point: acl_s_lock' AND
+         info = 'SHOW CREATE USER root@localhost';
+--source include/wait_condition.inc
+
+--echo # Start 5000 connection will go into ACL cache lock waiting state
+--let $i = 1
+while ($i <= 5000) {
+ #--echo "starting new connection $i that will blocked"
+ --exec_in_background $MYSQL -usatya -phello -h 127.0.0.1 -P 13000 test -e "SELECT 1" 2>&1 > /dev/null
```
percona-ysorokin: 13000 may vary in general case
```suggestion
 --exec_in_background $MYSQL -usatya -phello -h 127.0.0.1 -P $MASTER_MYPORT test -e "SELECT 1" 2>&1 > /dev/null
```

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+--echo # They will not be able to release the lock/ticket to internal rwlock
+--echo # taken as SHARED by Deadlock check threads. These threads need exclusive
+--echo # mode to remove ticket from granted list
+SET DEBUG_SYNC = 'now SIGNAL new_xlock';
+
+--echo # database freezes now and new connections are blocked
+--echo # and finally we hit max_connections limit
+--echo # Freeze start..  Also observe high cpu usage.. mysql server is doing deadlock checks
+--echo # try connecting from external client
+--echo # ../runtime_output_directory/mysql -usatya -phello -h 127.0.0.1 -P 13000 test -e "SELECT 1"
```
percona-ysorokin:
```suggestion
--echo # ../runtime_output_directory/mysql -usatya -phello -h 127.0.0.1 -P $MASTER_MYPORT test -e "SELECT 1"
```

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+--echo # and to measure time to connect use
+--echo # time ../runtime_output_directory/mysql -usatya -phello -h 127.0.0.1 -P 13000 test -e "SELECT 1"
```
percona-ysorokin:
```suggestion
--echo # time ../runtime_output_directory/mysql -usatya -phello -h 127.0.0.1 -P $MASTER_MYPORT test -e "SELECT 1"
```

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+
+SELECT NOW() into @start;
```
percona-ysorokin: formatting
```suggestion
SELECT NOW() INTO @start;
```

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+
+--connection default
+--echo # Waiting for connections to finish
+--echo # initial_con_count is $initial_con_count
+--let $wait_timeout = 600
+LET $wait_condition=
+  SELECT COUNT(*) = $initial_con_count FROM information_schema.processlist;
+--source include/wait_condition.inc
+
+SELECT NOW() into @end;
```
percona-ysorokin: formatting
```suggestion
SELECT NOW() INTO @end;
```

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+SET DEBUG_SYNC= 'acl_x_lock SIGNAL x_locked WAIT_FOR all_wait_s_locks_done';
+--send FLUSH PRIVILEGES
+
+--echo # 200 connections try for SHARED ACL lock and go into waiting list
+--echo # because FLUSH PRIVILEGS hold EXCLUSIVE lock
+--let $i = 1
+while ($i <= 200) {
+ --echo # Connection con$i
+ --connection con$i
+ SET DEBUG_SYNC = 'now WAIT_FOR x_locked NO_CLEAR_EVENT';
```
percona-ysorokin, line 137: Here and everywhere, any reason for having `NO_CLEAR_EVENT`?

sql/sys_vars.cc
```diff
@@ -278,6 +278,14 @@ static bool check_session_admin(sys_var *self MY_ATTRIBUTE((unused)), THD *thd,
   DBUG_ASSERT(self->scope() !=
               sys_var::GLOBAL);  // don't abuse check_session_admin()
   Security_context *sctx = thd->security_context();
+
+#ifndef DBUG_OFF
+  /* Skip ACL checks for SET DEBUG and SET DEBUG_SYNC */
```
percona-ysorokin: This rises security concerns. Basically, you allow any user to set `DEBUG` and `DEBUG_SYNC` variables here.

percona-ysorokin (changes requested):

client/mysqltest.cc
```diff
@@ -7442,7 +7442,7 @@ static struct my_option my_long_options[] = {
      REQUIRED_ARG, 500, 1, 10000, nullptr, 0, nullptr},
     {"max-connections", OPT_MAX_CONNECTIONS,
      "Max number of open connections to server", &opt_max_connections,
-     &opt_max_connections, nullptr, GET_INT, REQUIRED_ARG, 128, 8, 5120,
+     &opt_max_connections, nullptr, GET_INT, REQUIRED_ARG, 1024, 8, 9120,
```
venkatesh-prasad-v: Why not pass `DEFAULT_MAX_CONN` instead of the hardcoded value?

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+   }
+
+   printf("\n\nfinish executing main\n\n");
+   return(0);
+}
+EOF
+--echo # Require g++ to compile this program. If the exec fails "sh: 1: g++: not found". Please install g++ and retry
+--exec g++ $MYSQL_TMP_DIR/thread_limit_test.cc -lpthread -o $MYSQL_TMP_DIR/thread_limit 2>&1 > $MYSQL_TMP_DIR/thread_limit_gcc_error.txt
+
+--echo # if the below exec fails, it means you the operating system has limit of threads less than 12K
```
venkatesh-prasad-v: Remove `you`

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+--remove_file $MYSQL_TMP_DIR/thread_limit_gcc_error.txt
+--remove_file $MYSQL_TMP_DIR/thread_limit
+
+--echo # Precondition check of system over. System can handle 12K concurrent threads
+
+create user satya@'%' identified by 'hello';
+grant all  on  test.* to satya@'%';
+grant all  on  mysql.* to satya@'%';
+
+select @@global.max_connections;
```
venkatesh-prasad-v: Let these SQL commands be in uppercase.

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+--echo # echo 12000 > /proc/sys/kernel/threads-max
+--echo # echo 60000 > /proc/sys/vm/max_map_count
+--echo # echo 12000 > /proc/sys/kernel/pid_max
+--echo # echo 12000 > /sys/fs/cgroup/pids/user.slice/user-$UID.slice/pids.max
+--exec $MYSQL_TMP_DIR/thread_limit 2>&1 > $MYSQL_TMP_DIR/thread_limit_gcc_error.txt
+--remove_file $MYSQL_TMP_DIR/thread_limit_test.cc
+--remove_file $MYSQL_TMP_DIR/thread_limit_gcc_error.txt
+--remove_file $MYSQL_TMP_DIR/thread_limit
+
+--echo # Precondition check of system over. System can handle 12K concurrent threads
```
venkatesh-prasad-v: Can we change it to 
```suggestion
--echo # Precondition check of the system is completed. The system can handle 12K concurrent threads.
```

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+ SET DEBUG_SYNC = 'now WAIT_FOR x_locked NO_CLEAR_EVENT';
+ SET DEBUG_SYNC = 'acl_s_lock SIGNAL s_locked WAIT_FOR new_xlock NO_CLEAR_EVENT';
+ --send SHOW CREATE USER root@localhost
+ --inc $i
+}
+
+--echo # Connection default. Waiting for 200 connections to reach blocking state
+--connection default
+--let $wait_timeout = 600
+LET $wait_condition=
```
venkatesh-prasad-v: s/LET/--let/ to make it look uniform with rest of the test case.

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+--let $i = 201
+while ($i <= 700) {
+  --echo "starting new connection $i that will blocked in Deadlock check
+  --connection con$i
+  SET DEBUG_SYNC = 'now WAIT_FOR s_locked NO_CLEAR_EVENT';
+  SET DEBUG_SYNC = 'mdl_acquire_lock_wait SIGNAL before_dead_lock WAIT_FOR start_dead_lock_check NO_CLEAR_EVENT';
+  SET DEBUG_SYNC = 'acl_mdl_dead_lock SIGNAL dead_locked WAIT_FOR resume_wait_graph NO_CLEAR_EVENT';
+  --send SHOW CREATE USER root@localhost
+  --inc $i
+}
```
venkatesh-prasad-v: Not a strong opinion though! These statements seem to flood the result.file. If it makes sense, can we place this loop between the `--disable_query_log` and `--enable_query_log`?

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+--let $wait_timeout = 60
+LET $wait_condition=
+  SELECT COUNT(*) = 500 FROM information_schema.processlist
+    WHERE STATE = 'debug sync point: mdl_acquire_lock_wait'
+    AND info = 'SHOW CREATE USER root@localhost';
+#--sleep 30
+
+--echo # Signal 500 threads to start deadlocks checks
+SET DEBUG_SYNC = 'now SIGNAL resume_wait_graph';
+--echo # New let the initial 200 readers who took the SHARED lock to resume
```
venkatesh-prasad-v: s/New/Now/

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+--echo # and to measure time to connect use
+--echo # time ../runtime_output_directory/mysql -usatya -phello -h 127.0.0.1 -P 13000 test -e "SELECT 1"
+
+SELECT NOW() into @start;
+
+--connection default
+--echo # Waiting for connections to finish
+--echo # initial_con_count is $initial_con_count
+--let $wait_timeout = 600
+LET $wait_condition=
```
venkatesh-prasad-v: s/LET/--let/ to make it look uniform with rest of the test case.

sql/sys_vars.cc
venkatesh-prasad-v: I agree with Yura. Instead of this, we could add something like
```
DBUG_EXECUTE_IF("skip_session_admin_check", { return false });
```
and enable the debug point from the test case, similar to the DD test cases.
```
./dd_is_concurrency.test:SET GLOBAL debug= '+d,skip_dd_table_access_check';
./dd_is_concurrency.test:SET GLOBAL debug= '-d,skip_dd_table_access_check';
./dd_is_debug.test:SET debug = '+d,skip_dd_table_access_check';
./dd_is_debug.test:SET DEBUG="+d,skip_dd_table_access_check";
./dd_is_debug.test:SET DEBUG="-d,skip_dd_table_access_check";
```

sql/mdl.cc
```diff
@@ -3438,7 +3438,14 @@ bool MDL_context::acquire_lock(MDL_request *mdl_request,
...
 
-  find_deadlock();
+  /* We skip deadlock checks for ACL cache lock as there is severe contention
+  on internal RW lock of MDL lock. Since ACL cache lock is used during
+  connection process, all connections get stuck and will soon run out of
+  max_connections. All we need is good RW lock implementation. By removing this
+  deadlock checks, we remove contention on internal RW lock of MDL lock. This
+  means if ever there is deadlock due to code bug, we rely on
+  ACL_CACHE_LOCK_TIMEOUT value. */
+  if (mdl_request->key.mdl_namespace() != MDL_key::ACL_CACHE) find_deadlock();
```
venkatesh-prasad-v, line 3454: I'm fine with the current approach as it is perfect for now. 

Please consider this as a heads up that Oracle is implementing a deadlock detection infrastructure in replication using MDLs to avoid deadlocks with MDLs which may land anytime. Once it happens, we may have to re-evaluate/revert this change in very near future. 

If you are interested in knowing how ACL_CACHE can inlove in deadlock on slave, then please execute the below test.

```
➜  mysql-test git:(8.0) ✗ cat ../../mysql-test/suite/rpl/t/rpl_mts_spco_acl_deadlock.test

CREATE TABLE t1(a INT PRIMARY KEY);
CREATE TABLE t2(a INT PRIMARY KEY);

--source include/sync_slave_sql_with_master.inc

--source include/rpl_connection_slave.inc
BEGIN;
INSERT INTO t2 VALUES(1);

--source include/rpl_connection_master.inc
SET @@GLOBAL.DEBUG='+d,set_commit_parent_100';
INSERT INTO t2 VALUES(1); INSERT INTO t2 VALUES(2); CREATE USER u1;

--sleep 8
--source include/rpl_connection_slave.inc

SHOW PROCESSLIST;
SHOW PROCESSLIST;
SHOW PROCESSLIST;

--echo # we cant do anything more and slave server wont accept new
--echo # connections as well after this command

INSERT INTO t1 VALUES(1);
```
Run the test with 
```
./mtr rpl_mts_spco_acl_deadlock --mysqld=--slave-parallel-workers=8 --mysqld=--slave-parallel-type=1 --mysqld=--slave-preserve-commit-order=1 --mysqld=--binlog-format=statement --nocheck-testcases --mysqld=--innodb-lock-wait-timeout=5

```

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+--echo # because FLUSH PRIVILEGS hold EXCLUSIVE lock
+--let $i = 1
+while ($i <= 200) {
+ --echo # Connection con$i
+ --connection con$i
+ SET DEBUG_SYNC = 'now WAIT_FOR x_locked NO_CLEAR_EVENT';
+ SET DEBUG_SYNC = 'acl_s_lock SIGNAL s_locked WAIT_FOR new_xlock NO_CLEAR_EVENT';
+ --send SHOW CREATE USER root@localhost
+ --inc $i
+}
```
venkatesh-prasad-v: Not a strong opinion though! These statements seem to flood the result.file. If it makes sense, can we place this loop between the --disable_query_log and --enable_query_log?

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+--echo # Connection default: Make 500 threads stop just before deadlock check
+--connection default
+SET DEBUG_SYNC = 'now WAIT_FOR before_dead_lock NO_CLEAR_EVENT';
+
+--echo # Now signal waiting threads to start deadlock checks
+SET DEBUG_SYNC = 'now SIGNAL start_dead_lock_check';
+
+--echo # Waiting for 500 threads to reach before deadlock check state
+--let $wait_timeout = 60
+LET $wait_condition=
```
venkatesh-prasad-v: s/LET/--let/ to make it look uniform with rest of the test case.

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+   --connection con_flush
+   --send FLUSH PRIVILEGES;
+ }
+ --inc $i
+}
+
+--echo # Connection default: Wait for those 5000 new connections to reach ACL cache lock waiting state
+--connection default
+--let $wait_timeout = 600
+LET $wait_condition=
```
venkatesh-prasad-v: s/LET/--let/ to make it look uniform with rest of the test case.

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,255 @@
...
+--source include/wait_condition.inc
+
+--echo # Now release X-lock by FLUSH PRIVILEGES. Now All 200 readers get the
+--echo # SHARED lock and they wait on debug sync point after acquiring SHARED
+--echo # ACL MDL lock
+SET DEBUG_SYNC = 'now SIGNAL all_wait_s_locks_done';
+
+--echo # Wait for 200 connection to reach acquired SHARED ACL MDL lock state
+--let $wait_timeout = 600
+LET $wait_condition=
```
venkatesh-prasad-v: s/LET/--let/ to make it look uniform with rest of the test case.

venkatesh-prasad-v (commented):
Hi Satya, 

Thank you for working on this issue and I really liked the tests, great work!
However, I have a few suggestions. If you don't mind, please see below.

Thank you for your time.

sql/sys_vars.cc
satya-bodapati: This is only on debug builds. The problem is we cannot use debug or debug_sync for this testcase.. The SET DEBUG/DEBUG_SYNC would be stuck waiting for ACL cache lock

satya-bodapati: ifndef DBUG_OFF means.. ifdef DBUG_ON (confusin at first look)

satya-bodapati: DBUG_EXECUTE_IF() would also do ACL check and get stuck

venkatesh-prasad-v: Did you try enabling the debug point globally before the test and disabling it in the end? It worked on my system.
```
 86 select @@global.max_connections;
+87 --let $debug_point = skip_session_admin_check
+88 --source include/add_debug_point.inc
 89 
 90 --connect (con_flush,localhost,root,,)
 91 --connect (con0,localhost,root,,)
```
```
 253 DROP USER satya@'%';
+254 --source include/remove_debug_point.inc
 255 SET DEBUG_SYNC = RESET;
```

satya-bodapati: Good point, Will try.. thanks @venkatesh-prasad-v

sql/mdl.cc
satya-bodapati: I was not able to reproduce this behaviour with or without fix..

on side note:
Lock Ordering issues (in code) are un-acceptable. We should acquire in same order. Server now has some lock ordering check mechanism. Unfortunately it is not automatic.

8.0.18 has one such fix to solve lock ordergin (deadlock) issue. Now all the ACL TABLES are opened first and only then the ACL cache lock.

mysql-test/t/acl_mdl_scalability.test
satya-bodapati: moved it after the program check

mysql-test/t/acl_mdl_scalability.test
satya-bodapati: Thanks :) Without this, the mtr testcase wouldn't work reliably

mysql-test/t/acl_mdl_scalability.test
satya-bodapati: Always thought, explicit std API would generate correct compiler code

mysql-test/t/acl_mdl_scalability.test
satya-bodapati: As long as it is not zero, we are OK

mysql-test/t/acl_mdl_scalability.test
satya-bodapati: Typically when there is only one thread waiting for event, a signal would notify the wiating thread. Incase of many threads waiting for same event, we don't want to clear it as soon as on thread receives it. Something like  pthread_cond_signal() vs broadcast().

See https://dev.mysql.com/doc/internals/en/debug-sync-facility.html

mysql-test/t/acl_mdl_scalability.test
satya-bodapati: Intentional, to do observability, we need the user to create connection. This is time a user has to establish a connection when the test case is running

In fact, I used this approach to fix debug sync issues and arriving at this test case

mysql-test/t/acl_mdl_scalability.test
satya-bodapati: This I removed. Not so useful

mysql-test/t/acl_mdl_scalability.test
satya-bodapati: fixed

sql/mdl.cc
satya-bodapati: I am able to reproduce this hang. Agree that we have to evaluate if Oralce changes ACL MDL or ordering around that. Btw, even now, it is hang and not a deadlock because not all the parties involved are MDL locks.

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,262 @@
...
+#--sleep 15
+#--echo # sleep over
+
+--echo # connection con0: Hold ACL x lock and let readers wait
+--connection con0
+SET DEBUG_SYNC= 'acl_x_lock SIGNAL x_locked WAIT_FOR all_wait_s_locks_done';
+--send FLUSH PRIVILEGES
+
+--echo # 200 connections try for SHARED ACL lock and go into waiting list
+--echo # because FLUSH PRIVILEGS hold EXCLUSIVE lock
```
venkatesh-prasad-v: `s/PRIVILEGS/PRIVILEGES`

sql/sys_vars.cc
```diff
@@ -278,6 +278,10 @@ static bool check_session_admin(sys_var *self MY_ATTRIBUTE((unused)), THD *thd,
   DBUG_ASSERT(self->scope() !=
               sys_var::GLOBAL);  // don't abuse check_session_admin()
   Security_context *sctx = thd->security_context();
+
+  /* Skip ACL checks for SET DEBUG and SET DEBUG_SYNC */
```
venkatesh-prasad-v: This is not specific to only DEBUG and DEBUG_SYNC. Probably we should change it to.
```suggestion
  /* Skip ACL checks for SET commands */
```

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,262 @@
...
+let $wait_condition=
+ SELECT COUNT(*) = 200 FROM information_schema.processlist
+   WHERE state = 'debug sync point: acl_s_lock' AND
+         info = 'SHOW CREATE USER root@localhost';
+--source include/wait_condition.inc
+
+--echo # Start 5000 connection will go into ACL cache lock waiting state
+--let $i = 1
+while ($i <= 5000) {
+ #--echo "starting new connection $i that will blocked"
```
venkatesh-prasad-v: Commented code. Please remove this If it is not required.

satya-bodapati:
Submitted to Jenkins to run only this testcase: https://ps80.cd.percona.com/job/percona-server-8.0-param/749/
Will submit full run after the above results

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,261 @@
+--source include/have_debug.inc
```
venkatesh-prasad-v, line 1: The test seems to fail with valgrind. Probably it is better to skip this test under ASAN and valgrind.
~~~
2020-08-27T12:05:37.509547Z 0 [System] [MY-010931] [Server] /home/venki/work/ps/80/bld/runtime_output_directory/mysqld: ready for connections. Version: '8.0.20-11-debug'  socket: '/home/venki/work/ps/80/bld/mysql-test/var/tmp/mysqld.1.sock'  port: 13000  Source distribution.
Use --max-threads=INT to specify a larger number of threads
and rerun valgrind

valgrind: the 'impossible' happened:
   Max number of threads is too low

host stacktrace:
==133403==    at 0x58046FFA: ??? (in /usr/lib/x86_64-linux-gnu/valgrind/memcheck-amd64-linux)
==133403==    by 0x58047127: ??? (in /usr/lib/x86_64-linux-gnu/valgrind/memcheck-amd64-linux)
==133403==    by 0x58047390: ??? (in /usr/lib/x86_64-linux-gnu/valgrind/memcheck-amd64-linux)
```

Comment (satya-bodapati) [reply mysql-test/t/acl_mdl_scalability.test:1]: ok, let me skip for valgrind. Don't know how to skip for asan.

Comment (venkatesh-prasad-v) [reply mysql-test/t/acl_mdl_scalability.test:1]: --source include/not_asan.inc
--source include/not_valgrind.inc

Comment (satya-bodapati) [reply mysql-test/t/acl_mdl_scalability.test:1]: On asan builds, its passing. Should I keep it?

Comment (venkatesh-prasad-v) [reply mysql-test/t/acl_mdl_scalability.test:1]: Yes. It passes on ASAN build. So, there is no need to skip on asan. But it needs to be skipped on valgrind.

**Comment by satya-bodapati** [comment]
Submitted on Jenkins: https://ps80.cd.percona.com/job/percona-server-8.0-param/752/. Removed big_test intentionally. Because we do not run big tests frequently.

Will see if Jenkins can handle and based on the test result, I will decide if this should be big-test or not :)

File: mysql-test/t/acl_mdl_scalability.test [hunk 32]
```diff
@@ -0,0 +1,262 @@
...
+--echo # They will not be able to release the lock/ticket to internal rwlock
+--echo # taken as SHARED by Deadlock check threads. These threads need exclusive
+--echo # mode to remove ticket from granted list
+SET DEBUG_SYNC = 'now SIGNAL new_xlock';
+
+--echo # database freezes now and new connections are blocked
+--echo # and finally we hit max_connections limit
+--echo # Freeze start..  Also observe high cpu usage.. mysql server is doing deadlock checks
+--echo # try connecting from external client
+--echo # ../runtime_output_directory/mysql -usatya -phello -h 127.0.0.1 -P $MASTER_MYPORT test -e "SELECT 1"
```
venkatesh-prasad-v: We may probably need to mask the port number. As test sometimes fails with
```
@@ -54,9 +58,9 @@
 # and finally we hit max_connections limit
 # Freeze start..  Also observe high cpu usage.. mysql server is doing deadlock checks
 # try connecting from external client
-# ../runtime_output_directory/mysql -usatya -phello -h 127.0.0.1 -P 13000 test -e "SELECT 1"
+# ../runtime_output_directory/mysql -usatya -phello -h 127.0.0.1 -P 13210 test -e "SELECT 1"
 # and to measure time to connect use
-# time ../runtime_output_directory/mysql -usatya -phello -h 127.0.0.1 -P 13000 test -e "SELECT 1"
+# time ../runtime_output_directory/mysql -usatya -phello -h 127.0.0.1 -P 13210 test -e "SELECT 1"
 SELECT NOW() INTO @start;
 # Waiting for connections to finish
 # initial_con_count is 704
```

venkatesh-prasad-v (commented):
Hi Satya, the test is unstable on Jenkins and sometimes fails with timeout error. Could you please have a look at this failure?
```
CURRENT_TEST: main.acl_mdl_scalability
--- /tmp/results/PS/mysql-test/r/acl_mdl_scalability.result	2020-08-28 09:55:52.000000000 +0300
+++ /tmp/results/PS/mysql-test/var/4/log/acl_mdl_scalability.reject	2020-08-28 14:55:49.079520534 +0300
@@ -37,9 +37,13 @@
 # Start 5000 connection will go into ACL cache lock waiting state
 FLUSH PRIVILEGES;;
 # Connection default: Wait for those 5000 new connections to reach ACL cache lock waiting state
+Timeout in wait_condition.inc for SELECT COUNT(*) = 5000 FROM information_schema.processlist
+WHERE USER = 'unauthenticated user' AND STATE = 'Waiting for acl cache lock'
 # Start another 500 connections and stop them in just before deadlock check iteration
 # Connection default: Make 500 threads stop just before deadlock check
 SET DEBUG_SYNC = 'now WAIT_FOR before_dead_lock NO_CLEAR_EVENT';
+Warnings:
+Warning	1639	debug sync point wait timed out
 # Now signal waiting threads to start deadlock checks
 SET DEBUG_SYNC = 'now SIGNAL start_dead_lock_check';
 # Waiting for 500 threads to reach before deadlock check state
```

satya-bodapati:
> Hi Satya, the test is unstable on Jenkins and sometimes fails with timeout error. Could you please have a look at this failure?
> 
> ```
> CURRENT_TEST: main.acl_mdl_scalability
> --- /tmp/results/PS/mysql-test/r/acl_mdl_scalability.result	2020-08-28 09:55:52.000000000 +0300
> +++ /tmp/results/PS/mysql-test/var/4/log/acl_mdl_scalability.reject	2020-08-28 14:55:49.079520534 +0300
> @@ -37,9 +37,13 @@
>  # Start 5000 connection will go into ACL cache lock waiting state
>  FLUSH PRIVILEGES;;
>  # Connection default: Wait for those 5000 new connections to reach ACL cache lock waiting state
> +Timeout in wait_condition.inc for SELECT COUNT(*) = 5000 FROM information_schema.processlist
> +WHERE USER = 'unauthenticated user' AND STATE = 'Waiting for acl cache lock'
>  # Start another 500 connections and stop them in just before deadlock check iteration
>  # Connection default: Make 500 threads stop just before deadlock check
>  SET DEBUG_SYNC = 'now WAIT_FOR before_dead_lock NO_CLEAR_EVENT';
> +Warnings:
> +Warning	1639	debug sync point wait timed out
>  # Now signal waiting threads to start deadlock checks
>  SET DEBUG_SYNC = 'now SIGNAL start_dead_lock_check';
>  # Waiting for 500 threads to reach before deadlock check state
> ```

Yes, best to make it non-parallel and enhance Jenkins to include --non-parallel-test option. This will ensure all such tests are run at end sequentially.

Will add this patch and submit to Jenkins again

mysql-test/t/acl_mdl_scalability.test
satya-bodapati: right

satya-bodapati:
https://ps80.cd.percona.com/job/percona-server-8.0-param/756/

satya-bodapati:
@venkatesh-prasad-v implemented the testcase with debug assertion. We have two testcases now

sql/mdl.h
```diff
@@ -1654,7 +1654,7 @@ class MDL_context {
   bool fix_pins();
 
  public:
-  void find_deadlock();
+  void find_deadlock(MDL_request *mdl_request=nullptr);
```
venkatesh-prasad-v: Please run `clang-format-10` of all .h and .cc files.

sql/mdl.cc
```diff
@@ -3996,7 +4006,10 @@ bool MDL_context::visit_subgraph(MDL_wait_for_graph_visitor *gvisitor) {
         fact by setting VICTIM status to its wait slot.
 */
 
-void MDL_context::find_deadlock() {
+void MDL_context::find_deadlock(MDL_request *mdl_request) {
```
venkatesh-prasad-v: Azure Pipelines are failing for release builds with below error.
```
[ 48%] Building CXX object unittest/gunit/CMakeFiles/sqlgunitlib.dir/__/__/sql/stateless_allocator.cc.o
[ 48%] Building CXX object unittest/gunit/CMakeFiles/sqlgunitlib.dir/__/__/sql-common/sql_string.cc.o
/home/vsts/work/1/s/sql/mdl.cc:4009:46: error: unused parameter 'mdl_request' [-Werror=unused-parameter]
 void MDL_context::find_deadlock(MDL_request *mdl_request) {
                                              ^
cc1plus: all warnings being treated as errors
unittest/gunit/CMakeFiles/sqlgunitlib.dir/build.make:92: recipe for target 'unittest/gunit/CMakeFiles/sqlgunitlib.dir/__/__/sql/mdl.cc.o' failed
make[2]: *** [unittest/gunit/CMakeFiles/sqlgunitlib.dir/__/__/sql/mdl.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
Scanning dependencies of target ddl_rewriter-t
```
Please place the argument within `#ifndef DBUG_OFF` and `#endif` blocks.

mysql-test/t/acl_mdl.test
```diff
@@ -0,0 +1,41 @@
+--source include/have_debug_sync.inc
+--source include/have_debug.inc
+--source include/count_sessions.inc
+
+--echo #
+--echo # PS-7143 : Server freezes and reaches max connections due to ACL Cache Lock waits
+--echo #
+
+--echo # This test will ensure that when deadlock checks are done for waiting lock, it
+--echo # cannot be ACL Lock. A debug assertion will fail
```
venkatesh-prasad-v: This makes the reader to think that the following test will cause an assertion failure, which is wrong. IMO, it is better to remove the last sentence from the line and just keep the rest as is.

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,265 @@
...
+--source include/not_windows.inc
+# Because of 6K connections, mysqld consumes close to
+# 14G virtual memory
+--source include/big_test.inc
+--source include/not_valgrind.inc
+# Because of resource consumption, run this test alone
+# Use --non-parallel-test option to mtr
+--source include/not_parallel.inc
+
+--echo # Check if system can handle 12k concurrent threads
```
venkatesh-prasad-v, line 14: Please add a reference to the JIRA issue, similar to `acl_mdl.test`.

sql/mdl.cc
satya-bodapati: fixed using MY_ATTRIBUTE

satya-bodapati:
https://ps80.cd.percona.com/job/percona-server-8.0-param/765/  -> main.acl_mdl_scability
https://ps80.cd.percona.com/job/percona-server-8.0-param/766/ -> main suites

@venkatesh-prasad-v  tests are complete, Results look good

venkatesh-prasad-v (approved):
Hi Satya! 

Thank you for the excellent analysis on this bug and for addressing all the review comments. Great job!
The patch looks good now and I have no more comments. So, it is approved from my side.

But please wait for Yura's approval as I believe he will have something to tell.

Thank you for your time.

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,269 @@
...
+let $wait_condition=
+ SELECT COUNT(*) = 200 FROM information_schema.processlist
+   WHERE state = 'debug sync point: acl_s_lock' AND
+         info = 'SHOW CREATE USER root@localhost';
+--source include/wait_condition.inc
+
+--echo # Start 5000 connection will go into ACL cache lock waiting state
+--let $i = 1
+while ($i <= 5000) {
+ --exec_in_background $MYSQL -usatya -phello -h 127.0.0.1 -P 13000 test -e "SELECT 1" 2>&1 > /dev/null
```
percona-ysorokin: @satya-bodapati `1300` may still be an incorrect value in general case (even if run in non-parallel mode)
I believe `$MASTER_MYPORT` should be used instead.

mysql-test/t/acl_mdl_scalability.test
percona-ysorokin: This was my mistake - putting `$MASTER_MYPORT` into echo was a very bad idea :(

mysql-test/t/acl_mdl_scalability.test
percona-ysorokin: Likewise

percona-ysorokin (changes requested):

mysql-test/t/acl_mdl_scalability.test
satya-bodapati: Ah, this is the reason, it was not working with --parallel. Good find. Fixed

mysql-test/t/acl_mdl_scalability.test
satya-bodapati: on changing the comment to $MY_PORT, it is tricky.

1. --replace_result doesn't work on --echo
2.  The reason behind this message is to help user when debugging. (possibly without result file. That is how I measured before adding the code for freeze time). On second thought, since we measure hang time, we can get rid of this prints altogether.

mysql-test/t/acl_mdl_scalability.test
```diff
@@ -0,0 +1,264 @@
...
+--echo # Check if system can handle 12k concurrent threads
+--echo # 12K because server creates 6K threads and clients
+--echo # another 6k threads/processes
+
+--echo # There is no easy way to check if the system can
+--echo # handle 12K threads. We compile C-program (need g++)
+--echo # and execute it and check return status
+
+--write_file $MYSQL_TMP_DIR/thread_limit_test.cc
+#include <pthread.h>
```
kamil-holubicki, line 23: Do we need to use pthread here? Wouldn't std::thread do the job? 
Also 
printf => std::cout
sleep => std::this_thread::sleep_for

satya-bodapati: that can work too

mysql-test/t/acl_mdl_scalability.test
kamil-holubicki: What prevents us from stopping/joining previously created threads here to avoid the leak?

satya-bodapati: we can.. since this is a test program for testcase and since we terminate, they will get cleaned up.

satya-bodapati:
Latest Jenkins Link: https://ps80.cd.percona.com/job/percona-server-8.0-param/768/#showFailuresLink

The test is now a normal big-test. Can be run parallelly too.

@percona-ysorokin ^

percona-ysorokin (approved):
LGTM

PR #4006, opened 2020-08-26
PS-7270: admin_port should not accept proxied connections

imp3142:


percona-ysorokin (commented):
@impimp Everything looks logical
However, ideally this commit should also include an MTR test case which simulates the scenario mentioned in the Jira ticket

PR #4009, opened 2020-08-31
PS-7245: Block redo log disabling with LTFB

dutow:
Problem: upstream blocks the ALTER INSTANCE DISABLE INNODB REDO_LOG
statements if there is a backup lock is held. This is required to avoid
taking corrupt backups.
PS however also has the LOCK TABLES FOR BACKUP lock, which is preferred
by xtrabackup when taking a backup from a Percona Server. This lock
currently doesn't block the above alter instance statement.

Fix: block the statement when LTFB lock is held.

### DIFF
--- a/sql/sql_alter_instance.cc
+++ b/sql/sql_alter_instance.cc
@@ -269,6 +269,13 @@ bool Innodb_redo_log::execute() {
     return true;
   }
 
+  // Acquire Percona's LOCK TABLES FOR BACKUP lock
+  if (m_thd->backup_tables_lock.abort_if_acquired() ||
+      m_thd->backup_tables_lock.acquire_protection(
+          m_thd, MDL_TRANSACTION, m_thd->variables.lock_wait_timeout)) {
+    return true;
+  }
+
   if (hton->redo_log_set_state(m_thd, m_enable)) {
     /* SE should have raised error */
     DBUG_ASSERT(m_thd->get_stmt_da()->is_error());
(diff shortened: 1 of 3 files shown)


sql/sql_alter_instance.cc
```diff
@@ -269,6 +269,12 @@ bool Innodb_redo_log::execute() {
     return true;
   }
```
satya-bodapati, line 271: Please add a comment that this lock is Percona's LOCK TABLES FOR BACKUP.

sql/sql_alter_instance.cc
```diff
@@ -269,6 +269,12 @@ bool Innodb_redo_log::execute() {
...
 
+  if (m_thd->backup_tables_lock.abort_if_acquired() ||
+      m_thd->backup_tables_lock.acquire_protection(
+          m_thd, MDL_STATEMENT, m_thd->variables.lock_wait_timeout)) {
```
satya-bodapati: why MDL_STATEMENT? shouldn't it be MDL_TRANSACTION?

satya-bodapati (changes requested):

sql/sql_alter_instance.cc
dutow: Commend added.

sql/sql_alter_instance.cc
dutow: Statement seemed to be enough, but now that I've checked it, the other backup lock is also acquired with MDL_TRANSACTION. Changed.

satya-bodapati (changes requested):
Fix looks good but some sugar coating required  :blush: 

Please add to commit message and description in the following format

Problem:
------------

Fix:
----

Also submit Jenkins link as comment here

sql/sql_alter_instance.cc
satya-bodapati: thanks

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/767/#showFailuresLink

satya-bodapati (approved):

PR #4014, opened 2020-09-04
Fixed PS-7279 (Stacktrace and coredumps not generating on Percona ser…

altmannmarcelo:
…ver) (5.7)

https://jira.percona.com/browse/PS-7279

This is a regression introduced on PS-7114. It was assumed that a
BuildID would always be avaialbe on ELF Header section.
This fix makes the server print BuildID: Not Available in case the
server has been compiled with --build-id=none.

### DIFF
--- a/mysys/stacktrace.c
+++ b/mysys/stacktrace.c
@@ -259,6 +259,7 @@ void my_print_buildID()
   unsigned char *build_id;
   char           buff[3];
   uint           i;
+  uint           bytes_read;
   snprintf(proc_path, sizeof(proc_path), "/proc/%d/exe", getpid());
   readlink_bytes= readlink(proc_path, mysqld_path, sizeof(mysqld_path));
   if (readlink_bytes < 1)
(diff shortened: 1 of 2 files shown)


altmannmarcelo:
https://ps57.cd.percona.com/job/percona-server-5.7-param/336/

mysys/stacktrace.c
```diff
@@ -285,10 +286,17 @@ void my_print_buildID()
...
+  bytes_read= sizeof(ElfW(Nhdr));
   while (nhdr->n_type != NT_GNU_BUILD_ID)
   {
     nhdr= (ElfW(Nhdr) *)((size_t)nhdr + sizeof(ElfW(Nhdr)) + nhdr->n_namesz +
                          nhdr->n_descsz);
+    bytes_read+= nhdr->n_namesz + nhdr->n_descsz;
+    if (bytes_read > phdr->p_filesz)
+    {
+      my_safe_printf_stderr("Build ID: Not Available \n");
+      return;
```
dutow, line 299: There's a missing fclose here

altmannmarcelo: @dutow good catch. Fixed

dutow (approved):

PR #4015, opened 2020-09-07
Fixed PS-7279 (Stacktrace and coredumps not generating on Percona ser…

altmannmarcelo:
…ver) (8.0)

https://jira.percona.com/browse/PS-7279

This is a regression introduced on PS-7114. It was assumed that a
BuildID would always be available on ELF Header section.
This fix makes the server print BuildID: Not Available in case the
server has been compiled with --build-id=none.


altmannmarcelo:
https://ps80.cd.percona.com/job/percona-server-8.0-param/769/

dutow (approved):

PR #4016, opened 2020-09-07
PS-7284: Fix MTR test innodb.percona_changed_page_bmp_requests_debug

oleksandr-kachan:
The test used bitmap files number after each action to find out if
that action actually took place and bitmap data is updated.
This approach became not relaiable. The amount of data written
to bitmap files during each test run is the same but checkpoints
are generated at different LSN positions. As a result number of
generated bitmap files may be different.

To make test stable instead of just comparing results with the
predefined bitmap file names check how currently active file
name and its size changes after each action. Added separate
.inc file which checks current state of bitmap files.

### DIFF
--- a/mysql-test/suite/innodb/t/percona_changed_page_bmp_requests_debug.test
+++ b/mysql-test/suite/innodb/t/percona_changed_page_bmp_requests_debug.test
@@ -68,6 +68,10 @@ EOF
 --let $assert_cond = @@innodb_file_per_table = 0
 --source include/assert.inc
 
+# Read current state of bitmaps after restarts
+--let $bmp_data_dir = $CUSTOM_INNODB_DATA_HOME_DIR
+--source suite/innodb/include/percona_get_last_bmp_file_name_size.inc
+
 --echo
 --echo # Creating a simple table and filling it with some records.
 CREATE TABLE t1(id INT) ENGINE=InnoDB STATS_PERSISTENT=0 STATS_AUTO_RECALC=0;
@@ -79,6 +83,16 @@ FLUSH TABLES t1;
 SET @@GLOBAL.innodb_log_checkpoint_now=TRUE;
 SET @@GLOBAL.innodb_track_redo_log_now=TRUE;
 
+--let $prev_last_bmp_file_name = $last_bmp_file_name
+--let $prev_last_bmp_file_size = $last_bmp_file_size
+--source suite/innodb/include/percona_get_last_bmp_file_name_size.inc
+
+--echo
+--echo # Checking if bitmap files updated after the first flush
+if (`SELECT "$last_bmp_file_name" = "$prev_last_bmp_file_name" AND $last_bmp_file_size = $prev_last_bmp_file_size`) {
+    --die "Bitmap files should be updated after the flush"
+}
+
 --echo
 --echo # Inserting and flushing one more time for compatibility with 5.5.
 INSERT INTO t1 VALUES (1), (2), (3), (4), (5), (6), (7), (8), (9), (10);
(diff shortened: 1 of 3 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/771

mysql-test/suite/innodb/t/percona_changed_page_bmp_requests_debug.test
```diff
@@ -116,23 +139,31 @@ eval PURGE CHANGED_PAGE_BITMAPS BEFORE $previous_lsn;
...
---replace_regex /_[[:digit:]]+\.xdb$/_<nnn>.xdb/
---list_files $CUSTOM_INNODB_DATA_HOME_DIR *.xdb
+--echo # Checking .xdb files after the restart with tracking disabled.
+--echo # There must stay some .xdb files.
+--let $prev_last_bmp_file_name = $last_bmp_file_name
+--source include/percona_get_last_bmp_file_name_size.inc
+
+if (`SELECT $bmp_files_count = 0`) {
+    --die "There must stay some .xdb files after restart (tracking enabled)"
+}
```
satya-bodapati: How do we verify that there is a new .xdb file? Should we track the count before the restart and after and check the difference is 1?

mysql-test/suite/innodb/t/percona_changed_page_bmp_requests_debug.test
```diff
@@ -68,6 +68,10 @@ EOF
 --let $assert_cond = @@innodb_file_per_table = 0
 --source include/assert.inc
 
+# Read current state of bitmaps after restarts
+--let $bmp_data_dir = $CUSTOM_INNODB_DATA_HOME_DIR
+--source include/percona_get_last_bmp_file_name_size.inc
+
```
satya-bodapati: Since this is debug test, I would have skipped checkpoints to get stability SET GLOBA innodb_checkpooint_disable=TRUE;

As you have already added the .inc logic, OK to keep this.

mysql-test/include/percona_get_last_bmp_file_name_size.inc
```diff
@@ -0,0 +1,62 @@
+# Purpose:
+# To get name and size of the latest file written
+# by online log tracking. In additional it returns
+# number of bitmap files found in the provided directory.
+#
+# Parameters:
+# --let $bmp_data_dir = path/to/bmp/files/dir
```
satya-bodapati: for InnoDB related stuff, the inc file should be in mysql-test/suite/innodb/include

satya-bodapati (changes requested):

mysql-test/include/percona_get_last_bmp_file_name_size.inc
oleksandr-kachan: Done

mysql-test/suite/innodb/t/percona_changed_page_bmp_requests_debug.test
oleksandr-kachan: We could do so but when testing it I noticed that this number of files differ for different runs - sometimes there are as expected two files but sometimes there may stay only one but of bigger size. So I decided just to check that at least one file is present at this point. Looks like this should be enough. Do you think it is important to have exactly two files here?

satya-bodapati: Ok, fine to keep it as is

satya-bodapati (approved):
LGTM!

PR #4017, opened 2020-09-07
PS-7296: Fix existing bitmap files processing during initialization (8.0)

oleksandr-kachan:
In log_online_read_init there is an attempt to check the most recent
bitmap file and continue writing it or rotate if needed. At the moment
server fails to open the most recent bitmap file in case
--innodb-data-home-dir and --datadir are set to different directories.

With such configuration current working directory differs from innodb
data home directory. At the same time during initialization
log_bmp_sys->bmp_file_home isn't taken into account when trying
to open existing bitmap file. As a result it tries to find it
in current working directory.

### DIFF
--- a/storage/innobase/log/log0online.cc
+++ b/storage/innobase/log/log0online.cc
@@ -934,11 +934,8 @@ void log_online_read_init(void) {
         bitmap_dir->dir_entry[i].mystat->st_size > 0) {
       log_bmp_sys->out_seq_num = file_seq_num;
       last_file_start_lsn = file_start_lsn;
-      /* No dir component (log_bmp_sys->bmp_file_home) here, because
-that's the cwd */
-      strncpy(log_bmp_sys->out.name, bitmap_dir->dir_entry[i].name,
-              FN_REFLEN - 1);
-      log_bmp_sys->out.name[FN_REFLEN - 1] = '\0';
+      snprintf(log_bmp_sys->out.name, sizeof(log_bmp_sys->out.name), "%s%s",
+               log_bmp_sys->bmp_file_home, bitmap_dir->dir_entry[i].name);
     }
   }
   my_dirend(bitmap_dir);
--- a/mysql-test/suite/innodb/r/percona_changed_page_bmp.result
+++ b/mysql-test/suite/innodb/r/percona_changed_page_bmp.result
@@ -80,6 +80,7 @@ SELECT @@innodb_track_changed_pages;
 12th restart
 # restart:--innodb-data-home-dir=tmpdatadir --datadir=MYSQLD_DATADIR1
 ib_modified_log_1
+ib_modified_log_2
 include/assert.inc [INFORMATION_SCHEMA.INNODB_CHANGED_PAGES is not empty]
 13th restart
 # restart:--innodb-track-changed-pages=TRUE


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/774

storage/innobase/log/log0online.cc
```diff
@@ -900,26 +900,17 @@ void log_online_read_init(void) {
...
-  ut_a(srv_data_home_len < FN_REFLEN);
-  strcpy(log_bmp_sys->bmp_file_home, srv_data_home);
-  if (srv_data_home_len &&
-      log_bmp_sys->bmp_file_home[srv_data_home_len - 1] != SRV_PATH_SEPARATOR) {
-    ut_a(srv_data_home_len < FN_REFLEN - 1);
-    log_bmp_sys->bmp_file_home[srv_data_home_len] = SRV_PATH_SEPARATOR;
-    log_bmp_sys->bmp_file_home[srv_data_home_len + 1] = '\0';
-  }
+  /* Initialize bitmap file directory with a full path to srv_data_home. */
+  log_bmp_sys->bmp_file_home = Fil_path::get_real_path(srv_data_home);
```
satya-bodapati: I don't understand this full path requirement. We should be able to open the file with the relative pathname also "./ib_modified_log*.xb" isn't it?

satya-bodapati (changes requested):

storage/innobase/log/log0online.cc
oleksandr-kachan: Sorry Satya, I originally found it while testing a different issue so confused a root cause a bit... I've updated fix and description.

oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/780

storage/innobase/log/log0online.cc
```diff
@@ -934,11 +934,8 @@ void log_online_read_init(void) {
         bitmap_dir->dir_entry[i].mystat->st_size > 0) {
       log_bmp_sys->out_seq_num = file_seq_num;
       last_file_start_lsn = file_start_lsn;
-      /* No dir component (log_bmp_sys->bmp_file_home) here, because
-that's the cwd */
-      strncpy(log_bmp_sys->out.name, bitmap_dir->dir_entry[i].name,
-              FN_REFLEN - 1);
-      log_bmp_sys->out.name[FN_REFLEN - 1] = '\0';
+      snprintf(log_bmp_sys->out.name, sizeof(log_bmp_sys->out.name), "%s%s",
+               log_bmp_sys->bmp_file_home, bitmap_dir->dir_entry[i].name);
```
satya-bodapati, line 938: are you sure the bmp_file_home always contains the path separator at end?

satya-bodapati (changes requested):

storage/innobase/log/log0online.cc
satya-bodapati: Thanks for fixing the description and commit message, it is better now

storage/innobase/log/log0online.cc
oleksandr-kachan: It's initialized in the following way, so it should always be there

> /* Initialize bitmap file directory from srv_data_home and add a path
>   separator if needed.  */
>   const size_t srv_data_home_len = strlen(srv_data_home);
>   ut_a(srv_data_home_len < FN_REFLEN);
>   strcpy(log_bmp_sys->bmp_file_home, srv_data_home);
>   if (srv_data_home_len &&
>       log_bmp_sys->bmp_file_home[srv_data_home_len - 1] != SRV_PATH_SEPARATOR) {
>     ut_a(srv_data_home_len < FN_REFLEN - 1);
>     log_bmp_sys->bmp_file_home[srv_data_home_len] = SRV_PATH_SEPARATOR;
>     log_bmp_sys->bmp_file_home[srv_data_home_len + 1] = '\0';
>   }
>

satya-bodapati (approved):

PR #4021, opened 2020-09-08
Fixed PS-7297 (proxy_protocol_admin_port test failing on 8.0.21)

altmannmarcelo:
https://jira.percona.com/browse/PS-7297

MySQL 8.0.21 introduced a new SSL context for the admin port. This allows
the admin interface has its own SSL configuration. In case you do not
specify SSL certificates for the admin interface, a new warning will
be displayed indicating the missing SSL certs and the main SSL context will
also be used for the admin interface. This was causing the MTR test to fail due
to the new warning.
To fix it the test now disables SSL.


altmannmarcelo:
https://ps80.cd.percona.com/job/percona-server-8.0-param/775

percona-ysorokin (approved):
LGTM
s/especify/specify/ in the commit message

PR #4024, opened 2020-09-14
PS-7289: Limit innodb_encryption_threads to 255 (5.7)

oleksandr-kachan:
Server crashes in case innodb_encryption_threads
is set to too big value. Limiting the max allowed
value to 255.


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/338/

satya-bodapati (approved):

PR #4025, opened 2020-09-14
PS-7289: Limit innodb_encryption_threads to 255 (8.0)

oleksandr-kachan:
Server crashes in case innodb_encryption_threads
is set to too big value. Limiting the max allowed
value to 255.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -206,6 +206,7 @@ static my_bool	innodb_optimize_fulltext_only		= FALSE;
 
 static char*	innodb_version_str = (char*) INNODB_VERSION_STR;
 
+static const uint MAX_ENCRYPTION_THREADS = 255;
 extern uint srv_fil_crypt_rotate_key_age;
 extern uint srv_n_fil_crypt_iops;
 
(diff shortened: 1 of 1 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/776/

storage/innobase/handler/ha_innodb.cc
```diff
@@ -24265,8 +24266,8 @@ static MYSQL_SYSVAR_UINT(
     encryption_threads, srv_n_fil_crypt_threads_requested, PLUGIN_VAR_RQCMDARG,
     "Number of threads performing background key rotation and "
     "scrubbing",
-    innodb_encryption_threads_validate, innodb_encryption_threads_update,
-    srv_n_fil_crypt_threads_requested, 0, UINT_MAX32, 0);
+    innodb_encryption_threads_validate, innodb_encryption_threads_update, 0, 0,
```
satya-bodapati, line 24269: why did you remove srv_n_fil_crypt_threads_requested?

satya-bodapati (changes requested):

storage/innobase/handler/ha_innodb.cc
satya-bodapati: Hmm, the variable is assigned to 0, so the default is 0. OK to use explicit value.

satya-bodapati (approved):

PR #4030, opened 2020-09-15
PS-7300 : Sesssion temporary tablespace truncation on connection disc…

satya-bodapati:
…onnect causes high CPU usage

Background
----------
Session temporary tablespaces (aka IBT: InnodB Temporary tablespaces) introduced
in 8.0 are allocated per connection on three things

1) CREATE TEMPORARY TABLE by user
2) SELECT queries that use optimizer internal temporary table.(aka intrinsic
   tables). It is rare for SELECTs to use InnoDB for optimizer
   internal tables. Server mostly uses temp table engine. To force SELECT to use
   InnoDB temporary tables, memory has to exceed temptable_max_ram and
   temptable_use_mmap has to be OFF.
3) Information schema queries that bring in large data, use InnoDB temporary table.

On connection disconnect, the tablespaces are truncated and released back to the
pool of IBT tablespaces. This way all the disk space used by the above tables is
reclaimed on disconnect

Problem
--------
The IBT truncation at session disconnect iterates LRU to remove all pages
belonging to IBT tablespace. This was done so that space_id could be reused.
Iteration of LRU is expensive if the buffer pool is big (like 200G/300G/500G so
on) and filled with pages. The LRU list in each instance would be huge. Incase
there is some IO happening on any page, the scan resarts and make it even more
expensive in terms of CPU usage.

High CPU usage caused by the truncation can affect other queries that are
running when disconnect started

Fix:
----
We will follow the new undo truncation style introduced in 8.0.21 (almost).
we will use drop+create and use different space_ids. This approach also removes
LRU scan involved (see BUF_REMOVE_NONE).

The reason I say almost is that we cannot use a checkpoint lower watermark
(lwm) to remove the dropped tablespace object. When InnoDB determines,
checkpoint LSN, the oldest modification of the entire buffer pool, the pages
belonging to temporary tablespace are skipped. This is fine because temp
tablespaces changes are not redo logged and by considering them we cause
un-necessary flushing of the temp tablespaces page for lwm.

We will still rely on the removal of dirty pages of dropped IBTs (Just removal from
flush_list. Not flushed to disk). But.. we don't do this on every drop of IBT.

For every 7,000 IBTs, we drop the in-mem tablespace object aka fil_space_t. The
reason for choosing 7K is because the memory taken by 7K objects is 7K * 931
bytes ~ 6.2MB

Before freeing the 7K IBTs, we just do one iteration of flush_list to remove
(and not flush to disk) IBT pages.

Note that even after flush list removal of pages, we can still have pages
belonging to IBT tablespaces. If there is page creation and the page exists in
LRU, we re-initialize page and use it. If the page didn't exist in the buffer pool,
we would have asked for a free block. So it is safe to let the pages of IBTs 
"age-out". Also, we disable AHI for temp tables, so there is no need to remove
AHI entries of pages in LRU.

Note: On Debug builds, we chose the frequency of IBT fil_space_t removal as 10 to
stress the deletion process.


satya-bodapati:
https://ps80.cd.percona.com/job/percona-server-8.0-param/797/

storage/innobase/include/srv0tmp.h
```diff
@@ -307,6 +312,10 @@ class Tablespace_pool {
   Pool *m_active;
   /** Mutex to protect concurrent operations on the pool */
   ib_mutex_t m_mutex;
+
+  /** Next available space_id for tablespace. These are
+  hardcoded space_ids at higher range */
+  static space_id_t m_last_used_space_id;
```
percona-ysorokin, line 357: Shouldn't we use some kind of protection for this `m_last_used_space_id` (atomic var or mutex). Or even for the whole `Tablespace_pool::expand()` function body.
As far as I understood `Tablespace_pool::expand()` is called from `Tablespace_pool::get()` which in turn can be called concurrently?

satya-bodapati: There is protection yura:   ut_ad(!m_pool_initialized || mutex_own(&m_mutex)); -> at beginning of function

percona-ysorokin: Oops, sorry. My bad :(
There is indeed

satya-bodapati: np yura :)

lalitc18:
**Test scenarios:**

```
create temp table and disconnet:
 --query="CREATE TEMPORARY TABLE t1(a INT, b varchar (30) ,PRIMARY KEY (a),KEY b_1 (b)); INSERT INTO t1 VALUES (1,'asdgas'),(2,'sadads'),(3,'weuawgd'),(4, 'wgGWDA'),(5, 'AJDSFksd'); SELECT * FROM t1; DROP TEMPORARY TABLE t1; "
  --iterations=200

```
**Test-Case1:** mysqlslap with 64,128,256,512,1024 threads and small Buffer Pool (10GB) with LRU filled.
**Test-Case2:** mysqlslap with 64,128,256,512,1024 threads and bigger Buffer Pool (170GB) with LRU filled.

Tested both cases for PS-8.0.21-GA and PS-7300 Fix Patch.

In the test, we found that,

**With PS 8.0.21-GA:**
 
- With a small buffer pool(10GB) mysqlslap test with 64,128,256,512,1024 threads complete successfully. 
mysqlslap results: --iterations=200

```
Benchmark
 Average number of seconds to run all queries: 0.020 seconds
 Minimum number of seconds to run all queries: 0.015 seconds
 Maximum number of seconds to run all queries: 0.082 seconds
 Number of clients running queries: 64
 Average number of queries per client: 1

Benchmark
	Average number of seconds to run all queries: 0.347 seconds
	Minimum number of seconds to run all queries: 0.293 seconds
	Maximum number of seconds to run all queries: 0.608 seconds
	Number of clients running queries: 1024
	Average number of queries per client: 1
 
```

- With a large buffer pool (170GB) mysqlslap test with 64 thread taking a long time even after 2 hours, it was still running. This is due to the Temp table,  session disconnect, and high CPU usage when we have a large buffer pool.
 

**With Fix Patch:**  https://github.com/percona/percona-server/pull/4030

Regardless of the buffer size small or big, all mysqlslap tests (tested with 64,1024 threads) completed very quickly (~2 mins for 1024 threads test run)

mysqlslap results: --iterations=200

```
Benchmark
	Average number of seconds to run all queries: 0.022 seconds
	Minimum number of seconds to run all queries: 0.017 seconds
	Maximum number of seconds to run all queries: 0.073 seconds
	Number of clients running queries: 64
	Average number of queries per client: 1

Benchmark
	Average number of seconds to run all queries: 0.409 seconds
	Minimum number of seconds to run all queries: 0.331 seconds
	Maximum number of seconds to run all queries: 0.804 seconds
	Number of clients running queries: 1024
	Average number of queries per client: 1
 
```

 Here are the details for both tests with the report.

storage/innobase/fil/fil0fil.cc
```diff
@@ -821,6 +821,23 @@ class Fil_shard {
     mutex_release();
   }
 
+  void clear_ibts() {
+    mutex_acquire();
+
+    for (auto it : m_ibt_deleted) {
```
percona-ysorokin: Avoid copying, `Space` seems to be a pretty complicated structure
```suggestion
    for (const auto& it : m_ibt_deleted) {
```

storage/innobase/fil/fil0fil.cc
```diff
@@ -821,6 +821,23 @@ class Fil_shard {
...
+      auto space = it.second;
```
percona-ysorokin, line 828: Likewise
```suggestion
      const auto& space = it.second;
```

storage/innobase/fil/fil0fil.cc
```diff
@@ -1397,6 +1430,18 @@ class Fil_system {
...
   }
 
+  /** Delete the in memory tablespace(fil_space_t*) entries of the IBT
+  tablespaces from the IBT deleted hash. This is done after ensuring all dirty
+  pages of deleted IBTs are removed. Note: we cannot use checkpoint because when
+  considering checkpoint, we dont consider LSN from temporary tablespaces. So
+  the pages of IBT will still remain in flush_list even after checkpoint moves
+  its lwm */
+  void clear_ibts() {
+    for (auto shard : m_shards) {
```
percona-ysorokin: Likewise
```suggestion
    for (const auto& shard : m_shards) {
```

storage/innobase/fil/fil0fil.cc
```diff
@@ -8224,6 +8204,11 @@ dberr_t Fil_shard::do_io(const IORequest &type, bool sync,
     for (auto pair : m_deleted) {
       ut_ad(pair.first != page_id.space());
     }
+
+    /* Should never attempt to read from a IBT deleted tablespace. */
+    for (auto pair : m_ibt_deleted) {
```
percona-ysorokin: Likewise
```suggestion
    for (const auto& pair : m_ibt_deleted) {
```

percona-ysorokin (approved):
LGTM with minor stylistic improvements

storage/innobase/fil/fil0fil.cc
satya-bodapati: this didn't work because fil_space_free() is expecting non const space, it frees object inside space_t* and also space to nullptr

satya-bodapati: It doesn't accept reference "&" too.

Binding value of type 'fil_space_t *const' to reference to type 'fil_space_t *' drops 'const' qualifier  fil0fil.cc:1186:44: note: pass832,11         6%

satya-bodapati: this looks strange to me

satya-bodapati: @percona-ysorokin ok to revert to original?

percona-ysorokin: Yes, @satya-bodapati , here the original version is fine

satya-bodapati: Thanks yura

satya-bodapati:
https://ps80.cd.percona.com/job/percona-server-8.0-param/805/

satya-bodapati:
https://ps80.cd.percona.com/job/percona-server-8.0-param/809/ -> only innodb suites to check if temp_table_encrypt is passing.

PR #4042, opened 2020-09-17
PS-7162 feature: UDF functions for PITR in PXC operator

percona-ysorokin:
https://jira.percona.com/browse/PS-7162


mysql-test/t/percona_binlog_utils_udf.test
```diff
@@ -0,0 +1,312 @@
...
+    SET @saved_gtid_executed = @@global.gtid_executed;
+    START TRANSACTION;
+    --let $i = 1
+    while ($i <= $substage)
+    {
+      INSERT INTO t1 VALUES(DEFAULT);
+      --inc $i
+    }
+    COMMIT;
+    eval INSERT INTO captured_gtid VALUES($stage, $substage, '$binlog_file_name', GTID_SUBTRACT(@@global.gtid_executed, @saved_gtid_executed));
```
inikep, line 60: Can be ignored: I prefer to use `--eval` instead of `eval ... ;` to better distinguish `mysqltest` commands from SQL statements.

inikep (approved):
LGTM

percona-ysorokin:
Jenkins build link
https://ps80.cd.percona.com/job/percona-server-8.0-param/849/

mysql-test/t/percona_binlog_utils_udf.test
percona-ysorokin: @inikep To be honest a also prefer `--command ...` syntax rather than `command ...;`.
However, for executing SQL statements that require `$var` substitutions I usually use the latter because this syntax supports multiline statements (and very often this is the case).
So, this is actually where I bend the rule deliberately to avoid situations where I will be using `--eval ...` for single line and `eval ...;` for multiline statements - just to make everything uniform.

PR #4046, opened 2020-09-21
Fixed PS-7325 - SELECT missing data from MyRocks table when GROUP BY …

altmannmarcelo:
…is used

https://jira.percona.com/browse/PS-7325

Cherry-picked facebook commit[1] and ported to 8.0. Added a MTR
test case.

[1]:
commit 98b0d40af27051ded465a5ca81fac8e3110cc2c1
Author: Manuel Ung <mung@fb.com>
Date:   Tue Mar 19 07:44:54 2019 -0700

    Fix error handling in group by plans

    Summary:
    For certain GROUP BY plans, the QUICK_RANGE_SELECT class is used to iterate through all possible prefixes that form a group. The current iteration process calls read_range_first with both a start and end range, and then iterates until we've moved past the end range. However, the storage engine may choose to honour the end range (because of index condition pushdown, rocksdb iterator bounds, or rocksdb prefix seek mode) and return HA_ERR_KEY_NOT_FOUND or HA_ERR_END_OF_FILE instead. Instead of moving onto the next range, this causes the QUICK_RANGE_SELECT::get_next_prefix to return early without actually advancing onto the next prefix. This can cause an infinite loop in QUICK_GROUP_MIN_MAX_SELECT::get_next, or QUICK_GROUP_MIN_MAX_SELECT::get_next to return HA_ERR_END_OF_FILE too early.

    The fix is to detect if HA_ERR_KEY_NOT_FOUND or HA_ERR_END_OF_FILE is returned, and continue onto the next range instead of returning early.

    Reviewed By: hermanlee

    Differential Revision: D14521084

    fbshipit-source-id: 6963b17081f

### DIFF
--- a/mysql-test/suite/rocksdb/r/select.result
+++ b/mysql-test/suite/rocksdb/r/select.result
@@ -349,4 +349,38 @@ a	b
 200	bar
 200	bar
 200	bar
-DROP TABLE t1, t2;
+CREATE TABLE `t3` (
+`col1` bigint unsigned NOT NULL AUTO_INCREMENT,
+`col2` varbinary(16) NOT NULL,
+`col3` varchar(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL,
+`col4` bigint NOT NULL,
+`col5` mediumblob NOT NULL,
+PRIMARY KEY (`col1`),
+UNIQUE KEY `uc_key` (`col2`,`col3`,`col4`)
+) ENGINE=ROCKSDB;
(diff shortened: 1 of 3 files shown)


altmannmarcelo:
https://ps80.cd.percona.com/job/percona-server-8.0-param/782/ - rocksdb.select was failing
https://ps80.cd.percona.com/job/percona-server-8.0-param/783/ - ran with rocksdb.select test only

inikep (approved):
LGTM

PR #4048, opened 2020-09-21
PS-7291: Allow option value validation to be run for 'set persist_only' (8.0)

oleksandr-kachan:
Currently there is no new value check done in case variable value
is updated via 'set persist_only'. It's written to config 'as is'.
This was done in order to disable checks for readonly variables
which are not applicable at runtime, otherwise it's not possible to
do 'set persist_only' for a readonly variable.
But as a result of this change some invalid value for a variable may
be written to a config file and crash the server after restart.
New value check can be enabled in this case to avoid such issues.
This check will differ a bit for usual sys variables and for
plugin variables.

For usual sys variables there is a basic check and also custom on_check
method provided. We could skip custom on_check only in case of
'set persist_only'. This would allow to do default checks for a value
(min/max value, allowed values) depending on its type and skipping some
specific checks which may be not applicable to runtime check for
readonly variables.

For plugin variables there is a default method which validates value
according to parameters provided at variable definition time. It is
also possible to provide custom validation method in which case it
will replace default one. Here again checks not applicable to readonly
variables at runtime will be done in custom validator. We still
could use default validation method with 'set persist_only' to
do default checks for provided value.

Add separate method which converts a variable value for suitable
string representation to be used with 'set persist_only'.


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/785/

sql/sql_plugin_var.cc
```diff
@@ -588,6 +632,54 @@ void sys_var_pluginvar::saved_value_to_string(THD *, set_var *var,
...
 
+/**
+  Set a PERSIST_ONLY value for a variable. The resulting value should be
+  set to a string representation of the actual value.
+
+  @param[in]     thd   Thread context.
+  @param[in]     var   Plugin variable.
+  @param[in,out] dest  Destination memory pointer.
+*/
+void sys_var_pluginvar::persist_only_update(THD *, set_var *var, char *dest) {
```
kamil-holubicki: For me the method name is somehow confusing. It converts set_var into its string representation. Why not persist_only_to_string() or something?

sql/persisted_variable.cc
```diff
@@ -332,17 +334,19 @@ bool Persisted_variables_cache::set_variable(THD *thd, set_var *setvar) {
...
     String bool_str;
     if (setvar->value) {
-      res = setvar->value->val_str(&str);
+      setvar->var->persist_only_update(thd, setvar, str.ptr());
+      res = &str;
+
       if (system_var->get_var_type() == GET_BOOL) {
-        if (res == nullptr ||
-            check_boolean_value(res->c_ptr_quick(), bool_str)) {
+        if (check_boolean_value(res->c_ptr_quick(), bool_str)) {
```
kamil-holubicki: Wouldn't it be better to move this 'boolean hack' into Sys_var_bool::saved_value_to_string()?

sql/persisted_variable.cc
```diff
@@ -768,13 +792,29 @@ bool Persisted_variables_cache::set_persist_options(bool plugin_options) {
...
+            dynamic_cast<Sys_var_enum *>(sysvar) ||
+            dynamic_cast<Sys_var_multi_enum *>(sysvar)) {
+          res = create_item(thd, iter->value);
+        } else {
+          res = new (thd->mem_root)
+              Item_string(iter->value.c_str(), iter->value.length(),
+                          &my_charset_utf8mb4_bin);
+        }
+        break;
+      }
```
satya-bodapati: is this change done to handle ENUMs? Item_string() doesn't work ENUMs? How did it work before? Please try to add comments to the new code. Don't explain what the code is trying to do, explain what is achieving, or why we need this change?

sql/persisted_variable.cc
```diff
@@ -768,13 +792,29 @@ bool Persisted_variables_cache::set_persist_options(bool plugin_options) {
...
+      case SHOW_BOOL:
+      case SHOW_MY_BOOL:
+        res = create_item(thd, iter->value);
```
satya-bodapati: BOOLs will use Item_int() or Item_string()?

sql/sys_vars.h
```diff
@@ -416,6 +416,10 @@ class Sys_var_enum : public Sys_var_typelib {
   const uchar *global_value_ptr(THD *, LEX_STRING *) {
     return pointer_cast<const uchar *>(typelib.type_names[global_var(ulong)]);
   }
+  void persist_only_update(THD *, set_var *var, char *dest) {
+    strcpy(dest,
```
kamil-holubicki: Multiple occurrences: strcpy does not prevent buffer overflow. use strncpy instead.

sql/persisted_variable.cc
```diff
@@ -332,17 +334,19 @@ bool Persisted_variables_cache::set_variable(THD *thd, set_var *setvar) {
     uint dummy_err;
     String bool_str;
     if (setvar->value) {
-      res = setvar->value->val_str(&str);
+      setvar->var->persist_only_update(thd, setvar, str.ptr());
```
kamil-holubicki: Here we pass str.ptr() to the function where we copy data to this buffer. But inside we do not know about how big the buffer is. Maybe it is better to pass String* instead of char* and construct the string in safe manner?

sql/set_var.cc
```diff
@@ -346,7 +346,7 @@ uchar *sys_var::global_var_ptr() {
 
 bool sys_var::check(THD *thd, set_var *var) {
   if ((var->value && do_check(thd, var)) ||
-      (on_check && on_check(this, thd, var))) {
+      (var->type != OPT_PERSIST_ONLY && on_check && on_check(this, thd, var))) {
```
satya-bodapati, line 350: So if I understand correctly, the do_check() now does do_def_check() for PERSIST_ONLY and we skip the on_check() for PERSIST_ONLY.  Right? [the core change for sys_vars]

sql/set_var.cc
```diff
@@ -1013,7 +1013,7 @@ int set_var::check(THD *thd) {
     my_error(ER_WRONG_TYPE_FOR_VAR, MYF(0), var->name.str);
     return -1;
   }
-  int ret = (type != OPT_PERSIST_ONLY && var->check(thd, this)) ? -1 : 0;
+  int ret = var->check(thd, this) ? -1 : 0;
```
satya-bodapati, line 1032: var->check() calls either the  sys_var::check() or the plugin_var::check() ?

sql/sql_plugin_var.h
```diff
@@ -277,6 +277,16 @@ class sys_var_pluginvar : public sys_var {
...
+
+  /**
+    Set a PERSIST_ONLY value for a variable. The resulting value should be
+    set to a string representation of the actual value.
+
+    @param[in]     thd   Thread context.
+    @param[in]     var   Plugin variable.
+    @param[in,out] dest  Destination memory pointer.
+  */
+  void persist_only_update(THD *, set_var *var, char *dest);
```
satya-bodapati: function name is misleading. I was thinking it is updating somehing. AFAIU, this is about geting string representation of variable value.  may be get_string() or to_string() is sufficient.

sql/persisted_variable.cc
satya-bodapati: There is a lot of code to get string representation. why the current code doesn't work? set_var->value->val_str(&str)?

satya-bodapati (changes requested):
Still trying to understand the PR at code level [core, non-core etc]. Please bear with the questions.

sql/persisted_variable.cc
```diff
@@ -768,13 +792,29 @@ bool Persisted_variables_cache::set_persist_options(bool plugin_options) {
...
             Item_int(iter->value.c_str(), (uint)iter->value.length());
         break;
-      case SHOW_CHAR:
       case SHOW_LEX_STRING:
-      case SHOW_BOOL:
-      case SHOW_MY_BOOL:
         res = new (thd->mem_root) Item_string(
             iter->value.c_str(), iter->value.length(), &my_charset_utf8mb4_bin);
         break;
+      case SHOW_CHAR: {
```
kamil-holubicki: Is my understanding correct that this part is needed for proper handling of wrong values saved previously in mysqld-auto.cnf file?
Having this patch all bools are saved as ON/OFF and enums are saved as their string representation (not corresponding int representation).

If the above is true, do we need this part of the patch? Isn't it enough to have fix for storing part?

kamil-holubicki (changes requested):
I would suggest extending MTR test. We should cover the handling of all touched types vs PERSIST ONLY (including plugins). We should test positive and negative scenarios.

sql/sql_plugin_var.cc
oleksandr-kachan: Renamed to "persist_only_to_string"

sql/sql_plugin_var.h
oleksandr-kachan: Changed it according to Kamil's suggestion - "persist_only_to_string"

sql/persisted_variable.cc
oleksandr-kachan: This was removed according to Kamil's suggestion

sql/persisted_variable.cc
oleksandr-kachan: Same as above

sql/set_var.cc
oleksandr-kachan: In case of PERSIST_ONLY do_check will invoke do_def_check() for plugin vars and will do "default system" check for sys vars. And on_check() will be skipped here.

sql/set_var.cc
oleksandr-kachan: Here sys_var::check() is called and it in turn will decide what to call - sys_var_pluginvar::do_check() or do_check() for corresponding sys var.

sql/persisted_variable.cc
oleksandr-kachan: As I understand it value->val_str(&str) returns original value. When check method is invoked it stores resulting (and possibly changed) value in a different place (var->save_result) and it should be handled differently.

sql/persisted_variable.cc
oleksandr-kachan: Not sure about this. The Sys_var_bool::saved_value_to_string() is supposed to work with default values and this one is for a bit different scenario.

sql/sys_vars.h
oleksandr-kachan: This part was changed to use String.set()

sql/persisted_variable.cc
oleksandr-kachan: Changed to pass a String instead of *char.

sql/persisted_variable.cc
oleksandr-kachan: While working on this change I noticed that some incorrect values could still stay in config file. that is why I left this part. But for a resulting change you are probably right, there is not much sense in this code. Removed it.

oleksandr-kachan:
> I would suggest extending MTR test. We should cover the handling of all touched types vs PERSIST ONLY (including plugins). We should test positive and negative scenarios.

We already have some test for persisted variables. Some were updated in this change. Will take a look if there is anything to do additionally.

sql/persisted_variable.cc
satya-bodapati: Ok, You didn't answer the question :) but I understood from result file. So enums were written as numbers in persist file. With your change, it is written as string.

Is it possible to seprate out this  enum/bool "value" to string change?

sql/set_var.cc
satya-bodapati: Ok got it

sql/sql_plugin_var.h
satya-bodapati: Ok, looks better..

sql/persisted_variable.cc
satya-bodapati: Understood, so this about getting the string representation. Can you please extract out as separate PR/change?

sql/persisted_variable.cc
oleksandr-kachan: Sorry, I thought that is not actual as code was removed... This conversion is done in sys_var_pluginvar::persist_only_to_string(). In general it should be possible to continue writing ENUMs as numbers in one change and do a separate change which will change persist_only_to_string() behavior for ENUMs. This also will affect some tests. I can try to separate that but what benefit will that give us?

satya-bodapati: Ok to keep it for now.. since you already wrote it and tested.

1. limit the scope to actual bug will help reviewer and then if we revert we end up reverting only the actual code 
2. Easier to test
3. Lesser scope == less chance for errors and the fix could be approved faster.

mysql-test/t/persisted_variables.test
```diff
@@ -126,7 +126,7 @@ SET PERSIST sort_buffer_size=256000;
 SET PERSIST max_heap_table_size=999424, slave_net_timeout=124;
 SET PERSIST long_query_time= 2.999999;
 SET PERSIST sql_require_primary_key= true;
-SET PERSIST default_table_encryption= ONLINE_TO_KEYRING;
+SET PERSIST default_table_encryption= ON;
```
satya-bodapati: Is it possible to set this variable to ON without keyring plugin?

Also any reason why ONLINE_TO_KEYRING is changed to ON?

sql/set_var.h
```diff
@@ -308,6 +309,10 @@ class sys_var {
 
   void save_default(THD *thd, set_var *var) { global_save_default(thd, var); }
 
+  virtual void persist_only_to_string(THD *thd, set_var *var, String *dest) {
+    saved_value_to_string(thd, var, dest->c_ptr());
```
satya-bodapati: Don't we have the string buffer issue? Why not pass dest to save_value_to_string()

sql/sys_vars.h
```diff
@@ -831,8 +851,10 @@ class Sys_var_charptr : public sys_var {
     var->save_result.string_value.length = ptr ? strlen(ptr) : 0;
   }
   void saved_value_to_string(THD *, set_var *var, char *def_val) {
-    memcpy(def_val, var->save_result.string_value.str,
-           var->save_result.string_value.length);
+    if (var->save_result.string_value.str) {
```
satya-bodapati: please change it != nullptr
and in which case this would be called with nullptr?

satya-bodapati (changes requested):
looks good mostly

mysql-test/t/persisted_variables.test
oleksandr-kachan: This appears to be a separate issue. I somehow missed this test while checking changes for PS-7340 (Improve value validation for default_table_encryption variable) and just fixed it here together with other variable related tests. 
Keyring plugin is required only when default_table_encryption is set to ONLINE_TO_KEYRING or ONLINE_FROM_KEYRING_TO_UNENCRYPTED. Changed it here to ON as there is actually no intention to check validation for this variable in this test. Only need to make sure that persisting works.
I should probably fix this test in separate ticket?

sql/sys_vars.h
oleksandr-kachan: In case of empty string.

mysql-test/t/persisted_variables.test
satya-bodapati: ok, changing to ON is fine. I believe you handled this in another PR. So no need for another PR.

sql/persisted_variable.cc
kamil-holubicki: Now I see that check_boolean_value() is probably not needed at all, because for bool will always return ON or OFF, so check_boolean_value() will always return true.

What different scenario are you referring to?

oleksandr-kachan: > What different scenario are you referring to?

I meant that saved_value_to_string() is supposed to set default value for a variable and here we set value provided by user.
As for this check_boolean_value()  - most probably it isn't needed now, I just wanted to minimize code changes. Do you think we should remove it at all?

satya-bodapati (approved):

sql/persisted_variable.cc
kamil-holubicki: I think if it is not needed, we should remove it. What is your opinion @satya-bodapati ?

satya-bodapati: Existing changes are anyway above minimal :) , OK to remove

oleksandr-kachan: > I think if it is not needed, we should remove it.

Done. Still use it for default value case. It would require additional code changes to remove it.

oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/850/

kamil-holubicki (approved):

PR #4055, opened 2020-09-28
PS-7338 : Assertion failure : os0file.cc : 1767 : key_version_read_fr…

satya-bodapati:
…om_page != ENCRYPTION_KEY_VERSION_NOT_ENCRYPTED

Problem:
--------
With tablespaces that are created/altered with explicit "ENCRYPTION=N" and a
server crash makes server unrecoverable. The assertion says key version of
keyring encrypted tables should be unencrypted.

Even though all tables are MasterKey(MK) encrypted tables, server wrongly
assumes the tables as KEYRING encrypted tables. Based on presence of
crypt_data parsed from redo, Server thinks all crypt_datas are from KEYRING
encrypted tables.

It missed the fact, ENCRYPTION="N" table/tablespaces will have crypt_data
but of type UNENCRYPTED scheme.

Fix:
----
If the destination tablespace is encrypted and crypt_data found is
unencrypted : Do not set crypt_data to fil_space_t (tablespace is MK
encrypted)

If the destination tablespace is unencrypted and crypt_data found
is unencrypted: Just set crypt_data to fil_space_t (tablespace is created
with explicit ENCRYPTION=N)

Othercases when crypt_data schem is ENCRYPTED,
mark the tablespace as encrypted, set crypt_data to fil_space_t and to type
KEYRING.


mysql-test/suite/innodb/t/crypt_data_redo.test
```diff
@@ -0,0 +1,57 @@
...
+while ($i <= 10) {
+--let $new_checksum = query_get_value(CHECKSUM TABLE tt_$i, Checksum, 1)
+--let $old_checksum = query_get_value(SELECT a FROM checksum WHERE id = $i, a, 1)
+--let $assert_text = Table checksum must match after crash recovery
+--let $assert_cond = $old_checksum = $new_checksum
+--source include/assert.inc
+--inc $i
+}
+
+--echo # test cleanup
```
oleksandr-kachan, line 33: Most of the data being cleaned up here comes from 'inc' file. Would it be better to move those data cleanup into a separate 'inc' file as well?

oleksandr-kachan (approved):
LGTM, the comment related to test data cleanup is more like an idea and it's up to you if we need to change anything there.

mysql-test/suite/innodb/t/crypt_data_redo.test
satya-bodapati: Yeah. frankly, inc file is not necessary. I just merged all .inc's by mohit

satya-bodapati:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-pipeline/23185/

satya-bodapati:
New Jenkins link: https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-pipeline/23218/

(Fixed a mistake: should use crypt_data->type instead of crypt_data->should_encrypt())

PR #4057, opened 2020-09-28
PS-7289: Fix min/max value check for innodb_encryption_threads (8.0)

oleksandr-kachan:
The min and max variable value provided with variable definition
don't take effect in case there is a custom validator for a variable.
Added min and max allowed value check into custom validator
for innodb_encryption_threads variable.

(cherry picked from commit dc73cfb8f04825d539851c00cf850d06aa10c987)

### DIFF
--- a/mysql-test/suite/sys_vars/r/innodb_encryption_threads_basic.result
+++ b/mysql-test/suite/sys_vars/r/innodb_encryption_threads_basic.result
@@ -1,3 +1,52 @@
+SET GLOBAL innodb_file_per_table = ON;
+INSTALL PLUGIN keyring_file SONAME 'keyring_file.so';
+SET GLOBAL innodb_encryption_threads = 10;
+SELECT @@global.innodb_encryption_threads;
+@@global.innodb_encryption_threads
+10
+SET GLOBAL innodb_encryption_threads = 1000;
+Warnings:
+Warning	1292	Truncated incorrect innodb_encryption_threads value: '1000'
+SELECT @@global.innodb_encryption_threads;
+@@global.innodb_encryption_threads
+255
+SET GLOBAL innodb_encryption_threads = -1000;
+Warnings:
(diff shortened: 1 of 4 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/786/

satya-bodapati (changes requested):
Why cherrypick? is it because 8.0.21 release branch not created? Is it possible to regular merge from 5.7 now?

oleksandr-kachan:
> Why cherrypick?

The change to 5.7 was done on top of the first commit done in scope of this task. So it is ahead of last GCA point now. Not sure if it is Ok to do merge to 8.0 in this case (at least there will be some additional conflicts). Should I still do a merge in this case?

george-lorch:
> > Why cherrypick?
> 
> The change to 5.7 was done on top of the first commit done in scope of this task. So it is ahead of last GCA point now. Not sure if it is Ok to do merge to 8.0 in this case (at least there will be some additional conflicts). Should I still do a merge in this case?

Yes, as discussed, since the 5.7 commit for this change is based upon a previous 5.7 change that occurs _after_ the 5.7 GCA point, it is needed that this commit be based on the 5.7 HEAD and cherry-picked forward to 8.0

satya-bodapati:
> > > Why cherrypick?
> > 
> > 
> > The change to 5.7 was done on top of the first commit done in scope of this task. So it is ahead of last GCA point now. Not sure if it is Ok to do merge to 8.0 in this case (at least there will be some additional conflicts). Should I still do a merge in this case?
> 
> Yes, as discussed, since the 5.7 commit for this change is based upon a previous 5.7 change that occurs _after_ the 5.7 GCA point, it is needed that this commit be based on the 5.7 HEAD and cherry-picked forward to 8.0

Thanks for the explanation @georgelorchpercona

satya-bodapati (approved):

PR #4058, opened 2020-09-28
PS-7351: CA_PATH is ignored by the ldap plugin

dutow:
Issue: self signed certificates can't be verified by the simple ldap
plugin. Unless the LDAPTLS_REQCERT environment variable is set to
never, TLS/SSL can't be used with self signed certificates

This happened because the ldap plugin set the TLS options after
initializing the ldap connection. At that point, TLS already failed.

Fix: TLS parameters are set before initializing the ldap connection.


percona-ysorokin (approved):
LGTM

PR #4060, opened 2020-09-29
PS-7340: Improve value validation for default_table_encryption variable (8.0)

oleksandr-kachan:
Problem:
--------
It's possible to set default_table_encryption to ONLINE_TO_KEYRING or
ONLINE_FROM_KEYRING_TO_UNENCRYPTED in case when keyring plugin isn't
loaded. This leads to server crash when trying to encrypt/decrypt data.
    
Fix:
--------
Add additional validation for default_table_encryption variable.
It can be set to ONLINE_TO_KEYRING or ONLINE_FROM_KEYRING_TO_UNENCRYPTED
only in case keyring plugin is loaded.

### DIFF
--- a/mysql-test/suite/sys_vars/r/default_table_encryption_basic.result
+++ b/mysql-test/suite/sys_vars/r/default_table_encryption_basic.result
@@ -118,13 +118,15 @@ SELECT @@session.default_table_encryption;
 @@session.default_table_encryption
 OFF
 SET SESSION default_table_encryption=ONLINE_TO_KEYRING;
+ERROR HY000: The default_table_encryption option cannot be changed, keyring plugin is not available
 SELECT @@session.default_table_encryption;
 @@session.default_table_encryption
-ONLINE_TO_KEYRING
+OFF
 SET SESSION default_table_encryption=ONLINE_FROM_KEYRING_TO_UNENCRYPTED;
+ERROR HY000: The default_table_encryption option cannot be changed, keyring plugin is not available
 SELECT @@session.default_table_encryption;
 @@session.default_table_encryption
-ONLINE_FROM_KEYRING_TO_UNENCRYPTED
+OFF
 SET GLOBAL default_table_encryption=OFF;
 SELECT @@global.default_table_encryption;
 @@global.default_table_encryption
@@ -139,13 +141,15 @@ SELECT @@global.default_table_encryption;
 @@global.default_table_encryption
 ON
 SET PERSIST default_table_encryption=ONLINE_TO_KEYRING;
+ERROR HY000: The default_table_encryption option cannot be changed, keyring plugin is not available
 SELECT @@global.default_table_encryption;
 @@global.default_table_encryption
-ONLINE_TO_KEYRING
+ON
 SET PERSIST default_table_encryption=ONLINE_FROM_KEYRING_TO_UNENCRYPTED;
+ERROR HY000: The default_table_encryption option cannot be changed, keyring plugin is not available
 SELECT @@global.default_table_encryption;
 @@global.default_table_encryption
-ONLINE_FROM_KEYRING_TO_UNENCRYPTED
+ON
 SET PERSIST default_table_encryption=ON;
 SELECT @@global.default_table_encryption;
 @@global.default_table_encryption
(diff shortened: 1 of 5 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/787/

mysql-test/suite/sys_vars/r/default_table_encryption_basic.result
```diff
@@ -139,13 +141,15 @@ SELECT @@global.default_table_encryption;
 @@global.default_table_encryption
 ON
 SET PERSIST default_table_encryption=ONLINE_TO_KEYRING;
+ERROR 42000: Variable 'default_table_encryption' can't be set to the value of 'ONLINE_TO_KEYRING'
```
satya-bodapati: why it doesn't print that keyring plugin is not loaded? Specially this message
```
          thd, Sql_condition::SL_WARNING, ER_WRONG_ARGUMENTS,
                        " (innodb_undo_log_encrypt) and try again.");
          "The default_table_encryption option cannot be changed, "
    return true;
          "keyring plugin is not available");
```

satya-bodapati (changes requested):

mysql-test/suite/sys_vars/r/default_table_encryption_basic.result
oleksandr-kachan: This is done similar to other checks in that method now. The message goes to warnings and can be seen with 'show warnings'. I'll take a look how we could change error message in this case.

satya-bodapati: Please check. "cannot be set" is not user friendly. We should suggest an action that the user can execute. If you cannot reuse this existing message, add "Please check keyring plugin is loaded" (or something on that lines). Also, add "SHOW WARNINGS".

oleksandr-kachan: > "cannot be set"

This one comes from standard error messages

> ER_WRONG_VALUE_FOR_VAR 42000
>        eng "Variable '%-.64s' can't be set to the value of '%-.200s'"

satya-bodapati: Right :disappointed: ! This is for uses case when user sets wrong values.. (like out of the boundary, string). Doesn't need an explanation. 

Maybe use different ER_ code for ONLINE_KEYRING mode failures?

oleksandr-kachan: Updated it a bit, now it prints "keyring plugin is not available" error.

mysql-test/suite/sys_vars/r/default_table_encryption_basic.result
```diff
@@ -182,13 +186,15 @@ SELECT @@global.default_table_encryption;
...
 @@global.default_table_encryption
-ONLINE_TO_KEYRING
+OFF
 SET PERSIST default_table_encryption=ONLINE_FROM_KEYRING_TO_UNENCRYPTED;
+ERROR HY000: The default_table_encryption option cannot be changed, keyring plugin is not available
 SELECT @@global.default_table_encryption;
 @@global.default_table_encryption
-ONLINE_FROM_KEYRING_TO_UNENCRYPTED
+OFF
 SET PERSIST default_table_encryption=ON;
```
satya-bodapati, line 198: can you add a restart with keyring plugin and ensure that we are able to change these modes?

satya-bodapati: I mean a new test case scenario

satya-bodapati (changes requested):

mysql-test/suite/sys_vars/r/default_table_encryption_basic.result
oleksandr-kachan: We already have a separate test which checks this with keyring plugin enabled - encryption.default_table_encryption_var. Or would you like to test something in addition?

satya-bodapati: Ok, that testcase sets all modes. Should be sufficient.

satya-bodapati (approved):
Looks good
@oleksandr-kachan Please update the description and commit message to have the Problem: and Fix: sections

PR #4066, opened 2020-10-01
PS-7352, PS-7353: Fix LDAP issues on reconfiguration

dutow:
Issues:
* Specifying a non existent server with to the LDAP variables could
  result in a slow timeout, making the SET command slow to return,
  blocking connections in the meantime.
* Specifying a non existent CA file for TLS/SSL when there are already
  active connections will result in a crash in the ldap library

Fix: make the variables that requires reconnecting a new server static,
so no runtime reconfiguration is required. These parameters only have to
be changed when the LDAP server is moved, which is not a common event.t


percona-ysorokin (approved):
LGTM
Just a few typos in the commit message
s/reconnecting a new server/reconnecting to a new server/
s/event.t/event/

PR #4070, opened 2020-10-05
PS-7360: clang-4.0 compilation fails from '-Winconsistent-missing-destructor-override'

inikep:
Don't call `MY_COMPILER_CLANG_DIAGNOSTIC_IGNORE("-Winconsistent-missing-destructor-override")` for clang-4.0 or older.


percona-ysorokin (approved):
LGTM

PR #4073, opened 2020-10-05
5.6 - Fixed PS-7346 - Huge memory usage for audit plugin when large queries…

altmannmarcelo:
… are executed

query, we escape it.
outlen to buflen_estimated  and return NULL. Then audit_log_notify will

Issue:
1) buflen_estimated was doubling the size. This is no longer required
non escaped utf8mb4 query.
2) When estimating the size of the non escaped utf8mb4 query, it was

### DIFF
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -617,6 +617,14 @@ char *audit_log_general_record(char *buf, size_t buflen,
   query_length= my_charset_utf8mb4_general_ci.mbmaxlen *
                 event->general_query_length;
 
+  /* Note: query_length is the maximun size using utf8m4(4 bytes) that
+   * event->general_query_length may use. In the if branch, we convert it to
+   * utf8mb4. We store the recalculated (real size) length to query variable
+   * and use the remaing of the buffer for the output that will be printed to
+   * audit log. Parameter char *buf must be big enough to store
+   * the query (using utf8mb4) + the full output of audit event, which will
+   * contain the query again. At the else branch we estime this size.
+   */
   if (query_length < (size_t) (endbuf - endptr))
   {
     uint errors;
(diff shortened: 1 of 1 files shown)


altmannmarcelo:
https://ps56.cd.percona.com/job/percona-server-5.6-param/130/

altmannmarcelo:
5.6 is a bit difficult to have a test case without memory instrumentation on PS. Where is the example of 5.7 https://github.com/altmannmarcelo/percona-server/commit/0db8c9eff5820c8f8a4950111f9e1f05fa7d9ef2 before and after the patch (audit code is exactly the same for 5.6 and 5.7):

Before
```
*************************** 1. row ***************************
       event_name: memory/sql/THD::variables
    current_count: 4
    current_alloc: 400.00 MiB
current_avg_alloc: 100.00 MiB
       high_count: 6
       high_alloc: 400.00 MiB
   high_avg_alloc: 66.67 MiB
```

After:
```
*************************** 1. row ***************************
       event_name: memory/sql/THD::variables
    current_count: 4
    current_alloc: 80.00 MiB
current_avg_alloc: 20.00 MiB
       high_count: 6
       high_alloc: 80.00 MiB
   high_avg_alloc: 13.33 MiB
```

percona-ysorokin (approved):
LGTM

percona-ysorokin:
@altmannmarcelo Please, also double check with ASan and Valgrind. No need to run full set of suites. Just try to do the same measurements you did for regular Debug build under these checkers.

altmannmarcelo:
@percona-ysorokin 
Ran all tests related to audit_log (audit_log_buffer_size_overflow audit_log_charset audit_log_csv audit_log_default_db audit_log_filter_commands audit_log_filter_users audit_log_install_bug1435606 audit_log_install audit_log_json audit_log_long_records audit_log_many_connections audit_log_new audit_log_old audit_log_rotate audit_log_startup audit_log_syslog audit_log audit_log_threadpool) :

https://ps56.cd.percona.com/job/percona-server-5.6-asan-param/11/ - ASAN ok
https://ps56.cd.percona.com/job/percona-server-5.6-asan-param/12/ - Valgrind doesn't have any definitely lost

percona-ysorokin:
Good to go then

PR #4076, opened 2020-10-07
5.7 - Fixed PS-7346 - Huge memory usage for audit plugin when large queries…

altmannmarcelo:
… are executed

query, we escape it.
outlen to buflen_estimated and return NULL. Then audit_log_notify will

Issue:

buflen_estimated was doubling the size. This is no longer required
non escaped utf8mb4 query.
When estimating the size of the non escaped utf8mb4 query, it was

### DIFF
--- a/plugin/audit_log/tests/mtr/audit_log_memory_consumption.test
+++ b/plugin/audit_log/tests/mtr/audit_log_memory_consumption.test
@@ -1,0 +1,23 @@
+--source include/not_embedded.inc
+--source include/have_perfschema.inc
+
+--let $MYSQLD_DATADIR= `SELECT @@datadir`
+--let $log_file=$MYSQLD_DATADIR/test_audit.log
+
+--remove_file $log_file
+SET GLOBAL audit_log_flush=ON;
+--disable_query_log
+--disable_result_log
+UPDATE performance_schema.setup_instruments
+SET ENABLED = 'YES'
+WHERE NAME LIKE 'memory/%';
+CREATE TABLE test (col1 LONGBLOB NOT NULL);
+# 10M query
+--let $very_long_text=`SELECT REPEAT('a', 10485760)`
+eval INSERT INTO test (col1) VALUES ('$very_long_text');
+--enable_query_log
+--enable_result_log
+--let $assert_text = Memory usage is less than 100M
+--let $assert_cond= "[SELECT current_alloc FROM sys.memory_global_by_current_bytes WHERE event_name = \'memory/sql/THD::variables\']" <= 100
+--source include/assert.inc
+DROP TABLE test;
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -617,6 +617,14 @@ char *audit_log_general_record(char *buf, size_t buflen,
   query_length= my_charset_utf8mb4_general_ci.mbmaxlen *
                 event->general_query_length;
 
+  /* Note: query_length is the maximun size using utf8m4(4 bytes) that
+   * event->general_query_length may use. In the if branch, we convert it to
+   * utf8mb4. We store the recalculated (real size) length to query variable
+   * and use the remaing of the buffer for the output that will be printed to
+   * audit log. Parameter char *buf must be big enough to store
+   * the query (using utf8mb4) + the full output of audit event, which will
+   * contain the query again. At the else branch we estime this size.
+   */
   if (query_length < (size_t) (endbuf - endptr))
   {
     uint errors;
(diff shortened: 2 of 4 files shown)


altmannmarcelo:
https://ps57.cd.percona.com/job/percona-server-5.7-param/343/ - full run
https://ps57.cd.percona.com/job/percona-server-5.7-asan-param/25/ - ASAN
https://ps57.cd.percona.com/job/percona-server-5.7-asan-param/26/ - Valgrind

plugin/audit_log/tests/mtr/audit_log_memory_consumption.test
```diff
@@ -0,0 +1,25 @@
...
+# 10M query
+let $very_long_text=`SELECT REPEAT('a', 10485760)`;
+eval INSERT INTO test (col1) VALUES ('$very_long_text');
+--enable_query_log
+--enable_result_log
+--let $assert_text = Memory usage is less than 100M
+--let $query_result = query_get_value(SELECT current_alloc FROM sys.memory_global_by_current_bytes WHERE event_name = 'memory/sql/THD::variables', current_alloc, 1)
+--let $assert_cond= "$query_result" <= 100
+--source include/assert.inc
+DROP TABLE test;
```
percona-ysorokin, line 23: @altmannmarcelo  Should there be some kind of disabling performance schema instrumentation  after you did
```
UPDATE performance_schema.setup_instruments
SET ENABLED = 'YES'
WHERE NAME LIKE 'memory/%';
```
`mtr-check` may not detect this but it's possible that other tests running on the same mysqld instance will be affected.

plugin/audit_log/tests/mtr/audit_log_memory_consumption.test
```diff
@@ -0,0 +1,25 @@
+--source include/not_embedded.inc
+--source include/have_perfschema.inc
+
+let $MYSQLD_DATADIR= `select @@datadir`;
```
percona-ysorokin:
```suggestion
--let $MYSQLD_DATADIR= `SELECT @@datadir`
```

plugin/audit_log/tests/mtr/audit_log_memory_consumption.test
```diff
@@ -0,0 +1,25 @@
...
+let $log_file=$MYSQLD_DATADIR/test_audit.log;
```
percona-ysorokin:
```suggestion
--let $log_file=$MYSQLD_DATADIR/test_audit.log
```

plugin/audit_log/tests/mtr/audit_log_memory_consumption.test
```diff
@@ -0,0 +1,25 @@
...
+
+--remove_file $log_file
+SET GLOBAL audit_log_flush=ON;
+
+--disable_query_log
+--disable_result_log
+UPDATE performance_schema.setup_instruments
+SET ENABLED = 'YES'
+WHERE NAME LIKE 'memory/%';
+CREATE TABLE test (col1 longblob NOT NULL);
```
percona-ysorokin:
```suggestion
CREATE TABLE test (col1 LONGBLOB NOT NULL);
```

plugin/audit_log/tests/mtr/audit_log_memory_consumption.test
```diff
@@ -0,0 +1,25 @@
...
+# 10M query
+let $very_long_text=`SELECT REPEAT('a', 10485760)`;
```
percona-ysorokin:
```suggestion
--let $very_long_text=`SELECT REPEAT('a', 10485760)`
```

plugin/audit_log/tests/mtr/audit_log_memory_consumption.test
```diff
@@ -0,0 +1,25 @@
...
+eval INSERT INTO test (col1) VALUES ('$very_long_text');
+--enable_query_log
+--enable_result_log
+--let $assert_text = Memory usage is less than 100M
+--let $query_result = query_get_value(SELECT current_alloc FROM sys.memory_global_by_current_bytes WHERE event_name = 'memory/sql/THD::variables', current_alloc, 1)
+--let $assert_cond= "$query_result" <= 100
```
percona-ysorokin:
```suggestion
--let $assert_cond= "[SELECT current_alloc FROM sys.memory_global_by_current_bytes WHERE event_name = \'memory/sql/THD::variables\']" <= 100
```

percona-ysorokin (changes requested):
One comment regarding disabling memory instrumentation
and a few regarding consistency in `--command` vs `command;` style;

plugin/audit_log/tests/mtr/audit_log_memory_consumption.test
altmannmarcelo: @percona-ysorokin I've checked other tests that change setup_instruments and the clean-up is enable everything:
few examples:
https://github.com/percona/percona-server/blob/8.0/mysql-test/suite/sysschema/include/ps_setup_instruments_cleanup.inc
https://github.com/percona/percona-server/blob/8.0/mysql-test/suite/perfschema/include/memory_aggregate_cleanup.inc#L40

percona-ysorokin (approved):
LGTM

PR #4077, opened 2020-10-07
8.0 - Fixed PS-7346 - Huge memory usage for audit plugin when large queries…

altmannmarcelo:
… are executed

query, we escape it.
outlen to buflen_estimated and return NULL. Then audit_log_notify will

Issue:

buflen_estimated was doubling the size. This is no longer required
non escaped utf8mb4 query.
When estimating the size of the non escaped utf8mb4 query, it was

### DIFF
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -617,6 +617,14 @@ char *audit_log_general_record(char *buf, size_t buflen,
   query_length= my_charset_utf8mb4_general_ci.mbmaxlen *
                 event->general_query_length;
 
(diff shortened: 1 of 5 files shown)


percona-ysorokin (approved):
LGTM

altmannmarcelo:
Test results are ok:
https://ps80.cd.percona.com/job/percona-server-8.0-param/796/ - full
https://ps80.cd.percona.com/job/percona-server-8.0-asan-param/5/ - ASAN
https://ps80.cd.percona.com/job/percona-server-8.0-asan-param/6/ - Valgrind

PR #4078, opened 2020-10-07
PS-7363: Release MDL when UNDO truncate is cancelled because of LTFB

dutow:
Issue: executing drop undo tablespace while LTFB + UNLOCK TABLES is
executed on another session could result in DROP UNDO TABLESPACE getting
stuck.

This is caused by the undo truncation thread: when undo truncation is
cancelled because of LTFB, the purge thread doesn't release the MDL
locks held for truncation.

Fix: locks are correctly released on failure.


satya-bodapati (approved):
Looks safe

PR #4095, opened 2020-10-21
PS-7379:  Add macOS to Azure Pipelines (8.0)

inikep:


azure-pipelines.yml
```diff
@@ -279,45 +329,56 @@ jobs:
         -DENABLE_DOWNLOADS=1
         -DDOWNLOAD_BOOST=1
         -DWITH_BOOST=$(BOOST_DIR)
-        -DMYSQL_MAINTAINER_MODE=ON
-        -DCMAKE_C_COMPILER_LAUNCHER=/usr/bin/ccache
```
percona-ysorokin, line 283: @inikep Not sure why exactly `ccache` is invoked after removing these lines.

inikep: I solved the issue with Linux vs macOS using `-DCMAKE_C_COMPILER_LAUNCHER=$(which ccache)`.

percona-ysorokin (approved):
LGTM

PR #4110, opened 2020-10-28
PS-7388: CircleCI: avoid Docker Hub rate limits (8.0)

inikep:
On November 1st, 2020 Docker Hub will begin limiting anonymous image pulls.

This change will influence our clang-format validation with CircleCI.
But a simple workaround is to create a dummy Docker Hub account and use it within `.circleci/config.yml`.


percona-ysorokin (approved):
LGTM

PR #4111, opened 2020-10-28
PS-7394 Select correct VERSION file

EvgeniyPatlan:


inikep:
We have the same issue with 5.7 e.g. https://ps57.cd.percona.com/job/percona-server-5.7-param/348/

build-ps/version.sh
```diff
@@ -3,7 +3,14 @@ DIR=$(dirname $(realpath $0))
 BRANCH=$(git branch --show-current)
 IFS='.' read -r MAJOR MINOR PATCH <<< $(echo $BRANCH | awk -F'-' '{print $2}')
 EXTRA=$(echo $BRANCH | awk -F'-' '{print $3}')
-source ${DIR}/../VERSION
+if [ -f VERSION ]; then
```
vorsel: Shouldn't we check ${DIR}/../VERSION instead of VERSION ?

build-ps/version.sh
```diff
@@ -3,7 +3,14 @@ DIR=$(dirname $(realpath $0))
...
+    source ${DIR}/../VERSION
+elif [ -f MYSQL_VERSION ]; then
+    source ${DIR}/../MYSQL_VERSION
```
vorsel, line 9: The same as above.

build-ps/version.sh
EvgeniyPatlan: fixed

build-ps/version.sh
EvgeniyPatlan: fixed

vorsel (approved):

PR #4126, opened 2020-11-11
PS-7419 update document titles

patrickbirch:
Update the current document titles to "Percona Server for MySQL" instead of "Percona Server"


nastena1606 (approved):
LGTM

PR #4142, opened 2020-12-03
PS-7162 hotfix: UDF functions for PITR in PXC operator (custom --log-bin)

percona-ysorokin:
https://jira.percona.com/browse/PS-7162

Fixed problem with Binary Log UDFs not working properly when MySQL Server is
started with non-default '--log-bin' option.

Functions accepting binlog name now expect it to be in short form, meaning
that they will be read from the current binary log directory. If a name with
the path separator '/' is provided, an error is returned.

Functions returning binlog name now always return a name in short form
(without path component). If necessary, users can always prepend it with
'@@global.log_bin_basename'.

'main.percona_binlog_utils_udf' MTR test case updated to reflect the changes
 mentioned above. It was also extended with a few additional checks for
calling functions on non-existing binary logs.

### DIFF
--- a/mysql-test/r/percona_binlog_utils_udf.result
+++ b/mysql-test/r/percona_binlog_utils_udf.result
@@ -31,7 +31,7 @@ ERROR HY000: Can't initialize function 'get_last_gtid_from_binlog'; Function req
 SELECT get_last_gtid_from_binlog('blah', 'blah');
 ERROR HY000: Can't initialize function 'get_last_gtid_from_binlog'; Function requires exactly one argument
 SELECT get_last_gtid_from_binlog('non_existing.000001');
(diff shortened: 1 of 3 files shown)


percona-ysorokin:
Jenkins build link (single test `main.percona_binlog_utils_udf`)
https://ps80.cd.percona.com/job/percona-server-8.0-param/853/ (normal)
https://ps80.cd.percona.com/job/percona-server-8.0-param/854/ (--mysqld=--log-bin=/tmp/relocated_binlog)

venkatesh-prasad-v (approved):
Hi Yura! 

Thank you for working on this. Everything is working as expected. So, it is approved from my side.

inikep (approved):
LGTM

PR #4144, opened 2020-12-07
PS-7458 Fix format error and update Yum installation (8.0)

patrickbirch:
modified:   source/installation/yum_repo.rst

### DIFF
--- a/doc/source/installation/yum_repo.rst
+++ b/doc/source/installation/yum_repo.rst
@@ -30,15 +30,15 @@ Each of the |Percona Server| RPM packages have a particular purpose.
    * - Package
      - Contains
    * - percona-server-server
-     - The server itself (the ``mysqld`` binary)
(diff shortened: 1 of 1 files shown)


doc/source/installation/yum_repo.rst
```diff
@@ -85,7 +92,7 @@ You can install Percona yum repository by running the following commands as a ``
    <tokudb_intro>` and :ref:`MyRocks <myrocks_intro>` storage engines. These
    storage engines are installed as plugins. You can find more information on how
    to install and enable the |TokuDB| storage in the :ref:`tokudb_installation`
-   guide. More information about how to install |MyRocks| can be found in the
```
nastena1606, line 88: What if to make this call to action in an active voice? 
"Find how to install |MyRocks| in ..."

nastena1606 (approved):

PR #4147, opened 2020-12-09
PS-7049: AppArmor and SELinux configuration

kamil-holubicki:
https://jira.percona.com/browse/PS-7049

1. Removed check of SELinux mode from ps-admin script. SELinux does not
need to be disabled for TokuDB and LD_PRELOAD. However, SELinux will
prevent accessing 'tokudb_backup_dir'. This needs SELinux policy
adjustment or manual correction of the given directory's SELinux tag
2. Removed unused PS SELinux policy files
3. AppArmor policy included in percona-server deb package

Used upstream MySql AppArmor profile, but modified to work in 'complain'
mode by default instead of 'enforce' mode.


build-ps/debian/percona-server-server.postinst
```diff
@@ -116,9 +116,30 @@ case "$1" in
...
+		if [ "${CNF_ACTION}" = "Use NEW my.cnf" ]; then
+			update-alternatives --force --install /etc/mysql/my.cnf my.cnf "/etc/mysql/mysql.cnf" 200
+		fi
+
+		PROFILE_ACTION="Use NEW AppArmor profile"
+		# If the existing AppArmor module/local profile is the proper file, we back it up
+		if [ -f "/etc/apparmor.d/usr.sbin.mysqld" -o -f "/etc/apparmor.d/local/usr.sbin.mysqld" ]; then
+			cp /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/usr.sbin.mysqld.bak 2> /dev/null || true
+			cp /etc/apparmor.d/local/usr.sbin.mysqld /etc/apparmor.d/local/usr.sbin.mysqld.bak 2> /dev/null || true
+			db_input high percona-server-server/existing_profile_file || true
```
EvgeniyPatlan: as we discussed for pxc - it is better to create a backup if the $RET is "Use NEW AppArmor profile"

EvgeniyPatlan (changes requested):

build-ps/debian/percona-server-server.postinst
kamil-holubicki: Fixed. I forgot that we have the same part here :)

EvgeniyPatlan (approved):

PR #4149, opened 2020-12-11
fix ret value in `bool Bloom_filter::test_and_set(ulong key)`

rsy56640:


dutow:
Hello

Could you elaborate the issue?

Looking at the code, it seems to me that the return value of `test_and_set` is correct, and it's what its only caller, `access_distinct_page` expects.

The doc comment about the return value could be confusing, but the function itself works as it should.

dutow:
Hello!

I'm closing this for now, if you still have an issue, please comment.

PR #4154, opened 2020-12-15
PS-7451 Document SELinux and AppArmor (8.0)

patrickbirch:
modified:   source/index.rst
new file:   source/security/selinux.rst
new file:   source/security/apparmor.rst

### DIFF
--- a/doc/source/security/apparmor.rst
+++ b/doc/source/security/apparmor.rst
@@ -1,0 +1,405 @@
+.. _enable-apparmor:
+
+=================================================
+Working with Percona Server with AppArmor 
+=================================================
+
+The operating system has a Discretionary Access Controls (DAC) system. AppArmor supplements the DAC with a Mandatory Access Control (MAC) system. AppArmor is the default security module for Ubuntu or Debian systems and uses profiles to define how programs access resources. 
+
+AppArmor is path-based and restricts processes by using profiles. Each profile contains a set of policy rules. Some applications may install their profile along with the application. If an installation does not also install a profile, then that application is not part of the AppArmor subsystem. You can also create profiles since they are simple text files stored in the ``/etc/apparmor.d`` directory. 
+
+Profiles are one of the following modes:
+
+* Enforce - the default setting, applications are prevented from taking actions restricted by the profile rules.
+
+* Complain - applications are allowed to take restricted actions, and the actions are logged.
+
+* Disabled - Applications are allowed to take restricted actions, and the actions are not logged. 
+
+ You can mix enforce profiles and complain profiles. 
+
+Checking the Current Status
+-----------------------------
+
+As root or using ``sudo``, you can check the AppArmor status:
+
+.. sourcecode:: bash
+
+    $ sudo aa-status
+    apparmor module is loaded.
+    34 profiles are loaded.
+    32 profiles in enforce mode.
+    ...
+        /usr/sbin/mysqld
+    ...
+    2 profiles in complain mode.
+    ...
+    3 profiles have profiles defined.
+    ...
+    0 processes are in complain mode.
+    0 processes are unconfined but have a profile defined.
+
+.. _complain-one:
+
+Change a Profile to Complain mode
+-----------------------------------------
+
+To improve a profile, change to the ``complain`` mode and then run any actions that generates log entries. Update the profile from the log information.
+
+To change a profile to complain mode and the program is in your path:
+
+.. sourcecode:: bash
+
+    $ sudo aa-complain <program>
+
+If needed, specify the path:
+
+.. sourcecode:: bash
+
+    $ sudo aa-complain /sbin/<program>
+
+If the profile is not in stored in /etc/apparmor.d, use the following command:
+
+.. sourcecode:: bash
+
+    $ sudo aa-complain /path/to/profiles/<program>
+
+.. _enforce-one:
+
+Change a Profile to Enforce mode
+---------------------------------
+
+To change a profile to enforce mode and the program is in your path:
+
+.. sourcecode:: bash
+
+    $ sudo aa-enforce <program>
+
+If needed, specify the path:
+
+.. sourcecode:: bash
+
+    $ sudo aa-enforce /sbin/<program>
+
+To change enforce mode and the profile is not stored in /etc/apparmor, use the following command:
+
+.. sourcecode:: bash
+
+    $ sudo aa-enforce /path/to/profile
+
+.. _disable-one:
+
+Disabling one profile
+------------------------------
+
+You can disable a profile but change the profile to the ``complain`` mode if you are troubleshooting. Use either of the following methods:
+
+.. sourcecode:: bash
+
+    $ sudo ln -s /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/disable/ 
+    $ sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld
+
+or
+
+.. sourcecode:: bash
+
+    $ aa-disable /etc/apparmor.d/usr.sbin.mysqld
+
+Reloading all profiles
+-----------------------
+
+Run the following command to reload all profiles:
+
+.. sourcecode:: bash
+
+    $ sudo service apparmor reload
+
+.. _reload-one:
+
+Reloading one profile
+----------------------
+
+To reload one profile, run the following:
+
+.. sourcecode:: bash
+
+    $ sudo apparmor_parser -r /etc/apparmor.d/<profile>
+
+For some changes to take effect, you may need to restart the program.
+
+Disabling AppArmor
+--------------------
+
+AppArmor should not be disabled, if possible. If you have a profile with issues, change the profile to ``complain`` mode during troubleshooting. If AppArmor must be disabled, run the following commands:
+
+1. Check the status.
+
+.. sourcecode:: bash
+
+    $ sudo apparmor_status
+
+2. Stop and disable AppArmor.
+
+.. sourcecode:: bash
+
+    $ sudo systemctl stop apparmor
+    $ sudo systemctl disable apparmor
+
+Modifying the mysqld profile
+---------------------------------
+
+Install the mysqld profile with the following procedure: 
+
+1. Install the ``apparmor-utils`` package to work with profiles:
+
+    .. sourcecode:: bash
+
+        $ sudo apt-get -y install apparmor-utils
+        Reading package lists... Done
+        Building dependency tree
+        ...
+        The following additional packages will be installed:
+            python3-apparmor python3-libapparmor
+        ...
+
+2. Download the current version of the AppArmor from upstream:
+
+    ..  sourcecode:: bash
+
+        $ wget https://raw.githubusercontent.com/mysql/mysql-server/8.0/packaging/deb-in/extra/apparmor-profile
+        ...
+        Saving to 'apparamor-profile`
+        ...
+
+3. Move the file to `/etc/apparmor.d/usr.sbin.mysqld`
+
+    .. sourcecode:: bash
+
+        $ mv apparmor-profile /etc/apparmor.d/usr.sbin.mysqld
+
+4. Create an empty file for editing:
+
+    .. sourcecode:: bash
+
+        $ touch /etc/apparmor.d/local/usr.sbin.mysqld
+
+5. Load the profile:
+
+    .. sourcecode:: bash
+
+        $ apparmor_parser -r -T -W /etc/apparmor.d/usr.sbin.mysqld
+
+6. Restart |Percona Server|:
+
+    .. sourcecode:: bash
+
+        $ systemctl restart mysql
+
+7. Check the profile status:
+
+    .. sourcecode:: bash
+
+        $ sudo aa-status
+        ...
+        processes are in enforce mode
+        ...
+        /usr/sbin/mysqld (100840)
+        ...
+
+Editing the mysqld profile
+---------------------------
+
+Only edit :file:`/etc/apparmor.d/local/usr.sbin.mysql`. Follow the steps to :ref:`complain-one`. Open the file in any text editor to make the changes. Follow the steps in :ref:`reload-one` to enable the profile. When you are satisfied with the update, follow the steps to :ref:`enforce-one`.
+
+Configuring a custom data directory
+------------------------------------
+
+You can change the data directory to a non-default location, like `/var/lib/mysqlcustom`. You should enable audit mode and edit the profile to allow access.
+
+.. sourcecode:: bash
+
+    $ cat /etc/mysql/mysql.conf.d/mysqld.cnf 
+    #
+    # The Percona Server 8.0 configuration file.
+    #
+    # For explanations see
+    # http://dev.mysql.com/doc/mysql/en/server-system-variables.html
+
+    [mysqld]
+    pid-file    = /var/run/mysqld/mysqld.pid
+    socket        = /var/run/mysqld/mysqld.sock
+    *datadir    = /var/lib/mysqlcustom*
+    log-error    = /var/log/mysql/error.log
+
+Enable audit mode for mysqld. The security policy is enforced and all access is logged.
+
+.. sourcecode:: bash
+
+    $ aa-audit mysqld
+
+Restart Percona Server for MySQL.
+
+.. sourcecode:: bash
+
+    $ systemctl mysql restart
+
+The restart fails because AppArmor blocked access to the custom data directory.  
+
+.. sourcecode:: bash
+
+    ...
+    Dec 07 12:17:08 ubuntu-s-4vcpu-8gb-nyc1-01-aa-ps audit[16013]: AVC apparmor="DENIED" operation="mknod" profile="/usr/sbin/mysqld" name="/var/lib/mysqlcustom/binlog.index" pid=16013 comm="mysqld" requested_mask="c" denied_mask="c" fsuid=111 ouid=111
+    Dec 07 12:17:08 ubuntu-s-4vcpu-8gb-nyc1-01-aa-ps kernel: audit: type=1400 audit(1607343428.022:36): apparmor="DENIED" operation="mknod" profile="/usr/sbin/mysqld" name="/var/lib/mysqlcustom/mysqld_tmp_file_case_insensitive_test.lower-test" pid=16013 comm="mysqld" requested_mask="c" denied_mask="c" fsuid=111 ouid=111
+    ...
+
+In :file:`/etc/apparmor.d/local/usr.sbin.mysqld`, change the following entries in this section in your text editor:
+
+..  sourcecode:: text
+
+        # Allow data dir access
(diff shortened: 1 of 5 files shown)


doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+
+Security-Enhanced Linux (SELinux) is a Linux kernel security module. This module provides a way to support access control policies, including mandatory access controls (MAC). SELinux defines how confined processes interact with files, network ports, directories, other processes, and other parts of the server. 
+
+Each process and resource is assigned an SELinux label, which provides a security context. Policy rules determine how a process interacts with another process or with a resource. 
+
+SELinux has confined processes which restricts everything unless it is explicitly allowed. An unconfined process is not controlled by SELinux. 
+
+SELinux has three modes. The following list defines the modes:
+
+* Disabled - No SELinux policies loaded, which disables policies
```
kamil-holubicki: The idea is described in a proper way however, I think it would be good to stick to the naming used in SELinux world:

SELinux 'kernel security module' has only 1 policy loaded (e.g. targeted). This policy consists of 'policy modules' (aka modules). Modules contain rules.
So something like /etc/selinux/usr.sbin.mysqld is the policy module containing rules which describe what is allowed for mysqld.
This policy module is loaded into SELinux and merged together with another loaded policy modules into one SELinux policy.

So below if we say 'policy' in fact it is 'SELinux policy module'
If we say 'Targeted' or 'Strict' policy, we talk about SELinux policy.

However, I agree that sometimes it is easier/clearer to talk about SELinux policy, not the policy module like in the sentence:
The default policy allows mysqld to write to var_log_t tagged files
If we wanted to be strict here, he would write:
Rules in the default SELinux policy module for mysqld allow mysqld to write to var_log_t tagged files.

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+List SELinux Types associated with files
+----------------------------------------
+
+The SELinux label contains the context; such as user, role, and type. For Type Enforcement, by policy, interactions between processes and resources use SELinux types. SELinux type names often end with a ``_t``. 
+
+Add the ``-Z`` switch to many of the utilities to view the SELinux contexts. For example, to view the SELinux types associated with directories and files, run the following command:
+
+.. code-block:: bash
+
+    $ ls -1aZ /var/lib/
```
kamil-holubicki: ls -laZ

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+    -rw-r-----.  1 mysql mysql system_u:object_r:var_lib_t:s0   12582912 Dec  1 14:29  ibdata1
+    ...
+
+To solve the issue, you have two methods:
+
+* Set the proper labels for ``mysqlcustom`` files
+
+* Change the mysqld SELinux policy to allow mysqld access to ``var_lib_t`` files.
+
+The recommended solution is to set the labels. The following procedure assumes you have already created and set ownership to the custom data directory displays the process:
```
kamil-holubicki: Is this sentence OK? (2nd one). I'm not sure about the last part 'displays the process', but maybe my English is not good enough ;)

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+
+.. code-block:: text
+
+    log-error=/logs/mysqld.log
+
+Verify the logs with the following command:
+
+.. code-block:: bash
+
+    $ ls -1aZ
```
kamil-holubicki: ls -laZ /
1 -> l + trailing slash

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+            *type default_t*;
+            class file getattr;
+            *class dir write*;
+        }
+
+        #============= mysqld_t ==============
+        *allow mysqld_t default_t:dir write*;
+        allow mysqld_t var_lib_t:file getattr;
+
+    The policy contains rules for the custom data directory and the custom logs directory. We have set the proper labels for the data directory, and applying this autogenerated policy would loosen our hardening by allowing mysqld to access ``var_lib_t`` tags.
```
kamil-holubicki, line 403: I would say that the generated policy module **may** include rules not related to what we are doing right now (log location), because it may contain some rules generated from our previous experiments (in this cases I have played with data directory and SELinux generated errors now converted into rules). 
So it is important to review this file, detect what is related to what we want to achieve, remove everything else.
That's why we use --start parameter below. Let' say that we played with data directory at 9:35:00 which caused SELinux journal logs generation.
Now it's 10:12:00 and we play with custom logs location. So we want to check the journal only from the time point when we started with this (let's say 10:00:00), as earlier events/errors/logs are not relevant.

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+
+        $ journalctl -xe
+
+7. Stop the service:
+
+    .. code-block:: bash
+
+        $ service mysqld stop 
+
+8. Generate a local mysqld policy:
```
kamil-holubicki: In place of 10:00:00 use the time got in point 4.

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+.. note::
+
+    You can use this procedure to adjust the local mysqld policy module. You should review the changes generated to ensure the rules are not too permissive.
+
+.. _selinux-secure-file-priv:
+
+Setting ``secure_file_priv`` directory
+---------------------------------------
+
+Depending on the |Percona Server| version, you may be required to set the SELinux tags on the ``/var/lib/mysql-files/`` directory. This directory is used for ``SELECT ... INTO OUTFILE`` or similar operations.
```
kamil-holubicki: I would add the info that mysqld simply needs r/w access to the destination directory and this has to be allowed by SELinux rules.

BTW, where from we have got this info about "depending on the ..."?

kamil-holubicki (commented):
My main concern is about the terminology used when we talk about policy, policy module, module. I'm not sure if we should follow the strict terminology, or we can assume that the user has some experience and will figure it out from the context.

doc/source/security/selinux.rst
patrickbirch: Interrupted thought.  Fixed.

doc/source/security/apparmor.rst
```diff
@@ -0,0 +1,398 @@
...
+
+=================================================
+Running Percona Server with AppArmor 
+=================================================
+
+AppArmor supplements the operating system's Discretionary Access Controls (DAC) as a Mandatory Access Control (MAC) system. AppArmor is the default security module for Ubuntu or Debian systems and uses profiles to define how programs access resources. An AppArmor profile contains a set of policy rules for specific applications. AppArmor profiles are simple text files. 
+
+Profiles are either of the following modes:
+
+* Enforcement - Rule violations are logged and not allowed.
```
kamil-holubicki: this mode is 'enforce'. Which in fact means that the profile is enforcED.
But in AA world:
```
>aa-enforce mysqld
Setting /usr/sbin/mysqld to enforce mode.

>aa-status
....
1 processes are in enforce mode.
   /usr/sbin/mysqld (14953) 
....

```

doc/source/security/apparmor.rst
```diff
@@ -0,0 +1,398 @@
...
+
+    $ sudo aa-complain <program>
+
+If needed, specify the path:
+
+.. sourcecode:: bash
+
+    $ sudo aa-complain /sbin/<program>
+
+If the program is not in stored in /etc/apparmor.d, use the following command:
```
kamil-holubicki: program -> profile
It looks like it is OK in 'Change a Profile to Enforce mode' chapter below.

doc/source/security/apparmor.rst
```diff
@@ -0,0 +1,398 @@
...
+.. sourcecode:: bash
+
+    $ sudo aa-complain /path/to/profiles/<program>
+
+.. _enforce-one:
+
+Change a Profile to Enforce mode
+---------------------------------
+
+The default status of a profile is enabled. Violations are logged and not permitted.
```
kamil-holubicki: enabled or enforce?

doc/source/security/apparmor.rst
```diff
@@ -0,0 +1,398 @@
...
+
+Disabling one profile
+------------------------------
+
+You can disable one profile, if needed. It is recommended to change the profile to complain if you are troubleshooting.
+
+.. sourcecode:: bash
+
+    $ sudo ln -s /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/disable/ 
+    $ sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld
```
kamil-holubicki, line 100: this will remove the profile from AA. To disable, we should use aa-disable (module will be still loaded, but disabled). Probably the effect will be the same, but...

doc/source/security/apparmor.rst
```diff
@@ -0,0 +1,398 @@
...
+
+        $ systemctl restart mysql
+
+7. Check the profile status:
+
+    .. sourcecode:: bash
+
+        $ sudo aa-status
+        ...
+        processes are in enforce mode
```
kamil-holubicki, line 204: We said above that v8.0.23 installs AA profile in complain mode. This is because I have added 'complain' flag to the upstream profile.
As we see here, when we get upstream profile it installs in enforce mode. I think it is worth mentioning it here like:
"Please note that aa profile after installation is in enforce mode. If you want to switch it to complain (as default for PS 8.0.23) use the following command: aa-complain /etc/..."

doc/source/security/apparmor.rst
```diff
@@ -0,0 +1,398 @@
...
+
+In :file:`/etc/apparmor.d/local/usr.sbin.mysqld`, change the following entries in this section in your text editor:
+
+..  sourcecode:: text
+
+    # Allow data dir access
+    /var/lib/mysqlcustom/ r,
+    /var/lib/mysqlcustom/** rwk,
+
+In :file:`etc/apparmor.d/usr.sbin.mysqld`, comment out, using the `#` symbol, the entries in the `Allow data dir access` section.
```
kamil-holubicki: Maybe it is worth mentioning that this step is optional? Without doing this, mysqld will simply still have access to the default data directory location, but this is probably not what we want.

doc/source/security/apparmor.rst
```diff
@@ -0,0 +1,398 @@
...
+    pid-file    = /var/run/mysqld/mysqld.pid
+    socket        = /var/run/mysqld/mysqld.sock
+    datadir    = /var/lib/mysql
+    log-error    = /*custom-log-dir*/mysql/error.log
+
+Verify the directory exists.
+
+.. sourcecode:: bash 
+
+    $ ls -1a /custom-log-dir/
```
kamil-holubicki: -1a -> la

doc/source/security/apparmor.rst
```diff
@@ -0,0 +1,398 @@
...
+Restart Percona Server.
+
+..  sourcecode:: bash
+
+    $ service mysql start
+    Job for mysql.service failed because the control process exited with error code.
+    See "systemctl status mysql.service" and "journalctl -xe" for details.
+
+
+    root@ubuntu-s-4vcpu-8gb-nyc1-01-aa-ps:~#
```
kamil-holubicki: root@ubuntu-s-4vcpu-8gb-nyc1-01-aa-ps:~# journalctl -xe

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+    SELinux root directory:         /etc/selinux
+    Loaded policy name:             targeted
+    Current mode:                   enforcing
+    Mode from config file:          enforcing
+    Policy MLS status:              enabled
+    Policy deny_unknown status:     allowed
+    Memory protection checking:     actual (secure)
+    Max kernel policy version:      31
+
+To see a view of SELinux on various services, add the ``-b`` option to ``sestatus``. The header information is the same but each policy is specified with either an ``on`` or an ``off``.
```
kamil-holubicki: but each policy module?

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+    $ semanage permissive -d mysqld_t
+
+.. _disabling-selinux::
+
+Disabling SELinux
+-----------------
+
+If you plan to use the enforcing mode at another time, do not disable SELinux, instead use the Permissive mode. While SELinux is disabled, the system may contain mislabeled objects, or objects with no label. The ``autorelabel=1`` option forces a system relabel. Reboot the system in Permissive mode to allow the ``autorelabel`` process to complete before changing the mode to enforcing. 
+
+To disable SELinux, on boot, set the ``selinux=0`` option. With this option, the kernel does not load the SELinux infrastructure.
```
kamil-holubicki: set the ``selinux=0`` kernel option

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+To disable SELinux, on boot, set the ``selinux=0`` option. With this option, the kernel does not load the SELinux infrastructure. 
+
+Edit the ``/etc/selinux/config`` file and change ``SELINUX=diabled`` and then reboot the system. 
+
+.. _list-selinux-types:
+
+List SELinux Types associated with files
+----------------------------------------
+
+The SELinux label contains the context; such as user, role, and type. For Type Enforcement, by policy, interactions between processes and resources use SELinux types. SELinux type names often end with a ``_t``.
```
kamil-holubicki: SELinux label is the context. These 2 terms are used interchangeably. Context consists of user/role/type parts. SELinux uses only type part.

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+    --rw-r---- .    1 mysql     mysql   system_u:object_r:mysqld_db_t:s0
+    1317 Dec 1 13:57 /var/log/mysqld.log
+    -rwxr-xr-x.     1 root      root    system_u:object_r:mysqld_exec_t:s0
+    Oct 7 22:11 /usr/sbin/mysqld
+    -rw-r--r--.     1 root      root    system_u:object_r:mysqld_exec_t:s0
+    1286 Oct 7 22:08 /etc/my.cnf
+
+.. _selinux-starting-mysql:
+
+Starting MySQL in SELinux
```
kamil-holubicki: I'm not sure if we should include this chapter in the official doc. It describes the current implementation of the startup script with its problems.

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,507 @@
...
+        require {
+            type mysqld_t;
+            type default_t;
+            class dir write;
+        }
+
+        #============= mysqld_t ==============
+        allow mysqld_t default_t:dir write;
+
+    This policy allows mysqld writing into the directories tagged with the default ``_t``.
```
kamil-holubicki: \`\`default_t\`\` ?

kamil-holubicki (changes requested):

doc/source/security/selinux.rst
patrickbirch: Fixed.

doc/source/security/selinux.rst
patrickbirch: Fixed.

doc/source/security/selinux.rst
patrickbirch: Fixed.

doc/source/security/selinux.rst
patrickbirch: The -1aZ and the -laZ results were exactly the same.
Changed to -laZ

doc/source/security/selinux.rst
patrickbirch: If the process has been updated with our scripts, I'll remove this section.

doc/source/security/selinux.rst
patrickbirch: Fixed.

doc/source/security/selinux.rst
patrickbirch: Fixed.

doc/source/security/selinux.rst
patrickbirch: Rewritten

doc/source/security/selinux.rst
patrickbirch: Fixed

doc/source/security/selinux.rst
patrickbirch: Rewritten

doc/source/security/apparmor.rst
patrickbirch: Fixed.

doc/source/security/apparmor.rst
patrickbirch: fixed

doc/source/security/apparmor.rst
patrickbirch: Fixed.

doc/source/security/apparmor.rst
patrickbirch: In conversation

doc/source/security/apparmor.rst
patrickbirch: Added note

doc/source/security/apparmor.rst
patrickbirch: Added.

doc/source/security/apparmor.rst
patrickbirch: Fixed.

doc/source/security/apparmor.rst
patrickbirch: Fixed.

doc/source/security/selinux.rst
```diff
@@ -0,0 +1,578 @@
...
+.. note::
+
+    You can use this procedure to adjust the local mysqld policy module. You should review the changes generated to ensure the rules are not too permissive.
+
+.. _selinux-secure-file-priv:
+
+Setting ``secure_file_priv`` directory
+---------------------------------------
+
+|Percona Server| version, you may be required to set the SELinux tags on the ``/var/lib/mysql-files/`` directory. This directory is used for ``SELECT ... INTO OUTFILE`` or similar operations. The server only needs read/write access to the destination directory and these operations must be allowed by SELinux rules.
```
kamil-holubicki: ```|Percona Server| version, you may be required to set the SELinux tags on the ``/var/lib/mysql-files/`` directory.```
This sentence is OK?

kamil-holubicki (approved):

doc/source/security/apparmor.rst
patrickbirch: Fixed.

doc/source/security/selinux.rst
patrickbirch: re-written

doc/source/security/selinux.rst
patrickbirch: Removed.

doc/source/security/selinux.rst
patrickbirch: Fixed.

PR #4160, opened 2021-01-04
PS-1956: Adjust datatype for some timers used for extended slow query log (5.6)

oleksandr-kachan:
The ulong and ulint are used for such timers as innodb_io_reads_wait_timer,
innodb_lock_que_wait_timer and innodb_innodb_que_wait_timer. They are
32-bit at some platforms which may be not enough in some cases.
Adjust corresponding datatypes to 64-bit.

### DIFF
--- a/include/mysql/plugin.h
+++ b/include/mysql/plugin.h
@@ -587,9 +587,9 @@ void increment_thd_innodb_stats(MYSQL_THD thd,
                     unsigned long long trx_id,
                     long io_reads,
                     long long io_read,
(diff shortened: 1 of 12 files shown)


oleksandr-kachan:
https://ps56.cd.percona.com/job/percona-server-5.6-param/135/

satya-bodapati (changes requested):
How do we know this change fixed the bug? how did you verify?

satya-bodapati (approved):

PR #4165, opened 2021-01-12
PS-7493: clang-format checks stopped working on TravisCI and CircleCI (5.6)

inikep:
Use local `/usr/bin/clang-format-diff-5.0` instead of external `clang-format-diff.py`.


inikep:
Tested at https://travis-ci.org/github/inikep/percona-server/builds/754105761

oleksandr-kachan (approved):
Just to recheck... Do we need similar change for other branches as well? I see there is "wget https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/clang-format-diff.py" for 5.7 for example.

inikep:
Yes, this is GCA for 5.6. I will do the same for 5.7 and 8.0.

PR #4166, opened 2021-01-12
PS-7492: Update slow log formatting for tmp tables related stats (8.0)

oleksandr-kachan:
Depending on slow log config it is possible that stats for tmp
tables will be printed on a separate line without "#" symbol
in front of it.
Update log format so that tmp tables related stats will allways be
printed on a separate line.


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/870/

altmannmarcelo (approved):
LGTM

PR #4170, opened 2021-01-13
PS-7492: Update slow log formatting for tmp tables related stats (5.7)

oleksandr-kachan:
Update log format so that tmp tables related stats will be
printed on a separate line.


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/383/

altmannmarcelo (approved):
LGTM

PR #4171, opened 2021-01-13
PS-7290: Using RocksDB as a temp table can lead to an assertion on the debug build (8.0)

inikep:
MySQL 8.0.21 moves a check for `HA_LEX_CREATE_TMP_TABLE` from `check_engine()` to `get_viability()`.
This patch adds a similar check for MyRocks in `ha_rocksdb::create`.


george-lorch (approved):

inikep:
Tested at https://ps80.cd.percona.com/job/percona-server-8.0-param/867

PR #4174, opened 2021-01-14
PS-7298: Fix assertion in Fil_shard::close_file during fast shutdown

oleksandr-kachan:
The assertion ut_a(file->n_pending_flushes == 0) in Fil_shard::close_file
failes in case fast server shutdown is done with INNODB_FAST_SHUTDOWN=2.
The reason for this is that IO threads are still active at the time
all files are being closed. Those threads may start checking pending
flushes at any time.
To fix this wait for any active flushes on a file to finish before
actually closing it.


satya-bodapati (commented):
Looks good to me. Can you check if the same test fails on upstream?

oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/871/

oleksandr-kachan:
> Can you check if the same test fails on upstream?

I wasn't able to reproduce same issue on upstream 8.0 code.

PR #4177, opened 2021-01-15
PS-7495 Block Tablespace DDL with Lock Tables for backup

rahulmalik87:
PS-7495 Block Tablespace DDL with Lock Table For Backup

    Block CREATE/ALTER/DROP TABLESPACE statements when the backup lock
    is taken, and hold the lock until the operation is completed.

### DIFF
--- a/mysql-test/suite/innodb/r/tablespace_encrypt_1.result
+++ b/mysql-test/suite/innodb/r/tablespace_encrypt_1.result
@@ -47,6 +47,41 @@ SELECT @@innodb_file_per_table;
 1
 ALTER TABLESPACE innodb_system ENCRYPTION='Y';
 ERROR 42000: InnoDB: `innodb_system` is a reserved tablespace name.
+LOCK TABLES FOR BACKUP;
+CREATE TABLESPACE tbs01 ADD DATAFILE 'tbs01.ibd' ENCRYPTION='y';
+ERROR HY000: Can't execute the query because you have a conflicting backup lock
+SET @@session.lock_wait_timeout=1;
+CREATE TABLESPACE tbs01 ADD DATAFILE 'tbs01.ibd' ENCRYPTION='y';
+ERROR HY000: Lock wait timeout exceeded; try restarting transaction
+UNLOCK TABLES;
+CREATE TABLESPACE tbs01 ADD DATAFILE 'tbs01.ibd' ENCRYPTION='y';
+CREATE TABLE t1(i int) TABLESPACE=tbs01 ENCRYPTION='y';
+LOCK TABLES FOR BACKUP;
+ALTER TABLESPACE tbs01 ENCRYPTION='n';
+ERROR HY000: Can't execute the query because you have a conflicting backup lock
+INSERT into t1 values(100);
+SET @@session.lock_wait_timeout=1;
+ALTER TABLESPACE tbs01 ENCRYPTION='n';
+ERROR HY000: Lock wait timeout exceeded; try restarting transaction
+UNLOCK TABLES;
+ALTER TABLESPACE tbs01 ENCRYPTION='n';
+LOCK TABLES FOR BACKUP;
+ALTER TABLESPACE tbs01 rename to tbs01_renamed;
+ERROR HY000: Can't execute the query because you have a conflicting backup lock
+SET @@session.lock_wait_timeout=1;
+ALTER TABLESPACE tbs01 rename to tbs01_renamed;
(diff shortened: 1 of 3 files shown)


sql/sql_tablespace.cc
```diff
@@ -453,6 +453,13 @@ bool Sql_cmd_create_tablespace::execute(THD *thd) {
     return true;
   }
 
+  // Acquire Percona's LOCK TABLES FOR BACKUP lock
+  if (thd->backup_tables_lock.abort_if_acquired() ||
+      thd->backup_tables_lock.acquire_protection(
+          thd, MDL_TRANSACTION, thd->variables.lock_wait_timeout)) {
+    return true;
+  }
+
```
satya-bodapati: Where is upstream (LIFB) taken? It is safe to take at the same place. Also, the order is important. Either LIFB first and then LTFB or LTFB first and then LIFB. Please check this order in other places (like undo truncation for example)

satya-bodapati (changes requested):

sql/sql_tablespace.cc
rahulmalik87: It is taken at [line](https://github.com/percona/percona-server/blob/3d545ea8f8e72d3c59b5342788ca98170a6b09b2/sql/sql_tablespace.cc#L246) inside  lock_tablespace_names .
But undo it done [at](https://github.com/percona/percona-server/blob/3d545ea8f8e72d3c59b5342788ca98170a6b09b2/sql/sql_tablespace.cc#L1399)

for undo and my patch the order is sam, we  first we try to take [LTFB](https://github.com/percona/percona-server/blob/3d545ea8f8e72d3c59b5342788ca98170a6b09b2/sql/sql_tablespace.cc#L1382) and then [LIFB](https://github.com/percona/percona-server/blob/3d545ea8f8e72d3c59b5342788ca98170a6b09b2/sql/sql_tablespace.cc#L1399)

rahulmalik87: should I try moving LTFB also inside lock_tablespace_names?

satya-bodapati: global read lock is FTWRL right?

rahulmalik87: yes.

sql/sql_tablespace.cc
```diff
@@ -247,6 +247,13 @@ bool lock_tablespace_names(THD *thd, Names... names) {
...
```
satya-bodapati, line 256: can you also please add a test that shows LTFB is taken and DMLS on tablespaces happen after LTFB is taken?

satya-bodapati (changes requested):

sql/sql_tablespace.cc
rahulmalik87: done

rahulmalik87:
https://ps80.cd.percona.com/job/percona-server-8.0-param/883/testReport/

sql/sql_tablespace.cc
satya-bodapati: I dont see it

mysql-test/suite/innodb/r/tablespace_encrypt_1.result
```diff
@@ -47,6 +47,41 @@ SELECT @@innodb_file_per_table;
...
+SET @@session.lock_wait_timeout=1;
+CREATE TABLESPACE tbs01 ADD DATAFILE 'tbs01.ibd' ENCRYPTION='y';
+ERROR HY000: Lock wait timeout exceeded; try restarting transaction
+UNLOCK TABLES;
+CREATE TABLESPACE tbs01 ADD DATAFILE 'tbs01.ibd' ENCRYPTION='y';
+CREATE TABLE t1(i int) TABLESPACE=tbs01 ENCRYPTION='y';
+LOCK TABLES FOR BACKUP;
+ALTER TABLESPACE tbs01 ENCRYPTION='n';
+ERROR HY000: Can't execute the query because you have a conflicting backup lock
+INSERT into t1 values(100);
```
rahulmalik87, line 62: Here I am doing insert

sql/sql_tablespace.cc
rahulmalik87: check this in patch INSERT into t1 values

satya-bodapati (changes requested):
Can you please expand LTFB in title and also improve commit message a bit?
I will approve as soon as this change is done

satya-bodapati (approved):
LGTM! Ok to merge if trunk is open for merges (please ask on channel)

PR #4178, opened 2021-01-15
PS-7364 Adaptive network buffers

ldonoso:
https://jira.percona.com/browse/PS-7364

CI:

- [Matrix](https://ps80.cd.percona.com/job/percona-server-8.0-param/888/)
- [PS-7364-8.0 with ASAN](https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/27373/)
- [PS-7364-8.0 with ASAN and `-mysqld=--net_buffer_shrink_interval=2`](https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/27245/)
- [percona/8.0 with ASAN](https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/27244/)

## Net_buffer_length status variable

New `net_buffer_length` status variable shows the size of the buffer of the
current connection. If SHOW GLOBAL is used then cumulative size of buffers of
all connections is shown This variable will help to estimate how big the
network buffers overhead in every particular case.

## Periodic network buffers shrink

When long lived connections are used, network buffers tend to grow towards
`max_allowed_packet` size and never shrink automatically unless the connection
is terminated. For example, occasional big query will eventually increase
buffers of all connections(if connections picked at random from the pool),
while at any moment of time only one such buffer is sufficient. People usually
set `max_allowed_packet_size` to some relaxed value, like 64-128Mb, which,
along with significant number of connections(256-1024), may lead to enormous
memory overhead.

This feature allows periodically shrink network buffers according to the
workload during the specified interval.  Thread's network buffer size is set to
the maximum packet size seen in the last interval.  Thus it allows to "forget"
spikes in the packet size which happened before the last interval.

Similar affect can also be achieved by disconnecting and reconnecting TCP
connections, but that much heavier process(and it reconnects connections with
small buffers as well), while this feature is basically a no-op during the
stable workload.

Shrink check timeout can be configured through global
`net_buffer_shrink_interval` MySQL variable.

| **Command-Line Format** | --net-buffer-shrink-interval=# |                            |
| ---                     | ---                            | ---                        |
| **System Variable**     | **Name**                       | net_buffer_shrink_interval |
|                         | **Variable Scope**             | Global                     |
|                         | **Dynamic Variable**           | Yes                        |
| **Permitted Values**    | **Type**                       | integer (seconds)          |
|                         | **Default**                    | 0 (disabled)               |
|                         | **Min Value**                  | 0                          |
|                         | **Max Value**                  | 31536000                   |

## Rebased commits from [original PR](https://github.com/percona/percona-server/pull/1497/):

- aa7b261ec2a - Fixing nits. Tests
- 3d39948334a - Increase client library minor version
- 60750c8cabb - Net_buffer_length show status variable
- d64ff4de652 - Adaptive network buffers MTR test
- d3e28c11619 - Adaptive network buffers

Additional work:
- Fix test `main.all_persisted_variables` and `main.variables`
- Initialize `net_buffer_shrink_time`
- Use `unsigned long` in `my_net_shrink_buffer`
- Realloc the net buffer after `dispatch_command`
- Remove `libperconaserverclient18.2.*` files
- Assign `net_buffer_length` to 0 after `add_status`
- Use lock when accessing `net_buffer_shrink_time`
- The initial value of `max_interval_packet` should not include headers


sql/sql_parse.cc
```diff
@@ -1334,6 +1334,16 @@ bool do_command(THD *thd) {
   /* Restore read timeout value */
   my_net_set_read_timeout(net, thd->variables.net_read_timeout);
 
+  if (net_buffer_shrink_interval &&
```
kamil-holubicki: Shouldn't we access 'net_buffer_shrink_interval' under LOCK_global_system_variables lock?

sql/sql_parse.cc
```diff
@@ -1334,6 +1334,16 @@ bool do_command(THD *thd) {
...
+          thd->start_utime / 1000000 > thd->net_buffer_shrink_time +
+          net_buffer_shrink_interval)
+  {
+      my_net_shrink_buffer(net, thd->variables.net_buffer_length);
```
kamil-holubicki: This can fail and we do nothing in case of realloc failure.

sql-common/net_serv.cc
```diff
@@ -2042,6 +2048,37 @@ static size_t net_read_packet(NET *net, size_t *complen) {
...
+     * so no shrink needed. We allow 10% variance in workload to reduce number
+     * of reallocs */
+    if(max_interval_packet * 110 / 100 >= net->max_packet)
+        return false;
+
+    /* Buffer cannot be smaller then, default, net_buffer_length + header */
+    if(max_interval_packet < min_buf_size)
+        max_interval_packet = min_buf_size;
+
+    /* In the last interval packets were significantly smaller then max_packet,
```
kamil-holubicki: then -> than (multiple occurrences)

sql-common/net_serv.cc
```diff
@@ -161,6 +161,7 @@ bool my_net_init(NET *net, Vio *vio) {
   net->reading_or_writing = 0;
   net->where_b = net->remain_in_buf = 0;
   net->last_errno = 0;
+  net->max_interval_packet = net->max_packet + NET_HEADER_SIZE + COMP_HEADER_SIZE;
```
kamil-holubicki: Should we record headers size here as well?
In my_net_shrink_buffer() we are comparing net->max_packet (which does not include headers) to net->max_inerval_packet.

sql/system_variables.h
```diff
@@ -470,7 +470,7 @@ struct System_variables {
 
 /**
   Per thread status variables.
-  Must be long/ulong up to last_system_status_var so that
+  Must be ulonglong up to LAST_STATUS_VAR so that
```
kamil-holubicki: Leave trunk version/upstream version of this part

sql/system_variables.h
```diff
@@ -542,8 +544,8 @@ struct System_status_var {
   ulong com_stat[(uint)SQLCOM_END];
 
   /*
-    IMPORTANT! See last_system_status_var definition below. Variables after
-    'last_system_status_var' cannot be handled automatically by add_to_status()
+    IMPORTANT! See LAST_STATUS_VAR definition below. Variables after
```
kamil-holubicki: Leave trunk version/upstream version of this part

include/mysql_com.h
```diff
@@ -900,6 +900,7 @@ typedef struct NET {
     to maintain the server internal instrumentation for the connection.
   */
   void *extension;
+  unsigned long max_interval_packet;
```
kamil-holubicki: Please add description what it is for.

sql/sql_class.cc
```diff
@@ -885,6 +885,8 @@ void THD::set_new_thread_id() {
 void THD::cleanup_connection(void) {
   mysql_mutex_lock(&LOCK_status);
   add_to_status(&global_status_var, &status_var);
+  /* net_buffer_length is not cumulative */
+  global_status_var.net_buffer_length -= status_var.net_buffer_length;
```
kamil-holubicki: It is OK, but I wonder if it would not be much clearer if we set global_status.net_buffer_lentgth = 0 here. In fact, this member should always be zero in global_status.

sql/sql_class.cc
```diff
@@ -1090,6 +1092,8 @@ void THD::release_resources() {
   mysql_mutex_lock(&LOCK_status);
   /* Add thread status to the global totals. */
   add_to_status(&global_status_var, &status_var);
+  /* net_buffer_length is not cumulative */
+  global_status_var.net_buffer_length -= status_var.net_buffer_length;
```
kamil-holubicki: It is OK, but I wonder if it would not be much clearer if we set global_status.net_buffer_lentgth = 0 here. In fact, this member should always be zero in global_status.

sql/sql_show.cc
```diff
@@ -3778,8 +3778,11 @@ class Add_status : public Do_THD_Impl {
 void calc_sum_of_all_status(System_status_var *to) {
   DBUG_TRACE;
   mysql_mutex_assert_owner(&LOCK_status);
-  /* Get global values as base. */
+
+  /* Get global values as base but net_buffer_length. */
   *to = global_status_var;
+  to->net_buffer_length = 0;
+
```
kamil-holubicki: Is it needed? It looks like we do not modify global_status_var.net_buffer_length value anywhere. It should always be zero.

If it is needed, we should do the same in PFS_connection_iterator::visit_global() after capturing global_status_var.net_buffer_lenght. 

How to test this path? It is not the path we go when we execute 'show global status like 'net_buffer_length';'
mysqladmin or other client?

This change also causes change of the client-server protocol (additional field to be sent). Is it backward compatible?

sql/sql_parse.cc
ldonoso: Definitely, good call!

sql-common/net_serv.cc
ldonoso: Very good call @kamil-holubicki

In top of that, it is assigned `pkt_data_len` which doesn't include the length of the headers.

sql/sql_class.cc
ldonoso: I agree with you. I think it is clearer if we set it to 0.

sql/sql_show.cc
ldonoso: I guess it is not necessary, I will remove it. I will add a couple of tasks:

- [x] test this path
- [x] check the backward compatibility

ldonoso: @kamil-holubicki, I do not know if it is possible to test this path. The reason is that `calc_sum_of_all_status` is just used in <https://github.com/ldonoso/percona-server/blob/7b919163899d5aa2ebf1b7703d3393e56899a293/sql/sql_parse.cc#L2126-L2140>

Please, note that just the status variables `long_query_count` and `opened_tables` are used and exposed. There seems there is no way of checking out the value of `net_buffer_length`.

ldonoso: @kamil-holubicki, It'd say it is backward compatible. The status vars are [modelled as a table](https://github.com/ldonoso/percona-server/blob/7b919163899d5aa2ebf1b7703d3393e56899a293/storage/perfschema/table_global_status.h#L84).

From the point of view of the client/server protocol, returning a new row in a table
should not be a breaking change.

kamil-holubicki (approved):
lgtm
@percona-ysorokin I checked the logic and it looks fine, but could you take a brief look please and check if you don't see something?
@ldonoso Please squash commits before merge

sql/sql_parse.cc
ldonoso: @kamil-holubicki , please note the error is already managed in the function call by `my_net_shrink_buffer` named `net_realloc`. I have checkout how `net_realloc` is used in other parts of the application and in those cases there is no other extra error management.

kamil-holubicki (approved):

include/mysql_com.h
percona-ysorokin: Any chance we can put this new field `max_interval_packet` inside a custom structure in `extension` to avoid change in client library interface?

mysql-test/t/net_buffer_shrink_interval.test
```diff
@@ -0,0 +1,107 @@
+--source include/restart_mysqld.inc
```
percona-ysorokin: Is this really necessary?
If yes, there is an alternative of putting `--force-restart` in the `.opt` file.

mysql-test/t/net_buffer_shrink_interval.test
```diff
@@ -0,0 +1,107 @@
...
+--echo See the buffer size of the default connection
+show status like 'net_buffer_length';
```
percona-ysorokin: Here and everywhere, instead of relying on recorded values in `.result` files MTR best practices suggest to extract values for such statements into MTR variables
`--let $original_net_buffer_length= query_get_value(show status like 'net_buffer_length', Value, 1)` and then using `--source include/assert.inc` specify your expectations.

mysql-test/t/net_buffer_shrink_interval.test
```diff
@@ -0,0 +1,107 @@
...
+disconnect con_int2;
+
+--source include/wait_until_count_sessions.inc
+
+DROP TABLE t1;
+
+show global status like 'net_buffer_length';
+
+set global net_buffer_length=default;
+set global net_buffer_shrink_interval=default;
```
percona-ysorokin: Here, instead of restoring to defaults,  best practices suggest to save values of the global system variables being changed at the beginning of the test and restore them to saved values at the end.

sql-common/net_serv.cc
```diff
@@ -2042,6 +2048,30 @@ static size_t net_read_packet(NET *net, size_t *complen) {
...
+  if (net->max_packet <= min_buf_size) return false;
+
+  ulong max_interval_packet = net->max_interval_packet;
+  /*Reset buffer size for next interval */
+  net->max_interval_packet = min_buf_size;
+
+  /* In the last interval, packets were not smaller than 90% of the max_packet,
+   * so no shrink needed. We allow 10% variance in workload to reduce number
+   * of reallocs */
+  if (max_interval_packet * 110 / 100 >= net->max_packet) return false;
```
percona-ysorokin: I would put this 10% into a constant.

percona-ysorokin: Also, if `ulong` happens to be a 32-bit value, this may overflow if `max_interval_packet` is about 40M, which is not an unrealistic value for this parameter.

storage/innobase/include/buf0buf.ic
```diff
@@ -64,17 +64,17 @@ struct buf_chunk_t {
 
   /** Advices the OS that this chunk should not be dumped to a core file.
   Emits a warning to the log if could not succeed.
-  @return true iff succeeded, false if no OS support or failed */
```
percona-ysorokin: Probably, not in the scope of this change

mysql-test/t/net_buffer_shrink_interval.test
ldonoso: Good call, it is not, I will remove it.

mysql-test/t/net_buffer_shrink_interval.test
ldonoso: Good point. I will do

mysql-test/t/net_buffer_shrink_interval.test
ldonoso: Good point, the tests are more readable an explicit. I will do.

sql-common/net_serv.cc
ldonoso: Good point, but please, note that `max_interval_packet` should always be smaller than `net->max_packet`, which is also an `ulong`, and that `net->max_packet` is limited to `net->max_packet_size`.

storage/innobase/include/buf0buf.ic
ldonoso: I see, this was present beforehand, I will remove it.

mysql-test/t/net_buffer_shrink_interval.test
```diff
@@ -0,0 +1,165 @@
...
+--echo See the total usage after connections have been created
+--let $net_buffer_length= query_get_value(show global status like 'net_buffer_length', Value, 1)
+--let $assert_text= net_buffer_length shall be 24576
+--let $assert_cond= $net_buffer_length = 24576
+--source include/assert.inc
+
+#Insert big value(8096 bytes) to increase buffers in both connections
+connection con_int1;
+
+INSERT INTO t1 (data) VALUES ("
```
percona-ysorokin: To avoid ugly-looking long strings you can create an MTR variable
```
--let $long_string = `SELECT REPEAT('EF', <big_number>)`
```
and then use it in the `INSERT` statement
```
eval INSERT INTO t1 (data) VALUES ("$long_string");
```

percona-ysorokin (approved):
LGTM
with minor comment that can be ignored

mysql-test/t/net_buffer_shrink_interval.test
ldonoso: good idea @percona-ysorokin , I will do

PR #4179, opened 2021-01-20
PS-7499 [5.7]: Improve error log when MyRocks fails with rocksdb_validate_tables=1

inikep:
1. Add a table name for `AUTOINC mismatch` error.
2. Add a suggestion how to ignore `rocksdb_validate_tables=1` errors.


george-lorch (approved):
Perfect, thanks!

PR #4184, opened 2021-01-21
PS-5846: MyRocks - Add support for default value expressions (8.0)

inikep:
1. Add support for default value expressions in MyRocks.
2. Add `rocksdb.default_as_expr` MTR test.


inikep:
Tested at https://ps80.cd.percona.com/job/percona-server-8.0-param/881/

george-lorch (approved):
We may also want to send this to Facebook along with the virtual columns.

george-lorch (approved):

PR #4188, opened 2021-01-26
PS-7520: Travis CI: compilation issues with boost with macOS (8.0)

inikep:
1. Use `xcode12` instead of `xcode11`.
2. Use `port` instead od `homebrew` to install `ccache`.


percona-ysorokin (approved):
LGTM

PR #4189, opened 2021-01-28
Typo

nasiralamreeki:


dutow:
This is really a typo, but it has more occurrences than this diff, as the variable is really misspelled in the entire file, and all of its uses in other test files.

Its also present in MySQL, not specific to Percona Server.

Please open a new ticket at https://bugs.mysql.com/ or a pull request for https://github.com/mysql/mysql-server

PR #4199, opened 2021-02-15
PS-7568: Exclude coredumper related files from core files search pattern (5.7)

oleksandr-kachan:
https://jira.percona.com/browse/PS-7568

Don't include coredumper related files into a core files
search pattern to avoid false positives when searching core files
in mysql-test-run.


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/419/

percona-ysorokin (approved):
LGTM

PR #4205, opened 2021-02-19
PS-7507: Make processlist info fetched from different locations to be similar

oleksandr-kachan:
https://jira.percona.com/browse/PS-7507

The processlist data fetched from different sources contains
different fields.

The following changes where done to fix this:
- Add time_ms, rows_sent and rows_examined fields into
  performance_schema.processlist.
- Add time_ms column to 'show processlist' results.
- Make 'show processlist' return same data not depending on
  performance_schema_show_processlist variable value.


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/27242/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/27246/

PR #4214, opened 2021-02-26
PS 7459 - Backport of WL#9387 to PS 5.7

ldonoso:
https://jira.percona.com/browse/PS-7459

This is a null merge as the functionality is already present in 8.0


ldonoso:
The gca process was not followed so this merge is not valid

PR #4215, opened 2021-02-26
PS-4641 innodb_flush_method is not clear enough (5.7)

patrickbirch:
Merged 5.6 version
both modified:   source/scalability/innodb_io


inikep (approved):
LGTM

PR #1496, opened 2017-03-13
Jemalloc memory profiling integration

efirs:
This change provides low overhead way to get memory allocation statistics of the running instance.

One can get allocation summary using performance_schema.malloc_stats_summary table. Example:
```
mysql> select * from performance_schema.malloc_stats_summary;

+------------+------------+------------+------------+------------+-----------+
| ALLOCATED  | ACTIVE     | MAPPED     | RESIDENT   | RETAINED   | METADATA  |
+------------+------------+------------+------------+------------+-----------+
| 4444587952 | 5305044912 | 7284843752 | 5427371072 | 1237219680 | 196446784 |
+------------+------------+------------+------------+------------+-----------+
1 row in set (0.02 sec)
```
Or more detailed statistics by the allocation size and number of requests:
```
mysql> select * from performance_schema.malloc_stats;
+-------+------------+------------+---------+-------------+
| TYPE  | ALLOCATED  | NMALLOC    | NDALLOC | NREQUESTS   |
+-------+------------+------------+---------+-------------+
| small | 1981056488 | 1240672368 |       0 | 12119092401 |
| large |  344781832 |  101634376 |       0 |  2921581156 |
| huge  | 1904418320 |         53 |       0 |          53 |
+-------+------------+------------+---------+-------------+
3 rows in set (0.02 sec)
```
Most detailed view of the state of the memory can be produced by dumping memory profile:
```
mysql> show variables like '%jemalloc%';
+--------------------+-------+
| Variable_name      | Value |
+--------------------+-------+
| jemalloc_detected  | ON    |
| jemalloc_profiling | ON    |
+--------------------+-------+
2 rows in set (0.00 sec)

mysql> flush memory profile;
Query OK, 0 rows affected (0.01 sec)

[root@/ #]ls /tmp/jeprof_mysqld*
/tmp/jeprof_mysqld.1.0.170013202213
[root@/ #]jeprof --dot /usr/sbin/mysqld /tmp/jeprof_mysqld.1.0.170013202213 > /tmp/jeprof1.dot
[root@/ #]dot -Tpng /tmp/jeprof1.dot > /tmp/jeprof1.png
```
Here is an example of allocation graph: https://github.com/jemalloc/jemalloc/wiki/Use-Case%3A-Leak-Checking

Enabling profiling is a three fold process:
1. Jemalloc needs to be built with profiling enabled (./autogen.sh  --enable-prof ...)
2. Jemalloc and profiling needs to be loaded on the instance start:
`
[...]

PR #4223, opened 2021-03-08
PS 7366 - Add jemalloc memory allocation profiling

ldonoso:
Jira: <https://jira.percona.com/browse/PS-7366>
CI: <https://ps80.cd.percona.com/job/percona-server-8.0-param/910/>

**Description:**

Two new performance schema tables are introduced to get memory allocation statistics of the running instance:

```
> use performance_schema;

> select * from malloc_stats_totals;

ALLOCATED       ACTIVE  MAPPED  RESIDENT        RETAINED        METADATA
390977528       405291008       520167424       436813824       189718528       9933744

> select * from malloc_stats;

TYPE    ALLOCATED       NMALLOC NDALLOC NREQUESTS
small   23578872        586156  0       2649417
large   367382528       2218    0       6355
huge    0       0       0       0
```

Most detailed view of the state of the memory can be produced by dumping memory profile by running the new command:

```
> flush memory profile;
Query OK, 0 rows affected (0.01 sec)
```

The memory profile is saved on `/tmp`:

```
$ ls /tmp/jeprof_mysqld*
/tmp/jeprof_mysqld.1.0.170013202213
```

The memory profile can be processed further, for instance, for plotting a graph of memory use:

```
$ jeprof --dot /usr/sbin/mysqld /tmp/jeprof_mysqld.1.0.170013202213 > /tmp/jeprof1.dot
$ dot -Tpng /tmp/jeprof1.dot > /tmp/jeprof1.png
```

or here is an example of allocation graph: <https://github.com/jemalloc/jemalloc/wiki/Use-Case%3A-Leak-Checking>.

Two global variables have been added:

- `jemalloc_detected`. This is true if Jemalloc with PROFILING enabled detected. For this to be true:
  - Jemalloc should be compiled with profiling enabled.
  - The env variable MALLOC_CONF set to `prof:true`.
  - PS have been configure to use Jemalloc (e.i. by using the env variable `LD_PRELOAD`)
- `jemalloc_profiling`. To Enable jemalloc profiling. To be activated, it requires `jemalloc_detected`.

**Commits from [PR](https://github.com/percona/percona-server/pull/1496)**:

- a37caed185b - Integrated Jemalloc memory introspection tool
- 4acd636e259 - MTR: Fix perfschema mysql-test suite for malloc_stats

**Addtional work**:

- Exclude changes on tests `perfschema/r/pfs_upgrade_*.t` as they don't exist in 8.0.
- Remove check `include/not_embedded` as 8.0 does not allow embedded server.
- Rename table `table_malloc_stats_summary` to `table_malloc_stats_totals` because the test suite truncates tables of the perfschema whose name contains `summary` or `history` and this table is not truncable.

**TODO**:

- When merging upstream add new tables to list on `pds_dd_version`.h  and check/fix `dd_version_check.test`. This can not be done in the current branch as mysql is updating `PFS_DD_VERSION` in 8.0.23.


sql/sql_profile.cc
```diff
@@ -162,6 +163,63 @@ int make_profile_table_for_show(THD *thd, ST_SCHEMA_TABLE *schema_table) {
...
+typedef int (*je_mallctl_func)(const char *name, void *oldp, size_t *oldlenp,
+                               void *newp, size_t newlen);
+
+bool opt_jemalloc_profiling_enabled = false;
+bool opt_jemalloc_detected = false;
+static je_mallctl_func mallctl_p = NULL;
+static unsigned jemalloc_initialized = 0;
+static unsigned jemalloc_profile_counter = 0;
+
+int jemalloc_mallctl(const char *name, void *oldp, size_t *oldlenp, void *newp,
```
kamil-holubicki, line 192: Document params and return value

sql/sql_profile.cc
```diff
@@ -162,6 +163,63 @@ int make_profile_table_for_show(THD *thd, ST_SCHEMA_TABLE *schema_table) {
...
+    mallctl_p = (je_mallctl_func)dlsym(RTLD_DEFAULT, "mallctl");
+    jemalloc_initialized = 1;
+  }
+
+  if (mallctl_p) return mallctl_p(name, oldp, oldlenp, newp, newlen);
+
+  return 1;
+}
+
+int jemalloc_profiling_dump() {
```
kamil-holubicki, line 213: missing documentation

sql/sql_profile.cc
```diff
@@ -162,6 +163,63 @@ int make_profile_table_for_show(THD *thd, ST_SCHEMA_TABLE *schema_table) {
...
+    strftime(fn + i, 13, "%y%m%d%H%M%S", &ltm);
+    pfn = fn;
+    return jemalloc_mallctl("prof.dump", NULL, 0, &pfn, sizeof(pfn));
+  }
+  return jemalloc_mallctl("prof.dump", NULL, 0, NULL, 0);
+}
+
+/* Detected means Jemalloc loaded and profiling initialized using prof:true in
+ * MALLOC_CONF environment variable */
+bool jemalloc_detected() {
```
kamil-holubicki, line 245: missing return value documentation

sql/sql_profile.cc
```diff
@@ -162,6 +163,63 @@ int make_profile_table_for_show(THD *thd, ST_SCHEMA_TABLE *schema_table) {
...
+  char active;
+  size_t sz = sizeof(active);
+  int n = jemalloc_mallctl("opt.prof", &active, &sz, NULL, 0);
+  return !n && active;
+}
+
+/* Actually enable or disable profiling during runtime, for existing and
+ * newly created threads */
+int jemalloc_profiling_enable(bool enable) {
```
kamil-holubicki, line 258: missing return value documentation

sql/sql_profile.cc
kamil-holubicki, line 213: 1. This function uses magic constant values (13, 64)
2. string copying do not use safe functions
3. We ignore return value when we call this function. Should we a. handle return code/ b.  make this function being void?

sql/sys_vars.cc
```diff
@@ -2192,6 +2193,31 @@ static Sys_var_bool Sys_ft_query_extra_word_chars(
...
+    "Enable jemalloc profiling"
+    "Start MySQL with MALLOC_CONF set to 'prof:true' for this feature to work"
+    "Also Jemalloc should be compiled with profiling enabled",
+    GLOBAL_VAR(opt_jemalloc_profiling_enabled), CMD_LINE(OPT_ARG),
+    DEFAULT(false), NO_MUTEX_GUARD, NOT_IN_BINLOG,
+    ON_CHECK(jemalloc_profiling_check), ON_UPDATE(jemalloc_profiling_update),
+    NULL, sys_var::PARSE_EARLY);
+
+static Sys_var_bool Sys_jemalloc_detected(
+    "jemalloc_detected",
```
kamil-holubicki, line 2210: shouldn't we call this variable 'jemalloc_with_profiling'?
I understand that we can have jemalloc without profiling. In such a case 'jemaloc_detected==false' will be misleading.

storage/perfschema/table_malloc_stats.h
```diff
@@ -0,0 +1,116 @@
...
+};
+
+class table_malloc_stats_totals : public PFS_engine_table {
+ public:
+  /** Table share */
+  static PFS_engine_table_share m_share;
+  static PFS_engine_table *create(PFS_engine_table_share *);
+  static ha_rows get_row_count();
+
+  virtual int rnd_next() override;
```
kamil-holubicki: Multiple occurrences: unneeded virtual keyword when overriding.

storage/perfschema/table_malloc_stats.cc
```diff
@@ -0,0 +1,269 @@
...
+    make_row(m_pos.m_index);
+    m_next_pos.set_after(&m_pos);
+    return 0;
+  }
+
+  return HA_ERR_END_OF_FILE;
+}
+
+int table_malloc_stats::rnd_pos(const void *pos) {
+  set_position(pos);
```
kamil-holubicki, line 214: shouldn't we care about m_next_pos here as well?

storage/perfschema/table_malloc_stats.cc
```diff
@@ -0,0 +1,269 @@
...
+  return HA_ERR_RECORD_DELETED;
+}
+
+void table_malloc_stats::make_row(int index) {
+  char s[64];
+  size_t sz = sizeof(epoch);
+
+  memset(&m_row, 0, sizeof(m_row));
+
+  epoch++;
```
kamil-holubicki: Is the access protected by mutex? What if different threads are accessing this table?

storage/perfschema/table_malloc_stats.cc
```diff
@@ -0,0 +1,269 @@
...
+  unsigned n = 0;
+  sz = sizeof(unsigned);
+  /* Reading the index of merged arenas statistics, which equals to
+   * number of arenas. */
+  jemalloc_mallctl("arenas.narenas", &n, &sz, nullptr, 0);
+
+  sz = sizeof(size_t);
+  for (unsigned i = 0; i < NUM_STAT; i++) {
+    sprintf(s, "stats.arenas.%u.%s.%s", n, row_type[m_row.type], stat_type[i]);
+    jemalloc_mallctl(s, &m_row.stat[i], &sz, nullptr, 0);
```
kamil-holubicki, line 240: What happens if jemalloc_mallctl fails here (will not grab stat). Then m_row_stat[i] is uninitialized and we will see garbage in the table?

sql/sql_profile.cc
ldonoso: Good points. I have solved points 1 and 2.

About 3, I'd rather keep the return `int` although it is not checked. It doesn't hurt having that information available in case is needed.

sql/sys_vars.cc
ldonoso: please, mind there is another variable `jemalloc_profiling` to enable or disable the PS profiling. It is pretty similar to `jemalloc_with_profiling` so we could use `jemalloc_with_profiling_detected` but this is too long.

I think `jemalloc_detected` is fine as the doc of the variable clearly states the need of having enabled the profiling:
> This is true if Jemalloc with PROFILING enabled detected

storage/perfschema/table_malloc_stats.cc
ldonoso: Good point, I have check out the doc of the base class and it seems this methods is used just to return a row but it doesn't affect the global position.

https://github.com/ldonoso/percona-server/blob/d54bfaef0298f48864e0321cc080e389508c60c0/storage/perfschema/pfs_engine_table.h#L147-L151

Please, mind that as `m_next_pos` is not changed, `rnd_next` will continue to traverse the table in the correct order.

I have checkout out other perfschema tables, and they do implement this method in a similar way. e.i.

https://github.com/ldonoso/percona-server/blob/d54bfaef0298f48864e0321cc080e389508c60c0/storage/perfschema/table_global_variables.cc#L154-L158

storage/perfschema/table_malloc_stats.cc
ldonoso: None of the other methods (`rnd_next`, etc) of the rest of the `perfschema` tables are explicitly protected by a mutex so I guess it must be ok.

Please, are you talking of the access to `epoch`?

storage/perfschema/table_malloc_stats.cc
ldonoso: Unfortunately that's so. I will modify the code so we set it to 0 in case of an error.

sql/sql_profile.cc
```diff
@@ -163,6 +164,104 @@ int make_profile_table_for_show(THD *thd, ST_SCHEMA_TABLE *schema_table) {
...
+  It writes to /tmp/jeprof_mysqld.<PID>.<COUNTER>.<DATE> or
+  to a file determined by `prof_prefix` if `prof_prefix`
+  is different from default.
+
+  @return 0 on success.
+*/
+int jemalloc_profiling_dump() {
+  char *pfn;
+  size_t sz = sizeof(pfn);
+  int n = jemalloc_mallctl("opt.prof_prefix", &pfn, &sz, NULL, 0);
```
kamil-holubicki: I didn't notice it for the first time, but 8.0 code should use nullptr instead of NULL

sql/sql_profile.cc
```diff
@@ -163,6 +164,104 @@ int make_profile_table_for_show(THD *thd, ST_SCHEMA_TABLE *schema_table) {
...
+
+  /* Only write to custom file if user doesn't overwrite prof_prefix in
+   * MALLOC_CONF */
+  if (n || !strcmp(pfn, "jeprof")) {
+    char buff[64];
```
kamil-holubicki, line 221: Consider:
```	
static constexpr int buff_size = 64;
char buff[buff_size];
```

storage/perfschema/table_malloc_stats.cc
kamil-holubicki: Yes, I'm about epoch.

ldonoso: Thanks for the clarification. I think this is not protected by the static mutex of the table class. I will turn it into atomic.

storage/perfschema/table_malloc_stats.cc
ldonoso: I have just noticed that we were already doing that, at the beginning of the method we zero the array. I will revert these changes.

sql/sql_profile.cc
ldonoso: hmmm, I'd rather keep the call to `sizeof` and not use `buff_size`. The reason is that `buff_size` is not really a size but a number of elements. In this case it is the same as the size as the type of element is `char`.

As we are calling `sizeof` anyway, we would use `buff_size` only once, in the definition of `buff`, so I'd say it is not worth it to define `buff_size`.

storage/perfschema/table_malloc_stats.cc
```diff
@@ -0,0 +1,270 @@
...
+#include "storage/perfschema/pfs_column_values.h"
+#include "storage/perfschema/pfs_global.h"
+#include "storage/perfschema/pfs_instr.h"
+
+THR_LOCK table_malloc_stats_totals::m_table_lock;
+
+static std::atomic<uint64_t> epoch{0};
+
+static void set_jemalloc_epoch() {
+  auto epoch_copy = epoch++;
```
kamil-holubicki, line 34: I don't know if it makes any difference, but for the first time epoch_copy will have the value of 0, and it will be used in jemalloc_mallctl call.
Before converting it to atomic, we used incremented value here.

kamil-holubicki (approved):

kamil-holubicki:
Before merging, please ensure that it builds on all supported platforms.
@percona-ysorokin , any comments from your side?

PR #4224, opened 2021-03-08
[5.7] PS-7578: Replication failure with UPDATE when replica server has a PK and source not

venkatesh-prasad-v:
https://jira.percona.com/browse/PS-7578

Problem
----
Replication fails with the following error when executing an
UPDATE/DELETE statement in the source(where the table does not have a
PK) and the replica has a PK.

Last_SQL_Errno 1032
Last_SQL_Error Could not execute Update_rows event
on table test.t1; Can't find record in 't1', Error_code: 1032; handler
error HA_ERR_KEY_NOT_FOUND; the event's master log master-bin.000001,
end_log_pos 857

Analysis
----
When there are extra columns on slave server and when any of these
columns are in the key parts of the key used for searching the record in
the replica table, due to the use_index_extensions optimizer switch
(enabled by default even for slave applier threads), the function
`actual_key_parts()` function called from `calculate_key_len()` during
`index_read_map()` considers index_extensions into account and
incorrectly calculates the length of the `KEY::actual_key_parts` instead
of the `KEY::user_defined_key_parts` (in simple, this length the sum of
lengths of secondary key and primary key), and then searches the index
with wrong length, thereby causing "Can't find record in t1" error.

Consider this example.

1. source>  CREATE TABLE t1 (field1 char(5), KEY idx(field1));
2. source>  INSERT INTO t1 VALUES ('abc');
3. replica> ALTER TABLE t1 ADD COLUMN id_pk INT NOT NULL AUTO_INCREMENT PRIMARY KEY;
5. replica> SELECT * FROM t1;
            -------------
            field1  id_pk
            -------------
            abcd    1
            -------------
5. source>  UPDATE t1 SET field1='abcd';
6. Replication breaks.

Since there are extra columns on slave, the applier thread prepares full row
with default values and then merges with the before image row we got from
master and this newly prepared row is then sent to storage engine for updating.
```
Replicated row           : FE 03 61 62 63                <- Writeset from source server.
Prepared row (on replica): FD 61 62 63 20 20 00 00 00 00 <- bytes 7-10 corresponds to default value of INT column
Key data                 : 00 61 62 63 20 20             <- Key prepared from writeset.
Row present in index     : FD 61 62 63 20 20 01 00 00 00 <- bytes 7-10 corresponds to autoinc value 1 of INT column

Key length using actual_key_parts(index extensions) = 10 (1 null byte + 5 bytes for field1 + 4 bytes for id_pk)
Key length using user_defined_key_parts(no index extensions) = 6 (1 null byte + 5 bytes for field1)

Record compare with actual_key_parts (length 10) results in key not found
Prepared row:         FD 61 62 63 20 20 00 00 00 00
Row present in index: FD 61 62 63 20 20 01 00 00 00

Record compare with user_defined_key_parts (length 6) results in successful search
Prepared row:         FD 61 62 63 20 20
Row present in index: FD 61 62 63 20 20
```
Thus in the second case, when used with user_defined_key_parts, the applier
thread will be successfully able to search for the record.

Fix
---
Disable OPTIMIZER_SWITCH_USE_INDEX_EXTENSIONS for Row Based Replication
and make the applier thread to always use `KEY::user_defined_key_parts`
for searching the record.

Testing Done
---
https://ps57.cd.percona.com/job/percona-server-5.7-param/413/testReport/
Note: All failing tests are unrelated to fix proposed in this PR.

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_add_pk_on_extra_columns_on_replica.test
+++ b/mysql-test/suite/rpl/t/rpl_add_pk_on_extra_columns_on_replica.test
@@ -1,0 +1,176 @@
+# ==== Purpose ====
+#
+# This test verifies that update and delete events on a table are properly
+# applied by the replication applier threads when replica table has extra
+# columns, has primary key defined on any of the extra columns.
+#
+# ==== References ====
+#
+# PS-7578: Replication failure with UPDATE when replica server has a PK and
+#          source not
+
+--source include/have_binlog_format_row.inc
+--source include/master-slave.inc
+
+--source include/rpl_connection_slave.inc
+SET @saved_slave_rows_search_algorithms= @@global.slave_rows_search_algorithms;
+
+# Create an empty helper file which will be later used to load to the server.
+--let $LOAD_FILE= $MYSQLTEST_VARDIR/tmp/query_map.dat
+--write_file $LOAD_FILE
+EOF
+
+################################################################################
+# Here we write queries into the above file in the format
+#
+#          "query_identifier|query"
+#
+# so that this file can be used to load the queries into the table and fetch
+# the queries based on the identifier and execute it on the server for testing.
+#
+# Example: "CREATE_0|CREATE TABLE t1 (field1 char(15));"
+#          query_identifier = CREATE_0
+#          query = CREATE TABLE t1 (field1 char(15));
+#
+################################################################################
+
+############################
+# Variants of CREATE TABLE #
+############################
+# 1. Table with no keys.
+--exec echo "CREATE_0|CREATE TABLE t1 (field1 char(15));" >> $LOAD_FILE
+# 2. Table with a key.
+--exec echo "CREATE_1|CREATE TABLE t1 (field1 CHAR(15), field2 INT DEFAULT 1010, KEY idx (field1));" >> $LOAD_FILE
+# 3. Table with a composite key.
+--exec echo "CREATE_2|CREATE TABLE t1 (field1 CHAR(15), field2 INT DEFAULT 1010, KEY idx (field1, field2));" >> $LOAD_FILE
+
+########################
+# Query to insert data #
+########################
+--exec echo "DATA|INSERT INTO t1(field1) VALUES ('row1'); INSERT INTO t1(field1) VALUES ('row2'); INSERT INTO t1(field1) VALUES ('row3');" >> $LOAD_FILE
+
+###########################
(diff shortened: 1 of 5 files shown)


mysql-test/suite/rpl/t/rpl_add_pk_on_extra_columns_on_replica.test
```diff
@@ -0,0 +1,176 @@
...
+###########################
+# 1. No PK.
+--exec echo "ALTER_0|ALTER TABLE t1 ADD COLUMN extra_col INT NOT NULL DEFAULT 1;" >> $LOAD_FILE
+# 2. PK with AUTO_INCREMENT on additional column (adding one column).
+--exec echo "ALTER_1|ALTER TABLE t1 ADD COLUMN extra_col INT NOT NULL AUTO_INCREMENT PRIMARY KEY;" >> $LOAD_FILE
+# 3. Composite PK on both source's and additional columns.
+--exec echo "ALTER_2|ALTER TABLE t1 ADD COLUMN extra_col INT NOT NULL DEFAULT 1, ADD PRIMARY KEY (field1, extra_col);" >> $LOAD_FILE
+# 4. PK with AUTO_INCREMENT on additional column (adding two columns).
+--exec echo "ALTER_3|ALTER TABLE t1 ADD COLUMN extra_col1 INT NOT NULL DEFAULT 100, ADD COLUMN extra_col2 INT NOT NULL AUTO_INCREMENT, ADD PRIMARY KEY (extra_col2);" >> $LOAD_FILE
+
```
satya-bodapati, line 63: can you add a column at the beginning and see if errors out? or is it covered already?

satya-bodapati (changes requested):
Looks good. One doubt on testcase scenario, please check below

mysql-test/suite/rpl/t/rpl_add_pk_on_extra_columns_on_replica.test
venkatesh-prasad-v: This is an unsupported case and it has been covered by https://github.com/percona/percona-server/blob/5.7/mysql-test/extra/rpl_tests/rpl_extra_col_slave.test#L458-L802.

satya-bodapati: Great! no need for any changes then

satya-bodapati (approved):

PR #4228, opened 2021-03-10
Support for RocksDB filesystem plugins

yhr:
### This is a port to percona-server from upstream myrocks PR https://github.com/facebook/mysql-5.6/pull/1160. The upstream change will trickle down eventually, but this gives a heads up on the coming support for custom file system support in rocksdb. It would be great to get feedback on the build integration (especially any percona-server-specifics - the port required some changes)

RocksDB now supports building plugins from external repos (https://github.com/facebook/rocksdb/pull/7918) and this pull request adds the needed cmake magic to include these in percona server builds. The plugins to include are specified by ROCKSDB_PLUGINS: 

```
mkdir storage/rocksdb/plugin
git clone https://github.com/ajkr/dedupfs storage/rocksdb/plugin/dedupfs
mkdir build
cd build
cmake .. <build flags> -DROCKSDB_PLUGINS=dedupfs
```

Note: as this branch is not yet including the above pull request, the plugin directory does not exist, but can be created manually and the above example will work.

A configuration parameter, rocksdb_fs_uri is also added to allow the user to specify custom file systems.
```
rocksdb_fs_uri=dedupfs
```

These two changes allows external file systems to be included and used in myrocks. I've tested using dedupfs and a custom filesystem that I've developed for zoned block devices: zenfs (https://github.com/westerndigitalcorporation/zenfs)

### DIFF
--- a/storage/rocksdb/CMakeLists.txt
+++ b/storage/rocksdb/CMakeLists.txt
@@ -50,9 +50,30 @@ ELSE()
   # split the list into lines
   STRING(REGEX MATCHALL "[^\n]+" ROCKSDB_LIB_SOURCES ${SCRIPT_OUTPUT})
 
+  LIST(APPEND PLUGINS ${ROCKSDB_PLUGINS})
+  FOREACH(PLUGIN IN LISTS PLUGINS)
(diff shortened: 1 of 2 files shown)


percona-ysorokin:
Closing this PR at was superseded by https://github.com/percona/percona-server/pull/4354

PR #4236, opened 2021-04-01
PS-7617 - Timestamp is not set to CURRENT_TIMESTAMP in grant tables

venkatesh-prasad-v:
Background
---
This issue was found while cleaning up MTR test cases in PXC

1. Due to PS-6837 / Bug#98495, the `Timestamp` field of the `mysql.tables_priv`
table is set to NULL timestamp for all GRANT/REVOKE queries.

2. During server startup, the `role mysql.pxc.sst.role` is created and
is given access to `PERCONA_SCHEMA.xtrabackup_history` table.

3. Since the role `mysql.pxc.sst.role` is created using GRANT query, due
to PS-6837 / Bug#98495, the timestamp will have NULL value.

3. The tests execute
`CREATE TABLE <table_name> AS SELECT * FROM mysql.tables_priv;`

Since the column `Timestamp` is defined as

`timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`

the CREATE TABLE query fails with an error

`Incorrect datetime value: '0000-00-00 00:00:00' for column 'Timestamp'`

when it tries to insert the row having `mysql.pxc.sst.role` to the new
table.

Problem
-------
All the grant tables have a column called Timestamp to display when the
last change occurred for a user.

However, it was not fully implemented and caused the Timestamp field of
the mysql.*_priv table to be set to NULL timestamp("0000-00-00
00:00:00") for all GRANT/REVOKE queries and was preventing the logical
restore of grant tables to zero date.

Fix
---
We now write the the query start time as the timestamp for the
GRANT/REVOKE queries.

NOTE
----
For older datadirs, even after the upgrade, the existing entries can
still have zero timestamps. In such cases, for performing a logical
backup, users are expected to

1. Run GRANT/REVOKE query and then take backup.

   OR

2. Update the timestamp of the mysql.*_priv tables.

   Eg: UPDATE mysql.tables_priv SET Timestamp = CURRENT_TIMESTAMP WHERE User = 'u1';

   OR

3. Take backup and then edit the dump file and replace zero timestamp in the
   with CURRENT_TIMESTAMP.

  For example:

  change the contents
    REPLACE INTO `tables_priv` VALUES ('%','test','u1','t1','root@localhost','0000-00-00 00:00:00','','Select');
  to
    REPLACE INTO `tables_priv` VALUES ('%','test','u1','t1','root@localhost',CURRENT_TIMESTAMP,'','Select');

Testing Done
--
Jenkins: https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/903/testReport/
Failures: 
1. `rpl.rpl_deadlock_globalsystemvariable_channelmap`,`main.server_offline_debug` and`rpl.rpl_perfschema_log_status_blockage` : failing due to PS-7626.
2. `main.log_buffered-big` : Fixed in 8.0.23 merge branch
3. bitmap tests : Failing sporadically
4. `main.cast` - Fixed in 8.0.23 merge branch


sql/auth/sql_user_table.cc
```diff
@@ -1213,6 +1213,10 @@ int replace_column_table(THD *thd, GRANT_TABLE *g_t, TABLE *table,
       store_record(table, record[1]);  // copy original row
     }
 
+    timeval tm;
+    tm = thd->query_start_timeval_trunc(0);
+    table->field[5]->store_timestamp(&tm);
```
satya-bodapati, line 1218: can we merge definition and assignment?
timeval tm = thd->query_start_timeval_trunc(0)

satya-bodapati (changes requested):
Also to confirm if this fixes the original issue, can we add this to testcase
```
To clarify why this is an issue for a production environment:

mysqldump --compact --replace --skip-add-locks --skip-lock-tables --no-create-info --skip-triggers --set-gtid-purged=OFF mysql tables_priv | \
mysql --database=mysql ?
```
https://bugs.mysql.com/bug.php?id=98495

sql/auth/sql_user_table.cc
venkatesh-prasad-v: It is intentional. Because, it fails with 
```
[ 79%] Building CXX object sql/CMakeFiles/sql_main.dir/auth/sql_user_table.cc.o
/home/venki/work/ps/7617/sql/auth/sql_user_table.cc: In function 'int replace_routine_table(THD*, GRANT_NAME*, TABLE*, const LEX_USER&, const char*, const char*, bool, ulong, bool)':
/home/venki/work/ps/7617/sql/auth/sql_user_table.cc:1693:1: error: jump to label 'table_error' [-fpermissive]
 table_error:
 ^~~~~~~~~~~
/home/venki/work/ps/7617/sql/auth/sql_user_table.cc:1614:12: note:   from here
       goto table_error;
            ^~~~~~~~~~~
/home/venki/work/ps/7617/sql/auth/sql_user_table.cc:1646:18: note:   crosses initialization of 'timeval tm'
   struct timeval tm = thd->query_start_timeval_trunc(0);
```
As it is possible to skip the initialization because of a `goto` in the same function.

satya-bodapati: ah, gotos dont like initialization in middle. You can ingore

satya-bodapati (approved):

PR #4237, opened 2021-04-05
PS-1659 LP #1508909: Connect without proxy information hangs if "prox…

patrickbirch:
…y_protocol_networks" is enabled (5.7)

 Changes to be committed:
	modified:   source/flexibility/proxy_protocol_support.rst

### DIFF
--- a/doc/source/flexibility/proxy_protocol_support.rst
+++ b/doc/source/flexibility/proxy_protocol_support.rst
@@ -4,17 +4,16 @@
  Support for PROXY protocol
 ============================
 
-The proxy protocol allows an intermediate proxying server speaking proxy protocol (ie. HAProxy) between the server and the ultimate client (i.e. mysql client etc) to provide the source client address to the server, which normally would only see the proxying server address instead.
 
-As the proxy protocol amounts to spoofing the client address, it is disabled by default, and can be enabled on per-host or per-network basis for the trusted source addresses where trusted proxy servers are known to run. Unproxied connections are not allowed from these source addresses.
+The proxy server provides the benefits of load balancing, and security. Proxy protocol adds a header with connection information to the client's connection to the destination. Use the proxy protocol to pass the client IP address to a destination, which, normally, would only see the proxy server's address.
+
+You can enable the protocol on a per-host or a per-network basis for the trusted source addresses, where trusted proxy servers are known. You should specify from which network the destination expects the proxy headers. The `proxy_protocol_networks` should be set to the proxy address or a dedicated network range. The protocol is supported for TCP over IPv4 and IPv6 connections only. UNIX socket connections can not use the proxy protocol and do not fall under the effect of `proxy-protocol-networks`='*'. As a special exception, it is forbidden for the proxy protocol IP address to be ``127.0.0.1`` or ``::1``.
+
+Connections that do not use the proxy server are not allowed from the proxy-dedicated range. When the `proxy_protocol_networks` is enabled, the client communicates first. When a connection does not use the protocol, the server communicates first with an Initial Handshake packet and the client responds. If `proxy_protocol_networks` is enabled, a dedicated client address cannot connect to the destination without the proxy header. This connection stalls with no response and no error.
 
 .. note:: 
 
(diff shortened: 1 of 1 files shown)


doc/source/flexibility/proxy_protocol_support.rst
```diff
@@ -4,17 +4,22 @@
...
 
-As the proxy protocol amounts to spoofing the client address, it is disabled by default, and can be enabled on per-host or per-network basis for the trusted source addresses where trusted proxy servers are known to run. Unproxied connections are not allowed from these source addresses.
+The proxy server provides the benefits of load balancing, and security. Proxy protocol adds a header with connection information to the client's connection to the destination. Use the proxy network protocol to pass the client IP address to a destination, which, normally, would only see the proxy server's address.
+
+The proxy protocol is disabled by default. You can enable the protocol on a per-host or a per-network basis for the trusted source addresses, where trusted proxy servers are known. You should specify from which network the destination expects the proxy headers. The proxy_protocol_network should be set to the proxy address or a dedicated network range.
```
nastena1606:
```suggestion
The proxy protocol is disabled by default. You can enable the protocol on a per-host or a per-network basis for the trusted source addresses, where trusted proxy servers are known. You should specify from which network the destination expects the proxy headers. The ``proxy_protocol_network`` should be set to the proxy address or a dedicated network range.
```

doc/source/flexibility/proxy_protocol_support.rst
```diff
@@ -4,17 +4,22 @@
...
+
+Connections that do not use the proxy server are not allowed from the proxy-dedicated range. When the proxy_network_protocol is enabled, the client communicates first. When a connection does not use the protocol, the server communicates first with an Initial Handshake packet and the client responds. If proxy_network_protocol is enabled, a dedicated client address cannot connect to the destination without the proxy header. This connection stalls with no response and no error.
```
nastena1606:
```suggestion
Connections that do not use the proxy server are not allowed from the proxy-dedicated range. When the ``proxy_network_protocol`` is enabled, the client communicates first. When a connection does not use the protocol, the server communicates first with an Initial Handshake packet and the client responds. If ``proxy_network_protocol`` is enabled, a dedicated client address cannot connect to the destination without the proxy header. This connection stalls with no response and no error. 
```

doc/source/flexibility/proxy_protocol_support.rst
```diff
@@ -33,7 +38,7 @@ System Variables
   :dyn: No
   :default: ``(empty string)``
 
-This variable is a global-only, read-only variable, which is either a ``*`` (to enable proxying globally, a non-recommended setting), or a list of comma-separated IPv4 and IPv6 network and host addresses, for which proxying is enabled. Network addresses are specified in CIDR notation, i.e. ``192.168.0.0/24``. To prevent source host spoofing, the setting of this variable must be as restrictive as possible to include only trusted proxy hosts.
+This variable is a global-only, read-only variable, which is either a ``*`` (to enable proxy globally, a non-recommended setting), or a list of comma-separated IPv4 and IPv6 network and host addresses, for which proxy is enabled. Network addresses are specified in CIDR notation, i.e. ``192.168.0.0/24``. To prevent source host spoofing, the setting of this variable must be as restrictive as possible to include only trusted proxy hosts.
```
nastena1606: I would state the recommended value first and then mention * as an option. And add why it's not recommended using it and when one may use it.

doc/source/flexibility/proxy_protocol_support.rst
patrickbirch: fixed.

doc/source/flexibility/proxy_protocol_support.rst
patrickbirch: Fixed.

doc/source/flexibility/proxy_protocol_support.rst
patrickbirch: Rewritten

nastena1606 (approved):
LGTM

mohitj1988 (approved):
LGTM

PR #4239, opened 2021-04-08
PS-7631 [5.7]: Travis CI and Azure: installation of clang-9 fails on Ubuntu Bionic

inikep:
Do not try to install `clang++` what matches a regex.


inikep:
Tested at https://travis-ci.org/github/inikep/percona-server/builds/766460588

oleksandr-kachan (approved):

PR #4243, opened 2021-04-09
PS-7574: Improve communication between MTR test worker and main process (5.7)

oleksandr-kachan:
https://jira.percona.com/browse/PS-7574

This is a follow up improvement for
https://github.com/percona/percona-server/pull/3899.

When MTR worker finishes its job it sends shutdown and valgrind
reports to main process. It exits right after that. Two issues
are possible at this point:
- main process may lose last sent data from a worker;
- from time to time main MTR ptocess may silently exit right
  after closing a worker.

To solve both these issues communication protocol between
main MTR process and workers is modified in this change.
The following changes are done:
- Worker shutdown process is split into two steps - collecting
  and sending shutdown/valgrind reports, actual worker process shutdown.
- The GETREPORTS message gets sent to a worker to make it collect and
  send shutdown reports. Worker process stays active at this point.
- Once main process receives all reports it sends BYE to a worker.
  Worker process exits after receiving this message.
Above changes make sure main MTR process has time to collect
all data from a worker before it exits.

### DIFF
--- a/mysql-test/mysql-test-run.pl
+++ b/mysql-test/mysql-test-run.pl
@@ -879,8 +879,10 @@ sub run_test_server ($$$) {
 	      # Test has failed, force is off
 	      push(@$completed, $result);
 	      return $completed unless $result->{'dont_kill_server'};
-	      # Prevent kill of server, to get valgrind report
-	      print $sock "BYE\n";
+          # Prevent kill of server, to get shutdown/valgrind report from the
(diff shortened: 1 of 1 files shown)


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/428/

percona-ysorokin (approved):
LGTM

PR #4245, opened 2021-04-09
PS-7593: transaction-isolation level is not honored when changed at session level

kamil-holubicki:
https://jira.percona.com/browse/PS-7593

Problem:
1. Conditions: log_bin = ON, transaction_isolation = READ-COMMITTED,
binlog_format = ROW
2. Create two tables and execute REPLACE INTO ... SELECT
3. Change transaction_isolation to REPEATABLE-READ and binlog_format to
STATEMENT
4. Execute REPLACE INTO ... SELECT again
5. Error informing that it is not possible to execute a query is reported.

Cause:
1. When the table is created and then accessed, it is cached in thd's
table cache. It contains, inter alia, handler::cached_table_flags which
is calculated upon transaction_isolation value.
2. After transaction_isolation and binlog_format change
open_tables_for_query() and then sql_base.cc::open_table() finds
table in Table_cache
2. As the table was cached with READ-COMMITTED it has
HA_BINLOG_STMT_CAPABLE flag not set (handler::cached_table_flags)
3. Sql_cmd_insert_select::execute() and then handle_query() calls
select->prepare(thd) BEFORE call to lock_tables()
4. Commit e8cc22afb14914e56e535cb3e61a2b4310033d0e added call to
thd->decide_logging_format(). Decision is made upon cached value of
HA_BINLOG_STMT_CAPABLE flag (not set). This leads to immediate error
as the combination of READ-COMMITED and binlog_format STATEMENT is
not allowed.
5. lock_tables is called which would cause HA_BINLOG_STMT_CAPABLE flag
refresh (handler::cached_table_flags), but it is too late.

Solution:
Invalidate thd's table cache when transaction isolation changes.
That will cause opening the table and setting the proper value of flags
which are based on the current isolation level.


mysql-test/r/transaction_isolation_change.result
```diff
@@ -0,0 +1,15 @@
+SET @saved_transaction_isolation = @@session.transaction_isolation;
+SET @saved_binlog_format = @@session.binlog_format;
+SET @@session.transaction_isolation = "READ-COMMITTED";
+SET @@session.binlog_format = ROW;
+CREATE TABLE t1 (a INT);
+CREATE TABLE t2 (a INT);
+REPLACE INTO t1(a) SELECT a FROM t2;
+SET @@session.transaction_isolation = "REPEATABLE-READ";
```
satya-bodapati: Should we test other combinations of isolation levels too? Or is the disallowed combination only STATEMENT+READ_COMMITTED?

satya-bodapati (changes requested):
LGTM, one minor comment

mysql-test/r/transaction_isolation_change.result
kamil-holubicki: Hi @satya-bodapati I don't think it depends on disallowed combinations logic. This fix causes flags recalculation regardless of what combination is allowed or not. 
In other words, we may theoretically say that combination A + B is not allowed from now on, and the fix will still work.

satya-bodapati: @kamil-holubicki I understood the fix. Let me put my question in a different way.

Is there any other combination that fails after the regression commit and worked before the regression commit?

kornelpal: I was able to reproduce the issue with different isolation levels (all fixed by this PR):
- Both INSERT and REPLACE are affected
- Anything less strict that REPEATABLE READ is expected to cause an error for this INSERT/REPLACE with binlog_format=STATEMENT, so the only other option is SERIALIZABLE
- The query before INSERT/REPLACE can be SELECT too, in that case both READ UNCOMMITTED and READ COMMITTED can be used, and there is no need to change the binlog_format between the queries

A slightly different test case with different parameters:
```
SET @@session.binlog_format = STATEMENT;
SET @@session.transaction_isolation = "READ-UNCOMMITTED";
SELECT a FROM t1;
SET @@session.transaction_isolation = "SERIALIZABLE";
INSERT INTO t1(a) SELECT a FROM t2;
```

satya-bodapati: Thank you kornelpal, This is exactly what I was looking for.
@kamil-holubicki can you please add this scenario to testcase.

satya-bodapati (commented):
@kamil-holubicki I have one more concern with cache invalidation and the global mutex to do it.
I know that changing trx isolation is not so common, then it should be OK.

If there is work load that thinks SET SESSION tx_isolation level=READ_COMMITTED is harmless and does before every transaction, then there will be a hit in performance.

Is it possible to solve the original regression causing commit in a different way?

mbenshoof:
For reference, in the wild, this regression was found with a workload that ran entirely in read-committed/RBR.  The issue surfaced with pt-table-checksum and the need to set repeatable read for the change to SBR.

I would tend to agree to changing tx-isolation isn't super common, but definitely understand the concern about the extra mutex.

kornelpal:
I run into this regression with a workload that is using READ UNCOMMITTED for most SELECTs and the default REPEATABLE READ for writes, without changing the configured binlog_format=STATEMENT for any statements, so I share the performance concerns about this change.

Per-query variable statements make this very simple:
```
SET STATEMENT transaction_isolation='READ-UNCOMMITTED' FOR SELECT a FROM t1;
INSERT INTO t1(a) SELECT a FROM t2;
```

kamil-holubicki:
2 more proposals of solving the problem (last 2 commits pushed).
Both do not clear the TC, only reevaluate cached table flags.

1. Use of non-cached table flags in the problematic path. The drawback of this solution is that flags are recalculated always, even if transaction isolation didn't change

2. Invalidate cached table flags when transaction isolation changes. Then recalculate flags when requested for the first time. The advantage is that we are using cached flags always, and recalculate them only if transaction isolation changed. The drawback is that we have to introduce more changes. 
We could recalculate cached table flags immediately when transaction isolation changes (handler::init() call), however in most cases such immediate recalculation is not needed. We need it only handler::ha_table_flags() is called before handler::ha_external_lock() which happens only in Sql_cmd_insert_select::execute() path.

I would appreciate all your comments.

satya-bodapati (commented):
@kamil-holubicki This approach seems better. Can you please proceed with testing this fix?

satya-bodapati:
> 2 more proposals of solving the problem (last 2 commits pushed).
> Both do not clear the TC, only reevaluate cached table flags.
> 
>     1. Use of non-cached table flags in the problematic path. The drawback of this solution is that flags are recalculated always, even if transaction isolation didn't change
> 
>     2. Invalidate cached table flags when transaction isolation changes. Then recalculate flags when requested for the first time. The advantage is that we are using cached flags always, and recalculate them only if transaction isolation changed. The drawback is that we have to introduce more changes.
>        We could recalculate cached table flags immediately when transaction isolation changes (handler::init() call), however in most cases such immediate recalculation is not needed. We need it only handler::ha_table_flags() is called before handler::ha_external_lock() which happens only in Sql_cmd_insert_select::execute() path.
> 
> 
> I would appreciate all your comments.

I saw the implementation of flag calculation by SE's. It is not complex(no mutexes etc). Just ORing of some flags based on trx isolation levels. I would say this (1) is good enough for now.

kornelpal:
I think that the tests should be added to 8.0 too to avoid a similar regression in the future, I am not sure whether that happens normally.

PR #4251, opened 2021-04-12
PS-7182 : Create functionality to expose defined data from procfs for…

george-lorch:
… agentless environment (procfs)

  The ProcFS plugin is intended to provide access to Linux performance counters
  by running SQL queries against a Percona Server 8.0 (or other MySQL 8.0
  compatible variant). This is useful for Cloud and MySQL-as-service
  installations.

  The plugin reads files from /proc and /sys filesystems and returns the files
  name and content as is within individual rows of a new view
  INFORMATION_SCHEMA.PROCFS.  The schema definition of this view is:

  CREATE TEMPORARY TABLE `PROCFS` (
   `FILE` varchar(1024) NOT NULL DEFAULT '',
   `CONTENTS` longtext NOT NULL
  ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 

  Access to /proc and /sys dfiles and directories is limited by the system
  variable procfs_files_spec.  This variable is a global, read only variable,
  meaning that it may only be set either via the mysqld command line or within
  my.cnf and is not changeable at run time.  This is for security purposes to
  prevent a compromised account from giving itself greater access to the /proc
  or /sys filesystems.  Any files that are not included within the
  procfs_files_spec are ignored and considered to not exist.  You can write file
  access patterns in glob(7) style:  /sys/block/sd[a-z]/stat;/proc/version*

  The default value for procfs_files_spec is : /proc/cpuinfo;/proc/irq//;
  /proc/loadavg/proc/net/dev;/proc/net/sockstat;/proc/net/sockstat_rhe4;
  /proc/net/tcpstat;/proc/self/net/netstat;/proc/self/stat;/proc/self/io;
  /proc/self/numa_maps/proc/softirqs;/proc/spl/kstat/zfs/arcstats;/proc/stat;
  /proc/sys/fs/file-nr;proc/version;/proc/vmstat

  To install this plugin, copy the procfs.so file to the mysql plugin
  installation directory and execute "INSTALL PLUGIN procfs SONAME 'procfs.so';"

  Access to the INFORMATION_SCHEMA.PROCFS is limited to users with the
  ACCESS_PROCFS dynamic privilege.  Upon plugin startup, this new dynamic
  privilege is registered with the server.  After the plugin has been installed,
  grant a user access to the INFORMATION_SCHEMA.PROCFS view by executing:
  "GRANT ACCESS_PROCFS ON *.* TO 'user'@'host';"

  Authorized users can now obtain information from individual files by
  specifying the exact file name within a WHERE clause. Example:
  "SELECT * FROM INFORMATION_SCHEMA.PROCFS WHERE FILE = '/proc/version';".
  Specifying specific files is critical to limiting the impact of the plugin to
  the servers performance.  Limited file globbing is supported by the LIKE
  expression.  Failure to limit the set of files scanned via a WHERE clause
  through equality, LIKE or IN qualifiers can lead to lengthy query response
  times and high load and memory usage on the server as all files that match the
  procfs_files_spec must be opened, read, held in memory, and finally returned
  to the client.

  Some basic metrics are provided by status variables:

  Name                        Description
  -----------------------------------------------------------------------------
  procfs_access_violations    number of attempted queries by users without the
                              ACCESS_PROCFS privilege
  procfs_queries              number of queries made against the procfs view
  procfs_files_read           number of files read to provide content
  procfs_bytes_read           number of bytes read to provide content

 Credit goes to Nickolay Ihalainen (nickolay.ihalainen@percona.com) for the
 original implementation on which this work is based.  Thank you Nickolay!


plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,504 @@
...
+
+  while ((dir_entry = readdir(dir)) != 0) {
+    if (!fnmatch(pattern.c_str(), dir_entry->d_name,
+                 FNM_FILE_NAME | FNM_CASEFOLD | FNM_NOESCAPE | FNM_PERIOD)) {
+      if (dir_entry->d_type != DT_DIR) {
+        std::string fname =
+            std::string(path).append("/").append(dir_entry->d_name);
+
+        if (realpath(fname.c_str(), real_file_path.data())) {
+          if (strncmp("/sys/", real_file_path.data(), strlen("/sys/")) == 0 ||
```
dutow: strlen("/sys") (and similar) could be a const variable to avoid recalculating it in debug builds.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,504 @@
...
+  if (reg == nullptr) {
+    LogPluginErrMsg(ERROR_LEVEL, ER_LOG_PRINTF_MSG,
+                    "mysql_plugin_registry_acquire() returns NULL");
+    deinit_logging_service_for_plugin(&reg_srv, &log_bi, &log_bs);
+    return 1;
+  }
+  my_service<SERVICE_TYPE(dynamic_privilege_register)> reg_priv(
+      "dynamic_privilege_register", reg);
+  if (reg_priv.is_valid()) {
+    if (reg_priv->register_privilege(PRIVILEGE_NAME, strlen(PRIVILEGE_NAME))) {
```
dutow, line 449: shouldn't uninstall deregister the privilege?

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,504 @@
...
+  std::string::size_type first_star = query.find_first_of("*?[");
+  std::string::size_type last_sep_before_star =
+      query.find_last_of("/", first_star);
+  std::string::size_type first_sep_after_star =
+      query.find_first_of("/", first_star);
+
+  if (last_sep_before_star == std::string::npos) {
+    path = std::string(".");
+    pattern = query;
+  } else if (first_sep_after_star == std::string::npos) {
```
dutow, line 303: can this happen? if the last after the first_star exists, then there should be a filst too, even if it is the same.

plugin/procfs/tests/mtr/procfs_basic.result
```diff
@@ -0,0 +1,14 @@
+INSTALL PLUGIN procfs SONAME 'procfs.so';
+CREATE USER mysqltest_1@localhost;
+GRANT ALL PRIVILEGES ON mysqltest.* TO mysqltest_1@localhost;
+GRANT ACCESS_PROCFS ON *.* TO 'mysqltest_1'@'localhost';
+SELECT COUNT(*) FROM information_schema.procfs WHERE file = '/proc/version';
+COUNT(*)
+0
```
dutow: why is this 0, shouldn't /proc/version exists?

plugin/procfs/procfs.cc
george-lorch: good catch, should be using constexpr where possible

plugin/procfs/procfs.cc
george-lorch: true, not sure though what that means once granted to a user if MySQL cleans up users or not and if the priv ID can get reassigned to another.

plugin/procfs/procfs.cc
george-lorch: dunno, unfamiliar w/ the context as this wasn't my code originally, will need to look closer

plugin/procfs/tests/mtr/procfs_basic.result
george-lorch: another good catch, I suspect the default access mask is not allowing access to version, need to investigate.

plugin/procfs/procfs.cc
george-lorch: OK, not sure what you are getting at, there is last_sep_BEFORE_star and first_sep_AFTER_star, both can be present or missing or, more likely, I am missing the point of the question :)

dutow: My mistake, find_first_of begins the search at pos, while find_last_of searches until pos, I didn't remember this. I assumed both searches the [0,pos) interval.

plugin/procfs/tests/mtr/plugin.defs
```diff
@@ -0,0 +1,35 @@
+# Definition file for plugins.
```
percona-ysorokin: In 8.0 there should be no individual `plugin.defs` for new plugins - everything is put inside `mysql-test/include/plugin.defs`

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,527 @@
...
+        if (files_found.size() >= static_cast<size_t>(max_results)) break;
+      }
+    }
+  }
+
+  closedir(dir);
+  return files_found;
+}
+
+static std::vector<std::string> limited_glob(std::string query, int max_results,
```
percona-ysorokin:
```suggestion
static std::vector<std::string> limited_glob(const std::string &query, int max_results,
```

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,527 @@
...
+
+  if (strcmp(files_spec, DEFAULT_FILES_SPEC) != 0) {
+    LogPluginErrMsg(WARNING_LEVEL, ER_LOG_PRINTF_MSG,
+                    "default procfs_files_spec has been overridden with "
+                    "\"%s\", sensitive system information may be exposed via "
+                    "the information_schema.procfs table",
+                    files_spec);
+  }
+
+  SERVICE_TYPE(registry) *reg = mysql_plugin_registry_acquire();
```
percona-ysorokin: At this point plugin registry has already been acquired by the `init_logging_service_for_plugin()` call - you can simply use `reg_srv` here.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,527 @@
...
+      my_malloc(PSI_NOT_INSTRUMENTED, MAX_CONTENTS_SIZE, MY_ZEROFILL));
+
+  return 0;
+}
+
+static int view_deinit(void *) {
+  LogPluginErrMsg(INFORMATION_LEVEL, ER_LOG_PRINTF_MSG,
+                  "Plugin procfs de-initializing...");
+
+  SERVICE_TYPE(registry) *reg = mysql_plugin_registry_acquire();
```
percona-ysorokin: Likewise

percona-ysorokin (changes requested):

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,527 @@
...
+          }
+        }
+        return false;
+      }
+    }
+  }
+  return true;
+}
+
+static std::vector<std::string> limited_glob_files(std::string path,
```
kamil-holubicki: const std::string&

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,527 @@
...
+      if (dir_entry->d_type == DT_DIR || dir_entry->d_type == DT_LNK ||
+          dir_entry->d_type == DT_UNKNOWN) {
+        std::string new_pattern = std::string(path)
+                                      .append("/")
+                                      .append(dir_entry->d_name)
+                                      .append("/")
+                                      .append(rest);
+        std::vector<std::string> v = limited_glob(
+            new_pattern, max_results - files_found.size(), max_depth - 1);
+        files_found.insert(files_found.end(), v.begin(), v.end());
```
kamil-holubicki: I don't know how much are we concerned here about efficiency, but maybe
`files_found.insert(files_found.end(), std::make_move_iterator(v.begin()), std::make_move_iterator(v.end()));`
?

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,527 @@
...
+    if (path.rfind("/proc", 0) != 0 && path.rfind("/sys", 0) != 0) continue;
+
+    if (path.find_first_of("*?[") == std::string::npos) {
+      files.push_back(path);
+      continue;
+    }
+
+    std::vector<std::string> v1 =
+        limited_glob(path, MAX_FILES_PER_PATTERN, MAX_PATTERN_DEPTH);
+    files.insert(files.end(), v1.begin(), v1.end());
```
kamil-holubicki: Maybe consider std::make_move_iterator as above?

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,527 @@
...
+      in_args[eq_arg] = true;
+    } else if (!get_like_condition_argument(cond, &like_arg, I_S_FILE) &&
+               !like_arg.empty()) {
+    } else {
+      get_in_condition_argument(cond, in_args, I_S_FILE);
+    }
+  }
+
+  std::vector<std::string> files;
+  fill_files_list(files);
```
kamil-holubicki, line 403: Here we are passing the caller allocated vector to be filled in.
We also have limited_glob() and limited_glob_files() functions which effectively fill the vector as well, but in that cases we return local vector instead of passing caller-allocated one as the parameter and let the compiler to figure out that local vector should be moved to caller.

Just a matter of style, but within one file maybe we should stick to one way of another?

plugin/procfs/tests/mtr/plugin.defs
george-lorch: Right, right

plugin/procfs/procfs.cc
george-lorch: good catch

plugin/procfs/procfs.cc
george-lorch: and another good catch

plugin/procfs/procfs.cc
george-lorch: yeah, a little extra efficiency is always a good thing

plugin/procfs/procfs.cc
george-lorch: yup, another good call, also a bit of an efficiency improvement to pass reference rather than return by value

george-lorch: actually, no, now I see that there is a reason that these are done this way and that is to ultimately limit the number of found results.  It would still be possible to achieve this limiting with some re-writing, but, IMHO I would rather not go too deep down that hole at this time since this is already written and tested.

plugin/procfs/procfs.cc
george-lorch: Ahh, right, the example I got this from a structured this way but I see that these are the same type and interchangeable, that simplifies things greatly.

plugin/procfs/procfs.cc
george-lorch: seems to work OK

plugin/procfs/procfs.cc
george-lorch: Hmm, no, I'm wrong, this can be done easily w/o breaking anything.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,514 @@
...
+    }
+  }
+  return true;
+}
+
+static void limited_glob_files(const std::string &path,
+                               const std::string &pattern, int max_results,
+                               std::vector<std::string> &files_found) {
+  static const constexpr char SYS_NAME[] = "/sys/";
+  static const constexpr int SYS_NAME_LEN = sizeof(SYS_NAME);
```
dutow, line 252: previously this was strlen(SYS_NAME), which returns 5 (without the trailing 0). Now it's sizeof, which includes the zero, resulting in a length of 6. Based on the use I think this is what we actually want, just mentioning it so you can verify it.

plugin/procfs/procfs.cc
```diff
@@ -0,0 +1,514 @@
...
+    std::string path;
+    std::getline(list, path, ';');
+    if (path.rfind("/proc", 0) != 0 && path.rfind("/sys", 0) != 0) continue;
+
+    if (path.find_first_of("*?[") == std::string::npos) {
+      files.push_back(path);
+      continue;
+    }
+
+    limited_glob(path, MAX_FILES_PER_PATTERN, MAX_PATTERN_DEPTH, files);
```
dutow, line 358: This is definitely out of scope of this PR, but there's an optimization (refactoring) possibility here, and we probably should at least create a jira ticket about it.

Right now, we traverse the proc/sys ft for every wildcard entry in the list. We probably want to do that only once, and compare it against all possible patterns.

plugin/procfs/procfs.cc
george-lorch: yes, exactly and I did myself in fact when I made this change.  The same goes for the PROC_NAME which would be strlen = 6 and sizeof = 7.

plugin/procfs/procfs.cc
george-lorch: +1 see PS-7703

george-lorch:
@dutow, @kamil-holubicki, @percona-ysorokin, thanks guys for the reviews so far.  I think I have all suggestions incorporated now, can you all please re-review.  I am trying to get a jenkins run on the new changes but we seem to be having some difficulty there at the moment.  The simple mtr test passe locally and some manual testing shows things appear to be working as expected.  Thanks!

george-lorch:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/931/

kamil-holubicki (approved):
I have no more concerns. Please note that I didn't check the logic in every detail, focused on C++ compliance aspect.

PR #4255, opened 2021-04-14
PS-7636 [5.7]: Run a binary after a CPU or compiler feature detection in MyRocks

inikep:
Some platforms allow code to be compiled but fails to run a binary:
```
$ g++ avx2.cc -o avx2 -std=c++11 -mavx2
$ ./avx2
Illegal instruction (core dumped)
$ cat avx2.cc
#include <cstdint>
#include <immintrin.h>
int main() {
  const auto a = _mm256_setr_epi32(0, 1, 2, 3, 4, 7, 6, 5);
  const auto b = _mm256_permutevar8x32_epi32(a, a);
}
```

This patch checks if a feature can be run successfully after a compilation.

### DIFF
--- a/cmake/compiler_features.cmake
+++ b/cmake/compiler_features.cmake
@@ -16,20 +16,21 @@
 
 # Functions to detect features supported by compiler
 
-include(CheckCXXSourceCompiles)
+include(CheckCXXSourceRuns)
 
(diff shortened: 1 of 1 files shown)


inikep:
Tested at https://travis-ci.org/github/inikep/percona-server/builds/767005944

oleksandr-kachan (approved):

PR #4258, opened 2021-04-14
PS-7626 - Remove lock when reading net_buffer_shrink_time

ldonoso:
LOCK_global_system_variables is not owned when reading other
global variables.

As the type of net_buffer_shrink_interval is `ulong` the read/writes
are atomic but there still the hazards of using stale values.

As this is a mild problem, the risk is accepted.


kamil-holubicki (approved):
lgtm

PR #4259, opened 2021-04-15
PS-7593: transaction-isolation level is not honored when changed at session level

kamil-holubicki:
https://jira.percona.com/browse/PS-7593

Problem:
1. Conditions: log_bin = ON, transaction_isolation = READ-COMMITTED,
binlog_format = ROW
2. Create two tables and execute REPLACE INTO ... SELECT
3. Change transaction_isolation to REPEATABLE-READ and binlog_format to
STATEMENT
4. Execute REPLACE INTO ... SELECT again
5. Error informing that it is not possible to execute query is reported.

Cause:
1. When the table is created and then accessed, it is cached in thd's
table cache. It contains, inter alia, handler::cached_table_flags which
is calculated upon transaction_isolation value.
2. After transaction_isolation and binlog_format change
open_tables_for_query() and then sql_base.cc::open_table() finds
table in Table_cache
2. As the table was cached with READ-COMMITTED it has
HA_BINLOG_STMT_CAPABLE flag not set (handler::cached_table_flags)
3. Sql_cmd_insert_select::execute() and then handle_query() calls
select->prepare(thd) BEFORE call to lock_tables()
4. Commit e8cc22afb14914e56e535cb3e61a2b4310033d0e added call to
thd->decide_logging_format(). Decision is made upon cached value of
HA_BINLOG_STMT_CAPABLE flag (not set). This leads to immediate error
as the combination of READ-COMMITED and binlog_format STATEMENT is
not allowed.
5. lock_tables is called which would cause HA_BINLOG_STMT_CAPABLE flag
refresh (handler::cached_table_flags), but it is too late.

Solution:
Force recalculation of handler::cached_table_flags if they are used
before locking the table (call introduced in above-mentioned regression
commit)


kamil-holubicki:
Hi @kornelpal, @mbenshoof, 
I'm not sure why the original PR https://github.com/percona/percona-server/pull/4245 was automatically closed by GitHub after rebasing the solution to the GCA point, so creating new PR with the agreed solution.

kornelpal:
I was able to reproduce the issue between sessions too, so I think the stale cached flags can be reused between sessions:
```
#!/bin/bash
service mysql restart
mysql -A -e "drop database if exists test; create database test; create table test.t1 (a int); create table test.t2 (a int)"
for i in {0..100}
do
  mysql -A -e "set @@session.transaction_isolation = 'READ-UNCOMMITTED'; select a from test.t1"
done
mysql -A -e "set @@session.binlog_format = STATEMENT; insert into test.t1(a) select a from test.t2"
```

The current patch fixes this case too, but some of the other fixes discussed in PR #4245 might not fix all the affected cases.

I am not sure whether there is an easy way to a add a similar test case, but having more regression tests could prevent the regression from being reintroduced when switching to the upstream fix, once available.

kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-param/430/

satya-bodapati (approved):
LGTM!

PR #4262, opened 2021-04-20
PS-5364: Fix Hashicorp Vault plugin build with GNU 5.4.0

oleksandr-kachan:
https://jira.percona.com/browse/PS-5364


percona-ysorokin (approved):
LGTM

PR #4265, opened 2021-04-20
PS-7655 Document 'audit_log_timestamps' (8.0)

patrickbirch:
Changes to be committed:
	modified:   source/management/audit_log_plugin.rst

### DIFF
--- a/doc/source/management/audit_log_plugin.rst
+++ b/doc/source/management/audit_log_plugin.rst
@@ -680,6 +680,22 @@ This variable is used to specify the ``priority`` value for syslog. This
 variable has the same meaning as the appropriate parameter described in the
 `syslog(3) manual <http://linux.die.net/man/3/syslog>`_.
 
+.. variable:: audit_log_timestamps
+
+   :cli: Yes
+   :scope: Global
+   :dyn: No
+   :vartype: Numeric
+   :default: UTC
+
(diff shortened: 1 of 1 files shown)


doc/source/management/audit_log_plugin.rst
```diff
@@ -680,6 +680,22 @@ This variable is used to specify the ``priority`` value for syslog. This
...
+
+   :cli: Yes
+   :scope: Global
+   :dyn: No
+   :vartype: Numeric
+   :default: UTC
+
+This variable controls the timestamp time zone added to events in the audit log. 
+
+The available vales are the following:
```
alina-derkach-oaza: values

oleksandr-kachan:
@georgelorchpercona, looks like corresponding PR is still opened https://github.com/percona/percona-server/pull/4231. Am I missing something?

doc/source/management/audit_log_plugin.rst
patrickbirch: Fixed.

george-lorch:
yes, it has been determined that the PR in question can not be utilized as it changes format of audit log record without an option too enable/disable.  The code change to the call that enables this cna not be controlled in this way, so we will likely not take in this PR at all.

PR #4267, opened 2021-04-21
PS-7658: Fix compilation issues with clang-12 (5.7)

oleksandr-kachan:
https://jira.percona.com/browse/PS-7658

With clang-12 there is a check for missing comma in string array
initialization. Added extra parentheses around the string to supress error.


plugin/query_response_time/CMakeLists.txt
```diff
@@ -1,3 +1,9 @@
+# Suppress warnings for clang-12
```
percona-ysorokin: @oleksandr-kachan It looks like the error message is a bit misleading
I found out that the problem is in the calculation of `POSITIVE_POWER_COUNT` and `NEGATIVE_POWER_COUNT` constants - using floating point arithmetic cannot be done in constant evaluation context.
Here is how I resolved the issue.
```
diff --git a/plugin/query_response_time/query_response_time.cc b/plugin/query_response_time/query_response_time.cc
index 2ef8d63a11d..ca2abf71bcc 100644
--- a/plugin/query_response_time/query_response_time.cc
+++ b/plugin/query_response_time/query_response_time.cc
@@ -75,8 +75,8 @@

   Last counter always use for time overflow
 */
-#define POSITIVE_POWER_COUNT ((int)(3.32192809 * TIME_STRING_POSITIVE_POWER_LENGTH))
-#define NEGATIVE_POWER_COUNT ((int)(3.32192809 * TIME_STRING_NEGATIVE_POWER_LENGTH))
+#define POSITIVE_POWER_COUNT ((int)(332192809ULL * TIME_STRING_POSITIVE_POWER_LENGTH / 100000000ULL))
+#define NEGATIVE_POWER_COUNT ((int)(332192809ULL * TIME_STRING_NEGATIVE_POWER_LENGTH / 100000000ULL))
 #define OVERALL_POWER_COUNT (NEGATIVE_POWER_COUNT + 1 + POSITIVE_POWER_COUNT)

 #define MILLION ((unsigned long)1000 * 1000)
```

percona-ysorokin (changes requested):

percona-ysorokin (approved):
LGTM

PR #4270, opened 2021-04-21
PS-7252 : Investigate the unused member `least_occupied_workers` of  `Relay_log_info` class

nitendra-bhosle:
PS-7252 : Investigate the unused member `least_occupied_workers` of  `Relay_log_info` class

https://jira.percona.com/browse/PS-7252

Fix : Removed least_occupied_workers member from class Relay_log_info
      As it was not used in server code

### DIFF
--- a/sql/rpl_mts_submode.cc
+++ b/sql/rpl_mts_submode.cc
@@ -947,9 +947,10 @@ Mts_submode_logical_clock::get_least_occupied_worker(Relay_log_info *rli,
 Slave_worker*
 Mts_submode_logical_clock::get_free_worker(Relay_log_info *rli)
 {
-  for (Slave_worker **it= rli->workers.begin(); it != rli->workers.end(); ++it)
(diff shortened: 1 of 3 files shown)


nitendra-bhosle:
Jenkins run , after removing dead code : https://ps57.cd.percona.com/job/percona-server-5.7-param/434/
Jenkins run , before removing code : https://ps57.cd.percona.com/view/5.7/job/percona-server-5.7-param/435/

venkatesh-prasad-v:
Look like the intention of this sorted array based on workers' length is only to reduce the time taken by the co-ordinator thread to quickly choose the least occupied worker, but it was never implemented. IMO, it would be better if we retain this worker array and use it in places where we try to check/get the least occupied worker.

Can we do some more investigation to fulfill its original intention, by using the `rli->least_occupied_workers` instead of `rli->workers` in [Mts_submode_logical_clock::get_free_worker()](https://github.com/percona/percona-server/blob/5.7/sql/rpl_mts_submode.cc#L950), which returns the least occupied worker in case of Logical_clock MTS mode?

PR #4272, opened 2021-04-21
PS-7648 Optimizer switch "favor_range_scan" is not documented (5.7)

patrickbirch:
Changes to be committed:
	modified:   source/performance/query_limit_records.rst


george-lorch (approved):
Might want to have Yves take a read as well

altmannmarcelo (approved):

PR #4274, opened 2021-04-22
PS-7595: Same version upgrade from PS->PXC needs mysql_upgrade

kennt-percona:
Issue
The performance_schema.pxc_cluster_view table is not created when upgrading
from PS to PXC (same version).

Solution
If --upgrade=FORCE is used, always perform the performance schema initialization.
Previously, it would do a version check with the DD version (and would
only upgrade if the DD version changed).


kennt-percona:
https://ps80.cd.percona.com/job/percona-server-8.0-param/906/

satya-bodapati (approved):

PR #4284, opened 2021-05-03
PS-7657 crash seen with update query on a partition table having comp…

nitendra-bhosle:
…ressed column

PS-7657 crash seen with update query on a partition table having compressed column
    
    https://jira.percona.com/browse/PS-7657
    
    Analysis :
    Update statements execution flow on tables with partitions by key and having compressed column:
    
    1.First it searches for the row and positions the cursor (the blob is decompressed and stored in blob_heap and record is returned to server)
    2. Builds new row and then calls update(old_row, new_row). Since the blob field doesn't change, the blob from old row is stored in the new row (mysql record)
    3. Parition hanlder detects that the updated row should be in different partition
    4. Insert the new row in new parition. During insertion, we empty the blob heap. This releases the blocks used by blob and corrupts the blob data in new row
    5. So we end up inserting garbage to the new partition after the update
    
    Fix :
    Avoided prebuilt->compress_heap from being empty
    till it's required


storage/innobase/row/row0mysql.cc
```diff
@@ -2299,9 +2299,6 @@ row_insert_for_mysql_using_ins_graph(
 		return(DB_TABLE_CORRUPT);
 	}
 
-	if (UNIV_LIKELY_NULL(prebuilt->compress_heap))
-		mem_heap_empty(prebuilt->compress_heap);
-
```
satya-bodapati, line 2304: @nitendra-bhosle Can you please explain how the freed compress_heap causes update to fail (not find the record?)?

satya-bodapati (changes requested):

storage/innobase/row/row0mysql.cc
nitendra-bhosle: @satya-bodapati 'mem_heap_empty' function Empties a memory heap but first memory block of the heap is not freed 
hence inside 'function row_compress_column' prebuilt->compress_heap is not NULL 
But memory is emptied.

satya-bodapati: @nitendra-bhosle what happens after that? is the buffer insufficient to compress the column?

satya-bodapati: and the question how this translates to update not being able to find the row still remains

nitendra-bhosle: Inside method : 'row_compress_column'
buf = static_cast<byte*>(mem_heap_zalloc(
			prebuilt->compress_heap,buf_len));
this call operates on prebuilt->compress , which tries to memset memory , which it doesn't have ownership,
Which leads to memory pointed by pointer "const byte*	data" to be getting corrupted/emptied.
And hence buf returned from this method is empty .Which is causing update not being able to find the row

satya-bodapati: how did you prove that it doesn't have ownership? 
memset() is working on new block of memory returned by heap_zalloc().

I am not sure if I understand the fix. row_mysql_prebuilt_free_compress_heap() seems to be called too.

nitendra-bhosle: Inside function 'row_compress_column' 
buf = static_cast<byte*>(mem_heap_zalloc(
			prebuilt->compress_heap,buf_len));
this statement empties 'data' . Which happens when prebuilt->compress_heap doesn't have memory allocated.
I checked this by manually executing 
prebuilt->compress_heap =
			mem_heap_create(ut_max(UNIV_PAGE_SIZE, buf_len)); 
statement everythime.

'mem_heap_create'  create heap of size maximum of UNIV_PAGE_SIZE or buf_len , which is most likely to be UNIV_PAGE_SIZE

'mem_heap_zalloc' allocates and memsets  'buf_len' bytes out of heap block created by  'mem_heap_create'

'row_mysql_prebuilt_free_compress_heap' free's heap created by 'mem_heap_create' , this executes on end of statement.

Ideally there should be pair of 'mem_heap_create' and 'row_mysql_prebuilt_free_compress_heap' for each sql statement

satya-bodapati: row_mysql_prebuilt_free_compress_heap() is already called at end of statement. once heap is created, other iterations need not bother about creating the heap again. mem_heap_t is dynamic memory allocator. ie it will allocate new blocks requested. Something else is going @nitendra-bhosle

satya-bodapati: Specially since this happening with partitions, check if we need separate compressed heaps for paritions. 
Check m_blob_heap_parts/ clear_blob_heaps(). Do we need something like that?

satya-bodapati (changes requested):

storage/innobase/row/row0mysql.cc
nitendra-bhosle: I will explore this.

satya-bodapati: @nitendra-bhosle 

I debugged a bit and I think your fix makes sense but here is the explanation

So the way blobs work is storage engines only return a pointer to the blob (not make a copy) to the server. This pointer is from the prebuilt->blob_heap. If blob_heap is emptied, the blob returned by SE is corrupted.

Coming to the UPDATE statement:

1. First it searches for the row and positions the cursor (the blob is decompressed and stored in blob_heap and record is returned to server)
2. Builds new row and then calls update(old_row, new_row). Since the blob field doesn't change, the blob from old row is stored in the new row (mysql record)
3. Parition hanlder detects that the updated row should be in different partition
4. Insert the new row in new parition. During insertion, we empty the blob heap. This releases the blocks used by blob and corrupts the blob data in new row
5. So we end up inserting garbage to the new partition after the update

Requests
--------
1. Update commit message explaining the sequence clearly

2. Fix this blob heap code in row_sel_store_mysql_rec()
           if (UNIV_LIKELY_NULL(prebuilt->blob_heap)) {
                row_mysql_prebuilt_free_blob_heap(prebuilt);
        }

        if (UNIV_LIKELY_NULL(prebuilt->compress_heap))
                mem_heap_empty(prebuilt->compress_heap);
  
         call free for compress_heap as well
        
3. Bring the replication testcase from PS-7599

satya-bodapati: 4. Run the mtr testcase on valgrind/ASAN and paste the result here.

nitendra-bhosle:
```
mtr test-case on valgrind results 
./mtr --valgrind percona_bug_ps7599.test

TEST                                      RESULT   TIME (ms) or COMMENT

worker[1] Using MTR_BUILD_THREAD 300, with reserved ports 13000..13009
[ 20%] rpl.percona_bug_ps7599 'mix'             [ skipped ]  Doesn't support --binlog-format='mixed'
[ 40%] rpl.percona_bug_ps7599 'row'             [ pass ]  15462
[ 60%] rpl.percona_bug_ps7599 'stmt'            [ skipped ]  Doesn't support --binlog-format='statement'
[ 80%] shutdown_report                          [ pass ]       
[100%] valgrind_report                          [ pass ]       
Spent 15.462 of 48 seconds executing testcases

Completed: All 3 tests were successful.

2 tests were skipped, 0 by the test itself.

./mtr --valgrind xtradb_compressed_columns.test

TEST                                      RESULT   TIME (ms) or COMMENT

worker[1] Using MTR_BUILD_THREAD 300, with reserved ports 13000..13009
[ 33%] innodb.xtradb_compressed_columns         [ pass ]  54374
[ 66%] shutdown_report                          [ pass ]       
[100%] valgrind_report                          [ pass ]       
Spent 54.374 of 78 seconds executing testcases

Completed: All 3 tests were successful.
```

mysql-test/suite/rpl/t/percona_bug_ps7599.test
```diff
@@ -0,0 +1,20 @@
+#
```
venkatesh-prasad-v: IMO, using JIRA number in the testcase name is not very intuitive and will make the person to actually look into the test while categorizing the failure in case the test fails in future.

Can we rename the testcase to `rpl_update_partitioned_table_with_compressed_column.test` or something like that?

mysql-test/suite/rpl/t/rpl_update_partitioned_table_with_compressed_column.test
```diff
@@ -0,0 +1,20 @@
+#
+# PS-7599 : Node inconsistency seen in PS during update on compressed column table
+#
+
+--source include/have_binlog_format_row.inc
+--source include/master-slave.inc
+
+DROP TABLE IF EXISTS tt_2_p;
```
satya-bodapati: remove this

mysql-test/suite/rpl/t/rpl_update_partitioned_table_with_compressed_column.test
```diff
@@ -0,0 +1,20 @@
...
+
+CREATE TABLE tt_2_p (f1 INT, f2 BLOB COLUMN_FORMAT COMPRESSED) ENGINE=INNODB PARTITION BY HASH (f1) PARTITIONS 2;
```
satya-bodapati: just use name t1

mysql-test/suite/innodb/t/xtradb_compressed_columns.test
```diff
@@ -220,3 +220,18 @@ CREATE TABLE t1 (c1 VARBINARY(5) COLUMN_FORMAT COMPRESSED) ENGINE=InnoDB;
 ALTER TABLE t1 ADD c2 CHAR(10) FIRST;
 INSERT INTO t1 VALUES(0,11111);
 DROP TABLE t1;
+
+--echo #
+--echo # Bug PS-7657 crash seen with update query on a partition table having compressed column
+--echo #
+CREATE TABLE t_ps_7657 (ip_col INT, i0 INT, v1 VARCHAR(29), v2 VARCHAR(23), lt3 LONGTEXT COLUMN_FORMAT COMPRESSED, b4 BLOB , v5 VARCHAR(10), g6  CHAR(9) AS  (CONCAT(SUBSTRING(ip_col,1,1),SUBSTRING(i0,1,3),SUBSTRING(lt3,1,3),SUBSTRING(v1,1,2))), INDEX tt_47_pi1(i0, v2 ASC, ip_col, v5 DESC, g6, v1)  )  ENGINE=INNODB PARTITION BY KEY (ip_col) PARTITIONS 13;
```
satya-bodapati: use t1 as name

mysql-test/suite/innodb/t/xtradb_compressed_columns.test
```diff
@@ -220,3 +220,18 @@ CREATE TABLE t1 (c1 VARBINARY(5) COLUMN_FORMAT COMPRESSED) ENGINE=InnoDB;
...
+
+update t_ps_7657 set ip_col=2680 where ip_col = 23486;
+update t_ps_7657 set ip_col=23486 where ip_col = 2680;
+
```
satya-bodapati, line 233: also add SELECT to ensure the blob stored isn't corrupted

satya-bodapati (changes requested):
LGTM! minor suggestions

mysql-test/suite/rpl/t/rpl_update_partitioned_table_with_compressed_column.test
```diff
@@ -0,0 +1,20 @@
...
+
+REPLACE INTO tt_2_p(f1,f2)VALUES(0,'123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456');
+
+UPDATE tt_2_p SET f1=1 WHERE f1 = 0;
+
+--source include/sync_slave_sql_with_master.inc
+
+DROP TABLE tt_2_p;
```
venkatesh-prasad-v: Including`include/sync_slave_sql_with_master.inc` changes the connection to slave in the end.
So, this DELETE is executed on slave and not on the master.

So, please use
```
--source include/rpl_connection.inc
DROP TABLE tt_2_p;
--source include/rpl_end.inc
```

mysql-test/suite/innodb/t/xtradb_compressed_columns.test
```diff
@@ -220,3 +220,18 @@ CREATE TABLE t1 (c1 VARBINARY(5) COLUMN_FORMAT COMPRESSED) ENGINE=InnoDB;
...
+DROP TABLE t_ps_7657;
+
+
+
```
venkatesh-prasad-v: Remove these blank lines in the end of the test.

mysql-test/suite/innodb/t/xtradb_compressed_columns.test
venkatesh-prasad-v: It is probably better to use asserts in addition to SELECTs.

mysql-test/suite/rpl/t/rpl_update_partitioned_table_with_compressed_column.test
```diff
@@ -0,0 +1,18 @@
...
+--source include/master-slave.inc
+
+CREATE TABLE t1 (f1 INT, f2 BLOB COLUMN_FORMAT COMPRESSED) ENGINE=INNODB PARTITION BY HASH (f1) PARTITIONS 2;
+
+REPLACE INTO t1 (f1,f2)VALUES(0,'123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456');
+
+UPDATE t1 SET f1=1 WHERE f1 = 0;
+
+--source include/sync_slave_sql_with_master.inc
+--source include/rpl_connection_master.inc
```
venkatesh-prasad-v: In between these, can we validate that both master and slave have same data?

We can just source diff_tables.inc
```
--let $diff_tables= master:t1,slave:t1
--source include/diff_tables.inc
```

venkatesh-prasad-v (changes requested):

mysql-test/suite/innodb/t/xtradb_compressed_columns.test
satya-bodapati: either way is fine. at least for SELECTs. Otherwise, entire mtr suite should have been using asserts and never SELECTs. isn't it?

Abort on error is good for debugging, not so much from a coverage perspective.

satya-bodapati (changes requested):
1. Please fix GitHub description
2. Please add Jenkins link as a comment

nitendra-bhosle:
Jenkins : https://ps57.cd.percona.com/view/5.7/job/percona-server-5.7-param/446/

mysql-test/suite/innodb/t/xtradb_compressed_columns.test
george-lorch: Right, newer semantics, especially with built in mtr assert means that an explicit assert on single valued expressions is clearer in the test case and can not be washed away with a blind re-record.  So assert is preferred over recorded results for single value checks now.

satya-bodapati: if you want to cheat, add one more row to table :smiling_imp:

nitendra-bhosle: :) , I have added assert in test case

venkatesh-prasad-v (approved):
Thank you for addressing the review comments. The patch looks good and is approved from my side. But please wait for @satya-bodapati for the final approval.

mysql-test/suite/innodb/t/xtradb_compressed_columns.test
```diff
@@ -220,3 +220,20 @@ CREATE TABLE t1 (c1 VARBINARY(5) COLUMN_FORMAT COMPRESSED) ENGINE=InnoDB;
...
+UPDATE t1 SET ip_col=2680 WHERE ip_col = 23486;
+UPDATE t1 SET ip_col=23486 WHERE ip_col = 2680;
+
+--let $query_out=`SELECT COUNT(*) FROM t1 WHERE lt3='S6GaiUP5cMt7uS38D6B31KSJW0mkt7EQBu7sbK7G7LgYcQ3p6EE9HbTSGI9bMU4vQ8bJABIUbsjp0hIDtwgZOpSkc9xHvYPCx99DK6IDkEqioZu4808QqFCrtcGJWFceOMVyfG8f2NOQT58WAYPKBq5jkuHOtHBSIXyA1DFvWhifC0u0xXCULBrd9EF0Te0l5NcaQZTvJeKjoDjpRA4zfR7Z0t04QOhwVcad2ZbBWOfFiKiEz5DzOKzImrRLmcOtmgbPpTdBU8lND8r2FKSn0VwT7aYyyJHHJwvTlo8qepdbi4zdN1GPUAKeRtPLECGUkqQwLHacNWB2ZNlSDg788CXMsUgYpzB7TNqCB897PmOOPse0N'`
```
satya-bodapati, line 234: count(*) doesn't verify contents. the garbage inserted can be still counted as 1. So please use content comparison instead of COUNT(*)

satya-bodapati (changes requested):

mysql-test/suite/innodb/t/xtradb_compressed_columns.test
nitendra-bhosle: Inside Where condition I have used actual content , so count(*) will return 0 if content is not same as what we have inserted earlier.

satya-bodapati (approved):

satya-bodapati (approved):

PR #4286, opened 2021-05-05
PS-7665 : performance_schema.metadata_locks m_column_name_length is u…

satya-bodapati:
…ninitialized for MDL_key::FOREIGN_KEY

https://jira.percona.com/browse/PS-7665

Problem:
--------
When a PFS query on performance_schema.metadata_locks is executed, server crashes randomly. This happens
when there are DDLs in progress on a table with foreign key constraints. i.e. MDL with MDL_key::FOREIGN_KEY
should be acquired and not yet released. At this point of time, if we execute the SELECT query on
performance_schema.metadata_locks, there is uninitialized memory access and this causes crashes randomly.

Fix:
----
PFS engine has temporary object(PFS_column_row) to store each row of a query.
Initialize m_column_length to 0. This column name is used only for MDL_key::COLUMN_STATISTICS
So rest all MDL key types should initialize to 0.

### DIFF
--- a/sql/sql_table.cc
+++ b/sql/sql_table.cc
@@ -16588,6 +16588,8 @@ bool mysql_alter_table(THD *thd, const char *new_db, const char *new_name,
                                        thd->variables.lock_wait_timeout))
       return true;
 
+    DEBUG_SYNC_C("after_mdl_locks_acquired");
+
     /*
       If we are executing ALTER TABLE RENAME under LOCK TABLES we also need
       to check that all previously orphan tables which reference new table
--- a/sql/sql_table.cc
+++ b/sql/sql_table.cc
@@ -16588,7 +16588,7 @@ bool mysql_alter_table(THD *thd, const char *new_db, const char *new_name,
                                        thd->variables.lock_wait_timeout))
       return true;
 
-    DEBUG_SYNC_C("after_mdl_locks_acquired");
+    DEBUG_SYNC(thd, "after_mdl_locks_acquired");
 
     /*
       If we are executing ALTER TABLE RENAME under LOCK TABLES we also need
--- a/storage/perfschema/table_helper.cc
+++ b/storage/perfschema/table_helper.cc
@@ -805,11 +805,13 @@ int PFS_column_row::make_row(const MDL_key *mdl) {
       m_object_type = OBJECT_TYPE_FOREIGN_KEY;
       m_schema_name_length = mdl->db_name_length();
       m_object_name_length = mdl->name_length();
+      m_column_name_length = 0;
       break;
     case MDL_key::CHECK_CONSTRAINT:
       m_object_type = OBJECT_TYPE_CHECK_CONSTRAINT;
       m_schema_name_length = mdl->db_name_length();
       m_object_name_length = mdl->name_length();
+      m_column_name_length = 0;
       break;
     case MDL_key::NAMESPACE_END:
     default:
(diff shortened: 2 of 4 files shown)


satya-bodapati:
https://ps80.cd.percona.com/job/percona-server-8.0-valgrind-param/10/

sql/sql_table.cc
```diff
@@ -16588,6 +16588,8 @@ bool mysql_alter_table(THD *thd, const char *new_db, const char *new_name,
                                        thd->variables.lock_wait_timeout))
       return true;
 
+    DEBUG_SYNC_C("after_mdl_locks_acquired");
```
venkatesh-prasad-v: IIUC, `DEBUG_SYNC_C` uses `current_thd` which in turn uses `pthread_getspecific()` and may make things slower. Please use `DEBUG_SYNC(thd,"after_mdl_locks_acquired")` since we have thd here.

storage/perfschema/table_helper.cc
```diff
@@ -805,11 +805,13 @@ int PFS_column_row::make_row(const MDL_key *mdl) {
       m_object_type = OBJECT_TYPE_FOREIGN_KEY;
       m_schema_name_length = mdl->db_name_length();
       m_object_name_length = mdl->name_length();
+      m_column_name_length = 0;
```
venkatesh-prasad-v: I saw your question on upstream bug
>What if there is new MDL type added and m_column_length is not set to zero?

For this, wouldn't it be better if we do either of these
1. `memset(&object_row, 0, sizeof(object_row));` in `PFS_index_metadata_locks_by_object::match()`
OR
2. Set `m_column_name_length = 0;` before the `switch` statement in `PFS_column_row::make_row()` function (Line no. 700 in same file)?

satya-bodapati: Implemented 2.

satya-bodapati:
./mtr --mem --suite=perfschema --parallel=4 --retry-failure=0 --force|tee mtr.log
Logging: /home/satya/WORK/PS-8.0-trunk/mysql-test/mysql-test-run.pl  --mem --suite=perfschema --parallel=4 --retry-failure=0 --force
MySQL Version 8.0.23
Checking supported features

[ 99%] perfschema.idx_compare_session_variables w3  [ pass ]   4443
[ 99%] perfschema.idx_compare_setup_actors      w1  [ pass ]   4782
[ 99%] perfschema.idx_compare_prepared_statements_instances w2  [ pass ]  34297
[ 99%] perfschema.error_log                     w2  [ pass ]  52969
[100%] shutdown_report                          w0  [ pass ]       
The servers were restarted 128 times
Spent 1463.722 of 752 seconds executing testcases

Completed: All 710 tests were successful.

7 tests were skipped, 5 by the test itself.

-------
./mtr --mem perfschema.fk  --valgrind

Logging: /home/satya/WORK/PS-8.0-trunk/mysql-test/mysql-test-run.pl  --mem perfschema.fk --valgrind
MySQL Version 8.0.23
Turning on valgrind for all executables
Running valgrind with options " --tool=memcheck --num-callers=16 --show-reachable=yes --suppressions=/home/satya/WORK/PS-8.0-trunk/mysql-test/valgrind.supp "
Turning off --check-testcases to save time when valgrinding
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Checking leftover processes
Removing old var directory
Creating var directory '/home/satya/WORK/PS-8.0-trunk/bld/mysql-test/var'
 - symlinking 'var' to '/dev/shm/var_auto_68Ah'
Installing system database
Using parallel: 1

[ 33%] perfschema.fk                             [ pass ]  19974
[ 66%] shutdown_report                           [ pass ]       
[100%] valgrind_report                           [ pass ]       
Spent 19.974 of 141 seconds executing testcases

venkatesh-prasad-v (approved):
Approved from my side. Please squash the commits before merging.

PR #4300, opened 2021-05-12
PS-1116 : LP #1719506: Audit plugin reports "command_class=error" for…

nitendra-bhosle:
… server-side prepared statements

https://jira.percona.com/browse/PS-1116

Analysis
--------
server-side prepared statements sets
thd->lex->sql_command to SQLCOM_END , which results in
COMMAND_CLASS to set as 'error'

Fix
---
Inside function mysql_audit_notify we have
set COMMAND_CLASS to correct type based on msg type
This change is specifically made for audit logs only


sql/sql_audit.cc
```diff
@@ -451,7 +451,48 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
   event.general_host= sctx->host();
   event.general_external_user= sctx->external_user();
   event.general_rows= thd->get_stmt_da()->current_row_for_condition();
-  event.general_sql_command= sql_statement_names[thd->lex->sql_command];
+  if(thd->lex->sql_command == SQLCOM_END && msg_len > 0 )
+  {
+    if(!strcmp(msg,"Prepare"))
```
percona-ysorokin: There is `command_name` array declared in `sql_parse.h`. You can iterate over its elements and find the index which you then will be able to cast to `COM_XXX` constant. After that you can `switch` over this constant and fill `event.general_sql_command` for known ones
```
  COM_STMT_PREPARE,
  COM_STMT_EXECUTE,
  COM_STMT_SEND_LONG_DATA,
  COM_STMT_CLOSE,
  COM_STMT_RESET,
  COM_STMT_FETCH,
  ...
```
with exactly the same `event.general_sql_command= sql_statement_names[<found_index>];`
This will help to remove `magic constants` from the code

sql/sql_audit.cc
```diff
@@ -451,7 +451,48 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+    {
+      event.general_sql_command.str= "stmt_prepare";
```
percona-ysorokin: BTW, here, you could do `event.general_sql_command = C_STRING_WITH_LEN("...")`

percona-ysorokin (changes requested):

altmannmarcelo (changes requested):
Hi @nitendra-bhosle . The fix itself looks good. I have a few questions / requests:

- Where did you get the list of missing statements (stmt_send_long_data, stmt_fetch, ...). The reason for asking is, perhaps there is more statements
- Our documentation mention that the list of commands come from `SELECT name FROM performance_schema.setup_instruments WHERE name LIKE "statement/sql/%" ORDER BY name;`  and the ones you added are from `select * from setup_instruments where name like '%statement/com%';` - https://www.percona.com/doc/percona-server/8.0/management/audit_log_plugin.html#filtering-by-sql-command-type - Documentation needs to be updated to reflect those changes.
- We need test cases showing those commands will be logged with the desired command class
- Test cases must ensure we can filter in/out the new classes

sql/sql_audit.cc
nitendra-bhosle: I had tried with C_STRING_WITH_LEN("...") ,But I was getting error " error: extended initializer lists only available with ‘-std=c++11’ or ‘-std=gnu++11’ [-Werror]" 
This is warning and warnings are treated as error , so , it's not building with  this change.

May be because of this sql_statement_names is built using 'str' and 'len' members explicitly : 
./sql/mysqld.cc:      sql_statement_names[com_index].str= var->name;
./sql/mysqld.cc:      sql_statement_names[com_index].length= strlen(var->name);
 and 

event.general_command.str= msg;
event.general_command.length= msg_len;

satya-bodapati: Please try like this:  = { C_STRING_WITH_LEN("Host") }. (With curly brackets)

nitendra-bhosle: I was trying with curly brackets only.
event.general_sql_command = { C_STRING_WITH_LEN("Host") };

is not compiling .

nitendra-bhosle: I think inside makefile , CXX flags must be set as : -std=c++03

nitendra-bhosle: I have verified this with test program 
g++ test.cpp -o test -std=c++03
test.cpp: In function ‘int main()’:
test.cpp:19:33: warning: extended initializer lists only available with ‘-std=c++11’ or ‘-std=gnu++11’
   19 |  t.test = {"test",strlen("test")};

g++ test.cpp -o test -std=c++11 :  No warning

satya-bodapati: How does it work in other places. It is used a lot in 5.7 codebase. Something else is the reason? :thinking:

nitendra-bhosle: Inside file : sql_audit.cc , I am not able to use any of c++ 11 features.I am not sure why

satya-bodapati: AFAIR, c+11 is from 8.0 code base only. C_STRING_WITH_LEN has nothing do with c+11 feature? :thinking:

satya-bodapati: also compare compilation flags of sql_audit.cc with other source files? (use VERBOSE=1 make)

nitendra-bhosle: > AFAIR, c+11 is from 8.0 code base only. C_STRING_WITH_LEN has nothing do with c+11 feature?

Yeah right, I was trying to use auto and lamda's

nitendra-bhosle: > also compare compilation flags of sql_audit.cc with other source files? (use VERBOSE=1 make)

With VERBOSE=1 , I can see sql_audi.cc is compiling with -std=gnu++03 option

satya-bodapati: ok, did you avoid lambdas, autos and C+11 features in 5.7? do you see still same error?

nitendra-bhosle: I have figured out issue , with -std=c++03
You can initialise a struct with extended initializer lists , but you can't assign .
That means event.general_sql_command = { C_STRING_WITH_LEN(msg)} ; doesn't work 
But, MYSQL_LEX_CSTRING command_class = { C_STRING_WITH_LEN(msg)} ; works.

plugin/audit_log/tests/mtr/prepare_stmt_audit_log.py
```diff
@@ -0,0 +1,54 @@
...
+
+	if py_version == 3:
+		cnx = mysql.connector.connect(database=db, user=user,password=p,port=port)
+		cursor = cnx.cursor(prepared=True)
+		stmt = "SHOW GLOBAL VARIABLES WHERE Variable_name = ?"
+		cursor.execute(stmt, ("max_allowed_packet",))
+	else:
+		cnx = MySQLdb.connect(user=user, host=host, port=port, db=db,password=p)
+		cursor = cnx.cursor()
+		#cursor.execute("SHOW GLOBAL VARIABLES WHERE Variable_name = %s",("max_allowed_packet",))
```
satya-bodapati: I think this should be uncommented?

plugin/audit_log/tests/mtr/prepare_stmt_audit_log.py
```diff
@@ -0,0 +1,54 @@
...
+		cursor.execute("select 1")
```
satya-bodapati: remove this

plugin/audit_log/tests/mtr/prepare_stmt_audit_log.py
```diff
@@ -0,0 +1,54 @@
...
+	host = sys.argv[2]
+	port = int(sys.argv[3])
+	db = sys.argv[4]
+	p = sys.argv[5]
+	cursor = None
+
+	if py_version == 3:
+		cnx = mysql.connector.connect(database=db, user=user,password=p,port=port)
+		cursor = cnx.cursor(prepared=True)
+		stmt = "SHOW GLOBAL VARIABLES WHERE Variable_name = ?"
```
satya-bodapati: cant we use 
PREPARE stmt FROM SHOW GLOBAL VARIABLES WHERE Variable_name = ?;
SET @var2='blah';
EXECUTE insert_stmt USING @var2;

Do we need this python script ?

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
   event.general_host= sctx->host();
   event.general_external_user= sctx->external_user();
   event.general_rows= thd->get_stmt_da()->current_row_for_condition();
-  event.general_sql_command= sql_statement_names[thd->lex->sql_command];
+  if(thd->lex->sql_command == SQLCOM_END && msg_len > 0 )
```
satya-bodapati:
```suggestion
  if (thd->lex->sql_command == SQLCOM_END && msg_len > 0)
```

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+  {
+    int found_index = -1;
+    for(size_t i=0; i< get_command_name_len();++i)
```
satya-bodapati:
```suggestion
    for (size_t i=0; i < get_command_name_len(); ++i)
```

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+    {
+      if(!strcmp(command_name[i].str,msg))
```
satya-bodapati:
```suggestion
      if (!strcmp(command_name[i].str, msg))
```

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+      {
+        found_index = i;
+        break;
+      }
+    }
+
+    if(found_index == -1)
```
satya-bodapati:
```suggestion
    if (found_index == -1)
```

sql/sql_parse.cc
```diff
@@ -179,6 +179,11 @@ const LEX_STRING command_name[] MY_ATTRIBUTE((unused)) = {
   { C_STRING_WITH_LEN("Error") }  // Last command number
 };
 
+size_t get_command_name_len(void)
+{
+  return sizeof(command_name)/sizeof(command_name[0]);
```
satya-bodapati:
```suggestion
  return sizeof(command_name) / sizeof(command_name[0]);
```

satya-bodapati (changes requested):

plugin/audit_log/tests/mtr/prepare_stmt_audit_log.py
nitendra-bhosle: These statements will generate Audit logs with name "Query" not name "Prepare".
But MySQLdb python driver used for python2 in mtr tests also generates "Query" logs .
So , I will remove python script and use these statements.

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
   event.general_host= sctx->host();
   event.general_external_user= sctx->external_user();
   event.general_rows= thd->get_stmt_da()->current_row_for_condition();
-  event.general_sql_command= sql_statement_names[thd->lex->sql_command];
+  if (thd->lex->sql_command == SQLCOM_END && msg_len > 0)
+  {
+    int found_index = -1;
```
satya-bodapati:
```suggestion
    int found_index= -1;
```

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+    for (size_t i=0; i < get_command_name_len(); ++i)
+    {
+      if (!strcmp(command_name[i].str, msg))
+      {
+        found_index = i;
```
satya-bodapati:
```suggestion
        found_index= i;
```

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
   event.general_host= sctx->host();
   event.general_external_user= sctx->external_user();
   event.general_rows= thd->get_stmt_da()->current_row_for_condition();
-  event.general_sql_command= sql_statement_names[thd->lex->sql_command];
+  if (thd->lex->sql_command == SQLCOM_END && msg_len > 0)
+  {
+    int found_index = -1;
+    for (size_t i=0; i < get_command_name_len(); ++i)
```
satya-bodapati:
```suggestion
    for (size_t i= 0; i < get_command_name_len(); ++i)
```

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+    {
+      if (!strcmp(command_name[i].str, msg))
+      {
+        found_index = i;
+        break;
+      }
+    }
+
+    if (found_index == -1)
+      event.general_sql_command = sql_statement_names[thd->lex->sql_command];
```
satya-bodapati:
```suggestion
      event.general_sql_command= sql_statement_names[thd->lex->sql_command];
```

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+    else
+    {
+      switch (found_index)
+      {
+      case COM_STMT_PREPARE:
+        event.general_sql_command = sql_statement_names[SQLCOM_PREPARE];
```
satya-bodapati:
```suggestion
        event.general_sql_command= sql_statement_names[SQLCOM_PREPARE];
```

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+        break;
+      case COM_STMT_EXECUTE:
+        event.general_sql_command = sql_statement_names[SQLCOM_EXECUTE];
```
satya-bodapati:
```suggestion
        event.general_sql_command= sql_statement_names[SQLCOM_EXECUTE];
```

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+        break;
+      case COM_STMT_RESET:
+        event.general_sql_command = sql_statement_names[SQLCOM_RESET];
```
satya-bodapati:
```suggestion
        event.general_sql_command= sql_statement_names[SQLCOM_RESET];
```

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+        break;
+      default:
+        MYSQL_LEX_CSTRING command_class = { C_STRING_WITH_LEN(msg)} ;
```
satya-bodapati:
```suggestion
        MYSQL_LEX_CSTRING command_class= { C_STRING_WITH_LEN(msg)} ;
```

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+        event.general_sql_command = command_class;
```
satya-bodapati:
```suggestion
        event.general_sql_command= command_class;
```

satya-bodapati (changes requested):
5.7 style is a bit different. Rest all looks good.

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+        event.general_sql_command= sql_statement_names[SQLCOM_PREPARE];
+        break;
+      case COM_STMT_EXECUTE:
+        event.general_sql_command= sql_statement_names[SQLCOM_EXECUTE];
+        break;
+      case COM_STMT_RESET:
+        event.general_sql_command= sql_statement_names[SQLCOM_RESET];
+        break;
+      default:
+        MYSQL_LEX_CSTRING command_class= { C_STRING_WITH_LEN(msg)} ;
```
satya-bodapati:
```suggestion
        MYSQL_LEX_CSTRING command_class= { C_STRING_WITH_LEN(msg) } ;
```

satya-bodapati (approved):
LGTM! one format change requested.

satya-bodapati (approved):
typo in commit msg: "chnage"

Please wait for @altmannmarcelo's approval before merging

plugin/audit_log/tests/mtr/audit_log_charset.result
```diff
@@ -67,5 +67,5 @@ set global audit_log_flush= ON;
 "Query","<ID>","<DATETIME>","drop_db","<CONN_ID>",0,"DROP DATABASE `???`","root[root] @ localhost []","localhost","","","ฐานข้อมูล"
 "Query","<ID>","<DATETIME>","change_db","<CONN_ID>",0,"use test","root[root] @ localhost []","localhost","","","test"
 "Query","<ID>","<DATETIME>","set_option","<CONN_ID>",0,"SET @@character_set_client=cp1256","root[root] @ localhost []","localhost","","","test"
-"Query","<ID>","<DATETIME>","error","<CONN_ID>",1064,"CREATE t \217\355ف?\355ف?\355\335(\217\260\241\217\260\241\217\260\241 char) DEFAULT CHARSET=ujis engine=TokuDB","root[root] @ localhost []","localhost","","","test"
+"Query","<ID>","<DATETIME>","Query","<CONN_ID>",1064,"CREATE t \217\355ف?\355ف?\355\335(\217\260\241\217\260\241\217\260\241 char) DEFAULT CHARSET=ujis engine=TokuDB","root[root] @ localhost []","localhost","","","test"
```
altmannmarcelo: Why does this changed from `error` to `Query` ?

From https://github.com/percona/percona-server/blob/5.7/plugin/audit_log/tests/mtr/audit_log_charset.test#L64 we are expecting a Parsing error, and those should be reported as error on audit log.

plugin/audit_log/tests/mtr/audit_log_filter_commands.result
```diff
@@ -169,7 +169,7 @@ END","root[root] @ localhost []","localhost","","","db1"
 "Query","<ID>","<DATETIME>","change_db","<CONN_ID>",0,"USE test","root[root] @ localhost []","localhost","","","test"
 "Query","<ID>","<DATETIME>","drop_table","<CONN_ID>",0,"DROP TABLE db1.t1","root[root] @ localhost []","localhost","","","test"
 "Query","<ID>","<DATETIME>","drop_db","<CONN_ID>",0,"DROP DATABASE db1","root[root] @ localhost []","localhost","","","test"
-"Ping","<ID>","<DATETIME>","error","<CONN_ID>",0,"","root[root] @ localhost []","localhost","","","test"
+"Ping","<ID>","<DATETIME>","Ping","<CONN_ID>",0,"","root[root] @ localhost []","localhost","","","test"
```
altmannmarcelo: seems like we use lower case for all command classes, and here we are now adding the command classes with the initial upper case.

plugin/audit_log/tests/mtr/audit_log_filter_prepare_stmt.test
```diff
@@ -0,0 +1,37 @@
...
+
+let $MYSQLD_DATADIR= `select @@datadir`;
+let $log_file= $MYSQLD_DATADIR/test_audit.log;
+
+SET GLOBAL audit_log_flush=ON;
+--remove_file $log_file
+SET GLOBAL audit_log_flush=ON;
+
+#Filter by excluding select command
+SET GLOBAL audit_log_exclude_commands= 'set_option,prepare_sql,execute_sql';
```
altmannmarcelo, line 17: I see you covered the case from https://jira.percona.com/browse/PS-1116. We also have https://jira.percona.com/browse/PS-3530 linked to to it.

If a user runs
1. SELECT field FROM table;
2. PREPARE xxx FROM 'SELECT field FROM table'
3. EXECUTE xxx;

Does item 3 will show up as `execute_sql` command class or `select`?
If I filter by excluding SELECTs, will item 3 show up in audit logs?

sql/sql_audit.cc
```diff
@@ -451,7 +452,42 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
   event.general_host= sctx->host();
   event.general_external_user= sctx->external_user();
   event.general_rows= thd->get_stmt_da()->current_row_for_condition();
-  event.general_sql_command= sql_statement_names[thd->lex->sql_command];
+  if (thd->lex->sql_command == SQLCOM_END && msg_len > 0)
+  {
+    int found_index= -1;
+    for (size_t i= 0; i < get_command_name_len(); ++i)
+    {
+      if (!strcmp(command_name[i].str, msg))
```
altmannmarcelo, line 461: just to be on the safe side, any performance issue could arise here by doing a string comparison inside a loop if the workload is mostly done via prepared statements?

altmannmarcelo (changes requested):
@nitendra-bhosle thanks for working on the test cases. A few comments that are still unclear.

plugin/audit_log/tests/mtr/audit_log_filter_commands.result
nitendra-bhosle: These command classes are from query : select * from performance_schema.setup_instruments where name like '%statement/com%';
Which are having initial letter uppercase

plugin/audit_log/tests/mtr/audit_log_filter_prepare_stmt.test
nitendra-bhosle: Item 3 will show up as 'execute_sql' in audit log ,if you exclude by SELECTS's

sql/sql_audit.cc
nitendra-bhosle: Command classes generated based on 'msg' variable , hence strcmp is used , If we want to avoid this, we need to make changes to sql_parse.cc file

plugin/audit_log/tests/mtr/audit_log_filter_prepare_stmt.test
altmannmarcelo: This would be a FR, not the context of this bug. Marking as solved.

plugin/audit_log/tests/mtr/audit_log_filter_commands.result
altmannmarcelo: Make it lower case to be consistent.

sql/sql_audit.cc
altmannmarcelo: I'm just wondering that this could be a performance bummer for a workload that is 100% PS (some ORM do this. PHP also heavily uses PS for anti SQL injection).
Is it possible for you to run a quick test with and without your bug, having a workload that is PS (sysbench has it disable by default, but you can enable it https://github.com/akopytov/sysbench/issues/310 )

nitendra-bhosle: I have run sysbench with PS , I didn't notice any performance degradation. 
I have used these commands : 
sysbench --table-size=1000000 --db-ps-mode='auto' --mysql-db=test --mysql-host=127.0.01.1 --mysql-port=3306 --mysql-user=root --mysql-password=<password> /usr/share/sysbench/oltp_read_write.lua prepare
sysbench --table-size=1000000 --db-ps-mode='auto' --mysql-db=test --mysql-host=127.0.01.1 --mysql-port=3306 --mysql-user=root --mysql-password=<password> /usr/share/sysbench/oltp_read_write.lua run

These are results : 
1. With My patch included : 

Running the test with following options:
Number of threads: 1
Initializing random number generator from current time


Initializing worker threads...

Threads started!

SQL statistics:
    queries performed:
        read:                            2940
        write:                           840
        other:                           420
        total:                           4200
    transactions:                        210    (20.90 per sec.)
    queries:                             4200   (418.01 per sec.)
    ignored errors:                      0      (0.00 per sec.)
    reconnects:                          0      (0.00 per sec.)

General statistics:
    total time:                          10.0463s
    total number of events:              210

Latency (ms):
         min:                                   37.48
         avg:                                   47.83
         max:                                  106.89
         95th percentile:                       65.65
         sum:                                10044.63

Threads fairness:
    events (avg/stddev):           210.0000/0.00
    execution time (avg/stddev):   10.0446/0.0

2. Without my patch : 
Running the test with following options:
Number of threads: 1
Initializing random number generator from current time


Initializing worker threads...

Threads started!

SQL statistics:
    queries performed:
        read:                            2926
        write:                           836
        other:                           418
        total:                           4180
    transactions:                        209    (20.81 per sec.)
    queries:                             4180   (416.27 per sec.)
    ignored errors:                      0      (0.00 per sec.)
    reconnects:                          0      (0.00 per sec.)

General statistics:
    total time:                          10.0384s
    total number of events:              209

Latency (ms):
         min:                                   37.20
         avg:                                   48.02
         max:                                   87.09
         95th percentile:                       61.08
         sum:                                10036.33

Threads fairness:
    events (avg/stddev):           209.0000/0.00
    execution time (avg/stddev):   10.0363/0.00

altmannmarcelo: thanks for testing it

plugin/audit_log/tests/mtr/audit_log_echo.inc
```diff
@@ -35,7 +35,7 @@ perl;
     }
     $line =~ s/"([a-zA-Z_ ]*)","([0-9]+)_[0-9_ :T-]*","[0-9_ :A-Z-]*"/"$1","<ID>","<DATETIME>"/;
     $line =~ s/"(Connect|Quit|Change user)","<ID>","<DATETIME>","[0-9]+"/"$1","<ID>","<DATETIME>","<CONN_ID>"/;
-    $line =~ s/"([A-Za-z ]+)","<ID>","<DATETIME>","([a-z_]+)","[0-9]+"/"$1","<ID>","<DATETIME>","$2","<CONN_ID>"/;
+    $line =~ s/"([A-Za-z ]+)","<ID>","<DATETIME>","(Ping|init db|Query|[a-z_]+)","[0-9]+"/"$1","<ID>","<DATETIME>","$2","<CONN_ID>"/;
```
altmannmarcelo: are Ping and Query suppose to have the initial as upper case here?

altmannmarcelo (approved):
@nitendra-bhosle thanks for working on those. I have just the observation of the regexp. If required fix it and it's good to go from my side (no need to send for re-review).

plugin/audit_log/tests/mtr/audit_log_echo.inc
nitendra-bhosle: Ping and Query will not be required here , [a-z_]  will handle that , I will remove these two from here , will keep "init db" so that I don't have to keep random space

satya-bodapati (approved):

PR #4308, opened 2021-05-17
PS-7710: Azure Pipelines: clang-6 to clang-8 compilation with libstdc++-11

inikep:
Use libstdc++-9 instead of libstdc++-11 for clang-6 to clang-8.


percona-ysorokin (approved):
LGTM

PR #4318, opened 2021-05-21
PS-6802 : Configure fails with make-4.3 with CMake Error at storage/r…

george-lorch:
…ocksdb/CMakeLists.txt:152 (STRING)

- Incorporated suggest patch to bash script


george-lorch:
https://ps57.cd.percona.com/view/5.7/job/percona-server-5.7-param/456/

inikep (approved):
LGTM

PR #4325, opened 2021-05-25
PS-7526 : Unexpected quoting and dropping of comments in DROP TABLE c…

nitendra-bhosle:
…ommands

https://jira.percona.com/browse/PS-7526

Analysis
---------
Feature request to add binlog_ddl_query_log_events option
so that when it's set ON log drop table statements as it is
without comments being striped out and without adding quotes

Fix
----
Adde new option binlog_ddl_query_log_events , which is
by default OFF, and could be enabled by 'SET binlog_ddl_query_log_events = ON'
If enabled drop table ddl statements will be logged as it is
without striping of comments and adding quotes


nitendra-bhosle:
Jenkins Job : https://ps57.cd.percona.com/job/percona-server-5.7-param/459/

sql/sys_vars.cc
```diff
@@ -1232,6 +1232,12 @@ static Sys_var_mybool Sys_binlog_rows_query(
        SESSION_VAR(binlog_rows_query_log_events),
        CMD_LINE(OPT_ARG), DEFAULT(FALSE));
 
+static Sys_var_mybool Sys_binlog_ddl_query(
+       "binlog_ddl_query_log_events",
+       "Allow writing of DDl_query_log events into binary log.",
+       SESSION_VAR(binlog_ddl_query_log_events),
+       CMD_LINE(OPT_ARG), DEFAULT(FALSE));
+
```
satya-bodapati, line 1240: 1. Why stripping of comments is a problem? (from the user perspective)?
2. Why do we need a new option? Will the presence of comments create any backward compatibility issues?
3. Why fix only for DROP TABLE? what about ALTER, CREATE?
4. Aren't DMLs affected?

satya-bodapati (changes requested):

sql/sys_vars.cc
nitendra-bhosle: Please find below answers for questions : 
1. Stripping of comments is problem because Customer's application depends on metadata they include into queries.
2. We need option so that only subset of customers , who wish to change default behaviour of drop table could enable option , for rest of customers's no change , since this option is disabled by default, so no compatibility issue.
3. Fix for drop table only , because drop table only prepares queries to be logged in bin log dynamically , I will check again this for ALTER table. For create table there is no problem.
4. This is only for binlog logging.

satya-bodapati: I don't think this change warrants a new option. It just that, after the fix, comments are logged for DROP TABLE. Does it break any compatibility?

Unless, we are changing query behaviour (output change), disk format changes, we don't need an option.

If there are no compatibility issues, I suggest to get rid of the option.

satya-bodapati: also let me know your findings on ALTER TABLE comments

nitendra-bhosle: ALTER TABLE doesn't strip comments , logs statements as they are entered on mysql client.
New option is added to keep default behaviour of drop commands , Only in exceptional cases like 1 customer request , we are just providing a way to get away from default behaviour and keep comments to be bin-logged .
By default option is disabled , so no compatibility problem for anyone.

satya-bodapati: I understood the behaviour/impact of the option. My question is, if it is enabled by default, is there any problem for other users (not customer)? Will it break any application/format for others?

sql/sys_vars.cc
```diff
@@ -1232,6 +1232,12 @@ static Sys_var_mybool Sys_binlog_rows_query(
        SESSION_VAR(binlog_rows_query_log_events),
        CMD_LINE(OPT_ARG), DEFAULT(FALSE));
 
+static Sys_var_mybool Sys_binlog_ddl_query(
+       "binlog_ddl_query_log_events",
+       "Allow writing of DDl_query_log events into binary log.",
+       SESSION_VAR(binlog_ddl_query_log_events),
+       CMD_LINE(OPT_ARG), DEFAULT(FALSE));
```
satya-bodapati, line 1239: Side note:
The name of the option and the comments are misleading too.
We do log DDL events before this patch too. This option is specifically about comments of a DDL.

satya-bodapati (changes requested):

sql/sql_table.cc
```diff
@@ -2900,12 +2900,25 @@ int mysql_rm_table_no_locks(THD *thd, TABLE_LIST *tables, bool if_exists,
         built_query.append(" /* generated by server */");
         int error_code = (non_tmp_error ?
           (foreign_key_error ? ER_ROW_IS_REFERENCED : ER_BAD_TABLE_ERROR) : 0);
-        error |= thd->binlog_query(THD::STMT_QUERY_TYPE,
+
+        if(thd->variables.binlog_ddl_query_log_events)
```
venkatesh-prasad-v, line 2904: Few questions:
1. Why was it necessary to implement a query builder for only DROP TABLE instead of logging the query as-is?
2. Let's say we try to execute `DROP TABLE t1,t2` where t1 is a normal table and  t2 is a temporary table, Since t2 is a temp table, it would not have been replicated to slave in RBR. In such cases, if we log DROP TABLE t1,t2 as is, then wouldn't it cause replication failure on slave with an error that it could not find table t2.
3. What happens in such cases when we enable the variable and execute `DROP TEMPORARY normal_table, temp_table;` it in a cluster environment like GR/PXC.?

satya-bodapati: @venkatesh-prasad-v I honestly don't feel the necessity of an option. Extra comments in binlog wouldn't break any format changes or introduce disk format incompatibilities. Can you please confirm this?

nitendra-bhosle: I am open for removing option ,if everyone is fine with this.

nitendra-bhosle:
This feature is 8.0 only.

PR #4328, opened 2021-05-26
PS-7622 merge: Merge MySQL 8.0.24 (fix sys_vars.innodb_log_writer_thr…

ldonoso:
…eads_basic)

The thread running `log_resume_writer_threads` might get waiting
indefinitely for an ack if an unset followed by a set of
`global.innodb_log_writer_threads` occurrs.

The functions `log_write_notifier` and `log_flush_notifier` might not
notice the unset and set of the variable and consider there has
been no change but `log_resume_writer_threads` is waiting for ack.

The solution is checking if log_resume_writer_threads is waiting for an
ack although the `log.writer_threads_paused` is false and, in that case,
signal an ack.

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -5216,8 +5216,8 @@ int init_common_variables() {
     get corrupted if accesses with names of different case.
   */
   DBUG_PRINT("info", ("lower_case_table_names: %d", lower_case_table_names));
-  lower_case_file_system = test_if_case_insensitive(mysql_real_data_home);
-  if (!lower_case_table_names && lower_case_file_system == 1) {
+  lower_case_file_system = (test_if_case_insensitive(mysql_real_data_home) == 1);
+  if (!lower_case_table_names && lower_case_file_system) {
     if (lower_case_table_names_used) {
       LogErr(ERROR_LEVEL, ER_LOWER_CASE_TABLE_NAMES_CS_DD_ON_CI_FS_UNSUPPORTED);
       return 1;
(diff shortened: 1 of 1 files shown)


oleksandr-kachan (commented):
LGTM

storage/innobase/log/log0write.cc
```diff
@@ -2667,6 +2667,11 @@ void log_write_notifier(log_t *log_ptr) {
       log.write_notifier_resume_lsn.store(0, std::memory_order_release);
 
       log_write_notifier_mutex_enter(log);
+    } else if (UNIV_UNLIKELY(log.write_notifier_resume_lsn.load(
```
percona-ysorokin: @ldonoso Luis, It took me a few hours to understand the race conditions here :)
However, I am not sure this change is 100%. safe.
Something tells me that the proper fix would be to add something similar to what Oracle did with the `while` construct in the `log_resume_writer_threads ()` function
https://github.com/mysql/mysql-server/blob/98b2ccb470de120d36bc4a623c814cdfded958ec/storage/innobase/log/log0log.cc#L909
to the end of `log_pause_writer_threads()`.
Basically,  we need to guarantee that at the end of `log_pause_writer_threads()` both `log_write_notifier` and `log_flush_notifier` threads are waiting on `writer_threads_resume_event` here
https://github.com/mysql/mysql-server/blob/98b2ccb470de120d36bc4a623c814cdfded958ec/storage/innobase/log/log0write.cc#L2582
and here
https://github.com/mysql/mysql-server/blob/98b2ccb470de120d36bc4a623c814cdfded958ec/storage/innobase/log/log0write.cc#L2700

PR #4329, opened 2021-05-27
PS-7722: Multiple-Column Index using Column Prefix Key Parts fails with Index Condition Pushdown in MyRocks

inikep:
1. Fix the issue using code from https://github.com/facebook/mysql-5.6/commit/48dfd8c710b
2. Add MTR test case to cover the issue.


inikep:
Tested at https://ps80.cd.percona.com/job/percona-server-8.0-param/937/

george-lorch (approved):

PR #4331, opened 2021-05-28
PS-7645 : Incorrect strncpy usage

kennt-percona:
Issue
Manual code review of strncpy() usage in the MySQL source tree
revealed incorrect usage.  There are instances where the destination
buffer was not being NULL-terminated.

Solution
Ensure that the buffer is being NULL-terminated in the case where
the entire buffer is being filled.


george-lorch:
clang-format issues https://app.circleci.com/pipelines/github/percona/percona-server/4568/workflows/64f60ed5-aa6d-4bda-a1c8-9a1f285a8190/jobs/4550

PR #4332, opened 2021-05-28
PS-7622 - Revert PS-7688

ldonoso:
The 8.0 version is not affected just 5.7.


percona-ysorokin (approved):
LGTM
please update the commit message as discussed on slack

PR #4347, opened 2021-06-09
PS-1116 : LP #1719506: Audit plugin reports "command_class=error" for server-side prepared statements

nitendra-bhosle:
PS-1116 : LP #1719506: Audit plugin reports "command_class=error" for server-side prepared statements

https://jira.percona.com/browse/PS-1116

Analysis
--------
server-side prepared statements sets
thd->lex->sql_command to SQLCOM_END , which results in
COMMAND_CLASS to set as 'error'

Fix
---
Inside function mysql_audit_notify we have
set COMMAND_CLASS to correct type based on msg type
This change is specifically made for audit logs only


nitendra-bhosle:
Jenkins : https://ps80.cd.percona.com/job/percona-server-8.0-param/969/

altmannmarcelo:
@nitendra-bhosle LGTM, please just adjust the PR description (same as 5.7)

altmannmarcelo (approved):

satya-bodapati (approved):
Please resubmit to Jenkins again. 948 failed..

nitendra-bhosle:
Jenkins : https://ps80.cd.percona.com/job/percona-server-8.0-param/953/

satya-bodapati (changes requested):
I see lot of MySQL.audit_log.audit_log_filter_commands test failures. Logs aren't available for some reason.

Can you please check if u can find the reason for test failure? if not resubmit just with audit_log suite?

nitendra-bhosle:
> I see lot of MySQL.audit_log.audit_log_filter_commands test failures. Logs aren't available for some reason.
> 
> Can you please check if u can find the reason for test failure? if not resubmit just with audit_log suite?

I was trying to figure out , problem with Jenkins , Finally got what is failing .
On some platforms say 'debian:buster' code which I had written to convert a c string in lowercase wasn't working
correctly , and hence coming 'command_class"  as empty string , and failing tests.
I was trying to reproduce same on my local ubuntu machine , everything is working correctly .
To solve this , I used map , which will have mapping between string's to their lowercase counterpart.

sql/sql_audit.cc
```diff
@@ -360,6 +362,58 @@ class Ignore_event_error_handler : public Audit_error_handler {
...
+
+const char *get_lower_case_cmd_class(const char *msg) {
+  if (command_class_mapping.empty()) {
+    command_class_mapping["Sleep"] = "sleep";
+    command_class_mapping["Quit"] = "quit";
+    command_class_mapping["Init DB"] = "init db";
+    command_class_mapping["Query"] = "query";
+    command_class_mapping["Field List"] = "field list";
+    command_class_mapping["Create DB"] = "create db";
+    command_class_mapping["Drop DB"] = "drop db";
```
satya-bodapati: I haven't checked the earlier code to convert to lower case. What were you using?

satya-bodapati (changes requested):

sql/sql_audit.cc
nitendra-bhosle: Previous code :  

```
std::string cmd_class = msg;
+          std::transform(cmd_class.begin(), cmd_class.end(), cmd_class.begin(),
+                         ::tolower);
+          MYSQL_LEX_CSTRING command_class = {
+              STRING_WITH_LEN(cmd_class.c_str())};

```

satya-bodapati: Hey @nitendra-bhosle ,
I think this is better, instead of the map and all this code.. Feels like bloatware..

In match_valid_donor_address():
```
 /* Host in MySQL is case insensitive and converted to lower case. */
    auto transform_lower = [](unsigned char c) {
      return static_cast<unsigned char>(std::tolower(c));
    };
    std::transform(valid_host.begin(), valid_host.end(), valid_host.begin(),
                   transform_lower);
```
so the issue might be something else? :thinking: Can you please check compiler version used on debian buster?
Or if the above piece of code works?

Lets see first what the compiler error message is? :thinking:

nitendra-bhosle: std::string cmd_class = msg;
This is declared locally , that means on stack.
We are changing cmd_class to uppercase , which is fine.

Then we are extracting c string out of this local string.
We are going out of scope and during stack unwinding memory of cmd_class is dealloacated  
and hence command_class is empty string.

I think this is problem.

We can solve this by multiple ways : 
1. have some other c-string which will be copy of 'msg' c-string , change this to lowercase and use it.
    But the problem with this approach is lifetime of memory address , which is created by copying , where should we 
    allocate and de-allocate. we can use global variable string , if it is ok.
2. Second is map , approach , which I have implemented.

I think map approach is better of these two.

satya-bodapati: I see the problem now. I thought it is about std::transform/lower not being available on platform.
Am still not convinced this is all required Just for case conversion. @altmannmarcelo do you see any problem with leaving case as is? or is this for consistency reasons?

1. How do you ensure new commands should be added to this map ? Will there be assertions etc?
2. How do you make this map strict read-only? How do u prevent concurrent modifications?

nitendra-bhosle: > 1. How do you ensure new commands should be added to this map ? Will there be assertions etc?

Currently no assertions provided , since this map is fallback option for cases when we don't get command_class from 'sql_statement_names'

> 2. How do you make this map strict read-only? How do u prevent concurrent modifications?

We can add use mutex here , but I think this may not be required.In previous patch we were assigning std:string cmd_class = msg; this statement itself is not atomic , so should be kept under mutex lock.

nitendra-bhosle: I agree , this all could be avoided , if we soften on requirement to make commands to lowercase.
@altmannmarcelo  : Please provide your inputs on this.

altmannmarcelo: If this is too complicated we can leave it as it is. However, at the moment, all command classes are lower case. We are adding new command classes that are capitalization which will be different from what we currently have. 

> std::string cmd_class = msg;
This is declared locally , that means on stack.
We are changing cmd_class to uppercase , which is fine.
Then we are extracting c string out of this local string.
We are going out of scope and during stack unwinding memory of cmd_class is dealloacated
and hence command_class is empty string.

Can't we pass cmd_class as `std::copy` instead or that will also go out of scope?

satya-bodapati: I think we need this to be alive until some upper layers? @nitendra-bhosle ?
typically THD heap or TABLE heap is used (but we have to be very careful about its lifecycle)

altmannmarcelo: can't this be force at the plugin level then. At the moment we are writting the string to file, we use strtolower(whatever_var_we_are_writing) ?

nitendra-bhosle: Yeah right , We need to manage lifecycle. I think std::copy will not help .

nitendra-bhosle: > can't this be force at the plugin level then. At the moment we are writting the string to file, we use strtolower(whatever_var_we_are_writing) ?

Problem is we are using 'msg'  const char* , If we modify it , it will reflect to all parent reference  of it.That means , if we modify it by any means ,even just for printing , whole c-string be lowercased.
Second Issue is at the time of logging , we simply can't just convert it to lowercase , because all the encoding and local's will be lost , which are applied to it.

altmannmarcelo: @nitendra-bhosle you mean that at https://github.com/percona/percona-server/blob/5.7/plugin/audit_log/audit_log.c#L691 we cannot modify the code in any way that will result of `event->general_sql_command.str` been written as lower case?
I'm not sure I follow what encoding and locals are you referring to here, we only want someplace in the stack that will have a lower version of this string and pass this memory address to snprintf. We won't be changing encoding nor its size which could result of unknown behavior. 

I'm sorry for been a bit hard here, but we have a lot of tools in the ecosystem that are built on top of the audit log and use regular expressions for filtering. If those are not properly configured to match using case insensitive things will start to fail. pt-query-digest, PMM are a few examples of our own tools that do it. Community users also have some other tools that rely on audit log.

satya-bodapati: if plugin can do it just before doing it, it would be OK. Encoding shouldn't be problem for at least command class which are all ascii? :thinking:

nitendra-bhosle: Let me try this.

altmannmarcelo (approved):
LGTM. Please do a final round of jenkis tests and make sure audit plugin-related tests pass on all platforms and also ASAN.

sql/sql_audit.cc
```diff
@@ -360,6 +362,8 @@ class Ignore_event_error_handler : public Audit_error_handler {
   const char *m_event_name;
 };
 
+std::string cmd_class_lowercase;
+
```
satya-bodapati: Global string object with concurrent threads modifying and accessing it is still not acceptable.

sql/sql_audit.cc
```diff
@@ -384,7 +388,41 @@ int mysql_audit_notify(THD *thd, mysql_event_general_subclass_t subclass,
...
+          break;
+        case COM_STMT_EXECUTE:
+          event.general_sql_command = sql_statement_names[SQLCOM_EXECUTE];
+          break;
+        case COM_STMT_RESET:
+          event.general_sql_command = sql_statement_names[SQLCOM_RESET];
+          break;
+        default:
+          cmd_class_lowercase = msg;
+          boost::to_lower(cmd_class_lowercase);
```
satya-bodapati: like user_buf object on stack here. why can't we use std::string local to function?

satya-bodapati: Just placing std::string at start of function is sufficient? please check

satya-bodapati: Also would prefer std library compared to boost (when possible)

satya-bodapati: Check how user_buff access via event.general_user.str is safe( or not safe?)

satya-bodapati (changes requested):

sql/sql_audit.cc
nitendra-bhosle: It's safe , Since logging to audit_log happens before cmd_class_lowercase goes out of scope.

satya-bodapati (approved):
LGTM!

sql/sql_audit.cc
satya-bodapati: Thanks for getting rid of this.

PR #4349, opened 2021-06-10
PS-7622 merge: Merge MySQL 8.0.24 (fix sys_vars.innodb_log_writer_thr…

ldonoso:
…eads_basic)

The thread running `log_resume_writer_threads` might get waiting
indefinitely for an ack if an unset followed by a set of
`global.innodb_log_writer_threads` occurrs.

The functions `log_write_notifier` and `log_flush_notifier` might not
notice the unset and set of the variable and consider there has
been no change but `log_resume_writer_threads` is waiting for ack.

The solution is checking if log_resume_writer_threads is waiting for an
ack although the `log.writer_threads_paused` is false and, in that case,
signal an ack.


percona-ysorokin (approved):
LGTM

PR #4351, opened 2021-06-14
PS-7582 - Check module data_masking loaded when calling its functions

ldonoso:
https://jira.percona.com/browse/PS-7582
https://ps57.cd.percona.com/view/5.7/job/percona-server-5.7-param/468/

---

**Bug:**

If a function of the data_masking module is called without the module
has been loaded, it generates an exception.

**Solution:**

As it is done with other modules, use a variable to check if
data_masking has been initialized. This condition is checked when the
function is invoked.

**Extra work:**

- Use `my_bool` instead of bool for `extern "C" functions`.

### DIFF
--- a/mysql-test/suite/data_masking/r/error_not_loaded_module.result
+++ b/mysql-test/suite/data_masking/r/error_not_loaded_module.result
@@ -1,3 +1,3 @@
 CREATE FUNCTION gen_blacklist RETURNS STRING SONAME 'data_masking.so';
 select gen_blacklist();
-ERROR HY000: Can't initialize function 'gen_blacklist'; This function requires data_masking plugin which is not installed. Please instal
+ERROR HY000: Can't initialize function 'gen_blacklist'; It requires the data_masking plugin installed. Please, install it and try again.
--- a/mysql-test/suite/data_masking/r/error_not_loaded_module.result
+++ b/mysql-test/suite/data_masking/r/error_not_loaded_module.result
@@ -1,0 +1,3 @@
+CREATE FUNCTION gen_blacklist RETURNS STRING SONAME 'data_masking.so';
+select gen_blacklist();
+ERROR HY000: Can't initialize function 'gen_blacklist'; This function requires data_masking plugin which is not installed. Please instal
--- a/plugin/data_masking/src/plugin.cc
+++ b/plugin/data_masking/src/plugin.cc
@@ -75,15 +75,14 @@ mysql_declare_plugin(data_masking){
     0,                               // flags
 } mysql_declare_plugin_end;
 
-// Returns true and populates the msg with an error if the module has not been initialized
-bool data_masking_validate_status(char *msg, size_t msg_len) {
+// Returns false and populates the msg with an error if the module has not been initialized
+bool data_masking_is_inited(char *msg, size_t msg_len) {
   if (!data_masking_init) {
-    strncpy(
+    std::snprintf(
         msg,
-        "It requires the data_masking plugin installed. Please, install it and try again.",
-        msg_len - 1);
-    msg[msg_len - 1]= '\0';
+        msg_len,
+        "It requires the data_masking plugin installed. Please, install it and try again.");
   }
 
-  return !data_masking_init;
+  return data_masking_init;
 }
--- a/plugin/data_masking/src/plugin.cc
+++ b/plugin/data_masking/src/plugin.cc
@@ -80,8 +80,7 @@ bool data_masking_validate_status(char *msg, size_t msg_len) {
   if (!data_masking_init) {
     strncpy(
         msg,
-        "This function requires data_masking plugin which is not installed."
-        " Please install keyring_udf plugin and try again.",
+        "It requires the data_masking plugin installed. Please, install it and try again.",
         msg_len - 1);
     msg[msg_len - 1]= '\0';
   }
--- a/plugin/data_masking/src/plugin.cc
+++ b/plugin/data_masking/src/plugin.cc
@@ -15,6 +15,8 @@
 
 #include "../include/plugin.h"
 
+static bool data_masking_init = false;
+
 static int data_masking_plugin_init(void *p) {
   DBUG_ENTER("data_masking_plugin_init");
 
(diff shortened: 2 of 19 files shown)


plugin/data_masking/src/udf/udf_gen_blacklist.cc
```diff
@@ -13,23 +13,28 @@
...
+#include <my_global.h>
 #include "../../include/plugin.h"
 #include "../../include/udf/udf_utils.h"
 #include "../../include/udf/udf_utils_string.h"
 
 #include <algorithm>
 
 extern "C" {
-  bool gen_blacklist_init(UDF_INIT *initid, UDF_ARGS *args, char *message);
+  my_bool gen_blacklist_init(UDF_INIT *initid, UDF_ARGS *args, char *message);
```
satya-bodapati, line 24: what happens with bool?
How did it work before? for example, std::snprintf() was used with bool ?

mysql-test/suite/data_masking/r/error_not_loaded_module.result
```diff
@@ -0,0 +1,3 @@
+CREATE FUNCTION gen_blacklist RETURNS STRING SONAME 'data_masking.so';
+select gen_blacklist();
+ERROR HY000: Can't initialize function 'gen_blacklist'; This function requires data_masking plugin which is not installed. Please instal
```
satya-bodapati: is the error message truncated by mtr framework?

satya-bodapati (changes requested):

plugin/data_masking/src/udf/udf_gen_blacklist.cc
ldonoso: Please, mind these functions are define with C interface. I changed because `bool` is not part of C and the other plugins use `my_bool`.

I guess it worked beforehand because `bool` and `my_bool` are compatible somewhow.

`DBG_RETURN` calls use a conversion:

> If the source type is bool, the value false is converted to zero and the value true is converted to the value one of the destination type (note that if the destination type is int, this is an integer promotion, not an integer conversion).

mysql-test/suite/data_masking/r/error_not_loaded_module.result
ldonoso: No, it is not. It is cut off by the framework in https://github.com/ldonoso/percona-server/blob/32278aa683149716fdc5119939899b08de1d0851/mysys/my_error.c#L213

It uses the following format:

```
(gdb) p format                                                                                                                                                                                                                                         
+p format                                                                                                                                                                                                                                              
$9 = 0x555558375abd "Can't initialize function '%-.192s'; %-.80s
```

So the effective msg is effectively 80 chars long.

I have reduced the error message so it fits the available space.

plugin/data_masking/src/udf/udf_gen_blacklist.cc
satya-bodapati: Ok, thanks @ldonoso I think this change is fine for 5.7 only.
For 8.0, Please keep the bool as is.

plugin/data_masking/src/plugin.cc
```diff
@@ -67,3 +74,17 @@ mysql_declare_plugin(data_masking){
...
+
+// Returns true and populates the msg with an error if the module has not been initialized
+bool data_masking_validate_status(char *msg, size_t msg_len) {
+  if (!data_masking_init) {
+    strncpy(
+        msg,
+        "This function requires data_masking plugin which is not installed."
+        " Please install keyring_udf plugin and try again.",
+        msg_len - 1);
+    msg[msg_len - 1]= '\0';
```
satya-bodapati: Maybe use std::snprintf? (I don't see any advantage but just for consistency reason)

plugin/data_masking/src/plugin.cc
```diff
@@ -67,3 +74,17 @@ mysql_declare_plugin(data_masking){
     NULL,                            // config options
     0,                               // flags
 } mysql_declare_plugin_end;
+
+// Returns true and populates the msg with an error if the module has not been initialized
+bool data_masking_validate_status(char *msg, size_t msg_len) {
```
satya-bodapati: rename this as "is_plugin_inited()"
return true if the plugin is inited, else false.

mysql-test/suite/data_masking/r/error_not_loaded_module.result
satya-bodapati: Looks good @ldonoso

plugin/data_masking/src/udf/udf_gen_dictionary_drop.cc
```diff
@@ -21,21 +21,26 @@
...
 
-bool gen_dictionary_drop_init(UDF_INIT *initid, UDF_ARGS *args,
+my_bool gen_dictionary_drop_init(UDF_INIT *initid, UDF_ARGS *args,
                                      char *message) {
-  DBUG_ENTER("gen_blacklist_init");
+  DBUG_ENTER("gen_dictionary_drop_init");
+
+  if (data_masking_validate_status(message, MYSQL_ERRMSG_SIZE)) {
+    DBUG_RETURN(true);
+  }
```
satya-bodapati: based on my suggestion, this would become

if (!is_plugin_inited(..)) {
}

satya-bodapati (changes requested):

plugin/data_masking/src/udf/udf_gen_blacklist.cc
ldonoso: Yes, I will do. Please, note @satya-bodapati that this commit is just intended for 5.7.

satya-bodapati: yes, luis.

plugin/data_masking/src/plugin.cc
satya-bodapati: on second thought, I feel this should be an optional review comment ;)

satya-bodapati: so upto you :)

satya-bodapati (approved):
@ldonoso Looks good

Please squash and merge

PR #4354, opened 2021-06-17
PS-7757 feature 8.0: Integrate ZenFS RocksDB plugin into Percona Server

percona-ysorokin:
https://jira.percona.com/browse/PS-7749


george-lorch:
@percona-ysorokin I see that you updated the Rocksdb submodule commit pointer.  We need to be very careful with this and ensure that it points exactly and only to a commit that has appeared within one of the MyRocks production tag merges, otherwise we have no guarantee of integration testing done between MyRocks<->RocksDB.

percona-ysorokin:
> @percona-ysorokin I see that you updated the Rocksdb submodule commit pointer. We need to be very careful with this and ensure that it points exactly and only to a commit that has appeared within one of the MyRocks production tag merges, otherwise we have no guarantee of integration testing done between MyRocks<->RocksDB.

@georgelorch This was done deliberately. RocksDB (from Facebok) introduced filesystem plugins (that allowed to use zenfs plugin) only in their version `v6.19.3`. Unfortunately, even the most recent Facebook production tag that we merged to Percona Server is `6.16.3` which is far behind what we need.
This is a quick solution that definitely does not expose all new vars but at least it seems to work with ZenFS.
As soon Facebook has a production tag based on at least `v6.19.3` and we merge it to Percona Server, this PR will be rebased.

PR #4356, opened 2021-06-22
PS-7746 - Manage deallocation of `ha_innobase::m_share` correctly

ldonoso:
https://ps57.cd.percona.com/view/5.7/job/percona-server-5.7-param/471/

---
When `free_share(m_share)` is called, we need to:

- Assign it the NULL value.
- Call `dict_table_close`

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -1319,6 +1319,17 @@ free_share(
 /*=======*/
 	INNOBASE_SHARE*	share);		/*!< in/own: share to free */
 
+
+/************************************************************************/ /**
+Calls free_share and assign NULL to share. */
+static void
+free_share_and_nullify(
+    INNOBASE_SHARE **share) /*!< in/own: table share to free */
+{
+	free_share(*share);
+	*share = NULL;
+}
+
 /*****************************************************************//**
 Frees a possible InnoDB trx object associated with the current THD.
 @return 0 or error number */
@@ -7030,7 +7041,7 @@ ha_innobase::open(
 		if (UNIV_UNLIKELY(m_share->ib_table &&
 				  m_share->ib_table->is_corrupt &&
 				  srv_pass_corrupt_table <= 1)) {
-			free_share(m_share);
+			free_share_and_nullify(&m_share);
 			dict_table_close(ib_table, FALSE, FALSE);
 			DBUG_RETURN(HA_ERR_CRASHED_ON_USAGE);
 		}
@@ -7056,7 +7067,7 @@ ha_innobase::open(
 		or force recovery can still use it, but not others. */
 		ib_table->set_file_unreadable();
 		ib_table->corrupted = true;
-		free_share(m_share);
+		free_share_and_nullify(&m_share);
 		dict_table_close(ib_table, FALSE, FALSE);
 		ib_table = NULL;
 		is_part = NULL;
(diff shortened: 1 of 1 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -7030,8 +7041,8 @@ ha_innobase::open(
 		if (UNIV_UNLIKELY(m_share->ib_table &&
 				  m_share->ib_table->is_corrupt &&
 				  srv_pass_corrupt_table <= 1)) {
-			free_share(m_share);
-
+			free_share_and_nullify(&m_share);
+			dict_table_close(ib_table, FALSE, FALSE);
```
venkatesh-prasad-v, line 7045: I think this should be
```suggestion
			dict_table_close(m_share->ib_table, FALSE, FALSE);
```

satya-bodapati: Consider the intrinsic table opening, I would say ib_table is fine

storage/innobase/handler/ha_innodb.cc
```diff
@@ -7066,6 +7066,7 @@ ha_innobase::open(
 			  srv_pass_corrupt_table <= 1)) {
 
 		free_share(m_share);
+		dict_table_close(ib_table, FALSE, FALSE);
```
satya-bodapati, line 7079: I think this is good idea to call dict_table_close() on error paths. We have increased the ref_count in open_table() and for intrinsic tables, we did a acquire() which increases ref_count.

dict_table_close() can handle intrinsic tables too. But we have to be careful in passing NULL to dict_table_close(). Some paths set ib_table to NULL. and following dict_table_close(ib_table) can cause crash.

Just analyze and ensure dic_table_close(NULL) doesn't happen.

ldonoso: Yes, I have double checked that `dict_table_close` is provided a not null `ib_table`.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1319,6 +1319,17 @@ free_share(
 /*=======*/
 	INNOBASE_SHARE*	share);		/*!< in/own: share to free */
 
+
+/************************************************************************/ /**
+Calls free_share and assign NULL to share. */
```
satya-bodapati: it is not necessary to use old decoration
you could write the new doxygen comment header in 5.7 too

```
/** Calls free_share and assign NULL to share
@param[in,out]    share  table share to free */
```

storage/innobase/handler/ha_innodb.cc
satya-bodapati: Great!

satya-bodapati (changes requested):

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1320,8 +1320,8 @@ free_share(
 	INNOBASE_SHARE*	share);		/*!< in/own: share to free */
 
 
-/************************************************************************/ /**
-Calls free_share and assign NULL to share. */
+/* Calls free_share and assign NULL to share
```
satya-bodapati: /**  Calls
(two stars)

satya-bodapati (approved):

satya-bodapati:
@ldonoso Please update PR with squashed commit.

venkatesh-prasad-v (approved):
Approved! Please don't forget to squash the commits before merging.

satya-bodapati (commented):
Please squash the commits luis for final approval

PR #4358, opened 2021-06-22
PS-7748: Implement buffered error logging ability

dutow:
This commits adds two new variables, buffered_error_log_filename and
buffered_error_log_size.

When buffered_error_log_size is specified and non zero, a buffer is
allocated with the specified size, and certain (debug) log messages are
written there.

When this buffer gets full, the server shuts down, or crashes, the
buffer is written to buffered_error_log_filename.

This is useful in situations where debug logs create lots of output and
slow down normal server operation, but are helpful for figuring out
issues, as writing to a memory buffer has less overhead.

### DIFF
--- a/mysql-test/suite/sys_vars/r/buffered_error_log_filename_basic.result
+++ b/mysql-test/suite/sys_vars/r/buffered_error_log_filename_basic.result
@@ -1,0 +1,20 @@
+SET @start_value = @@global.buffered_error_log_filename;
+SELECT @start_value;
+@start_value
+
+SET @@global.buffered_error_log_filename = "/tmp/bar.txt";
+SELECT @@global.buffered_error_log_filename;
+@@global.buffered_error_log_filename
+/tmp/bar.txt
+SET @@global.buffered_error_log_filename = '';
+SELECT @@global.buffered_error_log_filename;
+@@global.buffered_error_log_filename
+
+SET @@session.buffered_error_log_filename = OFF;
+ERROR HY000: Variable 'buffered_error_log_filename' is a GLOBAL variable and should be set with SET GLOBAL
+SELECT @@session.buffered_error_log_filename;
+ERROR HY000: Variable 'buffered_error_log_filename' is a GLOBAL variable
+SET @@global.buffered_error_log_filename = @start_value;
+SELECT @@global.buffered_error_log_filename;
+@@global.buffered_error_log_filename
+
--- a/mysql-test/suite/sys_vars/r/buffered_error_log_size_basic.result
+++ b/mysql-test/suite/sys_vars/r/buffered_error_log_size_basic.result
@@ -1,0 +1,31 @@
+SET @start_value = @@global.buffered_error_log_size;
+SELECT @start_value;
+@start_value
+0
+SET @@global.buffered_error_log_size = 420;
+SELECT @@global.buffered_error_log_size;
+@@global.buffered_error_log_size
+420
+SET @@global.buffered_error_log_size = 0;
+SELECT @@global.buffered_error_log_size;
+@@global.buffered_error_log_size
+0
+SET @@session.buffered_error_log_size = 0;
+ERROR HY000: Variable 'buffered_error_log_size' is a GLOBAL variable and should be set with SET GLOBAL
+SELECT @@session.buffered_error_log_size;
+ERROR HY000: Variable 'buffered_error_log_size' is a GLOBAL variable
+SET @@global.buffered_error_log_size = -10;
+Warnings:
+Warning	1292	Truncated incorrect buffered_error_log_size value: '-10'
+SELECT @@global.buffered_error_log_size;
+@@global.buffered_error_log_size
+0
+SET @@global.buffered_error_log_size = "/foo/bar.txt";
+ERROR 42000: Incorrect argument type to variable 'buffered_error_log_size'
+SELECT @@global.buffered_error_log_size;
+@@global.buffered_error_log_size
+0
+SET @@global.buffered_error_log_size = @start_value;
+SELECT @@global.buffered_error_log_size;
+@@global.buffered_error_log_size
+0
--- a/mysql-test/t/buffered_error_log.test
+++ b/mysql-test/t/buffered_error_log.test
@@ -1,0 +1,24 @@
+call mtr.add_suppression("Attempting backtrace. You can use the following information to");
+
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+
+--disable_query_log
+--let SEARCH_FILE= $MYSQLTEST_VARDIR/log/buffered_error_log.err
+--eval SET GLOBAL buffered_error_log_filename="$SEARCH_FILE"
+--enable_query_log
+
+# Send a kill -11 to mysqld, should cause a buffered_error_log log
+--let $_kill_signal = 11
+--source include/send_kill_to_mysqld.inc
+
+--let $restart_parameters=restart:
+--source include/start_mysqld.inc
+
+--echo Testing that buffered_error_log log contains something logged
+--let ABORT_ON = NOT_FOUND
+--let SEARCH_PATTERN=Server socket created on IP
+--source include/search_pattern_in_file.inc
+
+--let $restart_parameters=restart:
+--source include/start_mysqld.inc
--- a/mysys/buffered_error_log.cc
+++ b/mysys/buffered_error_log.cc
@@ -1,0 +1,69 @@
+
+#include "buffered_error_log.h"
+
+Buffered_error_logger buffered_error_log;
+char *buffered_error_log_filename = nullptr;
+
+void Buffered_error_logger::resize(std::size_t buffer_size) {
+  std::lock_guard<std::mutex> lk{data_mtx};
+
+  if (data.get() != nullptr && buffer_size == data->capacity()) {
+    return;
+  }
+
+  // Write out what we have currently, that way we don't have
+  // to deal with old data in the buffer
+  write_to_disk_();
+
+  if (buffer_size == 0) {
+    data.reset();
+    return;
+  }
+
+  data_t new_buffer(new std::string());
+  new_buffer->reserve(buffer_size);
+  data.swap(new_buffer);
+}
+
+Buffered_error_logger::~Buffered_error_logger() { write_to_disk(); }
+
+void Buffered_error_logger::log(const char *msg, size_t len) {
+  std::lock_guard<std::mutex> lk{data_mtx};
+  if (data.get() == nullptr || data->capacity() == 0) return;
+  const auto msg_end = data->size() + len;
+  if (msg_end > data->capacity() - 1) {
+    write_to_disk_();
+  }
+  *data += msg;
+  *data += '\n';
+}
+
+void Buffered_error_logger::write_to_disk() {
+  std::lock_guard<std::mutex> lk{data_mtx};
+  write_to_disk_();
+}
+
+bool Buffered_error_logger::is_enabled() {
+  std::lock_guard<std::mutex> lk{data_mtx};
+  return data.get() != nullptr && data->size() != 0;
+}
+
+void Buffered_error_logger::write_to_disk_() {
+  if (buffered_error_log_filename == nullptr ||
+      strlen(buffered_error_log_filename) == 0) {
+    return;
+  }
+  if (data.get() == nullptr || data->size() == 0) {
+    return;
+  }
+  auto fdd = fopen(buffered_error_log_filename, "a");
+  fwrite(data->data(), data->size(), 1, fdd);
+  fclose(fdd);
+
+  // the C++ standard doesn't guarantee that clear doesn't deallocate the
+  // buffer but it seems to be the case in libstdc++ and libc++ just to be on
+  // the safe side, we call reserve after clear
+  const auto curr_size = data->capacity();
+  data->clear();
+  data->reserve(curr_size);
+}
--- a/mysys/buffered_error_log.h
+++ b/mysys/buffered_error_log.h
@@ -1,0 +1,43 @@
+
+#pragma once
+
+#include <stdio.h>
+#include <cstring>
+#include <memory>
+#include <mutex>
+#include <vector>
+
+extern char *buffered_error_log_filename;
+
+/** Stores log messages in a fixed size buffer, which can be dumped by large
+ * chunks instead of line by line, to increase performance for high throughput
+ * scenarios.
+ *
+ * The buffer is also dumped by the crash reporting code and during shutdown to
+ * ensure that nothing is lost, only delayed.
+ */
+class Buffered_error_logger {
+  // We are using a pointer to string to guarantee that we can decrease the size
+  // of the buffer
+  using data_t = std::unique_ptr<std::string>;
+  // memory buffer storing the log messages
+  std::mutex data_mtx;
+  data_t data;
+
+ public:
+  void resize(std::size_t buffer_size);
+
+  ~Buffered_error_logger();
+
+  void log(const char *msg, size_t len);
+
+  void write_to_disk();
+
+  bool is_enabled();
+
+ private:
+  // requires holding data_mtx
+  void write_to_disk_();
+};
+
+extern Buffered_error_logger buffered_error_log;
--- a/sql/server_component/log_sink_trad.cc
+++ b/sql/server_component/log_sink_trad.cc
@@ -23,6 +23,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA */
 #include "log_sink_trad.h"
 #include "log_sink_perfschema.h"  // log_sink_pfs_event
 #include "my_systime.h"           // my_micro_time()
+#include "mysys/buffered_error_log.h"
 #include "sql/log.h"  // log_write_errstream(), log_prio_from_label()
 
 extern int log_item_inconsistent(log_item *li);
@@ -387,8 +388,23 @@ int log_sink_trad(void *instance [[maybe_unused]], log_line *ll) {
         log_sink_trad_last = microtime;
       }
 
+      bool log_to_buffered_error_log = false;
+      bool log_only_to_buffered_error_log = false;
+
+      // Add product specific buffered_error_log logic based on subsys & prio
+      // here
+      DBUG_EXECUTE_IF("redirect_message_to_buffered_error_log", {
+        log_to_buffered_error_log = true;
+        // also log to the normal log for MTR
+      });
+
+      if (log_to_buffered_error_log) {
+        buffered_error_log.log(buff_line, len);
+      }
+
       // write log-event to log-file
-      log_write_errstream(buff_line, len);
+      if (!log_only_to_buffered_error_log || !buffered_error_log.is_enabled())
+        log_write_errstream(buff_line, len);
     }
   }
 
--- a/sql/sys_vars.cc
+++ b/sql/sys_vars.cc
@@ -52,6 +52,7 @@
 #include <limits>
 
 #include "include/compression.h"
+#include "mysys/buffered_error_log.h"
 
 #include "my_loglevel.h"
 #include "mysql/components/services/log_builtins.h"
@@ -5993,6 +5994,18 @@ static bool check_log_path(sys_var *self, THD *, set_var *var) {
   return false;
 }
 
+static bool check_log_path_allow_empty(sys_var *self, THD *thd, set_var *var) {
+  if (!var->value) return false;
+
+  if (!var->save_result.string_value.str) return false;
+
+  if (strlen(var->save_result.string_value.str) == 0) return false;
+
+  if (!check_log_path(self, thd, var)) return false;
+
+  return true;
+}
+
 static bool fix_general_log_file(sys_var *, THD *, enum_var_type) {
   bool res;
 
@@ -8010,6 +8023,26 @@ static Sys_var_enum Sys_terminology_use_previous(
     terminology_use_previous_names, DEFAULT(terminology_use_previous::NONE),
     NO_MUTEX_GUARD, NOT_IN_BINLOG, ON_CHECK(nullptr), ON_UPDATE(nullptr),
     DEPRECATED_VAR(""));
+
+std::size_t buffered_error_log_size;
+
+static bool buffered_error_log_size_update(sys_var *, THD *, enum_var_type) {
+  buffered_error_log.resize(buffered_error_log_size * 1024);
+  return false;
+}
+
+static Sys_var_charptr Sys_var_buffered_error_log_filename(
+    "buffered_error_log_filename", "Filename of the buffered error log",
+    GLOBAL_VAR(buffered_error_log_filename), CMD_LINE(REQUIRED_ARG),
+    IN_FS_CHARSET, DEFAULT(""), NO_MUTEX_GUARD, NOT_IN_BINLOG,
+    ON_CHECK(check_log_path_allow_empty));
+
+static Sys_var_ulonglong Sys_var_buffered_error_log_size(
+    "buffered_error_log_size", "Size of the buffered error log (kB)",
+    GLOBAL_VAR(buffered_error_log_size), CMD_LINE(REQUIRED_ARG),
+    VALID_RANGE(0, ULLONG_MAX), DEFAULT(0), BLOCK_SIZE(1), NO_MUTEX_GUARD,
+    NOT_IN_BINLOG, ON_CHECK(nullptr),
+    ON_UPDATE(buffered_error_log_size_update));
 
 #ifndef NDEBUG
 Debug_shutdown_actions Debug_shutdown_actions::instance;
--- a/mysql-test/r/all_persisted_variables.result
+++ b/mysql-test/r/all_persisted_variables.result
@@ -42,7 +42,7 @@ include/assert.inc [Expect 500+ variables in the table. Due to open Bugs, we are
 
 # Test SET PERSIST
 
-include/assert.inc [Expect 497 persisted variables in the table.]
+include/assert.inc [Expect 499 persisted variables in the table.]
 
 ************************************************************
 * 3. Restart server, it must preserve the persisted variable
@@ -50,9 +50,9 @@ include/assert.inc [Expect 497 persisted variables in the table.]
 ************************************************************
 # restart
 
-include/assert.inc [Expect 497 persisted variables in persisted_variables table.]
-include/assert.inc [Expect 497 persisted variables shown as PERSISTED in variables_info table.]
-include/assert.inc [Expect 497 persisted variables with matching peristed and global values.]
+include/assert.inc [Expect 499 persisted variables in persisted_variables table.]
+include/assert.inc [Expect 499 persisted variables shown as PERSISTED in variables_info table.]
+include/assert.inc [Expect 499 persisted variables with matching peristed and global values.]
 
 ************************************************************
 * 4. Test RESET PERSIST IF EXISTS. Verify persisted variable
--- a/mysql-test/r/buffered_error_log.result
+++ b/mysql-test/r/buffered_error_log.result
@@ -1,0 +1,5 @@
+call mtr.add_suppression("Attempting backtrace. You can use the following information to");
+Sending kill signal 11 to mysqld ...
+# restart:
+Testing that buffered_error_log log contains something logged
+# restart:
--- a/mysql-test/r/mysqld--help-notwin.result
+++ b/mysql-test/r/mysqld--help-notwin.result
@@ -245,6 +245,10 @@ The following options may be given as the first argument:
  values are COMMIT_ORDER, WRITESET and WRITESET_SESSION.
  --block-encryption-mode=name 
  mode for AES_ENCRYPT/AES_DECRYPT
+ --buffered-error-log-filename=name 
+ Filename of the buffered error log
+ --buffered-error-log-size=# 
+ Size of the buffered error log (kB)
  --bulk-insert-buffer-size=# 
  Size of tree cache used in bulk insert optimisation. Note
  that this is a limit per thread!
@@ -1755,6 +1759,8 @@ binlog-transaction-compression-level-zstd 3
 binlog-transaction-dependency-history-size 25000
 binlog-transaction-dependency-tracking COMMIT_ORDER
 block-encryption-mode aes-128-ecb
+buffered-error-log-filename 
+buffered-error-log-size 0
 bulk-insert-buffer-size 8388608
 caching-sha2-password-digest-rounds 5000
 caching-sha2-password-private-key-path private_key.pem
--- a/mysql-test/suite/sys_vars/t/buffered_error_log_filename_basic.test
+++ b/mysql-test/suite/sys_vars/t/buffered_error_log_filename_basic.test
@@ -1,0 +1,19 @@
+SET @start_value = @@global.buffered_error_log_filename;
+SELECT @start_value;
+
+
+SET @@global.buffered_error_log_filename = "/tmp/bar.txt";
+SELECT @@global.buffered_error_log_filename;
+SET @@global.buffered_error_log_filename = '';
+SELECT @@global.buffered_error_log_filename;
+
+
+--Error ER_GLOBAL_VARIABLE
+SET @@session.buffered_error_log_filename = OFF;
+--Error ER_INCORRECT_GLOBAL_LOCAL_VAR
+SELECT @@session.buffered_error_log_filename;
+
+
+SET @@global.buffered_error_log_filename = @start_value;
+SELECT @@global.buffered_error_log_filename;
+
--- a/mysql-test/suite/sys_vars/t/buffered_error_log_size_basic.test
+++ b/mysql-test/suite/sys_vars/t/buffered_error_log_size_basic.test
@@ -1,0 +1,26 @@
+SET @start_value = @@global.buffered_error_log_size;
+SELECT @start_value;
+
+
+SET @@global.buffered_error_log_size = 420;
+SELECT @@global.buffered_error_log_size;
+SET @@global.buffered_error_log_size = 0;
+SELECT @@global.buffered_error_log_size;
+
+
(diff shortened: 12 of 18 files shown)


george-lorch:
PXC-3574 is "Port wsrep_Mode variable and Functionality from Galera 8.0.22-26.5"  Do you mean PS-7748? :)

dutow:
Ops, yes, I mixed up two jira tabs when writing the commit message. Fixed.

sql/sys_vars.cc
```diff
@@ -7625,6 +7626,24 @@ static Sys_var_ulonglong Sys_tf_sequence_table_max_upper_bound(
...
+  blackbox.resize(blackbox_size);
+  return false;
+}
+
+static Sys_var_ulonglong Sys_var_blackbox_size(
+    "blackbox_size", "Size of the blackbox ringbuffer",
+    GLOBAL_VAR(blackbox_size), CMD_LINE(REQUIRED_ARG),
+    VALID_RANGE(0, ULLONG_MAX), DEFAULT(0), BLOCK_SIZE(1), NO_MUTEX_GUARD,
+    NOT_IN_BINLOG, ON_CHECK(nullptr), ON_UPDATE(blackbox_size_update));
+
```
satya-bodapati: @dutow the option names are not very clear. Can we use something like memory_log_filename or memory_log_filesize
or mem_log_filename, or mem_log_filesize or shared_mem_log_filename?

satya-bodapati (changes requested):
sorry for the delay, will start the review tomorrow

sql/sys_vars.cc
dutow: I'm not sure about changing size to filesize, as this is primarily a memory buffer. We only write it out into a file if the server crashes.

As for changing blackbox to memory_log or mem_log, we can do that, I only used blackbox because that's the name mariadb uses.

mysys/blackbox.h
```diff
@@ -0,0 +1,114 @@
...
+
+#include <stdio.h>
+#include <cstring>
+#include <memory>
+#include <mutex>
+#include <vector>
+
+extern char *blackbox_filename;
+
+/** Stores log messages in a fixed size ringbuffer, which can be dumpeb by the
```
satya-bodapati: dumped*

mysys/blackbox.h
```diff
@@ -0,0 +1,114 @@
...
+ * crash detection code to provide additional debug information.
+ */
+class blackbox_logger {
+  using data_t = std::unique_ptr<std::vector<char>>;
```
satya-bodapati: any reason why std::vector<char> is preferred to std::string?

mysys/blackbox.h
```diff
@@ -0,0 +1,114 @@
...
+    if (looped && buffer_size >= new_buffer->size()) {
+      if (curr_offset == 0) {
+        // only one block
+        memcpy(data->data(), new_buffer->data(),
+               std::min(buffer_size, new_buffer->size()));
+        looped = false;
+        curr_offset = new_buffer->size();
+      } else {
+        // we are resizing a looped buffer into something bigger.
+        // TODO
```
satya-bodapati: whats the TODO left here?

mysys/blackbox.h
```diff
@@ -0,0 +1,114 @@
...
+      looped = true;
+    } else {
+      memcpy(data->data() + curr_offset, msg, len);
+      curr_offset += len + 1;
+    }
+    (*data)[curr_offset - 1] = '\n';
+  }
+
+  void write_to_disk() {
+    std::unique_lock<std::mutex> lk{data_mtx};
```
satya-bodapati: why not std::lock_guard?

sql/sys_vars.cc
satya-bodapati: Did we take the implementation from MariaDB?

sql/server_component/log_sink_trad.cc
```diff
@@ -349,8 +350,22 @@ int log_sink_trad(void *instance MY_ATTRIBUTE((unused)), log_line *ll) {
         log_sink_trad_last = microtime;
       }
 
+      bool log_to_blackbox = false;
+      bool log_only_to_blackbox = false;
+
+      // Add product specific blackbox logic based on subsys & prio here
+      //
```
satya-bodapati: remove this extra comment line

satya-bodapati (changes requested):

satya-bodapati (changes requested):
@dutow CircleCI (clang-format) test failed?
Also please provide jenkins link

satya-bodapati:
@dutow Regarding the performance impact, can we get some numbers?

I was reading the Logger functionality code and found that we do not strictly write every line to stderr and flush. There is a buffering logic and based on some interval duration, the buffer is flushed to disk.

Is it possible to adjust that time based logic to improve performance? Do we need this feature at all? :smile:

sql/sys_vars.cc
dutow: No. This is only in mariadb enterprise, and they used a shared memory buffer.

satya-bodapati: Great! then we are free to choose whatever we want :+1:

dutow:
@satya-bodapati I think flushing is only part of the issue, the amount of log is also there. PXC with wsrep_debug>0, especially with 4 will result in huge log files.

wsrep_on=1 results in a 25% increased execution time, 12% with this feature. If I only modify the original code by removing the actual file writing calls (logging calls are done, but nothing is written to he memory/disk) that result in 10-11%, which means that this blackbox logging feature only adds a 1 percent overhead over that. And most of that 10% is spent in the string formatting code, we can't do anything about that.

With wsrep=4 these numbers are 70%+ without, 39% with this code.

I got these numbers with a release build, and writing logs to a PCIE4 NVME SSD. If I store the logs on a slower SSD, it is much worse.

dutow:
Also, generally you only need the content of these additional logs when a crash occurs. During normal operation these just add additional noise to the normal server logs.

satya-bodapati:
> @satya-bodapati I think flushing is only part of the issue, the amount of log is also there. PXC with wsrep_debug>0, especially with 4 will result in huge log files.
> 
> wsrep_on=1 results in a 25% increased execution time, 12% with this feature. If I only modify the original code by removing the actual file writing calls (logging calls are done, but nothing is written to he memory/disk) that result in 10-11%, which means that this blackbox logging feature only adds a 1 percent overhead over that. And most of that 10% is spent in the string formatting code, we can't do anything about that.
> 
> With wsrep=4 these numbers are 70%+ without, 39% with this code.
> 
> I got these numbers with a release build, and writing logs to a PCIE4 NVME SSD. If I store the logs on a slower SSD, it is much worse.

Ok, if there is a way to measure performance, then it is really good. 
What about messages in ring buffer on recycle? Do we discard them?

Also, can you please verify how many messages are in buffer for each flush. I think this buffering isn't happening in all cases? :thinking: 
in log_sink_trad(), please check about this part

```
   341               if (out_types & LOG_ITEM_LOG_BUFFERED) {                                                                                                       │
│   342                 // abort if we've already printed this (to trad log only!)                                                                                   │
│   343                 if (microtime <= log_sink_trad_last) {                                                                                                       │
│   344                   out_fields = LOG_SERVICE_NOTHING_DONE;                                                                                                     │
│   345                   goto done;  
```

dutow:
@satya-bodapati Looks like that buffering is only there for early server startup. As soon as the logging system is properly initialized, it is no longer used.

For the ring buffer: we only keep the messages that fit into the buffer, and discard earlier ones. The idea is that users can set a buffer size which will most likely contain enough messages when an error occurs.

satya-bodapati:
> @satya-bodapati Looks like that buffering is only there for early server startup. As soon as the logging system is properly initialized, it is no longer used.
> 
> For the ring buffer: we only keep the messages that fit into the buffer, and discard earlier ones. The idea is that users can set a buffer size that will most likely contain enough messages when an error occurs.

@dutow, is it possible to implement a similar buffer logic and write that buffer after it is filled? 

Regarding the current ring buffer logic, I feel it is very hard for users to estimate the buffer size required. For example, how much should be the buffer size if I want to log messages that are generated in a day?

Hmm, can we write ring buffer before discarding it? (same request as above, but I guess this change will be in your new code?)

sql/sys_vars.cc
dutow: I've changed the name to buffered_log. With the ringbuffer / discarding behavior gone, I think that's a good description.

mysys/blackbox.h
dutow: fixed.

mysys/blackbox.h
dutow: Only that the size of the vector is more explicit than the capacity of a string, but that's not that important, I changed it to string.

mysys/blackbox.h
dutow: Nothing, removed.

mysys/blackbox.h
dutow: Changed to lock_guard.

sql/server_component/log_sink_trad.cc
dutow: removed.

mysys/buffered_log.h
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  bool is_enabled() {
+    std::lock_guard<std::mutex> lk{data_mtx};
+    return data.get() != nullptr && data->size() != 0;
+  }
+
+ private:
+  // requires holding data_mtx
+  void write_to_disk_() {
```
satya-bodapati: is there a way to assert mutex is taken?

mysys/buffered_log.h
```diff
@@ -0,0 +1,96 @@
...
+    if (buffered_log_filename == nullptr || strlen(buffered_log_filename) == 0)
+      return;
+    if (data.get() == nullptr || data->size() == 0) {
+      return;
+    }
```
satya-bodapati: better to follow same style w.r.t to braces for both if condidtions

mysys/buffered_log.h
```diff
@@ -0,0 +1,96 @@
...
+    auto fdd = fopen(buffered_log_filename, "a");
+    fwrite(data->data(), data->size(), 1, fdd);
+    fclose(fdd);
```
satya-bodapati: Should we keep closing the file descriptor after each buffer flush?

mysys/buffered_log.h
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  ~buffered_logger() { write_to_disk(); }
+
+  void log(const char *msg, size_t len) {
+    std::lock_guard<std::mutex> lk{data_mtx};
+    if (data.get() == nullptr || data->capacity() == 0) return;
+    if (msg[len - 1] == 0) {
+      len--;
+    }
```
satya-bodapati: what will happen if we dont do this? write \0 to file?

mysys/buffered_log.h
```diff
@@ -0,0 +1,96 @@
...
+#include <stdio.h>
+#include <cstring>
+#include <memory>
+#include <mutex>
+#include <vector>
+
+extern char *buffered_log_filename;
+
+/** Stores log messages in a fixed size buffer, which can be dumped by large
+ * chunks instead of line by line, to increase performace for high throughput
```
satya-bodapati: performance

sql/server_component/log_sink_trad.cc
satya-bodapati: I still see it

sql/server_component/log_sink_trad.cc
```diff
@@ -349,8 +350,23 @@ int log_sink_trad(void *instance MY_ATTRIBUTE((unused)), log_line *ll) {
         log_sink_trad_last = microtime;
       }
 
+      bool log_to_buffered_log = false;
+      bool log_only_to_buffered_log = false;
```
satya-bodapati: what is the use of this variable? Can be removed?

sql/sys_vars.cc
```diff
@@ -7625,6 +7626,24 @@ static Sys_var_ulonglong Sys_tf_sequence_table_max_upper_bound(
     GLOBAL_VAR(tf_sequence_table_max_upper_bound), CMD_LINE(REQUIRED_ARG),
     VALID_RANGE(1024, ULLONG_MAX), DEFAULT(1048576), BLOCK_SIZE(1));
 
+static Sys_var_charptr Sys_var_buffered_log_filename(
+    "buffered_log_filename", "Filename for buffered log",
```
satya-bodapati: I think it we should mention error_log instead of just log

buffer_error_log_filename

satya-bodapati (changes requested):
@dutow Looks really good. Few suggestions

Now that we have decided on new names, Please fix
1. Github description
2. Remove blackbox everywhere
3. Commit message

sql/sys_vars.cc
```diff
@@ -7625,6 +7626,24 @@ static Sys_var_ulonglong Sys_tf_sequence_table_max_upper_bound(
...
+
+std::size_t buffered_log_size;
+
+static bool buffered_log_size_update(sys_var *, THD *, enum_var_type) {
+  buffered_log.resize(buffered_log_size);
+  return false;
+}
+
+static Sys_var_ulonglong Sys_var_buffered_log_size(
+    "buffered_log_size", "Size of the buffered log",
```
satya-bodapati: buffered_error_log_size

Also please change all variable name to indicate error or err

satya-bodapati (changes requested):

sql/server_component/log_sink_trad.cc
satya-bodapati: log_only_to_buffered_log

mysys/buffered_log.h
dutow: No, this only affects the length calculation, it's not that critical. This len is only used for the buffer full check below it. It was more important with the vector version, where we did write the \0 to the file otherwise.

In the current code I could remove it, and then sometimes we might flush the buffer one message earlier than we should, that doesn't really matter.

satya-bodapati: Ok, will leave it to you :) if you want to retain, please add comment

mysys/buffered_log.h
satya-bodapati: Just check. Not very important.

sql/server_component/log_sink_trad.cc
dutow: No, these both have real usecases.

log_to_buffered_log tells us if we should write something to the buffered log.
log_only_to_buffered_log tells us that we shouldn't write that message to the normal error log, only to the buffered log.

We definitely need log_only_to_buffered_log. 

Log_to_buffered_log is a different question. For performance testing, I've written all log messages to the buffered error log too, but we might have use cases where we only want to write some of the log messages there. The mariadb implementation is like that (based on the documentation), they only write wsrep messages to the blackbox, that's why both variables are here.

mysys/buffered_log.h
dutow: I don't think there's any without a custom wrapper class around std::mutex, or adding an std::unique_lock as a data member. std::mutex doesn't have this functionality. Even try_lock + unlock doesn't work, because calling try_lock if the current thread already owns the mutex is UB.

sql/server_component/log_sink_trad.cc
satya-bodapati: I mean, it will be compiler optimized anyway. I see the variable used
bool x  = false
if (!false || ..)

dutow: Because in the current code, there's nothing using this logging other than that DBUG_EXECUTE_IF code (and that only checks that we write to the buffered log correctly).

PXC will set log_only_to_buffered_log true if we are writing a wsrep/galera log message above the warning level.

mysys/buffered_log.h
satya-bodapati: ok, ignore this

sql/server_component/log_sink_trad.cc
satya-bodapati: So if I understand correctly, there are two functionalities we want to achieve?
1. Bufffered logging (log to buffer and write to disk when buffer is full)
2. Buffer-only Logging (write only to buffer, when buffer is full, we discard/overwrite. Buffer is written to disk only at crash)?

satya-bodapati: And regarding 1. the buffered logging, can't we use same error log file ? ie. write buffered messages to default mysqld error log.

dutow: No, we only want 2. Changing the basic error logging isn't a goal of this PR, as that doesn't produce enough output to be worth it.

But we want to be able to do "2" in two ways:

* write normal messages to the classic error log. Write both normal and high output debug messages to the buffered error log
* write normal messages to the classic error log. Write high output debug messages to the buffered error log

mysys/buffered_log.h
dutow: Good question. I implemented this way because (depending on the buffer size) writes aren't that frequent. With this the user should also be free to move / rotate / etc the logfile,as we always reopen it before writing to it.

mysys/buffered_log.h
dutow: I removed it, it shouldn't cause any noticable difference.

mysys/buffered_log.h
dutow: fixed

sql/server_component/log_sink_trad.cc
satya-bodapati: If not for log rotation, I would love to see a simpler logic. Just use buffered functionality for all messages and write buffer to mysqld error log. 

Maybe we should keep the log rotation problem separate and fix that separately. WIth that my proposal would be

1. No separate error log file (no option for this)
2. If buferred_error_log_buffer_size is not 0, use buffering and write the buffer to mysqld error log file.
3. If buferred_error_log_buffer_size is 0, normal logging (as if this feature is disabled)

mysys/buffered_error_log.h
```diff
@@ -0,0 +1,95 @@
...
+extern char *buffered_error_log_filename;
+
+/** Stores log messages in a fixed size buffer, which can be dumped by large
+ * chunks instead of line by line, to increase performance for high throughput
+ * scenarios.
+ *
+ * The buffer is also dumped by the crash reporting code and during shutdown to
+ * ensure that nothing is lost, only delayed.
+ */
+class buffered_error_logger {
```
satya-bodapati: upstream style is Buffered_error_logger

mysys/buffered_error_log.h
```diff
@@ -0,0 +1,95 @@
...
+    data.swap(new_buffer);
+  }
+
+  ~buffered_error_logger() { write_to_disk(); }
+
+  void log(const char *msg, size_t len) {
+    std::lock_guard<std::mutex> lk{data_mtx};
+    if (data.get() == nullptr || data->capacity() == 0) return;
+    const auto msg_end = data->size() + len;
+    if (msg_end > data->capacity() - 1) {
```
satya-bodapati: should we move implementation to .cc file?

mysys/buffered_error_log.h
```diff
@@ -0,0 +1,95 @@
...
+  // requires holding data_mtx
+  void write_to_disk_() {
+    if (buffered_error_log_filename == nullptr ||
+        strlen(buffered_error_log_filename) == 0) {
+      return;
+    }
+    if (data.get() == nullptr || data->size() == 0) {
+      return;
+    }
+    auto fdd = fopen(buffered_error_log_filename, "a");
```
satya-bodapati: mysql-secure-file-priv check missing?

sql/sys_vars.cc
```diff
@@ -7655,6 +7656,25 @@ static Sys_var_ulonglong Sys_tf_sequence_table_max_upper_bound(
...
+    GLOBAL_VAR(buffered_error_log_filename), CMD_LINE(REQUIRED_ARG),
+    IN_FS_CHARSET, DEFAULT(""));
+
+std::size_t buffered_error_log_size;
+
+static bool buffered_error_log_size_update(sys_var *, THD *, enum_var_type) {
+  buffered_error_log.resize(buffered_error_log_size);
+  return false;
+}
+
```
satya-bodapati: Better to keep these separate and all variables in same codeblock

sql/sys_vars.cc
```diff
@@ -7655,6 +7656,25 @@ static Sys_var_ulonglong Sys_tf_sequence_table_max_upper_bound(
     GLOBAL_VAR(tf_sequence_table_max_upper_bound), CMD_LINE(REQUIRED_ARG),
     VALID_RANGE(1024, ULLONG_MAX), DEFAULT(1048576), BLOCK_SIZE(1));
 
+static Sys_var_charptr Sys_var_buffered_error_log_filename(
```
satya-bodapati: Also put them in alphabetical sort position? :thinking:

sql/server_component/log_sink_trad.cc
```diff
@@ -349,8 +350,23 @@ int log_sink_trad(void *instance MY_ATTRIBUTE((unused)), log_line *ll) {
         log_sink_trad_last = microtime;
       }
 
+      bool log_to_buffered_error_log = false;
+      bool log_only_to_buffered_error_log = false;
+
+      // Add product specific buffered_error_log logic based on subsys & prio
+      // here
```
satya-bodapati, line 395: Should we add the control flag elsewhere?
LIke this?: LogEvent().module("blah").buffered(true)

satya-bodapati (changes requested):

sql/server_component/log_sink_trad.cc
satya-bodapati: @dutow what do u think? u want to keep buffered log file as separate?

sql/server_component/log_sink_trad.cc
dutow: I'm not sure about this, as the idea is to log everything printed by a submodule or loglevel to this log. That condition is easier to implement directly in log_sink_trad instead of adding it to all the macros/places log messages are generated.

sql/sys_vars.cc
dutow: That isn't an alphabetical list, entries before this are also in a random order.

sql/sys_vars.cc
dutow: Fixed.

mysys/buffered_error_log.h
dutow: It isn't used in any of the log variables. They use a different log file name check functions, I've added that.

mysys/buffered_error_log.h
dutow: Done.

mysys/buffered_error_log.h
dutow: Changed.

sql/server_component/log_sink_trad.cc
satya-bodapati: @dutow please let me know your thoughts on this (avoiding separate error log file) :)

sql/sys_vars.cc
```diff
@@ -7655,6 +7656,25 @@ static Sys_var_ulonglong Sys_tf_sequence_table_max_upper_bound(
...
+    "buffered_error_log_filename", "Filename of the buffered error log",
+    GLOBAL_VAR(buffered_error_log_filename), CMD_LINE(REQUIRED_ARG),
+    IN_FS_CHARSET, DEFAULT(""), NO_MUTEX_GUARD, NOT_IN_BINLOG, ON_CHECK(check_log_path));
+
+static Sys_var_ulonglong Sys_var_buffered_error_log_size(
+    "buffered_error_log_size", "Size of the buffered error log",
+    GLOBAL_VAR(buffered_error_log_size), CMD_LINE(REQUIRED_ARG),
+    VALID_RANGE(0, ULLONG_MAX), DEFAULT(0), BLOCK_SIZE(1), NO_MUTEX_GUARD,
+    NOT_IN_BINLOG, ON_CHECK(nullptr),
+    ON_UPDATE(buffered_error_log_size_update));
```
satya-bodapati: I think we should not allow arbirtray sizes. what if we set just 100 bytes? and a single error message needs 200 bytes?
May be multiples of 16K or 1MB? or start value for which we saw decent improvement?

sql/server_component/log_sink_trad.cc
satya-bodapati: ok, lets leave it as is

mysys/buffered_error_log.cc
```diff
@@ -0,0 +1,71 @@
...
+  data_t new_buffer(new std::string());
+  new_buffer->reserve(buffer_size);
+  data.swap(new_buffer);
+}
+
+Buffered_error_logger::~Buffered_error_logger() { write_to_disk(); }
+
+void Buffered_error_logger::log(const char *msg, size_t len) {
+  std::lock_guard<std::mutex> lk{data_mtx};
+  if (data.get() == nullptr || data->capacity() == 0) return;
```
satya-bodapati, line 32: when do we come log and the buffer capacity is null or 0?

sql/sys_vars.cc
satya-bodapati: Also, if the buffer is 1MB, the buffer is filled only until 512K, we will never see the remaining 512k until shutdown or crash?

satya-bodapati (changes requested):
few concerns on the buffer size length possibilities
Also please check my question on alternate possibility (avoiding separate error log file)

sql/server_component/log_sink_trad.cc
dutow: Yes, I'd like to keep it separe for several reasons:

1. buffered logging is delayed. Users might want to reveive normal log messages without significant delays.
2. buffered logging is intended for debug logs (high throughput, huge logfiles). This is only required if something bad happens, and we want to investigate how that happened, while the normal logs could be useful for normal server operation. Also, because of the larger size, keeping it separate means users can use different rotation/size limits (e.g. keep normal logs for a few months, but keep buffered logs only for the last day, or something like that)

sql/sys_vars.cc
dutow: Yes, we only write on buffer full and on exit.

mysys/buffered_error_log.cc
dutow: If `log_to_buffered_error_log` is true, but the user didn't configure buffered logging.

sql/sys_vars.cc
satya-bodapati: @dutow can you please tell what happens when buffer size is 100 bytes and a single error message is 200 bytes?

sql/server_component/log_sink_trad.cc
satya-bodapati: ok @dutow I think the answers are convincing enough to keep it as is

sql/sys_vars.cc
dutow: Since we are using an std::string as buffer and operator+= for adding the new log line to the buffer, it will automatically extend the buffer.

satya-bodapati (approved):
LGTM! Please ensure clean jenkins run before merging

mysys/buffered_error_log.h
satya-bodapati: @dutow I think, it is important to add file-priv check. THis is because, other log-file variables are startup and read-only. Since this is a dynamic option, we have to do privilege check. Apart from that, we have to ensure that the file pointed doesn't overwrite an existing file? What if the user points to mysql.ibd? or any other file inside datadir?

satya-bodapati (changes requested):
@dutow please check my comment on privilege check and other vulnerabilities.

mysys/buffered_error_log.h
satya-bodapati: @dutow Please check if the above validations are done for https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_general_log_file?

sql/sys_vars.cc
```diff
@@ -4010,6 +4010,14 @@ static Sys_var_charptr Sys_secure_file_priv(
...
     DEFAULT(DEFAULT_SECURE_FILE_PRIV_DIR));
 
+static Sys_var_charptr Sys_secure_log_path(
+    "secure_log_path",
+    "Limit location of general log, slow log and buffered error log"
+    "within specified directory",
+    READ_ONLY NON_PERSIST GLOBAL_VAR(opt_secure_log_path),
+    CMD_LINE(REQUIRED_ARG), IN_FS_CHARSET,
+    DEFAULT(DEFAULT_SECURE_FILE_PRIV_DIR));
+
```
satya-bodapati: @dutow Change looks good. Please add some negative tests with buffered_error_log and the sys_vars.test

satya-bodapati (changes requested):

dutow:
@satya-bodapati I rebased this based on the now merged log restricttion patch.

mysql-test/t/buffered_error_log.test
```diff
@@ -0,0 +1,21 @@
+call mtr.add_suppression("Attempting backtrace. You can use the following information to");
```
satya-bodapati, line 1: @dutow why did you suppress this? Does it mean there was assertion failure and core file generated?

mysql-test/suite/sys_vars/r/buffered_error_log_filename_basic.result
```diff
@@ -0,0 +1,20 @@
+SET @start_value = @@global.buffered_error_log_filename;
+SELECT @start_value;
+@start_value
+
+SET @@global.buffered_error_log_filename = "/tmp/bar.txt";
+SELECT @@global.buffered_error_log_filename;
+@@global.buffered_error_log_filename
+/tmp/bar.txt
+SET @@global.buffered_error_log_filename = '';
```
satya-bodapati, line 9: when the buffered file name is empty, what is the filename chosen?

mysql-test/suite/sys_vars/r/buffered_error_log_size_basic.result
```diff
@@ -0,0 +1,31 @@
...
+0
+SET @@global.buffered_error_log_size = 420;
+SELECT @@global.buffered_error_log_size;
+@@global.buffered_error_log_size
+420
+SET @@global.buffered_error_log_size = 0;
+SELECT @@global.buffered_error_log_size;
+@@global.buffered_error_log_size
+0
+SET @@session.buffered_error_log_size = OFF;
```
satya-bodapati: if you use GLOBAL.buffer_err_log_size=OFF, will it accept?

mysql-test/t/buffered_error_log.test
dutow: The test intentionally crashes the server. (to test that we do write the log)

mysql-test/suite/sys_vars/r/buffered_error_log_filename_basic.result
dutow: Nothing, then it's off.

mysql-test/t/buffered_error_log.test
satya-bodapati: simple, kill_mysqld.inc is not enough?

mysql-test/suite/sys_vars/r/buffered_error_log_filename_basic.result
satya-bodapati: ok, makes sense. Can you show the output of filename instead of error_log_size?

mysql-test/suite/sys_vars/r/buffered_error_log_size_basic.result
dutow: No, it fails with incorrect argument type. I'll change this to 0 too.

mysql-test/t/buffered_error_log.test
dutow: It's not, it doesn't suppress anything. You can see this suppression in many tests that use it.

mysql-test/suite/sys_vars/r/buffered_error_log_filename_basic.result
dutow: I don't understand the question, the select below shows the filename? This file doesn't mention size at all.

mysql-test/suite/sys_vars/r/buffered_error_log_size_basic.result
dutow: Fixed.

mysql-test/suite/sys_vars/r/buffered_error_log_filename_basic.result
satya-bodapati: sorry, I didn't see that you are showing filename.

mysql-test/t/buffered_error_log.test
satya-bodapati: I mean, instead of send_kill_to_mysqld.inc, why not use kill_mysqld.inc which sends signal 9?

satya-bodapati: ok, we can want to check if error log is saved to disk on core file generation/assert failures. Got it

satya-bodapati (approved):

kamil-holubicki:
My understanding is that it works in the way:
1. Log only to ram-buffered log
2. Log only to the disk log
3. Log to 1 AND 2
No filtering, logs splitting based on log severity, and so on.

Just a guess:
In this comment https://github.com/percona/percona-server/pull/4358#issuecomment-891560720 you showed benchmark analysis. Do you have similar results for locating the standard error log on /dev/shm? Do we still have the performance problem there? If not, maybe it would be better to implement the functionality for circular logging for the standard logging system?


In other comment I read:
```wsrep_on=1 results in a 25% increased execution time, 12% with this feature. If I only modify the original code by removing the actual file writing calls (logging calls are done, but nothing is written to he memory/disk) that result in 10-11%```
Do I understand correctly that original logging, if disk write is removed is as fast as the new feature? If yes, I would check the behavior with /dev/shm and consider adding circular logging feature for standard logging system.

dutow:
@kamil-holubicki I did consider that, but that would mean delaying normal log messages, while users might expect them to appear in time as usual.

This solution is a compromise: it keeps the existing log functioning as is, while adding a second sink for bulk output that we don't expect to be readable immediately.

PR #4360, opened 2021-06-24
PS-7753: Select HEX encoded string when checking data in encrypted column

oleksandr-kachan:
https://jira.percona.com/browse/PS-7753

In case of innodb.import_compress_encrypt test failure it adds
substrings selected from encrypted VARCHAR column directly into xml
report. This results in unreadable characters being added into report
and corrupted xml file.
Use hex() encoded substrings for results comparison instead.


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/473/

percona-ysorokin (approved):
LGTM

PR #4364, opened 2021-06-25
PS-7666: Assertion failure: fil0fil.cc:13097:fil_system->shard_by_id(…

satya-bodapati:
…id)->mutex_owned()

https://jira.percona.com/browse/PS-7666

Problem:
--------
On debug builds, a crash during DROP TABLE operation, makes the database unusable.
InnoDB will enter crash recovery but it will keep crashing with the
same assertion failure

For the DROP TABLE operation, InnoDB writes MLOG_FILE_DELETE redo log entry.
A subsequent start, InnoDB will enter crash recovery and tries to replay the
operation. During this replay, it tries to delete the file again and remove
the in-memory tablespace object (fil_space_t).

Removal of the in-memory tablespace object fails with assertion failure as
it expect the Fil_system->Fil_shard mutex to be taken.

Fix:
----
Acquire Fil_system->Fil_shard mutex for the in-memory space removal and release
after it is done with removal

### DIFF
--- a/mysql-test/suite/innodb/r/drop_table_crash_debug.result
+++ b/mysql-test/suite/innodb/r/drop_table_crash_debug.result
@@ -1,0 +1,23 @@
+#
+# PS-7666 : Assertion failure: fil0fil.cc:13097:fil_system->shard_by_id(id)->mutex_owned()
+#
+CREATE TABLE t1(a INT);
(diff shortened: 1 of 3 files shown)


satya-bodapati:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/987/

rahulmalik87 (approved):
LGTM

PR #4369, opened 2021-06-29
PS-7772: Fix rocksdb_stress MTR tests to work with old python connectors

oleksandr-kachan:
https://jira.percona.com/browse/PS-7772

In PS 8.0 default authentication plugin was changed to
caching_sha2_password. Python connector installed on most of the
platforms tested on Jenkins doesn't support this plugin and as a result
tests fail to run.
Updated rocksdb stress tests to use separate user with mysql_native_password
auth plugin for Python scripts.


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/989/

percona-ysorokin (approved):
LGTM

PR #4373, opened 2021-06-30
Bug#26330279 - ASSERT ON ROW_PURGE_REMOVE_SEC_IF_POSS_LEAF| INNOBASE/…

altmannmarcelo:
…ROW/ROW0PURGE.CC

Problem:
--------

On a table with virtual column and a secondary index on the prefix of virtual column.
A replace operation is executed with the same values of existing record. So as part of
REPLACE, the existing record is delete-marked and subsequent insert operation is
treated as update-by-modify(un-delete mark) the record.

Later when purge tries to process the update-undo, it finds the record eligible
for purge but the record is no longer delete-marked. Hence it asserts.

Ideally, purge shouldn't find the record eligibile for purge. The problem occurs
when trying to match the secondary index record to its corresponding cluster index
record.

If there are no versions matching the secondary index record, we declare the
secondary index as purgeable.

In the bug case, when trying to match the secondary index (which is on vcol),
the comparision goes wrong because the actual secondary index field size is
less than the field size logged for virtual column.

Fix:
----
When trying to match secondary index on virtual column to its corresponding
cluster index record, the field size on prefix index will be adjusted to its
actual secondary index field size.

Reviewed-By: Jimmy Yang <jimmy.yang@oracle.com>
RB: 17706
(cherry picked from commit ef4a3ee7329afe4dedd8aeee9274fb41fce8aafd)


altmannmarcelo:
https://ps57.cd.percona.com/job/percona-server-5.7-param/475/

satya-bodapati (approved):
LGTM! You can null merge (self approve) to 8.0

PR #4381, opened 2021-07-02
PS 7754 - Null merge from 5.7 of fb-prod202006

ldonoso:
https://jira.percona.com/browse/PS-7754
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/994/


inikep (approved):
LGTM

PR #4389, opened 2021-07-06
PS-1346: Reset enable_slow_log thd variable once log query is applied for replica thread (5.7)

oleksandr-kachan:
https://jira.percona.com/browse/PS-1346

In Query_log_event::do_apply_event() thd->enable_slow_log is
unconditionally set to TRUE after log query is applied on replica
server. As a result of this the log_slow_applicable() function will not
be able to properly detect if next applied record should be logged to a
slow log. It will be able to find this out only in
Query_logger::slow_log_write() after checking
opt_log_slow_slave_statements value.

To fix this the code which is supposed to reset enable_slow_log thd
variable is reverted back to upstream state.

### DIFF
--- a/sql/log_event.cc
+++ b/sql/log_event.cc
@@ -5000,7 +5000,7 @@ int Query_log_event::do_apply_event(Relay_log_info const *rli,
         thd->enable_slow_log is set to the value of
         opt_log_slow_admin_statements).
       */
-      thd->enable_slow_log= TRUE;
+      thd->enable_slow_log= opt_log_slow_slave_statements;
     }
     else
     {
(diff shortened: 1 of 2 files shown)


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/480/

sql/log_event.cc
```diff
@@ -5000,7 +5000,7 @@ int Query_log_event::do_apply_event(Relay_log_info const *rli,
         thd->enable_slow_log is set to the value of
         opt_log_slow_admin_statements).
       */
-      thd->enable_slow_log= TRUE;
+      thd->enable_slow_log= opt_log_slow_slave_statements;
```
satya-bodapati, line 5003: what about the other place mentioned by bug reporter?
sql/slave.cc init_slave_thread

satya-bodapati (changes requested):

sql/log_event.cc
oleksandr-kachan: It's already up to date with upstream
`  thd->slave_thread = 1;
  thd->enable_slow_log= opt_log_slow_slave_statements;
`

satya-bodapati: ok :+1:

satya-bodapati (approved):

PR #4397, opened 2021-07-08
PS 7780 [8.0] - Null merge of FB production tag fb-prod202101

ldonoso:
<https://jira.percona.com/browse/PS-7780>
<https://ps80.cd.percona.com/job/percona-server-8.0-param/1003/>

---
The following FB commits from 5.7 are new in 8.0:
- [d902f20e47c29b96dfda147085610f1a1176b8c4 Adding rocksdb_enable_pipelined_write variable](https://github.com/facebook/mysql-5.6/commit/d902f20e47c29b96dfda147085610f1a1176b8c4)
    
As a consequence of the new commits, these are the new variables:
- `rocksdb_enable_pipelined_write`


inikep (changes requested):
I think you should remove all changes except `rocksdb_enable_pipelined_write` and the fix for `rocksdb_tracing`. `decode_value = true` and other formatting changes should be skipped as it can lead to conflicts later in 8.0.

inikep (approved):
LGTM

PR #4398, opened 2021-07-09
PS-1346: Reset enable_slow_log thd variable once log query is applied for replica thread (8.0)

oleksandr-kachan:
https://jira.percona.com/browse/PS-1346

In Query_log_event::do_apply_event() thd->enable_slow_log is
unconditionally set to TRUE after log query is applied on replica
server. As a result of this the log_slow_applicable() function will not
be able to properly detect if next applied record should be logged to a
slow log. It will be able to find this out only in
Query_logger::slow_log_write() after checking
opt_log_slow_slave_statements value.

To fix this the code which is supposed to reset enable_slow_log thd
variable is reverted back to upstream state.


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1004/

satya-bodapati (approved):

PR #4400, opened 2021-07-12
PS-6002: Create/alter event state to keep enabled on replicas

venkatesh-prasad-v:
https://jira.percona.com/browse/PS-6002

This feature introduces a global config variable `--replica-enable-event`
that accepts the regex patterns of the events in format
"db_pattern.event_pattern" for those events that need to be re-enabld
after the creation or modification of event on replica server, unlike
the current behavior where the replicated events are by default set to
the state SLAVESIDE_DISABLED.

The below shall be the behavior when a replicated event matches the pattern
specified in the `--replica-enable-event`.

1. When an event is created as ENABLED on source, it shall remain ENABLED
   on replica as well.

2. When an event is created as DISABLED it shall remain DISABLED on
   replica as well.

3. If an event in ENABLED state is altered as DISABLED on source, then it
   shall become DISABLED on replica as well.

4. If an event in ENABLED state is altered and if it remains ENABLED on
   source, then it shall remain ENABLED on replica as well.

5. If an event in DISABLED state is altered and if it remains DISABLED on
   source then, shall remain DISABLED on replica as well.

6. If an event in DISABLED state is altered to ENABLED state on the source
   server, then it shall become ENABLED on replica as well.

Design
---
https://docs.google.com/document/d/1jXLmqt4LEz0ZgmpQZvJ3W1P0n0_XN3AC1LJDlxvbzsk/edit?usp=sharing

Testing Done
---
Jenkins: https://ps80.cd.percona.com/job/percona-server-8.0-param/1006/ 
Failing tests are unrelated to the current changes.


sql/mysqld.cc
```diff
@@ -8754,6 +8755,16 @@ struct my_option my_long_options[] = {
...
+    {"replica-enable-event", OPT_REPLICA_ENABLE_EVENT,
+     "Tells the replication applier thread to enable the events that match "
+     "the specified wildcard pattern without setting it as SLAVESIDE_DISABLED. "
+     "To specify more than one event, use the directive multiple times, once "
+     "for each event. This will work for cross-database events. Example: "
+     "replica-enable-event=foo%.bar% will enable the events in all "
+     "databases on replica server that start with 'foo' and whose event names "
+     "start with 'bar'.",
+     nullptr, nullptr, nullptr, GET_STR, REQUIRED_ARG, 0, 0, 0, nullptr, 0,
+     nullptr},
```
satya-bodapati: sys_vars.option_basic.test is not necessary?

satya-bodapati: Can't we do SET GLOBAL on this option? any reason why it is 'start-up' only?

Many times, we will not know the events to be replicated at the time of starting server

venkatesh-prasad-v: There are some overheads and problems associated with making the values to be updated via SET GLOBAL.

1. It requires a separate parser to add/modify the values.
2. It is not very extensible when someone wants to add more and more patterns to the variable.
3. Requires additional locking for setting the values from multiple threads. With the current approach, there is no need for locking.
4. It becomes a problem on server restart when the variable is not persisted, or it may also need some infrastructure to store the previous value in the system tables.

Because of the above reasons, I found it to be better to be settable through cnf file than through SQL command.
So, there was no need to add a testcase in the `sys_vars` suite and the test `rpl_enable_event_cmd_line_validation.test` covers most of the cases.

satya-bodapati: well, it is extra work, but we can do, maybe as an extension or improvement

1. well, just set is fine. For simplicity, we can take the entire new value provided by the user
2. same as 1 take all or nothing. latest provided by user
3. concurrent SET GLOBAL of an option are already synchronized by mutex
4. SET PERSIST

satya-bodapati: sys_vars suite will mandatorily ask for option_basic.test file isn't it? :thinking: 
It does not depend on the option type

mysql-test/suite/rpl/t/rpl_enable_event_cmd_line_validation.test
```diff
@@ -0,0 +1,94 @@
...
+--source extra/rpl_tests/rpl_enable_event_cmd_line_validation.inc
+
+# Positiv case 2 : DB.EVENT
+--let $wild_pattern=d%b%.%event%
+--let $expect_abort= no
+--source extra/rpl_tests/rpl_enable_event_cmd_line_validation.inc
+
+# Restart the replica server and start the applier thread.
+--let $rpl_server_number= 2
+--let $rpl_server_parameters= --replica-enable-event=db%.event1% --replica-enable-event=d%b%.%event%
```
satya-bodapati, line 69: does any other mysqld option work like this? I mean, used multiple times and all occurrences are accumulated.
Can we do a comma-separated list instead?

satya-bodapati (changes requested):

mysql-test/suite/rpl/t/rpl_enable_event_cmd_line_validation.test
venkatesh-prasad-v: All `--replicate-(do|ignore)-(db|table)` options work the same way. https://github.com/percona/percona-server/blob/8.0/sql/mysqld.cc#L8820-L8884.

satya-bodapati: I see. MySQL behaviour is a bit inconsistent here. For some options, it would take the last value only. for example, sort_buffer_size or innodb_buffer_pool_size if passed multiple times, only the last one is considered.

Ok to match the other replicate-* options and keep this one as is

mysql-test/suite/rpl/t/rpl_enable_event_cmd_line_validation.test
```diff
@@ -0,0 +1,94 @@
...
+--let $reason= The pattern has multiple dot characters. Please restart with format <db_pattern>.<event_pattern>
+--source extra/rpl_tests/rpl_enable_event_cmd_line_validation.inc
+
+# Negative case 6 : Channel separator
+--let $wild_pattern=channel:db%.event%
+--let $expect_abort= yes
+--let $reason= This server doesn't support per-channel --replica-enable-event feature
+--source extra/rpl_tests/rpl_enable_event_cmd_line_validation.inc
+
+# Positiv case 1 : DB.EVENT
```
laurynas-biveinis: Here and below "Positive"

Long time no see! :)

satya-bodapati: Laurynas is back!!! :dancing_panda:

sql/mysqld.cc
satya-bodapati: I debugged a bit and it seems select * from performance_schema.global_variables doesn't show this variable. That is why sys_vars.all_vars suite doesn't complain.

Can you please check why pfs.global_variables doesn't show this newly added variable. I doesn't show other replicate-* variables as well.. something is happening.

satya-bodapati (changes requested):

sql/mysqld.cc
venkatesh-prasad-v: Hi @satya-bodapati! Yes. This is expected because of the fact that it is not defined as a subclass of the `sys_var` class to be able to be considered as a system variable.

> Can you please check why pfs.global_variables doesn't show this newly added variable. I doesn't show other replicate-* variables as well.. something is happening.

As replicate-* variables behave in a different way (new value gets on accumulated) when compared to other variables which overrides the previous values when the same variable is specified multiple times.

Upstream introduced new perfschema table for the observability of replicate-* variables. But however, we cannot do the same. So, instead we need to implement a special type of `Sys_var` to be reported under `performance_schema.global_variables` and `SHOW GLOBAL VARIABLES`.

I'm currently working on such an implementation using `Sys_var_charptr_func` and will update the PR if it works.

venkatesh-prasad-v: Done. `replica_enable_event` is now a non persistable read only global variable.
```
mysql> show global variables like '%replica_en%';
+----------------------+---------------------------+
| Variable_name        | Value                     |
+----------------------+---------------------------+
| replica_enable_event | "db%.event%","%db%.%ev1%" |
+----------------------+---------------------------+
1 row in set (0.00 sec)

mysql> select * from performance_schema.global_variables where variable_name like '%Replica_en%';
+----------------------+---------------------------+
| VARIABLE_NAME        | VARIABLE_VALUE            |
+----------------------+---------------------------+
| replica_enable_event | "db%.event%","%db%.%ev1%" |
+----------------------+---------------------------+
1 row in set (0.01 sec)

mysql> select @@replica_enable_event;
+---------------------------+
| @@replica_enable_event    |
+---------------------------+
| "db%.event%","%db%.%ev1%" |
+---------------------------+
1 row in set (0.00 sec)

mysql> select @@session.replica_enable_event;
ERROR 1238 (HY000): Variable 'replica_enable_event' is a GLOBAL variable

mysql> set session replica_enable_event = 'db.ev';
ERROR 1238 (HY000): Variable 'replica_enable_event' is a read only variable

mysql> set global replica_enable_event = 'db.ev';
ERROR 1238 (HY000): Variable 'replica_enable_event' is a read only variable
```

venkatesh-prasad-v:
Jenkins: https://ps80.cd.percona.com/job/percona-server-8.0-param/1020/console

mysql-test/suite/sys_vars/r/all_vars.result
```diff
@@ -52,6 +52,8 @@ regexp_stack_limit
 regexp_stack_limit
 regexp_time_limit
 regexp_time_limit
+replica_enable_event
+replica_enable_event
```
satya-bodapati: You already added  _basic.test this, shouldn't be present. Can you please check?

sql/rpl_event_ctx.cc
```diff
@@ -0,0 +1,134 @@
...
+  std::string pattern(entry);
+  auto it =
+      std::find(m_event_wild_list->begin(), m_event_wild_list->end(), pattern);
+  if (it == m_event_wild_list->end()) {
+    m_event_wild_list->push_back(pattern);
+  }
+}
+
+bool Rpl_event_ctx::event_needs_reenable(const std::string &&db,
+                                         const std::string &&event) const {
```
satya-bodapati: Why const R-value reference? Generally, const and r-value reference doesn't make sense (except for some corner case)

mysql-test/suite/sys_vars/r/all_vars.result
venkatesh-prasad-v: Good find! Thanks for checking this. I didn't know about this.

Looks like this intention of test has been misunderstood and has been recorded multiple times as we still find many variables without a `_basic` tests.

sql/rpl_event_ctx.cc
venkatesh-prasad-v: I think I added it to suppress a warning from the initial version of the unittest. So, it is a leftover from the initial implementation. Changing it to
```
bool Rpl_event_ctx::event_needs_reenable(const std::string &db,
                                         const std::string &event) const {
```

satya-bodapati (approved):
LGTM! Please squash before merge

mysql-test/suite/sys_vars/r/all_vars.result
satya-bodapati: right! looks like some variables crept in wihtout creation of _basic.test

sql/mysqld.cc
```diff
@@ -8835,6 +8836,16 @@ struct my_option my_long_options[] = {
...
      nullptr, nullptr, nullptr, GET_STR, REQUIRED_ARG, 0, 0, 0, nullptr, 0,
      nullptr},
+    {"replica-enable-event", OPT_REPLICA_ENABLE_EVENT,
+     "Tells the replication applier thread to enable the events that match "
+     "the specified wildcard pattern without setting it as SLAVESIDE_DISABLED. "
+     "To specify more than one event, use the directive multiple times, once "
+     "for each event. This will work for cross-database events. Example: "
+     "replica-enable-event=foo%.bar% will enable the events in all "
+     "databases on replica server that start with 'foo' and whose event names "
+     "start with 'bar'.",
```
satya-bodapati: Sorry, one last minute doubt. Should we mention that, this option should be used on events that are read-only? :thinking:

venkatesh-prasad-v: I think that is implied that using this feature with non-read-only events will cause inconsistency. But however, let us do it during the documentation phase.

satya-bodapati: Nothing in JIRA, github description, commit message tells about using this option ONLY for READ-ONLY events ;)

I feel, user has access to this help text immediately without opening docs etc. Lets extend the last line. Anyway, it didn't reach 80 chars :).  "Should be used for read-only events" (or some variations like this)

satya-bodapati: @venkatesh-prasad-v can you please reply to the above comment?

venkatesh-prasad-v: I'm sorry. I missed the ping for your last comment.

Sure. I will add a line in the end that it is recommended to use this feature only for read-only events to avoid data inconsistencies.

satya-bodapati: Thanks @venkatesh-prasad-v :+1:

satya-bodapati (approved):

venkatesh-prasad-v:
This will be handled as part of https://github.com/percona/percona-server/pull/4861

PR #4403, opened 2021-07-13
PS-1344: Rearrange checks in log_slow_statement() function (8.0)

oleksandr-kachan:
https://jira.percona.com/browse/PS-1344

Rearrange PS specific checks in log_slow_statement() function:
- Check state of thd->enable_slow_log before doing a bunch
  of PS specific checks. This allows to find out if slow log is
  enabled early and avoid unnecessary checks.
- Move collecting of query_exec_time up, do it right after
  thd->enable_slow_log is checked. Getting query_exec_time earlier
  before performing PS speciffic checks will give a more precise
  duration.
- Copy global vars into session ones before doing checks where this
  value is actually used, do it as the next step after query_exec_time
  calculation.

### DIFF
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -1880,15 +1880,43 @@ bool log_slow_applicable(THD *thd)
   if (unlikely(thd->in_sub_stmt))
     DBUG_RETURN(false);                         // Don't set time for sub stmt
 
+  /*
+    Do not log administrative statements unless the appropriate option is
+    set.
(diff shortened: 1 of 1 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1011/

percona-ysorokin (approved):
LGTM

PR #4418, opened 2021-07-20
PS-2346: Implement logging certain errors into slow query log

oleksandr-kachan:
https://jira.percona.com/browse/PS-2346

Added an option log_query_errors which enables logging queries failing
with certain error codes into a slow query log independently of their
execution time. The log_query_errors can be passed a comma separated list
of error codes: log_query_errors=code1,code2,code3. It can be set to
'all' to log all failing queries with nonzero error code.

### DIFF
--- a/mysql-test/suite/sys_vars/t/log_query_errors_basic.test
+++ b/mysql-test/suite/sys_vars/t/log_query_errors_basic.test
@@ -1,0 +1,170 @@
+SET @old_log_query_errors = @@global.log_query_errors;
+
+--echo # check default empty value
+SELECT @@global.log_query_errors;
+SELECT @@session.log_query_errors;
+SHOW GLOBAL VARIABLES LIKE 'log_query_errors';
+SHOW SESSION VARIABLES LIKE 'log_query_errors';
+--disable_warnings
+SELECT * FROM performance_schema.global_variables WHERE variable_name='log_query_errors';
+SELECT * FROM performance_schema.session_variables WHERE variable_name='log_query_errors';
+--enable_warnings
+
+--echo
+--echo # set global to ALL, session is unset
+SET GLOBAL log_query_errors=all;
+
+SELECT @@global.log_query_errors;
+SELECT @@session.log_query_errors;
+SHOW GLOBAL VARIABLES LIKE 'log_query_errors';
+SHOW SESSION VARIABLES LIKE 'log_query_errors';
+--disable_warnings
+SELECT * FROM performance_schema.global_variables WHERE variable_name='log_query_errors';
+SELECT * FROM performance_schema.session_variables WHERE variable_name='log_query_errors';
+--enable_warnings
+
+--echo
+--echo # unset global, set session to ALL
+SET GLOBAL log_query_errors="";
+SET SESSION log_query_errors=all;
+
+SELECT @@global.log_query_errors;
+SELECT @@session.log_query_errors;
+SHOW GLOBAL VARIABLES LIKE 'log_query_errors';
+SHOW SESSION VARIABLES LIKE 'log_query_errors';
+--disable_warnings
+SELECT * FROM performance_schema.global_variables WHERE variable_name='log_query_errors';
+SELECT * FROM performance_schema.session_variables WHERE variable_name='log_query_errors';
+--enable_warnings
+
+--echo
+--echo # set both to ALL
+SET GLOBAL log_query_errors=all;
+SET SESSION log_query_errors=all;
+
+SELECT @@global.log_query_errors;
+SELECT @@session.log_query_errors;
+SHOW GLOBAL VARIABLES LIKE 'log_query_errors';
+SHOW SESSION VARIABLES LIKE 'log_query_errors';
+--disable_warnings
+SELECT * FROM performance_schema.global_variables WHERE variable_name='log_query_errors';
+SELECT * FROM performance_schema.session_variables WHERE variable_name='log_query_errors';
+--enable_warnings
+
+--echo
+--echo # set to default
+SET SESSION log_query_errors=default;
+SET GLOBAL log_query_errors=default;
+SELECT @@global.log_query_errors;
+SELECT @@session.log_query_errors;
+
+--echo
+--echo # set some codes
+SET GLOBAL log_query_errors="1010,1011,1012";
+SET SESSION log_query_errors="2010,2011,2012";
+SELECT @@global.log_query_errors;
+SELECT @@session.log_query_errors;
+
+--echo
+--echo # set one code
+SET GLOBAL log_query_errors="1000";
+SET SESSION log_query_errors="2000";
+SELECT @@global.log_query_errors;
+SELECT @@session.log_query_errors;
+
+--echo
+--echo # set max allowed number of codes
+SET GLOBAL log_query_errors="1001,1002,1003,1004,1005,1006,1007,1008,1009,1010,1011,1012,1013,1014,1015,1016,1017,1018,1019,1020,1021,1022,1023,1024,1025,1026,1027,1028,1029,1030,1031,1032,1033,1034,1035,1036,1037,1038,1039,1040,1041,1042,1043,1044,1045,1046,1047,1048,1049,1050,1051,1052,1053,1054,1055,1056,1057,1058,1059,1060,1061,1062,1063,1064";
+SELECT @@global.log_query_errors;
+
+--echo
+--echo # set more than max allowed number of codes
+--error ER_TOO_MANY_ERROR_CODES
+SET GLOBAL log_query_errors="1001,1002,1003,1004,1005,1006,1007,1008,1009,1010,1011,1012,1013,1014,1015,1016,1017,1018,1019,1020,1021,1022,1023,1024,1025,1026,1027,1028,1029,1030,1031,1032,1033,1034,1035,1036,1037,1038,1039,1040,1041,1042,1043,1044,1045,1046,1047,1048,1049,1050,1051,1052,1053,1054,1055,1056,1057,1058,1059,1060,1061,1062,1063,1064,1065";
+
+--echo
+--echo # check ALL in upper and in lower case is accepted
+SET GLOBAL log_query_errors=all;
+SELECT @@global.log_query_errors;
+SET GLOBAL log_query_errors=ALL;
+SELECT @@global.log_query_errors;
+
+--echo
+--echo # check it fails with non-numeric code added to the list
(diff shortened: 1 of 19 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1069/

sql/system_variables.h
```diff
@@ -199,6 +201,144 @@ struct fragmentation_stats_t {
...
+
+      codes[error_index] = err_code;
+      ++error_index;
+
+      while (!my_isdigit(system_charset_info, *p) && *p) {
+        p++;
+      }
+    }
+
+    return true;
```
satya-bodapati, line 312: where do we check for "comma" separation here?

sql/system_variables.h
```diff
@@ -199,6 +201,144 @@ struct fragmentation_stats_t {
...
+      ;
+
+    clear_all();
+    uint8_t error_index = 0;
+
+    for (const char *p = val; *p;) {
+      long err_code;
+      if (!(p = str2int(p, 10, 0, LONG_MAX, &err_code))) {
+        break;
+      }
```
satya-bodapati, line 297: will SET overwrite the old values?

sql/system_variables.h
```diff
@@ -199,6 +201,144 @@ struct fragmentation_stats_t {
...
+class Query_errors_set {
+ public:
+  static constexpr uint8_t MAX_CODES_NUM = 64;
+  static constexpr uint32_t MAX_TEXT_LENGTH =
+      std::numeric_limits<uint32_t>::digits10 * MAX_CODES_NUM + MAX_CODES_NUM;
+  static const String LOG_ALL;
+
+ private:
+  bool is_all_set;
+  uint32_t codes[MAX_CODES_NUM];
```
satya-bodapati, line 213: Default constructor will initialize them to 0? Better to have constructor to initialize these

sql/system_variables.h
```diff
@@ -199,6 +201,144 @@ struct fragmentation_stats_t {
...
+      }
+
+      if (error_index == MAX_CODES_NUM) {
+        // Too many codes, should have been detected in check()
+        return true;
+      }
+
+      codes[error_index] = err_code;
+      ++error_index;
+
```
satya-bodapati, line 306: Shouldn't we set the rest of the codes array to 0?

mysql-test/suite/sys_vars/t/log_query_errors_basic.test
```diff
@@ -0,0 +1,110 @@
...
+--echo
+--echo # check it fails with non-numeric code added to the list
+--error ER_WRONG_VALUE_FOR_VAR
+SET GLOBAL log_query_errors="1010,aaaa,1012";
+
+--echo
+--echo # check spaces only string, should be processed as empty string
+SET GLOBAL log_query_errors=' ';
+SELECT @@global.log_query_errors;
+SET GLOBAL log_query_errors="  ";
```
satya-bodapati, line 103: Should we accept these configurations?
What about SET GLOBAL log_query_errors='' / ""

mysql-test/suite/sys_vars/t/log_query_errors_basic.test
```diff
@@ -0,0 +1,110 @@
...
+--echo
+--echo # check negative code
+--error ER_WRONG_VALUE_FOR_VAR
+SET GLOBAL log_query_errors="-1000";
+
+--echo
+--echo # check some wrong types
+--error ER_WRONG_TYPE_FOR_VAR
+SET GLOBAL log_query_errors=1.1;
+--error ER_WRONG_TYPE_FOR_VAR
```
satya-bodapati, line 117: what would SET GLOBAL log_query_errors=default do?
Also a test for SET PERSIST log_query_errors, restart and check

satya-bodapati (changes requested):

sql/system_variables.h
oleksandr-kachan: It doesn't check for comma explicitly. It converts number till first non digit symbol. Then skips it and tries to convert rest of the string. Similar approach is used in add_replica_skip_errors().

sql/system_variables.h
oleksandr-kachan: Yes, this set_codes() will clear all parse new string.

sql/system_variables.h
oleksandr-kachan: Will clear var value in Sys_var_errors_set constructor. Adding default values here leads to default constructor changes for the type. As a result this class wouldn't be able to use in System_variables as it supports only primitive types (memset() is used with System_variables).

sql/system_variables.h
oleksandr-kachan: It clears the whole array before parsing a string so rest of codes array is zeroed already.

mysql-test/suite/sys_vars/t/log_query_errors_basic.test
oleksandr-kachan: > Should we accept these configurations?

All spaces are skipped while parsing value so there seem to be no harm in this.

> What about SET GLOBAL log_query_errors='' / ""

This will fail with "wrong value" error. Will add it to the test.

mysql-test/suite/sys_vars/t/log_query_errors_basic.test
oleksandr-kachan: > what would SET GLOBAL log_query_errors=default do? 

Should restore it to default (empty) sate.

> Also a test for SET PERSIST log_query_errors, restart and check

Added for set persist_only as well

sql/system_variables.h
satya-bodapati: @oleksandr-kachan Hmm, if I do, "1-34,23-13" will it accept as a valid string? I guess validation is done before reaching this function? :thinking:

mysql-test/suite/sys_vars/t/log_query_errors_basic.test
```diff
@@ -0,0 +1,159 @@
...
+--echo # set to default
+SET SESSION log_query_errors=default;
+SET GLOBAL log_query_errors=default;
+SELECT @@global.log_query_errors;
+SELECT @@session.log_query_errors;
+
+--echo
+--echo # set some codes
+SET GLOBAL log_query_errors="1010,1011,1012";
+SET SESSION log_query_errors="2010,2011,2012";
```
satya-bodapati, line 64: 1. Do we need to treat them as string? ie, will it work without the double or single quotes? 
2. Also can you try with the longest number possible? 2^64
3. Does it check if an error code is a valid or not? ie if an error code is > the "max allowed errorcode " or not? etc
4. For a single error number, is it possible to do this SET GLOBAL log_query_errors=1010?

satya-bodapati (changes requested):

sql/system_variables.h
oleksandr-kachan: The "1-34,23-13" will not be accepted. I'll add it as one more test for a wrong value.

> I guess validation is done before reaching this function?

Looks like I misunderstood your question previously. Validation is done in Query_errors_set::check() method, it makes sure there are only digits and "," in a string.

mysql-test/suite/sys_vars/t/log_query_errors_basic.test
oleksandr-kachan: > Do we need to treat them as string? ie, will it work without the double or single quotes?

Yes, it should be a string

> Also can you try with the longest number possible? 2^64

There was no check for biggest acceptable number. I've added one.

> Does it check if an error code is a valid or not? ie if an error code is > the "max allowed errorcode " or not? 

No, at the moment it doesn't know about existing error codes.

> For a single error number, is it possible to do this SET GLOBAL log_query_errors=1010?

No, it still should be a string - "1010"

satya-bodapati (commented):
@oleksandr-kachan LGTM! can you please submit to Jenkins again on the latest commit?

oleksandr-kachan:
> @oleksandr-kachan LGTM! can you please submit to Jenkins again on the latest commit?

Sure, https://ps80.cd.percona.com/job/percona-server-8.0-param/1114/. I've also rebased it on top of current 8.0.

https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34836/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34835/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34834/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34839/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34840/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34841/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34843/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34842/

https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34862/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34861/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34860/

satya-bodapati:
> > @oleksandr-kachan LGTM! can you please submit to Jenkins again on the latest commit?
> 
> Sure, https://ps80.cd.percona.com/job/percona-server-8.0-param/1114/. I've also rebased it on top of current 8.0.
> 
> https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34836/ https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34835/ https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34834/ https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34839/ https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34840/ https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34841/ https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34843/ https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34842/
> 
> https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34862/ https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34861/ https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/34860/

@oleksandr-kachan All Debug builds failed??

oleksandr-kachan:
> @oleksandr-kachan All Debug builds failed??

@satya-bodapati, that is due to dropped AWS workers. I started all failed via pipeline job (see that long list). There was no other way to get results for all platforms.

satya-bodapati (approved):

PR #4422, opened 2021-07-23
Port of security or crash fixes from MySQL 5.7.33 to 5.7.34 

avbelov23:


sql/sql_yacc.yy
```diff
@@ -3152,8 +3167,34 @@ sp_decl:
...
+              if (i == first_var_num) {
+                first_spvar = spvar;
+              } else if (has_default_clause && !is_const_item) {
+                Item_splocal *item =
+                  new (thd->mem_root)
+                    Item_splocal(first_spvar->name, first_spvar->offset,
+                                       first_spvar->type, 0, 0);
+                if (item == NULL)
+                  MYSQL_YYABORT; // OOM
+#ifndef NDEBUG
```
kamil-holubicki: I think we need to change it to DBUG_OFF, like in commit 6051df17b88a9151910bb2cf2e70ca8a359de8cb
?
m_sp member is under #ifndef DBUG_OFF in item.h

storage/innobase/handler/ha_innodb.cc
```diff
@@ -5966,6 +5974,7 @@ ha_innobase::open(
 		/* Mark this table as corrupted, so the drop table
 		or force recovery can still use it, but not others. */
 		ib_table->corrupted = true;
+		free_share(share);
```
kamil-holubicki: I'm not sure about the flow here.
We are calling free_share, then if we get into the next 'if (UNIV_UNLIKELY)' we will call free_share again...

I think the solution is in commit 9a507dc4add6c3676d7aa18bcf4dc9c1ed54b604 from PS

mysys/my_conio.c
```diff
@@ -210,8 +211,8 @@ void
 my_win_console_write(const CHARSET_INFO *cs, const char *data, size_t datalen)
 {
   static wchar_t u16buf[MAX_CONSOLE_LINE_SIZE + 1];
-  size_t nchars= my_mbstou16s(cs, (const uchar *) data, datalen,
-                              u16buf, sizeof(u16buf));
+  size_t nchars = my_mbstou16s(cs, (const uchar *) data, datalen, u16buf,
```
kamil-holubicki: s/nchars =/nchars=

kamil-holubicki (changes requested):
Hi, thanks for working on these backports. 
I have created release branch release-5.6.51-92.0. The PR should be against this branch, not the already released version. Shouldn't be a big deal to change, as there is no ongoing development on 5.6.

Please also see my other comments.

storage/innobase/handler/ha_innodb.cc
avbelov23: Condition is `if (UNIV_UNLIKELY(**ib_table** &&`, but `ib_table = NULL;` after `free_share(share);`

kamil-holubicki: I agree in this case, indeed. But why we introduced 'free_share_and_nullify()' in the mentioned commit? Are we safe in 5.6 without doing this?
@ldonoso, could you advise, please?

ldonoso: @avbelov23 , please, apply the changes on the commit above. Some other functions of Percona Server expect `share` to be available if not null.

kamil-holubicki:
LGTM. I started Jenkins job to check if everything is OK
https://ps56.cd.percona.com/job/percona-server-5.6-param/151/

kamil-holubicki:
Hi @avbelov23 , unfortunately after applying the patch the following tests fail with segfault:
innodb.xtradb_rseg_bug_1260259
innodb.innodb-wl5980-linux
innodb.innodb_mysql
innodb.percona_changed_page_bmp_requests_debug
innodb.percona_bug_1433197
innodb.percona_changed_page_bmp
innodb.percona_changed_page_bmp_1364315

kamil-holubicki:
https://ps56.cd.percona.com/job/percona-server-5.6-param/153/

kamil-holubicki (approved):

PR #4437, opened 2021-08-17
PS-7778 Wrongly handled definer for triggers when PS used

i-rinat:
It's possible to get spurious access denied errors when a query causes a trigger to execute, and is executed as a prepared statement. Access checking logic tries to check if the current user has access to all tables that a trigger touches, even if that trigger is bound to run in a context of another user.

As all additional tables are added to the list after all own tables of a query, it's enough to stop traversing the table list once any of non-own tables are met.

This regression was introduced in 67c3c70e4895874d43434f1df556f9f30d781b48.

### DIFF
--- a/mysql-test/t/ps_trigger_access_check_regression.test
+++ b/mysql-test/t/ps_trigger_access_check_regression.test
@@ -1,0 +1,154 @@
+#
+# PS-7778  Wrongly handled definer for triggers when prepared statements are used
+#
+
+CREATE USER 'logger'@'%' IDENTIFIED BY '';
+CREATE USER 'user'@'%' IDENTIFIED BY '';
+
+CREATE DATABASE test1;
+USE test1;
+
+# A table that only 'logger' can access.
+CREATE TABLE logger_internal (data VARCHAR(100));
+INSERT INTO logger_internal SET data = 'secret';
+GRANT ALL PRIVILEGES ON logger_internal TO 'logger'@'%';
+
+# A table that nobody can access.
+CREATE TABLE forbidden_for_everyone (data VARCHAR(100));
+INSERT INTO forbidden_for_everyone SET data = '______';
+
+# A table 'logger' logs to.
+CREATE TABLE log (val VARCHAR(100));
+GRANT ALL PRIVILEGES ON log TO 'logger'@'%';
+
+--echo # -----------------------------------------------------------------------
+--echo # 1. Trigger SELECTs a table that 'user' can't access, but 'logger' can.
+--echo #
+
+CREATE TABLE data (val VARCHAR(100));
+INSERT INTO data (val) VALUES ('original-data');
+GRANT ALL PRIVILEGES ON data TO 'user'@'%';
+
+GRANT TRIGGER ON data TO 'logger'@'%';
+CREATE DEFINER = 'logger'@'%' TRIGGER trg
+  AFTER UPDATE ON data
+  FOR EACH ROW
+    INSERT INTO log (val)
+      (SELECT CONCAT('trg/1/', data) FROM logger_internal LIMIT 1);
+
+--connect (con_user, localhost, user, , )
+USE test1;
+
+UPDATE data SET val = 'new-data';
+PREPARE s FROM 'UPDATE data SET val = \'new-data\'';
+EXECUTE s;
+
+--echo # -----------------------------------------------------------------------
+--echo # 2. Trigger SELECTs something both 'user' and 'logger' cannot read.
+--echo #
+
+--connection default
+DROP TABLE data;
+CREATE TABLE data (val VARCHAR(100));
+INSERT INTO data (val) VALUES ('original-data');
+GRANT ALL PRIVILEGES ON data TO 'user'@'%';
+
+GRANT TRIGGER ON data TO 'logger'@'%';
+CREATE DEFINER = 'logger'@'%' TRIGGER trg
+  AFTER UPDATE ON data
+  FOR EACH ROW
+    INSERT INTO log (val)
+      (SELECT CONCAT('trg/2/', data) FROM forbidden_for_everyone LIMIT 1);
+
+--connection con_user
+
+--error ER_TABLEACCESS_DENIED_ERROR
+UPDATE data SET val = 'new-data';
+PREPARE s FROM 'UPDATE data SET val = \'new-data\'';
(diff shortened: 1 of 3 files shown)


altmannmarcelo:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/1035/

altmannmarcelo (changes requested):
Hi @i-rinat 

Can you please cover a few more tests cases:

1) when definer is different from user executing the trigger - trigger update a different table and definer lack privileges on update different table.
2) definer has no SELECT access on the table and trigger uses OLD.column NEW.column
3) definer has no UPDATE access on the table and trigger uses `SET NEW.column`
4) definer has TRIGGER privilege on the table, creates the trigger, then remove the TRIGGER privilege from definer user and execute the INSERT/UPDATE using a different user.
4) Repeat the above with PS and without PS

i-rinat:
@altmannmarcelo,

Sure, will do the changes.
Also, as far as I can understand, all additional test cases you mentioned should generate access errors. Is that right?

altmannmarcelo:
Correct. This is to validate we are not by-passing any requirement ACL restrictions with this patch.

mysql-test/t/ps_trigger_access_check_regression.test
```diff
@@ -0,0 +1,52 @@
...
+
+--echo # Create users 'user01' and 'user02'.
+CREATE USER 'user01'@'%' IDENTIFIED BY 'user01';
+CREATE USER 'user02'@'%' IDENTIFIED BY 'user02';
+GRANT ALL PRIVILEGES ON *.* to 'user01'@'%';
+
+--echo #
+--echo # Fill tables.
+CREATE DATABASE test01;
+use test01;
```
kamil-holubicki: Multiple occurrences in the file:
s/use/USE

mysql-test/t/ps_trigger_access_check_regression.test
```diff
@@ -0,0 +1,52 @@
...
+--echo #
+--echo # Log in as 'user02'.
+--connect (con_user02, localhost, user02, user02, )
+--connection con_user02
+
+use test;
+
+--echo #
+--echo # Update should succeed.
+update tbl01 set a = now();
```
kamil-holubicki: s/update/UPDATE
s/set/SET

mysql-test/t/ps_trigger_access_check_regression.test
```diff
@@ -0,0 +1,52 @@
...
+--echo #
+--echo # Same update as a prepare-execute pair should also succeed.
+prepare s from 'update tbl01 set a = now()';
```
kamil-holubicki: From here to the end of the file: all SQL keywords should be upper case

sql/sql_select.cc
```diff
@@ -946,9 +946,10 @@ bool Sql_cmd_select::check_privileges(THD *thd) {
 
 bool Sql_cmd_dml::check_all_table_privileges(THD *thd) {
   // Check for all possible DML privileges
+  TABLE_LIST *first_not_own_table = lex->first_not_own_table();
 
-  for (TABLE_LIST *tr = lex->query_tables; tr != nullptr;
-       tr = tr->next_global) {
+  for (TABLE_LIST *tr = lex->query_tables;
```
kamil-holubicki, line 934: Is only select affected, or there are others as well?

mysql-test/t/ps_trigger_access_check_regression.test
i-rinat: I've rewritten the test, and hopefully have all keywords in upper case now.

sql/sql_select.cc
i-rinat: This bug affect UPDATEs, INSERTs and DELETEs. It could hypothetically affect SELECTs, but as far as I know, it's not possible to create a trigger on SELECTs.

Although this code is in `sql_select.cc` file, it's the implementation of the `Sql_cmd_dml` class, so it's inherited by all DML commands. All descendant classes have own implementation of `<...>::check_privileges()`, but they all also call `Sql_cmd_dml::check_all_table_privileges()`, so changes here affect all DML commands.

i-rinat:
> cover a few more tests cases

Added mentioned test cases. Also rewritten the original test case to make it more readable. Additional test cases do generate errors as expected.

Tried to find if there are any other ways to trigger a similar bug. But as one can only attach a trigger only to insert, update, and delete operations, only those three may cause a similar issue. Handlers of all those commands call `Sql_cmd_dml::check_all_table_privileges()` during the privilege checking phase, so changes in `Sql_cmd_dml::check_all_table_privileges()` are enough to cover all cases.

mysql-test/t/ps_trigger_access_check_regression.test
i-rinat: I've rewritten the test, and hopefully have all keywords in upper case now.

mysql-test/t/ps_trigger_access_check_regression.test
i-rinat: I've rewritten the test, and hopefully have all keywords in upper case now.

altmannmarcelo (approved):
LGTM.
Also, I see the same logic (checking for first_not_own_table) been used and various other different places.
Jenkins looks good when compared to trunk job too.

altmannmarcelo:
@i-rinat can I ask you just to adjust the commit message:

```
Title: Fixed (JIRA KEY) - (JIRA Title)

JIRA LINK

Problem:
Description of the problem

Fix:
Description of the fix
```

One example from PXB - https://github.com/percona/percona-xtrabackup/pull/1172/commits/bf91d8989b1c12cb3c4e86b284a6de4dad57d1eb

i-rinat:
@altmannmarcelo

I've updated the commit message to include a description of the problem, why it happens, and how it's fixed. Could you please take a look?

Since the moment I initially created the pull request, `8.0` was updated. Do I need to rebase the patch onto the current state of `8.0`? Any other branch?

kamil-holubicki (approved):

altmannmarcelo:
@i-rinat can you please rebase it on top of https://github.com/percona/percona-server/tree/release-8.0.26-16 and I will get it merged.

Thanks

i-rinat:
> rebase it on top of <...> release-8.0.26-16

Rebased.

PR #4441, opened 2021-08-18
PS-7811 merge: Merge MySQL 8.0.26 (fix minor use-after-free error in

ldonoso:
mysqldump.cc)

The cause is probabibly a merging error.

The solution follows the same pattern as upstream:
https://github.com/mysql/mysql-server/blob/beb865a960b9a8a16cf999c323e46c5b0c67f21f/client/mysqldump.cc#L4986-L5002

### DIFF
--- a/client/mysqldump.cc
+++ b/client/mysqldump.cc
@@ -37,6 +37,7 @@
 #include <sys/types.h>
 #include <forward_list>
 #include <list>
+#include <memory>
 #include <string>
 
 #include "client/client_priv.h"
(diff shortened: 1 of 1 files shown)


client/mysqldump.cc
```diff
@@ -5737,23 +5737,24 @@ static int do_show_master_status(MYSQL *mysql_con,
   char binlog_pos_file[FN_REFLEN];
   char binlog_pos_offset[LONGLONG_LEN + 1];
   char *file, *offset;
+  MYSQL_RES *master = nullptr;
```
oleksandr-kachan: Would it be possible to use unique_ptr with custom deleter here?

ldonoso: Yes, of course. Please, check again.

oleksandr-kachan (approved):

PR #4446, opened 2021-08-20
Allow the rocksdb_lock_wait_timeout variable to set dynamically throu…

ovaistariq:
Allow the rocksdb_lock_wait_timeout variable to set dynamically through the SET_VAR syntax.

Summary:
Allow the rocksdb_lock_wait_timeout variable to be set dynamically through the SET_VAR syntax.
This extend https://jira.percona.com/browse/PS-3966.


george-lorch:
Hello Ovais, thanks for the contribution and sorry for the delay in processing.  We originally had some concern about the change which lead to the stall in processing so that we could do some research.  The research has alleviated our concerns and we should be able to move forward with the contribution, which means running it through our internal CI/CD and preparing it and sending it upstream to Facebook.  Thanks again.

kamil-holubicki:
Hi @ovaistariq , 
thank you for your contribution. I had to refactor the test a little bit. It will be included as the part of https://github.com/percona/percona-server/pull/4639

ovaistariq:
Thanks for refactoring it and including the change.

PR #4451, opened 2021-08-31
PS-7706 Add options to explicitly enable TokuDB and TokuBackup post deprecation

i-rinat:
Introducing new BOOL options 'tokudb_enabled' and 'tokudb_backup_enabled' that enable TokuDB storage and TokuDB backup plugin respectively. Both options are FALSE by default.

Adding a new test suite, called tokudb_deprecation, to check that the storage and the plugin are disabled by default.


george-lorch:
This looks perfect, jenkins run before merging https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/1040/

george-lorch (approved):

PR #4452, opened 2021-09-01
PS-7811 merge: Merge MySQL 8.0.26 (fix update of trx_sys->min_active_…

ldonoso:
…trx_id)

When assigning a `trx->preallocated_id` the value of
`trx_sys->min_active_trx_id` needs to be updated.

This issue does not manifest in MySQL because the use of
`trx->preallocated_id` is a functionality of Percona and the
preallocated_id is not provided in ascending order.

This issue was detected by a failing assertion introduced in
e66d48b and occurred when testing `main.clone_consistent_snapshot`.

### DIFF
--- a/storage/innobase/trx/trx0trx.cc
+++ b/storage/innobase/trx/trx0trx.cc
@@ -1221,17 +1221,20 @@ void trx_assign_rseg_durable(trx_t *trx) {
 /** Assign an id for this RW transaction and insert it into trx_sys->rw_trx_ids
 @param trx	transaction to assign an id for */
 static void trx_assign_id_for_rw(trx_t *trx) {
-  ut_ad(mutex_own(&trx_sys->mutex));
+  ut_ad(trx_sys_mutex_own());
 
   trx->id =
       trx->preallocated_id ? trx->preallocated_id : trx_sys_allocate_trx_id();
 
   if (trx->preallocated_id) {
-    // Maintain ordering in rw_trx_ids
-    trx_sys->rw_trx_ids.insert(
-        std::upper_bound(trx_sys->rw_trx_ids.begin(), trx_sys->rw_trx_ids.end(),
-                         trx->id),
-        trx->id);
+    // preallocated_id might not be received in ascending order,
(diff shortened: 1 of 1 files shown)


storage/innobase/trx/trx0trx.cc
```diff
@@ -1221,12 +1221,16 @@ void trx_assign_rseg_durable(trx_t *trx) {
...
 @param trx	transaction to assign an id for */
 static void trx_assign_id_for_rw(trx_t *trx) {
-  ut_ad(mutex_own(&trx_sys->mutex));
+  ut_ad(trx_sys_mutex_own());
 
   trx->id =
       trx->preallocated_id ? trx->preallocated_id : trx_sys_allocate_trx_id();
 
   if (trx->preallocated_id) {
+    if (trx->id < trx_sys->min_active_trx_id.load()) {
```
percona-ysorokin: This is a perfectly valid fix. However, I would like to suggest a minor improvement.
Similarly to what Oracle did here
https://github.com/ldonoso/percona-server/commit/e66d48b0c73d5fec278f81784bd5697502990263#diff-52a34c53a0da3f03047e6e1e58eadefe983c05059c00414787c80af5f7c0a486R1752
we can read/update min_active_trx_id (touch the cache line) only when it is indeed needed.
Basically, if the `trx_sys->rw_trx_ids.insert()` returns `trx_sys->rw_trx_ids.begin()`
Something like
```c++
auto insert_result = trx_sys->rw_trx_ids.insert(...);
if (insert_result == trx_sys->rw_trx_ids.begin())
  trx_sys->min_active_trx_id.store(trx->id);
```

percona-ysorokin (changes requested):
LGTM

storage/innobase/trx/trx0trx.cc
ldonoso: Good point @percona-ysorokin , it is definitely a better way.

percona-ysorokin (approved):
LGTM

satya-bodapati (approved):

PR #4460, opened 2021-09-10
PS-7806: Column compression breaks async replication on PS

nitendra-bhosle:
https://jira.percona.com/browse/PS-7806

Analysis
compress_heap memory was released inside
method row_sel_store_mysql_rec , even though ,it was reused.
Which caused incorrect bin logging for comressed column,
leading to async replication to break.

Fix

Removed compress_heap free statements,
hence memory is reused.


nitendra-bhosle:
Jenkins : https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/1050/

storage/innobase/row/row0sel.cc
```diff
@@ -2987,8 +2987,6 @@ bool row_sel_store_mysql_rec(byte *mysql_rec, row_prebuilt_t *prebuilt,
     mem_heap_empty(blob_heap);
   }
 
-  if (UNIV_LIKELY_NULL(prebuilt->compress_heap))
-    row_mysql_prebuilt_free_compress_heap(prebuilt);
```
satya-bodapati, line 2991: Can you plz explain the following?
1. How it is reused? This function is supposed to transfer/convert a innodb record to server/mysql record
2. When is it freed?
3. How about emptying it? Like blob_heap above. Can't we do the same?

satya-bodapati (changes requested):

mysql-test/suite/binlog/t/binlog_stm_delete_with_compress_column.test
```diff
@@ -0,0 +1,5 @@
+# This is a wrapper for binlog_delete_with_compress_column.test so that the same
+# test case can be used for both statement and row based logging
+
+-- source include/have_binlog_format_mixed_or_statement.inc
+-- source extra/binlog_tests/binlog_delete_with_compress_column.test
```
venkatesh-prasad-v: I don't think there is no any need to run the test in statement format, as corrupted data is not propagated in statement based replication. So, I suggest you to please remove this test.

mysql-test/extra/binlog_tests/binlog_delete_with_compress_column.test
```diff
@@ -0,0 +1,20 @@
+#
+# PS-7806: Column compression breaks async replication on PS
+#
```
venkatesh-prasad-v: The tests pass even without fix, so please add a replication test (rpl suite) as this bug was initially found in a replication scenario.
```
[ 14%] binlog.binlog_row_delete_with_compress_column 'mix'  [ skipped ]  Doesn't support --binlog-format = 'mixed'
[ 28%] binlog.binlog_stm_delete_with_compress_column 'mix'  [ pass ]   4712
[ 42%] binlog.binlog_row_delete_with_compress_column 'row'  [ pass ]   5768
[ 57%] binlog.binlog_stm_delete_with_compress_column 'row'  [ skipped ]  Doesn't support --binlog-format = 'row'
[ 71%] binlog.binlog_row_delete_with_compress_column 'stmt'  [ skipped ]  Doesn't support --binlog-format = 'statement'
[ 85%] binlog.binlog_stm_delete_with_compress_column 'stmt'  [ pass ]   4456
[100%] shutdown_report                           [ pass ]
```

venkatesh-prasad-v (changes requested):

storage/innobase/row/row0sel.cc
```diff
@@ -2987,8 +2987,9 @@ bool row_sel_store_mysql_rec(byte *mysql_rec, row_prebuilt_t *prebuilt,
...
+  if (prebuilt->compress_heap != nullptr) {
+    mem_heap_empty(prebuilt->compress_heap);
```
satya-bodapati, line 2991: If we are reusing, even emptying is a problem isn't it? I think partitions are supposed to their partition-specific blob heaps? Can you please check again?

satya-bodapati (approved):

storage/innobase/row/row0sel.cc
satya-bodapati: Sorry used the wrong button.. "not approved yet" :)

mysql-test/suite/rpl/t/rpl_column_compression_async_rpl.test
```diff
@@ -0,0 +1,32 @@
+#
+# PS-7806 : Column compression breaks async replication on PS
+#
+
+--source include/not_group_replication_plugin.inc
+--source include/master-slave.inc
+
+sync_slave_with_master;
+
+--source include/rpl_connection_master.inc
```
venkatesh-prasad-v: You can remove these lines. The rationale is 
1. `--source include/master-slave.inc` by default leaves the connection on master, so there is no need to change the connection to master.
2. As we haven't done anything in the test, there is nothing to sync.

mysql-test/suite/rpl/t/rpl_column_compression_async_rpl.test
```diff
@@ -0,0 +1,32 @@
+#
+# PS-7806 : Column compression breaks async replication on PS
+#
+
+--source include/not_group_replication_plugin.inc
+--source include/master-slave.inc
```
venkatesh-prasad-v, line 5: Also, please source `include/have_binlog_format_row.inc` as there is no need to run this test in all formats.

mysql-test/suite/rpl/t/rpl_column_compression_async_rpl.test
```diff
@@ -0,0 +1,32 @@
+#
+# PS-7806 : Column compression breaks async replication on PS
+#
+
+--source include/not_group_replication_plugin.inc
```
venkatesh-prasad-v: This is not required.

mysql-test/suite/rpl/t/rpl_column_compression_async_rpl.test
```diff
@@ -0,0 +1,32 @@
...
+
+INSERT INTO t  ( ip_col ,c0 ,lb1 ,d2 ,lb3 ,f4 ,v5 ) VALUES( 4314, 'C', '9v2pBuu9PACawxmCC5eoYEsDHNmOaJLKqoW994gvKqEO35F8goMG9o', 0.00000, 'vio1lMJ3e9sGa5tKRz1iJwsHXjKw8uQelcv9o5tBzmgP2YDjG4bB5OK9IaBpWaf3I1Muwd06FFgT8pZKc2GyHCQ3bcqab83yKXe8BWpWFpw3sOb1rdM7uOCvGmSvGKzfBgaPfGCgqbx29kTcQxOzYJd5zGuEd32gQfocLNdttxPiBjyIIPg3kTI8US73sNN4UoEzCcC4XZJ8lWKAqVbDQmNDwV7kZZGb7s8JMjsJ0XTcAKf0sl9lw5L6albce2GqwsSDv', 0.00, '3VK82eQXa' ) ;
+
+DELETE FROM t WHERE d2 = 0.00000 ;
+
+DROP TABLE t;
+
+DROP DATABASE test7806;
+
+--source include/rpl_connection_slave.inc
```
venkatesh-prasad-v: Please source `sync_slave_sql_with_master.inc` file to make sure that the updates reach replica and are applied correctly on replica.

mysql-test/suite/rpl/t/rpl_column_compression_async_rpl.test
```diff
@@ -0,0 +1,32 @@
...
+
+let $last_sql_error= query_get_value(SHOW REPLICA STATUS, Last_SQL_Error, 1);
+--echo $last_sql_error
```
venkatesh-prasad-v: Can you please use `rpl_diff.inc` to ensure data consistency across both the source and replica.

venkatesh-prasad-v (changes requested):

mysql-test/suite/rpl/t/rpl_column_compression_async_rpl.test
```diff
@@ -0,0 +1,27 @@
+#
+# PS-7806 : Column compression breaks async replication on PS
+#
+--source include/master-slave.inc
+--source include/have_binlog_format_row.inc
```
venkatesh-prasad-v, line 4: Please change the order of inclusion of inc files.
```suggestion
--source include/have_binlog_format_row.inc
--source include/master-slave.inc
```

mysql-test/suite/rpl/t/rpl_column_compression_async_rpl.test
```diff
@@ -0,0 +1,27 @@
...
+
+INSERT INTO t  ( ip_col ,c0 ,lb1 ,d2 ,lb3 ,f4 ,v5 ) VALUES( 4314, 'C', '9v2pBuu9PACawxmCC5eoYEsDHNmOaJLKqoW994gvKqEO35F8goMG9o', 0.00000, 'vio1lMJ3e9sGa5tKRz1iJwsHXjKw8uQelcv9o5tBzmgP2YDjG4bB5OK9IaBpWaf3I1Muwd06FFgT8pZKc2GyHCQ3bcqab83yKXe8BWpWFpw3sOb1rdM7uOCvGmSvGKzfBgaPfGCgqbx29kTcQxOzYJd5zGuEd32gQfocLNdttxPiBjyIIPg3kTI8US73sNN4UoEzCcC4XZJ8lWKAqVbDQmNDwV7kZZGb7s8JMjsJ0XTcAKf0sl9lw5L6albce2GqwsSDv', 0.00, '3VK82eQXa' ) ;
+
+DELETE FROM t WHERE d2 = 0.00000 ;
+
+--source include/sync_slave_sql_with_master.inc
+--let $rpl_diff_statement = SELECT * FROM test7806.t;
+--source include/rpl_diff.inc
+
+connection master;
```
venkatesh-prasad-v: Please source `rpl_connection_master.inc` instead.

venkatesh-prasad-v (changes requested):

venkatesh-prasad-v (approved):
Looks fine with the test. But please wait for final approval from @satya-bodapati

storage/innobase/row/row0sel.cc
satya-bodapati: @nitendra-bhosle can you please reply to these questions?

I don't understand the issue. Specifically why free is a problem. and empty is not?
AFAIU, empty doesn't the remove the first block. So may be the compressed heap is staying within the first block of heap.

Can you please create a blob text such that it occupies more than the first block in heap?

This is outdated changes.

satya-bodapati (changes requested):
Please see the question on thread

nitendra-bhosle:
@satya-bodapati  :  https://github.com/percona/percona-server/pull/4460#discussion_r714653808 
This is outdated. I have removed both empty and free heap statements.

satya-bodapati:
> @satya-bodapati : [#4460 (comment)](https://github.com/percona/percona-server/pull/4460#discussion_r714653808)
> This is outdated. I have removed both empty and free heap statements.

Well, I still the empty function.

If you remove this empty, Do we call empty at the end of statement execution?
Can you explain how the 'reuse' is happening?

nitendra-bhosle:
> > @satya-bodapati : [#4460 (comment)](https://github.com/percona/percona-server/pull/4460#discussion_r714653808)
> > This is outdated. I have removed both empty and free heap statements.
> 
> Well, I still the empty function.
> 
> If you remove this empty, Do we call empty at the end of statement execution?

This empty statement , I have added just to be consistent with blob_heap .
If we remove this, we will not call empty at the end of statement execution, as it's already being 
getting free's inside : row_prebuilt_free 

> Can you explain how the 'reuse' is happening?

We are allocating memory column_get_compress_header and row_compress_column 
using statements like : 
if (!prebuilt->compress_heap)
    prebuilt->compress_heap = mem_heap_create(ut_max(UNIV_PAGE_SIZE, buf_len));

I reffered reuse , because this is diverging from blob_heap memory management in a sense that memory is getting
allowcated inside these two methods , and reuse might be happening , otherwise , no need to allocated and check whether it's already allocated inside other method.

bratao:
Any chance of this making into 8.0.26 release? I just got bitten by this

nitendra-bhosle:
> Any chance of this making into 8.0.26 release? I just got bitten by this

I think this Issue will be part of 8.0.27 release

storage/innobase/row/row0sel.cc
satya-bodapati: Can you please bring this back and test? Only then we will know if your new fix is working or not?

satya-bodapati (changes requested):

satya-bodapati (changes requested):
Also please test PS-7657 by reverting the fix and using this fix

nitendra-bhosle:
> Also please test PS-7657 by reverting the fix and using this fix

Tested , All tests are passing

satya-bodapati:
> > Also please test PS-7657 by reverting the fix and using this fix
> 
> Tested , All tests are passing

Thats great news! Please include the revert of PS-7657 code change in this PR

nitendra-bhosle:
> > > Also please test PS-7657 by reverting the fix and using this fix
> > 
> > 
> > Tested , All tests are passing
> 
> Thats great news! Please include the revert of PS-7657 code change in this PR

Revert is conflicting with this change : 
if (UNIV_LIKELY_NULL(prebuilt->compress_heap))
    row_mysql_prebuilt_free_compress_heap(prebuilt);

Earlier we had 
if (UNIV_LIKELY_NULL(prebuilt->compress_heap))
mem_heap_empty(prebuilt->compress_heap);

satya-bodapati:
> > > > Also please test PS-7657 by reverting the fix and using this fix
> > > 
> > > 
> > > Tested , All tests are passing
> > 
> > 
> > Thats great news! Please include the revert of PS-7657 code change in this PR
> 
> Revert is conflicting with this change : if (UNIV_LIKELY_NULL(prebuilt->compress_heap)) row_mysql_prebuilt_free_compress_heap(prebuilt);
> 
> Earlier we had if (UNIV_LIKELY_NULL(prebuilt->compress_heap)) mem_heap_empty(prebuilt->compress_heap);

You can add a new commit which will bring to original state.

nitendra-bhosle:
Also this change of PS-7657 is still required - 
if (UNIV_LIKELY_NULL(prebuilt->compress_heap))
    mem_heap_empty(prebuilt->compress_heap);

Inside function row_insert_for_mysql_using_ins_graph

satya-bodapati:
> Also this change of PS-7657 is still required - if (UNIV_LIKELY_NULL(prebuilt->compress_heap)) mem_heap_empty(prebuilt->compress_heap);
> 
> Inside function row_insert_for_mysql_using_ins_graph

Yes @nitendra-bhosle, please bring that emptying back. with the new logic, emptying or freeing compress_heap shouldn't matter.

satya-bodapati:
> > Also this change of PS-7657 is still required - if (UNIV_LIKELY_NULL(prebuilt->compress_heap)) mem_heap_empty(prebuilt->compress_heap);
> > Inside function row_insert_for_mysql_using_ins_graph
> 
> Yes @nitendra-bhosle, please bring that emptying back. with the new logic, emptying or freeing compress_heap shouldn't matter.

@nitendra-bhosle I still don't see the emptying code in row_insert_for_mysql_using_ins_graph()
```
  if (UNIV_LIKELY_NULL(prebuilt->compress_heap))
    mem_heap_empty(prebuilt->compress_heap);
```
This was present before PS-7657 and it should be brought back. It will not have any effect with your latest fix.

satya-bodapati (changes requested):
Please check my latest comment

nitendra-bhosle:
> > Also this change of PS-7657 is still required - if (UNIV_LIKELY_NULL(prebuilt->compress_heap)) mem_heap_empty(prebuilt->compress_heap);
> > Inside function row_insert_for_mysql_using_ins_graph
> 
> Yes @nitendra-bhosle, please bring that emptying back. with the new logic, emptying or freeing compress_heap shouldn't matter.




> > > Also this change of PS-7657 is still required - if (UNIV_LIKELY_NULL(prebuilt->compress_heap)) mem_heap_empty(prebuilt->compress_heap);
> > > Inside function row_insert_for_mysql_using_ins_graph
> > 
> > 
> > Yes @nitendra-bhosle, please bring that emptying back. with the new logic, emptying or freeing compress_heap shouldn't matter.
> 
> @nitendra-bhosle I still don't see the emptying code in row_insert_for_mysql_using_ins_graph()
> 
> ```
>   if (UNIV_LIKELY_NULL(prebuilt->compress_heap))
>     mem_heap_empty(prebuilt->compress_heap);
> ```
> 
>This was present before PS-7657 and it should be brought back. It will not have any effect with your latest fix.

After my investigation today, I found that above change is required to fix - PS-7657.It will not have any effect on my fix for PS-7806.
But this will break for PS-7657 , hence I kept it as is.

With current patch , which is part of these commits , all tests for PS--7657 and PS-7806 are passing

storage/innobase/row/row0mysql.cc
```diff
@@ -2030,6 +2030,9 @@ static dberr_t row_insert_for_mysql_using_ins_graph(const byte *mysql_rec,
 
   row_mysql_convert_row_to_innobase(node->row, prebuilt, mysql_rec, &blob_heap);
 
+  if (UNIV_LIKELY_NULL(prebuilt->compress_heap))
+    mem_heap_empty(prebuilt->compress_heap);
+
```
satya-bodapati, line 2035: previously before 7657, the empty was before trx->op_info = "inserting" line. Any reason, it is placed here? Does it have to be done after conversion of row to innodb format?

Can it be moved to before trx->op_info?

satya-bodapati (changes requested):

nitendra-bhosle:
> previously before 7657, the empty was before trx->op_info = "inserting" line. Any reason, it is placed here? Does it have to be done after conversion of row to innodb format?
> 
> Can it be moved to before trx->op_info?

We can't move it before trx->op_info , as row_mysql_convert_row_to_innobase call uses prebuilt->compress_heap

storage/innobase/row/row0mysql.cc
satya-bodapati: I see that you replied below.
So this means, we cannot free a buffer from the previous handler call and it is still in use?

Ideally, we should be able to free **before** the mysql_to_innodb row format conversion. The buffer would have last row insertion/handler call. So technically we should be allowed to free it.

I think I asked this question (may be on slack), can you please add empty or freeing of normal heap before trx->op_info/row_mysql_convert_row_to_innobase()? Will it have the same problem if we free earlier?

satya-bodapati (changes requested):
Please check my last comment

kamil-holubicki:
Will be handled by 
https://github.com/percona/percona-server/pull/5161
https://github.com/percona/percona-server/pull/5162

PR #4462, opened 2021-09-14
PS-7887: RocksDB encryption

kamil-holubicki:


sql/sql_alter_instance.cc
```diff
@@ -165,6 +165,77 @@ bool Rotate_innodb_master_key::execute() {
...
+    /* SE should have raised error */
+    assert(m_thd->get_stmt_da()->is_error());
+    return true;
+  }
+
+  if (log_to_binlog()) {
+    /*
+      Though we failed to write to binlog,
+      there is no way we can undo this operation.
+      So, covert error to a warning and let user
```
LenzGr:
```suggestion
      So, convert error to a warning and let user
```

storage/rocksdb/enc_env_encryption_myrocks.cc
```diff
@@ -0,0 +1,762 @@
...
+
+namespace myrocks {
+using namespace rocksdb;
+
+namespace {
+// MyRocksEncryptedFileSystemImpl implements an FileSystemWrapper that adds
+// encryption to files stored on disk.
+// This implementation is heavily based on EncryptedFileSystemImpl from rocksdb
+// however, we need the logic to be a bit different + we need to add master
+// key encryption support. Morover EncryptedFileSystemImpl is private to RocksDB
```
LenzGr:
```suggestion
// key encryption support. Moreover, EncryptedFileSystemImpl is private to RocksDB
```

storage/rocksdb/enc_aes_ctr_encryption_provider.cc
```diff
@@ -123,49 +140,72 @@ rocksdb::Status AesCtrEncryptionProvider::ReencryptPrefix(
...
+                                                &newestMasterKeyId)) {
+    ROCKS_LOG_ERROR(logger_, "Failed to get the most recent master key");
+    return rocksdb::Status::IOError();
+  }
 
   uint32_t fileMasterKeyId =
       rocksdb::DecodeFixed32(prefix.data() + MASTER_KEY_ID_OFFSET);
 
   if (newestMasterKeyId == fileMasterKeyId) {
+    ROCKS_LOG_INFO(logger_, "Newest MK already used. Reencryption skipped.");
```
LenzGr:
```suggestion
    ROCKS_LOG_INFO(logger_, "Newest master key already used. Reencryption skipped.");
```

storage/rocksdb/enc_keyring_master_key_manager.cc
```diff
@@ -38,6 +50,7 @@ void KeyringMasterKeyManager::InitKeyringServices() {
   SERVICE_TYPE(registry) *reg_svc = mysql_plugin_registry_acquire();
 
   if (reg_svc == nullptr) {
+    ROCKS_LOG_ERROR(logger_, "Failed to get pluging registry");
```
LenzGr:
```suggestion
    ROCKS_LOG_ERROR(logger_, "Failed to get plugin registry");
```

storage/rocksdb/enc_keyring_master_key_manager.cc
```diff
@@ -14,21 +16,31 @@ namespace myrocks {
...
+      seedUuid_(uuid),
+      logger_(logger) {
   std::call_once(master_key_mutex_init_once_flag, []() {
     master_key_id_mutex_.init(rdb_master_key_mutex_key, MY_MUTEX_INIT_FAST);
   });
 
+#ifdef USE_DEVEL_KEY
+  ROCKS_LOG_WARN(
+      logger_,
+      "KeyringMasterKeyManger uses development (hardcoded) master key.");
```
LenzGr:
```suggestion
      "KeyringMasterKeyManger uses hardcoded development master key.");
```

storage/rocksdb/enc_keyring_master_key_manager.cc
```diff
@@ -88,6 +103,7 @@ void KeyringMasterKeyManager::DeinitKeyringServices() {
   SERVICE_TYPE(registry) *reg_svc = mysql_plugin_registry_acquire();
 
   if (reg_svc == nullptr) {
+    ROCKS_LOG_ERROR(logger_, "Failed to get pluging registry");
```
LenzGr:
```suggestion
    ROCKS_LOG_ERROR(logger_, "Failed to get plugin registry");
```

storage/rocksdb/enc_keyring_master_key_manager.cc
```diff
@@ -276,9 +300,9 @@ int KeyringMasterKeyManager::GetMostRecentMasterKey(std::string *masterKey,
 
   if (retval == -1) {
     masterKey->clear();
-    fprintf(stderr,
-            "Encryption can't find master key, please check the keyring is "
-            "loaded\n");
+    ROCKS_LOG_ERROR(logger_,
+                    "Encryption can't find master key, please check the "
+                    "keyring is installed.");
```
LenzGr:
```suggestion
                    "Encryption can't find master key, please check that "
                    "the keyring is installed.");
```

.gitmodules
```diff
@@ -6,7 +6,7 @@
 	url = https://github.com/percona/Percona-TokuBackup.git
 [submodule "storage/rocksdb/rocksdb"]
 	path = storage/rocksdb/rocksdb
-	url = https://github.com/facebook/rocksdb.git
+	url = https://github.com/kamil-holubicki/rocksdb.git
```
Oneiroi, line 9: Why the deviation from upstream here ? surely we should be tracking against the upstream code, not a fork there of ?

george-lorch: Once our work is finalized, it will be submitted to Rocsdb and may take up to a year to get accepted, then incorporated into Facebook MySQL, then merged to Percona Server.  In the meanwhile, we will have no choice but to maintain a Percona fork of RocksDB.  We have had to do this once or twice in the past in order to keep a single Percona commit on top of the appropriate point-in-time that matches what we have merged in from the Facebook MySQL/MyRocks<-->RocksDB.

kamil-holubicki: Not a big deal here. As you can see, the change is only in 2 files. bool EncryptedWritableFile::IsSyncThreadSafe() override is missing. We can provide our own implementation existing in the MyRocks layer, without touching RocksDB at all. 
This way was easier/quicker for now, and in fact... It should be there from the beginning :)

PR #4463, opened 2021-09-15
Fixed PS-7827 (partition table created with RocksDB engine crashes on a select query)

i-rinat:
https://jira.percona.com/browse/PS-7827
https://jira.percona.com/browse/PS-7815

Problem:
Code in `Rdb_key_def::unpack_record()` and `Rdb_key_def::get_data_start_ptr()`
in MyRocks casts `handler*` pointers back to `ha_rocksdb*` assuming that
they do point to `ha_rocksdb` instances. However, when partitioning is in
use, `rocksdb_create_handler()` creates instances of `ha_rockspart`, which
cannot be casted to `ha_rocksdb`. Trying to do so results in memory
corruptions, and sometimes crashes.

Fix:
Extract buffer management parts of `ha_rocksdb` into a separate class, add
it as a mixin to both `ha_rocksdb` and `ha_rockspart`, and then recover from
`handler*` pointers with `dynamic_cast` at run time. RTTI is already enabled
and used in MyRocks, so no build flags change is necessary.

### DIFF
--- a/storage/rocksdb/rdb_datadic.cc
+++ b/storage/rocksdb/rdb_datadic.cc
@@ -2164,10 +2164,10 @@ int Rdb_key_def::unpack_record(TABLE *const table, uchar *const buf,
   }
 
   // Reset the blob buffer required for unpacking.
-  auto handler = (ha_rocksdb *)table->file;
   if (this->m_max_blob_length) {
-    auto handler = (ha_rocksdb *)table->file;
-    if (handler->reset_blob_buffer(this->m_max_blob_length)) {
+    auto bb = dynamic_cast<blob_buffer *>(table->file);
+    assert(bb);
+    if (bb->reset_blob_buffer(this->m_max_blob_length)) {
       return HA_ERR_OUT_OF_MEM;
     }
   }
@@ -2226,7 +2226,11 @@ int Rdb_key_def::unpack_record(TABLE *const table, uchar *const buf,
     } else {
       /* The checksums are present but we are not checking checksums */
     }
-    handler->m_validated_checksums++;
+
+    // At this point table->file may point to an instance of either ha_rocksdb
+    // or ha_rockspart, so we must be sure it actually is ha_rocksdb.
+    auto handler = dynamic_cast<ha_rocksdb *>(table->file);
+    if (handler) handler->m_validated_checksums++;
   }
 
   if (unlikely(reader.remaining_bytes())) return HA_ERR_ROCKSDB_CORRUPT_DATA;
@@ -3193,8 +3197,9 @@ uchar *Rdb_key_def::get_data_start_ptr(Rdb_field_packing *const fpi, uchar *dst,
   if (fpi->m_field_real_type == MYSQL_TYPE_VARCHAR) {
     data_start = dst + fpi->m_varlength_bytes;
   } else if (is_blob(fpi->m_field_real_type)) {
-    auto handler = static_cast<ha_rocksdb *>(ctx->table->file);
-    data_start = handler->get_blob_buffer(fpi->m_max_field_bytes);
+    auto bb = dynamic_cast<blob_buffer *>(ctx->table->file);
+    assert(bb);
+    data_start = bb->get_blob_buffer(fpi->m_max_field_bytes);
   } else {
     assert(false);
   }
--- a/mysql-test/suite/rocksdb/r/bug_ps7827.result
+++ b/mysql-test/suite/rocksdb/r/bug_ps7827.result
@@ -1,0 +1,1 @@
+bug_ps7827.test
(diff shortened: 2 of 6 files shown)


storage/rocksdb/rdb_datadic.cc
```diff
@@ -2155,7 +2154,8 @@ int Rdb_key_def::unpack_record(TABLE *const table, uchar *const buf,
     } else {
       /* The checksums are present but we are not checking checksums */
     }
-    handler->m_validated_checksums++;
+    auto handler = dynamic_cast<ha_rocksdb *>(table->file);
```
kamil-holubicki, line 2232: A suggestion to consider: 
Maybe it would be better to move all common functionality to the base class. So not blob_buffer class, but e.g ha_rocksbase.
All: I mean moving blob stuff and m_validated_checksums for now. Who knows what in the future.
Then we can just dynamic_cast at the beginning  of the method (once, as we need to cast at least once anyway (in current implementation 2 times in worst case scenario)) and then avoid below 'if'

mysql-test/suite/rocksdb/t/bug7827.test
```diff
@@ -0,0 +1,37 @@
+--source include/have_debug.inc
+--source include/have_rocksdb.inc
+
+# The following statements have caused memory corruption and subsequent server
+# crash. See https://jira.percona.com/browse/PS-7827 for details.
```
kamil-holubicki: Please write a short description of the problem here anyway

storage/rocksdb/rdb_datadic.cc
```diff
@@ -2093,10 +2093,9 @@ int Rdb_key_def::unpack_record(TABLE *const table, uchar *const buf,
   }
 
   // Reset the blob buffer required for unpacking.
-  auto handler = (ha_rocksdb *)table->file;
   if (this->m_max_blob_length) {
-    auto handler = (ha_rocksdb *)table->file;
-    if (handler->reset_blob_buffer(this->m_max_blob_length)) {
+    auto bb = dynamic_cast<blob_buffer *>(table->file);
```
kamil-holubicki, line 2168: Looking at the old code I understand that this dynamic_cast should always succeed, but as we have now dynamic_cast in the general case the result may be nullptr. We should handle this situation or at least add assert(bb) after the cast.

kamil-holubicki (changes requested):

mysql-test/suite/rocksdb/t/bug7827.test
i-rinat: There is no so much to add actually. I've just added a link thinking that there may be some additional information in the future.

For now, I've added an overview of the issue. I've also updated the test to match the version I've sent to facebook/mysql-5.6 repo. The original testcase did not trigger the desired crash, so I had to augment it a bit. File names changed in the process, so this comment thread may be marked as "outdated" by Github.

storage/rocksdb/rdb_datadic.cc
i-rinat: Yes, it should always succeed. I've added `assert(bb)` statements as you suggested. As for the handling, I don't think it a feasible idea. Having `bb == nullptr` is not an expected state, so it would not be possible to test handling of `bb == nullptr` regardless of input data.

storage/rocksdb/rdb_datadic.cc
i-rinat: Extracting the common functionality into an intermediate class was my first idea as well. However, there are some complications with the existing class hierarchy in MySQL that make it a difficult solution. Here is how classes are inherited in the code:
class-graph

As you can see, `ha_rocksdb` and `ha_rockspart` are not descendant of either of them. Instead, the nearest ancestor they have is `handler`. If we are to create an intermediate class, it needs to be inserted into MySQL core classes hierarchy. But then there will be MyRocks implementation details inside MySQL core classes, which something that I cannot advocate for.

The only solution I found was to use `dynamic_cast<>` to extract the mixin class. Yes, it looks a bit cumbersome. But the other alternative is an overhaul of MySQL core classes hierarchy which is way out of scope of the current issue.

i-rinat:
Updated the code to address review comments. Changed testcase to match the updated version that were submitted to `facebook/mysql-5.6` — it now tries harder to crash the server.

i-rinat:
Added a comment about checking results of `dynamic_cast<ha_rocksdb *>` (were mentioned in discussion off-Github).
Rebased to the current state of `8.0` branch.

kamil-holubicki (approved):

kamil-holubicki:
Hi @i-rinat , could you please rebase it on the HEAD of 8.0? I didn't merge it on time and now it has conflicts. As I see the conflict is easy to solve (only 1 file), but I have no permissions to push to your branch.

kamil-holubicki (approved):

PR #4464, opened 2021-09-16
release 8.0.26 16

ldonoso:

`deinit_logging_service_for_plugin` must be called after no more calls
to `LogPluginErrMsg` are done.  If not, the calls to `LogPluginErrMsg`
trigger a segmentation fault as `log_bi` is null.

`rocksdb_shutdown` calls `deinit_logging_service_for_plugin` so previous
calls to `deinit_logging_service_for_plugin` must be removed.

This problem occurred when running the test
`rocksdb.use_direct_reads_writes` as `rocksdb_shutdown` called
`LogPluginErrMsg`.


inikep (approved):
LGTM

PR #4467, opened 2021-09-17
Fixed PS-7873 (log_status table reports wrong gtid_executed)

altmannmarcelo:
https://jira.percona.com/browse/PS-7873

Problem
A commit has 3 stages:
1. Flushing - protected by LOCK_log
2. Syncing- we release LOCK_log and acquire LOCK_sync
3. Committing - we release LOCK_sync and acquire LOCK_commit. In this
stage we also acquire global_sid_lock and update gtid_executed.

When querying log_status table, we acquire LOCK_log for binlog at
Log_resource_binlog_wrapper::lock() and global_sid_lock for
gtid_executed at Log_resource_gtid_state_wrapper::lock().

In high level, Gtid works as follow:
1. a THD ask for the next GTID no for the server ID it's executing the
transaction.
2. it generates the binlog event with said GTID.
3. at stage 3 if THD owns any gtid, it updates gtid_executed

The problem comes when a query to ps.log_status comes after stage 1 of
binlog commit. At this stage, commit has released LOCK_log, binlg file
and position are already updated, Log_resource_binlog_wrapper::lock()
will move forward. Then while binlog is still processing stage 2,
Log_resource_gtid_state_wrapper::lock() will be invoked and will acquire
global_sid_lock. At this point stage 3 of commit will not be able to
update gtid_executed leaving the log_status query with a mismatch
between binlog position (new values) vs gtid_executed (old values).

Fix:
Increase the lock at log_status table to also take LOCK_sync and
LOCK_commit if GTID wrapper has been invoked. This way the query to
log_status will have to wait until gtid_executed has been fully updated
guaranteeing consistency with binlog position.

### DIFF
--- a/mysql-test/suite/binlog_gtid/t/binlog_gtid_log_status_mismatch.test
+++ b/mysql-test/suite/binlog_gtid/t/binlog_gtid_log_status_mismatch.test
@@ -1,0 +1,83 @@
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+# This test case is binary log format agnostic
+--source include/have_binlog_format_row.inc
+--source include/count_sessions.inc
+
+--connect (con1,localhost,root,,)
+--connect (con2,localhost,root,,)
+--echo # Connection default
+--connection default
+RESET MASTER;
+--let $server_uuid= `SELECT @@GLOBAL.SERVER_UUID`
+CREATE TABLE t1(c1 INT);
+INSERT INTO t1 VALUES(1);
+INSERT INTO t1 VALUES(2);
+FLUSH BINARY LOGS;
+
+#
+# Test 1 - Check we get correct GTID from log_status
+#
+
+SET debug_sync = "process_commit_stage_queue_before_handle_gtid SIGNAL insert_in_sync WAIT_FOR insert_continue";
+--send INSERT INTO t1 VALUES(3)
+
+--echo # Connection 1
+--connection con1
+SET DEBUG_SYNC= 'now WAIT_FOR insert_in_sync';
+SET debug_sync = "log_resource_gtid_lock SIGNAL gtid_in_sync WAIT_FOR gtid_continue";
+
+--send SELECT REPLACE(LOCAL->"$.gtid_executed", '"', '') INTO @ps_log_status FROM performance_schema.log_status
+
+--echo # Connection 2
+--connection con2
+SET DEBUG_SYNC= 'now WAIT_FOR gtid_in_sync';
+SET DEBUG_SYNC= 'now SIGNAL gtid_continue';
+SET DEBUG_SYNC= 'now SIGNAL insert_continue';
+
+--echo # Connection 1
+--connection con1
+--reap
+--let $assert_text= Gtid_executed is updated with GTID 4
+--let $assert_cond= "[SELECT @ps_log_status]" = "$server_uuid:1-4"
+--source include/assert.inc
+
+
+--connection default
+--reap
+
+#
+# Test 2 - Check the query to log_status is waiting on LOCK_commit
+#
+
+SET debug_sync = "process_commit_stage_queue_before_handle_gtid SIGNAL insert_in_sync WAIT_FOR insert_continue";
+--send INSERT INTO t1 VALUES(4)
+
+--echo # Connection 1
+--connection con1
+SET DEBUG_SYNC= 'now WAIT_FOR insert_in_sync';
+SET debug_sync = "log_resource_gtid_lock SIGNAL gtid_in_sync WAIT_FOR gtid_continue";
+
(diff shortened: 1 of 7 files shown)


altmannmarcelo:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/1052/

mysql-test/suite/binlog_gtid/t/binlog_gtid_log_status_mismatch.test
```diff
@@ -0,0 +1,37 @@
...
+FLUSH BINARY LOGS;
+# The transaction will not commit until it get 'continue' signal.
+SET debug_sync = "process_commit_stage_queue_before_handle_gtid SIGNAL in_sync WAIT_FOR continue";
+--send INSERT INTO t1 VALUES(3)
+
+--echo # Connection 1
+connection con1;
+SET DEBUG_SYNC= 'now WAIT_FOR in_sync';
+SET DEBUG_SYNC= 'now SIGNAL continue';
+--send SELECT LOCAL->"$.gtid_executed" AS gtid_executed FROM performance_schema.log_status;
```
venkatesh-prasad-v: Please add a wait_condition folowed by an assert here verifying that SELECT is waiting for the commit_lock.

We can use perfschema for validating this. We can use something like 
```
mysql> select PROCESSLIST_INFO from performance_schema.threads T, performance_schema.events_waits_current W where T.THREAD_ID=W.THREAD_ID AND W.EVENT_NAME LIKE '%LOCK_commit';
+-------------------------------------------------------------------------------------+
| PROCESSLIST_INFO                                                                    |
+-------------------------------------------------------------------------------------+
| SELECT LOCAL->"$.gtid_executed" AS gtid_executed FROM performance_schema.log_status |
+-------------------------------------------------------------------------------------+
1 row in set (0.01 sec)
```

```suggestion
--send SELECT LOCAL->"$.gtid_executed" AS gtid_executed FROM performance_schema.log_status;
< wait_condition.inc >
< assert.inc >
SET DEBUG_SYNC= 'now SIGNAL continue';
```

mysql-test/suite/binlog_gtid/t/binlog_gtid_log_status_mismatch.test
```diff
@@ -0,0 +1,37 @@
+--source include/have_debug.inc
```
venkatesh-prasad-v, line 1: This test seems to pass even without the fix. Can you check this?

venkatesh-prasad-v (changes requested):
Hi Marcelo!
Thank you for working on this isse. Not an easy one to fix. I appreciate your efforts on this. I have a few testcase related comments. Please see below.

mysql-test/suite/binlog_gtid/t/binlog_gtid_log_status_mismatch.test
altmannmarcelo: done

mysql-test/suite/binlog_gtid/t/binlog_gtid_log_status_mismatch.test
altmannmarcelo: done

altmannmarcelo:
Ran the whole binlog suite under --lock-order=1
```
marcelo@marce-bld:/work/ps/bld/8.0$ mysql-test/mtr --lock-order=1 --force --do-suite=binlog* --big-test --parallel=15

. . . 

Completed: Failed 3/452 tests, 99.34% were successful.

Failing test(s): binlog_gtid.binlog_xa_select_gtid_executed_explicitly_crash shutdown_report binlog.binlog_reset_master_to_option

The log files in var/log may give you some hint of what went wrong.

If you want to report this error, please read first the documentation
at http://dev.mysql.com/doc/mysql/en/mysql-test-suite.html

399 tests were skipped, 3 by the test itself.

```

mysql-test/suite/binlog_gtid/t/binlog_gtid_log_status_mismatch.test
```diff
@@ -1,37 +1,81 @@
 --source include/have_debug.inc
 --source include/have_debug_sync.inc
-# This test case is binary log format agnostic
---source include/have_binlog_format_row.inc
```
venkatesh-prasad-v: I would retain these lines for the same reason that this test is format agnostic.

venkatesh-prasad-v (commented):
Looks good now. Please squash after addresing the below comment.

mysql-test/suite/binlog_gtid/t/binlog_gtid_log_status_mismatch.test
altmannmarcelo: done

venkatesh-prasad-v (approved):
Approved! Please merge only after all the circle-ci and Azure Pipelines jobs are completed.

PR #4469, opened 2021-09-21
PS-7908 Replace all occurrences of apt-get with apt (5.7)

patrickbirch:
modified:   source/installation/apt_repo.rst
modified:   source/myrocks/install.rst
modified:   source/tokudb/removing_tokudb.rst
modified:   source/tokudb/tokudb_installation.rst
modified:   source/upgrading_guide_56_57.rst

### DIFF
--- a/doc/source/installation/apt_repo.rst
+++ b/doc/source/installation/apt_repo.rst
@@ -34,7 +34,7 @@ Installing |Percona Server| from Percona ``apt`` repository
 
    .. code-block:: bash
 
-      $ sudo apt-get install gnupg2
+      $ sudo apt install gnupg2
 
 2. Fetch the repository packages from Percona web:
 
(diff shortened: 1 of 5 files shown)


doc/source/installation/apt_repo.rst
```diff
@@ -34,34 +34,34 @@ Installing |Percona Server| from Percona ``apt`` repository
 
    .. code-block:: bash
 
-      $ sudo apt-get install gnupg2
+      sudo apt install gnupg2
```
LenzGr: Just curious: you removed the dollar prompt from some command line examples, while you kept it on others.

What's the general rule here? I usually prefer using the `$` to indicate a command that I enter as a regular user, while the `#` symbol indicates commands to be entered as the `root` user. SQL statements should be prefixed by e.g. `mysql>` when using the MySQL CLI.

patrickbirch: Added back.

LenzGr (approved):
LGTM - thank you for taking care of this!

PR #4482, opened 2021-10-01
PS-7894 Wrong error when query interrupted

avodaniel:
This PR tries to fix wrong error message when query interrupted and also to introduce changes to MTR test-cases to make them pass with newly returned error messages.

### DIFF
--- a/mysql-test/suite/rocksdb/t/issue7894.test
+++ b/mysql-test/suite/rocksdb/t/issue7894.test
@@ -1,0 +1,32 @@
+# issue 7894 : Wrong error when query interrupted due to MAX_EXECUTION_TIME
+--source include/have_rocksdb.inc
+
+connect (con1, localhost, root,,);
+connection default;
+
+CREATE TABLE test_table_rocksdb(id INT NOT NULL PRIMARY KEY, value CHAR(32)) ENGINE=ROCKSDB;
+
+INSERT INTO test_table_rocksdb VALUES(1, 'one');
+
+SET SESSION rocksdb_lock_wait_timeout=10;
+
+connection con1;
+BEGIN;
+SELECT /*+ MAX_EXECUTION_TIME(5) */ * FROM test_table_rocksdb WHERE id=1 FOR UPDATE;
+
+connection default;
+--error ER_QUERY_TIMEOUT
+SELECT /*+ MAX_EXECUTION_TIME(5) */ * FROM test_table_rocksdb WHERE id=1 FOR UPDATE;
+--error ER_LOCK_WAIT_TIMEOUT
+SELECT * FROM test_table_rocksdb WHERE id=1 FOR UPDATE;
+
+SET SESSION rocksdb_lock_wait_timeout=5;
+--error ER_LOCK_WAIT_TIMEOUT
+SELECT /*+ MAX_EXECUTION_TIME(20000) */ * FROM test_table_rocksdb WHERE id=1 FOR UPDATE;
+
+set max_execution_time=2000;
+--error ER_QUERY_TIMEOUT
+SELECT * FROM test_table_rocksdb WHERE id=1 FOR UPDATE;
+
+connection con1;
+DROP table test_table_rocksdb;
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -15691,10 +15691,22 @@ double ha_rocksdb::read_time(uint index, uint ranges, ha_rows rows) {
 }
 
 void ha_rocksdb::print_error(int error, myf errflag) {
-  if (error == HA_ERR_ROCKSDB_STATUS_BUSY) {
-    error = HA_ERR_LOCK_DEADLOCK;
+  switch (error) {
+    case HA_ERR_ROCKSDB_STATUS_BUSY:
+      handler::print_error(HA_ERR_LOCK_DEADLOCK, errflag);
+      break;
+    case HA_ERR_LOCK_WAIT_TIMEOUT:
+      if (error == HA_ERR_LOCK_WAIT_TIMEOUT && my_core::thd_killed(ha_thd())) {
+        my_error(ER_QUERY_TIMEOUT, errflag,
+                 table_share->table_name.str /*, error*/);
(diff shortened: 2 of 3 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15342,10 +15342,17 @@ double ha_rocksdb::read_time(uint index, uint ranges, ha_rows rows) {
 }
 
 void ha_rocksdb::print_error(int error, myf errflag) {
-  if (error == HA_ERR_ROCKSDB_STATUS_BUSY) {
-    error = HA_ERR_LOCK_DEADLOCK;
+  switch (error) {
```
kamil-holubicki, line 15694: Unfortunately, this fix spoils the situation when lock_wait_timeout occurs. It will print:
`ERROR 3024 (HY000): Query execution was interrupted, maximum statement execution time exceeded`
but should:
`ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction`

In other words the current behavior is (testcase from https://jira.percona.com/browse/PS-7894):
```
mysql> select /*+ MAX_EXECUTION_TIME(5) */ * from test_table_rocks where id=1 for update; 
ERROR 3024 (HY000): Query execution was interrupted, maximum statement execution time exceeded
mysql> select * from test_table_rocks where id=1 for update; 
ERROR 3024 (HY000): Query execution was interrupted, maximum statement execution time exceeded

mysql> select /*+ MAX_EXECUTION_TIME(5) */ * from test_table_inno where id=1 for update; 
ERROR 3024 (HY000): Query execution was interrupted, maximum statement execution time exceeded
mysql> select * from test_table_inno where id=1 for update; 
ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction
```
but should be:
```
mysql> select /*+ MAX_EXECUTION_TIME(5) */ * from test_table_rocks where id=1 for update; 
ERROR 3024 (HY000): Query execution was interrupted, maximum statement execution time exceeded
mysql> select * from test_table_rocks where id=1 for update; 
ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction

mysql> select /*+ MAX_EXECUTION_TIME(5) */ * from test_table_inno where id=1 for update; 
ERROR 3024 (HY000): Query execution was interrupted, maximum statement execution time exceeded
mysql> select * from test_table_inno where id=1 for update; 
ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction
```

mysql-test/suite/rocksdb/include/locking_issues_case1_1.inc
```diff
@@ -31,10 +31,10 @@ SELECT * FROM t0 WHERE id1=1 AND id2=5 FOR UPDATE;
 connection con2;
 eval SET SESSION TRANSACTION ISOLATION LEVEL $isolation_level;
 BEGIN;
---error ER_LOCK_WAIT_TIMEOUT
+--error ER_QUERY_TIMEOUT
```
kamil-holubicki: This and other re-recorded tests. I'm not sure this is OK. This may be the side effect of the fix which loses information about lock_wait_timeout. Now every ER_LOCK_WAIT_TIMEOUT is converted into ER_QUERY_TIMEOUT which I think is not OK. We need to distinguish these 2 cases like we do for InnoDB

doc/source/diagnostics/slow_extended.rst
```diff
@@ -298,7 +298,7 @@ Any other number      KILLED_NO_VALUE (Catches all other cases)
 .. seealso::
 
    |MySQL| Documentation: |MySQL| Server Error Codes
-      https://dev.mysql.com/doc/refman/8.0/en/server-error-reference.html
+      https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html
```
kamil-holubicki: Please rebase on 8.0 HEAD. PR should not contain any doc related stuff.

kamil-holubicki (changes requested):

kamil-holubicki (changes requested):
It looks like SELECT 'hint' case works as expected now, but I see that specifying 
`SET max_execution_time=2000;`
still suffers the same problem
For InnoDB:
```
mysql> set max_execution_time=2000;
Query OK, 0 rows affected (0.00 sec)
mysql> select * from test_table_innodb where id=1 for update;
ERROR 3024 (HY000): Query execution was interrupted, maximum statement execution time exceeded 
```
For RocksDB:
```
mysql> set max_execution_time=2000;
Query OK, 0 rows affected (0.00 sec)
select * from test_table_rocksdb where id=1 for update;
ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction 
```
This is because the current solution overrides the error code when it detects that hint was specified in the query. The above case with session variable is not handled.

kamil-holubicki:
I checked once again, and the current fix doesn't work for hint case as well
```
mysql> SET SESSION rocksdb_lock_wait_timeout=5;
Query OK, 0 rows affected (0.00 sec)mysql> select /*+ MAX_EXECUTION_TIME(20000) */ * from test_table_rocksdb where id=1 for update;
ERROR 3024 (HY000): Query execution was interrupted, maximum statement execution time exceeded
```
I think the error should be
```
ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction
```
as it is for InnoDB

kamil-holubicki (commented):
Hi! Looks fine. Please correct uppercase/lowercase in the test file. Please squash commits and write a proper commit message with a detailed explanation of the problem and the solution.

mysql-test/suite/rocksdb/t/issue7894.test
```diff
@@ -0,0 +1,33 @@
+# issue 7894 : Wrong error when query interrupted due to MAX_EXECUTION_TIME
+--source include/have_rocksdb.inc
+
+connect (con1, localhost, root,,);
+connection default;
+
+create table test_table_rocksdb(id int not null primary key, value char(32)) engine=rocksdb;
```
kamil-holubicki: please use UPPERCASE for all SQL keywords.

kamil-holubicki (approved):

PR #4484, opened 2021-10-05
PS-5168: Set stats.data_file_length in temptable SE

oleksandr-kachan:
https://jira.percona.com/browse/PS-5168

The stats.data_file_length isn't set in temptable storage
engine. As a result of this we always get "Tmp_table_sizes: 0"
in slow query log even in case temporary table was actually
used during query processing.

To fix the issue the Handler::info() for temptable was updated
to set stats.data_file_length. Needed info is fetched from the
allocator instance wich handles all memory allocations for
the temptable storage engine.
In addition invocation of Handler::info() added in Handler::open()
to properly update memory usage after table is opened.

### DIFF
--- a/sql/sql_tmp_table.cc
+++ b/sql/sql_tmp_table.cc
@@ -2371,7 +2371,6 @@ void close_tmp_table(TABLE *table) {
   DBUG_TRACE;
   DBUG_PRINT("enter", ("table: %s", table->alias));
 
-  // Possibly use current_thd instead of table->in_use
   if (table->file && table->in_use != nullptr)
     table->in_use->tmp_tables_size += table->file->stats.data_file_length;
 
--- a/storage/temptable/include/temptable/allocator.h
+++ b/storage/temptable/include/temptable/allocator.h
@@ -277,6 +277,12 @@ struct AllocatorState {
    * new block needs to be created.
    */
   size_t number_of_blocks = 0;
+
+  /**
+   * Allocated memory size counter. Used for statistics to determine
+   * maximum allocated memory space for a table.
+   */
+  size_t allocated_mem_counter;
 };
 
 /** Custom memory allocator. All dynamic memory used by the TempTable engine
@@ -413,6 +419,10 @@ class Allocator {
    * before other methods. */
   static void init();
 
+  uint64_t get_allocated_mem_counter() const {
+    return m_state->allocated_mem_counter;
+  }
+
   /**
     Shared state between all the copies and rebinds of this allocator.
     See AllocatorState for details.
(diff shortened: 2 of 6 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1067/

sql/sql_tmp_table.cc
```diff
@@ -2394,6 +2390,10 @@ void close_tmp_table(TABLE *table) {
     table->set_deleted();
   }
 
+  if (table->file && table->in_use != nullptr) {
+    table->in_use->tmp_tables_size += table->file->stats.data_file_length;
+  }
+
```
satya-bodapati: why did we move this down? I don't see any info() call in between.

satya-bodapati (changes requested):

sql/sql_tmp_table.cc
oleksandr-kachan: info() call is done by temptable storage engine itself when closing a table

"In addition invocation of Handler::info() added in Handler::close()
to properly update memory usage before table is closed."

satya-bodapati: ah, there is ha_close() in between. Got it!

storage/temptable/src/handler.cc
```diff
@@ -216,6 +216,9 @@ int Handler::close() {
 
   assert(m_opened_table != nullptr);
 
+  // Update stats before closing a table
+  info(HA_STATUS_VARIABLE);
```
satya-bodapati: I dont see such info() in ha_innobase::close(). How does it work for InnoDB?

oleksandr-kachan: Removed this one.

storage/temptable/include/temptable/allocator.h
```diff
@@ -413,6 +413,11 @@ class Allocator {
    * before other methods. */
   static void init();
 
+  uint64_t get_consumed_mem_size() const {
+    return MemoryMonitor::RAM::consumption() +
+           MemoryMonitor::MMAP::consumption();
+  }
```
satya-bodapati: Is this per table? or shared across tables?

satya-bodapati (changes requested):

storage/temptable/include/temptable/allocator.h
oleksandr-kachan: This one was shared but we need per table stats here... I've reworked change a bit.

storage/temptable/include/temptable/allocator.h
```diff
@@ -499,6 +509,7 @@ inline T *Allocator<T, AllocationScheme>::allocate(size_t n_elements) {
   T *chunk_data =
       reinterpret_cast<T *>(block->allocate(n_bytes_requested).data());
   assert(reinterpret_cast<uintptr_t>(chunk_data) % alignof(T) == 0);
+  m_state->allocated_mem_counter += n_bytes_requested;
```
satya-bodapati, line 512: is the free done only at close() of table object? I don't see any decrements. Not necessary? :thinking:

oleksandr-kachan: In my understanding we want to know the max used memory by tmp table while processing some query. So I'm not sure we need tracking deallocations here. What do you think?

satya-bodapati (approved):

PR #4485, opened 2021-10-05
PS-7890 : Unable to install RocksDB engine plugin if server is started with --loose-rocksdb_persistent_cache_size_mb

avbelov23:
Added checks to bypass RocksDB restrictions

The problem has been dragging on for a long time and there is little chance that it will be fixed.
https://github.com/facebook/mysql-5.6/issues/911
https://github.com/facebook/mysql-5.6/issues/713

I don't see an explanation of why the cache size is limited to 100 Mb. I found a person who wrote the code, but he has one contact through linkedin.com (https://www.linkedin.com/in/karrad/). I asked him a question, but even if he answers, this is third party code and won't affect anything in the near future. Therefore, I propose this patch.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -5235,6 +5235,39 @@ static int rocksdb_init_func(void *const p) {
   }
 
   if (rocksdb_persistent_cache_size_mb > 0) {
+    // TODO: This is the limitations in RocksDB.
+    // 1. Persistent cache size RocksDB has to be at least cache_file_size.
+    //
+    // utilities/persistent_cache/persistent_cache_tier.h:112
+    // Status ValidateSettings() const {
+    // ...
+    // if (cache_size < cache_file_size ...) {
+    // ...
+    // cache_file_size is set here.
+    //
+    // utilities/persistent_cache/persistent_cache_tier.h:165
+    // uint32_t cache_file_size = 100ULL * 1024 * 1024;
+    //
+    // 2. rocksdb_persistent_cache_path required persistent cache parameter
+    //
+    // utilities/persistent_cache/persistent_cache_tier.h:104
+    // Status ValidateSettings() const {
+    // ...
+    // if (!env || path.empty()) {
+    // ...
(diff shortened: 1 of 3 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5688,6 +5688,30 @@ static int rocksdb_init_internal(void *const p) {
   }
 
   if (rocksdb_persistent_cache_size_mb > 0) {
+    // TODO: This is the limitation in RocksDB.
```
kamil-holubicki: Please replace this comment with the explanation where `cache_file_size` is set to 100MB and where is the validation of the value provided with `rocksdb_persistent_cache_size_mb`. That will explain why we need to do the validation prior to calling `NewPersistentCache`

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5688,6 +5688,30 @@ static int rocksdb_init_internal(void *const p) {
...
+    // utilities/persistent_cache/block_cache_tier.cc:33
+    // status = opt_.ValidateSettings();
+    // assert(status.ok()); // <- THIS IS
+    // if (!status.ok()) {
+    // ...
+    // }
+    //
+    // BEGIN
+    if (rocksdb_persistent_cache_size_mb < 100) {
+      LogPluginErrMsg(ERROR_LEVEL, 0, "Invalid persistent cache options");
```
kamil-holubicki: Would be nice to write something like
`Invalid value for rocksdb_persistent_cache_size_mb. It has  to be at least XYZ`

also please do not hardcode `100` in such a case twice, rather use `static constexpr int persistent_cache_size_mb_min = 100;` and then use it.

kamil-holubicki (changes requested):
I think this is the right way, but we need some improvements:
1. The problem is reported to 5.7, so please provide the fix as well (remember about GCA)
2. Can we write MTR test for this?
3. Please also see my suggestions inline.

kamil-holubicki (approved):
lgtm

PR #4486, opened 2021-10-06
Update function comment for disableMaster() within semisync_master.h

blylei:
Just modify the comment


dutow:
Hello!

This is a bug in the MySQL sources, please submit a bugreport at https://bugs.mysql.com/ or open a pull request at https://github.com/mysql/mysql-server instead.

PR #4489, opened 2021-10-11
PS-7883 : Server crashes with ALTER query when --rocksdb_write_disable_wal is enabled

avbelov23:
https://jira.percona.com/browse/PS-7883

Sync writes has to enable WAL. Output a warning if wal is disabled and
sync writes is enabled, and ignore wal disabled

### DIFF
--- a/mysql-test/suite/rocksdb/t/locking_issues.test
+++ b/mysql-test/suite/rocksdb/t/locking_issues.test
@@ -1,5 +1,11 @@
 --source include/have_rocksdb.inc
 
+# rocksdb_write_disable_wal=ON is incompatible with
+# synchronous writes (rocksdb_flush_log_at_trx_commit=1).
+# See https://jira.percona.com/browse/PS-7883 for details.
+SET @rocksdb_flush_log_at_trx_commit_saved = @@global.rocksdb_flush_log_at_trx_commit;
+SET @@global.rocksdb_flush_log_at_trx_commit = 2;
+
 let $isolation_level = REPEATABLE READ;
 --source suite/rocksdb/include/locking_issues_case1_1.inc
 
@@ -71,3 +77,5 @@ let $isolation_level = REPEATABLE READ;
 
 let $isolation_level = READ COMMITTED;
 --source suite/rocksdb/include/locking_issues_case7.inc
+
+SET @@global.rocksdb_flush_log_at_trx_commit = @rocksdb_flush_log_at_trx_commit_saved;
--- a/mysql-test/suite/rocksdb/t/rocksdb_bug7883.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_bug7883.test
@@ -1,0 +1,19 @@
+--source include/have_rocksdb.inc
+
+# rocksdb_write_disable_wal=ON is incompatible with
+# synchronous writes (rocksdb_flush_log_at_trx_commit=1).
+# See https://jira.percona.com/browse/PS-7883 for details.
+
+call mtr.add_suppression("Plugin rocksdb reported:");
+call mtr.add_suppression("Plugin 'ROCKSDB'");
+
+--error ER_UNKNOWN_STORAGE_ENGINE
+CREATE TABLE tt_12 (ipkey INT AUTO_INCREMENT, i1 INT, d2 DOUBLE,
+                    PRIMARY KEY(ipkey), INDEX tt_12i0(d2, i1 ASC, ipkey),
+                    INDEX tt_12i1(ipkey, i1, d2))
+                   ROW_FORMAT=DYNAMIC ENGINE=RocksDB;
+
+--let $grep_file=$MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $grep_pattern=\[ERROR\] .* Plugin rocksdb reported: 'Invalid argument: Sync writes \(rocksdb_flush_log_at_trx_commit == 1\) has to enable WAL'
+--let $grep_output=boolean
+--source include/grep_pattern.inc
--- a/mysql-test/suite/rocksdb/t/rocksdb_bug7883_2.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_bug7883_2.test
@@ -1,0 +1,71 @@
+--source include/have_rocksdb.inc
+
+--connect(con1, localhost, root,,)
+--connect(con2, localhost, root,,)
+
+--connection con1
+SET @rocksdb_flush_log_at_trx_commit_saved = @@global.rocksdb_flush_log_at_trx_commit;
+SET @rocksdb_write_disable_wal_saved = @@session.rocksdb_write_disable_wal;
+SET @@global.rocksdb_flush_log_at_trx_commit = 2;
+SET @@session.rocksdb_write_disable_wal=ON;
+
+--connection con2
+SET @@session.rocksdb_write_disable_wal=OFF;
+SET @@global.rocksdb_flush_log_at_trx_commit = 1;
+
+--connection con1
+# Now rocksdb_flush_log_at_trx_commit = 1 and rocksdb_write_disable_wal=ON
+# The following statements have caused crash as
+# rocksdb_write_disable_wal=ON is incompatible with
+# synchronous writes (rocksdb_flush_log_at_trx_commit=1).
+# See https://jira.percona.com/browse/PS-7883 for details.
+
+# The incompatible value (ON) is retained until the variable is used and the
+# incompatible value will not be temporarily hidden by the compatible
+SHOW SESSION VARIABLES LIKE "rocksdb_write_disable_wal";
+
+CREATE TABLE tt_12 (ipkey INT AUTO_INCREMENT, i1 INT, d2 DOUBLE,
+                    PRIMARY KEY(ipkey), INDEX tt_12i0(d2, i1 ASC, ipkey),
+                    INDEX tt_12i1(ipkey, i1, d2))
+                   ROW_FORMAT=DYNAMIC ENGINE=RocksDB;
+
+REPLACE INTO tt_12  (ipkey, i1, d2) VALUES(2097, 6145, 0.00000);
+
+call mtr.add_suppression("Sync writes has to enable WAL. Switching to rocksdb_write_disable_wal = false for the time when sync writes are enabled.");
+ALTER TABLE tt_12 MODIFY COLUMN  ipkey INT(14), LOCK=DEFAULT, ALGORITHM=DEFAULT;
+
+--let $grep_file=$MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $grep_pattern=\[Warning\] .* Plugin rocksdb reported: 'Sync writes has to enable WAL. Switching to rocksdb_write_disable_wal = false for the time when sync writes are enabled.'
+--let $grep_output=boolean
+--source include/grep_pattern.inc
+
+# The compatible value (OFF) hides incompatible (ON) after ALTER TABLE
+SHOW SESSION VARIABLES LIKE "rocksdb_write_disable_wal";
+
+--connection con2
+SET @@global.rocksdb_flush_log_at_trx_commit = 2;
+
+--connection con1
+
+# The compatible value (OFF) that hides a previously incompatible value (ON) is
+# retained until the variable is used
+SHOW SESSION VARIABLES LIKE "rocksdb_write_disable_wal";
+
+call mtr.add_suppression("Sync writes disabled. Switching back to rocksdb_write_disable_wal = true");
+ALTER TABLE tt_12 MODIFY COLUMN  ipkey INT(15), LOCK=DEFAULT, ALGORITHM=DEFAULT;
+
+--let $grep_file=$MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $grep_pattern=\[Warning\] .* Plugin rocksdb reported: 'Sync writes disabled. Switching back to rocksdb_write_disable_wal = true'
+--let $grep_output=boolean
+--source include/grep_pattern.inc
+
+# The originally incompatible value (ON) is now compatible and we see and use it
+SHOW SESSION VARIABLES LIKE "rocksdb_write_disable_wal";
+
+DROP TABLE tt_12;
+
+SET @@session.rocksdb_write_disable_wal = @rocksdb_write_disable_wal_saved;
+SET @@global.rocksdb_flush_log_at_trx_commit = @rocksdb_flush_log_at_trx_commit_saved;
+
+--disconnect con1
+--disconnect con2
--- a/mysql-test/suite/rocksdb_sys_vars/r/rocksdb_flush_log_at_trx_commit_basic.result
+++ b/mysql-test/suite/rocksdb_sys_vars/r/rocksdb_flush_log_at_trx_commit_basic.result
@@ -1,3 +1,5 @@
+SET @rocksdb_flush_log_at_trx_commit_saved = @@global.rocksdb_flush_log_at_trx_commit;
+SET @rocksdb_write_disable_wal_saved = @@session.rocksdb_write_disable_wal;
 CREATE TABLE valid_values (value varchar(255)) ENGINE=myisam;
 INSERT INTO valid_values VALUES(2);
 INSERT INTO valid_values VALUES(1);
@@ -55,3 +57,9 @@ SELECT @@global.ROCKSDB_FLUSH_LOG_AT_TRX_COMMIT;
 1
 DROP TABLE valid_values;
 DROP TABLE invalid_values;
+SET @@global.rocksdb_flush_log_at_trx_commit=0;
+SET @@session.rocksdb_write_disable_wal = ON;
+SET @@global.rocksdb_flush_log_at_trx_commit=1;
+ERROR HY000: Got error 513 'rocksdb_flush_log_at_trx_commit = 1 and rocksdb_write_disable_wal = ON are not compatible' from ROCKSDB
+SET @@session.rocksdb_write_disable_wal = @rocksdb_write_disable_wal_saved;
+SET @@global.rocksdb_flush_log_at_trx_commit = @rocksdb_flush_log_at_trx_commit_saved;
--- a/mysql-test/suite/rocksdb_sys_vars/r/rocksdb_write_disable_wal_basic.result
+++ b/mysql-test/suite/rocksdb_sys_vars/r/rocksdb_write_disable_wal_basic.result
@@ -1,3 +1,6 @@
+SET @rocksdb_flush_log_at_trx_commit_saved = @@global.rocksdb_flush_log_at_trx_commit;
+SET @rocksdb_write_disable_wal_saved = @@session.rocksdb_write_disable_wal;
+SET @@global.rocksdb_flush_log_at_trx_commit = 2;
 CREATE TABLE valid_values (value varchar(255)) ENGINE=myisam;
 INSERT INTO valid_values VALUES(1);
 INSERT INTO valid_values VALUES(0);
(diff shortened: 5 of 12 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3999,6 +4003,10 @@ class Rdb_writebatch_impl : public Rdb_transaction {
     write_opts.sync = (rocksdb_flush_log_at_trx_commit == FLUSH_LOG_SYNC) &&
                       rdb_sync_wal_supported();
     write_opts.disableWAL = THDVAR(m_thd, write_disable_wal);
+    if (write_opts.disableWAL && write_opts.sync) {
```
kamil-holubicki: likewise

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3723,6 +3723,10 @@ class Rdb_transaction_impl : public Rdb_transaction {
...
```
kamil-holubicki: The idea is correct however, I think we should rather handle the validation in MYSQL_THDVAR_BOOL(write_disable_wal...) on_check callback rather than in the execution path.
We should detect that rocksdb_flush_log_at_trx_commit == FLUSH_LOG_SYNC and the user tries to write_disable_wal=1, write a proper warning to the user (what he has to do this setting), and prevent the wrong configuration.

The same in 
MYSQL_SYSVAR_UINT(rocksdb_flush_log_at_trx_commit...), we should validate the current setting of write_disable_wal and in case of conflict write a proper message informing how to change the setting and prevent conflict.

kamil-holubicki (changes requested):

kamil-holubicki:
I just thought about parameters passed as command-line options. We should handle this case as well. I think we should prevent rocksdb start in case of the wrong configuration is detected (+ appropriate message). That will be better than overriding some settings behind the scenes. What if the user DOES need to disable wal? We don't want automatic to happen in the background that enables it again. It is better to say it to the user right away.

storage/rocksdb/ha_rocksdb.cc
avbelov23: These are the checks (on_check) for SET expression. By a bug, the setting of options occurs/can occur through the command line options. I can only add these restrictions to the on_check callback so that SET expression swears, but I cannot delete this code

kamil-holubicki: The problem with the current implementation is that it changes the actual behavior of the code to the one as write_disable_wal was OFF, even if `show variables like 'rocksdb_write_disable_wal';` says that it is ON.
In other words:
The user sees that `show variables like 'rocksdb_write_disable_wal';` is ON, but the code acts as rocksdb_write_disable_wal was OFF.

I think we can handle command-line options validation at the beginning of `rocksdb_init_internal()`

It should work in the following way:
1. initialization: when incompatible options are detected, an error message should be printed out and the plugin initialization should fail
2. user executes SET...: validation should be done in on_check. If options are not compatible, the proper message should be printed to the user, the value should not change.

The above is how I think it should behave. The acceptable alternative I see might be the following:
1. initialization: when incompatible options are detected, fix them automatically, but in the way that values of variables reported to the user are the real ones used by execution paths + warning in the error log that we detected the wrong configuration and fixed it in the way A-B-C
2. As previously
To be honest, I'm not a big fan of doing such magic behind the scenes. It may confuse the user who would not notice the warning in the log and think that if I set it to ON and it started, it is ON. 

So I vote for the 1st described approach. 
@georgelorchpercona could you comment on this requirement, please?

george-lorch: @kamil-holubicki you are 100% correct, option 1 is the proper way to do this in MySQL.  It is unfortunate because it introduces some code redundancy between the initialization validation and the SET validation, but this can usually be extracted out into an inline that both locations can call.

avbelov23:
DONE

mysql-test/suite/rocksdb/t/locking_issues.test
```diff
@@ -71,3 +76,5 @@ let $isolation_level = REPEATABLE READ;
 
 let $isolation_level = READ COMMITTED;
 --source suite/rocksdb/include/locking_issues_case7.inc
+
+SET @@global.ROCKSDB_FLUSH_LOG_AT_TRX_COMMIT = 1;
```
kamil-holubicki: Please follow the pattern from other tests, when we change the global variable.
```
SET @variable_name_saved = @@variable_name
SET @@variable_name=some value
... do other stuff here ...
SET @@variable_name = @variable_name_saved
```

mysql-test/suite/rocksdb_sys_vars/t/rocksdb_write_disable_wal_basic.test
```diff
@@ -17,3 +22,5 @@ INSERT INTO invalid_values VALUES('\'aaa\'');
 
 DROP TABLE valid_values;
 DROP TABLE invalid_values;
+
+SET @@global.ROCKSDB_FLUSH_LOG_AT_TRX_COMMIT = 1;
```
kamil-holubicki: See previous comment for similar case.

mysql-test/suite/rocksdb/t/rocksdb_bug7883.test
```diff
@@ -0,0 +1,15 @@
...
+
+# The following statements have caused crash as
+# rocksdb_write_disable_wal=ON is incompatible with
+# synchronous writes (rocksdb_flush_log_at_trx_commit=1).
+# See https://jira.percona.com/browse/PS-7883 for details.
+
+call mtr.add_suppression("Plugin rocksdb reported:");
+call mtr.add_suppression("Plugin 'ROCKSDB'");
+
+--error 1286
```
kamil-holubicki: Is it possible to use error name instead of error code here?

mysql-test/suite/rocksdb/t/rocksdb_bug7883.test
```diff
@@ -0,0 +1,15 @@
...
+CREATE TABLE tt_12 (ipkey INT AUTO_INCREMENT, i1 INT, d2 DOUBLE,
+                    PRIMARY KEY(ipkey), INDEX tt_12i0(d2, i1 ASC, ipkey),
+                    INDEX tt_12i1(ipkey, i1, d2))
+                   ROW_FORMAT=DYNAMIC ENGINE=RocksDB;
```
kamil-holubicki, line 14: The original test attached by the bug reporter to the Jira ticket has REPLACE INTO and ALTER TABLE afterward. I understand that now we detect wrong configuration when RocksDB starts and print the message into the error log. 
1. Could we add the validation (grep) that proper message is written into error log?
2. "The following statements have caused crash" comment seems to be not relevant, as what we have in the test didn't cause the crash
3. Could you please add a test for the case when RocksDB is started and we try to change rocksdb_write_disable_wal to an incompatible value? (I thnk it should test when we change/manipulate with rocksdb_write_disable_wal and when we change/manipulate with ROCKSDB_FLUSH_LOG_AT_TRX_COMMIT)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -985,21 +985,10 @@ static int rocksdb_validate_flush_log_at_trx_commit(
...
-
-  if (rocksdb_db_options->allow_mmap_writes && new_value != FLUSH_LOG_NEVER) {
-    return HA_EXIT_FAILURE;
-  }
-
-  *static_cast<uint32_t *>(var_ptr) = static_cast<uint32_t>(new_value);
-  return HA_EXIT_SUCCESS;
-}
+    struct st_mysql_value *const value /* in: incoming value */);
+int rocksdb_check_write_disable_wal(
```
kamil-holubicki: static

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -14840,6 +14838,56 @@ int mysql_value_to_bool(struct st_mysql_value *value, bool *return_value) {
...
+  /* value is NULL */
+  if (value->val_int(value, &new_value)) {
+    return HA_EXIT_FAILURE;
+  }
+
+  if (rocksdb_db_options->allow_mmap_writes && new_value != FLUSH_LOG_NEVER) {
+    return HA_EXIT_FAILURE;
+  }
+
+  bool write_disable_wal = THDVAR(thd, write_disable_wal);
```
kamil-holubicki, line 14890: I started thinking about the case when we change rocksdb_validate_flush_log_at_trx_commit (global) to the value which is OK for the current thread's write_disable_wal (session), but not compatible with the value of write_disable_wal of another thread. 
This validation will not prevent setting the wrong combination and we should be able still to see the crash.
I started wondering how to solve this. For now, it looks like we would need the combination of the current fix and your original idea (with the change that the value reported to the user by SHOW VARIABLES LIKE .... should be the proper/real one). Any better idea? I cannot figure out anything better for now :).  Let's discuss the ideas before implementing.

kamil-holubicki (changes requested):

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5311,6 +5309,14 @@ static int rocksdb_init_internal(void *const p) {
   // Validate the assumption about the size of ROCKSDB_SIZEOF_HIDDEN_PK_COLUMN.
   static_assert(sizeof(longlong) == 8, "Assuming that longlong is 8 bytes.");
 
+  if (THDVAR(nullptr, write_disable_wal) &&
+      rocksdb_flush_log_at_trx_commit == FLUSH_LOG_SYNC) {
+    LogPluginErrMsg(ERROR_LEVEL, 0,
```
kamil-holubicki, line 5339: Is this message OK? FLUSH_LOG_SYNC expands to 1, not 0.
Wouldn't it be more clear to say:
"Invalid argument: Cannot disable wal while sync writes are enabled (rocksdb_write_disable_wal cannot true and rocksdb_flush_log_at_trx_commit cannot be 1 at the same time)"

Side question: I understand that having rocksdb_flush_log_at_trx_commit == FLUSH_LOG_BACKGROUND and rocksdb_write_disable_wal==true is the valid combination, right?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -14840,6 +14846,56 @@ int mysql_value_to_bool(struct st_mysql_value *value, bool *return_value) {
...
+static int rocksdb_validate_flush_log_at_trx_commit(
+    THD *const thd,
+    struct SYS_VAR *const var, /* in: pointer to system variable */
+    void *var_ptr,             /* out: immediate result for update function */
+    struct st_mysql_value *const value /* in: incoming value */) {
+  long long new_value;
+
+  /* value is NULL */
+  if (value->val_int(value, &new_value)) {
+    return HA_EXIT_FAILURE;
```
kamil-holubicki: (many times)
Not a strong opinion. I see that other functions use it as well, but HA_EXIT_FAILURE expands to 'true' (bool). Function type is 'int'.
I understand that bool to int conversion is implicit, but I can imagine potential problem if we change HA_EXIT_FAILURE macro definition :)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -14840,6 +14846,56 @@ int mysql_value_to_bool(struct st_mysql_value *value, bool *return_value) {
...
+    THD *const thd, struct SYS_VAR *var MY_ATTRIBUTE((__unused__)), void *save,
+    struct st_mysql_value *value) {
+  bool new_value;
+  if (mysql_value_to_bool(value, &new_value) != 0) {
+    return 1;
+  }
+
+  if (rocksdb_flush_log_at_trx_commit == FLUSH_LOG_SYNC && new_value) {
+    my_error(
+        ER_GET_ERRMSG, MYF(0), HA_ERR_ROCKSDB_STATUS_INVALID_ARGUMENT,
```
kamil-holubicki, line 14913: Likewise, FLUSH_LOG_SYNC expands to 1, not 0.

kamil-holubicki:
Please check also https://github.com/tempesta-tech/percona-server/pull/1

storage/rocksdb/ha_rocksdb.cc
avbelov23: Typo in LogPluginErrMsg()

kamil-holubicki: I'm not sure I understand. The error message is still wrong, no?

include/mysql/plugin.h
```diff
@@ -273,6 +273,23 @@ typedef int (*mysql_var_check_func)(MYSQL_THD thd, SYS_VAR *var, void *save,
 typedef void (*mysql_var_update_func)(MYSQL_THD thd, SYS_VAR *var,
                                       void *var_ptr, const void *save);
 
+/*
```
kamil-holubicki: We are changing the plugin API here. I wouldn't go this way, even if necessary. The bug itself is not so critical that we should go with API change. See my detailed thoughts on Slack.

sql/sql_plugin_var.cc
```diff
@@ -323,7 +323,54 @@ uchar *sys_var_pluginvar::real_value_ptr(THD *thd, enum_var_type type) {
     /* scope of OPT_PERSIST is always GLOBAL */
     if (type == OPT_GLOBAL || type == OPT_PERSIST) thd = nullptr;
 
-    return intern_sys_var_ptr(thd, *(int *)(plugin_var + 1), false);
+    uchar *ptr = intern_sys_var_ptr(thd, *(int *)(plugin_var + 1), false);
```
kamil-holubicki: I think `show variables like ...` doesn't go this path.... So how will it fix the case:
1. set global (to incompatible one) (session 1)
2. show session variable like ... (session 2)
?

kamil-holubicki (commented):
This does not compile for me reporting ABI compatibility check errors. Maybe I'm doing something wrong, but... I cleaned everything and started from scratch so...

It is nice research, but, I don't think we will go this way. Several things:

Most important: we changed plugin API (I think this is the reason of the error I get). We don't want to do this. The problem is not so important to do such change. I think our research showed that we cannot achieve what we need without changing the interface, and that is a good research/learning exercise, but no...
I'm not sure it works as it should. My research shows that show variables like ... does not use sys_var_pluginvar::real_value_ptr for session variables, so I'm not sure how this commit fixes the problem we discussed recently (set global from session 1, then without anything in between, show session variable from session 2)
Maybe I'm missing something, but as I said it does not compile for me, so there is no way to check. Please advise.

Anyway, I'm strongly against changing plugin API.

storage/rocksdb/ha_rocksdb.cc
avbelov23: Message "Invalid argument: Sync writes (**rocksdb_flush_log_at_trx_commit == 0**) has to enable WAL" was wrong.
Need "Invalid argument: Sync writes (**rocksdb_flush_log_at_trx_commit == 1**) has to enable WAL"

sql/sql_plugin_var.cc
avbelov23: You need to look at the idea of a class `sys_var_pluginvar`
It seemed to me there was a test, but I can see it just added `show variables like...` to the existing rocksdb_bug7883_2
Trace should be:
```sys_var_pluginvar::session_value_ptr()
sys_var_pluginvar::session_value_ptr()
sys_var_pluginvar::do_value_ptr()
sys_var_pluginvar::real_value_ptr()```

File: mysql-test/suite/rocksdb_sys_vars/r/rocksdb_flush_log_at_trx_commit_basic.result [hunk 14]
```diff
@@ -55,3 +57,9 @@ SELECT @@global.ROCKSDB_FLUSH_LOG_AT_TRX_COMMIT;
 1
 DROP TABLE valid_values;
 DROP TABLE invalid_values;
+SET @@global.rocksdb_flush_log_at_trx_commit=0;
+SET @@session.rocksdb_write_disable_wal = ON;
+SET @@global.rocksdb_flush_log_at_trx_commit=1;
+ERROR HY000: Got error 513 'Invalid argument: Cannot disable wal while sync writes are enabled (rocksdb_write_disable_wal cannot' from ROCKSDB
```
kamil-holubicki: The error message is trucated.

mysql-test/suite/rocksdb_sys_vars/r/rocksdb_write_disable_wal_basic.result
```diff
@@ -112,3 +115,9 @@ SELECT @@session.ROCKSDB_WRITE_DISABLE_WAL;
 0
 DROP TABLE valid_values;
 DROP TABLE invalid_values;
+SET @@session.rocksdb_write_disable_wal=OFF;
+SET @@global.rocksdb_flush_log_at_trx_commit = 1;
+SET @@session.rocksdb_write_disable_wal=ON;
+ERROR HY000: Got error 513 'Invalid argument: Cannot disable wal while sync writes are enabled (rocksdb_write_disable_wal cannot' from ROCKSDB
```
kamil-holubicki: The error message is truncated

mysql-test/suite/rocksdb_sys_vars/t/rocksdb_write_disable_wal_basic.test
```diff
@@ -17,3 +24,17 @@ INSERT INTO invalid_values VALUES('\'aaa\'');
...
+
+##########################################################################
+# Change the value of rocksdb_write_disable_wal to an incompatible value #
+##########################################################################
+# rocksdb_write_disable_wal=ON is incompatible with
+# synchronous writes (rocksdb_flush_log_at_trx_commit=1).
+# See https://jira.percona.com/browse/PS-7883 for details.
+SET @@session.rocksdb_write_disable_wal=OFF;
+SET @@global.rocksdb_flush_log_at_trx_commit = 1;
+--error 1296
```
kamil-holubicki: Please use ER_GET_ERRMSG instead of hardcoded error number

kamil-holubicki (commented):
I think it looks fine now. Just 3 cosmetic changes and we are good to go :)

kamil-holubicki (changes requested):
Error message is still truncated (only 1 occurrence has been fixed. 1 still left).
BTW, it looks like we don't have it covered by MTR

image

kamil-holubicki (commented):
LGTM. 
Great research, and thanks for your patient cooperation :)
Please squash all commits before merging (soft reset, then force push) and add a detailed message commit (what was the problem and how it was finally solved)

mysql-test/suite/rocksdb/t/rocksdb_bug7883_2.test
```diff
@@ -0,0 +1,39 @@
...
+                    PRIMARY KEY(ipkey), INDEX tt_12i0(d2, i1 ASC, ipkey),
+                    INDEX tt_12i1(ipkey, i1, d2))
+                   ROW_FORMAT=DYNAMIC ENGINE=RocksDB;
+
+REPLACE INTO tt_12  (ipkey, i1, d2) VALUES(2097, 6145, 0.00000);
+
+call mtr.add_suppression("Sync writes has to enable WAL. Switching to rocksdb_write_disable_wal = false for the time when sync writes are enabled.");
+ALTER TABLE tt_12 MODIFY COLUMN  ipkey INT(14), LOCK=DEFAULT, ALGORITHM=DEFAULT;
+
+--let $grep_file=$MYSQLTEST_VARDIR/log/mysqld.1.err
```
kamil-holubicki, line 37: Please add the check if rocksdb_write_disable_wal reports the proper value (false)

mysql-test/suite/rocksdb/t/rocksdb_bug7883_2.test
```diff
@@ -0,0 +1,39 @@
...
+SET @@session.rocksdb_write_disable_wal=OFF;
+SET @@global.rocksdb_flush_log_at_trx_commit = 1;
+
+--connection con1
+# Now rocksdb_flush_log_at_trx_commit = 1 and rocksdb_write_disable_wal=ON
+# The following statements have caused crash as
+# rocksdb_write_disable_wal=ON is incompatible with
+# synchronous writes (rocksdb_flush_log_at_trx_commit=1).
+# See https://jira.percona.com/browse/PS-7883 for details.
+
```
kamil-holubicki, line 22: Please check that rocksdb_write_disable_wal still reports the old (true) value. 
Add the comment that this is the issue not solved because it would need plugin API change.

kamil-holubicki (commented):
Sorry for spotting it out so late, but I have found that 1 MTR check is missed. Should be easy to add.
Also, the commit message says:
`The issue of displaying the currently compatible rocksdb_write_disable_wal
state to be used has not been resolved because it requires changing plugin
API. For example, by adding a callback that will adjust the value of the
variable when reading.`

But in general, we solved it! And it shows current value. The only problem is that just after setting global rocksdb_flush_log_at_trx_commit from session 2, we will still see the old value of rocksdb_write_disable_wal in session 1 until we execute the first transaction in session 1. After the transaction, the reported value is the proper one.
The same applies when we change rocksdb_flush_log_at_trx_commit back. We will see the "fallback value" of rocksdb_write_disable_wal until we execute the first transaction.

kamil-holubicki (approved):
LGTM

PR #4490, opened 2021-10-12
WDCZEN-9 feature 8.0: Allow running rocksdb MTR test cases on ZenFS storage in parallel

percona-ysorokin:
https://jira.percona.com/browse/WDCZEN-9

### DIFF
--- a/mysql-test/lib/My/ConfigFactory.pm
+++ b/mysql-test/lib/My/ConfigFactory.pm
@@ -459,7 +459,7 @@ sub post_check_client_groups {
 }
 
 sub resolve_at_variable {
-  my ($self, $config, $group, $option) = @_;
+  my ($self, $config, $group, $option, $worker) = @_;
 
   # Split the options value on last '.'
   my @parts       = split(/\./, $option->value());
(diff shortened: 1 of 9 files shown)


percona-ysorokin:
Jenkins build link (based on custom ps-build scripts with zenfs --parallel extensions)
https://ps3.cd.percona.com/job/percona-server-8.0-pipeline-PS-7749/58/

mysql-test/lib/My/ConfigFactory.pm
```diff
@@ -693,6 +695,9 @@ sub new_config {
     $hosts->{$host} = $baseport;
   }
 
+  # Worker ID
+  my $worker = $args->{'worker'};
```
oleksandr-kachan: It's used only in one place in this method. Shall we just pass it as &$rule($self, $config, $args->{'worker'}) below? Or maybe just declare it closer to the place where it is actually used.

mysql-test/mysql-test-run.pl
```diff
@@ -4177,6 +4177,7 @@ sub default_mysqld {
                                     baseport      => 0,
                                     user          => $opt_user,
                                     password      => '',
+                                    worker        => 1,
```
oleksandr-kachan: Just for better readability it may be better to add some constant for default worker ID. Up to you.

oleksandr-kachan (approved):
See some minor comments

oleksandr-kachan (approved):
LGTM

PR #4493, opened 2021-10-12
PS-1484: Fix slow log rotation when default log name is used

oleksandr-kachan:
https://jira.percona.com/browse/PS-1484

The slow query log rotation didn't work in case default log name is
used.

Reimplemented slow query log rotation functionality based on similar
fix at https://github.com/facebook/mysql-5.6/commit/30d9e653c71.

In addition:
    - Back ported fix for https://github.com/mysql/mysql-server/commit/02c845e6cc5c2679bcb429b621164431d06914de.
      It's needed in order to properly implement slow log rotation fix.
    - Reverted some changes in sql/binlog back to upstream state as they
      are not needed now.
    - Removed not implemented method File_query_log::new_file().
    - Removed a couple of duplicating MTR tests as all test cases for
      slow log rotation are now located in percona_slowlog_size_limits.


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/502/
https://ps57.cd.percona.com/job/percona-server-5.7-pipeline/23500/

sql/log.cc
```diff
@@ -2192,77 +2154,168 @@ bool Slow_log_throttle::log(THD *thd, bool eligible)
...
+  length = (size_t)(end - start);
+
+  MY_DIR *const dir_info = my_dir(buff, MYF(MY_DONT_SORT));
+  const struct fileinfo *file_info = dir_info->dir_entry;
+
+  for (uint i = dir_info->number_off_files; i--; file_info++)
+  {
+    if (strncmp(file_info->name, start, length) == 0 &&
+        file_info->name[length] == '.' &&
+        is_n_digit_number(file_info->name + length + 1, 6, &number))
```
satya-bodapati, line 2220: will this work if file have multiple '.' character ins slow query log file name?
Like 'slow-query.log' ? or 'mysqld.slow-query.log' ?

sql/log.cc
```diff
@@ -2192,77 +2154,168 @@ bool Slow_log_throttle::log(THD *thd, bool eligible)
...
-      DBUG_RETURN(error);
+  const char *start = str++;
+  while (isdigit(*str)) str++;
+  if (*str != 0 || str - start != n_digit) return false;
+
+  if (res) *res = atol(start);
+  return true;
+}
+
+static size_t get_last_extension(const char *const name)
```
satya-bodapati: would be nice if you can add function comment
1. the file patterns you are expecting 2. which part you are extracting
2. how do you deal with files that are not in the expected pattern?

sql/log.cc
```diff
@@ -2192,77 +2154,168 @@ bool Slow_log_throttle::log(THD *thd, bool eligible)
...
+{
+  DBUG_ENTER("get_last_extension");
```
satya-bodapati, line 2201: DBUG_TRACE everywhere?

sql/log.cc
```diff
@@ -2192,77 +2154,168 @@ bool Slow_log_throttle::log(THD *thd, bool eligible)
...
+  }
+
+  if (cur_log_ext > 0)
+  {
+    /* check if reached the maximum possible extension number */
+    if (cur_log_ext >= MAX_LOG_UNIQUE_FN_EXT)
+    {
+      sql_print_error("Log filename extension number exhausted: %06lu. \
+                       Please fix this by archiving old logs and \
+                       updating the index files.", cur_log_ext);
```
satya-bodapati, line 2270: may be add that option that will trigger auto deletion of slow query log files?

sql/log.cc
```diff
@@ -2192,77 +2154,168 @@ bool Slow_log_throttle::log(THD *thd, bool eligible)
...
+    else
+    {
+      snprintf(buff, sizeof(buff), "%s", name);
+    }
 
-end:
+    if (!mysql_file_stat(m_log_file_key, buff, &stat_area, MYF(0))) {
+      last_removed_ext = iter_log_ext;
+      DBUG_RETURN(false);
+    }
```
satya-bodapati, line 2346: why do we abort purging of log files if a file is missing? :thinking:

sql/log.cc
```diff
@@ -2192,77 +2154,168 @@ bool Slow_log_throttle::log(THD *thd, bool eligible)
...
+  DBUG_RETURN(false);
+}
+
+bool File_query_log::purge_logs()
+{
+  DBUG_ENTER("File_query_log::purge_logs");
+  mysql_mutex_assert_owner(&LOCK_log);
+
+  if (max_slowlog_files == 0 || cur_log_ext < 1 ||
+      last_removed_ext == cur_log_ext - 1)
```
satya-bodapati, line 2324: Lets say we intent to purge/remove the last few files and some of the files are missing. If we tolerate this and allow purge, can we keep this condition last_removed_ext == cur_log_ext -1 ?

satya-bodapati (changes requested):

sql/log.cc
oleksandr-kachan: Yes, it works fine. Will add test case for this.

sql/log.cc
oleksandr-kachan: Added some function comments

sql/log.cc
oleksandr-kachan: I guess there is no DBUG_TRACE in 5.7

sql/log.cc
oleksandr-kachan: Not sure we need this, at least for now. At the moment it behaves similarly to binlogs.

sql/log.cc
oleksandr-kachan: We stop checking files once we find one from the range of max_slowlog_files is already removed. No need to continue counting down to zero as older files were removed already.

sql/log.cc
oleksandr-kachan: This `last_removed_ext == cur_log_ext - 1` is here to check that there is at least one more file except currently used one. Otherwise there is nothing to purge.
So if there were 3 files and you change max_slowlog_files to 1 this condition shouldn't prevent it from removing all except currently active file.

sql/log.cc
satya-bodapati: @oleksandr-kachan ok, my point is what happens if an user deleted somefile manually. For example, .001, .002, .003, .004 exists and if user removes .003 manually, .001 and .002 will never be removed?

I guess, user shouln't touch these files? :thinking:

satya-bodapati (approved):

oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-pipeline/24562/testReport/

PR #4495, opened 2021-10-15
PS-6730: Use Last_errno in slow query log for errors only

oleksandr-kachan:
https://jira.percona.com/browse/PS-6730

Query execution warning gets added as Last_errno in slow query log.
Fix is based on upstream change
https://github.com/mysql/mysql-server/commit/0ae6ced64eb6243c3c5ab97ad8093ae60b56ccc5.
The thd->last_errno may contain not actual data at the moment when
slow log is printed. It also gets updated with warnings if there
are any during query execution.
Populate Last_errno with info from statement Diagnostics Area instead.


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/503/

satya-bodapati (approved):

PR #4497, opened 2021-10-18
PS-7917 - post-fix - Added check for ON variables

altmannmarcelo:
Added a check of tokudb-enabled / tokudb-backup-enabled to validate
if ON/on is set.


altmannmarcelo:
```
marcelo@marce-bld:/work/ps/src/8.0$ echo tokudb-enabled=on | grep -v '#' | egrep -i tokudb[_-]enabled | egrep -ic '=1|=true|=on'
1
marcelo@marce-bld:/work/ps/src/8.0$ echo tokudb-enabled=ON | grep -v '#' | egrep -i tokudb[_-]enabled | egrep -ic '=1|=true|=on'
1
```

george-lorch (approved):

PR #4507, opened 2021-10-22
WDCZEN-8 feature: Introduce WITH_ZBD CMake option to allow linking libzbd into ha_rocksdb.so statically

percona-ysorokin:
https://jira.percona.com/browse/WDCZEN-8

Added 'libzbd' as a submodule.
Introduced 'WITH_ZBD=system|bundled' CMake option ('bundled' by default).


oleksandr-kachan (approved):
LGTM

PR #4512, opened 2021-10-26
PS-6774: Add test for Last_errno in case DD table is accessed

oleksandr-kachan:
https://jira.percona.com/browse/PS-6774

In https://github.com/mysql/mysql-server/commit/a58bf5ef8806941bbd417f8ab17c818c48e21822
access to internal DD tables was prohibited. But SELECT statement
still should not fail in case DD table names are submitted as strings in
WHERE clauses. Add test case to verify Last_errno field in slow query
log doesn't contain any error in this case.
The issue with Last_errno not being empty in this case was previously
fixed in https://github.com/percona/percona-server/pull/4506.

### DIFF
--- a/mysql-test/t/percona_slow_extended_log_error.test
+++ b/mysql-test/t/percona_slow_extended_log_error.test
@@ -33,6 +33,18 @@ INSERT INTO t2(f1) VALUES (31.400191);
 --let grep_pattern = ^.*Last_errno: 0 .*\$
 --source include/log_grep.inc
 
+--echo
+--echo # PS-6774 - Unexpected error in Last_errno in case internal DD tables are
+--echo             accessed. Statement des not fail in case DD table names
+--echo             are submitted as strings in WHERE clause.
+--echo
+--let log_file=percona.slow_extended.log_error_4
+--source include/log_start.inc
+SELECT SCHEMA_NAME as Test FROM information_schema.schemata WHERE SCHEMA_NAME NOT IN ('mysql', 'performance_schema', 'information_schema');
+--source include/log_stop.inc
+--let grep_pattern = ^.*Last_errno: 0 .*\$
+--source include/log_grep.inc
+
 DROP TABLE t1, t2;
 
 --source include/log_cleanup.inc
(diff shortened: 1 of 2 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1082/

mysql-test/t/percona_slow_extended_log_error.test
```diff
@@ -33,6 +33,18 @@ INSERT INTO t2(f1) VALUES (31.400191);
...
 --source include/log_grep.inc
 
+--echo
+--echo # PS-6774 - Unexpected error in Last_errno in case internal DD tables are
+--echo             accessed. Statement des not fail in case DD table names
+--echo             are submitted as strings in WHERE clause.
+--echo
+--let log_file=percona.slow_extended.log_error_4
+--source include/log_start.inc
+SELECT SCHEMA_NAME as Test FROM information_schema.schemata WHERE SCHEMA_NAME NOT IN ('mysql', 'performance_schema', 'information_schema');
```
satya-bodapati, line 43: @oleksandr-kachan why does accessing DD tables names cause the err_no field to be filled? Can you please find out who fills the thd->m_da->mysql_errno?

satya-bodapati (changes requested):

mysql-test/t/percona_slow_extended_log_error.test
oleksandr-kachan: @satya-bodapati, to make sure we are on the same page here. The thd->get_stmt_da()->mysql_errno() is not populated in this case. The issue was with thd->last_errno before fix from https://github.com/percona/percona-server/commit/b79645143cd773cdc61c439f269c69520a2d3e31.

satya-bodapati: Right @oleksandr-kachan , so this question should have been more relevant with your other fix. WIthout the other fix, who has set the errno?

oleksandr-kachan: @satya-bodapati, it was set at https://github.com/percona/percona-server/blob/8.0/sql/sql_parse.cc#L6141

satya-bodapati: I see, does the query fail? If it doesn't, do we reset the diagnostic area before the completion of the query?

oleksandr-kachan: @satya-bodapati, query doesn't fail. As I can see, error in diagnostic area is not set in this case.

satya-bodapati: got it oleksandar. my_error() sets it may be cleared at later point of time

satya-bodapati (approved):

PR #4515, opened 2021-10-28
WDCZEN-13 feature: Check for libgflags presence into RocksDB CMakeLists.txt when WITH_ZENFS_UTILITY=ON

percona-ysorokin:
https://jira.percona.com/browse/WDCZEN-13

Added new check to the RocksDB CMakeLists.txt when 'WITH_ZENFS_UTILITY'
is 'ON' that ensures that 'gflags' libarary and headers are installed in
the system.


oleksandr-kachan (approved):
LGTM

PR #4518, opened 2021-10-30
Fixed PS-7882 (Assertion `m_output_buf->m_curr_offset <= m_output_buf->m_total_size' failed)

i-rinat:
https://jira.percona.com/browse/PS-7882

Problem:
There was a missing size check in `Rdb_index_merge::add()` just after `merge_buf_write()` emptied the unsorted buffer. With extremely small merge buffer sizes, moderately big data could trigger a buffer overflow. Shortly after, a failed assertion terminates the server (debug builds only).

Fix:
Check if the unsorted buffer size is enough to fit a key-value pair right after an unsorted buffer is flushed and cleared.

### DIFF
--- a/storage/rocksdb/rdb_index_merge.cc
+++ b/storage/rocksdb/rdb_index_merge.cc
@@ -164,13 +164,27 @@ int Rdb_index_merge::add(const rocksdb::Slice &key, const rocksdb::Slice &val) {
     */
     if (m_offset_tree.empty()) {
       LogPluginErrMsg(ERROR_LEVEL, 0,
-                      "Sort buffer size is too small to process merge. Please "
-                      "set merge buffer size to a higher value.");
+                      "Current value of rocksdb_merge_buf_size=%llu is too "
+                      "small. At least %u bytes required.",
+                      m_rec_buf_unsorted->m_total_size, total_offset);
       return HA_ERR_ROCKSDB_MERGE_FILE_ERR;
     }
 
     if (merge_buf_write()) {
       LogPluginErrMsg(ERROR_LEVEL, 0, "Error writing sort buffer to disk.");
+      return HA_ERR_ROCKSDB_MERGE_FILE_ERR;
+    }
+
+    /*
+      The unsorted buffer may be too small for the key-value pair.
+    */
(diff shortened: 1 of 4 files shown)


i-rinat:
Updated the patch.
Added `source include/have_debug.inc` to the test, since assertions can only be checked in debug builds.
Added and one more `INSERT` statement to the test. Helps to trigger the bug in facebook/mysql-5.6 version of the code.

storage/rocksdb/rdb_index_merge.cc
```diff
@@ -173,6 +173,18 @@ int Rdb_index_merge::add(const rocksdb::Slice &key, const rocksdb::Slice &val) {
...
       return HA_ERR_ROCKSDB_MERGE_FILE_ERR;
     }
+
+    /*
+      The unsorted buffer may be too small for the key-value pair.
+    */
+    const uint data_size = RDB_MERGE_CHUNK_LEN + RDB_MERGE_KEY_DELIMITER +
+                           RDB_MERGE_VAL_DELIMITER + key.size() + val.size();
+    if (data_size > m_rec_buf_unsorted->m_total_size) {
+      LogPluginErrMsg(ERROR_LEVEL, 0,
```
kamil-holubicki, line 184: Wouldn't it be more informative to print something like:
`Current value of rocksdb_merge_buf_size=ABC is too small. At least XYZ bytes required`

i-rinat: Yes, you are right, it is better as users will get a clear action plan alongside with the error message.

I was in doubt whenever I should make changes to the error message as the message in the patch is a duplicate of exactly the same message half a dozen lines above. The error of overrunning a buffer is conceptually the same, but is checked twice. I'd expect both cases to produce the same error, so they are identical. It's possible to easily remove duplication, but the only way I found involved `goto` backwards, which is not immediately understandable. For example a reader (human) may ask if infinite looping is possible. It's not, but it would take time to prove that.

I've updated the message in both places in the code and updated the test.

kamil-holubicki:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1147/

kamil-holubicki (approved):
lgtm

PR #4524, opened 2021-11-08
PS-7940 ON DELETE CASCADE with generated column crashes in innobase_g…

rahulmalik87:
…et_computed_value

https://jira.percona.com/browse/PS-7940

Problem:
Virtual column template of child table is not populated when cascade
delete happened on parent table.  Failing to initliaze it causes a crash
when we are trying remove the secondary index rows defined on virtual column .

Fix:

As part of fix e5e8c9d1b2e0 MySQL handle case when child virtual column
depends on forign key. To extend the fix we are now initliazing virtual column template
if child table has any virtual column.

### DIFF
--- a/mysql-test/suite/innodb/t/foreign_key.test
+++ b/mysql-test/suite/innodb/t/foreign_key.test
@@ -253,3 +253,28 @@ SELECT `date_sent` FROM `email_stats` WHERE `generated_sent_email` = '2020-';
 #clean up.
 DROP TABLE `email_stats`;
 DROP TABLE `emails`;
+
+--echo #
+--echo # PS-7940 :ON DELETE CASCADE with generated column crashes in innobase_get_computed_value
+--echo #
+
+CREATE TABLE `t1` (`t1_id` int unsigned NOT NULL AUTO_INCREMENT, PRIMARY KEY(`t1_id`)) ENGINE=InnoDB;
+INSERT INTO `t1` VALUES (10);
+CREATE TABLE `t2` (
+  `t2_id` int unsigned NOT NULL AUTO_INCREMENT,
+  `c1` decimal(14,2) NOT NULL DEFAULT '0.00',
+  `t1_id` int unsigned DEFAULT NULL,
+  `testycol` bit(1) GENERATED ALWAYS AS (`c1` <> 0) VIRTUAL,
+  PRIMARY KEY (`t2_id`),
+  UNIQUE KEY `t1_id` (`t1_id`),
+  KEY `testycol` (`testycol`),
+  CONSTRAINT `t2_ibfk_3` FOREIGN KEY (`t1_id`) REFERENCES `t1` (`t1_id`) ON DELETE CASCADE
+) ENGINE=InnoDB;
+INSERT INTO `t2` (`t2_id`, `c1`, `t1_id`) VALUES (1,5.00,10);
+
+--source include/restart_mysqld.inc
+
+DELETE FROM `t1` WHERE `t1_id`=10;
+
+DROP TABLE `t2`;
+DROP TABLE `t1`;
--- a/storage/innobase/row/row0ins.cc
+++ b/storage/innobase/row/row0ins.cc
@@ -1306,8 +1306,13 @@ row_ins_foreign_check_on_constraint(
 						 clust_index, tmp_heap);
 	}
 
-	if (cascade->is_delete && foreign->v_cols != NULL
-	    && foreign->v_cols->size() > 0
+	/* A cascade delete from the parent table triggers delete on the child
+	table. Before a clustered index record is deleted in the child table,
+	a copy of row is built to remove secondary index records. This copy of
+	the row requires virtual columns to be materialized. Hence, if child
+	table has any virtual columns, we have to initialize virtual column
+	template */
+	if (cascade->is_delete && dict_table_get_n_v_cols(table) > 0
 	    && table->vc_templ == NULL) {
 		innobase_init_vc_templ(table);
 	}
--- a/mysql-test/suite/innodb/r/foreign_key.result
+++ b/mysql-test/suite/innodb/r/foreign_key.result
@@ -297,3 +297,23 @@ date_sent
 2020-10-22 13:32:41
 DROP TABLE `email_stats`;
 DROP TABLE `emails`;
+#
+# PS-7940 :ON DELETE CASCADE with generated column crashes in innobase_get_computed_value
+#
+CREATE TABLE `t1` (`t1_id` int unsigned NOT NULL AUTO_INCREMENT, PRIMARY KEY(`t1_id`)) ENGINE=InnoDB;
+INSERT INTO `t1` VALUES (10);
+CREATE TABLE `t2` (
+`t2_id` int unsigned NOT NULL AUTO_INCREMENT,
+`c1` decimal(14,2) NOT NULL DEFAULT '0.00',
+`t1_id` int unsigned DEFAULT NULL,
+`testycol` bit(1) GENERATED ALWAYS AS (`c1` <> 0) VIRTUAL,
+PRIMARY KEY (`t2_id`),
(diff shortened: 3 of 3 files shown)


storage/innobase/row/row0ins.cc
```diff
@@ -1279,9 +1279,8 @@ row_ins_foreign_check_on_constraint(
 						 clust_index, tmp_heap);
 	}
 
-	if (cascade->is_delete && foreign->v_cols != NULL
-	    && foreign->v_cols->size() > 0
-	    && table->vc_templ == NULL) {
+	if (cascade->is_delete && dict_table_get_n_v_cols(table) > 0
+			&& table->vc_templ == NULL) {
 		innobase_init_vc_templ(table);
 	}
```
satya-bodapati: how about fixing the row_ins_foreign_fill_virtual()? I think it is not being called to set out the template initialization?

satya-bodapati (changes requested):

satya-bodapati (changes requested):
I don't think my comment is addressed

storage/innobase/row/row0ins.cc
rahulmalik87: yes. It is not called for same reason . we have a check  that[ foreign key](https://github.com/percona/percona-server/blob/a73ccf393d8c41410b7f25f657cc178b2d860a05/storage/innobase/row/row0ins.cc#L1330) should have the virtual column.

satya-bodapati: can we remove this part of the code and fix that instead? ie call row_ins_foreign_fill_virtual() based on child virtual columns presence?

rahulmalik87: we do not call row_ins_foreign_fill_virtual for this code path. It is called in case DICT_FOREIGN_ON_DELETE_SET_NULL. here the code flow DICT_FOREIGN_ON_DELETE_CASCADE

storage/innobase/row/row0ins.cc
```diff
@@ -1279,9 +1279,8 @@ row_ins_foreign_check_on_constraint(
 						 clust_index, tmp_heap);
 	}
 
-	if (cascade->is_delete && foreign->v_cols != NULL
-	    && foreign->v_cols->size() > 0
-	    && table->vc_templ == NULL) {
+	if (cascade->is_delete && dict_table_get_n_v_cols(table) > 0
+			&& table->vc_templ == NULL) {
```
satya-bodapati: Can you please add a comment on why we initialize virtual column template?

Some thing like
```
/* A cascade delete from the parent table triggers delete on the child table. Before a clustered index record is deleted in the child table, a copy of row is built to remove secondary index records. This copy of the row requires virtual columns to be materialized. Hence, if child table has any virtual columns, we have to initialize virtual column template */
```
Please lookout for 80 column char limit when pasting the comment to code :wink:

satya-bodapati (changes requested):

storage/innobase/row/row0ins.cc
satya-bodapati: ok then row_ins_foreign_fill_virtual() becomes dummy operation? Can you please check with SET_NULL and see if row_ins_foreign_fill_virtual() is called again? And this double initialization is safe?

storage/innobase/row/row0ins.cc
rahulmalik87: thank for well explained comment. Added to the code.

satya-bodapati: @rahulmalik87 Please consider fixing commit message too :)

storage/innobase/row/row0ins.cc
satya-bodapati: @rahulmalik87 did you check?

rahulmalik87: SET_NULL doens't go inside this method since we have additional check [cascade->delete](https://github.com/percona/percona-server/blob/a73ccf393d8c41410b7f25f657cc178b2d860a05/storage/innobase/row/row0ins.cc#L1282)

storage/innobase/row/row0ins.cc
rahulmalik87: Let me know if commit message is fine now.

storage/innobase/row/row0ins.cc
satya-bodapati: Rahul, am asking you check with constraint ON DELETE  SET NULL

rahulmalik87: in case of ON DELETE SET NULL. cascade->delete is set to 0. 
lldb) p cascade->is_delete
(ulint) $1 = 0
(lldb) p node->is_delete
(ulint) $2 = 1
(lldb)

satya-bodapati: cascade delete as 1 with ON DELETE as NULL not possible?

rahulmalik87: ON DELETE either SET NULL or Cascade. so CASCADE and SET NUL is not possible

storage/innobase/row/row0ins.cc
satya-bodapati: This && should start position below 'c' in cascade in above line

mysql-test/suite/innodb/t/foreign_key.test
```diff
@@ -113,3 +113,20 @@ WHERE table_name = 't2';
 
 DROP TABLE t2;
 DROP TABLE t1;
+CREATE TABLE `t1` (`t1_id` int unsigned NOT NULL AUTO_INCREMENT, PRIMARY KEY(`t1_id`)) ENGINE=InnoDB;
+INSERT INTO `t1` VALUES (10);
+CREATE TABLE `t2` (
```
satya-bodapati, line 263: Please add header
--echo #
--echo # PS-7940 : ON ..
--echo #

satya-bodapati (changes requested):
minor comments. looks good

storage/innobase/row/row0ins.cc
rahulmalik87: fixed

mysql-test/suite/innodb/t/foreign_key.test
rahulmalik87: fixed

storage/innobase/row/row0ins.cc
```diff
@@ -1279,9 +1279,14 @@ row_ins_foreign_check_on_constraint(
...
-	    && foreign->v_cols->size() > 0
-	    && table->vc_templ == NULL) {
+	/* A cascade delete from the parent table triggers delete on the child
+	table. Before a clustered index record is deleted in the child table,
+	a copy of row is built to remove secondary index records. This copy of
+	the row requires virtual columns to be materialized. Hence, if child
+	table has any virtual columns, we have to initialize virtual column
+	template */
+	if (cascade->is_delete && dict_table_get_n_v_cols(table) > 0
+            && table->vc_templ == NULL) {
```
satya-bodapati: @rahulmalik87 alignment still looks wrong, check tabs vs spaces :)

satya-bodapati:
@rahulmalik87 (incase you address your review comment and are ready for review), since Jenkins is not working, I suggest you  run all innodb suites and paste result here

rahulmalik87:
> @rahulmalik87 (incase you address your review comment and are ready for review), since Jenkins is not working, I suggest you run all innodb suites and paste result here

@satya-bodapati https://ps57.cd.percona.com/view/5.7/job/percona-server-5.7-param/525/testReport/

rahulmalik87:
latest result on lab machine
Spent 4792.967 of 1088 seconds executing testcases

Completed: Failed 4/316 tests, 98.73% were successful.

Failing test(s): innodb.mysqld_core_dump_without_buffer_pool_dynamic shutdown_report innodb.mysqld_core_dump_without_buffer_pool innodb.mysqld_core_dump_without_buffer_pool_with_resizing

satya-bodapati (approved):

PR #4526, opened 2021-11-10
PS-7958: Fix crash during fulltext search index with special character

oleksandr-kachan:
https://jira.percona.com/browse/PS-7958

Server crashes during fulltext search in case there is a null character
symbol in the middle of a search string. The issue reproducible for
NGRAM and MECAB parsers.
Updated ngram_parse() and mecab_parse() method to skip control characters
while parsing a string.


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1104/

satya-bodapati (approved):

PR #4531, opened 2021-11-11
PS-7968: Implement BiDi scan for PS via Azure pipelines

oleksandr-kachan:
https://jira.percona.com/browse/PS-7968


inikep (approved):
LGTM

PR #4532, opened 2021-11-11
PS-7968: Implement BiDi scan for PS via Azure pipelines

oleksandr-kachan:
https://jira.percona.com/browse/PS-7968

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,459 @@
+jobs:
+- job:
+  timeoutInMinutes: 240
+  pool:
+    vmImage: $(imageName)
+
+  variables:
+    UBUNTU_CODE_NAME: bionic
+    BOOST_VERSION: boost_1_59_0
+    BOOST_DIR: $(Pipeline.Workspace)/boost
+    USE_CCACHE: 1
+    CCACHE_DIR: $(Pipeline.Workspace)/ccache
+    CCACHE_COMPRESS: 1
+    CCACHE_COMPRESSLEVEL: 9
+    CCACHE_CPP2: 1
+    CCACHE_MAXSIZE: 2G
+    OS_NAME: $(Agent.OS)
+
+  strategy:
+    matrix:
+      macOS 10.14 Release:
+        imageName: 'macOS-10.14'
+        Compiler: clang
+        BuildType: RelWithDebInfo
+
+      ${{ if ne(variables['Build.Reason'], 'IndividualCI') }}:
+        macOS 10.14 Debug:
+          imageName: 'macOS-10.14'
+          Compiler: clang
+          BuildType: Debug
+
+      # clang-5 and newer compilers
+      Ubuntu Bionic clang-13 Release:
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -1,4 +1,16 @@
...
+  pool:
+    vmImage: 'ubuntu-18.04'
+
+  steps:
+  - checkout: self
+    fetchDepth: 32
+
+  - script: |
+      git fetch origin 5.7
+      python $(Build.SourcesDirectory)/scripts/find_unicode_control.py -p bidi -v $(git diff --name-only --relative --diff-filter AMR origin/5.7 -- . | tr '\n' ' ')
```
inikep: 8.0 instead of 5.7

inikep: It checks also e.g. `.test` and `.result` files. You may limit checking to some types of files e.g.: 
```git diff --name-only origin/5.7 *.c *.cc *.cpp *.h *.hpp *.i *.ic *.ih```

Comment (oleksandr-kachan) [reply azure-pipelines.yml]: Done

Comment (oleksandr-kachan) [reply azure-pipelines.yml]: @inikep, maybe better to exclude ones which we are sure we don't want to check (like .test and .result)?

File: azure-pipelines.yml [hunk 2]
```diff
@@ -1,4 +1,16 @@
...
+- job: BiDiScan
+  pool:
+    vmImage: 'ubuntu-18.04'
+
+  steps:
+  - checkout: self
+    fetchDepth: 32
+
+  - script: |
+      git fetch origin 8.0
```
inikep, line 11: Did you try without `fetchDepth: 32` ?
`git fetch origin 8.0` may be not required and Azure might be faster. I'm not sure but it's worth checking.

azure-pipelines.yml
inikep: Sure, it will be better. For now you can keep all files and improve it later.

azure-pipelines.yml
oleksandr-kachan: This 'fetchDepth: 32' came  from existing config. Do you think we could remove it there as well?

inikep: It's tricky. 'fetchDepth: 32' asks `git` to download only 32 last commits from a repo history. It makes less data to be downloaded for client. From the other side, it requires a git server to prepare and compress a repo history just for your call so it makes a git server slower. We use 'fetchDepth: 32' mainly because it takes less disk space on Azure VM.

inikep: For BiDi you can experiment what is faster (because we have a lot of free disk space) but I don't expect a huge difference.

inikep (approved):
LGTM

azure-pipelines.yml
oleksandr-kachan: @inikep, without fetchDepth and 'git fetch' the job runs for about 90s longer.

inikep: Thanks for checking

PR #4538, opened 2021-11-23
PS-7895 arch_oper_mutex is not released inside Arch_Page_Sys::track_page

rahulmalik87:
https://jira.percona.com/browse/PS-7895

Problem:

Hang in checkpoint flusher thread due to arch_oper_mutex can't be acquired.
It can happen when page-tracking in enabled and server is under heavy load.

The Arch_Page_Sys::track_page is supposed to be released and reacquire the
archive mutex through function wait_flush_archiver. If is_flushable is false
it never enters in while loop in wait_flush_archiver.
so arch_oper_mutex is never relased.

Fix:

Release and reacquire arch_oper_mutex inside Arch_Page_Sys::track_page

### DIFF
--- a/storage/innobase/arch/arch0page.cc
+++ b/storage/innobase/arch/arch0page.cc
@@ -32,6 +32,7 @@ this program; if not, write to the Free Software Foundation, Inc.,
 #include "arch0page.h"
 #include "arch0recv.h"
 #include "clone0clone.h"
+#include "raii/sentry.h"
 #include "srv0start.h"
 
 #ifdef UNIV_DEBUG
@@ -1716,9 +1717,10 @@ void Arch_Page_Sys::track_page(buf_page_t *bpage, lsn_t track_lsn,
   }
 
   /* We need to track this page. */
-  arch_oper_mutex_enter();
 
   while (true) {
+    arch_oper_mutex_enter();
+    raii::Sentry<> arch_oper_mutex_guard{[&]() { arch_oper_mutex_exit(); }};
     if (m_state != ARCH_STATE_ACTIVE) {
       break;
     }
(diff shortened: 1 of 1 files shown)


satya-bodapati (approved):

storage/innobase/arch/arch0page.cc
```diff
@@ -1716,10 +1716,11 @@ void Arch_Page_Sys::track_page(buf_page_t *bpage, lsn_t track_lsn,
   }
 
   /* We need to track this page. */
-  arch_oper_mutex_enter();
 
   while (true) {
+    arch_oper_mutex_enter();
     if (m_state != ARCH_STATE_ACTIVE) {
+      arch_oper_mutex_exit();
```
venkatesh-prasad-v: Not an issue.

FYI, `sentry` objects can be used in such cases to ensure that the lock is released when the control gets out of scope.

Usage is as below
```
#include "raii/sentry.h" /* raii::Sentry */
...
...
arch_oper_mutex_enter();
/* Create a sentry object to call `arch_oper_mutex_exit()` on exit */
raii::Sentry<> arch_oper_mutex_guard{[&]() { arch_oper_mutex_exit(); }};
```

satya-bodapati: Good suggestion. no need to create dummy class for RAII

venkatesh-prasad-v (approved):
Looks good. Please wait for the final approval from @satya-bodapati to merge it.

satya-bodapati (commented):
@rahulmalik87 did you test your fix with the latest changes (sentry object). Please check if the hang is not reproducible.

rahulmalik87:
Could not reproduce the bug on latest 8.0.27 branch

PR #4542, opened 2021-11-30
PS-7929 : Wrong row lock number in SEIS

satya-bodapati:
https://jira.percona.com/browse/PS-7929

Problem:
--------

The number of row locks in "SHOW ENGINE INNODB STATUS" OUTPUT is wrong.
When the trx inserts many duplicate records, the number of row locks
in, "SHOW ENGINE INNODB STATUS" output is wrong.

InnoDB maintains this locks as a counter in trx->lock.n_rec_locks. They are
incremented on lock acquisition and decremented on lock release.

In case of duplicate records, some lock acquisitions increment the counter
for the lock which is already acquired/set

Fix
---
Set the lock bit only if it is not already set/acquired.

### DIFF
--- a/storage/innobase/lock/lock0lock.cc
+++ b/storage/innobase/lock/lock0lock.cc
@@ -1842,7 +1842,9 @@ lock_rec_add_to_queue(
 
 		if (lock != NULL) {
 
-			lock_rec_set_nth_bit(lock, heap_no);
+			if (!lock_rec_get_nth_bit(lock, heap_no)) {
+				lock_rec_set_nth_bit(lock, heap_no);
+			}
 
 			return;
 		}


storage/innobase/lock/lock0lock.cc
```diff
@@ -1842,7 +1842,9 @@ lock_rec_add_to_queue(
 
 		if (lock != NULL) {
 
-			lock_rec_set_nth_bit(lock, heap_no);
+			if (!lock_rec_get_nth_bit(lock, heap_no)) {
```
rahulmalik87, line 1845: Do we need to same with lock_rec_reset_nth_bit   i.e reset only if it is set?

satya-bodapati: we already do. if bit !=0, it means it is set, only then we decrement.

rahulmalik87 (approved):
LGTM

PR #4544, opened 2021-12-01
PS-7988: Improve BiDiScan Azure job to properly handle empty commits (5.7)

oleksandr-kachan:
https://jira.percona.com/browse/PS-7988

It is possible to have PR which doesn't actually change any file
(Null merge for example). Improve BiDiScan Azure job to skip
actual check in case there are not any files changed in PR.


altmannmarcelo (approved):
LGTM

PR #4548, opened 2021-12-06
PS-7931: Implement slow query log rotation (8.0)

oleksandr-kachan:
https://jira.percona.com/browse/PS-7931

The implementation is ported from PS 5.7.

Added the following variables to controll the functionality:
- max_slowlog_size - Slow query log will be rotated automatically when
                     the size exceeds this value. The default is 0,
                     don't limit the size.
- max_slowlog_files - Maximum number of slow query log files. Used with
                      --max-slowlog-size this can be used to limit the total
                      amount of disk space used for the slow query log.
                      Default is 0, don't limit.

### DIFF
--- a/mysql-test/include/assert_number_of_files.inc
+++ b/mysql-test/include/assert_number_of_files.inc
@@ -1,0 +1,38 @@
+# ==== Purpose ====
+#
+# Assert that number of files matching given $file_spec is equal to expected.
+#
+# ==== Usage ====
+#
+# --let file_spec = PATH_REGEX
+# --let expected_number = INTEGER
+# --source include/assert_number_of_files.inc
+#
+# Parameters:
+#   $file_spec
+#     The path with wildcards e.g. $MYSQLTEST_VARDIR/abcd*
+#
+#   $expected_number
+#     The expected number of files
+
+--perl
+  use strict;
+  my $dir = $ENV{'MYSQL_TMP_DIR'} or die "MYSQL_TMP_DIR not set";
+  my $file_spec = $ENV{'file_spec'} or die "file_spec not set";
+  my @files = <$ENV{'file_spec'}>;
+  open (OUTPUT, ">$dir/number_of_files.inc") ;
+  print OUTPUT "--let \$number_of_files = ", scalar(@files), "\n";
+  close (OUTPUT);
+EOF
+--source $MYSQL_TMP_DIR/number_of_files.inc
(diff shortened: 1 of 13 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1126/

sql/sys_vars.h
```diff
@@ -1107,6 +1107,20 @@ class Sys_var_charptr : public sys_var {
...
+      if (!new_val) return true;
+    } else {
+      new_val = nullptr;
+    }
+    if (flags & ALLOCATED) my_free(global_var(char *));
+    flags |= ALLOCATED;
+    global_var(char *) = new_val;
+    return false;
+  }
+
```
satya-bodapati: Did we have this in 5.7 PR? Can you please point me to it?

satya-bodapati (changes requested):

sql/sys_vars.h
oleksandr-kachan: We don't have it for 5.7 (probably missed it while moving code to 5.7) but it looks like ALLOCATED flag should be handled for 5.7 as well (even though I don't see ALLOCATED to be actually used for any sysvar).

satya-bodapati: if it is not used, and it can work (check other variables) without setting ALLOCATED flag, better to remove the entire func in 8.0 too

oleksandr-kachan: @satya-bodapati, do you think it would be wise to remove this ALLOCATED flag at all (in case it is really not used anywhere)?

oleksandr-kachan: > better to remove the entire func in 8.0 too

Done

oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1128/

satya-bodapati (approved):

sql/sys_vars.h
satya-bodapati: btw, I see some other places where it is used. not sure if we can remove it.

satya-bodapati (changes requested):
@oleksandr-kachan circleCI build is failing. Please check formatting

oleksandr-kachan:
> @oleksandr-kachan circleCI build is failing. Please check formatting

Done

satya-bodapati (approved):

PR #4550, opened 2021-12-06
PS-7044 feature 8.0: Provide Enterprise Encryption UDFs for OpenSSL

percona-ysorokin:
https://jira.percona.com/browse/PS-7044


plugin/encryption_udf/encryption_udf.cc
```diff
@@ -0,0 +1,258 @@
...
+    ctx.set_arg_type(0, STRING_RESULT);
+
+    // arg1 - @key_len
+    ctx.mark_arg_nullable(1, false);
+    ctx.set_arg_type(1, INT_RESULT);
+  }
+  ~create_asymmetric_priv_key_impl() { DBUG_TRACE; }
+
+  mysqlpp::udf_result_t<STRING_RESULT> calculate(
+      const mysqlpp::udf_context &args);
```
oleksandr-kachan: Parameter name in declaration and implementation differs. Same for other 'calculate()' methods in this file.

plugin/encryption_udf/encryption_udf.cc
```diff
@@ -0,0 +1,258 @@
...
+
+mysqlpp::udf_result_t<STRING_RESULT> create_asymmetric_priv_key_impl::calculate(
+    const mysqlpp::udf_context &ctx) {
+  DBUG_TRACE;
+
+  auto algorithm = ctx.get_arg<STRING_RESULT>(0);
+  if (algorithm.data() == nullptr)
+    throw std::invalid_argument("Algorithm cannot be NULL");
+  if (algorithm != "RSA")
+    throw std::invalid_argument("Invalid algorithm specified");
```
oleksandr-kachan: May be useful to print actual value here

extra/opensslpp/include/opensslpp/big_number_fwd.hpp
```diff
@@ -0,0 +1,26 @@
...
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA */
+
+#ifndef OPENSSLPP_BIG_NUMBER_FWD_HPP
+#define OPENSSLPP_BIG_NUMBER_FWD_HPP
+
+namespace opensslpp {
+
+class big_number;
```
dutow, line 22: This is purely a cosmetic / opinion-based question: do we need separate header files for every forward declaration? It might be easier to maintain/use just a single "opensslpp_fwd.hpp" with all of them.

extra/opensslpp/include/opensslpp/core_error.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+#ifndef OPENSSLPP_CORE_ERROR_HPP
+#define OPENSSLPP_CORE_ERROR_HPP
+
+#include <stdexcept>
+#include <string>
+
+namespace opensslpp {
+
+class core_error : public std::runtime_error {
```
dutow, line 27: Shouldn't `raise_with_error_string` be used more in the code, or shouldn't the constructor use the same `err_get_error` logic? Calling it should be harmless if there's no error, and many places which simply construct a `core_error` could use the additional error details.

extra/opensslpp/include/opensslpp/dh_key.hpp
```diff
@@ -0,0 +1,92 @@
...
+#include <opensslpp/accessor_fwd.hpp>
+#include <opensslpp/big_number_fwd.hpp>
+
+namespace opensslpp {
+
+class dh_key final {
+  friend class accessor<dh_key>;
+
+ public:
+  static const std::uintmax_t generator_2;
```
dutow, line 37: These could be `constexpr`, and that would allow inline initialization.

extra/opensslpp/include/opensslpp/rsa_key.hpp
```diff
@@ -0,0 +1,86 @@
...
+#include <opensslpp/big_number_fwd.hpp>
+#include <opensslpp/rsa_padding_fwd.hpp>
+
+namespace opensslpp {
+
+class rsa_key final {
+  friend class accessor<rsa_key>;
+
+ public:
+  static const big_number exponent_3;
```
dutow, line 38: Could also be `constexpr`

extra/opensslpp/include_private/opensslpp/core_allocated_buffer.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+#ifndef OPENSSLPP_CORE_ALLOCATED_BUFFER
+#define OPENSSLPP_CORE_ALLOCATED_BUFFER
+
+#include <cstdint>
+#include <memory>
+
+namespace opensslpp {
+
+class core_allocated_buffer final {
```
dutow: I only see a single use of this class, which only treats it as a unique_ptr with a custom deleter, most of the functionality doesn't seem to be used anywhere.

Do we really need this class and all of its functionality?

extra/opensslpp/include_private/opensslpp/digest_context_accessor.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include <openssl/evp.h>
+
+#include <opensslpp/digest_context_fwd.hpp>
+
+#include "opensslpp/typed_accessor.hpp"
+
+namespace opensslpp {
+
+using digest_context_accessor = typed_accessor<digest_context, EVP_MD_CTX>;
```
dutow, line 28: I have a similar question for these as the for the _fwd files: wouldn't a single _accessor header be better?

extra/opensslpp/include_private/opensslpp/rsa_padding_conversions.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include <cassert>
+
+#include <openssl/rsa.h>
+
+#include <opensslpp/rsa_padding.hpp>
+
+namespace opensslpp {
+
+inline int rsa_padding_to_native_padding(rsa_padding padding) noexcept {
```
dutow, line 28: The enum class could use an `int` base type, native values, and then this function could be a simple `static_cast`

extra/opensslpp/src/opensslpp/dh_compute_operations.cpp
```diff
@@ -0,0 +1,102 @@
...
+    throw core_error{
+        "cannot compute shared key as DH key does not have private component"};
+
+  auto function = get_compute_key_function(padding);
+
+  // TODO: use c++17 non-const std::string::data() member here
+  using buffer_type = std::vector<unsigned char>;
+  buffer_type res(private_key.get_size_in_bytes());
+  auto compute_status = (*function)(
+      res.data(), DH_get0_pub_key(dh_key_accessor::get_impl(public_key)),
```
dutow: Can't we extract most of the code into a common function instead of duplicating it? The only difference between them seems to be using `DH_get0_pub_key` here.

extra/opensslpp/src/opensslpp/rsa_encrypt_decrypt_operations.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+  if (!key.is_private())
+    throw core_error{"RSA key does not have private components"};
+
+  if (input.size() != key.get_size_in_bytes())
+    throw core_error{
+        "decryption block size is not the same as RSA key length in bytes"};
+
+  // TODO: use c++17 non-const std::string::data() member here
+  using buffer_type = std::vector<unsigned char>;
```
dutow, line 124: This using and TODO comment could be in a common include, it's used at many places.

extra/opensslpp/src/opensslpp/rsa_key.cpp
```diff
@@ -0,0 +1,195 @@
...
+  assert(!is_empty());
+  return RSA_bits(rsa_key_accessor::get_impl(*this));
+}
+
+std::size_t rsa_key::get_size_in_bytes() const noexcept {
+  assert(!is_empty());
+  return RSA_size(rsa_key_accessor::get_impl(*this));
+}
+
+std::size_t rsa_key::get_max_block_size_in_bytes(
```
dutow, line 100: Similarly to the other enum class, this could also use the native values and a static_cast.

extra/opensslpp/src/opensslpp/rsa_key.cpp
```diff
@@ -0,0 +1,195 @@
...
+                                     nullptr, nullptr));
+  if (res.is_empty())
+    core_error::raise_with_error_string(
+        "cannot import RSA key from PEM PUBLIC KEY");
+
+  return res;
+}
+
+std::ostream &operator<<(std::ostream &os, const rsa_key &obj) {
+  assert(!obj.is_empty());
```
dutow, line 211: Are you sure an assert is enough here? The export functions also only do an assert, and the key can be empty. (And some other operators have the same issue)

components/encryption_udf/encryption_udf_component.cc
```diff
@@ -0,0 +1,796 @@
...
+  std::size_t max;
+  const char *var_name;
+  const char *var_description;
+};
+
+constexpr std::array<threshold_definition, number_of_thresholds> thresholds = {
+    threshold_definition{
+        1024U, 16384U, "rsa_bits_threshold",
+        "Maximum RSA key length in bits for create_asymmetric_priv_key()"},
+    // DSA max key length must be <= OPENSSL_DSA_MAX_MODULUS_BITS (10000)
```
dutow: This comment could be a `static_assert` with a message.

components/encryption_udf/encryption_udf_component.cc
```diff
@@ -0,0 +1,796 @@
...
+  mysqlpp::udf_result_t<STRING_RESULT> calculate(
+      const mysqlpp::udf_context &args);
+};
+
+mysqlpp::udf_result_t<STRING_RESULT> create_asymmetric_priv_key_impl::calculate(
+    const mysqlpp::udf_context &ctx) {
+  auto algorithm = ctx.get_arg<STRING_RESULT>(0);
+  if (algorithm.data() == nullptr)
+    throw std::invalid_argument("Algorithm cannot be NULL");
+  if (algorithm != "RSA" && algorithm != "DSA" && algorithm != "DH")
```
dutow: Shouldn't these be case insensitive?

components/encryption_udf/encryption_udf_component.cc
```diff
@@ -0,0 +1,796 @@
...
+        throw std::invalid_argument("Invalid RSA key length specified");
+      auto key = opensslpp::rsa_key::generate(length);
+      pem = opensslpp::rsa_key::export_private_pem(key);
+    } else if (algorithm == "DSA") {
+      if (!check_if_bits_in_range(length, threshold_index_type::dsa))
+        throw std::invalid_argument("Invalid DSA key length specified");
+      auto key = opensslpp::dsa_key::generate_parameters(length);
+      key.promote_to_key();
+      pem = opensslpp::dsa_key::export_private_pem(key);
+    }
```
dutow, line 249: DH algorithm doesn't need threshold checks?

percona-ysorokin:
Jenkins build link (--big-test --repeat=2 --do-test=percona_encryption_udf)
https://ps80.cd.percona.com/job/percona-server-8.0-param/1201/

extra/opensslpp/include/opensslpp/big_number_fwd.hpp
percona-ysorokin: From my experience, the approach with separate `_fwd.hpp` files is more flexible. In more generic cases when we have  not just `class foo;` but some smart pointers types defined on this class `foo` (say, `using foo_ptr=std::shared_ptr<foo>`), the `_fwd.hpp` file also needs to include `#include <memory>`. At the same time `class bar;` may not have smart pointers defined for it. 
Basically, when one of my implementation `.cpp` files needs just `class bar`, including `all_fwd.hpp` will result in including unnecessary header `<memory>`
Moreover, my rule of thumb is:
when in doubt of how this should be done look how it is already done in `boost`.

extra/opensslpp/include/opensslpp/core_error.hpp
percona-ysorokin: This was my initial idea as well, but it turned out that for most cases providing OpenSSL error details to the user is very confusing - they are too low level, so I decided to fall back to more simple error messages in most cases.
But this can definitely be revised if you think some case may benefit from more info.
Probably not 100% true, but I do `raise_with_error_string()` mostly for PEM read/write functions - in this case OpenSSL indeed gives more details on what went wrong.

extra/opensslpp/include/opensslpp/dh_key.hpp
percona-ysorokin: Unfortunately, no.
As one of the goals for the library public interface was to never include any OpenSSL headers,
we cannot include `dh.h` in this header to get the `DH_GENERATOR_2` constant.
Therefore the definition is hidden in the `.cpp` file.

extra/opensslpp/include/opensslpp/rsa_key.hpp
percona-ysorokin: likewise

extra/opensslpp/include_private/opensslpp/core_allocated_buffer.hpp
percona-ysorokin: Most probably you are right. In the initial stages of library evolution it seemed like this helper class will be used in a lot of places. It turned out that now iot is used only in `big_number`.

extra/opensslpp/include_private/opensslpp/digest_context_accessor.hpp
percona-ysorokin: Again, I disagree here. More granularity is better in my opinion.

extra/opensslpp/include_private/opensslpp/rsa_padding_conversions.hpp
percona-ysorokin: Again, this is a drawback of the `public headers never include openssl headers` rule.
In the public header file we cannot write
enum class rsa_padding {
  no = RSA_NO_PADDING,
  pkcs1 = RSA_PKCS1_PADDING
};
because RSA_NO_PADDING and RSA_PKCS1_PADDING are not available there.

extra/opensslpp/src/opensslpp/dh_compute_operations.cpp
percona-ysorokin: Agree here. Copy-pasting is evil :)

extra/opensslpp/src/opensslpp/rsa_encrypt_decrypt_operations.cpp
percona-ysorokin: This is a part of my c++17-fication internal todo list:
* avoid extra data copying (this item)
* use std::string_vew in interfaces
However, I decided to pay this technical debt in the next update and release as soon as possible.

extra/opensslpp/src/opensslpp/rsa_key.cpp
percona-ysorokin: The same answer as for padding.

extra/opensslpp/src/opensslpp/rsa_key.cpp
percona-ysorokin: Not sure I am completely following you here but my general principles regarding empty objects were
* default constructor always creates an empty object
* empty objects can only be used as targets for assignment
* you are not allowed to call any method (except is_empty()) on an empty object - we do not throw exceptions in such cases, we use asserts

components/encryption_udf/encryption_udf_component.cc
percona-ysorokin: Probably, it's better to define 9984 as
static constexpr std::size_t aligned_max_dsa_bits = OPENSSL_DSA_MAX_MODULUS_BITS / 64 *64;

components/encryption_udf/encryption_udf_component.cc
percona-ysorokin: Agree, even better, I think we should define a conversion function from a string to an enum
enum class algorithm_type { unknowm rsa, dsa, dh }
algorithm_type get_algorith_by_name(const std::string& name) noexcept {...}

components/encryption_udf/encryption_udf_component.cc
percona-ysorokin: DH threshold checks are performed in the `create_dh_parameters()` - this are the specifics of the Diffie-Hellman algorithm. The most computationally-heavy operations is not key generation but parameters generation.

extra/opensslpp/src/opensslpp/rsa_key.cpp
dutow: I'm not sure why we have the ostream operators, they aren't used in the code right now.

I assumed that they are for debugging or maybe logging in the future.

And in that case, accidentally printing out an empty object when debugging an issue / logging something out could dereference null pointers and crash. And that's something I wouldn't expect.

extra/opensslpp/include/opensslpp/big_number_fwd.hpp
dutow: I also prefer organizing the source code into smaller logical files, but my main concern here is that this is quite different compared to the mysql coding standards. Even our own files/projects (in pxc/pxb, and other repos too, where we have even more percona code) follow the usual "put everything into a few big files" approach. Compared to that, this library now even has "single using/declaration" header files, and one function definition source files.

So does this mean that we'll change our approach for this, and now we might use this style at other more self-contained parts of the code? For the coding style, so far the basic rule was "look how it is done in `mysql`".

And then we could face another issue in the long run: while these small files are logical and nice, they aren't compilation time friendly. Without precompiled headers and unity builds (or modules) they could add up to a measurable time increase in a few years if we keep using them more and more. (at least CMake now supports both, but they require 3.16)

Again, this is more of a question than a change request. I also prefer smaller files, and if go this way, I'll use them too in the future.

extra/opensslpp/include/opensslpp/core_error.hpp
dutow: > to the user is very confusing

`opensslpp` looks like a generic C++ OpenSSL wrapper library used in other projects, such as the UDF component, or other components/things in the future. And it's strange for a wrapper library to hide error messages from the underlying library.

On the other hand, I didn't check how those error messages would look like, I only checked the documentation of a few OpenSSL functions where these exceptions are generated, and most said that for more details about the error, use `err_get_error`.

Or these error messages might be confusing to users, but maybe they could be helpful for the next developer working on some code with `opensslpp`?

In that case, maybe `core_error` could store the real OpenSSL error in another data member with a separate accessor. The using code could decide if it wanted to display it to the user or not, and while debugging, it would be easily available to the developer?

dutow:
> As one of the goals for the library public interface was to never include any OpenSSL headers

This just reminded me that the library could use at least a short readme or some sort of minimal documentation explaining what it is in a few sentences.

E.g. while anybody can figure out that it is a C++ OpenSSL wrapper, hiding the C interface is not always included in the goals of wrapper libraries.

percona-ysorokin:
Proper Jenkins build link (after fixing `WITH_ENCRYPTION_UDF` issue)
(--big-test --repeat=2 --do-test=percona_encryption_udf)
https://ps80.cd.percona.com/job/percona-server-8.0-param/1204/

plugin/encryption_udf/encryption_udf.cc
percona-ysorokin: Fixed

extra/opensslpp/include/opensslpp/core_error.hpp
percona-ysorokin: @dutow My judging is based on the following example.
If, for instance, we have an error in the opensslpp::rsa_key copy constructor (most probably one of the memory allocation functions failed during RSA components (big numbers, BNs) cloning), we do not need nothing more than "Cannot duplicate rsa_key". Extracting SSL error message won't give users any useful info in this case - the error would look like "Failed to allocate memory for BN_copy()" (users won't even know what BN in this case is).
On the other hand from my experience almost everything PEM format-related has pretty good/useful error diagnostics

extra/opensslpp/include/opensslpp/big_number_fwd.hpp
percona-ysorokin: You are right regarding this being in contradiction with MySQL server standards. But those standards are 20 years old :)
So, for independent components like this one I would definitely prefer something more up-to-date.
BTW, I totally disagree regarding compilation times - the main reason of having modularized small files and forward declarations is to REDUCE build times. The principle to include only what you need to use instead of `#include <library/everything.hpp>` has dramatic impact on build time. Moreover, having interfaces of one class depending only on forward declarations of others helps a lot with it as less files need to be recompiled when something in implementation is changed.

extra/opensslpp/include_private/opensslpp/core_allocated_buffer.hpp
percona-ysorokin: `core_allocate_buffer` completely removed

extra/opensslpp/src/opensslpp/dh_compute_operations.cpp
percona-ysorokin: Logic extracted into common `compute_dh_key_internal()` method

components/encryption_udf/encryption_udf_component.cc
percona-ysorokin: Introduced `get_algorithm_id_by_label()` and `get_and_validate_algorithm_id_by_label()` functions

dutow (approved):

extra/opensslpp/include/opensslpp/operation_cancelled_error.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <opensslpp/operation_cancelled_error_fwd.hpp>
+
+#include <opensslpp/core_error.hpp>
+
+namespace opensslpp {
+
+class operation_cancelled_error : public core_error {
+ public:
+  operation_cancelled_error(const char *message) : core_error{message} {}
```
dutow: This could be simply `using core_error::core_error`

dutow (approved):

extra/opensslpp/include/opensslpp/operation_cancelled_error.hpp
percona-ysorokin: fixed

percona-ysorokin:
New Jenkins build link (after rebase on 8.0.28-19)
https://ps80.cd.percona.com/job/percona-server-8.0-param/1224/ (--do-test=percona_encryption_udf --repeat=2 --big-test)

PR #4555, opened 2021-12-07
Generate ZenFS version upon building percona-server

MaisenbacherD:
Description: For the RocksDB plugin ZenFS a semi automatic versioning that is
triggered by the make build environment was introduced with
https://github.com/westerndigitalcorporation/zenfs/pull/94.
In order to generate the ZenFS version information upon building
percona-server, the CMake build environment is ajusted to call the script
that is responsible for updateing the ZenFS version.

Signed-off-by: Dennis Maisenbacher <dennis.maisenbacher@wdc.com>


MaisenbacherD:
Forwarded the ZenFS submodule so that it points to master and made sure that building stops if the ZenFS version could not be generated.

percona-ysorokin:
@MaisenbacherD I encountered an interesting issue with the tag identification mechanism.
I am trying to get the latest known tag from the `zenfs` git repository using `git describe`
`git lg -5` gives me
```
99cb288 (HEAD, tag: v1.0.0) fs, util: Introducing semi automatic versioning for ZenFS.
f93bd5c ci: Always remove docker images
79244a7 ci: Always collect results
234531e ci: zbdbench workflow
fee6ebe fs: Add snapshot API
```
the last known tag here is `v1.0.0` (current head)

and at the same time `git describe` returns
```
v0.1.0-78-g99cb288
```
meaning `78` commits ahead of `v0.1.0` whereas what I expect is just `v1.0.0`

Not 100% sure here but I believe that this is connected to the fact that `v1.0.0` in terms of `git` is not an `annotated` tag, just a `lightweight` one.

Running `git describe --tags` solves the issue and returns proper
```
v1.0.0
```
however this is not what `generate-version.sh` does.

I see 3 solutions to this problem now:
1. You somehow update the `zenfs` repo to make `v1.0.0` an annotated tag (not sure if this can be done without force pushing)
2. You push a new commit to `zenfs` adding `--tags` to `git describe` call inside `generate-version.sh` and I advance submodule commit pointer even further. However, in this case the release will include a few new commits that were not tested (like "fs:Add support for getting ZBD status")
3. I rewrite the logic of zenfs version file generation using CMake commands in RocksDB's CMakeLists file and ignore (never call) `generate-version.sh`.

metaspace:
@percona-ysorokin let me see if I can change the tag without too much trouble.

metaspace:
@percona-ysorokin I was able to push an annotated version of the tag. Let me know if that solves the problem :)

percona-ysorokin:
Thanks @metaspace @MaisenbacherD 
The issue with the tag in the `zenfs` repository seems to be fixes now.
Everything works as expected.
Just be aware that after force tag update regular `git submodule update` did not do anything - I had to use `git fetch origin --force` for the submodule manually to update the local repo.

percona-ysorokin:
Closing as this PR has been cherry-picked to the release-8.0.26-17 branch
(commit 1338ba3)

PR #4556, opened 2021-12-07
PS-7993: Use dot to separate MTR test combination name in JUnit report (5.7)

oleksandr-kachan:
https://jira.percona.com/browse/PS-7993

Using underscore as a separator between test name and test combination
name makes it impossible for any script used to parse results file to
find out if there is a test combination in test name. Use dot as
a separator in this case to improve this.


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-pipeline/25032/

percona-ysorokin (approved):
LGTM

PR #4562, opened 2021-12-08
PS-7995: Update macOS version used for Azure builds to latest (5.7)

oleksandr-kachan:
https://jira.percona.com/browse/PS-7995

The macOS-10.14 environment is going to be deprecated soon.
Use macOS-10.15 instead.


inikep (approved):
LGTM

PR #4474, opened 2021-09-27
PS-7792: On replica Master_User should be blank if not storing in slave_master_info

venkatesh-prasad-v:
PS-7792: On replica Master_User should be blank if not storing in slave_master_info
BUG#27357189: MASTER_USER BECOMES 'TEST' IF IT IS EMPTY WHEN READ FROM REPO

This commit backports the bugfix for Oracle Bug#27357189 to PS-5.7.

When the user for the master connection is empty and there is a read
from the repository information the user becomes 'test'. This happens
as the default value when there is no information is 'test'.

The default value is now empty as well.

ReviewBoard: 24194
(cherry picked from commit 304e920403cecaab9a45834d041bc2ea5a0d2a57)

Testing Done
---
Jenkins: https://ps57.cd.percona.com/job/percona-server-5.7-param/501/console (In Progress)
Failing tests:

PR #4564, opened 2021-12-13
PS-7792: On replica Master_User should be blank if not storing in slave_master_info

venkatesh-prasad-v:
This is a Null Merge of https://github.com/percona/percona-server/pull/4474, as the bug is already fixed in 8.0.

PS-7792: On replica Master_User should be blank if not storing in slave_master_info
BUG#27357189: MASTER_USER BECOMES 'TEST' IF IT IS EMPTY WHEN READ FROM REPO

This commit backports the bugfix for Oracle Bug#27357189 to PS-5.7.

When the user for the master connection is empty and there is a read
from the repository information the user becomes 'test'. This happens
as the default value when there is no information is 'test'.

The default value is now empty as well.


satya-bodapati (approved):
Null merges can be self approved :)

PR #4566, opened 2021-12-17
PS-7891 - RocksDB table with virtual column crashes the server with update query

avodaniel:


storage/rocksdb/rdb_datadic.cc
```diff
@@ -185,6 +185,18 @@ int Rdb_convert_to_record_key_decoder::skip(
   if (fpi->m_skip_func == &Rdb_key_def::skip_variable_space_pad &&
       !fpi->m_unpack_info_stores_value && !covered_bitmap_format_enabled) {
     unp_reader->read(fpi->m_unpack_info_uses_two_bytes ? 2 : 1);
+  } else if (fpi->m_covered == Rdb_key_def::KEY_MAY_BE_COVERED &&
```
kamil-holubicki, line 188: I just started wondering if these 2 cases are not "hacks" that should be handled by proper `m_skip_func` function.
`m_skip_func` is used in several other places. Don't we have the same problem of 'partial skip' there?

avodaniel: @kamil-holubicki  It would be a bit tricky to re-implement skip function, because it does not take `unp_reader()` as an argument. It looks that other use of `m_skip_function()` are safe because they do not combine pack and unpack reader or they also skip unpacked bytes (`unpack_unknown_varlength()`).

kamil-holubicki: OK, agree.

mysql-test/suite/rocksdb/t/issue7891.test
```diff
@@ -0,0 +1,13 @@
+--source include/have_rocksdb.inc
```
kamil-holubicki: It looks like tese cases are 1:1 taken from pstress run.
Please:
1. Fix formatting (unneeded whitespaces, etc)
2. table/column/index naming
3. Are tables structures minimal? I've found this case to be minimal:
```
CREATE TABLE tt_5 (
	t0 TEXT,
	f3 FLOAT,
	g4 VARCHAR(4),
	INDEX tt_5i3(f3, t0(2), g4 ASC))
	ENGINE=RocksDB ;
INSERT INTO tt_5  ( t0 ,f3 ,g4 ) VALUES( 'WID5rpEI0cHsvwPtW77IEGYMAhVHBJKTVwMgUzaluvLR1se4uo3hZblvSW0I7xvcq9qa8H9ThW',  0.00, 'abcd' ) ;
UPDATE tt_5 SET g4 = 'ABCD' WHERE f3 = 0.00 ;
DROP TABLE tt_5;
```
Maybe we can narrow it down further?
4. Please change the name of the test to be meaningful

kamil-holubicki (changes requested):
The fix itself seems to be the way to go however, I've left a few comments asking for adjustments needed to comply with our coding standards.
Additionally, please write a proper commit message (incl Jira ticket link) containing a detailed explanation of the problem in the form problem/cause/solution.

kamil-holubicki:
https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/39/

avodaniel:
> https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/39/

@kamil-holubicki  I tried this issue with the case-insensitive filesystem (JFS) and everything looks to work.

kamil-holubicki (approved):
LGTM. Please squash commits.

dutow (changes requested):
Please run the entire rocksdb suite, this patch causes tests to fail. Including the already existing virtual_basic testcase. (on the currrent 8.0 branch)

PR #4568, opened 2021-12-17
PS-7856: update on partition tables crashes the server

voodam:
`write_set` wasn't used here, probably this is mistake. Replacing with `read_set` solves the problem, but is it correct that we skip `set_auto_increment_if_higher` when read_set is reseted? Or maybe we need to mark autoincrement column used in some appropriate place. I need an advice here.

Also bug is reproduced in 8.0, but binlog should be disabled, otherwise bitmap is set and no assertion error is produced. I created a PR for that branch too.

### DIFF
--- a/sql/partitioning/partition_handler.cc
+++ b/sql/partitioning/partition_handler.cc
@@ -669,12 +669,12 @@ int Partition_helper::ph_update_row(const uchar *old_data, uchar *new_data,
     index)
     mysql_update does not set table->next_number_field, so we use
     table->found_next_number_field instead.
-    Also checking that the field is marked in the write set.
+    Also checking that the field is marked in the read set.
   */
   if (m_table->found_next_number_field &&
       new_data == m_table->record[0] &&
       !m_table->s->next_number_keypart &&
-      bitmap_is_set(m_table->write_set,
+      bitmap_is_set(m_table->read_set,
                     m_table->found_next_number_field->field_index))
   {
     set_auto_increment_if_higher();
--- a/mysql-test/suite/innodb/r/partition_autoinc.result
+++ b/mysql-test/suite/innodb/r/partition_autoinc.result
@@ -1,0 +1,5 @@
+# PS-7856: update on partition tables crashes the server
+CREATE TABLE t1(ip_col INT, i1 INT AUTO_INCREMENT, INDEX tt_2_pi1(i1)) PARTITION BY HASH (ip_col);
+INSERT INTO t1(ip_col) VALUES(1);
+UPDATE t1 SET i1 = 1 WHERE ip_col = 1;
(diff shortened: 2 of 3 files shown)


sql/partitioning/partition_handler.cc
```diff
@@ -669,12 +669,12 @@ int Partition_helper::ph_update_row(const uchar *old_data, uchar *new_data,
...
     mysql_update does not set table->next_number_field, so we use
     table->found_next_number_field instead.
-    Also checking that the field is marked in the write set.
+    Also checking that the field is marked in the read set.
   */
   if (m_table->found_next_number_field &&
       new_data == m_table->record[0] &&
       !m_table->s->next_number_keypart &&
-      bitmap_is_set(m_table->write_set,
+      bitmap_is_set(m_table->read_set,
```
kamil-holubicki, line 677: If we go this way, we will skip `set_auto_increment_if_higher()` call, which is not OK.
The problem is that when binlog is disabled, we are skipping call to `TABLE::mark_columns_per_binlog_row_image()` which is supposed to set table's write/read sets bits. If we skip, then we hit the assert.
On the other hand the assert ASSERT_COLUMN_MARKED_FOR_READ inside `Field_long::val_int()` seems to be too restrictive in this case (that's why commit 74cc73d4619c94ceb1b93725b1bba7b802f2290d introduced `dbug_tmp_use_all_columns / dbug_tmp_restore_column_map` 

My proposal for the fix is:
```
  if (m_table->found_next_number_field && new_data == m_table->record[0] &&
      !m_table->s->next_number_keypart &&
      bitmap_is_set(m_table->write_set,
                    m_table->found_next_number_field->field_index())) {
    my_bitmap_map *old_map = dbug_tmp_use_all_columns(m_table, m_table->read_set);
    set_auto_increment_if_higher();
    dbug_tmp_restore_column_map(m_table->read_set, old_map);
  }
```

kamil-holubicki (changes requested):
This is the fix for 5.7 branch. We need to merge it to 8.0 branch. We should use GCA process here.

voodam:
> This is the fix for 5.7 branch. We need to merge it to 8.0 branch. We should use GCA process here.

These commits slightly differ:
https://github.com/percona/percona-server/pull/4567/files
https://github.com/percona/percona-server/pull/4568/files

For `8.0` it's need to disable binlog in `mysql-test/suite/innodb/t/partition_autoinc-master.opt` to reproduce error. Is it possible to handle this case via GCA? Also result and test files was created newly in `5.7`, maybe it also can be a problem, I'm not sure.

voodam:
New PR in https://github.com/percona/percona-server/pull/4589 to comply GCA process.

PR #4570, opened 2021-12-20
PS-7046: Add Query_tables and Query_digest fields into slow query log

oleksandr-kachan:
Extend slow query log records with additional query info. Two fields are added:
- Query_tables which contains a comma separated list of tables used by a query
- Query_digest field containing query digest.
Both fields are disabled by default. It is needed to set log-slow-verbosity=query_info
to enable them.


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1130/

mysql-test/t/percona_slowlog_query_tables.test
```diff
@@ -0,0 +1,42 @@
...
+
+--source include/log_start.inc
+
+SELECT * FROM t1, t2;
+
+--source include/log_stop.inc
+
+--let grep_pattern = ^# Query_tables: t1,t2
+--let log_expected_matches = 1
+--source include/log_grep.inc
```
satya-bodapati: does it work, if the query has inner queries, joins etc? also CTEs

satya-bodapati (changes requested):

mysql-test/t/percona_slowlog_query_tables.test
oleksandr-kachan: I've added some queries into the test. Please pay attention to CTE result, tables list includes CTE name as well.

sql/sys_vars.cc
```diff
@@ -6112,8 +6119,8 @@ static Sys_var_set Sys_log_slow_verbosity(
     "log_slow_verbosity",
     "Choose how verbose the messages to your slow log will be. "
     "Multiple flags allowed in a comma-separated string. [microtime, "
-    "query_plan, innodb, profiling, profiling_use_getrusage, minimal, "
-    "standard, full]",
+    "query_plan, innodb, profiling, profiling_use_getrusage, query_info, "
```
satya-bodapati, line 6122: can you please rename query_info to query_tables?

satya-bodapati (changes requested):

sql/sys_vars.cc
oleksandr-kachan: I'm planning to use this flag for a few different fields in slow log so used a generic name for it.

satya-bodapati (changes requested):
patch looks good. Please review test failures and fix them. One of them is related your fix

oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1133/

oleksandr-kachan:
> patch looks good. Please review test failures and fix them. One of them is related your fix

Done. I've added one more commit which adds one more related field. Waiting for new test results.

satya-bodapati:
@oleksandr-kachan some option or help tests are failing because of the new option query_info, please check

oleksandr-kachan:
> percona-server-8.0-param/1130/

@satya-bodapati, which one are you checking? percona-server-8.0-param/1130/ is the old one, I'm waiting for new test results now (there were issues with aws so it takes some time).

oleksandr-kachan:
@satya-bodapati, here are fresh test results. Please note that not all test suits are run here.

https://ps80.cd.percona.com/job/percona-server-8.0-param/1138/
https://ps80.cd.percona.com/job/percona-server-8.0-param/1140/

satya-bodapati (approved):

PR #4576, opened 2021-12-28
PS-7999: FEDERATED engine not reconnecting on wait_timeout exceeded

voodam:
https://jira.percona.com/browse/PS-7999

Original commit caused error: https://github.com/percona/percona-server/commit/14508bbb179
Error is caused by sending error message not to the client (where MYSQL_SERVER is undefined), but in case of Federated to the server. In original commit !defined(MYSQL_SERVER) check was used here. Probably author of the original commit assumed that only client can be target of this error message, but for Federated this is not true.


kamil-holubicki:
I didn't trace it in detail but can imagine that Federated SE uses net_read_packet_header() function (probably indirectly). Something had to change there as well because the server part still works the same as before the change. Only the part added is the one within `#if !defined(MYSQL_SERVER)`. As Federated defines MYSQL_SERVER, this part is out, but the caller probably expects the new behavior of  net_read_packet_header() and that is why we fail.

It looks that the change is OK, but could you give more insights why after the mentioned change we see the error? It would be good to describe this in the commit message as well.

voodam:
> I didn't trace it in detail but can imagine that Federated SE uses net_read_packet_header() function (probably indirectly). Something had to change there as well because the server part still works the same as before the change. Only the part added is the one within `#if !defined(MYSQL_SERVER)`. As Federated defines MYSQL_SERVER, this part is out, but the caller probably expects the new behavior of net_read_packet_header() and that is why we fail.

IIUIC, this part is here: https://github.com/percona/percona-server/commit/14508bbb179#diff-281245ca3dd9f832d0c1cda0cf51f2410801734265ae35e7ae5120d84bd3aeb0L1371

> It looks that the change is OK, but could you give more insights why after the mentioned change we see the error? It would be good to describe this in the commit message as well.

We see error "Got packets out of order" because packages are really out of order (`pkt_nr != (uchar)net->pkt_nr`: https://github.com/percona/percona-server/commit/14508bbb179#diff-281245ca3dd9f832d0c1cda0cf51f2410801734265ae35e7ae5120d84bd3aeb0R1454), but this state become not errorneous intentionally as part of feature in `14508bbb179` commit. Also two new errors: `NET_ERROR_SOCKET_NOT_WRITABLE` and `NET_ERROR_SOCKET_NOT_READABLE` was introduced just to send this error message.

Or if question is: why we see "The client was disconnected by the server because of inactivity..." error in test - because we set `wait_timeout` equal 4 and then sleep for 5 seconds.

I changed commit message, hope it's better now.

kamil-holubicki:
Thanks for the explanation. Let me rephrase my question:
1. Before 14508bb `net_read_packet_header()` returned `true`
2. After 14508bb `net_read_packet_header()` returns `false` and sets NET_ERROR_SOCKET_UNUSABLE and ER_NET_PACKETS_OUT_OF_ORDER

Where is the change in Federated that expects the new behavior (client side expecting (2))?

voodam:
I reexaminated all the test routune in gdb and want to give detailed explanation. I hope it should make all clear.
I added breakpoints in all code added in https://github.com/percona/percona-server/commit/14508bbb1790697c28659dd051fbc855cd3b5da9, so all not mentioned below code is not important (couple of places) or wasn't affected (most places). All reconnect stuff wasn't affected, for example.

1. Make all preparations and connect from gdb to n1 and n2:
```
n2 -e "set global wait_timeout=5"
n2 -e "create table test.f (id int primary key); insert into test.f values(1)"

n1 -e "create server test foreign data wrapper test1 options (user 'msandbox', password 'msandbox', host '127.0.0.1', port 26729, database 'test');"
n1 -e "create table test.f (id int primary key) engine=federated connection='test';"
```

2. Run queries:
```
n1 -e "select * from test.f; select sleep(10); select * from test.f;"
```

3. After timeout is reached (5 seconds) n2 breaks in https://github.com/percona/percona-server/commit/14508bbb1790697c28659dd051fbc855cd3b5da9#diff-281245ca3dd9f832d0c1cda0cf51f2410801734265ae35e7ae5120d84bd3aeb0L1374, so socket is `NET_ERROR_SOCKET_NOT_READABLE` now.

4. Next in this call we return back to https://github.com/percona/percona-server/commit/14508bbb1790697c28659dd051fbc855cd3b5da9#diff-33a630b3f254d9adcc1b2e97b6817e5c04da320d4523eadd1776fed78f920262R1413 and return `-1` from `Protocol_classic::read_packet`.

5. Next we return to `do_command` and `rc == -1`: https://github.com/percona/percona-server/blob/8.0/sql/sql_parse.cc#L1348. So we go to `send_statement_status`: https://github.com/percona/percona-server/blob/8.0/sql/sql_parse.cc#L1366.

6. Next we go to `net_write_packet`: https://github.com/percona/percona-server/blob/8.0/sql-common/net_serv.cc#L1316, trying to send package with error about timeout.

7. This package from n2 will be catched on n1 node: https://github.com/percona/percona-server/commit/14508bbb1790697c28659dd051fbc855cd3b5da9#diff-281245ca3dd9f832d0c1cda0cf51f2410801734265ae35e7ae5120d84bd3aeb0R1457. I didn't explored in detail why this packet has out of order number, probably because we didn't close the socket and maked it `NET_ERROR_SOCKET_NOT_READABLE` instead. Can explore it better if necessary.

8. Next, about returning with `false` from `net_read_raw_loop`. We return from all callers of `net_read_raw_loop` with normal, not-errorneous code until being here: https://github.com/percona/percona-server/blob/8.0/sql-common/client.cc#L1203. So it seems that returning `false` from `net_read_raw_loop` is not most legal way to handle error. From this point we return error code and handle packet as errorneous, show error and so on.

Federated code works as usual here calling common query functions or something.

I have some questions after that:
1. Is it required to find better way to return error from `net_read_raw_loop`? Maybe return true or something else.
2. I found suspicious thing here: https://github.com/percona/percona-server/commit/14508bbb1790697c28659dd051fbc855cd3b5da9#diff-33a630b3f254d9adcc1b2e97b6817e5c04da320d4523eadd1776fed78f920262R1413. In all other places `NET_ERROR_SOCKET_UNUSABLE` was replaced from `2` and only here from `3`. Should I to investigate it better?

kamil-holubicki:
TL;DR
I looked closer at the problem and in my opinion it doesn't work properly now.
8.0.23 - even if federated client-server times out, automatic reconnection invisible to the user happens under the hood. So no error is reported to the user. From his point of view he is able to query slave tables without any interruption

8.0.26 original version - When the connection times out the user gets `ERROR 1156 (08S01): Got packets out of order` on 1st query. Next query is OK if done before connection timeout

8.0.26 with the fix - When the connection times out the user gets `ERROR HY000: : 4031 : The client was disconnected by the server because of inactivity. See wait_timeout and interactive_timeout for configuring this behavior.` on 1st query. Next query is OK if done before connection timeout

The reconnection should be invisible to the user (client of the master)


Long version:

Please check the behavior of 8.0.23. Because of mentioned automatic reconnection it seems that `SET @@GLOBAL.WAIT_TIMEOUT= 4;` takes no effect. (but this is only the impression of "black-box" observer)
I have hacked it in the following way:
1. Change mysql_com.h -> #define NET_WAIT_TIMEOUT 6
Connect to n1. After 6 seconds connection to n1 times out. Make some activity just to avoid its timeout (show databases; or something) and after 10 seconds query federated table again. It still works without any problem.

2. Change ha_federated::real_connect() -> mysql->reconnect = false;
Do similar test as above. This time federated connection is not reconnected automatically and we get the error `ERROR 1158 (08S01): Got an error reading communication packets` and then `ERROR 1430 (HY000): : 2006 : MySQL server has gone away`


Back to 8.0.26. What I asked for is the callstack on n1 (federated client side):
```
net_read_packet_header(NET * net) (percona-server/sql-common/net_serv.cc:1469)
net_read_packet(NET * net, size_t * complen) (percona-server/sql-common/net_serv.cc:2059)
net_read_uncompressed_packet(NET * net, size_t & len) (percona-server/sql-common/net_serv.cc:2175)
my_net_read(NET * net) (percona-server/sql-common/net_serv.cc:2251)
cli_safe_read_with_ok(MYSQL * mysql, bool parse_ok, bool * is_data_packet) (percona-server/sql-common/client.cc:1142)
cli_safe_read(MYSQL * mysql, bool * is_data_packet) (percona-server/sql-common/client.cc:1267)
cli_read_query_result(MYSQL * mysql) (percona-server/sql-common/client.cc:7285)
mysql_real_query(MYSQL * mysql, const char * query, ulong length) (percona-server/sql-common/client.cc:7657)
ha_federated::real_query(ha_federated * const this, const char * query, size_t length) (percona-server/storage/federated/ha_federated.cc:3000)
ha_federated::info(ha_federated * const this, uint flag) (percona-server/storage/federated/ha_federated.cc:2642)
TABLE_LIST::fetch_number_of_rows(TABLE_LIST * const this) (percona-server/sql/table.cc:6667)
JOIN::init_planner_arrays(JOIN * const this) (percona-server/sql/sql_optimizer.cc:5287)
JOIN::make_join_plan(JOIN * const this) (percona-server/sql/sql_optimizer.cc:5114)
JOIN::optimize(JOIN * const this) (percona-server/sql/sql_optimizer.cc:608)
Query_block::optimize(Query_block * const this, THD * thd) (percona-server/sql/sql_select.cc:1807)
Query_expression::optimize(Query_expression * const this, THD * thd, TABLE * materialize_destination, bool create_iterators) (percona-server/sql/sql_union.cc:679)
Sql_cmd_dml::execute_inner(Sql_cmd_dml * const this, THD * thd) (percona-server/sql/sql_select.cc:759)
Sql_cmd_dml::execute(Sql_cmd_dml * const this, THD * thd) (percona-server/sql/sql_select.cc:574)
mysql_execute_command(THD * thd, bool first_level) (percona-server/sql/sql_parse.cc:4720)
dispatch_sql_command(THD * thd, Parser_state * parser_state, bool update_userstat) (percona-server/sql/sql_parse.cc:5325)
```

We should answer the question why the federated client does not autoreconnect.

kamil-holubicki (changes requested):
In my opinion it does not solve the problem

voodam:
I added fix that restores reconnect for Federated. Sorry, didn't mentioned this requirement, though it was right in title.

Commits will be squashed to single one and the last commit message will remain only.

voodam:
It seems that it's required some significant modification, not just patch with reconnect in suitable place.

I'll describe in detail.

[14508bbb179](https://github.com/percona/percona-server/commit/14508bbb179) introduces new logic: on timeout server doesn't break connection, but tries to send error first and after this socket stays unusable. And what happens here: we send error from the server (node 2) to the client (node 1), then on client node we check socket via `vio_is_connected`. Inside this function we trying to peek data here: https://github.com/percona/percona-server/blob/8.0/vio/viosocket.cc#L1481. And client sees the data - this is error message from server, so client decides that connection is available. And so it doesn't reconnect in `cli_advanced_command`.

Error in client occurs further, on `cli_safe_read`, and read commands can not have reconnect logic. We can reconnect after the full real_query anyway. `ha_federated::real_query` is good place for this (I put code in `ha_federated::info` first, but this was wrong, I'll update it). But maybe we need rework `vio_is_connected` instead? Or rework 14508bbb179 to it's purpose will be achieved in some other way, without sending error and then closing socket.

sql/protocol_classic.cc
```diff
@@ -1410,6 +1410,12 @@ int Protocol_classic::read_packet() {
   }
 
   bad_packet = true;
+  /*
+    XXX: This place seems errorneous: why we keep thread for unusable
```
kamil-holubicki: s/errorneous/erroneous

Please write more descriptive comment. This one is hard to understand without investigating what is the context (we keep thread... which thread? what keep, etc...). Probably describing better what happens to the caller now and what would happen if it was `NET_ERROR_SOCKET_RECOVERABLE` would be understood easier.

storage/federated/ha_federated.cc
```diff
@@ -2990,14 +2991,27 @@ int ha_federated::real_connect() {
...
   if (!query || !length) goto end;
 
-  rc = mysql_real_query(mysql, query, static_cast<ulong>(length));
+  rc = mysql_real_query(mysql, query, len);
+  if (rc) {
+    /*
+      We want to reconnect here because error can occur on reading query
+      result in the case of timeout exceeding. See PS-7999 for details.
+    */
+    if (!mysql || mysql->net.last_errno != CR_SERVER_LOST) goto end;
```
kamil-holubicki: We don't need the same condition as in mysql.cc::mysql_real_query_for_lazy()?

kamil-holubicki (commented):
Looks fine.
Please just 
1. correct the typo
2. answer my question about reconnect condition
3. sqash everything

kamil-holubicki (approved):
lgtm

PR #4578, opened 2021-12-31
PS-7887-2: RocksDB encryption

kamil-holubicki:
This work is based on top of https://github.com/percona/percona-server/pull/4462
1. Rebased to current 8.0 HEAD
2. SST files are immutable. Master Key rotation affects only new files
3. Introduced EncryptionInfoStorage interface with EncryptionInfoPlainFileStorage implementation

NOTE: There is 1 change in rocksdb repository: missing override added. It will be added in upstream in the future, or if not, we can put this class wrapper in myrocks namespace.

White: Parts provided by RocksDB, located in rocksdb namespace
Pink: Parts provided by RocksDB, located in rocksdb namespace that we cannot use (do not meet our requirements)
Blue: Parts provided by Percona, located in myrocks namespace.
RDB-encryption-version2-ClassDiagram


hulyav:
Given there is no intrest on the feature, we are closing the PR

PR #4581, opened 2022-01-04
PS 7989 - Migrate innodb_encrypt_online_alter_logs to 8.0.27

ldonoso:
https://jira.percona.com/browse/PS-7989
https://ps80.cd.percona.com/job/percona-server-8.0-param/1139/

### DIFF
--- a/storage/innobase/ddl/ddl0file-reader.cc
+++ b/storage/innobase/ddl/ddl0file-reader.cc
@@ -37,7 +37,7 @@ this program; if not, write to the Free Software Foundation, Inc.,
 namespace ddl {
 
 [[nodiscard]] inline static auto get_read_len_max(
-    const Write_offsets &offsets) {
+    const Write_offsets &offsets) noexcept {
   ut_a(!offsets.empty());
 
   os_offset_t len_max = 0;
@@ -57,7 +57,7 @@ dberr_t File_reader::prepare() noexcept {
   ut_a(m_ptr == nullptr);
   ut_a(m_mrec == nullptr);
   ut_a(m_buffer_size > 0);
-  ut_a(m_len == 0);
+  ut_a(m_read_len == 0);
 
   if (m_offset == m_size) {
     return DB_END_OF_INDEX;
@@ -109,8 +109,8 @@ dberr_t File_reader::prepare() noexcept {
   }
 
   ut_a(m_size > m_offset);
-  m_len = get_read_len_next();
-  const auto err = ddl::pread(m_fd, m_io_buffer.first, m_len, m_offset,
+  m_read_len = get_read_len_next();
+  const auto err = ddl::pread(m_fd, m_io_buffer.first, m_read_len, m_offset,
                               m_crypt_buffer.first, m_space_id);
 
   if (err != DB_SUCCESS) {
@@ -129,8 +129,8 @@ dberr_t File_reader::seek(os_offset_t offset) noexcept {
 
   m_offset = offset;
 
-  m_len = get_read_len_next();
-  const auto err = ddl::pread(m_fd, m_io_buffer.first, m_len, m_offset,
+  m_read_len = get_read_len_next();
+  const auto err = ddl::pread(m_fd, m_io_buffer.first, m_read_len, m_offset,
                               m_crypt_buffer.first, m_space_id);
 
   m_ptr = m_io_buffer.first;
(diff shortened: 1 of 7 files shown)


storage/innobase/include/ddl0impl-file-reader.h
```diff
@@ -132,12 +136,13 @@ struct File_reader : private ut::Non_copyable {
...
 
+  os_offset_t m_len{};
+
+  byte *get_io_buffer_end() { return m_io_buffer.first + m_len; }
+
   /** Pointer current offset within file buffer. */
   const byte *m_ptr{};
 
-  /** File buffer bounds. */
-  Bounds m_bounds{};
```
percona-ysorokin, line 139: Most probably `using Bounds = std::pair<const byte *, const byte *>;` can be removed as well

storage/innobase/ddl/ddl0file-reader.cc
```diff
@@ -257,4 +281,34 @@ dberr_t File_reader::next() noexcept {
   return DB_SUCCESS;
 }
 
+[[nodiscard]] inline static auto get_next_offset(const Write_offsets &offsets,
+                                                 os_offset_t offset) {
```
percona-ysorokin: `noexcept`

storage/innobase/ddl/ddl0merge.cc
```diff
@@ -158,6 +158,8 @@ struct Merge_file_sort::Output_file : private ut::Non_copyable {
   /** @return number of rows in the output file. */
   [[nodiscard]] uint64_t get_n_rows() const noexcept { return m_n_rows; }
 
+  Write_offsets m_write_offsets;
```
percona-ysorokin, line 161: Any chance we can predict the number of potential writes and preallocate `m_write_offsets`
`m_write_offsets.reserve()` (probably in the `init()` method)?

storage/innobase/ddl/ddl0file-reader.cc
```diff
@@ -36,33 +36,57 @@ this program; if not, write to the Free Software Foundation, Inc.,
 
 namespace ddl {
 
+[[nodiscard]] inline static auto get_read_len_max(
+    const Write_offsets &offsets) {
```
percona-ysorokin: `noexcept`

percona-ysorokin (commented):
Looks good with only minor comments.

General question: can we completely avoid filling/using `write_offsets` when `innodb_encrypt_online_alter_logs` is disabled?

I would also suggest to run `innodb` MTR test suite with explicitly enabled `innodb_encrypt_online_alter_logs`, something like `--mysqld=--innodb-encrypt-online-alter-logs=ON` (probably with additional file keyring preload options)

storage/innobase/ddl/ddl0merge.cc
ldonoso: There is no easy way. @percona-ysorokin , please, consider that it depends on the size of the new index and of the configuration variables `--innodb-ddl-threads` and `--innodb-ddl-buffer-size`.

 Anyway, the performance improvement of calling `reserve` in advance is insignificant considering all the operations are carried out during the `ALTER TABLE` command.

ldonoso:
> Looks good with only minor comments.
> 
> General question: can we completely avoid filling/using `write_offsets` when `innodb_encrypt_online_alter_logs` is disabled?
> 
> I would also suggest to run `innodb` MTR test suite with explicitly enabled `innodb_encrypt_online_alter_logs`, something like `--mysqld=--innodb-encrypt-online-alter-logs=ON` (probably with additional file keyring preload options)

ldonoso:
> Looks good with only minor comments.
> 
> General question: can we completely avoid filling/using `write_offsets` when `innodb_encrypt_online_alter_logs` is disabled?
> 
> I would also suggest to run `innodb` MTR test suite with explicitly enabled `innodb_encrypt_online_alter_logs`, something like `--mysqld=--innodb-encrypt-online-alter-logs=ON` (probably with additional file keyring preload options)

@percona-ysorokin, yes, there are ways of preventing filling `io_write_offsets` but as it is a small vector, I think it is not worth the trouble.

@percona-ysorokin, I've run the `innodb` suite with this feature on:
    
```
    --mysqld=--innodb-encrypt-online-alter-logs=ON \
    --suite=innodb
```

There is no need of setting a keyring plugin. If no encryption is set, AES is used.

There are just three test failing:

```
Completed: Failed 4/738 tests, 99.46% were successful.

Failing test(s): shutdown_report innodb.innodb-index-debug innodb.percona_changed_page_bmp_requests_debug innodb.percona_doublewrite_single_page_flush
```

The new one is `innodb.innodb-index-debug` but this one with `innodb-encrypt-online-alter-logs` is failing in 8.0.26 as well. We can deal with the issue but I think the proper way is using another tasks/PR.

percona-ysorokin (approved):
LGTM

PR #4582, opened 2022-01-05
PS-4772: Break down SP stats in slow log per statement

oleksandr-kachan:
https://jira.percona.com/browse/PS-4772

In case log_slow_sp_statements=ON stats for each separate statement
within stored procedure continue accumulating for each next statement.
The issue affects Rows_sent, Rows_examined and Rows_affected counters.

To improve this reset slow query log stats before executing each
statement within stored procedure. The following is done in order
to achieve this:
- the statement execution is wrapped with THD::reset_sub_statement_state_slow_extended()
  and THD::restore_sub_statement_state_slow_extended() methods.
- Updated THD::reset_sub_statement_state_slow_extended() and
  THD::restore_sub_statement_state_slow_extended() methods to make them
  safe to use separately from THD::reset_sub_statement_state().
  These methods should handle sent_row_count, examined_row_count
  and row_count_func attrs reset/restore as they are used by slow log.


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/35665/

PR #4587, opened 2022-01-10
PS-8028: Server crashes with Assertion `m_prebuilt->table->n_ref_coun…

kamil-holubicki:
…t > 0' failed

https://jira.percona.com/browse/PS-8028

Cause:
5.7 does not support atomic DDL. It may happen that MySql's frm file is
out of sync with InnodDB dictionary. In such a case the table that is
part of InnoDB partitioned table is not opened which causes asserts.

Fix:
We are not able to recover from this case automatically, but we can
improve the error handling.
If the partitioned table cannot be opened, and its 1st partition is
marked as corrupted, return error without proceeding with further
opening.

Also, logging was improved as it may happen that when the inconsistency
is detected (opening the table), partition names are not initialized
yet, which results with the spoiled log.

### DIFF
--- a/mysql-test/suite/innodb/t/partition_fmt_inconsistency.test
+++ b/mysql-test/suite/innodb/t/partition_fmt_inconsistency.test
@@ -1,0 +1,41 @@
+#
+# Test the case when server is aborted after updating InnoDB dictionary,
+# but before fmt file gets updated. This leads to the inconsistency which
+# cannot be auto healied. Informative error message should be printed
+# and there should be possible to drop such a table.
+#
+--source include/have_innodb.inc
+
+--disable_query_log
+call mtr.add_suppression('InnoDB: Partition .* contains 2 user defined columns in InnoDB, but 1 columns in MySQL');
+--enable_query_log
+
+CREATE TABLE t1 (id INT)
+  PARTITION BY RANGE(id) (
+    PARTITION p0 VALUES LESS THAN (100),
+    PARTITION p1 VALUES LESS THAN (200),
+    PARTITION p2 VALUES LESS THAN (300),
+    PARTITION p3 VALUES LESS THAN (400)
+);
+
+SET SESSION DEBUG = '+d,alter_table_crash_before_frm_replace';
+
+--source include/expect_crash.inc
+
+--error 2013
+ALTER TABLE t1 ADD a INT;
+
+--echo # Restart mysqld after the crash and reconnect.
+--source include/start_mysqld.inc
+
+--error ER_NO_SUCH_TABLE
+SELECT * FROM t1;
+
+# Check that the error log contains the information about corrupted table
+--let $assert_file = $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $assert_count = 4
+--let $assert_select = InnoDB: Partition .* contains 2 user defined columns in InnoDB, but 1 columns in MySQL.
+--let $assert_text = InnoDB vs MySql columns mismatch.
+--source include/assert_grep.inc
+
+DROP TABLE t1;
--- a/storage/innobase/handler/ha_innopart.cc
+++ b/storage/innobase/handler/ha_innopart.cc
@@ -197,7 +197,14 @@ Ha_innopart_share::open_one_table_part(
 		&& (m_table_share->fields
 		    != dict_table_get_n_user_cols(ib_table)
 		       + dict_table_get_n_v_cols(ib_table) - 1))) {
-		ib::warn() << "Partition `" << get_partition_name(part_id)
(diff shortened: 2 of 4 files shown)


storage/innobase/handler/ha_innopart.cc
```diff
@@ -197,7 +197,14 @@ Ha_innopart_share::open_one_table_part(
 		&& (m_table_share->fields
 		    != dict_table_get_n_user_cols(ib_table)
 		       + dict_table_get_n_v_cols(ib_table) - 1))) {
-		ib::warn() << "Partition `" << get_partition_name(part_id)
+        /* We may end up here when doing ha_innopart::open(), so before
+           populate_partition_name_hash() was called. In such a case
+           get_partition_name() returns nullptr. Fall back to any other
+           informative string instead of trying to print nullptr. */
```
satya-bodapati:
```suggestion
        /* We may end up here when doing ha_innopart::open(), so before
        populate_partition_name_hash() was called. In such a case
        get_partition_name() returns nullptr. Fall back to any other
        informative string instead of trying to print nullptr. */
```

storage/innobase/handler/ha_innopart.cc
```diff
@@ -1056,6 +1066,11 @@ ha_innopart::open(
 	m_clust_pcur_parts = NULL;
 	m_pcur_map = NULL;
 
+    if (ib_table->corrupted) {
+        close();
+        DBUG_RETURN(HA_ERR_TABLE_CORRUPT);
+    }
```
satya-bodapati: what does this close()? do :thinking: will it internally call dict_table_close()?

Also, can we do the same for normal tables? ie return HA_ERR_TABLE_CORRUPT on ib_table->corrupted?

satya-bodapati (changes requested):

satya-bodapati (commented):
@kamil-holubicki is it possible to create mtr testcase? it should be easy to crash after .frm changes are synced?

If required, pstress has a feature to crash with SIG4. this creates core and shows the alter thread at the time of crash.

storage/innobase/handler/ha_innopart.cc
kamil-holubicki: It iterates over all partitions and closes them.
If the partition is corrupted, dict_table_close() is not called.

For normal tables I see that ib_table is set to nullptr. This causes HA_ERR_NO_SUCH_TABLE to be returned (around line 7162). I changed the error returned from ha_innopart::open() to HA_ERR_NO_SUCH_TABLE to be consistent with ha_innodb::open() behavior.

satya-bodapati (changes requested):
Looks good.
Can you please address the align comment? also try for mtr test.

mysql-test/suite/innodb/t/partition_fmt_inconsistency.test
```diff
@@ -0,0 +1,45 @@
...
+    PARTITION p0 VALUES LESS THAN (100),
+    PARTITION p1 VALUES LESS THAN (200),
+    PARTITION p2 VALUES LESS THAN (300),
+    PARTITION p3 VALUES LESS THAN (400)
+);
+
+SET SESSION DEBUG = '+d,alter_table_crash_before_frm_replace';
+
+# Write file to make mysql-test-run.pl expect crash
+--exec echo "restart" > $_expect_file_name
```
satya-bodapati: use expect_crash.inc

mysql-test/suite/innodb/t/partition_fmt_inconsistency.test
```diff
@@ -0,0 +1,45 @@
...
+# and there should be possible to drop such a table.
+#
+--source include/have_innodb.inc
+
+--disable_query_log
+call mtr.add_suppression('InnoDB: Partition .* contains 2 user defined columns in InnoDB, but 1 columns in MySQL');
+--enable_query_log
+
+--let $_server_id = `SELECT @@server_id`
+--let $_expect_file_name = $MYSQLTEST_VARDIR/tmp/mysqld.$_server_id.expect
```
satya-bodapati: with expect_crash.inc, we dont need this I think

satya-bodapati (changes requested):
Looks good, minor test case comments

mysql-test/suite/innodb/t/partition_fmt_inconsistency.test
kamil-holubicki: done

mysql-test/suite/innodb/t/partition_fmt_inconsistency.test
kamil-holubicki: done

satya-bodapati (approved):

PR #4589, opened 2022-01-11

voodam:

This is new PR instead of https://github.com/percona/percona-server/pull/4568 to comply GCA process.

### DIFF
--- a/mysql-test/suite/innodb/t/partition_autoinc.test
+++ b/mysql-test/suite/innodb/t/partition_autoinc.test
@@ -1,0 +1,9 @@
+--source include/have_innodb.inc
+
+#################################
+--echo # PS-7856: update on partition tables crashes the server
+#################################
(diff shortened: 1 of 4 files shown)


mysql-test/suite/innodb/t/partition_autoinc.test
```diff
@@ -0,0 +1,7 @@
+#################################
```
kamil-holubicki, line 3: `--source include/have_innodb.inc`

kamil-holubicki (changes requested):

mysql-test/suite/innodb/t/partition_autoinc.test
voodam: Done

kamil-holubicki (approved):
LGTM. Please squash commits before continuing with merge to 8.0 (GCA)

PR #4590, opened 2022-01-11

voodam:

This is new PR instead of https://github.com/percona/percona-server/pull/4567 to comply GCA process.

### DIFF
--- a/mysql-test/suite/innodb/t/partition_autoinc.test
+++ b/mysql-test/suite/innodb/t/partition_autoinc.test
@@ -1,0 +1,7 @@
+#################################
+--echo # PS-7856: update on partition tables crashes the server
+#################################
+CREATE TABLE t1(ip_col INT, i1 INT AUTO_INCREMENT, INDEX tt_2_pi1(i1)) PARTITION BY HASH (ip_col);
+INSERT INTO t1(ip_col) VALUES(1);
+UPDATE t1 SET i1 = 1 WHERE ip_col = 1;
+DROP TABLE t1;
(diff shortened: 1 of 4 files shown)


satya-bodapati (changes requested):
1. please fix commit message and github description. I think the write_set part is not relevant anymore
2. This doesn't look a GCA commit. If this is fixed in 5.7, we should see two commits for 8.0 branch

voodam:
> 1. please fix commit message and github description. I think the write_set part is not relevant anymore

Done, my bad.

> 2. This doesn't look a GCA commit. If this is fixed in 5.7, we should see two commits for 8.0 branch

You mean second commit with resolving conflicts? Yes, but AFAIU I first need to merge PR for 5.7: https://github.com/percona/percona-server/pull/4589, otherwise this 5.7 PR will have wrong commit. Or is this wrong undestanding of GCA process?

UPD. I probably do things in the wrong order: I need to create 8.0 PR after merging 5.7, not at the same time, right? Now it's need to wait while 5.7 PR will be merged.

mysql-test/suite/innodb/t/partition_autoinc.test
```diff
@@ -0,0 +1,7 @@
+#################################
```
kamil-holubicki, line 3: `--source include/have_innodb.inc`

voodam: Done (for 5.7)

voodam:
I need to close PR again because of missunderstanding GCA, sorry for the mess. Now all should be OK.
Here is the new one: https://github.com/percona/percona-server/pull/4603

PR #4591, opened 2022-01-11
PS-7865 [8.0]: Dropping a table with discarded tablespace crashes the server

inikep:
This is how the bug occurs:
1. If table have a lot of data, these conditions are satisfied and `btr_search_update_block_hash_info` returns true: https://github.com/percona/percona-server/blob/8.0/storage/innobase/btr/btr0sea.cc#L588 on one of `LOAD DATA` commands execution. This causes the creation of a hash index.
2. On discarding table this line is reached: https://github.com/percona/percona-server/blob/8.0/storage/innobase/row/row0mysql.cc#L3994. This will be the problematic index on which assert fails because of `index->space == FIL_NULL` but `block->page.id.space()` stays the same.
3. This erros occurs only on a lot amount of data, otherwise a hash index will not be created and hence will not be droped.

I suppose that solution can be in widening assert condition to include `index->space == FIL_NULL` case. But I'am not sure, is it okay that we drop the reseted index in `btr_search_drop_page_hash_index`? But early returning from the function probably is not an option, because in that case `assert_block_ahi_empty(block)` is failed:
```
  if (block->index == nullptr ||
      block->index->space == FIL_NULL) {
    rw_lock_s_unlock(latch);
    assert_block_ahi_empty(block);
    return;
  }
```

Assert was originally added in https://github.com/percona/percona-server/commit/195760a57028 commit. Probably it's not good idea to widen other same asserts because other cases are not related with dropping index, but with creating, updating etc. But I'am also not sure.

I added a little faster version of a testcase. It differs from the original one, but works in the same way.

### DIFF
--- a/mysql-test/suite/innodb/r/discard_tablespace.result
+++ b/mysql-test/suite/innodb/r/discard_tablespace.result
@@ -1,4 +1,10 @@
 #
+# PS-7865: Dropping a table with discarded tablespace crashes the server
+#
+CREATE TABLE t1(a INT)ENGINE=innodb;
+ALTER TABLE t1 DISCARD TABLESPACE;
+DROP TABLE t1;
+#
 # Bug #29793800 ASSERT AFTER DISCARD TABLESPACE, RENAME TABLE AND
 # CREATE TABLE USING SAME NAME.
 #
--- a/mysql-test/suite/innodb/t/discard_tablespace.test
+++ b/mysql-test/suite/innodb/t/discard_tablespace.test
@@ -1,4 +1,24 @@
 --echo #
+--echo # PS-7865: Dropping a table with discarded tablespace crashes the server
+--echo #
+
+CREATE TABLE t1(a INT) ENGINE=innodb;
+
+--disable_query_log
+--let $n=2048
+BEGIN;
+while ($n)
+{
+INSERT INTO t1 VALUES (10);
+--dec $n
+}
+COMMIT;
+--enable_query_log
+
+ALTER TABLE t1 DISCARD TABLESPACE;
+DROP TABLE t1;
+
+--echo #
 --echo # Bug #29793800 ASSERT AFTER DISCARD TABLESPACE, RENAME TABLE AND
 --echo # CREATE TABLE USING SAME NAME.
 --echo #
--- a/storage/innobase/btr/btr0sea.cc
+++ b/storage/innobase/btr/btr0sea.cc
@@ -1120,7 +1120,7 @@ retry:
   ut_ad(!index->disable_ahi);
   ut_ad(btr_search_enabled);
 
-  ut_ad(block->page.id.space() == index->space);
+  ut_ad(block->page.was_stale() || block->page.id.space() == index->space);
   ut_a(index_id == index->id);
   ut_a(!dict_index_is_ibuf(index));
 #ifdef UNIV_DEBUG


mysql-test/suite/innodb/r/discard_tablespace.result
```diff
@@ -1,4 +1,31 @@
 #
+# PS-7865: Dropping a table with discarded tablespace crashes the server
+#
+CREATE TABLE t1(a INT)ENGINE=innodb;
+INSERT INTO t1 VALUES (10);
+SELECT * FROM t1 INTO OUTFILE 'tmp1.txt';
```
kamil-holubicki: Why this way? What is the advantage of having 10 separate files? (btw, files should be deleted at the end of the test).
Is it really so much faster with  LOAD DATA INFILE than INSERT INTO... SELECT?

Why not:
```
--let $n=10
--disable_qery_log
while ($n)
{
INSERT INTO t1 SELECT * FROM t1;
--dec $n
}
--enable_query_log
```

or just:
```
--disable_query_log
--let $n=2048
BEGIN;
while ($n)
{
INSERT INTO t1 VALUES (10);
--dec $n
}
COMMIT;
--enable_query_log
```

storage/innobase/btr/btr0sea.cc
```diff
@@ -1120,7 +1120,8 @@ void btr_search_drop_page_hash_index(buf_block_t *block) {
   ut_ad(!index->disable_ahi);
   ut_ad(btr_search_enabled);
 
-  ut_ad(block->page.id.space() == index->space);
+  ut_ad(index->space == FIL_NULL ||
```
kamil-holubicki: The root cause of the issue is:
  When space is discarded (ALTER TABLE ... DISCARD TABLESPACE)
  we set index->space = FIL_NULL in row_discard_tablespace()
  Then when we DROP TABLE ... we get here from
  btr_search_drop_page_hash_when_freed() => buf_page_get_gen() =>
  buf_page_free_stale() => buf_LRU_free_page()
  We can detect this case by examining
  block->page.get_space()->was_not_deleted or block->page.was_stale()
  
My proposal is to fix it in the following way:
```
ut_ad(block->page.was_stale() || block->page.id.space() == index->space);
```

satya-bodapati: Agree, stale page check is better.

mysql-test/suite/innodb/r/discard_tablespace.result
voodam: Yes, the last one is also working. Done

storage/innobase/btr/btr0sea.cc
voodam: Done

mysql-test/suite/innodb/t/discard_tablespace.test
```diff
@@ -3,30 +3,43 @@
...
-LOAD DATA INFILE 'tmp10.txt' INTO TABLE t1;
+--dec $n
+}
+COMMIT;
+--enable_query_log
+
 ALTER TABLE t1 DISCARD TABLESPACE;
 DROP TABLE t1;
 
+#INSERT INTO t1 VALUES (10);
```
kamil-holubicki: Dead code. Remove it, please.

kamil-holubicki (commented):
The solution looks fine.
Please:
1. Remove dead code (see comment)
2. Sqush commits
3. Write a proper commit description

mysql-test/suite/innodb/t/discard_tablespace.test
voodam: Done

mysql-test/suite/innodb/t/discard_tablespace.test
```diff
@@ -1,3 +1,23 @@
+--echo #
+--echo # PS-7865: Dropping a table with discarded tablespace crashes the server
+--echo #
+
+CREATE TABLE t1(a INT)ENGINE=innodb;
```
satya-bodapati:
```suggestion
CREATE TABLE t1(a INT) ENGINE=InnoDB;
```

satya-bodapati (changes requested):
Please explain a bit more in the commit message on what causes the assertion failure(I see GitHub description has more detailed info. You can summarize and write better commit message)

mysql-test/suite/innodb/t/discard_tablespace.test
voodam: Done, and tried to write good message.

rahulmalik87:
This would be fixed by https://jira.percona.com/browse/PS-8190

LenzGr:
Related PR: https://github.com/percona/percona-server/pull/4809

PR #4593, opened 2022-01-12
Make mysqldump compatible with ProxySQL: don't die when SHOW STATUS is unsupported by server

voodam:
https://jira.percona.com/browse/PS-7486

Do i need to create the same PR for 8.0 branch too? It probably should be GCA?

### DIFF
--- a/client/mysqldump.c
+++ b/client/mysqldump.c
@@ -1319,28 +1319,52 @@ static void maybe_die(int error_num, const char* fmt_reason, ...)
   some.
 
   SYNOPSIS
-    mysql_query_with_error_report()
-    mysql_con       connection to use
-    res             if non zero, result will be put there with
-                    mysql_store_result()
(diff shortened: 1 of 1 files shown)


kamil-holubicki (commented):
I'm not sure about embedding ProxySQL specific fixes in mysqldump. Do we really need this?

What about

--ignore-error=1045 ?

https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html#option_mysqldump_ignore-error

It seems to be designed exactly for such cases.

voodam:
> 



> What about
> 
> --ignore-error=1045 ?
> 
> https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html#option_mysqldump_ignore-error
> 
> It seems to be designed exactly for such cases.

Yes, you right. It seems that that PR is useless.

PR #4595, opened 2022-01-14
5.7 PS-8032 Inplace Add index with lock=exclusive doesn’t generate MLOG_A…

rahulmalik87:
…DD_INDEX redo

https://jira.percona.com/browse/PS-8032
Problem:
Server is not generating MLOG_ADD_INDEX in case of inplace add index
with exclusive lock DDL. It is very important for backup tool to know if
any transaction happened without redo logging.
Fix:
Generate MLOG_ADD_INDEX if index was created using btr_blk.

### DIFF
--- a/mysql-test/suite/innodb/t/log_inplace_alter_with_exclusiv_lock.test
+++ b/mysql-test/suite/innodb/t/log_inplace_alter_with_exclusiv_lock.test
@@ -1,0 +1,33 @@
+--source include/have_debug.inc
+
+--echo #
+--echo # PS-8032 Inplace Add index with lock=exclusive doesn't generate MLOG_ADD_INDEX redo
+--echo #
+
+SET GLOBAL innodb_page_cleaner_disabled_debug=1;
+
+--source include/no_checkpoint_start.inc
+CREATE TABLE t1 (a INT NOT NULL, b INT UNIQUE) ENGINE=InnoDB TABLESPACE=innodb_file_per_table;
+# MLOG_INDEX_LOAD will not be emitted for empty tables. Insert a row.
+INSERT INTO t1 VALUES (1,2);
+ALTER TABLE t1 ADD INDEX(a), ALGORITHM=INPLACE, LOCK=EXCLUSIVE;
+
+--let CLEANUP_IF_CHECKPOINT=DROP TABLE t1;
+--source include/no_checkpoint_end.inc
+
+--let $restart_parameters = restart: --debug=d,ib_log
+--source include/start_mysqld.inc
+
+let SEARCH_FILE = $MYSQLTEST_VARDIR/log/mysqld.1.err;
+# Look for MLOG_INDEX_LOAD in the error log.
+let SEARCH_PATTERN=scan .*:.*log rec MLOG_INDEX_LOAD;
+--source include/search_pattern.inc
+
+CHECK TABLE t1;
+
+# Remove the --debug=d,ib_log setting.
+--let $restart_parameters =
+--source include/restart_mysqld.inc
+
+DROP TABLE t1;
+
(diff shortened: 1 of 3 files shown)


storage/innobase/row/row0merge.cc
```diff
@@ -4718,6 +4718,9 @@ row_merge_build_indexes(
 				error = btr_bulk.finish(error);
 			}
 		}
+		if (error == DB_SUCCESS) {
+			row_merge_write_redo(indexes[i]);
+		}
```
satya-bodapati: it is wrong to write redo before flush.

Also, is it possible to reach here without the btr bluk ? How do we truly know that we have written rows using btr bluk insert?

satya-bodapati (changes requested):

storage/innobase/row/row0merge.cc
rahulmalik87: Now doing it inside the if block. IIUC it would be using btr bulk  except when building FTS index.

satya-bodapati: @rahulmalik87 the patch is still wrong. Writing redo log record before flush.

satya-bodapati: Also better to do it in one place instead of duplicating in all branches. Just check if btrbulk is inited and records written using btr bulk

satya-bodapati (changes requested):
Check my latest comments

storage/innobase/row/row0merge.cc
rahulmalik87: removed the duplicating in all branch

mysql-test/suite/innodb/t/log_inplace_alter_with_exclusiv_lock.test
```diff
@@ -0,0 +1,33 @@
+--source include/have_debug.inc
+
+--echo #
+--echo # Bug#106163 Inplace Add index with lock=exclusive doesn't generate MLOG_ADD_INDEX redo
+ --echo #
+
+SET GLOBAL innodb_page_cleaner_disabled_debug=1;
+
+--source include/no_checkpoint_start.inc
```
satya-bodapati, line 9: instead of checkpoint_end.inc and no_checkpoint, can we use disable checkpoitn_disabled=true?

mysql-test/suite/innodb/t/log_inplace_alter_with_exclusiv_lock.test
```diff
@@ -0,0 +1,33 @@
...
+
+let SEARCH_FILE = $MYSQLTEST_VARDIR/log/mysqld.1.err;
+# Look for MLOG_INDEX_LOAD in the error log.
+let SEARCH_PATTERN=scan .*:.*log rec MLOG_INDEX_LOAD;
+--source include/search_pattern.inc
+
+CHECK TABLE t1;
+
+# Remove the --debug=d,ib_log setting.
+--let $restart_parameters = restart;
```
satya-bodapati: --let $restart_parameters =

satya-bodapati (changes requested):
Looks good, minor comments

mysql-test/suite/innodb/t/log_inplace_alter_with_exclusiv_lock.test
rahulmalik87: `innodb_checkpoint_disabled`  is not defined in 5.7

satya-bodapati: @rahulmalik87 ok, please take care of this in the 8.0 merge

mysql-test/suite/innodb/t/log_inplace_alter_with_exclusiv_lock.test
```diff
@@ -0,0 +1,33 @@
+--source include/have_debug.inc
+
+--echo #
+--echo # Bug#106163 Inplace Add index with lock=exclusive doesn't generate MLOG_ADD_INDEX redo
```
satya-bodapati: Please use PS-8032

satya-bodapati (approved):
one minor comment.. just found. Also please don't merge before 8.0 version approval

satya-bodapati (approved):

PR #4601, opened 2022-01-20
PS-7890 : Unable to install RocksDB engine plugin if server is started with --loose-rocksdb_persistent_cache_size_mb

avbelov23:
Added checks to bypass RocksDB restrictions

The problem has been dragging on for a long time and there is little chance that it will be fixed.
facebook/mysql-5.6#911
facebook/mysql-5.6#713

I don't see an explanation of why the cache size is limited to 100 Mb. I found a person who wrote the code, but he has one contact through linkedin.com (https://www.linkedin.com/in/karrad/). I asked him a question, but even if he answers, this is third party code and won't affect anything in the near future. Therefore, I propose this patch.

### DIFF
--- a/mysql-test/suite/rocksdb/r/persistent_cache.result
+++ b/mysql-test/suite/rocksdb/r/persistent_cache.result
@@ -11,3 +11,9 @@ a
 1
 # restart
 DROP TABLE t1;
+call mtr.add_suppression("RocksDB:");
+call mtr.add_suppression("Plugin 'ROCKSDB'");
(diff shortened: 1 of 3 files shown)


kamil-holubicki (changes requested):
The patch itself looks fine. However, as it goes to 8.0 (https://github.com/percona/percona-server/pull/4485) as well we need to follow GCA process.

avbelov23:
Ok

kamil-holubicki (approved):
lgtm

PR #4603, opened 2022-01-20

voodam:

This is new PR instead of #4590 to comply GCA process (again).


voodam:
All commits will be squashed to single after review.

voodam:
I'll wait until https://github.com/percona/percona-server/pull/4589 will be approved, then rebase this PR in the right way and reopen it.

kamil-holubicki (approved):
lgtm

kamil-holubicki:
@voodam Patch is OK, but it looks like clang formating check fails. Please adjust.
image

PR #4604, opened 2022-01-20
Bug#33341623: SHOW PROCESSLIST with terminology_use_previous hits fai…

george-lorch:
…lsafe in race condition

SHOW PROCESSLIST may inspect a given thread from another at pretty much any time. If proc_info() is NULL at this point and @@global.terminology_use_previous is set (e.g. to BEFORE_8_0_26), the look-up for alternative terminology is attempted using said nullptr. Wrapping the look-up in an if() testing whether there actually is anything to look up prevents failsafes from triggering.

Change-Id: Ic30ebf1742ddc9403136c9ac040566567e0ce7b4
(cherry picked from commit 4bff15b94917c4996e2a54beb92c99d474cfb656)


george-lorch:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/1156/

venkatesh-prasad-v (approved):

PR #4616, opened 2022-01-30
PS-7586: Apple build fix

dutow:
Issue: undefined symbols aren't allowed on OSX modules by default.

Fix: instruct the linker to allow it.


percona-ysorokin (approved):
LGTM

PR #4617, opened 2022-01-31
PS-8064: Restrict dynamic log file locations

dutow:
Previously dynamic logs files could be moved to anywhere on the
filesystem, potentially overwriting database files.

This commit adds the 'secure-log-path' system variable, similar to
'secure-file-priv', which restricts these files to a specific directory.

### DIFF
--- a/mysql-test/t/secure_log_path-master.opt
+++ b/mysql-test/t/secure_log_path-master.opt
@@ -1,0 +1,1 @@
+--secure_log_path=$MYSQL_TMP_DIR
--- a/share/messages_to_error_log.txt
+++ b/share/messages_to_error_log.txt
@@ -339,28 +339,28 @@ ER_CANT_CREATE_SHUTDOWN_THREAD
   eng "Can't create thread to handle shutdown requests (errno= %d)"
 
 ER_SEC_FILE_PRIV_CANT_ACCESS_DIR
-  eng "Failed to access directory for --secure-file-priv. Please make sure that directory exists and is accessible by MySQL Server. Supplied value : %s"
+  eng "Failed to access directory for --%s. Please make sure that directory exists and is accessible by MySQL Server. Supplied value : %s"
 
 ER_SEC_FILE_PRIV_IGNORED
-  eng "Ignoring --secure-file-priv value as server is running with --initialize(-insecure)."
+  eng "Ignoring --%s value as server is running with --initialize(-insecure)."
 
 ER_SEC_FILE_PRIV_EMPTY
-  eng "Insecure configuration for --secure-file-priv: Current value does not restrict location of generated files. Consider setting it to a valid, non-empty path."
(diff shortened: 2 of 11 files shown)


share/messages_to_error_log.txt
```diff
@@ -339,28 +339,28 @@ ER_CANT_CREATE_SHUTDOWN_THREAD
   eng "Can't create thread to handle shutdown requests (errno= %d)"
 
 ER_SEC_FILE_PRIV_CANT_ACCESS_DIR
-  eng "Failed to access directory for --secure-file-priv. Please make sure that directory exists and is accessible by MySQL Server. Supplied value : %s"
+  eng "Failed to access directory for --%s. Please make sure that directory exists and is accessible by MySQL Server. Supplied value : %s"
 
 ER_SEC_FILE_PRIV_IGNORED
-  eng "Ignoring --secure-file-priv value as server is running with --initialize(-insecure)."
+  eng "Ignoring --%s value as server is running with --initialize(-insecure)."
```
satya-bodapati, line 346: should we duplicate all thse to ER_SECURE_LOG_PRIV variants?

mysql-test/t/secure_log_path-master.opt
```diff
@@ -0,0 +1 @@
+--secure_log_path=$MYSQL_TMP_DIR
```
satya-bodapati, line 1: Can you also add a test case for the scenario with default (ie without setting secure_log_path)? where will the general_log_file go?

satya-bodapati (changes requested):
Looks good. minor comments

share/messages_to_error_log.txt
dutow: That would make things more complex on the C++ side. We have common functions which use these error codes and add the argument name as a parameter to it.

If we duplicate them, we would have to use different error codes for the two variables, which would either need lookup arrays or many additional if/else in these functions.

Also, these error codes can only appear during startup, as these are read-only variables, I don't think anybody would want to differentiate them using the numerical error code.

satya-bodapati: Ok Zsolt! lets go with existing error messages then

mysql-test/t/secure_log_path-master.opt
dutow: Secure log path defaults to empty, which allows logs to be placed anywhere, as without this change. (the only difference is that now we display a warning about this during startup)

And this scenario is tested by several existing general_log/slow_log tests (main.rewrite_general_log, main.log_tables, main.log_tables_debug, ...)

satya-bodapati:
@dutow please provide the Jenkins link

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1165/

satya-bodapati (approved):

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1202/

PR #4622, opened 2022-02-02
8.0 PS-8032 Inplace Add index with lock=exclusive doesn’t generate MLOG_ADD_INDEX redo

rahulmalik87:
https://jira.percona.com/browse/PS-8032
    Problem:
    Server is not generating MLOG_ADD_INDEX in case of inplace add index
    with exclusive lock DDL. It is very important for backup tool to know if
    any transaction happened without redo logging.
    Fix:
    Generate MLOG_ADD_INDEX if index was created using btr_blk.

### DIFF
--- a/mysql-test/suite/innodb/t/log_inplace_alter_with_exclusiv_lock.test
+++ b/mysql-test/suite/innodb/t/log_inplace_alter_with_exclusiv_lock.test
@@ -1,0 +1,33 @@
+--source include/have_debug.inc
+
+--echo #
+--echo # PS-8032 Inplace Add index with lock=exclusive doesn't generate MLOG_ADD_INDEX redo
+--echo #
+
+SET GLOBAL innodb_page_cleaner_disabled_debug=1;
+
+--source include/no_checkpoint_start.inc
+CREATE TABLE t1 (a INT NOT NULL, b INT UNIQUE) ENGINE=InnoDB TABLESPACE=innodb_file_per_table;
+# MLOG_INDEX_LOAD will not be emitted for empty tables. Insert a row.
+INSERT INTO t1 VALUES (1,2);
+ALTER TABLE t1 ADD INDEX(a), ALGORITHM=INPLACE, LOCK=EXCLUSIVE;
+
+--let CLEANUP_IF_CHECKPOINT=DROP TABLE t1;
+--source include/no_checkpoint_end.inc
+
+--let $restart_parameters = restart: --debug=d,ib_log
+--source include/start_mysqld.inc
+
+let SEARCH_FILE = $MYSQLTEST_VARDIR/log/mysqld.1.err;
+# Look for MLOG_INDEX_LOAD in the error log.
+let SEARCH_PATTERN=scan .*:.*log rec MLOG_INDEX_LOAD;
+--source include/search_pattern.inc
+
+CHECK TABLE t1;
+
+# Remove the --debug=d,ib_log setting.
+--let $restart_parameters =
+--source include/restart_mysqld.inc
+
+DROP TABLE t1;
+
--- a/mysql-test/suite/innodb/r/log_inplace_alter_with_exclusiv_lock.result
+++ b/mysql-test/suite/innodb/r/log_inplace_alter_with_exclusiv_lock.result
@@ -1,0 +1,15 @@
+#
+# PS-8032 Inplace Add index with lock=exclusive doesn't generate MLOG_ADD_INDEX redo
+#
+SET GLOBAL innodb_page_cleaner_disabled_debug=1;
+CREATE TABLE t1 (a INT NOT NULL, b INT UNIQUE) ENGINE=InnoDB TABLESPACE=innodb_file_per_table;
+INSERT INTO t1 VALUES (1,2);
+ALTER TABLE t1 ADD INDEX(a), ALGORITHM=INPLACE, LOCK=EXCLUSIVE;
+# Kill the server
+# restart: --debug=d,ib_log
+Pattern "scan .*:.*log rec MLOG_INDEX_LOAD" found
+CHECK TABLE t1;
+Table	Op	Msg_type	Msg_text
+test.t1	check	status	OK
+# restart
+DROP TABLE t1;
--- a/storage/innobase/row/row0merge.cc
+++ b/storage/innobase/row/row0merge.cc
@@ -4725,21 +4725,23 @@ wait_again:
 		if (indexes[i]->type & DICT_FTS) {
 			row_fts_psort_info_destroy(psort_info, merge_info);
 			fts_psort_initiated = false;
-		} else if (error != DB_SUCCESS || !online) {
-			/* Do not apply any online log. */
-		} else if (old_table != new_table) {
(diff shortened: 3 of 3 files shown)


mysql-test/suite/innodb/t/log_inplace_alter_with_exclusiv_lock.test
```diff
@@ -0,0 +1,31 @@
+--source include/have_debug.inc
+
+--echo #
+--echo # PS-8032 Inplace Add index with lock=exclusive doesn't generate MLOG_ADD_INDEX redo
+--echo #
+
+SET GLOBAL innodb_checkpoint_disabled=1;
+--source include/no_checkpoint_start.inc
```
satya-bodapati: remove this checkpoint_start.inc and end.inc

storage/innobase/ddl/ddl0builder.cc
```diff
@@ -1961,11 +1961,13 @@ dberr_t Builder::finalize() noexcept {
   if (err == DB_SUCCESS) {
     write_redo(m_index);
 
-    DEBUG_SYNC_C_IF_THD(m_ctx.thd(), "row_log_apply_before");
+    if (apply_log) {
+      DEBUG_SYNC_C_IF_THD(m_ctx.thd(), "row_log_apply_before");
 
-    err = row_log_apply(m_ctx.m_trx, m_index, m_ctx.m_table, m_local_stage);
+      err = row_log_apply(m_ctx.m_trx, m_index, m_ctx.m_table, m_local_stage);
```
satya-bodapati, line 1967: is there a tab before err?

storage/innobase/ddl/ddl0builder.cc
```diff
@@ -2045,19 +2047,17 @@ dberr_t Builder::finish() noexcept {
...
-    ut_a(m_index->online_status == ONLINE_INDEX_COMPLETE);
-
-    auto observer = m_ctx.m_trx->flush_observer;
-    observer->flush();
-
-  } else {
-    err = finalize();
-
+    if (!m_ctx.m_online) {
+      apply_log = false;
```
satya-bodapati, line 2054: this should be true right? with EXCLUSIVE, m_online is false

satya-bodapati (changes requested):

storage/innobase/ddl/ddl0builder.cc
rahulmalik87: > this should be true right? with EXCLUSIVE, m_online is false

yes. with EXCLUSIVE m_online is false.

storage/innobase/ddl/ddl0builder.cc
rahulmalik87: no only spaces.

mysql-test/suite/innodb/t/log_inplace_alter_with_exclusiv_lock.test
rahulmalik87: If I remove no_checkpoint_start.inc and end.inc. I get an error
exec of 'echo "restart: --debug=d,ib_log" > ' failed, error: 512, status: 2, errno: 0.
safe_process[92454]: Child process: 92455, exit: 1

satya-bodapati: issue is something else, please check spacing etc

storage/innobase/ddl/ddl0builder.cc
satya-bodapati: I see a hightighted space in github, please check

storage/innobase/ddl/ddl0builder.cc
satya-bodapati: ok, apply_log doesn't have effect on flush observer flush and writing redo. So this is fine?

storage/innobase/ddl/ddl0builder.cc
rahulmalik87: I think this just tells the difference of old code.  I have checked it is all spaces and not tab.

satya-bodapati: ok

storage/innobase/ddl/ddl0builder.cc
rahulmalik87: yeah. we discussed that we need to flush if we have used btr_blk. so that's why we would be flushing in all three cases.  old_table!=new_table, not online and  when online.

satya-bodapati: right rahul, makes sense

satya-bodapati (changes requested):
@rahulmalik87 Jenkins link plz, I want to see how this test is behaving on Jenkins and other innodb *alter* tests after this change.

storage/innobase/ddl/ddl0builder.cc
```diff
@@ -2047,7 +2047,7 @@ dberr_t Builder::finish() noexcept {
   }
 
   dberr_t err{DB_SUCCESS};
-  if (get_error() == DB_SUCCESS) {
+  if (get_error() == DB_SUCCESS && !m_index->table->is_temporary()) {
```
satya-bodapati, line 2050: Don't we do bulk load for ADD INDEX on temporary tables?
isn't there online log to apply for temporary tables?

satya-bodapati (changes requested):

storage/innobase/ddl/ddl0builder.cc
rahulmalik87: we do bulk load. IIUC, there wont' be online log to apply on temporary tables since these tables are per session.

satya-bodapati: IMHO, we should only skip writing redo. Flushing should be done

satya-bodapati: for temp tables

rahulmalik87: Flushing would ensure data is persistant to disk. but in the case of temporary table we don't need that. if server crashes. we can safely discard that table.

satya-bodapati: Hmm, you are right, nobody copies this datafile and we don't need to necessarily wait. Can you please check without your patch if observer->flush() is called or not for temp tables?

satya-bodapati:
@rahulmalik87 I don't see Jenkins link to analyze failures. Please update PR

storage/innobase/ddl/ddl0builder.cc
rahulmalik87: no we do not Flush_observer::flush() for temp tables.

rahulmalik87:
> @rahulmalik87 I don't see Jenkins link to analyze failures. Please update PR

https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/1163

satya-bodapati (changes requested):
Please squash them for merge

satya-bodapati (approved):

PR #4623, opened 2022-02-02
PS-7949 - Stop buf_lru_manager_thread during calls to buf_pool_invalidate instance

ldonoso:
https://jira.percona.com/browse/PS-7949

---

*Problem*:

For buffer pool invalidation to proceed we must ensure there is no
write activity happening.

PS does not implement this guarantee as PS implements a multi-threaded
LRU Manager unlike MySQL that uses a single thread. Hence, flush
operations in the initial state might be possible.

This error manifests when running the test `clone.remote_ddl_basic`.
Although, it is not deterministic as it involves a particular
interleaving of execution of several threads.

*Solution*:

Stop `buf_lru_manager_thread` when `buf_pool_invalidate_instance`
by checking that the event `buf_pool_t->no_invalidate` is set.

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -39,6 +39,7 @@ this program; if not, write to the Free Software Foundation, Inc.,
 
 #include "my_config.h"
 
+#include <boost/scope_exit.hpp>
 #include "btr0btr.h"
 #include "buf0buf.h"
 #include "fil0fil.h"
@@ -1326,6 +1327,9 @@ static void buf_pool_create(buf_pool_t *buf_pool, ulint buf_pool_size,
     buf_pool->no_flush[i] = os_event_create();
   }
 
+  buf_pool->no_invalidate = os_event_create();
+  os_event_set(buf_pool->no_invalidate);
+
   buf_pool->watch = (buf_page_t *)ut::zalloc_withkey(
       UT_NEW_THIS_FILE_PSI_KEY, sizeof(*buf_pool->watch) * BUF_POOL_WATCH_SIZE);
   for (i = 0; i < BUF_POOL_WATCH_SIZE; i++) {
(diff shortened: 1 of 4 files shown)


storage/innobase/buf/buf0buf.cc
```diff
@@ -5885,27 +5891,29 @@ static void buf_pool_invalidate_instance(buf_pool_t *buf_pool) {
...
+    might be running several LRU manager threads at this stage.
+    Hence, a new write batch can be in initialization stage at this point. */
 
-    /* However, it is possible that a write batch that has
-    been posted earlier is still not complete. For buffer
-    pool invalidation to proceed we must ensure there is NO
+    /* For buffer pool invalidation to proceed we must ensure there is NO
     write activity happening. */
-    if (buf_pool->n_flush[i] > 0) {
+    if (buf_pool->n_flush[i] > 0 || buf_pool->init_flush[i]) {
```
satya-bodapati, line 5908: we just suppressed the assertion. I don't understand the purpose of the event. What exactly it is helping us with?

if you stopped the LRU thread, then this code shouldn't change?

satya-bodapati (changes requested):

storage/innobase/buf/buf0buf.cc
ldonoso: @satya-bodapati, the event is reset to notify the LRU thread to stop the execution. Please, mind the following steps:

- We enter `buf_pool_invalidate_instance`. LRU thread is still running.
- We reset event `no_invalidate` to notify the LRU thread to stop. Mind that:
    - LRU might be still running so we have to remove the assertion `buf_pool->init_flush[i] == FALSE`
- If LRU is still running (we find out be checking `n_flush` and `init_flush` we wait for it to end
- At this point we are sure that the LRU is not running: In the previous step we waited for it to end and after that, the LUR thread is waiting for the event `no_invalidate` to be set.

satya-bodapati: got it @ldonoso in the normal case (not invalidate), how do we ensure the LRU thread doesn't simply wait for flush_event to be set?

satya-bodapati: Hmm, once set, it will never wait unless reset?

ldonoso: Yes, that's right.

Please, mind that after the wait is finished the event is **not** reset automatically so the next time the LRU thread calls `os_event_wait` it will find it set unless `buf_pool_invalidate_instance` has reset it.

ldonoso:
Close in favor of #4624

PR #4624, opened 2022-02-04
PS-7949 - Stop buf_lru_manager_thread during calls to buf_pool_invali…

ldonoso:
…date_instance

https://jira.percona.com/browse/PS-7949

---

*Problem*:

For buffer pool invalidation to proceed we must ensure there is no
write activity happening.

PS does not implement this guarantee as PS implements a multi-threaded
LRU Manager unlike MySQL that uses a single thread. Hence, flush
operations in the initial state might be possible.

This error manifests when running the test `clone.remote_ddl_basic`.
Although, it is not deterministic as it involves a particular
interleaving of execution of several threads.

*Solution*:

Stop `buf_lru_manager_thread` when `buf_pool_invalidate_instance`
by checking that the event `buf_pool_t->no_invalidate` is set.

### DIFF
--- a/storage/innobase/include/buf0buf.h
+++ b/storage/innobase/include/buf0buf.h
@@ -2297,6 +2297,12 @@ struct buf_pool_t {
   running. Protected by flush_state_mutex. */
   os_event_t no_flush[BUF_FLUSH_N_TYPES];
 
+  /* This event is always set at startup, so LRU threads do not wait for this
+  event. Before invalidating bufferpool, this event is reset, so the next LRU
+  batch flushing will wait for the event. Bufferpool invalidation needs LRU
+  flushing to be stopped. */
+  os_event_t run_lru;
+
   /** A red-black tree is used exclusively during recovery to speed up
   insertions in the flush_list. This tree contains blocks in order of
   oldest_modification LSN and is kept in sync with the flush_list.  Each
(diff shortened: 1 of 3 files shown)


ldonoso:
Supersedes #4623

storage/innobase/include/buf0buf.h
```diff
@@ -2297,6 +2297,10 @@ struct buf_pool_t {
   running. Protected by flush_state_mutex. */
   os_event_t no_flush[BUF_FLUSH_N_TYPES];
 
+  /** This is event is set when there is no ongoing invalidation of the
+   * buff_pool_t. */
```
satya-bodapati: We can improve the comment a bit:

This event is always set at startup, so LRU threads do not wait for this event. Before invalidating bufferpool, this event is reset, so the next LRU batch flushing will wait for the event. Bufferpool invalidation needs LRU flushing to be stopped

storage/innobase/include/buf0buf.h
```diff
@@ -2297,6 +2297,10 @@ struct buf_pool_t {
...
+  os_event_t no_invalidate;
```
satya-bodapati: LRU threads are stopped by using this event. The user is "buf pool invalidate". Maybe we should rename it to stop_lru ? or pause_lru?

satya-bodapati (changes requested):
Looks good. Some minor comments

storage/innobase/include/buf0buf.h
ldonoso: Good point! Please, mind I have called it the other way round that suggested: `buf_pool->run_lru`

satya-bodapati: Looks good

satya-bodapati (approved):

PR #4644, opened 2022-02-18
PS-7967 Document how to make NVME ZBD permissions/ownership/scheduler…

patrickbirch:
… settings permanent (8.0)

 M source/myrocks/zenfs.rst


percona-ysorokin (approved):
LGTM
the only suggestion is to change the "Permanently change io scheduler" link to something `udev`-based.

PR #4657, opened 2022-03-01
PS-8102: Fixing memory leak in keyring_common

dutow:
get_default_mr in keyring_common created a new instance of the default
memory resource every time the funciton was called instead of using a
single static instance.

This commit fixes this by making it a static variable.


altmannmarcelo (approved):
LGTM. Will cherry-pick to PXB

ldonoso (approved):

altmannmarcelo (approved):
LGTM

components/keyrings/common/data/global_default_mr.h
```diff
@@ -19,7 +19,7 @@
 
 inline ::boost::container::pmr::memory_resource
     * ::boost::container::pmr::get_default_resource() BOOST_NOEXCEPT {
-  psi_memory_resource *global_default_mr = new psi_memory_resource{};
+  static psi_memory_resource *global_default_mr = new psi_memory_resource{};
```
ldonoso: Please, why not the following? On the other hand, is there not an already available `ps_memory_resource` published by the FW?

```
static psi_memory_resource global_default_mr;

inline ::boost::container::pmr::memory_resource
    * ::boost::container::pmr::get_default_resource() BOOST_NOEXCEPT {
  return &global_default_mr;
}
```

ldonoso (changes requested):

components/keyrings/common/data/global_default_mr.h
dutow: Please check PS-8004, as this leak is caused by the fix for that issue (the original code, before the fix was exactly what you suggested here].

ldonoso (approved):

ldonoso (changes requested):
Please, do we have concurrency problems over `global_default_mr`?

dutow:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1189/

ldonoso (approved):

PR #4658, opened 2022-03-01
PS-7890 - Remove calls to deinit_logging_service_for_plugin

ldonoso:
The logging service is used later, so if it is deinitialized, there will
be a crash.

The logging service is deinitialized anyway during the shutdown of
RocksDB.


kamil-holubicki (approved):
lgtm

PR #4665, opened 2022-03-07
PS-8103: Do not unload dynamic libraries in an ASAN build

dutow:
Issue: address sanitizer and valgrind can't resolve stack traces inside
plugins and components. The reason behind this is that MySQL unloads
dynamic libraries loaded dynamically during shutdown, making the code
inaccessible to the sanitizers.

This also means that suppressions do not work inside components and
plugins, if the suppression involves a step in the stacktrace which is
inside the plugin/component.

Fix: add a conditional around dlunload to remove it from asan/valgrind
builds.


satya-bodapati (approved):

altmannmarcelo (approved):
LGTM

PR #4680, opened 2022-03-17
PS-8047 - Fix compilation of pfs_instr-t

ldonoso:
https://jira.percona.com/browse/PS-8047

---

*Problem*:

gcc-7 compilation fails with the following error:

```
ld.lld: error: duplicate symbol: log_message(int, ...)
```
                                    ^
The function is defined in `pfs_server_stubs` and `sql_main`.

*Solution*:

Remove inclusion of `sql_main` which is not necessary.

### DIFF
--- a/libbinlogevents/src/gtids/gtidset.cpp
+++ b/libbinlogevents/src/gtids/gtidset.cpp
@@ -21,6 +21,7 @@
    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA */
 
 #include "libbinlogevents/include/gtids/gtidset.h"
+#include "my_compiler.h"
 #include <map>
 #include <string>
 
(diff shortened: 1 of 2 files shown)


libbinlogevents/src/gtids/gtidset.cpp
```diff
@@ -235,7 +236,10 @@ bool Gtid_set::is_empty() const { return m_gtid_set.empty(); }
 std::size_t Gtid_set::count() const {
   std::size_t count{0};
 
+  MY_COMPILER_DIAGNOSTIC_PUSH()
+  MY_COMPILER_GCC_DIAGNOSTIC_IGNORE("-Wunused-variable")
   for (auto const &[uuid, intervals] : m_gtid_set) {
```
dutow, line 241: Would `for([[maybe_unused]] auto const ...)` work?

ldonoso: @dutow , I tried but gcc 7 has no implemented `[[maybe_unused]]` for structure bindings.

PR #4681, opened 2022-03-17
PS-8115 - Fix `CREATE EVENT DEFINER`

ldonoso:
https://jira.percona.com/browse/PS-8115
https://ps57.cd.percona.com/job/percona-server-5.7-pipeline/26486/

*Problem*:

When the user contains the character `@` this is associated to the
hostname.

*Solution*:

Do as 8.0 and consider that the hostname is the part from the last '@'
till the end of the value of `DEFINER`.

'@' is an illegal character in the hostname.

### DIFF
--- a/mysql-test/r/events_bugs.result
+++ b/mysql-test/r/events_bugs.result
@@ -402,6 +402,24 @@ ERROR 42000: Access denied; you need (at least one of) the SUPER privilege(s) fo
 DROP EVENT e1;
 ERROR HY000: Unknown event 'e1'
 DROP USER mysqltest_u1@localhost;
+CREATE USER 'mysqltest_u1@corp'@'localhost';
+GRANT EVENT ON events_test.* TO 'mysqltest_u1@corp'@'localhost';
+CREATE DEFINER='mysqltest_u1@corp'@'localhost' EVENT e1 ON SCHEDULE EVERY 1 DAY STARTS '2006-01-01 00:00:00' DO SELECT 1;
+SHOW CREATE EVENT e1;
+Event	sql_mode	time_zone	Create Event	character_set_client	collation_connection	Database Collation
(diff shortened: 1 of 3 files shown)


mysql-test/r/events_bugs.result
```diff
@@ -402,6 +402,14 @@ ERROR 42000: Access denied; you need (at least one of) the SUPER privilege(s) fo
 DROP EVENT e1;
 ERROR HY000: Unknown event 'e1'
 DROP USER mysqltest_u1@localhost;
+CREATE USER 'mysqltest_u1@corp'@'localhost';
```
satya-bodapati, line 405: what about two @ @ in user? Can you please add one more test scenario?

Also a negative test for @ in hostname. for example: 'localhost@domain' in the host part

satya-bodapati, line 405: what about two @ @ in user? Can you please add one more test scenario?

Also a negative test for @ in hostname. for example: 'localhost@domain' in the host part

sql/event_data_objects.cc
```diff
@@ -446,18 +465,8 @@ Event_job_data::load_from_row(THD *thd, TABLE *table)
...
-  if (! ptr)
-    ptr= definer.str;
-
-  len= ptr - definer.str;
-  definer_user.str= strmake_root(&mem_root, definer.str, len);
-  definer_user.length= len;
-  len= definer.length - len - 1;
-  /* 1:because of @ */
-  definer_host.str= strmake_root(&mem_root, ptr + 1, len);
-  definer_host.length= len;
```
satya-bodapati, line 460: how does this change fix the issue? this seems like simple refactoring?

ldonoso: Yes, it is a refactoring to eliminate some duplication. This way we eliminate future errors when we modify the logic in some part of the code but forget to update the other one.

Please, mind that we are not using `strchr` anymore but `strrchar`

satya-bodapati: Thanks @ldonso My eyes couldn't catch the extra "r"

satya-bodapati (changes requested):
Looks good @ldonoso Can you please squash and submit to Jenkins pipeline, one platform should be sufficient?

satya-bodapati (approved):
@ldonoso Please squash and recreate the 8.0 commmit

PR #4684, opened 2022-03-22
PS-8115 - Fix `CREATE EVENT DEFINER`

ldonoso:
*Problem*:

When the user contains the character `@` this is associated to the
hostname.

*Solution*:

Do as 8.0 and consider that the hostname is the part from the last '@'
till the end of the value of `DEFINER`.

### DIFF
--- a/sql/event_data_objects.cc
+++ b/sql/event_data_objects.cc
@@ -22,6 +22,7 @@
 
 #include "event_data_objects.h"
 
+#include "my_user.h"
 #include "sql_parse.h"                          // parse_sql
 #include "strfunc.h"                           // find_string_in_array
 #include "sql_db.h"                        // get_default_db_collation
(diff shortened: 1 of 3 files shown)


sql/event_data_objects.cc
```diff
@@ -446,18 +464,8 @@ Event_job_data::load_from_row(THD *thd, TABLE *table)
   Event_creation_ctx::load_from_db(thd, &mem_root, dbname.str, name.str, table,
                                    &creation_ctx);
 
-  ptr= strchr(definer.str, '@');
-
-  if (! ptr)
-    ptr= definer.str;
-
-  len= ptr - definer.str;
-  definer_user.str= strmake_root(&mem_root, definer.str, len);
```
satya-bodapati: @ldonoso in commit message, you told 8.0 already uses strrchar instead of strchar()? :thinking:

satya-bodapati (approved):
Looks good. Please ensure the test passes on 8.0

PR #4692, opened 2022-03-28
PS-6029 - Fix format of result of data_masking.gen_rnd_us_phone

ldonoso:
https://jira.percona.com/browse/PS-6029


mysql-test/suite/data_masking/t/gen_rnd_us_phone-master.opt
```diff
@@ -0,0 +1 @@
+$DATA_MASKING_PLUGIN_OPT $DATA_MASKING_PLUGIN_LOAD
```
dutow: This could be `suite.opt` since all tests have the same master.opt in the suite.

dutow (approved):

PR #4695, opened 2022-03-29
PS-8141: percona_encrypt_tmp_files_debug MTR test

kamil-holubicki:
fails if executed with --mem option

https://jira.percona.com/browse/PS-8141

Problem:
When percona_encrypt_tmp_files_debug MTR test is executed with --mem
option, it fails with the assertion

Cause:
The test uses proc filesystem to find the file used by filesort.
When the test is executed with --mem option, mysqld tmp directory
is located in /dev/shm and procfs links directly to it, not to
the location reported by SELECT @@global.tmpdir. This causes file not
being found and the test fails.

Solution:
Do not use the result of SELECT @@global.tmpdir in grep in the test.
Use tmp/*/mysqld.1 value instead which is constant anyway.

### DIFF
--- a/mysql-test/t/percona_encrypt_tmp_files_debug.test
+++ b/mysql-test/t/percona_encrypt_tmp_files_debug.test
@@ -33,7 +33,13 @@ while($encryption_mode < 2)
 
   # Determining server temp dir where 'filesort()' temporary files will be
   # created.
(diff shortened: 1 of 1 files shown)


percona-ysorokin (approved):
LGTM

PR #4697, opened 2022-03-29
Fixed PS-8136 - LOCK TABLES FOR BACKUP is not preventing ALTER INSTAN…

altmannmarcelo:
…CE ROTATE INNODB MASTER KEY

https://jira.percona.com/browse/PS-8136

Problem:
Lightweight MDL LTFB should block operations that will result on
inconsistent backups. Rotating master key after the backup has started
will make tables header to point to a key that is not present on backup
cache of keys as PXB reads them once at the beginning of the backup.
Rotating the master key should be blocked.

Fix:
Enhanced Rotate_innodb_master_key to wait for up to lock_wait_timeout
seconds in case other session has acquired LTFB and also fail in case
LTFB has been acquired by the current session.

### DIFF
--- a/sql/sql_alter_instance.cc
+++ b/sql/sql_alter_instance.cc
@@ -90,6 +90,14 @@ Rotate_innodb_master_key::execute()
     return true;
   }
 
+  // Acquire Percona's LOCK TABLES FOR BACKUP lock
+  if (m_thd->backup_tables_lock.abort_if_acquired() ||
+      m_thd->backup_tables_lock.acquire_protection(
+          m_thd, MDL_TRANSACTION, m_thd->variables.lock_wait_timeout))
+  {
+    return true;
+  }
+
   if (hton->rotate_encryption_master_key())
   {
     /* SE should have raised error */
--- a/mysql-test/r/backup_locks.result
+++ b/mysql-test/r/backup_locks.result
@@ -909,6 +909,13 @@ CREATE TABLE t(a INT);
 LOCK TABLE t READ;
 LOCK BINLOG FOR BACKUP;
 DROP TABLE t;
+LOCK TABLES FOR BACKUP;
+SET SESSION lock_wait_timeout=1;
+ALTER INSTANCE ROTATE INNODB MASTER KEY;
+ERROR HY000: Lock wait timeout exceeded; try restarting transaction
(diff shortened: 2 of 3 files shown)


altmannmarcelo:
Single platform run - https://ps57.cd.percona.com/job/percona-server-5.7-pipeline/26602/

sql/sql_alter_instance.cc
```diff
@@ -90,6 +90,14 @@ Rotate_innodb_master_key::execute()
     return true;
   }
 
+  // Acquire Percona's LOCK TABLES FOR BACKUP lock
```
kamil-holubicki, line 93: I've got only one concern. I don't know if it matters or not. I would need to analyze it deeper. Maybe @satya-bodapati will be able to answer.
Let's consider the case without PXB.
Session 1 executes `ALTER INSTANCE ROTATE INNODB MASTER KEY`. It acquires this lock.
A the same time Session 2 gets to the point when `backup_tables_lock` lock is requested, like ```if (thd->backup_tables_lock.abort_if_acquired()) return true;``` in  ```lock.cc::lock_schema_name()```. There are a lot of such paths.

Without the fix, these paths would simply acquire the lock (or not return true). Now the flow will change (parallelism will decrease).
On the flip side, I can imagine that the code responsible for MK rotation is resilient to paths that are currently acquiring `backup_tables_lock`. If we make these paths mutually exclusive, probably some logic of MK handling will become obsolete.
Yet another possibility is that MK rotation occurs rather rarely, so most probably all above is true, but does not matter :)

satya-bodapati: @kamil-holubicki MK rotation already takes the upstream backup lock (LIFB). We extended the fix for percona's backup lock (LTFB).

Rotation is not a frequent operation IMHO.

altmannmarcelo: @kamil-holubicki not sure I get what you are saying here, but `abort_if_acquired` returns `ER_CANT_EXECUTE_WITH_BACKUP_LOCK` if LTFB has been acquired by the current session,
so there is no parallelism involved here.

kamil-holubicki (approved):
lgtm

satya-bodapati (approved):

rahulmalik87 (approved):
LGTM

PR #4699, opened 2022-03-31
PS-7769 - Fix use-after-return error in audit_log_exclude_accounts_validate

ldonoso:
*Problem:*

`st_mysql_value::val_str` might return a pointer to `buf` which after
the function called is deleted. Therefore the value in `save`, after
reuturnin from the function, is invalid.

In this particular case, the error is not manifesting as `val_str`
returns memory allocated with `thd_strmake` and it does not use `buf`.

*Solution:*

Allocate memory with `thd_strmake` so the memory in `save` is not local.

### DIFF
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -35,6 +35,7 @@
 #include "buffer.h"
 #include "audit_handler.h"
 #include "filter.h"
+#include "sql_const.h"
 
 #define PLUGIN_VERSION 0x0002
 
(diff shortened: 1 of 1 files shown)


plugin/audit_log/audit_log.c
```diff
@@ -1593,24 +1593,31 @@ static MYSQL_THDVAR_STR(query_stack,
                         PLUGIN_VAR_NOSYSVAR | PLUGIN_VAR_NOCMDOPT,
                         "Query stack.", NULL, NULL, "");
 
+static const char *val_strmake(MYSQL_THD              thd,
+                               struct st_mysql_value *mysql_val)
+{
+  char        buf[80];
```
oleksandr-kachan: Lets probably use STRING_BUFFER_USUAL_SIZE here

plugin/audit_log/audit_log.c
```diff
@@ -1593,24 +1593,31 @@ static MYSQL_THDVAR_STR(query_stack,
...
+  int         len= sizeof(buf);
+  const char *val= mysql_val->val_str(mysql_val, buf, &len);
+
+  if (val)
```
oleksandr-kachan: "val != NULL" ?

oleksandr-kachan (commented):
Please take a look at comments and add some description for the PR. Lets also run MTR tests (including ASAN enabled) for the plugin.

ldonoso:
> Please take a look at comments and add some description for the PR. Lets also run MTR tests (including ASAN enabled) for the plugin.

https://ps57.cd.percona.com/job/percona-server-5.7-asan-param/51/

oleksandr-kachan (approved):

PR #4701, opened 2022-03-31
Improve extraction of plugin sources.

MaisenbacherD:
Removes line continuation when collecting plugin sources.

With this change source files can be found which are defined over multiple
lines like:
```
plugin_SOURCES = \
	file1.cc \
	file2.cc \
	file3.cc
```

Signed-off-by: Dennis Maisenbacher <dennis.maisenbacher@wdc.com>

### DIFF
--- a/storage/rocksdb/CMakeLists.txt
+++ b/storage/rocksdb/CMakeLists.txt
@@ -98,6 +98,8 @@ ELSE()
       MESSAGE(SEND_ERROR "Missing RocksDB plugin makefile in ${PLUGIN_ROOT}")
     ELSE()
       FILE(READ ${PLUGINMKFILE} PLUGINMK)
+      #Remove line continuation/breaking
+      STRING(REGEX REPLACE "[ \t\r]*\\\\\n[ \t\r]*" " "  PLUGINMK ${PLUGINMK})
       STRING(REGEX MATCH "SOURCES[ ]*=[ ]*([^\n]*)" FOO ${PLUGINMK})
       SET(MK_SOURCES ${CMAKE_MATCH_1})
       SEPARATE_ARGUMENTS(MK_SOURCES)


storage/rocksdb/CMakeLists.txt
```diff
@@ -98,6 +98,8 @@ ELSE()
       MESSAGE(SEND_ERROR "Missing RocksDB plugin makefile in ${PLUGIN_ROOT}")
     ELSE()
       FILE(READ ${PLUGINMKFILE} PLUGINMK)
+      #Remove line continuation/breaking
+      STRING(REGEX REPLACE "\\\\\n[ \t]+" ""  PLUGINMK ${PLUGINMK})
```
percona-ysorokin: This may not work if there is no space before `\`
For instance, for
```
plugin_SOURCES =\
	file1.cc\
	file2.cc\
	file3.cc
```
we will have `file1.ccfile1.ccfile3.cc`

MaisenbacherD: That is correct. 
How about 
```
      STRING(REGEX REPLACE "[ \t]*\\\\\n[ \t]+" " "  PLUGINMK ${PLUGINMK})
```

percona-ysorokin: I would add something like
```suggestion
      STRING(REGEX REPLACE "\\\\\n" ""  PLUGINMK ${PLUGINMK})
      STRING(REGEX REPLACE "\\s+" " "  PLUGINMK ${PLUGINMK})
```

MaisenbacherD: "\s" does not seem to be a valid escape character in CMake. 
The [CMake specification](https://cmake.org/cmake/help/latest/command/string.html#regex-specification) mentions that `[ \t\r\n]` matches any single whitespace char. 

Something like that?:
```
STRING(REGEX REPLACE "[ \t\r]*\\\\\n[ \t\r]*" " "  PLUGINMK ${PLUGINMK})
```
Where `[ \t\r]*` is zero or more whitespace without linebreaks,
and `\\\\\n` is the "\\" followed by a linebreak,
and `[ \t\r]*` is zero or more whitespace which is the indentation.

percona-ysorokin: Yes, @MaisenbacherD
This one should work

MaisenbacherD: Changed the regex :)

percona-ysorokin (approved):
LGTM

percona-ysorokin:
Closed in favour of https://github.com/percona/percona-server/pull/4732

PR #4704, opened 2022-04-01
PS-8145: Fix MYSQL_LEX_CSTRING initialization in sql_audit::mysql_audit_notify() (5.7)

oleksandr-kachan:
https://jira.percona.com/browse/PS-8145

The sizeof(const char *) was used as a string length during general_sql_command
initialization mysql_event_general_subclass_t audit event. As a result
there was wrong string length in some cases.
Changed it to use std::string::length() to properly initialize string
length.


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/552/

nitendra-bhosle (approved):
LGTM

PR #4708, opened 2022-04-03
PS-7923 Sort metadata locks acquired during ALTER TABLESPACE

i-rinat:
https://jira.percona.com/browse/PS-7923

Problem:
Metadata locks upgraded during ALTER TABLESPACE are upgraded in
no particular order, which present deadlock possibility. Metadata deadlock
prevention mechanism kicks in and rolls back an attachable transaction
during data dictionary updates, which causes the server to terminate.

Fix:
Sort metadata lock requests before upgrade procedure.
Move metadata lock upgrading routine to MDL_context class to reuse the
sorting comparator.

### DIFF
--- a/sql/mdl.cc
+++ b/sql/mdl.cc
@@ -58,6 +58,7 @@
 extern MYSQL_PLUGIN_IMPORT CHARSET_INFO *system_charset_info;
 
 static PSI_memory_key key_memory_MDL_context_acquire_locks;
+static PSI_memory_key key_memory_MDL_context_upgrade_shared_locks;
 
 #ifdef HAVE_PSI_INTERFACE
 static PSI_mutex_key key_MDL_wait_LOCK_wait_status;
@@ -83,6 +84,9 @@ static PSI_cond_info all_mdl_conds[] = {{&key_MDL_wait_COND_wait_status,
 
 static PSI_memory_info all_mdl_memory[] = {
     {&key_memory_MDL_context_acquire_locks, "MDL_context::acquire_locks", 0, 0,
+     "Buffer for sorting lock requests."},
+    {&key_memory_MDL_context_upgrade_shared_locks,
+     "MDL_context::upgrade_shared_locks", 0, 0,
      "Buffer for sorting lock requests."}};
 
 /**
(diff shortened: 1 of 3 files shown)


satya-bodapati (changes requested):
This fix looks much better than the previous one (which blocked connections).
Why didn't you bring the testcase from that commit?

i-rinat:
> Why didn't you bring the testcase from that commit?

That testcase will fail every time unless this pull request will also include setting of thread ID in `fsp_init_resume_alter_encrypt_tablespace()`. Unless I clump everything together into a combined pull request, the testcase will have to go to the third one. I haven't made the third one yet, because it strongly depends on the patch in this pull request. Without third patch the testcase will not fail every time, but instead will fail intermittently.

I've also just had an idea. It may be possible to make a specific testcase for the lock ordering issue if I place debug points in the right places. I'll try to explore that idea.

i-rinat:
> It may be possible to make a specific testcase for the lock ordering issue

Turned out to be not feasible. It's required to add debug point into cycles, and essentially single-step those cycles in a testcase. That testcase would then be highly-dependent on particular table names in "mysql" tablespace, and would silently stop checking anything useful if another table was introduced. Moreover, such test would not prevent accidental out of order lock acquiring in new code. So essentially that test would be a dead weight.

satya-bodapati:
> > Why didn't you bring the testcase from that commit?
> 
> That testcase will fail every time unless this pull request will also include setting of thread ID in `fsp_init_resume_alter_encrypt_tablespace()`. Unless I clump everything together into a combined pull request, the testcase will have to go to the third one. I haven't made the third one yet, because it strongly depends on the patch in this pull request. Without third patch the testcase will not fail every time, but instead will fail intermittently.
> 
> I've also just had an idea. It may be possible to make a specific testcase for the lock ordering issue if I place debug points in the right places. I'll try to explore that idea.

@i-rinat Just setting thread_id is fine. we can include this and bring the testcase. Does it need InnoDB fil_io() fix as well?

i-rinat:
> Just setting thread_id is fine. we can include this and bring the testcase. Does it need InnoDB fil_io() fix as well?

The test will trigger the assertion in InnoDB, sometimes.

Without the test and setting thread_id, so as this pull request is right now, it doesn't not really depend on the InnoDB's fix.

satya-bodapati (approved):

PR #4711, opened 2022-04-04
PS-7769 - Fix use-after-return error in audit_log_exclude_accounts_validate

ldonoso:
---

*Problem:*

`st_mysql_value::val_str` might return a pointer to `buf` which after
the function called is deleted. Therefore the value in `save`, after
reuturnin from the function, is invalid.

In this particular case, the error is not manifesting as val_str`
returns memory allocated with `thd_strmake` and it does not use `buf`.

*Solution:*

Allocate memory with `thd_strmake` so the memory in `save` is not local.

### DIFF
--- a/plugin/audit_log/audit_log.c
+++ b/plugin/audit_log/audit_log.c
@@ -35,6 +35,7 @@
 #include "buffer.h"
 #include "audit_handler.h"
 #include "filter.h"
+#include "sql_const.h"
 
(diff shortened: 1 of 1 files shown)


ldonoso:
CI: https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/37593/

oleksandr-kachan (approved):

PR #4713, opened 2022-04-04
PS-7980: Basic implementation of LDAP authorization

dutow:
This patch adds the ability to provide roles from an external source to
MySQL users with authentication plugins, and adds a simple LDAP
implementation.

How it works:

We extend the authentication info structure used by authentication
plugins by one more type, which should contain role names.

During authorization, the server will automatically add these to the in
memory cache, but not to the persisted mysql database. This means that
the roles will shouw up on queries working on the cache, such as SHOW
GRANTS, but will be invisible on queries working on the persisted
database.

Another limitation is that in the current implementation, external roles
are only populated when a user logs on. Users who didn't log on since
the server was started will only show persisted roles.

If needed, both of these could be changed in the future - for example by
adding another column to the role mappings (showing that it's an
externally managed role) they could be persisted.

We could also extend the plugin API to provide additional methods for
checking external changes automatically - for example by using a
background thread.

As permissions are only important when a user actually logs in, these
were omitted from this first version.

Interaction with related SQL commands:

* FLUSH PRIVILEGES reloads the in memory cache from the persisted
  database. With this patch we also track external roles separately, and
  automatically re-add them at the same place. This way once an external
  node gets added, it will always stay visible.
* External roles are rechecked for every login. If the list changes in
  the external database,those changes will be reflected in the new
  session
* For now we allow REVOKEing external roles - but they will be readded
  automatically during the next login. Maybe it would be a better
  solution to disable REVOKE for these with an error message explaining
  that they should be removed in the external application.

### DIFF
--- a/plugin/auth_ldap/src/auth_ldap_impl.cc
+++ b/plugin/auth_ldap/src/auth_ldap_impl.cc
@@ -25,18 +25,37 @@ namespace mysql {
 namespace plugin {
 namespace auth_ldap {
 
+static std::map<std::string, std::string> calc_group_mappings(
+    std::string const &mapping) {
+  std::map<std::string, std::string> group_mapping;
+  std::vector<std::string> roles;
+  boost::algorithm::split(roles, mapping, boost::is_any_of(","));
+  for (auto const &role : roles) {
+    std::vector<std::string> r;
+    boost::algorithm::split(r, role, boost::is_any_of("="));
+    if (r.size() == 1) {
+      group_mapping[role] = role;
+    } else {
+      group_mapping[r[0]] = r[1];
+    }
+  }
+  return group_mapping;
+}
+
 AuthLDAPImpl::AuthLDAPImpl(const std::string &user_name,
                            const std::string &auth_string,
                            const std::string &user_search_attr,
                            const std::string &group_search_filter,
                            const std::string &group_search_attr,
-                           const std::string &bind_base_dn, Pool *pool)
+                           const std::string &bind_base_dn,
+                           const std::string &mapping, Pool *pool)
     : pool_(pool),
       user_search_attr_(user_search_attr),
       group_search_attr_(group_search_attr),
       group_search_filter_(group_search_filter),
       bind_base_dn_(bind_base_dn),
-      user_name_(user_name) {
+      user_name_(user_name),
+      group_role_mapping_(calc_group_mappings(mapping)) {
   std::vector<std::string> parts;
   boost::algorithm::split(parts, auth_string, boost::is_any_of("#"));
   user_auth_string_ = boost::algorithm::trim_copy(parts[0]);
@@ -48,8 +67,9 @@ AuthLDAPImpl::AuthLDAPImpl(const std::string &user_name,
 
 AuthLDAPImpl::~AuthLDAPImpl() {}
 
-bool AuthLDAPImpl::bind(const std::string &user_dn,
-                        const std::string &password) {
+bool AuthLDAPImpl::bind_internal(const std::string &user_dn,
+                                 const std::string &password,
+                                 Pool::pool_ptr_t *conn_out) {
   log_srv_dbg("AuthLDAPImpl::bind()");
   bool success = false;
   std::ostringstream log_stream;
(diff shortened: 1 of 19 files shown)


plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -188,7 +224,7 @@ groups_t AuthLDAPImpl::search_ldap_groups(const std::string &user_dn) {
   log_srv_dbg("AuthLDAPImpl::search_ldap_groups");
   groups_t list;
 
-  std::shared_ptr<Connection> conn = pool_->borrow_connection();
+  std::shared_ptr<Connection> conn = pool_->borrow_connection(true);
```
altmannmarcelo: default_connect already defaults to true, any reason to enforcing it now?

sql/auth/role_tables.cc
```diff
@@ -265,6 +266,8 @@ bool populate_roles_caches(THD *thd, TABLE_LIST *tablelst) {
       }
 
       grant_role(acl_role, acl_user, *with_admin_opt == 'Y' ? true : false);
+
+
```
altmannmarcelo: remove

sql/auth/role_tables.cc
```diff
@@ -331,5 +334,26 @@ bool populate_roles_caches(THD *thd, TABLE_LIST *tablelst) {
...
+    for(const auto& p: g_external_roles) {
+      const char* host = p.first.second.c_str();
+      const char* username = p.first.first.c_str();
+      ACL_USER* acl_user = find_acl_user(host, username, true);
+      if (acl_user == nullptr) {
+        continue;
+      }
+
+      for(const auto& r: p.second) {
+        const char* host = r.second.c_str();
```
altmannmarcelo: are we re-declaring host here?

altmannmarcelo: actually yes, cannot compile:
```
/work/ps/src/8.0/sql/auth/role_tables.cc:347:21: error: declaration of ‘host’ shadows a previous local [-Werror=shadow=compatible-local]
         const char* host = r.second.c_str();
                     ^~~~
/work/ps/src/8.0/sql/auth/role_tables.cc:339:19: note: shadowed declaration is here
       const char* host = p.first.second.c_str();
```

altmannmarcelo (changes requested):
got myself familiar with the current LDAP implementation and scanned through your changes.
Will play a bit more tomorrow, but this is the first round of review.

Question, anything users will have to adjust if they already have simple_ldap authentication or upgrade will go without any extra steps?

sql/auth/role_tables.cc
dutow: I compiled this without maintainer mode - it shadows the previous declaration, but it is a different variable. I'll fix that, but it wasn't critical for a prototype.

plugin/auth_ldap/src/auth_ldap_impl.cc
dutow: No, this was a debugging leftover.

dutow:
For the users:

The ldap part introduces a new variable, where they will have to specify mapping between the ldap groups and mysql roles. By default that new variable is empty, and only roles present in it will be mapped.

Also, previously users only had to specify the search DN and password if they were using the authenticated_as feature - and it's also required for this feature to work.

plugin/auth_ldap/src/auth_ldap_impl.cc
```diff
@@ -25,18 +25,37 @@ namespace mysql {
 namespace plugin {
 namespace auth_ldap {
 
+static std::map<std::string, std::string> calc_group_mappings(std::string const &mapping) {
+  std::map<std::string,std::string> group_mapping;
```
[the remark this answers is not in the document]
altmannmarcelo: `boost::algorithm::split` will return an array with 1 empty element in case of no commas. we could validate if mapping and interrupt the flow without going through the `for loops`

altmannmarcelo (approved):
PR LGTM.

One small suggestion - this FR requires authentication_ldap_simple_bind_root_dn and authentication_ldap_simple_bind_root_pwd been configured. Since both are read_only we could improve user experience by validating if they are at least set when updating  `authentication_ldap_simple_group_role_mapping` and at least produce a warning on thread message area or into error log.

Apart from that, tested and it works. Code looks good to.

dutow:
@altmannmarcelo We could add that (and probably also for authenticated_as), I'll add it.

Also, as part of the refactoring, that's another part of the code I'm fixing: For 99% of the time, we don't really need a globally configured user, because all authenticated users have access to group membership information.
If those variables aren't configured, we could simply try to use the connection authenticated as the actual user to check group memberships, and it should work on most setups, both on linux and windows.

PR #4722, opened 2022-04-08
PS-8143: Fix memory leak in File_query_log::set_rotated_name() (5.7)

oleksandr-kachan:
https://jira.percona.com/browse/PS-8143

The opt_slow_logname now uses allocated memory to keep file path.
Make sure previously allocated memory is freed when generating
rotated file name. Also add PREALLOCATED flag in variable declaration.

### DIFF
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -2282,6 +2282,9 @@ bool File_query_log::set_rotated_name(const bool need_lock)
     }
 
     if (need_lock) mysql_mutex_lock(&LOCK_global_system_variables);
+    if (opt_slow_logname != NULL) {
+      my_free(opt_slow_logname);
+    }
     opt_slow_logname = my_strdup(key_memory_LOG_name, log_file_name, MYF(MY_WME));
     if (need_lock) mysql_mutex_unlock(&LOCK_global_system_variables);
   }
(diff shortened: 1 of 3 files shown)


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/553/

sql/mysqld.cc
```diff
@@ -3377,7 +3377,10 @@ int init_common_variables()
   FIX_LOG_VAR(opt_general_logname,
               make_query_log_name(logname_path, QUERY_LOG_GENERAL));
   FIX_LOG_VAR(opt_slow_logname,
-              make_query_log_name(slow_logname_path, QUERY_LOG_SLOW));
+              my_strdup(
+                  key_memory_LOG_name,
+                  make_query_log_name(slow_logname_path, QUERY_LOG_SLOW),
+                  MYF(MY_WME)));
```
satya-bodapati, line 3383: strdup is not necessary opt_general_logname (above) because it is not dynamic?

sql/log.cc
```diff
@@ -2282,6 +2282,9 @@ bool File_query_log::set_rotated_name(const bool need_lock)
     }
 
     if (need_lock) mysql_mutex_lock(&LOCK_global_system_variables);
+    if (opt_slow_logname != NULL) {
+      my_free(opt_slow_logname);
```
satya-bodapati, line 2286: where do we free without rotation?

satya-bodapati (changes requested):

sql/mysqld.cc
oleksandr-kachan: Yes

sql/log.cc
oleksandr-kachan: This should be done in Sys_var_charptr::cleanup() which is invoked in sys_var_end().

satya-bodapati (approved):

PR #4726, opened 2022-04-12
PS-8121 Document do not use dot (.) on binlog log_bin and log_bin_index variables (8.0)

alina-derkach-oaza:
modified: source/flexibility/binlogging_replication_improvements.rst


patrickbirch (approved):
LGTM

PR #4734, opened 2022-04-19
PS-8116 IMPORT TABLESPACE with different data directory fails

rahulmalik87:
PS-8116 IMPORT TABLESPACE with different data directory fails
    https://jira.percona.com/browse/PS-8116

    Problem:
    Reimplement Import tablespace with different data directory.
    https://jira.percona.com/browse/PS-1697

    Feature was removed in 8.0.20 and 5.7.30 due to upstream marking
    mismatch in data directory as an error.
    https://github.com/mysql/mysql-server/commit/9909cd94f896

    Introduce the support again in 5.7 and 8.0

### DIFF
--- a/mysql-test/suite/innodb/t/import_tablespace_schema_missmatch.test
+++ b/mysql-test/suite/innodb/t/import_tablespace_schema_missmatch.test
@@ -23,11 +23,9 @@ CREATE TABLE t1 (id int unsigned NOT NULL PRIMARY KEY);
 ALTER TABLE t1 DISCARD TABLESPACE;
 
 --move_file $MYSQL_TMP_DIR/t1.ibd $MYSQLD_DATADIR/$DB/t1.ibd
---error ER_TABLE_SCHEMA_MISMATCH
 
 ALTER TABLE t1 IMPORT TABLESPACE;
 DROP TABLE t1;
---remove_file $MYSQLD_DATADIR/$DB/t1.ibd
 
 --echo #
 --echo # Test-case-2
@@ -44,14 +42,16 @@ UNLOCK TABLES;
 DROP TABLE t1;
 --replace_result $MYSQL_TMP_DIR MYSQL_TMP_DIR
 eval CREATE TABLE t1 (id int unsigned NOT NULL PRIMARY KEY) $data_directory;
+SELECT * from t1;
 ALTER TABLE t1 DISCARD TABLESPACE;
 
 --move_file $MYSQL_TMP_DIR/t1.ibd $MYSQL_TMP_DIR/alt_dir/$DB/t1.ibd
 
---error ER_TABLE_SCHEMA_MISMATCH
+--disable_warnings
 ALTER TABLE t1 IMPORT TABLESPACE;
+--enable_warnings
+SELECT * FROM t1;
 DROP TABLE t1;
---remove_file $MYSQL_TMP_DIR/alt_dir/$DB/t1.ibd
 
 
 --echo #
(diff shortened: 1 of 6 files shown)


storage/innobase/row/row0import.cc
```diff
@@ -1321,28 +1321,20 @@ row_import::match_schema(
...
 {
 	/* Do some simple checks. */
+	const unsigned relevant_flags = m_flags & ~DICT_TF_MASK_DATA_DIR;
+	const unsigned relevant_table_flags =
+	    m_table->flags & ~DICT_TF_MASK_DATA_DIR;
 
-	if (m_flags != m_table->flags) {
-		if (dict_tf_to_row_format_string(m_flags) !=
-				dict_tf_to_row_format_string(m_table->flags)) {
+	if (relevant_flags != relevant_table_flags) {
```
satya-bodapati, line 1328: I guess the original patch never updated the FSP_MASK_FLAG and DICT_TF_FLAG on import?
FSP flag is present on the disk.
DICT_TF flag in InnoDB dictionary (SYS_TABLES*)

After a restart, do we get a flag mismatch error or warning?

rahulmalik87: Hello @satya-bodapati , I tested show create table, SELECT/INSERT after  restart of server. It is not reporting any error or warning on the table.

satya-bodapati: I debugged a bit and found out that DATA_DIR flag is not used for flag comparison

` /* Validate this single-table-tablespace with the data dictionary,
        but do not compare the DATA_DIR flag, in case the tablespace was
        remotely located. *
`

satya-bodapati: Datafile::validate_to_dd()

satya-bodapati (changes requested):
@rahulmalik87 patch looks good. Please provide jenkins links (innodb suites is enough)

rahulmalik87:
@satya-bodapati https://ps57.cd.percona.com/view/5.7/job/percona-server-5.7-param/556/#showFailuresLink

mysql-test/suite/innodb/t/import_tablespace_schema_missmatch.test
```diff
@@ -46,7 +46,9 @@ ALTER TABLE t1 DISCARD TABLESPACE;
 
 --move_file $MYSQL_TMP_DIR/t1.ibd $MYSQL_TMP_DIR/alt_dir/$DB/t1.ibd
 
+--disable_warnings
 ALTER TABLE t1 IMPORT TABLESPACE;
+--enable_warnings
```
satya-bodapati, line 52: warning is caused by cfg missing. Why the cfg is removed with this PR?

satya-bodapati (changes requested):
please squash as well

mysql-test/suite/innodb/t/import_tablespace_schema_missmatch.test
rahulmalik87: It is existing test case, to test import without the cfg file.

satya-bodapati: and the warnings didn't exist before? :thinking:

rahulmalik87: It is upstream test case, MS doesn't allow but PS allows

satya-bodapati: Ok, this testcase is not very clear whether import succeeds or not. IMPORT TABLESPACE errored Can you please add different data in source and destination table and add SELECT (after IMPORT) that shows data is always from source table

satya-bodapati (changes requested):

satya-bodapati:
@rahulmalik87 please squash

satya-bodapati (approved):

PR #4735, opened 2022-04-19
8.0 Fixed PS-8136 - LOCK TABLES FOR BACKUP is not preventing ALTER INSTAN...

altmannmarcelo:
…CE ROTATE INNODB MASTER KEY

https://jira.percona.com/browse/PS-8136

Problem:
Lightweight MDL LTFB should block operations that will result on
inconsistent backups. Rotating master key after the backup has started
will make tables header to point to a key that is not present on backup
cache of keys as PXB reads them once at the beginning of the backup.
Rotating the master key should be blocked.

Fix:
Enhanced Rotate_innodb_master_key to wait for up to lock_wait_timeout
seconds in case other session has acquired LTFB and also fail in case
LTFB has been acquired by the current session.


altmannmarcelo:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-pipeline/37791/

satya-bodapati (approved):

kamil-holubicki (approved):
lgtm

PR #4738, opened 2022-04-26
Fixed PS-8174 - Assertion failure: buf0flu.cc:3567:UT_LIST_GET_LEN(bu…

altmannmarcelo:
…f_pool->flush_list) == 0

https://jira.percona.com/browse/PS-8174

Problem:
There is a possibility that at shutdown by the time we do the last
sweep on flushing the buffer pool there are still pages in the flush
list. Those pages are still marked as io_fix->BUF_IO_READ thus they are
not eligible for flushing from flush_list.

Where is the workflow:

1. ibuf_merge_in_background requested those pages to be read in order to
merge the ibuf changes. This will mark the page as BUF_IO_READ and
increment buf_pool->n_pend_reads by 1.
2. When IO threads pick them up, it will start to merge the insert
bugger changes.
3. On the first change, it will add the page to flush_list.
4. If there are more changes to apply, it will and continue on applying
the changes until it is done.
5. Once the io thread finishes applying ibuf records to this page, it
will mark the page as BUF_IO_NONE
6. the io thread decreases buf_pool->n_pend_reads by 1.

The last sweep on flushing buffer pool considers the round of flushes
completed when n_flushed == 0 which is not correct, if it runs when we
are at step 4.

Also, there is a still another race condition that by the time we tried
to flush a page it was still marked as BUF_IO_READ (step 5), but when
the page cleaner code checks buf_get_n_pending_read_ios, the IO thread
has already decremented leaving one last page on flush_list.

Fix:

For the last round flushing pages at shutdown, only consider the flush
completed if there are no pending read io operations and flush_list size
is 0.

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -429,6 +429,24 @@ buf_pool_get_oldest_modification(void)
 	return(oldest_lsn);
 }
 
+ulint
+buf_get_flush_list_len(const buf_pool_t *buf_pool) {
+	ulint pages = 0;
+	if (buf_pool == NULL) {
+		for (ulint i = 0; i < srv_buf_pool_instances; i++) {
+			buf_pool_t *buf_pool_instance;
+
+			buf_pool_instance = buf_pool_from_array(i);
+
+			pages +=
+			    UT_LIST_GET_LEN(buf_pool_instance->flush_list);
+		}
+	} else {
+		pages = UT_LIST_GET_LEN(buf_pool->flush_list);
+	}
+	return (pages);
+}
+
 /********************************************************************//**
 Get total buffer pool statistics. */
 void
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -3494,7 +3494,9 @@ DECLARE_THREAD(buf_flush_page_cleaner_coordinator)(
 
 		buf_flush_wait_batch_end(NULL, BUF_FLUSH_LIST);
 
-	} while (!success || n_flushed > 0);
+	} while (!success || n_flushed > 0 ||
+		 buf_get_n_pending_read_ios() > 0 ||
+		 buf_get_flush_list_len(NULL) > 0);
 
 	/* Some sanity checks */
 	ut_a(srv_get_active_thread_type() == SRV_NONE);
--- a/storage/innobase/include/buf0buf.h
+++ b/storage/innobase/include/buf0buf.h
@@ -1484,6 +1484,12 @@ buf_pool_watch_occurred(
 	const page_id_t&	page_id)
 MY_ATTRIBUTE((warn_unused_result));
 
+/** Get the current length of the flush list.
+@param[in] buf_pool buffer pool instance or nullptr to all instances
+@return number of pages in flush list */
+ulint
+buf_get_flush_list_len(const buf_pool_t *buf_pool);
+
 /********************************************************************//**
 Get total buffer pool statistics. */
 void


altmannmarcelo:
https://ps57.cd.percona.com/view/5.7/job/percona-server-5.7-pipeline/27007/

storage/innobase/include/buf0buf.h
```diff
@@ -1484,6 +1484,13 @@ buf_pool_watch_occurred(
 	const page_id_t&	page_id)
 MY_ATTRIBUTE((warn_unused_result));
 
+/** Get the current lenght of the flush list. It acquires and releases
```
satya-bodapati: typo: length

storage/innobase/buf/buf0flu.cc
```diff
@@ -3494,7 +3494,9 @@ DECLARE_THREAD(buf_flush_page_cleaner_coordinator)(
 
 		buf_flush_wait_batch_end(NULL, BUF_FLUSH_LIST);
 
-	} while (!success || n_flushed > 0);
+	} while (!success || n_flushed > 0 ||
+		 buf_get_n_pending_read_ios() > 0 ||
+		 buf_get_flush_list_len(NULL) > 0);
```
satya-bodapati, line 3499: I think these should be aligned to "!" in "!success"

storage/innobase/buf/buf0buf.cc
```diff
@@ -429,6 +429,29 @@ buf_pool_get_oldest_modification(void)
 	return(oldest_lsn);
 }
 
+ulint
+buf_get_flush_list_len(buf_pool_t *buf_pool) {
+	ulint i;
+	ulint pages = 0;
+	if (buf_pool == NULL) {
+		for (i = 0; i < srv_buf_pool_instances; i++) {
```
satya-bodapati: for (ulint i = 0..

satya-bodapati (changes requested):
minor comments. Testcase possible with debug sleep?

storage/innobase/buf/buf0buf.cc
```diff
@@ -429,6 +429,29 @@ buf_pool_get_oldest_modification(void)
...
+			buf_pool_t *buf_pool_instance;
+
+			buf_pool_instance = buf_pool_from_array(i);
+
+			buf_flush_list_mutex_enter(buf_pool_instance);
```
kamil-holubicki: Not a strong opinion here, but do we need to acquire this mutex here?
```#define UT_LIST_GET_LEN(BASE) (BASE).count```
So even if we dirty-read non-zero value, we will return zero next time, which will cause one additional loop turn (line 3485).

No other comments than from @satya-bodapati

Comment (altmannmarcelo) [reply storage/innobase/buf/buf0buf.cc]: Thinking about this particular case, I would say no.
We first test the number of pending reads on PC loop, the workflow is as follows:

```
IO_BUF_READ
n_pend_reads ++
add to flush_list
IO_BUF_NONE
n_pending_reads--
```

So `buf_get_n_pending_read_ios() > 0 ||` will evaluate true and we will not even call `buf_get_flush_list_len()`, and by the time we call `buf_get_flush_list_len` its guarantee that no other thread will be adding new entries to flush_list.

If you don't have any objections @satya-bodapati  I will remove it.

File: storage/innobase/buf/buf0flu.cc (hunk 2 above)
Comment (altmannmarcelo) [reply storage/innobase/buf/buf0flu.cc:3499]: @satya-bodapati , they seem to align already:
Screen Shot 2022-04-27 at 12 58 53

**Comment by altmannmarcelo** [comment]
> minor comments. Testcase possible with debug sleep?

@satya-bodapati unfortunately it's not deterministic. I can reproduce with initiating shutdown when we merge ibuf using full iops but that requires some random number of rans to crash the server.

File: storage/innobase/buf/buf0buf.cc (hunk 4 above)
Comment (satya-bodapati) [reply storage/innobase/buf/buf0buf.cc]: @kamil-holubicki @altmannmarcelo what if n_pending_reads is 0 and the flush_list_len len is also seen 0 (after a dirty page is added)? Isn't this possible?

File: storage/innobase/buf/buf0flu.cc (hunk 2 above)
Comment (satya-bodapati) [reply storage/innobase/buf/buf0flu.cc:3499]: Check if white space is used instead of tabs? the first few alignment should be done by tab. later spaces.

Comment (satya-bodapati) [reply storage/innobase/buf/buf0flu.cc:3499]: I still see mismatched alignment in github

File: storage/innobase/buf/buf0buf.cc (hunk 4 above)
Comment (altmannmarcelo) [reply storage/innobase/buf/buf0buf.cc]: @satya-bodapati I don't think this is possible, at this stage of shutdown, the only thing that is running is purge and the pending io reads. as explained on the above workflow, we increment n_pend_reads -> add page to flush_list so it's now possible to have a new page added if n_pend_reads == 0

Comment (satya-bodapati) [reply storage/innobase/buf/buf0buf.cc]: @altmannmarcelo I mean, due to instruction (load/store) reordering between n_pending_reads & flush_list.count. These are not atomics and no read/write barriers here.

Comment (altmannmarcelo) [reply storage/innobase/buf/buf0buf.cc]:
```
1. IO_BUF_READ - at srv_master_do_idle_tasks->ibuf_merge_in_background->ibuf_merge->ibuf_merge_pages->buf_read_ibuf_merge_pages->buf_read_page_low->buf_page_init_for_read
2.. n_pend_reads ++ - at srv_master_do_idle_tasks->ibuf_merge_in_background->ibuf_merge->ibuf_merge_pages->buf_read_ibuf_merge_pages->buf_read_page_low->buf_page_init_for_read
3. add to flush_list - io_handler_thread->fil_aio_wait->buf_page_io_complete->ibuf_merge_or_delete_for_page->...->buf_flush_insert_into_flush_list
4. IO_BUF_NONE - io_handler_thread->fil_aio_wait->buf_page_io_complete (last steps)
5. n_pending_reads-- - io_handler_thread->fil_aio_wait->buf_page_io_complete (last steps)
```
@satya - at the point of last sweep, 1 & 2 will not happen anymore, as by the time we do the last flush, srv_master_do_idle_tasks has already waited for master thread to set m_master_ready_for_dd_shutdown and then we will set SRV_SHUTDOWN_PURGE and we will only reach the last flush if SRV_SHUTDOWN_PURGE.

so when we reach 3 / 4 / 5 the only possibility for `n_pending_reads` is to decrease, but never increase. 
Reading n_pend_reads - it's an atomic variable and their add/sub is done via atomic calls, however, at `buf_get_n_pending_reads_ios` we access them directly, on 5.7 that doesn't impose a memory barrier but on 8.0 it will as it's declared as std::atomic . That said, I think this is still ok, due to the fact that a dirty read here will always mean we are reading and old higher value, like (1 when it should be 0) and it will mean one extra loop at worse.

For flush list, we will only call it when buf_get_n_pending_reads_ios returns 0, eg.: all pending IO's have been completed and all necessary pages have been added to flush_list and properly marked as BUF_IO_NONE. At this stage, no new pages can be brought to BP due to SRV_SHUTDOWN_PURGE so reading the flush_list without a mutex is ok as we check if we have to do a new loop only once PC has already finished, meaning no changes to flush_list can be happening from any other thread.

altmannmarcelo: just realized I pinged the wrong Satya - cc @satya-bodapati

satya-bodapati: @altmannmarcelo right, once pending_iops is 0, a page is already added to the flush list just before the last sweep. 
The flush coordinator thread at the same moment can see the total number of dirty pages to be 0. this is because the other thread (incremented it) but the current thread still sees old value due to the missing read barrier?

So instead of seeing count as 1, page cleaner coordinator thread can see the cached value of flush list count as 0? :thinking: isn't it?

altmannmarcelo: > The flush coordinator thread at the same moment can see the total number of dirty pages to be 0. this is because the other thread (incremented it) but the current thread still sees old value due to the missing read barrier?

@satya-bodapati here the same thread (io thread) will be adding the page to flush list, merge all ibuf records to this age, then once it is completed, it will mark the page as IO_BUF_NONE (at this point PC will consider the page eligible for flushing) and only then decrement the number of pending reads.

IMHO for the purposes of the last PC sweep, we do not need to have the read of flush_list under mutex as pending reads will only be zero once the page is already eligible for flushing.

To be honest, here we are not under too much multithreading here - IO threads are only processing the async IO that have been previously dispatched to them, no new operations will land (only the ones resulting from completing this IO, such ibuf), user threads have already been closed at this point. So even adding the mutex back will not mean too much overhead.

satya-bodapati: I see.  Before  n_pdening_reads is decremented, the flush list count is incremented. and the PC sweep would always check pending_ops first and then the flust list length.

Looks OK I guess. Please put some additional sleeps and test the patch (without the mutex)

satya-bodapati (approved):
Please squash and then merge

PR #4752, opened 2022-05-05
PS-8155: Implement multiple LDAP server support

dutow:
This change adds support to a fallback LDAP server, which is used when
the primary server can't be reached.

To configure a fallback server, the LDAP authentication plugin adds two
new parameters, authentication_ldap_simple_fallback_server_host and
authentication_ldap_simple_fallback_server_port. These valueas are used
if the host parameter is non empty.

The change also improves the logging capabilities of the authentication
plugin by forwarding the LDAP library debug messages to the MySQL error
log. This feature requires setting authentication_ldap_simple_log_status
to its new maximum value, 6.

### DIFF
--- a/plugin/auth_ldap/include/log_client.h
+++ b/plugin/auth_ldap/include/log_client.h
@@ -33,6 +33,7 @@ namespace plugin {
 namespace auth_ldap {
 struct ldap_log_type {
   enum ldap_type {
+    LDAP_LOG_LDAP_DBG,
     LDAP_LOG_DBG,
     LDAP_LOG_INFO,
     LDAP_LOG_WARNING,
@@ -45,7 +46,8 @@ enum ldap_log_level {
   LDAP_LOG_LEVEL_ERROR,
   LDAP_LOG_LEVEL_ERROR_WARNING,
   LDAP_LOG_LEVEL_ERROR_WARNING_INFO,
-  LDAP_LOG_LEVEL_ALL
+  LDAP_LOG_LEVEL_MYSQL_DBG,
+  LDAP_LOG_LEVEL_LDAP_DBG
 };
 
 class Ldap_log_writer_error {
@@ -72,8 +74,13 @@ template <ldap_log_type::ldap_type type>
 void Ldap_logger::log(const std::string &msg) {
   std::ostringstream log_stream;
   switch (type) {
+    case ldap_log_type::LDAP_LOG_LDAP_DBG:
+      if (LDAP_LOG_LEVEL_LDAP_DBG > m_log_level) {
+        return;
+      }
+      break;
     case ldap_log_type::LDAP_LOG_DBG:
-      if (LDAP_LOG_LEVEL_ALL > m_log_level) {
+      if (LDAP_LOG_LEVEL_MYSQL_DBG > m_log_level) {
         return;
       }
       break;
--- a/plugin/auth_ldap/include/plugin_variables.h
+++ b/plugin/auth_ldap/include/plugin_variables.h
@@ -39,6 +39,8 @@ static int log_status;
 static unsigned int max_pool_size;
 static char *server_host;
 static unsigned int server_port;
+static char *fallback_server_host;
+static unsigned int fallback_server_port;
 static bool ssl;
 static bool tls;
 static char *user_search_attr;
@@ -113,7 +115,7 @@ static MYSQL_SYSVAR_UINT(max_pool_size, max_pool_size, PLUGIN_VAR_RQCMDARG,
 static MYSQL_SYSVAR_INT(log_status, log_status, PLUGIN_VAR_RQCMDARG,
                         "The logging level", nullptr /* check */,
                         &update_sysvar<unsigned int> /* update */,
-                        1 /* default */, 1 /*minimum */, 5 /* maximum */,
+                        1 /* default */, 1 /*minimum */, 6 /* maximum */,
                         0 /* blocksize */);
 static MYSQL_SYSVAR_STR(server_host, server_host,
                         PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_MEMALLOC |
@@ -126,7 +128,20 @@ static MYSQL_SYSVAR_UINT(server_port, server_port,
                          "The LDAP server TCP/IP port number",
                          nullptr /* check */,
                          &update_sysvar<unsigned int> /* update */,
-                         389 /* default */, 1 /*minimum */, 32376 /* maximum */,
+                         389 /* default */, 1 /*minimum */, 65535 /* maximum */,
+                         0 /* blocksize */);
+static MYSQL_SYSVAR_STR(fallback_server_host, fallback_server_host,
+                        PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_MEMALLOC |
+                            PLUGIN_VAR_READONLY,
+                        "The fallback LDAP server host", nullptr /* check */,
+                        &update_sysvar<char *> /* update */,
+                        nullptr /* default */);
+static MYSQL_SYSVAR_UINT(fallback_server_port, fallback_server_port,
+                         PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_READONLY,
+                         "The fallback LDAP server TCP/IP port number",
+                         nullptr /* check */,
+                         &update_sysvar<unsigned int> /* update */,
+                         0 /* default */, 0 /*minimum */, 65535 /* maximum */,
                          0 /* blocksize */);
 static MYSQL_SYSVAR_BOOL(
     ssl, ssl, PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_READONLY,
@@ -167,6 +182,8 @@ static SYS_VAR *mpaldap_sysvars[] = {MYSQL_SYSVAR(auth_method_name),
                                      MYSQL_SYSVAR(max_pool_size),
                                      MYSQL_SYSVAR(server_host),
                                      MYSQL_SYSVAR(server_port),
+                                     MYSQL_SYSVAR(fallback_server_host),
+                                     MYSQL_SYSVAR(fallback_server_port),
                                      MYSQL_SYSVAR(ssl),
                                      MYSQL_SYSVAR(tls),
                                      MYSQL_SYSVAR(user_search_attr),
--- a/plugin/auth_ldap/src/connection.cc
+++ b/plugin/auth_ldap/src/connection.cc
@@ -6,20 +6,87 @@
 
 #include "plugin/auth_ldap/include/plugin_log.h"
 
+namespace {
+// example of this callback is in the OpenLDAP's
+// servers/slapd/back-meta/bind.c (meta_back_default_urllist)
+int cb_urllist_proc(LDAP * /* ld */, LDAPURLDesc **urllist, LDAPURLDesc **url,
+                    void * /* params */) {
+  if (urllist == url) return LDAP_SUCCESS;
+
+  LDAPURLDesc **urltail;
+  for (urltail = &(*url)->lud_next; *urltail; urltail = &(*urltail)->lud_next)
+    /* count */;
+
+  // all failed hosts go to the end of list
+  *urltail = *urllist;
+  // succeeded host becomes first
+  *urllist = *url;
+  // mark end of list
+  *url = nullptr;
+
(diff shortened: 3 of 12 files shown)


plugin/auth_ldap/src/connection.cc
```diff
@@ -29,23 +96,29 @@ Connection::~Connection() {
...
   // ldap function use c_strs from these variables
   // changing them during a connect call could lead to a crash
   std::lock_guard<std::mutex> lock(conn_mutex_);
   ldap_host_ = ldap_host;
   ldap_port_ = ldap_port;
+  ldap_fallback_host_ = fallback_host;
+  ldap_fallback_port_ = fallback_port;
   use_ssl_ = use_ssl;
   use_tls_ = use_tls;
-  ca_path_ = ca_path;
```
altmannmarcelo, line 42: This is not been used anymore. Also remove from function signature.

plugin/auth_ldap/src/connection.cc
```diff
@@ -5,20 +5,87 @@
 
 #include "plugin/auth_ldap/include/plugin_log.h"
 
+namespace {
+// example of this callback is in the OpenLDAP's
+// servers/slapd/back-meta/bind.c (meta_back_default_urllist)
+int cb_urllist_proc(LDAP *ld, LDAPURLDesc **urllist, LDAPURLDesc **url,
```
altmannmarcelo: ld and params are not in use. Please adjust.

plugin/auth_ldap/include/log_client.h
```diff
@@ -33,6 +33,7 @@ namespace plugin {
 namespace auth_ldap {
 struct ldap_log_type {
   enum ldap_type {
+    LDAP_LOG_LDAP_DBG,
```
altmannmarcelo, line 36: this is counter intuitive, we normally start from less to high verbose. But seems like it was like this before.

plugin/auth_ldap/include/log_client.h
```diff
@@ -45,7 +46,8 @@ enum ldap_log_level {
   LDAP_LOG_LEVEL_ERROR,
   LDAP_LOG_LEVEL_ERROR_WARNING,
   LDAP_LOG_LEVEL_ERROR_WARNING_INFO,
-  LDAP_LOG_LEVEL_ALL
+  LDAP_LOG_LEVEL_SRV_DBG,
+  LDAP_LOG_LEVEL_LDAP_DBG
```
altmannmarcelo, line 50: naming here is confusing, `LDAP_LOG_LEVEL_SRV_DBG` here can mean LDAP SERVER or MYSQL SERVER.

plugin/auth_ldap/include/plugin_variables.h
```diff
@@ -39,6 +39,8 @@ static int log_status;
 static unsigned int max_pool_size;
 static char *server_host;
 static unsigned int server_port;
+static char *fallback_server_host;
```
altmannmarcelo, line 42: I know we followed the server_host, server_port pattern here. But I would rather use a comma list of server:port. The current proposed implementation allow for only a single LDAP replicate, while it's common to have more than one (like one on same DC, one in different DC).

plugin/auth_ldap/src/connection.cc
```diff
@@ -5,20 +5,87 @@
...
+                    void *params) {
+  if (urllist == url) return LDAP_SUCCESS;
+
+  LDAPURLDesc **urltail;
+  for (urltail = &(*url)->lud_next; *urltail; urltail = &(*urltail)->lud_next)
```
altmannmarcelo, line 17: here, in case of ldap_host been out, and come back, we will always perform fail-back which is not always desired, like when the old primary gets back and it's still not in sync. MySQL connecting to it might allow revoked user to connect and so on.
I would rather make this behavior configurable.

altmannmarcelo (changes requested):
few comments, plus new debug logging is crashing on plugin initialization:

```
Thread 1 "mysqld" received signal SIGSEGV, Segmentation fault.
0x0000555558b5948e in my_plugin_log_message (plugin_ptr=0x7fffe93d1428 <auth_ldap_simple_plugin_info>, level=MY_INFORMATION_LEVEL, format=0x7fffe93a80c0 "%s") at /work/ps/src/8.0/sql/log.cc:2829
2829	  snprintf(format2, sizeof(format2) - 1, "Plugin %.*s reported: '%s'",
(gdb) bt
#0  0x0000555558b5948e in my_plugin_log_message (plugin_ptr=0x7fffe93d1428 <auth_ldap_simple_plugin_info>, level=MY_INFORMATION_LEVEL, format=0x7fffe93a80c0 "%s") at /work/ps/src/8.0/sql/log.cc:2829
#1  0x00007fffe93628f4 in mysql::plugin::auth_ldap::Ldap_log_writer_error::write (this=0x555560414b40, level=mysql::plugin::auth_ldap::ldap_log_type::LDAP_LOG_DBG, data=": Ldap_logger initialized") at /work/ps/src/8.0/plugin/auth_ldap/src/log_client.cc:59
#2  0x00007fffe936354a in mysql::plugin::auth_ldap::Ldap_logger::log<(mysql::plugin::auth_ldap::ldap_log_type::ldap_type)1> (this=0x555560415200, msg="Ldap_logger initialized") at /work/ps/src/8.0/plugin/auth_ldap/include/log_client.h:108
#3  0x00007fffe93646a1 in auth_ldap_simple_init (plugin_info=0x55556043a438) at /work/ps/src/8.0/plugin/auth_ldap/src/plugin_simple.cc:97
#4  0x0000555558dc5ce0 in plugin_initialize (plugin=0x55556043a438) at /work/ps/src/8.0/sql/sql_plugin.cc:1309
#5  0x0000555558dc6305 in plugin_init_initialize_and_reap () at /work/ps/src/8.0/sql/sql_plugin.cc:1449
#6  0x0000555558dc7226 in plugin_register_dynamic_and_init_all (argc=0x55555daee2d0 <remaining_argc>, argv=0x55555e180de8, flags=0) at /work/ps/src/8.0/sql/sql_plugin.cc:1766
#7  0x0000555558b7973e in init_server_components () at /work/ps/src/8.0/sql/mysqld.cc:6522
#8  0x0000555558b7dc19 in mysqld_main (argc=17, argv=0x55555e180de8) at /work/ps/src/8.0/sql/mysqld.cc:7677
#9  0x00005555588cbecd in main (argc=2, argv=0x7fffffffd7b8) at /work/ps/src/8.0/sql/main.cc:25

(gdb) f 3
#3  0x00007fffe93646a1 in auth_ldap_simple_init (plugin_info=0x55556043a438) at /work/ps/src/8.0/plugin/auth_ldap/src/plugin_simple.cc:97
97	  log_srv_dbg("Ldap_logger initialized");
(gdb) f 0
#0  0x0000555558b5948e in my_plugin_log_message (plugin_ptr=0x7fffe93d1428 <auth_ldap_simple_plugin_info>, level=MY_INFORMATION_LEVEL, format=0x7fffe93a80c0 "%s") at /work/ps/src/8.0/sql/log.cc:2829
2829	  snprintf(format2, sizeof(format2) - 1, "Plugin %.*s reported: '%s'",
(gdb) p plugin
$5 = (st_plugin_int *) 0x0
(gdb)
```

plugin/auth_ldap/src/connection.cc
dutow: Even if we make it configurable, we are also using a connection pool. We either implement a mechanism that forces all of them to reconnect when we detect that the primary is back, or they remain connected to the current server.

Also, for your suggestion, we have two options:
* either leave the URL-list as is, making all connection attempts slower (considerably so, if the ldap port doesn't report a connection refused immediately)
* or implement a custom logic that tries to detect if the primary comes back alive, and modify the urllist then

These are doable, but I'm not sure if it's worth it, all examples I could find in opensource project used this same callback.

plugin/auth_ldap/include/plugin_variables.h
dutow: I agree, but the server_host and server_port variables are upstream, and we shouldn't change them, and I wanted to keep the fallbacks the same.

I can change it to "fallback_hosts", or just make so that the port is optional, and the user can specify multiple hosts in the host variable.

plugin/auth_ldap/src/connection.cc
dutow: I left this as-is for now, mostly because of what I've written above.

plugin/auth_ldap/include/plugin_variables.h
dutow: I left the two variables, but changed the default of port to 0. If it is set to 0, the fallback_host is appended to the connection string as is, which means it can contain multiple servers.

plugin/auth_ldap/include/log_client.h
dutow: Fixed.

plugin/auth_ldap/src/connection.cc
dutow: fixed

plugin/auth_ldap/src/connection.cc
dutow: fixed.

plugin/auth_ldap/src/connection.cc
altmannmarcelo: Ack, maybe we just need to document the behavior as this is normally something people want to know when we have fail over capabilities.

plugin/auth_ldap/include/plugin_variables.h
```diff
@@ -140,7 +140,7 @@ static MYSQL_SYSVAR_UINT(fallback_server_port, fallback_server_port,
                          "The fallback LDAP server TCP/IP port number",
                          nullptr /* check */,
                          &update_sysvar<unsigned int> /* update */,
-                         389 /* default */, 1 /*minimum */, 32376 /* maximum */,
+                         0 /* default */, 1 /*minimum */, 32376 /* maximum */,
```
altmannmarcelo: Default 0 minimal 1
also shouldn't the max be 65535 ?

plugin/auth_ldap/src/connection.cc
```diff
@@ -332,8 +332,19 @@ std::string Connection::get_ldap_uri() {
...
   if (!ldap_fallback_host_.empty()) {
-    str_stream << "," << (use_ssl_ ? "ldaps://" : "ldap://")
-               << ldap_fallback_host_ << ":" << ldap_fallback_port_;
+    // We allow 2 formats for fallback:
+    // * only fallback_host is specified (port is 0): in this case, we add it to
+    // the connection string as is
+    // * port is also specified: in this case we add the protocal prefix and
+    // port to it
+    str_stream << ",";
+    if (!ldap_fallback_port_ != 0) {
```
altmannmarcelo: this logic does not seems correct for multiple `ldap_fallback_host_` in case we have multiple hosts, only the first one will have ldap[s]:// prepend  to it?

Also, we don't have any validation in terms of if the user adds host with ldap:// already we will prepend it for a second time.

altmannmarcelo (changes requested):

plugin/auth_ldap/include/plugin_variables.h
altmannmarcelo: This seems a bit hacky IMHO, like if I have a port set it looks just the one address otherwise we can have multiple. I would rather:
1) remove fallback_server_port entirely
2) rename `fallback_server_host` to `fallback_server_hosts`
3) populate the list of fallback servers based on `fallback_server_hosts` only

plugin/auth_ldap/src/connection.cc
dutow: The idea is that we have two configuration options:

* if the user specifies ldap_fallback_port, we tret is as the normal setting, and use multiple variables to build the connection string
* if fallback_port is 0, but fallback_host is specifed, we assume that the user specified one or more complete connection strings in it, and use it as is.

This way the user can specify as many ldap servers as required, and can also mix SSL/TLS and other settings.

plugin/auth_ldap/include/plugin_variables.h
dutow: Fixed. I copy pasted the normal setting, which also used this value, I adjusted that too.

altmannmarcelo (approved):
LGTM on the technical side.
However I would rather have a single variable that allows for multiple server:port,server:port list of fallback servers.

PR #4773, opened 2022-05-16
PS-8222 Fix the Run in Docker section after disabling psdom extension (8.0)

alina-derkach-oaza:
modified:   source/installation/docker.rst


patrickbirch (approved):
Yes

PR #4781, opened 2022-05-16
PS-8204: Fix XML escape rules for audit plugin (5.7)

oleksandr-kachan:

There was a wrong length specified for some XML
escape rules. It lead to queries truncation in audit
log file.


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/561/

percona-ysorokin (approved):
LGTM

PR #4800, opened 2022-05-25
PS-8204: Fix XML escape rules for audit plugin (8.0)

oleksandr-kachan:

Added a test to check SQL text is not truncated in audit log
in case some special characters are present in the SQL.


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1228/

percona-ysorokin (approved):
LGTM

PR #4801, opened 2022-05-25
PS-8204: Fix XML escape rules for audit plugin

oleksandr-kachan:

There was a wrong length specified for some XML
escape rules. As a result of this terminating null symbol from
replacement rule was copied into resulting string. This lead to
quer text truncation in audit log file.
In addition added empty replacement rules for '\b' and 'f' symbols
which just remove them from resulting string. These symboles are
not supported in XML 1.0.

(cherry picked from commit c2e8878750c983446171c1b0c948029db7c566ff)


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/566/

percona-ysorokin (approved):
LGTM

PR #4804, opened 2022-05-29
Fixed PS-8018 (Parallel replication with blackhole table on replica throws error 1863)

i-rinat:
https://jira.percona.com/browse/PS-8018

Problem:

Applying DELETE commands on a replica causes an error that stops
the replication process if the involved table uses BLACKHOLE engine and
is subject to table scans for finding a particular row. In other words,
has no usable indexes.

When applying DELETEs, Partition_helper::ph_delete_row() checks that
the partition ID that was determined by partition info, is exactly the
partition ID from Partition_helper::m_last_part. Latter is set in
Partition_helper::ph_rnd_next() to an ID of the first partition that
reports having the requested row. Since BLACKHOLE engine stores nothing,
it searches nothing. On replicas it happily reports that it has any
requested row. Thus making Partition_helper::ph_rnd_next() believe that
partition 0 is the one that has the row. Partition_helper::m_last_part is
therefore always set to 0 when BLACKHOLE engine is in use. That's fine for
rows in the first partition, but causes HA_ERR_ROW_IN_WRONG_PARTITION for
other partitions.

Fix:

BLACKHOLE engine has no need to track specific rows, so the solution is
to disable the check in Partition_helper::ph_delete_row() by making
ha_blackhole::rpl_lookup_rows() always return false.

handler::rpl_lookup_rows() was introduced as a part of optimizations for
TokuDB and RocksDB, and with that introduced a warning that explicitly
mentions "TokuDB/RocksDB" in its text. Receiving such warning when
BLACKHOLE engine is used would be confusing. So the text of the message is
changed to be engine-agnostic.

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_blackhole_partitioned.test
+++ b/mysql-test/suite/rpl/t/rpl_blackhole_partitioned.test
@@ -1,0 +1,46 @@
+--source include/have_blackhole.inc
+--source include/have_innodb.inc
+--source include/have_partition.inc
+--source include/master-slave.inc
+
+# This testcase checks that the bug which caused replication into a partitioned
+# BLACKHOLE table to stop if any other than the first partition was involved,
+# is no longer a bug.
+
+call mtr.add_suppression(".*read free replication is disabled for table.*continue with rows lookup");
+
+--source include/rpl_connection_master.inc
+# Source table needs to have at least two partitions and no indexes.
+CREATE TABLE t1(id INT)
+    Engine=InnoDB PARTITION BY RANGE (id)
+    (PARTITION p1 VALUES LESS THAN (10),
+     PARTITION p2 VALUES LESS THAN MAXVALUE);
+
+--source include/sync_slave_sql_with_master.inc
+
+# Replicating into a BLACKHOLE table.
+ALTER TABLE t1 ENGINE=BLACKHOLE;
+
+--source include/rpl_connection_master.inc
+# All partitions need to have some data in them.
+INSERT INTO t1 (id) VALUES (1), (5), (9), (13), (17);
+
+# Delete operation needs to do something in the second partition. Without
+# indexes this command will use table scans to find all affected rows.
+DELETE FROM t1 WHERE id > 4 AND id < 15;
+
+# Now try to touch the second partition with an UPDATE.
+UPDATE t1 SET id = 18 WHERE id = 17;
+
+--source include/sync_slave_sql_with_master.inc
+
+# Replica should successfully process DELETEs and UPDATEs from a source
+# instance.
+--source include/check_slave_is_running.inc
+
+--source include/rpl_connection_master.inc
+DROP TABLE t1;
+
+--source include/sync_slave_sql_with_master.inc
+
+--source include/rpl_end.inc
--- a/mysql-test/suite/rpl/r/bug93917.result
+++ b/mysql-test/suite/rpl/r/bug93917.result
@@ -7,6 +7,7 @@ Warnings:
 Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
 Note	####	Storing MySQL user name or password information in the master info repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START SLAVE; see the 'START SLAVE Syntax' in the MySQL Manual for more information.
 [connection master]
+call mtr.add_suppression(".*read free replication is disabled for table.*continue with rows lookup");
 [connection slave]
 [connection master]
 ********************************************************************************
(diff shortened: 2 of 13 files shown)


kamil-holubicki (commented):
Hi, The fix looks fine.
Would it be possible to extend the test with UPDATE? Looking at the code it seems that it should suffer the same problem
```
  int update_row(const uchar *old_data, uchar *new_data) override
  {
    return Partition_helper::ph_update_row(old_data, new_data,
                                           rpl_lookup_rows());
  }
```

kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-pipeline/27834/

i-rinat:
> Would it be possible to extend the test with UPDATE?

Sure. Added an UPDATE to the testcase.

> Looking at the code it seems that it should suffer the same problem

I've tried just to be sure. And UPDATEs do trigger the same issue. The patch helps as `Partition_helper::ph_update_row()` contains similar code that skips the error when `lookup_rows` is `false`.

i-rinat:
> Would it be possible to extend the test with UPDATE?

Sure. Added an UPDATE to the testcase.

> Looking at the code it seems that it should suffer the same problem

I've tried just to be sure. And UPDATEs do trigger the same issue. The patch helps as `Partition_helper::ph_update_row()` contains similar code that skips the error when `lookup_rows` is `false`.

kamil-holubicki:
https://ps57.cd.percona.com/job/percona-server-5.7-pipeline/27854/

mysql-test/suite/rpl/t/rpl_blackhole_partitioned.test
```diff
@@ -0,0 +1,52 @@
...
+--source include/have_partition.inc
+--source include/master-slave.inc
+
+# This testcase checks that the bug which caused replication into a partitioned
+# BLACKHOLE table to stop if any other than the first partition was involved,
+# is no longer a bug.
+
+call mtr.add_suppression(".*read free replication is disabled for table.*continue with rows lookup");
+
+--connection master
```
venkatesh-prasad-v: Please use `--source include/rpl_connection_master.inc`

mysql-test/suite/rpl/t/rpl_blackhole_partitioned.test
```diff
@@ -0,0 +1,52 @@
...
+# Source table needs to have at least two partitions and no indexes.
+CREATE TABLE t1(id INT)
+    Engine=InnoDB PARTITION BY RANGE (id)
+    (PARTITION p1 VALUES LESS THAN (10),
+     PARTITION p2 VALUES LESS THAN MAXVALUE);
+
+--save_master_pos
+--connection slave
+--sync_with_master 0
```
venkatesh-prasad-v: Please source `--source include/sync_slave_sql_with_master.inc` to sync the replica server with the source.

mysql-test/suite/rpl/t/rpl_blackhole_partitioned.test
```diff
@@ -0,0 +1,52 @@
...
+# Delete operation needs to do something in the second partition. Without
+# indexes this command will use table scans to find all affected rows.
+DELETE FROM t1 WHERE id > 4 AND id < 15;
+
+# Now try to touch the second partition with an UPDATE.
+UPDATE t1 SET id = 18 WHERE id = 17;
+
+--save_master_pos
+--connection slave
+--sync_with_master 0
```
venkatesh-prasad-v: Likewise

venkatesh-prasad-v (changes requested):

mysql-test/suite/rpl/t/rpl_blackhole_partitioned.test
i-rinat: Replaced all `connection master` with `source include/rpl_connection_master.inc`. There were no `connection slave` outside of synchronization command bundles, so those were replaced with `source include/sync_slave_sql_with_master.inc`.

mysql-test/suite/rpl/t/rpl_blackhole_partitioned.test
i-rinat: Replaced with `source include/sync_slave_sql_with_master.inc`.

mysql-test/suite/rpl/t/rpl_blackhole_partitioned.test
i-rinat: Replaced with `source include/sync_slave_sql_with_master.inc`.

i-rinat:
Made changes according to the suggestions from the review.

kamil-holubicki (approved):
lgtm

venkatesh-prasad-v (approved):

PR #4806, opened 2022-05-29
release 8.0.29 21

ldonoso:
- Bump the version number in MYSQL_VERSION
- Bug #33407054: Hypergraph : Join on Strings
- WL#13458 Global and session memory allocation limits. additional fix for MTR mem_cnt_sql_keys.test to make it worked with --hypergraph option.
- Bug #33336806: Hypergraph Offload issue : Due to the absence of CAST_AS_DATE Bug #33388751: Hypergraph Offload : Incompatible function arguments Bug #33410654: Hypergraph Offload Issue :CAST Missing
- Bug #33378581: More preliminary fixes for WL #14488 [interesting_order, noclose]
- Bug #33296504 Use the hypergraph optimizer for subqueries in INSERT statements
- WL#13458 Global and session memory allocation limits. additional fix for MTR mem_cnt_sql_keys.test to make it worked with --hypergraph option.
- Bug #33378581: More preliminary fixes for WL #14488 [Unique cost, noclose]
- Bug #33426283: Report sub-zero row counts more accurately
- Bug#33364732: Preliminary fixes for WL#14673 [no_cache, noclose]
- 
- Bug #33424846 	problems with decimals... bin2decimal ... Assertion `intg <= from->intg' failed.
- Bug #33426283: Report sub-zero row counts more accurately
- Bug #33409812 Invalid paths found in RPATH for group_replication.so
- Bug #33431753 Add WITH_CURL=system to Fedora RPM builds
- Bug#33364732: Preliminary fixes for WL#14673 [DeleteRowsIterator, noclose]
- Bug#33364732: Preliminary fixes for WL#14673 [sql_safe_updates, noclose]
- 
- Bug #33367371 Add support for missing regexp features in bundled ICU library
- 
- 
- 
- 
- 
- 
- Bug #33424846   problems with decimals... bin2decimal ... Assertion `intg <= from->intg' failed.
- Bug #33385252 CHANGE 'UTF8' TO 'UTF8MB3' when reporting invalid strings
- Bug #33395007 CHANGE 'UTF8' TO 'UTF8MB3' for SQL SHOW commands
- Bug #33395007 CHANGE 'UTF8' TO 'UTF8MB3' for SQL SHOW commands
- 
- Bug#33449268 MTR: ndb_opt.spj_rqg result mismatch
- Bug #33395007 CHANGE 'UTF8' TO 'UTF8MB3' for SQL SHOW commands
- Bug #33442929 Fix msbuild Warning MSB8065: File not created for NDB
- Bug#33300078: New warnings with clang 13
- Bug #30624990 NO UTF8MB3 IN INFORMATION_SCHEMA.CHARACTER_SETS
- Bug#31731752: INCORRECT (STRONG) MDL LOCK IS ACQUIRED WITH MULTI TABLE UPDATE COMMAND
- 
- Bug #30624990 NO UTF8MB3 IN INFORMATION_SCHEMA.CHARACTER_SETS
- Bug #33290245 Item_func_st_distance_sphere::val_real(): Assertion `is_nullable()' failed.
- 
- Bug#33466577 Remove NDB compile time check workarounds
- Bug #33468892 CHANGE 'UTF8' TO 'UTF8MB3' for mismatched collation/charset warnings
- BUG#33213841: MySQL crashes while executing `RESET REPLICA ALL`
- 
- WL#14346 PERFORMANCE_SCHEMA, SECONDARY_ENGINE STATS
- Bug #33477822: Push join conditions from parameterized paths as low as possible [noclose]
- Bug #33415840 "main.range_icp_mrr" fails sporadically with result mismatch on pb2 daily trunk
- 
- 
- 
- 
- WL#14346 PERFORMANCE_SCHEMA, SECONDARY_ENGINE STATS
- Bug#33452674 join_pushdown_*.test: Use regex filtering to shorten IN-value lists in *.result
- Bug#33449195: Remove MTR tests join_pushdown_none and join_pushdown_nobnl
- Bug#33451256 AbstractQueryPlan (AQP) return incorrect access type to be used for a table.
- 
- Bug #33491886: Nested loop joins allow very expensive inner sides in the hypergraph optimizer
- Bug#33451256 AbstractQueryPlan (AQP) return incorrect access type to be used for a table.
- 
- 
- Bug#33492244 Repeated logging of 'cluster failure for epoch'
- 
- Bug#33449000 Need a warning when a pushed join was not executed after all
- Bug#33452674 join_pushdown_*.test: Use regex filtering to shorten IN-value lists in *.result
- Bug#33452674 join_pushdown_*.test: Use regex filtering to shorten IN-value lists in *.result
- 
- Bug #32302724: Wrong result when use "select all" Bug #32950670: Potentially wrong results
- 
- Bug #33506809: Duplicate edges in the hypergraph
- 
- BUG#30832382 NDB_IMPORT NEEDS A --TABLE OPTION
- BUG#30832382 NDB_IMPORT NEEDS A --TABLE OPTION
- 
- BUG#24935788 COPYING ALTER LEAVES EXTRANEOUS TEMPORARY TABLE AND FOREIGN KEYS
- WL#14663 - Shared unreachable destinations quarantine for routing plugins
- Bug #32302724: Wrong result when use "select all"
- WL#14663 - Shared unreachable destinations quarantine for routing plugins
- 
- Bug#33516925 - Doxygen issue in mysql_routing.h
- 
- Bug #33517637: Speed up microbenchmarks in debug mode
- BUG#24935788 COPYING ALTER LEAVES EXTRANEOUS TEMPORARY TABLE AND FOREIGN KEYS
- Bug#33522651 - Missing logging headers in routing_common_unreachable_destinations.cc
- Bug#33522379 - Failed assertion in IgnoreSpacesInDestinations
- Bug #33521622: Allow developers to force certain subplans in the hypergraph join optimizer
- Bug#33364732: Preliminary fixes for WL#14673 [force_sort_rowids]
- 
- Bug#14139386 DDL APPLY SET MYSQL.NDB_APPLY_STATUS EPOCH=0
- Bug#33466554 Prevent optimizer from generating range scans on unique hash indexes Bug#33474345 Avoid table REF-access in query plan, when handler does full table scan
- Bug#33364732: Preliminary fixes for WL#14673 [bit fields]
- 
- Bug#33364732: Preliminary fixes for WL#14673 [bit fields]
- 
- Bug #33531039: Stop slow microbenchmarks after calibration
- Bug#33395007: CHANGE 'UTF8' TO 'UTF8MB3' for SQL SHOW commands
- 
- 
- BUG#30881109 PERFSCHEMA TABLES ARE NOT COLLATION AWARE
- Bug#33530262 use cmakes GenerateExportHeader
- Bug#33530155 routing plugin does not unregister from routing component
- 
- Bug#33541083 remove unused include common.h
- Bug#33541091 replace duplicated scope-guard
- Bug#33541103 replace rename_thread() by my_thread_self_setname()
- WL#14526: Deprecate disconnect-slave-event-count and abort-slave-event-count
- Bug#19947842 ASSERTION `HOST_LENGTH <= 60' FAILED IN SET_SETUP_ACTOR_KEY
- 
- Bug #33395007 CHANGE 'UTF8' TO 'UTF8MB3' for SQL SHOW commands
- WL#13720 : Deprecate system variables query_prealloc_size and transaction_prealloc_size
- Bug #33395007 CHANGE 'UTF8' TO 'UTF8MB3' for SQL SHOW commands
- Bug #33546200: Allow parameterized hash tables in the hypergraph join optimizer
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Revert "Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]"
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug#33542456: Mutex_lock is not movable
- 
- 
- 
- Bug #33555774: clang-tidy fixes in hypergraph optimizer
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33554887: Disallow redundant parameterized access paths
- Bug#32983256 LOCAL_ROUTE_ORD should check path and destination balance
- Bug #32738705: "OUT OF SORT MEMORY ERROR" HAS AN INCONSISTENT RELATIONSHIP WITH THE BUFFER SIZE Bug #33501541: Unmanageable Sort Buffer Behavior in 8.0.20+
- Bug #32738705: "OUT OF SORT MEMORY ERROR" HAS AN INCONSISTENT RELATIONSHIP WITH THE BUFFER SIZE Bug #33501541: Unmanageable Sort Buffer Behavior in 8.0.20+
- 
- Bug #33560764: Make hypergraph optimizer trace somewhat clearer
- 
- Bug#33550031 Routing plugin does not reuse BasePluginConfig
- Bug #33362907 mysql crashes due to small thread_stack value
- 
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug#32604161 ADDS A FEATURE IN NDB_RESTORE TO RESTORE NDB_APPLY_STATUS TABLE
- Bug#33300587: mysqld crash with unknown parameter and enable create_admin_listener_thread
- 
- Bug#33189767 - Improve logging of GCS communication issues
- Bug#33189767 - Improve logging of GCS communication issues
- Bug#33575155 Remove unused buddy files
- Bug#33564303 make routing extern-const constexpr
- Bug#33564247 implicit-copy-constructor MySQLRoutingBase
- Bug#33142135: Assert for fields not marked as READ [cleanup]
- Bug#33142135: Follow-up [cleanup]
- Bug#33321787: Possible SIGSEGV in set_parse_error_message
- Bug #33521497 result of GROUP_CONCAT is inconsistent when group_concat_max_len becomes larger
- WL#14823 Fail on unknown config option
- Post push fix for WL#14526.
- Bug#33570004 decouple option-decoders from BasePluginConfig
- Bug#33570218 Column performance_schema.processlist.host is too short
- Bug #33569770: Stackless CreateIteratorFromAccessPath()
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33037007: PRELIMINARY FIXES FOR WL #14488 [renaming, noclose]
- BUG#29892252 TEMP TABLE IS NOT CLEANING UP AFTER ALTER TABLE OPS FAILED ..
- Bug#33583337: Hypergraph: main.subquery_all takes a long time due to not using index lookup [noclose]
- Bug #33569770: Stackless CreateIteratorFromAccessPath()
- 
- 33583572 event_source_name option not recognised by Router option validation
- Bug#33142135: Follow-up [cleanup]
- Bug #33568767 Upgrade to libedit 20210910-3.1 [patch 1]
- Bug #33568767 Upgrade to libedit 20210910-3.1 [patch 2]
- Bug #33568767 Upgrade to libedit 20210910-3.1 [patch 3]
- Bug #33568767 Upgrade to libedit 20210910-3.1 [patch 4]
- Bug #33574176: Compile MySQL with GCC 12 [noclose]
- Bug #33574176: Compile MySQL with GCC 12 [noclose]
- Bug#33574630 export router_lib symbols
- Bug#33574698 split metadata_cache into lib and plugin
- Bug#33574737 split routing plugin and library
- Bug#33575138 remove duplicate router test
- Bug#33570004 decouple option-decoders from BasePluginConfig [post-fix]
- Bug#33559796: Single-argument constructors should be explicit (patch #1)
- BUG#29913991: HEARTBEAT EVENT CAN ONLY ADDRESS 32 BIT FILE OFFSET
- BUG#29913991: HEARTBEAT EVENT CAN ONLY ADDRESS 32 BIT FILE OFFSET
- BUG#29892252 TEMP TABLE IS NOT CLEANING UP AFTE...
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug#33594328: Minor fix in Performance Schema Notification doxygen documentation
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Revert "Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]"
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33592812: Clarify the description of WalkAccessPaths()
- Bug #33594247 INFO_SRC should include information on cloud update release number
- Bug #33594208 	Doxygen CMake script does not parse correctly doxygen error file with '[' and ']
- BUG#32523769 TEST INNODB_S_GAP_DATA_LOCKS FAIL SPORADICALLY
- Bug#33596116 doxygen warnings about [[maybe_unused]]
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- WL #14806: Index range scans in the hypergraph optimizer [patch 1/7, tests]
- WL #14806: Index range scans in the hypergraph optimizer [patch 2/7, reverse]
- WL #14806: Index range scans in the hypergraph optimizer [patch 3/7, set_keyread]
- WL #14806: Index range scans in the hypergraph optimizer [patch 4/7, keyparts]
- WL #14806: Index range scans in the hypergraph optimizer [patch 5/7, inexact]
- WL #14806: Index range scans in the hypergraph optimizer [patch 6/7, unindent]
- WL #14806: Index range scans in the hypergraph optimizer [patch 7/7, code]
- Bug#33142135: Doxygen fixes
- BUG#33531967: server failure during group member recovery
- 
- 
- Bug#33599001 doc-comment warnings in xplugin
- Bug#33599026 doc-comment warnings in innobase
- Bug#33599068 doc-comment warnings in perfschema
- Bug#33599173 doc-comment warnings in group-replication
- Bug#33599219 doc-comment warnings in optimizer
- Bug#33599268 doc-comment warnings in log-sink-perfschema
- Bug#33599320 doc-comment warnings in replication
- Bug#33599363 doc-comment warnings in sql-auth-cache
- WL #14807: Index merge in the hypergraph optimizer
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33603493: Multiple equalities get lost in hypergraph join optimizer pushdown
- 
- Bug#33584940 Pb2 test failure in windows due to apply stored grants failure [noclose]
- Bug #33595484 CMake should check exact versions of required Doxygen, PlantUML, GraphViz
- BUG#28694376 : Don't run valgrind leak checking if a test crashes
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33604507 Fix clang -Wdocumentation warnings [noclose]
- Bug#33148961 FAILURE TO UPGRADE FROM 5.7, INVALID UTF8 CHARACTER STRING
- Bug#33605366 Error: option 'DEFAULT.max_total_connections' is not supported
- Bug #33609599: Hypergraph: Arguments to NeighborhoodCache are swapped
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- WL#14363:Support multiple stand-by servers for keyring_okv
- 
- 
- Raise version number after cloning 5.6.55
- Update License Book
- Bug#33615013 Fix Doxygen warnings reported by Doxygen v1.9.2
- 
- 
- 
- Bug #30799495 - NDB_IMPORT: "SETTING NON-NULLABLE ATTR TO NULL" FOR PK COLUMN WITH AUTO_INCREMENT
- Bug #33101844 CREATE FULLTEXT INDEX CRASHES SERVER FOR LARGE TABLE
- 
- Bug#33549751 Bug#33549751 Placement of REJECT_IF for correlated subqueries in Access Path
- 
- 
- Bug #30799495 - NDB_IMPORT: "SETTING NON-NULLABLE ATTR TO NULL" FOR PK COLUMN WITH AUTO_INCREMENT
- Bug#28128077 Part 6: Move require out from ndb_global.h to util/require.h [noclose]
- Bug#28128077 Part 7: Replace ATTRIBUTE_UNUSED with [[maybe_unused]] [noclose]
- Bug#28128077 Part 8: Move ATTRIBUTE_FORMAT and ATTRIBUTE_NOINLINE from ndb_global.h to ndb_compiler.h [noclose]
- Bug#32233543 [#1] Add cstrbuf for safe string copy.
- Bug#32233543 [#2] REPLACE STRNCPY AND REMOVE STRINGOP-TRUNCATION COMPILER WARNINGS
- Bug#32233543 [#3] REPLACE STRNCPY AND REMOVE STRINGOP-TRUNCATION COMPILER WARNINGS
- Bug#32233543 [#4] Remove suppression of -Wstringop-truncation warnings.
- WL#14867 Open NDB table with missing index in MySQL
- 
- Bug#33616560 compiler warning dynamic memcmp
- BUG#30881109 PERFSCHEMA TABLES ARE NOT COLLATION AWARE
- Bug#33339129 : Remove the redundant code.
- 
- Bug#33311645 Fuzzy costs in the hypergraph join optimizer
- Bug#33514253: Ctrl-C to interrupt query results in               password prompt and reauth in CLI for               8.0.27
- Bug #31334762: INVALID-ENUM-VALUE IN READ_OK_EX RB#27415
- Bug#33594652 FIXES TIMEOUT FAILURE OF NDB_RESTORE_CONV_PK_NEGATIVE TEST
- 
- Bug#33300271: EXPLAIN + CTE + temporary table leads to crash mysqld
- Bug#33589961 Invalid numeric values such as '10 "M12L"' should not be allowed in config.ini
- Bug#33300235 cs fails to rejoin the old primary cluster back to cs after forced failover
- Bug #33571661 InnoDB: Enable level 4 warnings on VC++: Enable warning C4389 and C4245
- WL#14867 Open NDB table with missing index
- Bug#28128077 Fixup: Part 6: Move require out from ndb_global.h to util/require.h [noclose]
- Bug#32233543 Fixup: [#1] Add cstrbuf for safe string copy.
- 
- Bug#33506241  sql produces different resultsets when different internel storage engines r used
- Bug#33614908 perfschema.processlist_acl fails on daily-trunk
- 
- BUG#33559219 Crash in pthread_cond_signal on bus error on Mac OS X ARM
- 
- Bug #33571661 InnoDB: Enable level 4 warnings on VC++: Enable warning C4389 and C4245 [post-push fix]
- Bug #33608154 libmysqlclient C++20 compilation fails due to reserved 'requires' keyword
- 
- Bug #33571661 InnoDB: Enable level 4 warnings on VC++: Enable warning C4389 and C4245 [post-push fix]
- 
- 
- Bug#33507416 MTR test ndbinfo_plans need to specify result ordering
- WL #14806: Index range scans in the hypergraph optimizer
- 
- Bug #33608154 libmysqlclient C++20 compilation fails due to reserved 'requires' keyword
- 
- Bug#33624554: main.merge_myisam fails with --hypergraph
- BUG #33588734   Starting data node with AutomaticThreadConfig on fails with less than 8 cpu
- 
- BUG#33625488 COMPONENT SERVICES HEADERS: MOVE *_BITS TO SEPARATE SUBDIR
- 
- Bug#33621792 Export additional symbols from router_lib
- Bug#33529743 The component tests are very slow on Windows
- 
- Bug#33584940 test failure due to apply stored grants failure
- WL#14867 Open NDB table with missing index
- 
- Bug#33615893: EXPLAIN statement crashes in get_fields_used()
- Bug#33615893: EXPLAIN statement crashes in get_fields_used()
- Bug#33305617: Assertion `table_independent_conds == nullptr' failed.
- Bug#33630661 export access-rights API
- Bug#33630320 extra-semi warnings when building metadata_cache_export_all
- Bug#33630957 remove duplicated delete_file
- Bug#33631207 decouple routing from plugin-env
- Bug#31468581 - INVALID PID IN UNIX SOCKET LOCK FILE
- Bug#33619511: MySQL server exits in debug build when -p parameter used
- 
- Bug#33619511: MySQL server exits in debug build when -p parameter used
- 
- Bug#33616957: Parser error for TIME values in binary protocol
- Bug#33615893: EXPLAIN statement crashes in get_fields_used()
- Bug#33616957: Parser error for TIME values in binary protocol
- 
- 
- Bug#31468581 - INVALID PID IN UNIX SOCKET LOCK FILE
- Bug#31468581 - INVALID PID IN UNIX SOCKET LOCK FILE
- Bug#33635459 remove reference to PluginFuncEnv from shared quarantine
- Bug #33635092 Remove unused file mysqlx_charset.cc in the X plugin
- 
- Bug #33640145 WL#14684 Setting --DEFAULT.* disregards other cmdline config changes
- Bug #32749286 BUILDING ON MAC OS WITH SSL INCORRECTLY COPIES SSL LIBRARIES
- Bug #33624777 sum(least(a, NULL)) result error
- Bug #33624777 sum(least(a, NULL)) result error
- 
- Bug#33438883: Assertion `!thd->lex->is_view_context_analysis()
- Bug#33499071: 'keyread'(only) broken for index-merge access method
- Bug#33619511: MySQL server exits in debug build when -p parameter used
- 
- Bug#33619511: MySQL server exits in debug build when -p parameter used
- BUG#33635088 ATRT finding test case fails on blank lines after comment
- Bug#33578113: DROP privilege on performance_schema.* can't be revoked
- Bug#33578113: DROP privilege on performance_schema.* can't be revoked
- BUG#32327411 : OFFLINE KEYRING MIGRATION FROM KEYRING_FILE REPORTS SUCCESS, BUT MIGRATES NO KEYS
- BUG#33378364: group_replication.gr_perfschema_group_communication_information
- BUG#33615493: Assert failure `local_member_info->get_recovery_status() == MEMBER_OFFLINE`
- 
- Bug#33399379 : Missing PREFIX BLOB in calculating maximum length of                possible row
- Updated copyright year in user visible text
- Update License Book
- Raise version number after tagging 7.4.35
- Updated copyright year in user visible text
- Update License Book
- Raise version number after tagging 7.5.25
- Updated copyright year in user visible text
- Update License Book
- Raise version number after tagging 7.6.21
- Bug#33593833 - Race condition related to sockets container on shutdown
- 
- Bug #33619406: Fix fido client authentication client plugin documentation.
- Bug #33631144: Unable to log in using authentication_fido plugin in mysql client
- Bug #33619406: Fix fido client authentication client plugin documentation.
- Bug#33551163 Bootstraping router with InnoDB Cluster Set returns confusing error messages
- 
- WL#14700: Disallow using temp tables in XA transactions in 8.0 Bug#31599926: XA COMMIT FAILED WHEN STOP GROUP_REPLICATION WILL LEAD NODE ERROR
- Bug#33353637 UPGRADE BOOST TO VER. 1.77.0
- Bug#33652853: Single-argument constructors should be explicit
- Bug #33640631 No LARGE-PAGE support on FreeBSD
- Bug#33499071: 'keyread'(only) broken for index-merge access method
- 
- Bug #30799495 - NDB_IMPORT: "SETTING NON-NULLABLE ATTR TO NULL" FOR PK COLUM N WITH AUTO_INCREMENT
- Bug #32006725 - NDB_IMPORT FAILS TO PARSE A CSV FILE WITH WINDOWS STYLE LINE FEED
- Bug#33551163 Bootstraping router with InnoDB Cluster Set returns confusing error messages
- 
- 
- BUG#33581529: Backport Fix for Bug 29223844 to 5.7 version of Group Replication
- Bug#33649009: Inconsistency between protocols when writing values to DATE colmns
- Bug#33054071 - segmentation fault "mysql_prepare_create_table"
- Bug#33616019: Hypergraph: Assertion `static_cast<size_t>(bit_num) < x.capacity()' failed
- Bug#33653934: Assert in main.index_merge_myisam - 32-bit, --hypergraph
- 
- 
- Bug#33630835 Failed assertion while iterating events
- WL#14680 - Removing #defines from InnoDB
- BUG#33653824 AQP mismatched parameters dbug_dump
- Bug#33643512 Missing check for Thd_ndb in information_schema.partitions
- 
- Bug#33619811: Hypergraph: Assertion `nullptr != dynamic_cast<Target>(arg)' failed
- Bug#33611740: Hypergraph optimizer loses conditions in pushdown
- 
- Bug#33630225: Bad caching of converted constants in NULL-safe comparison
- Bug#31947466: ASSERTION FAILURE WITH GROUP_CONCAT(... ORDER BY) AND PREPARED STATEMENTS
- Bug#31848191:ITEM_SUM_STD AND _VARIANCE HAS INCORRECT IS_NULL()
- Bug#33649009: Inconsistency between protocols when writing values to DATE colmns
- BUG#30042589 MYSQLDUMP LEAKS MEMORY WHEN SELECTED TABLES ARE DUMPED WITH --ORDER-BY-PRIMARY
- WL#14670: Privilege to disable audit abort function RB#27363
- Bug#33538106: There is a problem when calling the system table in stored function
- BUG #32474988  TRANSACTION MEMORY PROBLEM
- 
- Bug #33661024 Validation of Array Index
- 
- BUG #32474988  TRANSACTION MEMORY PROBLEM
- Bug#33647601 A TransientPool that grows above 256MB will fail shrinking
- Bug #33659435 Bundled libevent: EVENT__HAVE_WORKING_KQUEUE not defined with Xcode 12 or later
- 
- 
- Bug #33659435 Bundled libevent: EVENT__HAVE_WORKING_KQUEUE not defined with Xcode 12 or later
- Bug#33635076 - mysqld crash at ddl::Merge_file_sort::Output_file::duplicate_check
- BUG #33401830  Missing return
- Bug #33661024 Validation of Array Index
- Bug#33649169 MERGE_LIBRARIES_SHARED() exports builtin_perfschema_plugin for all libs
- Bug#33603522 ODR violation around mysqlxmessages lib
- 
- Bug#33671403 WL14864 Setting --DEFAULT.* does not always propagate command line overrides
- Bug #33661024 Validation of Array Index
- Bug #33603036 	InnoDB: Fix doxygen errors, enable doxygen for .ic and .i files
- Bug#33459671 GCC 11 build failure: -Werror=array-bounds
- Bug#33647601 A TransientPool that grows above 256MB will fail shrinking
- Bug #33594208 	Doxygen CMake script does not parse correctly doxygen error file with '[' and ']
- Bug #33603036 	InnoDB: Fix doxygen errors, enable doxygen for .ic and .i files
- 
- WL#14444 Ndb replication : Support replica_parallel_workers=1
- Bug#33044886: STOP GROUP REPLICATION SPORADICALLY STUCK FOR EVER  AFTER REBOOT
- Bug#33334507 - 5.7 - utf8mb4 + REPLACE(UUID(),'-','') returns duplicate UUIDs!
- Bug #32749286 BUILDING ON MAC OS WITH SSL INCORRECTLY COPIES SSL LIBRARIES
- Bug #33668341 MySQL 5.7: Look for openssl@1.1 on Mac
- Bug #33659435 Bundled libevent: EVENT__HAVE_WORKING_KQUEUE not defined with Xcode 12 or later
- BUG#33589359 Fix PB2 release build issues on Apple ARM
- 
- Bug#33499071: 'keyread'(only) broken for index-merge access method
- Bug #33037007: PRELIMINARY FIXES FOR WL #14488 [renaming, noclose]
- WL#14680 - Removing #defines from InnoDB [post-push fix]
- Bug#33661337: more gcols corruption
- 
- Bug #33059387 TRX_RW_MIN_TRX_ID() CHECK IN LOCK_SEC_REC_READ_CHECK_AND_LOCK() IS REDUNDANT
- WL#14444 post-push fix. Support replica_parallel_workers=1.
- Bug#33635076 mysqld crash at ddl::Merge_file_sort::Output_file::duplicate_check post-push fix forced innodb.ddl_large_record test to use 16k page size.
- Bug#33663811 Ship standard debuginfo package for SLES|openSUSE RPM packages
- 
- 
- Bug #33664929 Reduce size of RelWithDebInfo binaries
- Bug#33661337: more gcols corruption
- 
- Bug#33399379 : Missing PREFIX BLOB in calculating maximum length of                possible row
- 
- Bug #33664929 Reduce size of RelWithDebInfo binaries
- WL#13899 : INSTANT DROP (and ADD) COLUMN
- Bug#33661337: more gcols corruption
- Bug#33405696 btr_insert_into_right_sibling() leads to ACID violation
- WL#14670: Privilege to disable audit abort function
- Bug#33635637: Regression in FIND_IN_SET() in 8.0.22
- Bug #33688395 macos12 ARM warning: 'SIZE_MAX' macro redefined [-Wmacro-redefined]
- Bug#33689707 update create-innodb-cluster URL
- 
- WL#14930: server option to enable/disable auto purging of binlog files
- 
- Bug#33394701: crash at JOIN::optimize_keyuse()
- WL#13899 : INSTANT DROP (and ADD) COLUMN
- 
- Bug#33664787: hypergraph: Incorrect cost calculation for implicitly grouped query with LIMIT
- Bug #33650498: Mixing PS and non-PS syntax breaks --fido-register-factor
- 
- 
- Bug#33619350: Hypergraph: Assertion `it != m_access_paths.end()' failed.
- BUG#33559219 Crash in pthread_cond_signal on bus error on Mac OS X ARM
- Bug#33639644: sig11 in CreateIteratorFromAccessPath | join_optimizer/access_path.cc
- Bug#28805764 STABILIZE THE 8.0 RELEASE TEST [noclose]
- Bug #33694353 macos12 ARM warning: ld: warning: direct access in function
- Bug #30624990 NO UTF8MB3 IN INFORMATION_SCHEMA.CHARACTER_SETS
- 
- 
- Bug #33516707 Range check causing high memory consumption in memory/sql/THD::main_mem_root
- 
- 
- WL#14592: Refactor system variable processing in parser and executor
- 
- Bug#33611112: Reverse index scan is not sorted on implicit InnoDB primary key
- Bug #32019119 SCHEMA DIST SETUP WARNINGS
- Bug #33073354 TOO FEW CHUNK FILES ARE CREATED FOR HYBRID HASH JOIN
- Bug#29520374 CLEAN UP THE VALGRIND TESTING [noclose]
- WL#14592: Refactor system variable processing in parser and executor
- 
- Bug #33675809 	Bump required Doxygen version to 1.9.2, remove all workarounds for earlier ver
- Bug #33571677 	InnoDB: Enable level 4 warnings on VC++: Enable warning C4100
- Bug#33611545 - regression: 8.0.27+ crash in heap_rkey (MEMORY) or number_of_indexes (TempTable)
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- Bug #33571677 	InnoDB: Enable level 4 warnings on VC++: Enable warning C4100 [post-push fix]
- Bug #33473759 InnoDB: Enable level 4 warnings on VC++: Enable C4127
- 
- 
- BUG #33459653 : coredump after add index for partion table with ibd file missing Problem: - If we delete .ibd file of a partition along with the 1st partition   then after restart when we open the partition table we get the   error message, which is the expected behaviour. - When we delete .ibd file of a partition other than the 1st   partition then after restart when we open the partition table we   hit the debug assert because ha_innopart::open() checks for   ibd_file_missing only for the 1st partition.
- Bug #33473759 InnoDB: Enable level 4 warnings on VC++: Enable C4127 [post-push fix]
- Bug #33473759 InnoDB: Enable level 4 warnings on VC++: Enable C4127 [post-push fix]
- Bug#29520374 CLEAN UP THE VALGRIND TESTING [noclose]
- Updated copyright year to 2022.
- Bug#33690501 ODR violations when building against mysqlxclient
- 
- 
- Bug#29520374 CLEAN UP THE VALGRIND TESTING [noclose]
- WL#14978 Decrease connection timeout to the endpoint servers
- WL#14585: Extend and improve group_replication_set_as_primary behavior
- WL#14673: Enable the hypergraph optimizer for DELETE [patch 1/2, multi-table]
- WL#14673: Enable the hypergraph optimizer for DELETE [patch 2/2, single-table]
- Bug #33635120 Rename character set names from 'utf8' to 'utf8mb3'
- WL#11968 NdbDictionary in ndbinfo.  Part 1 : Preliminary work
- 
- 
- Bug#33676070 (patch ver. 2) Various issues with Dictionary::List
- Post push fix
- WL#13730 - Condition pushdown to materialized derived tables            with set operations
- POST PUSH MTR failure FIX for WL#13899
- Bug#31946448:OPTIMIZE SUBQUERY CAUSES A CORE DUMP
- Bug #33406728 Item_charset_conversion::compute_max_char_length
- WL#14937: disable multi-threaded myisam repair in 8.0 RB#27440
- WL#14937: deprecate myisam_repair_threads the 5.7 version RB#27440
- WL#13730 - Condition pushdown to materialized derived tables            with set operations
- BUG#31769352 FIX WINDOWS COMPILER WARNINGS IN SQL_MAIN
- 
- Bug#33717719 test-fail if port 3306 is in use
- Bug#33717815 test-fail if called without slash
- 
- Bug#33725248 doc-comment for purge_source_logs_before_date is out-dated
- 
- 
- 
- Bug#33727551 ndbxfrm error gave wrong file name
- Bug#33727564 Remove superfluous ndbabort, require, and, return from AsyncFile
- Bug#33727629 Misc fixes for ndb_file on Windows
- Bug#33727647 Win32AsyncFile treat both ERROR_FILE_NOT_FOUND and ERROR_PATH_NOT_FOUND equal.
- Bug#33727669 Introduce ndb_openssl_evp::MAX_SALT_COUNT and ndbxfrm_progress enum
- Bug#33727678 Unittest ndb_zlib-t is confusing
- WL#13601 Deprecate arbitrary delimiters and whitespaces in time/date/datetime/timestamp literals
- 
- WL #14335: Graph simplification for hypergraph join optimizer [patch 1/11, DPhyp bug]
- WL #14335: Graph simplification for hypergraph join optimizer [patch 2/11, movable]
- WL #14335: Graph simplification for hypergraph join optimizer [patch 3/11, move TrivialReceiver]
- WL #14335: Graph simplification for hypergraph join optimizer [patch 4/11, FindOutputRowsForJoin]
- WL #14335: Graph simplification for hypergraph join optimizer [patch 5/11, base code]
- WL #14335: Graph simplification for hypergraph join optimizer [patch 6/11, PQ mark]
- WL #14335: Graph simplification for hypergraph join optimizer [patch 7/11, priority queue]
- WL #14335: Graph simplification for hypergraph join optimizer [patch 8/11, cycle encoding]
- WL #14335: Graph simplification for hypergraph join optimizer [patch 9/11, GetCardinality]
- WL #14335: Graph simplification for hypergraph join optimizer [patch 10/11, optimize ModifyEdge]
- WL #14335: Graph simplification for hypergraph join optimizer [patch 11/11, FindOutputRowsForJoin]
- Bug#33620355: WL#14335: SIGSEV at (anonymous namespace)::CombiningWouldViolateConflictRules
- WL#13601 Deprecate arbitrary delimiters and whitespaces in time/date/datetime/timestamp literals
- Bug#32774799: ASSERTION `ITEM->TYPE() != ITEM::FIELD_ITEM ||               STATIC_CAST<CONST ITEM_FIELD *>(ITEM)->FIELD->               IS_HIDDEN_BY_SYSTEM()'FAILED.
- 
- Bug #33612437 InnoDB: Remove all tabs from the code
- Bug #33612437 InnoDB: Remove all tabs from the code
- 
- Bug #33715446 Contribution: rpcgen.cmake: always build xdr_sizeof.c on macOS
- Bug #33709490 Contribution: ssl.cmake: Fix spelling in comment
- WL#13601 Deprecate arbitrary delimiters and whitespaces in time/date/datetime/timestamp literals
- WL#14370, add new test cases
- WL#14370, Reject Multi-range-read operations as 'child pushable'
- WL#14370, allow AT_MULTI_UNIQUE_KEY to be pushed
- WL#14370, Use '!m_scan_operations.contain(tab_no)' to detect lookup_operation
- WL#14370, preparation patch: Be explicit about allocating memory from 'thd->mem_root'
- WL#14370, Replace AQP::Equal_set_iterator with standard C++ iterator.
- WL#14370, Constify *TABLE returned from get_table()
- WL#14370, don't inline, leave that to optimizer.
- WL#14370, Remove Table_access::dbug_print().
- WL#14370, standardize wording in EXPLAIN_NO_PUSH messages.
- WL#14370, avoid using current_thd when having a THD member variable
- WL#14370, Make 'm_has_pending_cond' usage more consistent
- WL#14370, Prepare for AQP interface changes
- WL#14370, introduce a 'nest_stack'
- WL#14370, fix get_last_sj_inner(), and align with comming changes.
- WL#14370, Prepare for semi-join nest to be 'nested' as well.
- WL#14370, adapt handler interface for doing join pushdown.
- WL#14370, Move join-pushdown preparation code into ha_ndbcluster_push.cc
- WL#14370, Integrate condition push as well into ndbcluster_push_to_engine()
- WL#14370, prepare join_pushdown MTR test for AccessPath usage.
- WL#14370, introduce GetBasicTable(AccessPath* path)
- WL#14370, Introduce AccessPath based AQP
- WL#14370, fix spj_rqg.test which started failing after previous patch.
- wl#14370, Remove obsolete code for handling FOUND_MATCH-triggers.
- WL#14370, remove obsolete code for setting 'spill_to_disk = false'
- WL#14370, re-record MTR test results.
- WL#14370, use 'query_scope' bitmap to calculate set of 'const_expr_tables'.
- wl#14370, Use 'query_scope' to set up m_const_scope
- WL#14370, Remove last usage of uses_join_cache().
- WL#14370, replace AQP::is_sj_firstmatch() with ::is_semi_joined().
- WL#14370, fix, don't allow field substitute into semi-join nests.
- WL#14370, fix SPJ API to handle semi- / outer-join combination
- WL#14370, replace is_antijoin() with is_anti_joined(ancestor).
- WL#14370, cleanup, align implementation and variable names
- WL#14370, remove usage of filesort_before_join()
- WL#14370, Improve EXPLAIN_NO_PUSH warning message
- WL#14370, Enable join-pushdown for Hypergraph optimizer.
- WL#14370, remove assert added during development.
- WL#14370 Fix array out of bound problem
- WL#14370, Fix failing ndbcluster.pushed_join.test
- WL#14370, fix failing ndbcluster.fk.test
- WL#14370, Fix linker problem.
- WL#14370, fix failing ndb_binlog_format.test and rpl_ndb_row_basic.test
- Bug#33734011 doc-comment warnings in innobase
- WL#14285 [01] Add OM_READ_FORWARD to FsOpenReq.
- WL#14285 [02] Debug logging of failed ndbfs requests
- WL#14285 [03] Always fill in whole FsOpenReq.
- WL#14285 [04] Remove use of DUMMY_PASSWORD.
- WL#14285 [05] ndb_zlib, ndb_openssl_evp: Multi threaded I/O
- WL#14285 [06] Support compression with XTS-encryption.
- WL#14285 [07] ndb_ndbxfrm1: Add descriptions and methods.
- WL#14285 [08] ndb_az31: minor change adding const.
- WL#14285 [09] ndbxfrm_buffer: Skip trailer on read.
- WL#14285 [10] ndbxfrm_file: class for accessing backup and ndbfs files.
- WL#14285 [11] ndbxfrm, ndb_restore, ndb_print_backup_file: Use ndbxfrm_file.
- WL#14285 [12] AsyncIoThread: Store a ndb_open_ssl operation instance per thread.
- WL#14285 [13] AsyncFile: Release file buffer if only used for init, like for table space files.
- WL#14285 [14] AsyncFile use ndbxfrm_file
- WL#14285 [15] Make Ndbfs also aware of sparse init.
- WL#14285 [16] NDFBS: Init of encrypted TS needs at least two pages.
- WL#14285 [17] AsyncFile: Handle encryption of sparse file
- WL#14285 [18] AsyncFile: PBKDF2 was too slow for TDE - reduce kdf iter count to 1.
- WL#14285 [19] Remove ndbzio - ndb_zlib is used instead.
- WL#14285 [20] remove ndbxfrm_readfile, ndbxfrm_writefile
- WL#14285 [21] Option [DB]EncryptedFileSystem
- WL#14285 [22] Prepare defines for hardcoded testing of TDE.
- WL#14285 [23] Improve code returning read size via Ndbfs
- 
- Bug #33712370 MySQL 8.0.23 SLAVE HUNG AFTER SEMAPHORE WAIT
- 
- 
- Bug#33734206 doc-comment warning in sql/
- Bug#33734075 doc-comment warnings in authentication_ldap
- Bug#33734104 doc-comment warnings in sql_mfa
- 
- 
- 
- Bug#33734181 doc-comment warning in router
- Bug#33734188 broken doc-comment in xplugin
- Bug#33734129 doc-comment warnings in my_compress.cc
- Bug#33734062 doc-comment warnings in libchangestreams
- 
- Bug #33732729 Clean up sql_const.h
- Bug#33655246 - Cleanup shared quarantine on shutdown issue
- WL#14370: Enable join-pushdown for Hypergraph optimizer.
- 
- WL#14285 Fixup [10] ndbxfrm_file: fix bus error on sparc
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug #33550360: Inconsistent row estimations in the hypergraph optimizer [noclose]
- Bug#33725507: A bug in sql_lex.cc cause mysql-server to fail
- WL#13469: secure storage for sensitive system variables
- WL#14370, post-push fix
- 
- Bug#28805764: STABILIZE THE 8.0 RELEASE TEST [noclose]
- 
- 
- WL#11968 Patch 1(b), post-push fixes for patch #1.
- WL13954: Deprecate replica_parallel_type
- Bug#33743068: GraphSimplification unit tests are leaking memory
- Bug#33738447: Assertion `!(used_tabs & (~ read_tables & ~filter_for_table))' failed.
- 
- Bug #33045204 EQUI-HEIGHT HISTOGRAM ALWAYS ESTIMATES NON-INTEGRAL EQUALITY PREDICATE TO 1/ROWS [patch 1/2, selectivity fix]
- Bug #33045204 EQUI-HEIGHT HISTOGRAM ALWAYS ESTIMATES NON-INTEGRAL EQUALITY PREDICATE TO 1/ROWS [patch 2/2, efficiency fix]
- Bug#31477725 NDB_RESTART FAILS SPORADICALLY IN PB2
- Bug #33488047: Parser generated by Bison 3.8.1+ doesn't compile
- 
- Bug #33747036: Make the THD parameter to table_access_factory_v1::create   accept NULL as valid RB#27558
- 
- 
- WL#14633: Add support for jemalloc on Windows
- WL#14633: Add support for jemalloc on Windows
- Bug #33732729 Clean up sql_const.h
- Bug #33748221 Look for homebrew bison on macOS
- Bug #33748273 Silence warnings in code generated by homebrew protoc
- Bug #32749286 BUILDING ON MAC OS WITH SSL INCORRECTLY COPIES SSL LIBRARIES
- Bug#33729162: Hypergraph misses rows for SELECT DISTINCT ... GROUP BY WITH ROLLUP
- 
- 
- 
- 
- Bug#33165726 Fix freed memory access in performance schema table
- 
- WL#11968 NdbDictionary in ndbinfo. Part 2: Implementation
- WL#11968 NdbDictionary in ndbinfo. Part 3: tests
- 
- Bug#33714885 - LOAD DATA INFILE fails with ERROR 1054 : Unknown column XXXX in 'where clause'
- 
- WL#15000 Add ndbcluster plugin threads to P_S
- 
- BUG#27857780 PB2 PERFSCHEMA.START_SERVER_NO_MDL UNSTABLE AND FAILS SPORADICALLY
- 
- Bug#33491183: Wrong query result using select Union All Bug#33503357: comparing a `bigint` value with an identical `bigint unsigned` value fails Bug#33634109: hypergraph: Unstable results from TPC-H Q2 Bug#33674441: Wrong result with outer join and streaming aggregation
- WL#13075: Support TLS session reuse in the C API RB#27093
- BUG#33165693 Test case fix: Fix main.ssl-big
- WL#14722: Support IF NOT EXISTS clause in CREATE PROCEDURE/FUNCTION/TRIGGER
- Bug#32695670 ASSERTION 'M_CURRENT_ROLLUP_LEVEL >= 0 && M_CURRENT_ROLLUP_LEVEL < M_NUM_LEVELS'
- Adjust MYSQL_VERSION and restore .gitreview
- WL#13730: Condition pushdown to materialized derived tables having unions
- Bug#33491183: Wrong query result using select Union All
- Bug#33457946: main.index_merge_innodb fails sporadically on pb2 daily trunk
- WL#14633: Add support for jemalloc on Windows
- 
- 
- BUG#33362935 NDBD: redundant connect-retries entry in help message
- 
- 
- 
- WL#12771 reuse connection [01/26] - classic::auth, classic::quit
- WL#12771 reuse connection [02/26] - classic::query
- WL#12771 reuse connection [03/26] - classic::ping
- WL#12771 reuse connection [04/26] - classic::init-schema
- WL#12771 reuse connection [05/26] - classic::reset-connection
- WL#12771 reuse connection [06/26] - classic::kill
- WL#12771 reuse connection [07/26] - classic::reload
- WL#12771 reuse connection [08/26] - classic::statistics
- WL#12771 reuse connection [09/26] - classic::change-user
- WL#12771 reuse connection [10/26] - classic::list-fields
- WL#12771 reuse connection [11/26] - classic::stmt-*
- WL#12771 reuse connection [12/26] - classic::clone protocol
- WL#12771 reuse connection [13/26] - classic::clone
- WL#12771 reuse connection [14/26] - classic::binlog protocol
- WL#12771 reuse connection [15/26] - classic::binlog
- WL#12771 reuse connection [16/26] - xproto::auth
- WL#12771 reuse connection [17/26] - xproto::crud::*
- WL#12771 reuse connection [18/26] - xproto::prepare::*
- WL#12771 reuse connection [19/26] - xproto::expect::*
- WL#12771 reuse connection [20/26] - xproto::crud::*_view
- WL#12771 reuse connection [21/26] - xproto::cursor::*
- WL#12771 reuse connection [22/26] - xproto::sess::*
- WL#12771 reuse connection [23/26] - connection-pool
- WL#12771 reuse-connection [24/26] - integration connection pool
- WL#12771 reuse connection [25/26] - REST API
- WL#12771 reuse connection [26/26] - integration test
- 
- Bug #33454557 Innodb asserts due to corruption in the fsp layer while running MTR tests
- BUG#33362935 NDBD: redundant connect-retries entry in help message
- Bug #33600109 Segmentation fault at ha_innobase::index_type， Improper code handling
- Bug #33702755: heap-use-after-free failures for cli_read_query_result_nonblocking
- Bug#33722819: recent regression: crash/corruption with gcols and multitable delete...
- Bug#33752914 Allow receiveThread to become more active
- Bug#33752983  NDB API sendThread failed to raise its priority
- Bug#33670002 SPJ: 'Got error 4829 'FirstInner/Upper has to be an ancestor or a sibling'
- Bug#33752914 Allow receiveThread to become more active
- Bug #33755144 	FatalSignalDeathTest.CrashOnParallelAbort is unstable
- 
- Bug#33752983  NDB API sendThread failed to raise its priority
- 
- 
- Bug #32716838 ALTER TABLE MAY CRASH IF DDL LOG LOSE, IN NON-FILE-PER-TABLE MODE
- Bug#33753245 mysqld crashes in query with WINDOW function and WITH ROLLUP modifier
- 
- Bug#33688141: mysqlpump exits in regards to 'std::logic_error'
- Post-push fix: Bug#33688141: mysqlpump exits in regards to 'std::logic_error'
- 
- Bug#33594199 Assertion `in_transaction == false' at sql/binlog.cc [noclose]
- Bug#33760236 ClusterJ test fails when cross compiled in Java 11 for Java 8 [noclose]
- Bug #32130854 MOVE UTF8 CHARSET REPRESENTATION MANAGEMENT TO CHARSET_INFO
- Bug #33730302 Move to GCC 11 on el7 and el8 [choose compiler]
- Bug #33730302 Move to GCC 11 on el7 and el8 [compiler warnings]
- Bug #33730302 Move to GCC 11 on el7 and el8 [mtr results]
- Bug #33574176 Compile MySQL with GCC 12 [noclose]
- 
- 
- 
- 
- Bug#33688141: mysqlpump exits in regards to 'std::logic_error'
- 
- Bug#33669039 SPJ read stale row, undefined result, hit assert(key_cmp_if_same..
- Bug#33684418 Ship standard debuginfo package for el6 and el7 RPM package formats
- Bug#33757412 computeXorChecksum, improve performance by vectorize generated code
- Bug#33757412 computeXorChecksum, improve performance by vectorize generated code
- Bug#33757412 computeXorChecksum, improve performance by vectorize generated code
- Bug#33757412 computeXorChecksum, improve performance by vectorize generated code
- 
- Bug#33773243 stdx::expected<T, void> is deprecated
- Bug#25790312 MTR MAY USE INCORRECT VERSION OF CLUSTERJ JAR FILES
- Bug#33757412 computeXorChecksum, improve performance by vectorize generated code
- 
- Bug#33559169 Crash in touch_mem on ARM running MTR tests on MySQL Cluster 8.0.27
- Bug #33769505 Potentially incompatible system Boost gets picked even with -DWITH_BOOST=/path
- Bug#33671869 -  ndb_transaction_memory_shortage test does not cover the scan out of TC connect
- Bug#33783762 routertest_component_config fails with Address already in use
- Bug #33752981 Implement workaround for ancient version of bison on macOS
- Bug #33784168 UnknownClusterSetTargetCluster/5 fails randomly
- 
- Bug#33778845 stdx::expected<> monadic interface as
- Bug #33657235 Assertion failure: lock0lock.cc:NNN:!other_lock thread NNN
- Bug#33760236: ClusterJ test fails when cross compiled in Java 11 for Java 8
- 
- 
- 
- 
- Bug #33724166 assertion: m_prebuilt->new_rec_locks_count() == 0, "Unlock row could not find a 3 mode lock"
- 
- Bug#33535379 Wrong value for RAND() expression when aggregating on empty table.
- Bug#33789954 Add configuration for autotest up-/downgrade tests with ndbd
- 
- Bug#33786936 AppTest.CmdLineExtraConfigNoDefaultFail test fails
- Bug#33785422 error-counter incremented too late
- Bug#33787933 typo in comments
- 
- Bug#31884434: ASSERTION "HA_INNOBASE..TRY_SEMI_CONSISTENT_READ"
- 
- 
- 
- 
- 
- 
- 
- 
- WL#14722: Support IF NOT EXISTS clause in CREATE PROCEDURE/FUNCTION/TRIGGER
- Bug#33788832: wrong result with COALESCE function
- Bug#33791393 use string_view instead of raw pointers and size
- Bug#32929392 - TEMPTABLE STORAGE ENGINE LACKS OF THE ERROR HANDLER OF RESULT::RECORD_FILE_FULL
- 
- Bug#33783274 EventBuffer hash-key generation ignore most signifficant byte of key
- Bug#33778923 Collection of 'EventBytesRecvdCount' client statistics has too much overhead
- Bug#33783274 EventBuffer hash-key generation ignore most signifficant byte of key
- Bug#33778923 Collection of 'EventBytesRecvdCount' client statistics has too much overhead
- 
- BUG#33059773: AFTER_PRIMARY_ELECTION() IS CALLED WITH INCORRECT PARAMETER
- 
- Bug#33789954 Add configuration for autotest up-/downgrade tests with ndbd
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- Bug#33777821 Memory leak when checking a view of a view
- Bug 33764307 : row0pread.cc:474:!m_trx || m_trx->isolation_level == trx_t::READ_UNCOMMITTED...
- 
- 
- Bug #31862170 :  SASL PLUGIN DOES NOT HONOUR KERBEROS STANDARD OF PORTS IN HOSTNAME
- 
- Bug #33773026 Rename CHARSET_INFO.name to m_coll_name
- Bug #33791736 The POST_BUILD TARGET libmysql_api_test may fail when bulding with Xcode
- Bug #33764046 gcov build fails with GCC 11
- Revert "BUG#33059773: AFTER_PRIMARY_ELECTION() IS CALLED WITH INCORRECT PARAMETER"
- Bug #32984903 - DBUG_LOG LEADS TO A CRASH WHEN DBUG_PRINT HAS NOT BEEN USED EARLIER
- Bug #33800310 Add cmake target clang_tidy_prerequisites
- Bug#33769850 NDB_STARTUP FAILS IN PB2 WITH WRONG ERROR
- 
- 
- Bug#33795042 PB2 win: Argument "1/8" isn't numeric in numeric ge (>=) at mysql-test-run.pl
- 
- 
- 
- 
- 
- WL#14857 Add configurable variable to delay clone process
- Bug #33684418 Follow-up: allow builds with full debug for el6/el7
- Bug #33800310 Add cmake target clang_tidy_prerequisites
- 
- Bug#33769911 With ROLLUP, projection size may be needlessly large, twice in select list
- 
- Bug #11745929 CHANGE LOCK PRIORITY SO THAT THE TRANSACTION HOLDING S-LOCK GETS X-LOCK F
- Bug #33783709 mem_heap_alloc could provide better hints to VALGRIND
- Bug #33800310 Add cmake target clang_tidy_prerequisites
- 
- Bug#33748078: Disable thread_pool_max_transaction_limit for non-cloud build
- Bug#33542052 grant ndb_stored_user alters user with wrong hash
- Bug #11745929 CHANGE LOCK PRIORITY SO THAT THE TRANSACTION HOLDING S-LOCK GETS X-LOCK F
- Bug#33787914: int-decimal. using left join, the returned result is incorrect
- Bug#33794977: Wrong results from hash join with compound hash keys [patch 1/2, fix]
- Bug#33794977: Wrong results from hash join with compound hash keys [patch 2/2, only multi-column]
- Bug#33334928: SELECT COUNT(*) using multi-valued index reports wrong number of rows.
- Bug#33804178 : innodb.instant_ddl_upgrade_rollback is failing
- 
- Bug#33803583 NdbBlob::getBlobKeyHash hash-key generation ignore most signifficant byte of key
- 
- 
- Bug #33768149 Clean up usage of init_sql_alloc
- 
- Bug#33722819: recent regression: crash/corruption with gcols and multitable delete...
- Bug#33806018: Remove usage of MD5 in explain_access_path.cc
- Bug#33559169 Crash in touch_mem on ARM running MTR tests on MySQL Cluster 8.0.27
- 
- 
- 
- Bug #33788290 Look for winflexbison3 rather than Gnuwin32 bison
- 
- 
- BUG#33302400 SYS.SCHEMA_UNUSED_INDEXES IS NOT ACCURATE AND NOT UPDATING
- Bug#33776832: [InnoDB] Assertion failure: page0cur.ic:207:cmp_dtuple_rec(tuple, rec, index, *o
- 
- Bug #33791736 The POST_BUILD TARGET libmysql_api_test may fail when bulding with Xcode
- Bug#21040523 NDB REPLICATION : MYSQLD BATCH SIZE PARAMETER LIMIT TOO LOW
- Bug #33067891 ASSERTION : M_ID != LATCH_ID_NONE IN SYNC0TYPES.H
- Bug#21040523 NDB REPLICATION : MYSQLD BATCH SIZE PARAMETER LIMIT TOO LOW
- Bug#33791802: Query with multiple JOINs stopped working
- Bug #33809262 Run clang-tidy modernize-use-nullptr (again) [noclose]
- 
- Bug#33814906 Metadata cache defines unused exception connection_error
- Bug#33814983 Invalid doxygen in metadata-cache header
- Bug#33815026 Metada cache enum InstanceStatus is unused.
- Bug#33815053 enum ClusterStatus is internal and need not be exposed in metadata cache API
- Revert of patch for Bug#33752914 Allow receiveThread to become more active
- Bug#21040523 NDB REPLICATION : MYSQLD BATCH SIZE PARAMETER LIMIT TOO LOW
- Bug#21040523 NDB REPLICATION : MYSQLD BATCH SIZE PARAMETER LIMIT TOO LOW
- Bug #33414289 Bounds_checked_array [] Assertion `n < m_size' failed
- Bug#33728909 8.0.28 mysqld initial start stuck after 'Could not create index stat system tables'
- Bug#33673803 Re-bootstrap fails when Cluster name contains '-' or '.'
- Bug #33809262 Run clang-tidy modernize-use-nullptr (again) [noclose]
- Bug #33810558 Invalid Unicode characters can be inserted in column with UTF32 character set
- BUG #30509416 - MYSQL-8.0: NODES CRASHED DURING ROWS DELETION  WITH AN ERROR "SIGNAL 11 RECEIVED;
- BUG#33059773: AFTER_PRIMARY_ELECTION() IS CALLED WITH INCORRECT PARAMETER
- 
- Bug#29769061 ENABLE DEBUGINFO PKGS AND MOVE DEBUG SERVER TO OWN PKG ON EL8
- Bug#33334928: SELECT COUNT(*) using multi-valued index reports wrong number of rows.
- Bug#33796754 Hash-search for EventBufData is inneficient
- 
- Bug #33768149 Clean up usage of init_sql_alloc
- Bug#33673803 Re-bootstrap fails when Cluster name contains ‘-’ or ‘.’
- Bug #33810558 Invalid Unicode characters can be inserted in column with UTF32 character set
- Bug #33789526 srv_error_monitor_thread shouldn't call functions which can deadlock
- Bug#33791270  Binding query thread to cpu through ThreadConfig crashes data nodes
- 
- Bug#33414289: Bounds_checked_array [] Assertion `n < m_size' failed
- Bug #33819270 innodb.lob_recovery is failing for --initialize=--innodb-page-size=4k
- BUG#32232092: GTID_PURGED SHOULD NOT BE PERSISTABLE
- Bug #33795721 WalkAccessPaths handles APPEND Access Path incorrrectly
- Bug#33813951: Binary resultset for `SELECT ?` with date/time gives wrong type
- 
- Bug#33790919: Null-safe compare works incorrectly with TIMESTAMP in trigger
- BUG#33824392: Improve show_rpl_debug_info.inc
- Bug#33831340 mtr: ndbinfo plugin can not be off due to use in mtr_check.sql
- BUG#33830926: rpl_gtid.rpl_mts_spco_acl_commands_binlog fails on pb2
- Bug#33796754 Hash-search for EventBufData is inefficient
- 
- Bug #33809262 Run clang-tidy modernize-use-nullptr (again) [noclose]
- Bug#33815426: Inconsistent rows matched value for multi-table UPDATE with failed constraint
- Bug#33727690 Support compression for backup and LCP on Windows
- Bug#33727629 Misc fixes for ndb_file on Windows
- Bug#33727629 Misc fixes for ndb_file on Windows
- WL#14285 Fixup [10] ndbxfrm_file: Remove use of REQUIRE.
- 
- 
- 
- Bug#33822602: Improve MY-010928 your password has expired error on the server to show the user
- 
- Bug #30567295 	SOME PB2 TEST FAILS WITH OPERATING SYSTEM ERROR NUMBER 995 IN A FILE OPERATION.
- Bug #33800310 Add cmake target clang_tidy_prerequisites
- Bug#33803541 Excessive CPU consumed by Dblqh::findTransaction()
- Bug #33830073 Assertion 'num_codepoints >= scanner.get_char_index()'
- 
- Bug#33764260  THRConfig::reorganize_ldm_bindings
- Bug#33803487 Large transaction runs gradually slower, excessive Dbacc::getElement() overhead
- Bug #33730302 Move to GCC 11 on el7 and el8 [choose compiler]
- Bug#33803487 Large transaction runs gradually slower, excessive Dbacc::getElement() overhead
- WL#15060: Change group_replication_transaction_size_limit default in 5.7
- Bug #33800310 Add cmake target clang_tidy_prerequisites
- Bug #33809262 Run clang-tidy modernize-use-nullptr (again)
- Post-push fix: BUG#33824392: Improve show_rpl_debug_info.inc
- Bug #33824544: Add a primary key for mysql.firewall_membership RB#27661
- Bug #33830493: SET PERSIST statement is not atomic
- 
- Bug #33840645 	InnoDB: AIO doesn't need SYNC queue
- Bug #33794610 	Remove dependency on randomly-sampled statistics in MTR tests in EXPLAIN queries
- Bug#33729821: TP: Debug improvements
- Bug #33797931 Ndb : Avoid excessive operation list traversal when defining Blob operations
- Bug #33830493: SET PERSIST statement is not atomic
- Fix for failing test script dd_schema_definition_after_upgrade_80_debug.test
- 
- 
- Bug#33847722 - triggers - heap-use-after-free in JOIN::update_depend_map
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- Bug#33719861 FIXES NDB_RESTORE_APPLY_STATUS TEST
- Bug #33768584 	InnoDB: space->n_pending_flushes == 0 is triggered with srv_fast_shutdown == 2
- Bug #33690316 	Test `gr_crash_donor_server_continuously` fails on PB2
- 
- BUG#33822549 MYSQL 8.0.28 ROUTER BINARY MISSING IN INFO_BIN & INFO_SRC FOR WINDOWS PLATFORMS BUG#33842889 MEB 8.0.28 BINARY MISSING IN INFO_BIN & INFO_SRC FOR WINDOWS PLATFORMS
- Bug#33869347 MTR performance_schema.processlist_port does not test intended use case
- Bug #33813846 Upgrade to latest protobuf library [add sources]
- Bug #33813846 Upgrade to latest protobuf library [spec files]
- Bug #33813846 Upgrade to latest protobuf library [patches]
- Bug #33813846 Upgrade to latest protobuf library [remove old version]
- Bug #33813846 Upgrade to latest protobuf library [valgrind suppressions]
- Bug #33690316 	Test `gr_crash_donor_server_continuously` fails on PB2 [post-push fix]
- Bug #33768584 	InnoDB: space->n_pending_flushes == 0 is triggered with srv_fast_shutdown == 2 [post-push fix]
- Bug #33690316 	Test `gr_crash_donor_server_continuously` fails on PB2 [post-push fix]
- Bug#33838439: ExtractValue not working properly with COUNT
- Bug#33612166 DOXYGEN: UPGRADE TO VERSION 1.9.3 [no-close]
- Bug#33791802: Query with multiple JOINs stopped working
- Bug#33869388 performance_schema.processlist returns no rows   for anonymous account with PROCESS privs
- Bug#33874311 memory leak in runTestMgmdwithoutnodeid
- Bug#33872577 Main thread(tid:0) runs 'Out of job buffers' due to extra JB quota assigned.
- Bug#33856371 Too much overhead in ::execSEND_PACKED() after WL#13980
- Bug 26906787 - INCORRECT AUTO_INCREMENT IDS ISSUED WHEN CLOSE TO ULLONG_MAX
- 
- 
- 
- Bug#33830212 reuse_connection test fails with passthrough
- Bug#33830146 gr_notifications test fails GrNotificationMysqlxWaitTimeoutUnsup
- 
- 
- 
- Bug #33813846 Upgrade to latest protobuf library [remove old version]
- Bug#33840613: Empty transaction blocks group_replication_set_as_primary
- Bug#33862323 Replace systemd PermissionsStartOnly with executable prefix
- Bug#33869004 Range optimizer is not aware of the session memory limit.
- WL#13899 : INSTANT DROP (and ADD) COLUMN
- Bug 33859937: InnoDB: Doxygen complains that void methods should not have @retval or @retvalue
- Bug#33837691 UNIX_TIMESTAMP() works differently in 8.0.28 than in previous versions
- Bug#33611915 ArrayPool [-Wmaybe-uninitialized] [noclose]
- Bug#33611915 IntrusiveList [-Wmaybe-uninitialized] [noclose]
- Bug#33611915 TransientPool [-Wmaybe-uninitialized] [noclose]
- Bug#33611915 VMSignal [-Wmaybe-uninitialized] [noclose]
- Bug#33611915 DLHashTable [-Wmaybe-uninitialized] [noclose]
- Bug#33611915 Dbdict [-Wmaybe-uninitialized] [noclose]
- Bug#33611915 Dbdih [-Wmaybe-uninitialized] [noclose]
- Bug#33772598 Cleanup in Dblqh, use FsOpenReq.
- Bug#33772598 FSOPENREQ: add flags OM_ENCRYPT_CBC/XTS/PASSWORD/KEY
- Bug#33772598 Pass Om_ENCRYPT_XTS for REDO log
- Bug#33772598 Pass Om_ENCRYPT_XTS for TS data file
- Bug#33772598 Pass Om_ENCRYPT_XTS for UNDO
- Bug#33772598 Pass Om_ENCRYPT_XTS for LCP data
- Bug#33772598 Replace some const char* with string_view
- Bug#33772598 Replace EncryptionPasswordData with EncryptionKeyMaterial
- Bug#33772598 Make block fill in OM_ENCRYPT_KEY for LCP data and REDO
- Bug#33772598 Make block fill in OM_ENCRYPT_KEY for TS and UNDO
- Bug#33772598 Remove code guarded by NDBFS_TDE
- 
- Bug #33635445: SET PASSWORD has no effect on PAM authentication, but binlogged as ALTER USER RB#27689
- Bug #33830765 Explicit CAST allows to insert invalid character strings in strict mode
- Bug#33837691 UNIX_TIMESTAMP() works differently in 8.0.28 than in previous versions [follow-up]
- 
- 
- 
- Bug #33851256 crash on recursive decimal stored function call
- Bug #33865094 Assertion `m_addon_buf_length == sz' at sort_param.h
- Bug 33482747 clone failed leaving clone_progress stuck
- 
- 
- 
- 
- 
- 
- Bug#33881825 Unit test ndbd_malloc-t failing
- Bug#33822729 : Table version metadata is not reset when all partitions                are truncated
- Bug 33482747 clone failed leaving clone_progress stuck
- Bug#33060953 - UPGRADE FROM 5.7 TO 8.0 CHANGE EXECTION PLAN                IN PARTITIONED TABLE, EVEN AFTER REVERTING                NEW OPTIMIZER FLAG
- 
- 
- 
- Revert "Bug #33830765 Explicit CAST allows to insert invalid character strings in strict mode"
- Bug#33855163 / Bug#33825393 crashes in ha_ndbinfo::unpack_record()
- Bug33784672 : off by one in rec_convert_dtuple_to_rec_old in               innodb.instant_ddl_recovery_old_2
- 
- 
- 
- 
- 
- 
- Bug #33856332 InnoDB should let HP transaction timeout while waiting on a lock
- Bug#33809478: Installing firewall plugin back after persisting variable exits the server
- Bug#33270516: 8.0.22+: crash in KEY::has_records_per_key const
- Bug#33101025: ASSERTION `!TABLE || (!TABLE->READ_SET ||               BITMAP_IS_SET(TABLE->READ_SET, FIELD_INDEX()))' FAILED.
- Bug #33874978 GR notification thread can get stuck when new nodes are added to the Cluster
- BUG #32586721 : SERVER CRASH IN FIL_SPACE_IS_BEING_TRUNCATED
- Bug #33664929 Reduce size of RelWithDebInfo binaries
- Bug#33903639 Fix spell of log messages in mysql_safe
- Bug#33776774 ndbinfo_plans test case unstable
- Bug#33886102: Disable thread_pool_dedicated_listeners for non-cloud builds
- Bug#33134835 RPL_WRITESET_ADD_UNIQUE_KEY FAILS SPORADICALLY [noclose]
- Bug#33781534 - I_S.KEY_COLUMN_USAGE does not lists invisible columns                which has key constraints.
- Bug #33896399 NdbInfo pools cursor cookie robustness
- Bug #33896356 Check SysFile nodeId in range
- Bug #33896399 NdbInfo pools cursor cookie robustness
- 
- Bug #33896389 Improve request validation for NdbFs file operations
- Bug #33896389 Improve request validation for NdbFs file operations
- 
- Bug#33913908 Disable replica worker for NDB replication
- Bug#33855045 Test ndbinfo_upgrade fails on valgrind builds [noclose]
- Bug#31477725 NDB_RESTART FAILS SPORADICALLY IN PB2
- 
- Bug#33825480 dba.upgradeMetadata() does not recognize upgraded router version
- 
- 
- 
- 
- 
- 
- 
- 
- 
- Bug #33904267 Build fails on EL7 ARM when using self-provided shared libraries
- 
- 
- Bug#33755920: InnoDB Cluster Start of Secondary Results in Shutdown of Primary with MY-013309 and MY-011452
- 
- Bug#33916404 Ensure validity of nodeId for requested nodes in a node group
- Bug#33896409 Check validity of received NodeId
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- 
- Bug#33924532 : Assert failure in purge which tries to purge UNDO across INSTANT DDL
- Bug #33881339 Default collation causes inconsistent results
- Bug#33732907: Slow query log logs password in plain text on syntax error
- Bug#33732907: Slow query log logs password in plain text on syntax error
- Bug #33916134 Enhance checking of CREATE_FRAGMENTATION_REQ processing
- Update License Book
- Updating man pages from Docs
- 
- Bug#33924532 : Assert failure in purge which tries to purge UNDO across INSTANT DDL
- Bug#33856374: Query returns incorrect results in 8.0.22+
- Bug#33908146 perfschema.processlist_reg_user fails on daily-trunk
- Bug#33728209: set variable to function value cause double locking mutex & deadlock/hang
- Bug #33705056: Assertion failure on double LOCK_plugin inside page_track_service in debug mode
- Bug #33732576: innodb.innodb_extend_and_initialize_windows is failing on PB2 daily runs
- Bug #33705056: Assertion failure on double LOCK_plugin inside page_track_service in debug mode
- 
- 
- 
- 
- Bug#33956760 mysql-community-server-debug not obsoleted by any commercial package
- Bug#33956785 sl15/sles12 cluster-8.0.29 RPM builds are also building router packages
- Bug #33965055: add PK to mysql.firewall_membership with mysql_system_tables_fix.sql fails RB#27762
- Bug#33896428 Checking signal is not remote.
- Bug#33976268 NDB: gcc 11 Ubuntu 21.10 : build stops with [-Werror=maybe-uninitialized]
- 
- 
- 
- 
- 
- 
- 
- 
- 
- Implemented PS-8185 (Merge MySQL 8.0.29) (conflicts resolved)
- PS-8185 - Fix struct PSI_thread_service_v6
- PS-8185 - Fix uses of ib_uint32_t and ib_uint64_t
- PS-8185 - Fix simple compilation errors
- PS-8185 - Fix use of constants TRUE and FALSE
- PS-8185 - Fix uses of find_sys_var
- PS-8185 - Fix create_compression_dictionary_allowed_expr lexical token
- PS-8185 - Fix definition location of ZIP_COLUMN_HEADER_LENGTH
- PS-8185 - Fix uses of row_events_waits::m_object_name_length
- PS-8185 - Fix uses of rec_get_nth_field functions
- PS-8185 - Fix uses of log_buffer_write
- PS-8185 - Fix uses of log_write_encryption
- PS-8185 - Fix calls to mem_heap_create
- PS-8185 - Fix log_online_rec_has_page
- PS-8185 - Fix uses of event SRV_START_STATE_MONITOR
- PS-8185 - Fix uses of `trx_sys_t::rw_max_trx_id`
- PS-8185 - Solve YYSTYPE redefinition bug
- PS-8185 - Remove duplicate mutex creation of LATCH_ID_MASTER_KEY_ID_MUTEX
- PS-8185 - Fix test i_s_schema_definition_debug
- PS-8185 - Update PFS schema signature in dd_version_check.test
- PS-8185 - Fix uses of character set utf8
- PS-8185 - Re-record tests affected by changes in table_encrypt_3.inc
- PS-8185 - Re-record test perfschema.processlist_anonymous
- PS-8185 - Re-record test affected by deprecation of replica_parallel_type
- PS-8185 - Fix tests affected by WL#14592 (Refactor system variable processing in parser and executor)
- PS-8185 - Fix tests affected by the new variable xa_detach_on_prepare
- PS-8185 - Fix tests affected by the new type SENSITIVE
- PS-8185 - Fix tests group_replication.gr_applier_channel_stop_with_pending_transactions and gr_stop_procedure_hang
- PS-8185 - Fix test mysqlpump_bugs
- PS-8185 - Re-record test percona_sequence_table
- PS-8185 - Support hint updateable variables of type PLUGIN
- PS-8185 - Re-record tests affected by deprecation of myisam_repair_threads
- PS-8185 - Re-record tests affected by deprecation of delimiters and whitespaces in time/date/datetime/timestamp literals
- PS-8185 - Re-record test rocksb.index_merge_rocksdb2
- PS-8185 - Fix test innodb.hp_deadlock
- PS-8185 - Remove duplicate header inclusions
- PS-8185 - Re-record innodb.partition


ldonoso:
Jenkins ASAN: https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-pipeline/38678/

cmake/lz4.cmake
```diff
@@ -23,7 +23,6 @@
 # cmake -DWITH_LZ4=system|bundled
 # bundled is the default
 
-<<<<<<< HEAD
 SET(LIBLZ4_VERSION_REQUIRED "1.9.3")
 
 MACRO (CHECK_LZ4_VERSION)
```
oleksandr-kachan: This macro partially duplicates behavior of upstream's FIND_LZ4_VERSION(). The difference is only that here we have min version check. Please take a look if we could avoid duplication here.

storage/innobase/include/mach0data.h
```diff
@@ -119,14 +119,8 @@ static inline uint32_t mach_get_compressed_size(ulint n);
 @param[in,out]  b       pointer to memory where to read;
 advanced by the number of bytes consumed
 @return unsigned value */
-<<<<<<< HEAD
 MY_NODISCARD
```
oleksandr-kachan: Should we use [[nodiscard]] instead? Same below.

storage/innobase/include/page0zip.h
```diff
@@ -117,14 +117,8 @@ void page_zip_set_alloc(void *stream, mem_heap_t *heap);
 /** Compress a page.
  @return true on success, false on failure; page_zip will be left
  intact on failure. */
-<<<<<<< HEAD
 MY_NODISCARD
```
oleksandr-kachan: [[nodiscard]] ?

breisig:
It's been 2 months since the original release. Any update?

breisig:
Can you also please create .el9 releases since RHE9, AlamaLinux 9,Oracle 9, and very soon Rocky Linux 9 has been released and people are using it already.

cmake/lz4.cmake
ldonoso: Good point, I've implemented the change on https://github.com/percona/percona-server/commit/32c3607cbb0db2511630781f68fbb45f6ac82f25

breisig:
LGTM

PR #4808, opened 2022-05-30
PS-8258 feature: Make Percona Server buildable on MacOS with RocksDB enabled

percona-ysorokin:
ttps://jira.percona.com/browse/PS-8258

In the top-level 'CMakeList.txt' file we now include 'extra/libkmip'
subdirectory only when 'KMIP' keying component is configured to be built.

RocksDB Storage Engine 'CMakeLists.txt' file now detects MacOS ('Darwin')
platform and adds 'OS_MACOSX' preprocessor definition.

'rt' system library no longer added to 'ha_rocksdb.so' as a dependency
on MacOS.

'rdb_buff.h' header from the RocksDB Storage Engine code extended with
a set of macros for emulating big-endian / little-endian
byte-manipulation functions for MacOS.

Some RocksDB system variables declared with 'MYSQL_SYSVAR_ENUM' were
using global variables of type 'uint64_t'. This was incompatible with
the macro definition as it expected variables of type 'ulong'. On Linux
platform it was not a problem as 'uint64_t' was defined as 'unsigned long'
there. However, on MacOS 'uint64_t' is defined as 'unsigned long long'.
Fixed by changing the type of underlying variables from 'uint64_t' to
'ulong'.

The family of 'MYSQL_SYSVAR_XXX' macros extended with
'MYSQL_SYSVAR_UINT64_T' that is supposed to be used for 'uint64_t'
variables. Because of the same reasons ('uint64_t' being defined
differently on Linux and MacOS) we needed this new macro to be able
bind system variables directly to RocksDB library variables that have
'uint64_t' type (the type of which we cannot change as this is RocksDB
submodule code).

A number of 'xxxprintf()' calls in the RocksDB Storage Engine code were
incorrectly using '%lu' format specifier for 'uint64_t' types. Again,
this was not a problem on Linux as 'uint64_t' was defined as
'unsigned long' there. For compatibility with MacOS, fixed by using
proper 'PRIu64' format specifier instead.

Fixed problem with passing an empty string to the 'std::regexp'
constructor that used to work on Linux but is not allowed by the
standard.


storage/rocksdb/ib_ut0counter.h
```diff
@@ -84,7 +84,9 @@ struct thread_id_indexer_t : public generic_indexer_t<Type, N> {
   /* @return a random number, currently we use the thread id. Where
   thread id is represented as a pointer, it may not work as
   effectively. */
-  size_t get_rnd_index() const { return get_curr_thread_id(); }
+  size_t get_rnd_index() const {
+    return reinterpret_cast<std::uintptr_t>(get_curr_thread_id());
```
laurynas-biveinis, line 88: @percona-ysorokin the function returns `size_t` but the cast is to `std::uintptr_t`

percona-ysorokin: @laurynas-biveinis `get_curr_thread_id()` is basically `pthread_self()` that returns `pthread_t`. On all Linuxes and  MacOSes I tested, `pthread_t` was either `void *` or `<some_opaque_struct> *`, a pointer to something in other words. In the standard `std::uintptr_t` is defined as an unsigned type with the same size as a pointer. That is why it is safe to use `reinterpret_cast<std::uintptr_t>()` here. After that, that `std::uintptr_t` can be easily implicitly converted to `std::size_t` (which is either `unsigned long` or `unsigned long long`).
Moreover, although I did not test this, I believe the same construct should work o Windows as well where `get_curr_thread_id()` is  `GetCurrentThreadId()` that returns `DWORD`.
So, it is intentional.

laurynas-biveinis: @percona-ysorokin I see, this makes sense, thanks. Maybe I'd consider doing two casts in a row, but that could be too verbose

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1500,12 +1500,13 @@ static MYSQL_SYSVAR_INT(max_open_files, rocksdb_db_options->max_open_files,
...
-                          "DBOptions::max_total_wal_size for RocksDB", nullptr,
-                          nullptr, rocksdb_db_options->max_total_wal_size,
-                          /* min */ 0L, /* max */ LONG_MAX, 0);
+static MYSQL_SYSVAR_UINT64_T(max_total_wal_size,
+                             rocksdb_db_options->max_total_wal_size,
+                             PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_READONLY,
+                             "DBOptions::max_total_wal_size for RocksDB",
+                             nullptr, nullptr,
+                             rocksdb_db_options->max_total_wal_size,
+                             /* min */ 0L, /* max */ LONG_MAX, 0);
```
laurynas-biveinis, line 1509: @percona-ysorokin have you considered replacing `LONG_MAX` with `UINT64_MAX`?

percona-ysorokin: Yes, @laurynas-biveinis
Agree, looks like we should have done this.
Reported https://jira.percona.com/browse/PS-8359

PR #4814, opened 2022-06-09
PS-8277 fix: Assertion in vio_ssl_write() after executing asymmetric_verify() UDF

percona-ysorokin:
https://jira.percona.com/browse/PS-8277

Fixed problem in the 'opensslpp::verify_with_rsa_public_key()' where after
unsuccessful signature verification OpenSSL error code queue may not be
properly cleared affecting invoking code that may rely on 'ERR_get_error()'.

Reworked 'opensslpp::core_error::raise_with_error_string()' function - now
we also make sure that OpenSSL error code queue is empty before throwing
the exception.

Fixed problem in UDF wrappers with incorrect helper function signature for the
'INT_RESULT' UDFs.

### DIFF
--- a/extra/opensslpp/src/opensslpp/core_error.cpp
+++ b/extra/opensslpp/src/opensslpp/core_error.cpp
@@ -39,6 +39,7 @@ static constexpr std::size_t error_message_buffer_size = 256;
 
     ERR_error_string_n(err, buffer.data(), buffer.size());
     message += buffer.data();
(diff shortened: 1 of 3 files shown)


percona-ysorokin:
Jenkins build link (--do-test=percona_encryption_udf --repeat=2)
https://ps80.cd.percona.com/job/percona-server-8.0-param/1234/

PR #4815, opened 2022-06-10
PS-8257 Failed upgrade attempt to 8.0.29 corrupts the data dictionary

rahulmalik87:
https://jira.percona.com/browse/PS-8257

Problem:
Failed upgrade attempt to 8.0.29 corrupts the data dictionary

Analysis:
8.0.29 has changed the redo log format and shuffle the redo type code,
so lower version can't parse redo log file in case of failed upgrade

Fix:
Recreate the empty redo log file during failed upgrade
so lower version does not need to parse redo file

### DIFF
--- a/share/messages_to_error_log.txt
+++ b/share/messages_to_error_log.txt
@@ -11572,6 +11572,9 @@ ER_IB_WARN_MANY_NON_LRU_FILES_OPENED
 ER_IB_MSG_TRYING_TO_OPEN_FILE_FOR_LONG_TIME
   eng "Trying to open a file for %lld seconds. Configuration only allows for %zu open files. Consider setting innobase_open_files higher."
 
+ER_IB_MSG_DOWNGRADING_LOG_FILE
+  eng "Upgrade fail, cleaning up redo logs, so server can start with previous version"
+
 ER_GLOBAL_CONN_LIMIT
   eng "Connection closed. Global connection memory limit %llu bytes exceeded. Consumed %llu bytes."
 
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -1212,6 +1212,7 @@ bool opt_no_monitor = false;
 bool opt_no_dd_upgrade = false;
 long opt_upgrade_mode = UPGRADE_AUTO;
 bool opt_initialize = false;
+bool dd_init_failed_during_upgrade = false;
 bool opt_skip_replica_start = false;  ///< If set, slave is not autostarted
 bool opt_enable_named_pipe = false;
 bool opt_local_infile, opt_replica_compressed_protocol;
@@ -6586,6 +6587,10 @@ static int init_server_components() {
     if (!is_help_or_validate_option() &&
         dd::init(dd::enum_dd_init_type::DD_RESTART_OR_UPGRADE)) {
       LogErr(ERROR_LEVEL, ER_DD_INIT_FAILED);
+
+      if (!dd::upgrade::no_server_upgrade_required()) {
+        dd_init_failed_during_upgrade = true;
+      }
 
       /* If clone recovery fails, we rollback the files to previous
       dataset and attempt to restart server. */
--- a/storage/innobase/log/log0chkp.cc
+++ b/storage/innobase/log/log0chkp.cc
@@ -448,7 +448,7 @@ void meb_log_print_file_hdr(byte *block) {
 
 #ifndef UNIV_HOTBACKUP
 
-void log_files_downgrade(log_t &log) {
+void log_files_downgrade(log_t &log, uint32_t log_format) {
   ut_ad(srv_shutdown_state.load() >= SRV_SHUTDOWN_LAST_PHASE);
   ut_a(!log_checkpointer_is_active());
 
@@ -462,7 +462,7 @@ void log_files_downgrade(log_t &log) {
       static_cast<page_no_t>(dest_offset / univ_page_size.physical());
 
   /* Write old version */
-  mach_write_to_4(buf + LOG_HEADER_FORMAT, LOG_HEADER_FORMAT_5_7_9);
+  mach_write_to_4(buf + LOG_HEADER_FORMAT, log_format);
 
   log_block_set_checksum(buf, log_block_calc_checksum_crc32(buf));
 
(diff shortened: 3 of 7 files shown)


storage/innobase/srv/srv0start.cc
```diff
@@ -3936,21 +3944,63 @@ static lsn_t srv_shutdown_log() {
 
   srv_wake_log_tracker_thread();
 
-  if (srv_downgrade_logs) {
+  /* if upgrade is failed, delete and create new empty redo so old version do
+   not complaint  */
+  if (dd_init_failed_during_upgrade) {
+    ib::info(ER_IB_MSG_DOWNGRADING_LOG_FILE);
+
+    ut_ad(fil_space_t::s_redo_space != nullptr);
```
ldonoso, line 3952: Please, @rahulmalik87 , could we extract this functionality in a independent function (i.e.`recreate_redo_log_files)?

This way, `srv_shutdown_log` is not so long and the functionality is easier to maintain.

storage/innobase/srv/srv0start.cc
```diff
@@ -3936,21 +3944,63 @@ static lsn_t srv_shutdown_log() {
...
+    lsn_t new_checkpoint_lsn = 0;
+
+    dirnamelen = strlen(srv_log_group_home_dir);
+    ut_a(dirnamelen < (sizeof logfilename) - 10 - sizeof "ib_logfile");
+    memcpy(logfilename, srv_log_group_home_dir, dirnamelen);
+
+    /* Add a path separator if needed. */
+    if (dirnamelen && logfilename[dirnamelen - 1] != OS_PATH_SEPARATOR) {
+      logfilename[dirnamelen++] = OS_PATH_SEPARATOR;
+    }
```
ldonoso, line 3969: These same lines appear on `srv_start`. Please, maybe we could create a function and using it both in `srv_start` and here. This way if the logic changes on the future we do not need to remember to update two places.

ldonoso (changes requested):

share/messages_to_error_log.txt
```diff
@@ -11572,6 +11572,9 @@ ER_IB_WARN_MANY_NON_LRU_FILES_OPENED
 ER_IB_MSG_TRYING_TO_OPEN_FILE_FOR_LONG_TIME
   eng "Trying to open a file for %lld seconds. Configuration only allows for %zu open files. Consider setting innobase_open_files higher."
 
+ER_IB_MSG_DOWNGRADING_LOG_FILE
```
altmannmarcelo, line 11575: Shouldn't this be part of PS section?

sql/mysqld.cc
```diff
@@ -6587,6 +6588,10 @@ static int init_server_components() {
         dd::init(dd::enum_dd_init_type::DD_RESTART_OR_UPGRADE)) {
       LogErr(ERROR_LEVEL, ER_DD_INIT_FAILED);
 
+      if (!dd::upgrade::no_server_upgrade_required()) {
```
altmannmarcelo, line 6591: Are we calling no_server_upgrade_required a second time?
Why not set dd_init_failed_during_upgrade when the original upgrade failed?

storage/innobase/log/log0chkp.cc
```diff
@@ -661,7 +661,6 @@ void log_create_first_checkpoint(log_t &log, lsn_t lsn) {
   page_no_t block_page_no;
   uint64_t block_offset;
 
-  ut_a(srv_is_being_started);
```
altmannmarcelo, line 664: ut_a(srv_is_being_started || dd_init_failed_during_upgrade); ?

sql/mysqld.cc
altmannmarcelo:
```
  @retval false  ON SUCCESS
  @retval true   ON FAILURE
```

Those are the return values of that function, shouldn't we set `dd_init_failed_during_upgrade = true;` if no_server_upgrade_required returns true ?

altmannmarcelo (changes requested):
do we have any test case to stress this code path?

storage/innobase/srv/srv0start.cc
```diff
@@ -149,6 +149,9 @@ bool srv_startup_is_before_trx_rollback_phase = false;
 /** true if srv_start() has been called */
 static bool srv_start_has_been_called = false;
 
+/** current log format */
+static uint32_t server_log_format;
```
satya-bodapati, line 153: I would call log_format_before_upgrade and appropriate comment
/** Redo log format before upgrade, used to create redo log files of the same version on failed upgrades. */
Something along these lines

storage/innobase/srv/srv0start.cc
```diff
@@ -3959,7 +4009,9 @@ static lsn_t srv_shutdown_log() {
     ut_a(err == DB_SUCCESS);
   }
   buf_must_be_all_freed();
-  ut_a(lsn == log_get_lsn(*log_sys));
+  if (!dd_init_failed_during_upgrade) {
+    ut_a(lsn == log_get_lsn(*log_sys));
```
satya-bodapati, line 4013: instead of removing assert for failed upgrade, shouldn't we reset LSNs?

satya-bodapati (changes requested):

storage/innobase/srv/srv0start.cc
rahulmalik87: fixed

storage/innobase/srv/srv0start.cc
rahulmalik87: fixed

share/messages_to_error_log.txt
rahulmalik87: replaced it with ER_XB_MSG_DOWNGRADING_LOG_FILE

sql/mysqld.cc
rahulmalik87: these return values belongs to upgrade_system_schemas and not no_server_upgrade_required. 
we are doing it exactly after failed dd::int().

storage/innobase/log/log0chkp.cc
rahulmalik87: fixed

storage/innobase/srv/srv0start.cc
rahulmalik87: fixed

storage/innobase/srv/srv0start.cc
rahulmalik87: added, fixed

rahulmalik87:
Guys, Please review https://github.com/percona/percona-server/pull/4818, which is based on 8.0.29

PR #4818, opened 2022-06-13
PS-8257 Failed upgrade attempt to 8.0.29 corrupts the data dictionary

rahulmalik87:
https://jira.percona.com/browse/PS-8257

Problem:
Failed upgrade attempt to 8.0.29 corrupts the data dictionary

Analysis:
8.0.29 has changed the redo log format and shuffle the redo type code,
so lower version can't parse redo log file in case of failed upgrade

Fix:
Recreate the empty redo log file during failed upgrade
so lower version does not need to parse redo file

### DIFF
--- a/share/messages_to_error_log.txt
+++ b/share/messages_to_error_log.txt
@@ -11696,6 +11696,10 @@ ER_XB_MSG_6
 ER_XB_UNDO_DELETE_FAILURE
   eng "Deletion of 5.7 undo tablespace %s failed"
 
+ER_XB_MSG_DOWNGRADING_LOG_FILE
+  eng "Upgrade fail, cleaning up redo logs, so server can start with previous version"
+
+
 ER_UPGRADE_KEYRING_UNSUPPORTED_VERSION_ENCRYPTION
   eng "Upgrade is not possible as there are tables that are KERYING encrypted. KEYRING encryption is still experimental feature in the version you are upgrading from. To upgrade please decrypt the tables encrypted with KEYRING encryption."
 
--- a/share/messages_to_error_log.txt
+++ b/share/messages_to_error_log.txt
@@ -11772,7 +11772,7 @@ ER_KEYRING_COMPONENT_KMS_ERROR
   eng "KMS communication error: %s"
 
 ER_XB_MSG_DOWNGRADING_LOG_FILE
-  eng "Upgrade failed, cleaning up redo logs, so server can start with previous version"
+  eng "Upgrade failed, recreating redo logs with previous version"
 
 #
 # End of Percona Server 8.0 server error log messages
--- a/share/messages_to_error_log.txt
+++ b/share/messages_to_error_log.txt
@@ -11696,10 +11696,6 @@ ER_XB_MSG_6
 ER_XB_UNDO_DELETE_FAILURE
   eng "Deletion of 5.7 undo tablespace %s failed"
 
-ER_XB_MSG_DOWNGRADING_LOG_FILE
-  eng "Upgrade fail, cleaning up redo logs, so server can start with previous version"
-
-
 ER_UPGRADE_KEYRING_UNSUPPORTED_VERSION_ENCRYPTION
   eng "Upgrade is not possible as there are tables that are KERYING encrypted. KEYRING encryption is still experimental feature in the version you are upgrading from. To upgrade please decrypt the tables encrypted with KEYRING encryption."
 
@@ -11774,6 +11770,9 @@ ER_SEC_LOG_PATH_NULL
 
 ER_KEYRING_COMPONENT_KMS_ERROR
   eng "KMS communication error: %s"
+
+ER_XB_MSG_DOWNGRADING_LOG_FILE
+  eng "Upgrade failed, cleaning up redo logs, so server can start with previous version"
 
 #
 # End of Percona Server 8.0 server error log messages
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -149,6 +149,10 @@ bool srv_startup_is_before_trx_rollback_phase = false;
 /** true if srv_start() has been called */
 static bool srv_start_has_been_called = false;
 
+/** Redo log format before upgrade, used to create redo log files of the same
+version on failed upgrades*/
+static uint32_t log_format_before_upgrade;
+
 /** Bit flags for tracking background thread creation. They are used to
 determine which threads need to be stopped if we need to abort during
 the initialisation step. */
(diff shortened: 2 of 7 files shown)


share/messages_to_error_log.txt
```diff
@@ -11696,6 +11696,10 @@ ER_XB_MSG_6
 ER_XB_UNDO_DELETE_FAILURE
   eng "Deletion of 5.7 undo tablespace %s failed"
 
+ER_XB_MSG_DOWNGRADING_LOG_FILE
```
ldonoso: Please, maybe we should create the new error codes at the end of Percona's list.

It might not be important, but this way, we do not change the numeric code of the existing errors. Please, think that there might be scripts checking the error numeric code instead of the text representation.

storage/innobase/srv/srv0start.cc
```diff
@@ -2688,6 +2756,7 @@ dberr_t srv_start(bool create_new_db) {
     }
   }
 
+
```
ldonoso: Please, remove this extra line. Sorry, I introduced this one.

storage/innobase/srv/srv0start.cc
```diff
@@ -580,11 +584,82 @@ static void create_log_files_rename(
...
+  fil_space_t::s_redo_space = fil_space_get(dict_sys_t::s_log_space_first_id);
 
   /* For cloned database it is normal to resize redo logs. */
   ib::info(ER_IB_MSG_1068, ulonglong{lsn});
 }
 
+/* create the log file name
+@param[in/out] logfilename  buffer for log file name
+retrun the dir name lenght */
+static size_t create_log_file_name(char *logfilename) {
```
ldonoso: Please, mind we need to pass the size of `logfilename`. Previously, the `sizeof` in this function worked fine but now `sizeof logfilename` will always return 8 as `logfilename` is of type `char *`

ldonoso: We need to passed it with `[[maybe_unused]]` as the size is used just in `ut_a`

storage/innobase/srv/srv0start.cc
```diff
@@ -580,11 +584,82 @@ static void create_log_files_rename(
...
+
+  fil_flush_file_spaces(FIL_TYPE_TABLESPACE);
+
+  create_log_files_rename(logfilename, dirnamelen, new_checkpoint_lsn,
+                          logfile0);
+  ut_d(log_sys->disable_redo_writes = false);
+
+  flushed_lsn = new_checkpoint_lsn;
+
+  log_start(*log_sys, 0, flushed_lsn, flushed_lsn);
```
ldonoso, line 658: It is really cool we can remove redundant code by using this function in two places.

But mind this function is used in `srv_shutdown`, please, do we need to call to `log_start`?

ldonoso (changes requested):

share/messages_to_error_log.txt
rahulmalik87: fixed.

storage/innobase/srv/srv0start.cc
rahulmalik87: fixed.

storage/innobase/srv/srv0start.cc
rahulmalik87: ut_a is release assert, Is'nt it will always be called, even it debug or release build?

storage/innobase/srv/srv0start.cc
rahulmalik87: log_start is required to intialize in-memory object. If we don't call log_start it crashes srv0start.cc:3986:log_lsn_validate(lsn) |

storage/innobase/srv/srv0start.cc
ldonoso: Sorry, that's true

ldonoso (approved):
LGTM. Please, I am just qualified to review general errors. We need an additional approval from a expert to verify the logic to recreate the redo log files.

share/messages_to_error_log.txt
```diff
@@ -11771,6 +11771,9 @@ ER_SEC_LOG_PATH_NULL
 ER_KEYRING_COMPONENT_KMS_ERROR
   eng "KMS communication error: %s"
 
+ER_XB_MSG_DOWNGRADING_LOG_FILE
+  eng "Upgrade failed, cleaning up redo logs, so server can start with previous version"
```
satya-bodapati: Upgrade failed, recreating redo logs with the previous version

storage/innobase/srv/srv0start.cc
```diff
@@ -580,11 +584,82 @@ static void create_log_files_rename(
...
+  /* Close and free the redo log files, so that
+  we can replace them. */
+  fil_close_log_files(true);
+
+  RECOVERY_CRASH(5);
+
+  log_sys_close();
+
+  /* Finish clone file recovery before creating new log files. We
+      roll forward to remove any intermediate files here. */
```
satya-bodapati: comment alignment (of the line roll..)

storage/innobase/srv/srv0start.cc
```diff
@@ -3913,10 +3954,30 @@ static lsn_t srv_shutdown_log() {
 
   srv_wake_log_tracker_thread();
 
-  if (srv_downgrade_logs) {
+  /* if upgrade is failed, delete and create new empty redo so old version do
+   not complaint  */
```
satya-bodapati: If the upgrade fails, new redo log format is not compatible with older versions, recreate them

storage/innobase/srv/srv0start.cc
```diff
@@ -3913,10 +3954,30 @@ static lsn_t srv_shutdown_log() {
...
+
+    auto flushed_lsn = log_get_lsn(*log_sys);
+
+    ib::info(ER_XB_MSG_DOWNGRADING_LOG_FILE);
+
+    auto err =
+        recreate_redo_logs(logfilename, dirnamelen, logfile0,
+                           new_checkpoint_lsn, flushed_lsn, srv_n_log_files);
+
+    ut_a(err == DB_SUCCESS);
```
satya-bodapati: can be debug

storage/innobase/srv/srv0start.cc
```diff
@@ -580,11 +584,82 @@ static void create_log_files_rename(
...
+  auto err =
+      create_log_files(logfilename, dirnamelen, flushed_lsn,
+                       number_of_log_files, logfile0, new_checkpoint_lsn);
+
+  if (err != DB_SUCCESS) {
+    return err;
+  }
+
+  /* create_log_files() can increase system lsn that is
+      why FIL_PAGE_FILE_FLUSH_LSN have to be updated */
```
satya-bodapati: create_log_files() can increase LSN, update the FIL_PAGE_FILE_FLUSH_LSN to the last LSN

satya-bodapati (changes requested):

share/messages_to_error_log.txt
rahulmalik87: fixed

storage/innobase/srv/srv0start.cc
rahulmalik87: fixed

storage/innobase/srv/srv0start.cc
rahulmalik87: fixed

storage/innobase/srv/srv0start.cc
rahulmalik87: fixed

storage/innobase/srv/srv0start.cc
rahulmalik87: fixed

satya-bodapati (approved):

ldonoso (approved):

PR #4827, opened 2022-06-22
PS-8292 : MySQL 8.0.29 fails to perform crash recovery

satya-bodapati:
PS-8292 : MySQL 8.0.29 fails to perform crash recovery

https://jira.percona.com/browse/PS-8292

Problem 1:
----------
Crash after an INSTANT ADD/ DROP COLUMN DDL followed by DMLS make the
InnoDB corrupt tables or lead to irrecoverable instances. (ie.crash recovery
will never succeed)

Crash recovery asserts is because, recovery is trying to update the wrong
field. This causes the assertion failure len2 == len. The update field value
and the field it tries to update are different. Hence, it crashes.

For example, for this schema:

CREATE TABLE `tb1` ( `col1` VARCHAR(10) NOT NULL, `col2` char(13), `col3`
varchar(10), PRIMARY KEY (`col1`) ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4
COLLATE=utf8mb4_0900_ai_ci;

the record lay out (before drop) is
col1 (0), DB_TRX_ID (1), DB_ROLL_PTR(2), col2(3), col3 (4)
this is clustered index layout. (The physical pos in braces ().)

After ALTER TABLE tb1 DROP COLUMN col2, LOCK=DEFAULT;

record layout still remains same. ie col1 (0), DB_TRX_ID(1), DB_ROLL_PTR(2),
col2_dropped_v1(3), col3 (4).

After drop column col2, the clustered index metadata now has one less field.
So clustered index layout is

col1(0), DB_TRX_ID(1), DB_ROLL_PTR(2), col3 (4) [ col2_droppeed_v1(3) is
just present in metadata)

So the logical position of col3 is now 3 field in index but physical position
of the field is 4.

To update a field in the record, we do it by the field_no in the index.
This I would call it logical position of the field. But the physical_pos in the
record is different.

We only log the "logical" pos of the field to be updated (field_no). But during recovery,
we don't have all the information to build the physical pos of the index_fields.

The updated field (based on logical pos), will now update the wrong field during recovery.

In the above case, field number 3 is logged to be updated. The record layout is 

col1 (0), DB_TRX_ID (1), DB_ROLL_PTR(2), col2_dropped(3), col3 (4)

Without the phy_pos mapping, update now tries to update the field at col2_dropped(3)

Fix:
----
For updated fields, we can log directly the phys_pos of the field. The record to be
logged is already logged in physical order

Problem 2:
----------
We also believe INSTANT ADD COLUMN / DROP COLUMN is unstable. There are multiple
bugs caused by this feature

Fix for Problem 1:
------------------
We move the dropped columns that are not part of index to the last. This way the
correct fields are part of the index and recovery will update the right field.

Fix for Problem 2:
------------------
ALTER ADD COLUMN/DROP COLUMN without explicit ALGORITHM keyword, or ALGORITHM=DEFAULT
would use INPLACE/COPY instead of INSTANT.

ALGORITHM=INSTANT, will use the new INSTANT ADD/DROP COLUMN feature.


altmannmarcelo (approved):
LGTM - tested PXB also and it works fine.

satya-bodapati:
mysql-test-run.pl  --mem --suites=innodb,innodb_zip,innodb_fts,innodb_gis,innodb_undo,clone --retry-failure=0 --parallel=16 --force --max-test-fail=500
MySQL Version 8.0.29
Checking supported features
 - Binaries are debug compiled
..

The servers were restarted 197 times
The servers were reinitialized 29 times
Spent 25027.663 of 2014 seconds executing testcases

Completed: Failed 5/943 tests, 99.47% were successful.

Failing test(s): shutdown_report innodb.percona_changed_page_bmp_flush innodb.percona_changed_page_bmp_requests_debug innodb.percona_changed_pages innodb.percona_changed_page_bmp_debug

The log files in var/log may give you some hint of what went wrong.

If you want to report this error, please read first the documentation
at http://dev.mysql.com/doc/mysql/en/mysql-test-suite.html

148 tests were skipped, 56 by the test itself.

altmannmarcelo (approved):
LGTM

PR #4828, opened 2022-06-22
Fixed PS-8303 - MySQL 8.0.29 - Assertion failure: dict0mem.h:2498:pos…

altmannmarcelo:
… < n_def

https://jira.percona.com/browse/PS-8303

Problem:
dict_index_t / dict_table_t have metadata fields such as n_def,
n_fields, n_total_fields defined with bit-field explicit size of 10
bits, meaning they can hold up to 1023 then operations such as n_def++
will roll over back to 0.
New ADD/DROP INSTANT work makes each row be versioned after an INSTANT
ALTER and dropped columns are not removed, they are kept invisible to
users. If a user keeps adding and dropping columns n_def will still
hold the fields that have been dropped even if those fields are hidden
from the user and eventually, the counter will wrap around.

Fix:
Adjust alter handler to take this scenario into consideration and act
according to the algorithm selected:

1) DEFAULT / Not specified - we now default add/drop columns to INSTANT.
If a user has not specified an algorithm or has specified DEFAULT we
will fall back to INPLACE in case n_def has exceeded REC_MAX_N_FIELDS.
2) INSTANT - if a user has explicitly requested the INSTANT algorithm, we
error out with ER_TOO_MANY_FIELDS (Too many columns).


satya-bodapati (commented):
@altmannmarcelo I think this can be moved to innobase_support_instant()? Also please merge this PR after my PR. it will likely have a conflict with mine.

altmannmarcelo:
> @altmannmarcelo I think this can be moved to innobase_support_instant()? Also please merge this PR after my PR. it will likely have a conflict with mine.

@satya-bodapati as per slack conversation, at innobase_support_instant we return `Instant_Type`, we still want to return `ER_TOO_MANY_FIELDS` in case user specifically asked for algorithm INSTANT

satya-bodapati (approved):

PR #4829, opened 2022-06-28
Fixed PS-8291 - MySQL 8.0.29 crashes with ALTER DROP

altmannmarcelo:
MySQL 8.0.29 crashes with ALTER DROP

Problem:
TRUNCATE will erase all data from a table, it will also remove dropped
instant columns and reorder instant added columns from dict_index_t.
However, DD entry for the table was not removing dropped columns. It was just
removing se_private_date proprieties. Later when dd_copy_table_columns
identified the columns was a hidden column due to instant drop, it tried to
validate the column had se_private_data DD_INSTANT_VERSION_DROPPED set.
This meant that DD and dict_index_t can get out-of-sync after truncate.

Fix:
Adjust dd_clear_instant_table function to remove dropped columns from DD
entry of the table.


storage/innobase/dict/dict0dd.cc
```diff
@@ -1689,8 +1689,11 @@ void dd_copy_table_columns(const Alter_inplace_info *ha_alter_info,
   for (const auto old_col : old_table.columns()) {
     if (old_col->is_se_hidden() && !is_system_column(old_col->name().c_str()) &&
         (strcmp(old_col->name().c_str(), FTS_DOC_ID_COL_NAME) != 0)) {
-      /* Must be an alredy dropped column. */
-      ut_ad(dd_column_is_dropped(old_col));
+      /* TRUNCATE will erase INSTANT ADD/DROP information */
+      if (!first_row_version) {
+        /* Must be an alredy dropped column. */
+        ut_ad(dd_column_is_dropped(old_col));
+      }
```
satya-bodapati: @altmannmarcelo 
the only time a col is set hidden (this has different meaning at server layer) in InnoDB is this

```
* Add hidden FTS_DOC_ID column */
    dd::Column *col = new_table.add_column();
    col->set_hidden(dd::Column::enum_hidden_type::HT_HIDDEN_SE);
    col->set_name(FTS_DOC_ID_COL_NAME);
    col->set_type(dd::enum_column_types::LONGLONG);
    col->set_nullable(false);
    col->set_unsigned(true);
    col->set_collation_id(1);
  
```
Do we set this set_se_hidden() for instant dropped cols as well? if so, where? :thinking:

altmannmarcelo: decided to follow a different approach by removing dropped columns from DD so the entry from DD is in sync with the index object.

mysql-test/suite/innodb/t/instant_ddl_truncate_debug.test
```diff
@@ -0,0 +1,25 @@
...
+col1 INT,
+col2 INT,
+col3 INT
+) ENGINE=InnoDB ;
+ALTER TABLE tb1 DROP COLUMN col2, ALGORITHM=INSTANT;
+TRUNCATE TABLE tb1 ;
+ALTER TABLE tb1 DROP COLUMN col3, ALGORITHM=INSTANT;
+
+# clean-up
+DROP TABLE tb1;
```
satya-bodapati: Can you please extend to partitions as well?

With partitions, we want a single partition to be truncated. See if other versions data in other partitions is accessible. After that, we should be able to add and drop instant columns on all partitions.

satya-bodapati (changes requested):
Fix looks good. One test modifications requested

mysql-test/suite/innodb/t/instant_ddl_truncate_debug.test
altmannmarcelo: Ack, done @satya-bodapati

mysql-test/suite/innodb/t/instant_ddl_truncate_debug.test
```diff
@@ -0,0 +1,57 @@
+##########################################################################
+# 107613 - Assertion failure: dict0dd.cc:1685:dd_column_is_dropped(old_col)
+##########################################################################
+
+--source include/have_debug.inc
```
satya-bodapati: do we need debug?

mysql-test/suite/innodb/t/instant_ddl_truncate_debug.test
```diff
@@ -0,0 +1,57 @@
...
+);
+ALTER TABLE tb1_part DROP COLUMN col2, ALGORITHM=INSTANT;
+INSERT INTO tb1_part VALUES (1,3,4, '2017-01-01');
+INSERT INTO tb1_part VALUES (1,3,4, '2018-01-01');
+INSERT INTO tb1_part VALUES (1,3,4, '2019-01-01');
+INSERT INTO tb1_part VALUES (1,3,4, '2020-01-01');
+INSERT INTO tb1_part VALUES (1,3,4, '2021-01-01');
+INSERT INTO tb1_part VALUES (1,3,4, '2022-01-01');
+SELECT * FROM tb1_part PARTITION (p1);
+TRUNCATE TABLE tb1_part ;
```
satya-bodapati: please dont truncate all paritions first. May be at end ;)

satya-bodapati (changes requested):

mysql-test/suite/innodb/t/instant_ddl_truncate_debug.test
altmannmarcelo: Assertion is ut_ad

satya-bodapati: that is ok, its good to test release as well

satya-bodapati: we force debug only, when we use SET DEBUG or things like that

satya-bodapati (approved):
Please squash and merge. Ensure clean suites (at least locally)

altmannmarcelo:
```
$ LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libeatmydata.so mysql-test/mtr --suite innodb --big-test --parallel=24 --force 
 . . .
The servers were restarted 172 times
The servers were reinitialized 20 times
Spent 21556.034 of 1497 seconds executing testcases

Completed: All 787 tests were successful.

50 tests were skipped, 50 by the test itself.

```

satya-bodapati (approved):

PR #4830, opened 2022-06-28
PS-8255: Fido authentication

dutow:
This change adds a plugin named authentication_fido to the server,
which uses the already present client plugin with the same name.

For documentation, see
https://dev.mysql.com/doc/refman/8.0/en/fido-pluggable-authentication.html

### DIFF
--- a/plugin/authentication_fido/CMakeLists.txt
+++ b/plugin/authentication_fido/CMakeLists.txt
@@ -1,0 +1,38 @@
+#Copyright (c) 2022 Percona LLC and/or its affiliates. All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License, version 2.0,
+# as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License, version 2.0, for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA
+
+IF(NOT FIDO_FOUND)
+  MESSAGE(WARNING "Skipping the FIDO server authentication plugin")
+  RETURN()
+ENDIF()
+
+SET(AUTH_FIDO_SOURCES
+  src/plugin_fido.cc
+)
+### Configuration ###
+ADD_DEFINITIONS(-DLOG_COMPONENT_TAG="authentication_fido")
+
+INCLUDE_DIRECTORIES(SYSTEM ${BOOST_PATCHES_DIR} ${BOOST_INCLUDE_DIR})
+
+MY_INCLUDE_SYSTEM_DIRECTORIES(FIDO)
+
+DISABLE_MISSING_PROFILE_WARNING()
+
+MYSQL_ADD_PLUGIN(auth_fido ${AUTH_FIDO_SOURCES}
+  LINK_LIBRARIES
+  ${FIDO_LIBRARY}
+  ${OPENSSL_LIBRARIES}
+  MODULE_ONLY MODULE_OUTPUT_NAME "authentication_fido"
+  )
--- a/plugin/authentication_fido/src/plugin_fido.cc
+++ b/plugin/authentication_fido/src/plugin_fido.cc
@@ -1,0 +1,425 @@
+/* Copyright (c) 2022 Percona LLC and/or its affiliates. All rights reserved.
+
+   This program is free software; you can redistribute it and/or
+   modify it under the terms of the GNU General Public License
+   as published by the Free Software Foundation; version 2 of
+   the License.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA */
+
+#include <array>
+#include <condition_variable>
+#include <mutex>
+
+#include "base64.h"
+#include "m_string.h"
+#include "my_dbug.h"
+#include "my_sys.h"
+#include "mysql/components/my_service.h"
+#include "mysql/components/service_implementation.h"
+#include "mysql/components/services/mysql_authentication_registration.h"
+#include "mysql/plugin_auth.h"
+#include "mysqld_error.h"
+#include "sql/current_thd.h"
+#include "sql/sql_class.h"
+
+#include <fido.h>
+#include <fido/es256.h>
+
+#include <openssl/ec.h>
+#include <openssl/rand.h>
+
+static char *rpid;
+static MYSQL_SYSVAR_STR(
+    rp_id, rpid, PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_MEMALLOC,
+    "The relying party ID used for FIDO device registration and authentication",
+    nullptr, nullptr, "MySQL");
+
+struct SYS_VAR;
+static SYS_VAR *fido_sysvars[] = {MYSQL_SYSVAR(rp_id), nullptr};
+
+struct SYS_VAR {
+  MYSQL_PLUGIN_VAR_HEADER;
+};
+
+MYSQL_PLUGIN authentication_fido_pluogin_info;
+
+using fido_pubkey_t = std::array<unsigned char, 64>;
+bool reconstruct_fido_cred(fido_cred_t *cred, const char *data,
+                           std::size_t data_len, fido_pubkey_t &pubkey);
+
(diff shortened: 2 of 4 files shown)


plugin/authentication_fido/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+### Configuration ###
+ADD_DEFINITIONS(-DLOG_COMPONENT_TAG="authentication_fido")
+
+INCLUDE_DIRECTORIES(SYSTEM ${BOOST_PATCHES_DIR} ${BOOST_INCLUDE_DIR})
+
+MYSQL_ADD_PLUGIN(auth_fido ${AUTH_FIDO_SOURCES}
+  LINK_LIBRARIES
+  fido2
+  ${OPENSSL_LIBRARIES}
+  MODULE_ONLY MODULE_OUTPUT_NAME "authentication_fido"
```
altmannmarcelo, line 37: We need to make sure if we do it using dynamic linkage, that we link it to the correct library from private (the one we compiled).
Ubuntu 20.04 for example, only has fido2 1.3

plugin/authentication_fido/src/plugin_fido.cc
```diff
@@ -0,0 +1,414 @@
...
+#include <openssl/rand.h>
+
+struct SYS_VAR;
+static SYS_VAR *fido_sysvars[] = {nullptr};
+
+struct SYS_VAR {
+  MYSQL_PLUGIN_VAR_HEADER;
+};
+
+static const char *rpid = "MySQL";
```
altmannmarcelo: This should be configurable and users have to make it unique across the fleet.

plugin/authentication_fido/src/plugin_fido.cc
```diff
@@ -0,0 +1,414 @@
...
+         current_thd->security_context()->host().length);
+  pos += current_thd->security_context()->host().length;
+
+  pos += user_name_len;
+
+  assert(pos - data < 512);
+
+  // client expects base64 encoded data
+  (*outbuf) = new unsigned char[256];
+  base64_encode(data, pos - data, (char *)*outbuf, false);
```
altmannmarcelo, line 96: host field alone can have up to 255 characters. Any chance of overflow here?

plugin/authentication_fido/src/plugin_fido.cc
```diff
@@ -0,0 +1,414 @@
...
+  memcpy(pos, current_thd->security_context()->host().str,
+         current_thd->security_context()->host().length);
+  pos += current_thd->security_context()->host().length;
+
+  pos += user_name_len;
+
+  assert(pos - data < 512);
+
+  // client expects base64 encoded data
+  (*outbuf) = new unsigned char[256];
```
altmannmarcelo: Do we need to free outbuf?

plugin/authentication_fido/src/plugin_fido.cc
```diff
@@ -0,0 +1,414 @@
...
+      reinterpret_cast<my_h_service>(
+          const_cast<mysql_authentication_registration_t *>(
+              &SERVICE_IMPLEMENTATION(authentication_fido,
+                                      mysql_authentication_registration))));
+
+  mysql_plugin_registry_release(plugin_registry);
+  return result;
+}
+
+static int fido_deinit(MYSQL_PLUGIN plugin_info __attribute__((unused))) {
```
altmannmarcelo, line 181: on some function signatures we use ` __attribute__((unused))` and on others we use `/*VAR*/` . Can you adjust to use one form only?

altmannmarcelo (changes requested):
apart from code comments, testing the plugin, if I remove the USB stick and try to connect, mysql client crashes. It seems its a client issue trying to open a device that is not connected.

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7080cb7 in ?? () from /lib/x86_64-linux-gnu/libfido2.so.1
(gdb) bt
#0  0x00007ffff7080cb7 in ?? () from /lib/x86_64-linux-gnu/libfido2.so.1
#1  0x00007ffff70726b5 in fido_dev_open () from /lib/x86_64-linux-gnu/libfido2.so.1
#2  0x00007ffff7fc3e8f in fido_prepare_assert::sign_challenge (this=0x555555e700e0) at /work/ps/src/8.0/libmysql/authentication_fido/fido_assertion.cc:111
#3  0x00007ffff7fc43a0 in fido_assertion::sign_challenge (this=0x555555e700e0) at /work/ps/src/8.0/libmysql/authentication_fido/fido_assertion.cc:245
#4  0x00007ffff7fc4776 in fido_auth_client (vio=0x7fffffffbf00) at /work/ps/src/8.0/libmysql/authentication_fido/fido_client_plugin.cc:163
#5  0x00005555555be353 in authsm_run_second_authenticate_user (ctx=0x7fffffffbec0) at /work/ps/src/8.0/sql-common/client.cc:5834
#6  0x00005555555bd1ab in run_plugin_auth (mysql=0x555555e023a0 <mysql>, data=0x555555e4b630 "nj\001.64yp5cs\006\033bc\023\\A[g", data_len=21, data_plugin=0x555555e5477d "\005MySQL@sXi\264G>\254\313K\224\026ς,PP\377\231'6\017\004\016\063ٚ\273\321\034\020\200\001\214\271\245^eAJ\216\324\325\062\263\272\364k?1\004\"z\r`\327\"\035NMvyI4u", db=0x0) at /work/ps/src/8.0/sql-common/client.cc:5505
#7  0x00005555555c2719 in csm_authenticate (ctx=0x7fffffffc0b0) at /work/ps/src/8.0/sql-common/client.cc:6923
#8  0x00005555555bf3ee in mysql_real_connect (mysql=0x555555e023a0 <mysql>, host=0x555555e22e20 "127.0.0.1", user=0x555555e22df0 "u1", passwd=0x0, db=0x0, port=3411, unix_socket=0x0, client_flag=66560) at /work/ps/src/8.0/sql-common/client.cc:6088
#9  0x00005555555987c2 in sql_real_connect (host=0x555555e22e20 "127.0.0.1", database=0x0, user=0x555555e22df0 "u1", silent=0) at /work/ps/src/8.0/client/mysql.cc:4527
#10 0x0000555555598f2e in sql_connect (host=0x555555e22e20 "127.0.0.1", database=0x0, user=0x555555e22df0 "u1", silent=0) at /work/ps/src/8.0/client/mysql.cc:4715
#11 0x000055555558f4dd in main (argc=7, argv=0x555555e204b0) at /work/ps/src/8.0/client/mysql.cc:1337
```

I'm using libfido2 from ppa:yubico

plugin/authentication_fido/CMakeLists.txt
dutow: We discussed this on slack, we'll ship a dynamic library somehow.

plugin/authentication_fido/src/plugin_fido.cc
dutow: Changed, now it's `authentication_fido_rp_id`

plugin/authentication_fido/src/plugin_fido.cc
dutow: Updated the buffer to 512, this was a typo there, earlier I've used a 512 byte buffer, which should be enough.

plugin/authentication_fido/src/plugin_fido.cc
dutow: No, it's freed by the server code later.

plugin/authentication_fido/src/plugin_fido.cc
dutow: Updated all to use the attribute.

altmannmarcelo (changes requested):
I'm still getting crash when FIDO device is not plugged.
From https://developers.yubico.com/libfido2/Manuals/fido_dev_info_new.html 
>  The fido_dev_info_manifest() function always returns FIDO_OK. If a discovery error occurs, the olen pointer is set to 0. 

It seems that we need to adjust `fido_prepare_assert::sign_challenge` to also check for `dev_infos_len`

```
git diff libmysql/authentication_fido/fido_assertion.cc
diff --git a/libmysql/authentication_fido/fido_assertion.cc b/libmysql/authentication_fido/fido_assertion.cc
index 946691015f5..dd76a574e8e 100644
--- a/libmysql/authentication_fido/fido_assertion.cc
+++ b/libmysql/authentication_fido/fido_assertion.cc
@@ -98,7 +98,7 @@ bool fido_prepare_assert::sign_challenge() {
   fido_init(0);
   size_t dev_infos_len = 0;
   fido_dev_info_t *dev_infos = fido_dev_info_new(1);
-  if (fido_dev_info_manifest(dev_infos, 1, &dev_infos_len) != FIDO_OK) {
+  if (fido_dev_info_manifest(dev_infos, 1, &dev_infos_len) != FIDO_OK || dev_infos_len == 0) {
     fido_dev_info_free(&dev_infos, 1);
     get_plugin_messages("No FIDO device available on client host.",
                         message_type::ERROR);
```

dutow:
@altmannmarcelo I left that alone as we usually use the upstream mysql-client, fixing something in our code doesn't make much sense. We can create an upstream bug and submit a patch, and maybe they accept it, but fixing it in our code won't have much effect.

altmannmarcelo:
@dutow raised that with upstream https://bugs.mysql.com/bug.php?id=107894
I still think that leaving it broken is not the right thing to do having the user experience in mind. We know there is a crash, and there are users that use 100% Percona server across the fleet, those users will be affected by us not fixing it.

altmannmarcelo (approved):
LGTM

PR #4831, opened 2022-06-30
Fixed PS-8306 - Test innodb.create_isl_with_direct fails if mysql-test-run.pl is used with "--mem" parameter

i-rinat:
https://jira.percona.com/browse/PS-8306

Problem:
Test innodb.create_isl_with_direct fails when run on tmpfs. Running on
tmpfs causes InnoDB to generate a warning about missing support of
O_DIRECT. The test has a suppression for that exact warning, but it's
off by one space character.

Solution:
Update the warning suppression regex by adding one more space.


kamil-holubicki (approved):
lgtm

PR #4832, opened 2022-06-30
Follwup fix PS-8292 : MySQL 8.0.29 fails to perform crash recovery

satya-bodapati:
https://jira.percona.com/browse/PS-8292

Problem:
-------
The update fields logged to redo now use field phy pos.
We used the field_phy_pos stored in the update vector.

But this is not always calculated and also only available in
debug builds

Fix:
----
We retrieve the physical position using the index using the field_no
number from update vector.

Also added new assertions to catch problem early (before recovery)

### DIFF
--- a/storage/innobase/row/row0upd.cc
+++ b/storage/innobase/row/row0upd.cc
@@ -614,6 +614,7 @@ void row_upd_index_write_log(dict_index_t *index, const upd_t *update,
 
   log_ptr += mach_write_compressed(log_ptr, log_fields);
 
+  IF_DEBUG(std::vector<ulint> used_phy_pos;)
   for (i = 0; i < n_fields; i++) {
     static_assert(MLOG_BUF_MARGIN > 30, "MLOG_BUF_MARGIN <= 30");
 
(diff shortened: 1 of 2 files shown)


storage/innobase/row/row0upd.cc
```diff
@@ -635,17 +636,26 @@ void row_upd_index_write_log(dict_index_t *index, const upd_t *update,
...
-    indexes that do not have row_versions */
-    ulint field_off = index->has_row_versions() ? upd_field->field_phy_pos
-                                                : upd_field->field_no;
+    ulint field_phy_pos =
+        index->get_field(upd_field->field_no)->col->get_col_phy_pos();
+
+#ifdef UNIV_DEBUG
+    /* Check if multiple fields are updated, they should be different fields
+    in row */
+    if (j > 0) {
```
altmannmarcelo: Do we need `j` here ?
What if we reach continue at `if (upd_fld_is_virtual_col(upd_field) && !dict_index_has_virtual(index))` whouldn't `j` be pointing to the wrong filed?

Perhaps we can use `i` instead

altmannmarcelo: Other question. The assertion is validating if my current phy_pos is different from previous field pos. 
For example on below table:
```
ID INT pk
col2 INT
col3 INT
col4 INT
``` 

This will have [physical][logical]:
``` 
ID                       [0][0]
DB_TRX                   [1][1]
DB_ROLL_PTR              [2][2]
col2                     [3][3]
col3                     [4][4]
col4                     [5][5]
```
After a DROP col2 INSTANT
```
ID                       [0][0]
DB_TRX                   [1][1]
DB_ROLL_PTR              [2][2]
col3                     [4][3]
col4                     [5][4]
col2                     [3][5]
```

What we were seeing before at crash recover was col3 and col2 having the same phy_pos, but they are 2 logical positions apart now. So I don't this that comparing only with prev field will be enough here.

altmannmarcelo (changes requested):

storage/innobase/row/row0upd.cc
satya-bodapati: > Do we need `j` here ? What if we reach continue at `if (upd_fld_is_virtual_col(upd_field) && !dict_index_has_virtual(index))` whouldn't `j` be pointing to the wrong filed?
> 
> Perhaps we can use `i` instead
@altmannmarcelo 

if i got incremented due to virtual col, then next interaction, we will compare with skipped field. So is the need for j.

But i see your point about checking only with 'consecutive' fields. Typically they were all zero. So as such it is OK but I will improve the logic and get rid of this j and search across all used fields.

altmannmarcelo (approved):
LGTM

PR #4836, opened 2022-07-14
PS-8128: In-place upgrade of views that select from performance_schem…

voodam:
…a is broken

https://jira.percona.com/browse/PS-8128

View is broken because when it created, perfromance_schema wasn't imported yet. Here we swap code populating DD tables and code initializing performance_schema tables. This allows to upgrade views that refer to performance_schema tables.

I runned MTR tests from `perfschema` suite + all named `*upgrade*` test and didn't found any differences with 8.0.

Thoughs to consider:
- Are we safe, nothing can become broken here? Is that order of initialization code was arbitrary?
- Maybe it will be better not fix the issue, but add some notes to the documentation: https://dev.mysql.com/doc/refman/8.0/en/upgrade-prerequisites.html?


kamil-holubicki:
I think this change is fine. What is your opinion @satya-bodapati ?
@voodam could we introduce MTR test for this?

voodam:
> @voodam could we introduce MTR test for this?

I should add 5.7 datadir to tests data for this, but probably yes, I'm working on this.

voodam:
> @voodam could we introduce MTR test for this?

Done. I just added tests, so I used squash commit, I hope this doesn't make the review process less convinient. Also I can't re-request review, because there weren't code changes requested, so I hope comments are also delivered.

sql/mysqld.cc
```diff
@@ -6728,6 +6697,28 @@ static int init_server_components() {
...
 
+  if (!is_help_or_validate_option() && dd::upgrade_57::in_progress()) {
+    // Populate DD tables with meta data from 5.7
+    if (dd::init(dd::enum_dd_init_type::DD_POPULATE_UPGRADE)) {
+      LogErr(ERROR_LEVEL, ER_DD_POPULATING_TABLES_FAILED);
+      unireg_abort(1);
+    }
+    // Run after_dd_upgrade hook
+    if (RUN_HOOK(server_state, after_dd_upgrade_from_57, (nullptr)))
+      unireg_abort(MYSQLD_ABORT_EXIT);
```
satya-bodapati, line 6708: Am not sure if PFS tables depend on DD to be initialized or not.

@dlenev can you please check and confirm if this is OK?

dlenev: Hello!

Here are my comments about this issue after some digging:

Re-creation of PFS tables indeed depends on that DD tables exist (we create DD tables during dd::init(DD_RESTART_OR_UPGRADE). However, it should not depend on the fact whether information about tables in user databases was added to DD (this is what happens in dd::init(DD_POPULATE_UPGRADE) or not by the time PFS tables re-created. In fact we do pretty similar thing with I_S views - they are created early in the DD construction process, right after DD tables created themselves. So this patch is OK in this respect.

Now some thoughts about the general approach used in the patch/possible alternatives:

AFAICS the problem with upgrade of view dependent on P_S occurs due to combination of two factors:

1) P_S tables are skipped when we add information about user tables to new DD during dd::init(DD_POPULATE_UPGRADE). They are re-created in DD at later stage. So user view which depends on P_S tables can be considered missing underlying tables/broken during this dd::init() call.
2) The process of user view upgrade which happens during dd::init(DD_POPULATE_UPGRADE) consists of two steps: During the first step we simply add definition of view to DD. During the second step we evaluate information about column types of view and tables/routines on which view depends and add it to DD. The interesting fact that if view is broken/there are missing underlying tables for view, then the 1st step succeeds, but the 2nd fails and do not add even information about view dependencies to DD (which in theory should be still possible).

As result of the above information about user view which depends on P_S table is added to DD during upgrade, but information about its dependency on P_S table is not. This leads to this view not being detected as needing re-validation at the point when P_S tables are re-created and is incorrectly left as invalid.

The same issue will actually apply to any user view which depend on table which was dropped at some point. During upgrade its dependency on missing table won't be reflected in DD and it will stay invalid forever, even after its underlying table is back.

The proposed patch solves the problem reported by addressing factor 1) mentioned above, making user view dependent on P_S valid during dd::init(DD_POPULATE_UPGRADE) call.

One possible alternative approach is to address factor 2), by changing 2nd step of view upgrade process to store in DD dependencies even for invalid views. In theory this should also address issue with invalid views becoming irreparable after upgrade. However, such patch is likely to be more complex. And I doubt that problem with upgrading invalid views is really important for our users.

Taking into account the above I think it is OK to proceed with this patch (possibly leaving approach 2) for upstream to implement/postponing its implementation until real user need for it arises).

mysql-test/t/ps8128_upgrade_perfschema_view.test
```diff
@@ -0,0 +1,38 @@
...
+--source include/not_valgrind.inc
+
+--echo # Prepare datadir
+--let $file= data57_upgrade_perfschema_view
+--copy_file $MYSQLTEST_VARDIR/std_data/$file.zip $MYSQL_TMP_DIR/$file.zip
+--file_exists $MYSQL_TMP_DIR/$file.zip
+--exec unzip -qo $MYSQL_TMP_DIR/$file.zip -d $MYSQL_TMP_DIR/$file
+--let $MYSQLD_DATADIR1= $MYSQL_TMP_DIR/$file
+--let $MYSQLD_LOG= $MYSQLTEST_VARDIR/log/$file.log
+--replace_result $MYSQLD MYSQLD $MYSQLD_DATADIR1 MYSQLD_DATADIR1 $MYSQLD_LOG MYSQLD_LOG
```
dlenev: Is this line still necessary? Perhaps better to remove it?

dlenev (approved):

mysql-test/t/ps8128_upgrade_perfschema_view.test
voodam: No, was copy-pasted, removed.

sql/mysqld.cc
satya-bodapati: one question I have for @voodam is that if there is an invalid view in 5.7, does the upgrade fail?
I guess one can create an invalid view by dropping underlying tables (user table)?

voodam: In case of dropping user tables behaviour will be the same: server shows warning, but import will be successful:
`[Warning] [MY-010200] [Server] Resolving dependency for the view 'test.v1' failed. View is no more valid to use`

satya-bodapati: @voodam there is a tool called upgrade checker https://dev.mysql.com/doc/mysql-shell/8.0/en/mysql-shell-utilities-upgrade.html
please check if this detects such invalid views

voodam: @satya-bodapati, no, there are no relevant warnings after, I used 8.0.29 version of checking tool.

satya-bodapati: @voodam please create two bugs against upstream
1. invalid user views are upgraded silently. mysqldump fails (same bug as this). show that it is not specific pfs tables
2. upgrade checker failing to catch such invalid views

voodam: @satya-bodapati Done. I supposed that using existing issue is better than creating new one in both cases:

https://bugs.mysql.com/bug.php?id=106819
https://bugs.mysql.com/bug.php?id=108005

kamil-holubicki (approved):
lgtm

PR #4839, opened 2022-07-22
PS-8327 ALTER TABLE ... CHECK PARTITION inside stored procedure causes problem

voodam:
https://jira.percona.com/browse/PS-8327

## How to reproduce

```
CREATE TABLE t1(i INT PRIMARY KEY)
PARTITION BY RANGE (i)
(
  PARTITION p0 VALUES LESS THAN (2),
  PARTITION p1 VALUES LESS THAN (4)
);

DELIMITER $$
CREATE PROCEDURE p1()
BEGIN
  ALTER TABLE t1 CHECK PARTITION p0;
END $$
DELIMITER ;

CALL p1();
```

On 5.7: Get an assertion fail on the client on the next command entering:

`sql/net_serv.cc:234: void net_clear(NET*, my_bool): Assertion '!check_buffer || (vio_pending(net->vio) <= 1)' failed`

On 8.0: Client reconnecting on the next command entering.

## Fix

Add `sp_head::MULTI_RESULTS` flag for ALTER TABLE command because it may return a result set in case of CHECK PARTITION.

On 8.0 reconnect was introduced in https://github.com/percona/percona-server/commit/14508bbb179 and assert was removed there: https://github.com/percona/percona-server/commit/14508bbb179#diff-281245ca3dd9f832d0c1cda0cf51f2410801734265ae35e7ae5120d84bd3aeb0L273. But the problem remains the same.


venkatesh-prasad-v:
One side effect of this fix I can think of is that we return different error to client when ALTER TABLE is used inside Stored Functions or Triggers. I am not sure this is a good idea or there is another way to fix it.

Without the fix, the server errors out with 
```
mysql> CREATE FUNCTION f2() RETURNS INTEGER BEGIN ALTER TABLE t1 ENGINE=INNODB; RETURN 1; END $$
ERROR 1422 (HY000): Explicit or implicit commit is not allowed in stored function or trigger.
```

With this fix, the server errors out with
```
mysql> CREATE FUNCTION f2() RETURNS INTEGER BEGIN ALTER TABLE t1 ENGINE=INNODB; RETURN 1; END $$
ERROR 1415 (0A000): Not allowed to return a result set from a function
```

Please see `is_not_allowed_in_function()` for more details.

voodam:
@venkatesh-prasad-v, I added fix for your last comment. All default MTR tests wasn't broken after this, so probably it's safe. If it's ok, I'll add it to PXC-2687 PR: https://github.com/percona/percona-xtradb-cluster/pull/1625.

sql/sp.cc
```diff
@@ -2571,13 +2571,20 @@ uint sp_get_flags_for_command(LEX *lex)
   case SQLCOM_RESET:
     flags= sp_head::HAS_SQLCOM_RESET;
     break;
+  /*
+    ALTER TABLE may return a result set (in case of CHECK PARTITION
+    at least), but doesn't have to. See SQLCOM_EXECUTE comment why
```
kamil-holubicki: Would it be possible to explain "the why" here?

sql/sp_head.h
```diff
@@ -882,10 +882,10 @@ class sp_head : private Query_arena
   {
     if (m_flags & CONTAINS_DYNAMIC_SQL)
       my_error(ER_STMT_NOT_ALLOWED_IN_SF_OR_TRG, MYF(0), "Dynamic SQL");
-    else if (m_flags & MULTI_RESULTS)
-      my_error(ER_SP_NO_RETSET, MYF(0), where);
     else if (m_flags & HAS_SET_AUTOCOMMIT_STMT)
       my_error(ER_SP_CANT_SET_AUTOCOMMIT, MYF(0));
+    else if (m_flags & MULTI_RESULTS)
```
kamil-holubicki: why this reordering is needed?

kamil-holubicki (changes requested):
Could you please:
1. Add MTR test
2. I understand there will be a fix for 8.0 as well, right? In such a case we need to use GCA process. (In general, even if there is no fix for 8.0 it should branch from GCA and be followed by an empty merge to 8.0)
3. Please write a detailed description of the root cause and the fix in the commit message (why the order of 'ifs' was changed)
4. prodecure -> procedure

dlenev:
If change in error message is an issue, then perhaps it makes sense to change the code to mark ONLY variants of ALTER TABLE that can return result set with MULTI_RESULTS flag? This will ensure that for most of ALTER TABLE variants (which do not return result sets) the error message will stay the same.

sql/sp_head.h
voodam: Just to to keep the same error message for `ALTER TABLE`. Here is an example in @venkatesh-prasad-v message above.

voodam:
> If change in error message is an issue, then perhaps it makes sense to change the code to mark ONLY variants of ALTER TABLE that can return result set with MULTI_RESULTS flag? This will ensure that for most of ALTER TABLE variants (which do not return result sets) the error message will stay the same.

AFAIU, alter info flag `Alter_info::ALTER_ADMIN_PARTITION` is not set on this stage of handling SQL command; another way can be retrieving this information is to modify parser code. I thank this ifs reorder is the best option, especially since it seems this doesn't affects existing MTR tests.

sql/sp.cc
voodam: Done

voodam:
> Could you please:
> 
> 1. Add MTR test
> 2. I understand there will be a fix for 8.0 as well, right? In such a case we need to use GCA process. (In general, even if there is no fix for 8.0 it should branch from GCA and be followed by an empty merge to 8.0)
> 3. Please write a detailed description of the root cause and the fix in the commit message (why the order of 'ifs' was changed)
> 4. prodecure -> procedure

Hi @kamil-holubicki, all is done, I'm re-requesting review.

venkatesh-prasad-v (approved):

kamil-holubicki (changes requested):
I'm afraid GCA commit is wrong. It should be 39cdc5672f9c5c3cd5c9a476c441da0076be2225 not 5e37dea7bd99394ff72ab96e6159ad1516b90cc6.
Do we have PR for 5.7 -> 8.0 merge?

voodam:
> I'm afraid GCA commit is wrong. It should be [39cdc56](https://github.com/percona/percona-server/commit/39cdc5672f9c5c3cd5c9a476c441da0076be2225) not [5e37dea](https://github.com/percona/percona-server/commit/5e37dea7bd99394ff72ab96e6159ad1516b90cc6).

Yes, I had mistake in gca script, fixed.

> Do we have PR for 5.7 -> 8.0 merge?

No I didn't create it until 5.7 PR will be approved.

venkatesh-prasad-v (approved):

kamil-holubicki (approved):
LGTM.
Please create 8.0 merge PR

voodam:
> LGTM. Please create 8.0 merge PR

https://github.com/percona/percona-server/pull/4926

percona-ysorokin (approved):

kamil-holubicki (approved):

voodam:
Nothing changed in the code, rebased only.

PR #4846, opened 2022-07-28
PS-8188 : Make the clone_plugin to not force plugins to match on reci…

satya-bodapati:
…pient/donor

https://jira.percona.com/browse/PS-8188

Problem:
--------
Clone enforces donor and recipient have same plugins installed.
For example, if a donor has an extra authentication plugin or a data masking plugin
installed but not installed on recipient, it is not possible to clone.
It will fail with the below error.

'Clone Donor plugin <plugin_name> is not active in Recipient.'

Fix:
----
Introduce new clone option 'clone_exclude_plugins_match' to take list of plugins that
are not installed on recipient.

This option is used on recipient server. The following plugins are disallowed. They must
be installed on recipient.

"daemon_keyring_proxy_plugin",
"binlog",
"performance_schema",
"memory",
"innodb",
"keyring_file",
"keyring_vault",
"group_replication"

### DIFF
--- a/plugin/clone/src/clone_client.cc
+++ b/plugin/clone/src/clone_client.cc
@@ -29,6 +29,7 @@ Clone Plugin: Client implementation
 
 #include "plugin/clone/include/clone_client.h"
 #include "plugin/clone/include/clone_os.h"
+#include "plugin/clone/include/clone.h"
 
 #include "my_byteorder.h"
 #include "my_systime.h"  // my_sleep()
@@ -1043,6 +1044,26 @@ int Client::validate_remote_params() {
     last_error = ER_CLONE_PLUGIN_MATCH;
   }
 
+  /* Build list of plugins from comma separated value of the variable
+  clone_exclude_plugins_match */
+  std::vector<std::string> excl_plugins_vec;
+  if (clone_exclude_plugins_match != nullptr) {
+    std::string excl_plugins_str(clone_exclude_plugins_match);
+    std::transform(excl_plugins_str.begin(), excl_plugins_str.end(),
+                   excl_plugins_str.begin(), ::tolower);
+
+    std::stringstream exclude_stream(excl_plugins_str);
+
+    while (exclude_stream.good()) {
+      std::string substr;
+      getline(exclude_stream, substr, ',');
+      substr.erase(remove(substr.begin(), substr.end(), ' '), substr.end());
+      if (!substr.empty()) {
+        excl_plugins_vec.push_back(substr);
+      }
+    }
+  }
+
   /* Validate plugins and check if shared objects can be loaded. */
   for (auto &plugin : m_parameters.m_plugins_with_so) {
     assert(m_share->m_protocol_version > CLONE_PROTOCOL_VERSION_V1);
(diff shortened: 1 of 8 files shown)


sjmudd:
I do not understand the previous statement:
> This option is used on recipient server. The following plugins are disallowed. They must
> be installed on recipient.
> 
> "daemon_keyring_proxy_plugin",
> "binlog",
> "performance_schema",
> "memory",
> "innodb",
> "keyring_file",
> "keyring_vault",
> "group_replication"

For example if I'm not using group replication why MUST the `group_replication` plugin be installed on the recipient, especially if it's not installed on the source?
The other plugins perhaps might lead to similar questions like the binlog plugin. If I don't have binlogs configured on the replica why would I care if it's configured on the master and not the replica?

Perhaps you mean that the plugins must be present on the target if installed on the sources at least for the ones mentioned above?

So clarification of the last paragraph you wrote would be helpful to understand more clearly why these restrictions are expected to be there and at least with the `group_replication` plugin that requirement seems bogus to me.

sjmudd:
related on a random server I'm looking at: I don't see `keyring_file` or `keyring_vault` plugins but do see what may be similar `daemon_keyring_proxy_plugin` (on MySQL community 8.0.29).

The other plugins are installed by default on my MySQL community server.

I use native cloning for cloning from non-GR servers (and this works fine) so the requirement of the `group_replication` plugin to be installed seems to be unnecessary. In some larger clusters I use I may have async slaves replicating from a GR cluster to provide wider scaleout.  I believe that even when cloning from a GR cluster member to setup an additional async replica this works fine.

sjmudd:
The variable name suggested for exclusion, `clone_exclude_plugins_match` may not be ideal. This is a list and match to me implies some sort of regex or similar. I wonder if changing the name to `clone_exclude_plugins_list` might read better and be clearer.  Naming is hard but to me this would sound and read better than the name you propose.

plugin/clone/src/clone_client.cc
```diff
@@ -1058,6 +1079,25 @@ int Client::validate_remote_params() {
...
+    if (clone_exclude_plugins_match != nullptr) {
+      std::string plugin_str(plugin_name);
+      std::transform(plugin_str.begin(), plugin_str.end(), plugin_str.begin(),
+                     ::tolower);
+
+      if (std::find(excl_plugins_vec.begin(), excl_plugins_vec.end(),
+                    plugin_str) != excl_plugins_vec.end()) {
+        char info_mesg[128];
+        snprintf(info_mesg, 128,
+                 "Plugin %s is not installed on recipient but ignored due to "
```
sjmudd: The only reason the plugin not being installed is an issue is because it's on the source? If so say so as the reader of the error message won't realise why the plugin is being installed.

e.g. `"Ignoring plugin %s installed on the source server but not the recipient.  The mismatch is ignored due to 'clone_exclude_plugins_list'"`

I think that something like this slightly different comment makes the DBA aware of (i) why there's a mismatch and (ii) why it's excluded in this specific case.

plugin/clone/src/clone_plugin.cc
```diff
@@ -502,6 +518,62 @@ static int plugin_clone_remote_server(THD *thd, MYSQL_SOCKET socket) {
...
+  char buf[STRING_BUFFER_USUAL_SIZE];
+  int len = sizeof(buf);
+  const char *val = mysql_val->val_str(mysql_val, buf, &len);
+
+  if (val != NULL) val = thd_strmake(thd, val, len);
+
+  return val;
+}
+
+static bool plugin_is_ignorable(std::string &plugin_name) {
```
laurynas-biveinis: const

dutow:
Merged as #5032

PR #4854, opened 2022-08-05
PS-7745 - Stored procedure execution fails if contains a functional index

avodaniel:


mysql-test/suite/rocksdb/t/issue7745.test
```diff
@@ -0,0 +1,18 @@
+--source include/have_rocksdb.inc
```
dutow: This isn't specific to rocksdb.

Could you move the testcase to the proper directory and remove this check?

dutow (commented):
Looks good! With one simple comment.

PR #4861, opened 2022-08-15
PS-6002: CREATE/ALTER EVENT STATE TO KEEP ENABLED ON REPLICAS

venkatesh-prasad-v:
https://jira.percona.com/browse/PS-6002

This feature introduces a global config variable --replica-enable-event
that accepts the regex patterns of the events in format
"db_pattern.event_pattern" for those events that need to be re-enabld
after the creation or modification of event on replica server, unlike
the current behavior where the replicated events are by default set to
the state SLAVESIDE_DISABLED.

The below shall be the behavior when a replicated event matches the pattern
specified in the `--replica-enable-event`.

1. When an event is created as ENABLED on source, it shall remain ENABLED
   on replica as well.

2. When an event is created as DISABLED it shall remain DISABLED on
   replica as well.

3. If an event in ENABLED state is altered as DISABLED on source, then it
   shall become DISABLED on replica as well.

4. If an event in ENABLED state is altered and if it remains ENABLED on
   source, then it shall remain ENABLED on replica as well.

5. If an event in DISABLED state is altered and if it remains DISABLED on
   source then, shall remain DISABLED on replica as well.

6. If an event in DISABLED state is altered to ENABLED state on the source
   server, then it shall become ENABLED on replica as well.

Related approved PR: https://github.com/percona/percona-server/pull/4400


satya-bodapati (approved):
Please run tests against 8.0 (jenkins) and ensure clean run before merging

venkatesh-prasad-v:
Jenkins: https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/293/console

Test result:
image

PR #4868, opened 2022-08-31
PS-8330: Merge MySQL 8.0.30

dutow:
Note: two feature removal commits are added before the merge commit


percona-ysorokin:
@dutow I tried to build this branch on Ubuntu 22.04 with OpenSSL 3.0 and encountered a number of OpenSSL "using deprecated functions" warnings.
I noticed that you also removed
```
    IF("${OPENSSL_MAJOR_VERSION}.${OPENSSL_MINOR_VERSION}.${OPENSSL_FIX_VERSION}" VERSION_GREATER "3.0.0")
       ADD_DEFINITIONS(-DOPENSSL_SUPPRESS_DEPRECATED)
    ENDIF()
```
from the `ssl.cmake` include file.
Most of warnings are coming from Encryption UDFs component, which is expected, and most probably should be suppressed locally inside `extra/opensslpp/CMakeLists.txt`
```
target_compile_definitions(${PROJECT_NAME} PRIVATE OPENSSL_SUPPRESS_DEPRECATED
```
But there are also a few coming from the server source code that need to be dealt with individually.

percona-ysorokin (approved):
Hard to do full-blown review because of the number of conflicts but this is definitely OK for the release-8.0.30-22 starting point.

PR #4870, opened 2022-09-07
Implemented PS-8385 (change redo log consumer registrator to allow us…

altmannmarcelo:
…er to specify a consumer name)

https://jira.percona.com/browse/PS-8385

Problem:
Currently implementation of redo log consumer UDF does not allow users to specify a custom name for the consumer. This can be misleading when ER_IB_MSG_LOG_WRITER_WAIT_ON_CONSUMER error is issued as it will always report MEB as consumer name.

Fix:
Adjust the UDF to allow users to optionally specify a consumer name. In case of no consumer name specified, assume MEB.
Also adjusted Log_consumer class to have a consumer_type ENUM. This is used log_writer_wait_on_consumers to validate we are not waiting on CONSUMER_TYPE_SERVER (log archive or checkpointer).

### DIFF
--- a/storage/innobase/include/log0consumer.h
+++ b/storage/innobase/include/log0consumer.h
@@ -52,7 +52,7 @@ class Log_consumer {
   the oldest redo log file. */
   virtual void consumption_requested() = 0;
 
-  typedef enum { CONSUMER_TYPE_SERVER, CONSUMER_TYPE_USER } consumer_type;
+  enum class consumer_type { SERVER, USER };
 
   /** @return Type of this consumer. */
   virtual consumer_type get_consumer_type() const = 0;
--- a/storage/innobase/include/log0consumer.h
+++ b/storage/innobase/include/log0consumer.h
@@ -51,6 +51,11 @@ class Log_consumer {
   is the most lagging one and it is critical to consume
   the oldest redo log file. */
   virtual void consumption_requested() = 0;
+
+  typedef enum { CONSUMER_TYPE_SERVER, CONSUMER_TYPE_USER } consumer_type;
+
+  /** @return Type of this consumer. */
+  virtual consumer_type get_consumer_type() const = 0;
 };
 
 class Log_user_consumer : public Log_consumer {
@@ -69,6 +74,8 @@ class Log_user_consumer : public Log_consumer {
 
   void consumption_requested() override;
 
+  Log_consumer::consumer_type get_consumer_type() const override;
+
  private:
   /** Name of this consumer (saved value from ctor). */
   const std::string m_name;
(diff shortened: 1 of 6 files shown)


storage/innobase/log/log0write.cc
```diff
@@ -2088,11 +2088,14 @@ static void log_writer_wait_on_consumers(log_t &log, lsn_t next_write_lsn) {
...
+
     log_files_mutex_exit(*log_sys);
-    /* This should not be a checkpointer nor archiver, as we've used dedicated
-    log_writer_wait_on_checkpoint() and log_writer_wait_on_archiver() to wait
-    for them already */
-    ut_ad(name == "MEB");
+    /* This should not be a checkpointer nor archiver (CONSUMER_TYPE_SERVER), as
+    we've used dedicated log_writer_wait_on_checkpoint() and
+    log_writer_wait_on_archiver() to wait for them already */
+    ut_ad(consumer_type == Log_consumer::consumer_type::CONSUMER_TYPE_USER);
```
rahulmalik87: We can avoid adding a new enum Log_consumer::consumer_type if its purpose is only to ensure it is not an archive and checkpoint consumer. 
instead use ut_ad(name != "log_checkpointer" && name != "log_archiver").

rahulmalik87: on second thought. If server calls log_writer_wait_on_consumers frequent. then enum is the fast and clear approach.

storage/innobase/log/log0meb.cc
```diff
@@ -1667,7 +1667,8 @@ static bool redo_log_archive_flush(THD *thd) {
 static std::unique_ptr<Log_user_consumer> log_meb_consumer;
 static innodb_session_t *log_meb_consumer_session;
 
-static bool redo_log_consumer_register(innodb_session_t *session) {
+static bool redo_log_consumer_register(innodb_session_t *session,
+                                       std::string name) {
```
dutow: `std::string const&`

storage/innobase/log/log0meb.cc
```diff
@@ -2323,12 +2324,18 @@ long long innodb_redo_log_consumer_register(
...
     [[maybe_unused]] unsigned char *error) {
+  std::string name = "MEB";
   if (current_thd == nullptr ||
       verify_privilege(current_thd, backup_admin_privilege)) {
     return 1;
   }
-  return static_cast<long long>(
-      meb::redo_log_consumer_register(thd_to_innodb_session(current_thd)));
+
+  if (args->arg_count == 1) {
```
dutow: I would use `>=` here. Right now, we error out if the argument count is greater than 1, but later, if for some reason we do add more arguments, `>= 1` will still work.

storage/innobase/include/log0consumer.h
```diff
@@ -51,6 +51,11 @@ class Log_consumer {
   is the most lagging one and it is critical to consume
   the oldest redo log file. */
   virtual void consumption_requested() = 0;
+
+  typedef enum { CONSUMER_TYPE_SERVER, CONSUMER_TYPE_USER } consumer_type;
```
dutow: We could use an `enum class`, and simpler values (`SERVER` and `USER`)

dutow (approved):

rahulmalik87 (approved):
LGTM

PR #4871, opened 2022-09-08
PS-8246 bug: Adapt Encryption UDF MTR test cases to OpenSSL 3.0.x

percona-ysorokin:
https://jira.percona.com/browse/PS-8246

Refactored Encryption UDFs MTR test cases so that they now pass on platforms with OpenSSL 3.0.x.

'have_openssl_binary_version.inc' MTR include file extended so that in case when '$openssl_binary_version_do_not_skip' variable is set to non-zero, it does not skip the test but instead just reports whether the OpenSSL library version matches the provided pattern or not via '$openssl_binary_version_match' result variable.

Fixed how 'have_openssl_binary_version.inc' was used in some 'Encryption UDFs' MTR test cases - in some occasions '$openssl_binary_version' variable was not defined before this file was included.

Direct calls to '--exec openssl ...' were changed to more appropriate '--exec $OPENSSL_EXECUTABLE ...'.

'percona_encryption_udf_digest_table.inc' reworked so that we fill the 'digest_type' table with different values based on the hash functions available in the particular version of the OpenSSL library that is currently in use (1.0.2 / 1.1.x / 3.0.x).

Recorded PEM-format read/write error messages generated by Encryption UDFs functions trimmed with '--replace_regex' to just basic (defined inside 'opensslpp') messages in order to unify this output for different OpenSSL library versions.


percona-ysorokin:
Jenkins build link (--big-test --do-test=percona_encryption_udf)
https://ps80.cd.percona.com/job/percona-server-8.0-param/1253/

oleksandr-kachan (approved):

PR #4879, opened 2022-09-14
post-fix PS-8385 (Redo log consumer name)

altmannmarcelo:
Problem:
consumer_type variable was been considered unused on non debug builds.

Fix:
Adjusted ut_ad to use consumer->get_consumer_type() directly instead of storing its value into a variable.


percona-ysorokin (approved):
LGTM

PR #4883, opened 2022-09-26
PS-8330: Fixing remaining 8.0.30 failures

dutow:

PR #4889, opened 2022-09-29
PS-8418: Collision between compression_dictionary table id and id of the table created after it.

kamil-holubicki:
Note: This branch is based on 833stuff branch which has open PR, so https://github.com/percona/percona-server/pull/4883 should be merged 1st

PS-8418: Collision between compression_dictionary
table id and id of the table created after it.

https://jira.percona.com/browse/PS-8418

Problem:
During the upgrade to 8.0.30 there is a collision between IDs of
compression_dictionary (c_d) / compression_dictionary_cols (c_d_c) and
replication_asynchronous_connection_failover /
replication_asynchronous_connection_failover_managed tables.

Cause:
1. PS-5245 (commit https://github.com/percona/percona-server/commit/69e7b49b8fa411452216b8743053e92efe2e5557) introduced
the logic in bootstrapper.cc:check_and_create_compression_dict_tables()
which "fixes" IDs of c_d and c_d_c tables to be two next IDs just after
view_table_usage id. The fix logic is triggered by the detection of IDs
being not 3 consecutive numbers.
2. In 8.0.30 view_table_usage table definition changed. This triggers
its upgrade. As the result of the upgrade, ID of the table changes from
33 to 1080.
3. This triggers logic in bootstrapper.cc. ID's of c_d and c_d_c tables
are patched to 1081 and 1082.
4. InnoDB knows nothing about this patching, so DICT_HDR_TABLE_ID values
does not change (still 1080)
5. After we patch c_d and c_d_c tables we go directly to
upgrade_system_schemas() and creation of
mysql.replication_asynchronous_connection_failover and
mysql.replication_asynchronous_connection_failover_managed tables.
They get IDs 1081 and 1082 accordingly and the insertion into
mysql.tables fails with a duplicate key error, but the error is ignored
in ignore_error_and_execute().
Next table mysql.replication_group_member_actions gets ID 1083 and
everything works fine.

Side note: Why this fix was introduced?
1. 8.0.13-3 was released with a bug. During the upgrade from 5.7 when
c_d and c_d_c tables were created, their IDs were allocated as
regular tables. As the creation was done before upgrading user tables
(and patching the value of DICT_HDR_TABLE_ID inside InnoDB's dictionary
header), the allocated ID was shifted down by 1024. Later, when user
tables were upgraded, they were read and added to the InnoDB's
dictionary cache. Cache detected that the table with the same ID already
exists and assert was triggered (PS-5245).

2. The detailed process of handling IDs during 5.7 -> 8.0 upgrade:
We've got ids clash because 5.7 generates IDs without 1024 offset, so
user table can be 100 for example.
During the upgrade of user tables 5.7 -> 8.0, InnoDB cheats a little
bit, and reports IDs that are shifted by 1024
(dict0load.cc:dict_load_table_one()), but dict_header
(DICT_HDR_TABLE_ID) still contains 5.7 (not shifted) real value.
The value of DICT_HDR_TABLE_ID is "fixed" after all tables are upgraded
(in dd_upgrade_logs()). Starting from this point any new table will have
a proper ID.
The problem with 8.0.13-3 is that it creates compression_dictionary
table before the above-mentioned process. So the table gets ID which is
not shifted.

So the sequence is:
a. Create compression_dictionary table with the wrong ID. add it to the
cache.
b. Try to migrate the user table. To do this, open it and add it to the
cache.
But before adding check if there is no table with the same ID there.
c. Assert

We cannot postpone compression_dictionary table creation to be done
after the user's tables migration and fixing the value of
DICT_HDR_TABLE_ID, because this table is needed at DD upgrade stage
which happens earlier.

3. PS-5245 fixed two issues:
bootstrapper.cc part - patch c_d and c_d_c tables ids which were created
during 5.7 -> 8.0.13-3 upgrade. Having this, any upgrade
8.0.13-3 -> 8.0.x will fix bad ids generated by 8.0.13-3
dict0crea.cc part - fix generation of c_d and c_d_c ids when table is
created. Having this we generate the proper ID for these tables when
they are created (5.7 -> 8.0 upgrade or 8.0 fresh bootstrap).

Solution:
The problem is caused by the logic inside bootstrapper.cc. It was
introduced for upgrade scenarios 5.7 -> 8.0.13-3 -> 8.0.x, to be able to
get out of problematic situation if someone upgraded 5.7 -> 8.0.13-3.
It is rather safe to assume now that this path is not so probable, so I
reverted the problematic part of the patch. If anyone follows this path
the way to get out of spoiled 8.0.13-3 is to upgrade to < 8.0.30 first.

1. Removed percona_dd_upgrade_8013 MTR test as it is not relevant
anymore
2. Introduced percona_dd_upgrade_57_many_tables MTR test which upgrades
from 5.7 having 1200 user tables. This test is to prove that c_d and
c_d_c tables IDs are generated in the proper way
(without the dict0crea.cc fix part it will fail).


satya-bodapati (approved):
Thanks for detailed explanation in commit message.
Please run entire test suite. There are some tests related to compression dictionary i_s views.

kamil-holubicki:
https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/68/

mysql-test/r/ctype_euckr.result
```diff
@@ -5887,7 +5887,6 @@ F058
 F059
 F05A
 F061
-F062
```
kamil-holubicki, line 5890: Why this change? On PXC I see that F062 is still there.

dutow: This only happens on my system, not on jenkins, I reverted it.

kamil-holubicki:
Cherry picked to release 8.0.30 by commit b5c31f86e92ad621b3037b1d61654b37f471ab74

PR #4890, opened 2022-09-29
PS-8106: MySQL crashes if you stop replication while it is replicating a CTAS.

voodam:
https://jira.percona.com/browse/PS-8106

On rollback of `CREATE TABLE .. AS SELECT` statement DDL Log REMOVE CACHE is executed, but underlying procedure `dict_table_remove_from_cache_low` requires that table shouldn't have opened handles (`get_ref_count() == 0`). If tables were closed before rollback, this procedure can be executed freely.

This bug has nondeterministic character because statement can be commited already or not arrived yet to replica. Setting replica_parallel_workers != 0 makes rollback practically uncatchable, in most cases transaction will commit.

I runned all replication-related tests and did't found any degradation.

Also I tried to write test to this bug, but faced significant challenges: it seems that some special conditions are required to catch rollback, in my test I had commit in all tries. So to same some time I postponed writting of the test. Is it required? If yes, I'll try to finish and add it.

### DIFF
--- a/sql/rpl_rli.cc
+++ b/sql/rpl_rli.cc
@@ -1313,6 +1313,12 @@ void Relay_log_info::cleanup_context(THD *thd, bool error) {
 
   assert(info_thd == thd);
   /*
+    It's required to close thread tables before rollback, because some
+    rollback operations, like replay_remove_cache_log, require closed tables.
+  */
+  m_table_map.clear_tables();
+  slave_close_thread_tables(thd);
+  /*
     1) Instances of Table_map_log_event, if ::do_apply_event() was called on
     them, may have opened tables, which we cannot be sure have been closed
     (because maybe the Rows_log_event have not been found or will not be,
@@ -1344,8 +1350,6 @@ void Relay_log_info::cleanup_context(THD *thd, bool error) {
       assert(!debug_sync_set_action(info_thd, STRING_WITH_LEN(action)));
     };);
   }
-  m_table_map.clear_tables();
-  slave_close_thread_tables(thd);
   if (error) {
     /*
       trans_rollback above does not rollback XA transactions.
--- a/mysql-test/suite/rpl/r/percona_bug_ps8106.result
+++ b/mysql-test/suite/rpl/r/percona_bug_ps8106.result
@@ -1,0 +1,40 @@
+include/master-slave.inc
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
+Note	####	Storing MySQL user name or password information in the master info repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START SLAVE; see the 'START SLAVE Syntax' in the MySQL Manual for more information.
+[connection master]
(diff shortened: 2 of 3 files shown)


sql/rpl_rli.cc
```diff
@@ -1345,7 +1350,6 @@ void Relay_log_info::cleanup_context(THD *thd, bool error) {
     };);
   }
   m_table_map.clear_tables();
```
kamil-holubicki: Wouldn't it be good to move as well for better code readability?

kamil-holubicki (commented):
Would it be possible to introduce MTR test?

sql/rpl_rli.cc
voodam: I'm not sure that understand right, could you explain please: move `m_table_map.clear_tables();` before `slave_close_thread_tables(thd);`? Yes, but probably it requires to rerun all replication tests again.

voodam:
> Would it be possible to introduce MTR test?

I suppose yes. but it requires further investigations, I tried and so far haven't been able to catch rollback situation. So if this is 
 desirable, I'll work on test further.

sql/rpl_rli.cc
voodam: Done

voodam:
@kamil-holubicki Test added

kamil-holubicki:
https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/270/

voodam:
> https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/270/

Are these tests broken after changes in this commit or they are always broken? Sorry, I'm not familiar with pipelines, got access to it no so long ago.

kamil-holubicki:
Looks like something is wrong with the previous Jenkins run. I've restarted it
https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/290/

voodam:
> Looks like something is wrong with the previous Jenkins run. I've restarted it https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/290/

These tests are seem broken in 8.0:
- encryption.innodb-redo-page0-wrongkey
- encryption.innodb-redo-nokeys
- perfschema.memory_key_descriptions

I tried to re-run them locally on this branch and another one. Besides them there are no broken tests in this run.
UPD: they are broken outside of this PR.

kamil-holubicki (approved):
lgtm

venkatesh-prasad-v (approved):

voodam:
@kamil-holubicki Can I merge, or it's better just to ask you in such cases?

kamil-holubicki:
> @kamil-holubicki Can I merge, or it's better just to ask you in such cases?

Yes, once it it is approved and has no merge conflicts, go ahead and merge :)

kamil-holubicki (approved):

PR #4899, opened 2022-10-11
PS-1098: Fix audit_log_flush behavior when used with auto log rotation (5.7)

oleksandr-kachan:
https://jira.percona.com/browse/PS-1098

In case auto log rotation is enabled for audit log and 'SET GLOBAL audit_log_flush=1' is used right after log file is rotated it leads to just rotated file being reopend. Setting audit_log_flush=1 should do nothing except reopening current audit log file.

The reason of wrong behavior is that do_rotate() method uses data structures of logger handler to store temporary data while checking and renaming existing log files. Once it is done with this task logger handler may be not returned to its correct state in some scenarios.

Rewritten do_rotate() method not to use logger handler as temporary data storage.

### DIFF
--- a/plugin/audit_log/file_logger.c
+++ b/plugin/audit_log/file_logger.c
@@ -218,38 +218,45 @@ static char *logname(LOGGER_HANDLE *log, char *buf, size_t buf_len,
 }
 
 
-static int do_rotate(LOGGER_HANDLE *log)
-{
-  char namebuf[FN_REFLEN];
-  int result;
-  unsigned int i;
-  char *buf_old, *buf_new, *tmp;
+static int do_rotate(LOGGER_HANDLE *log) {
+  char new_name_buf[FN_REFLEN] = {0};
+  char old_name_buf[FN_REFLEN] = {0};
+  char *new_name = NULL;
+  char *old_name = NULL;
+  unsigned int i = 0;
 
-  if (log->rotations == 0)
-    return 0;
+  if (log->rotations == 0) return 0;
 
-  memcpy(namebuf, log->path, log->path_len);
+  memcpy(new_name_buf, log->path, log->path_len);
+  memcpy(old_name_buf, log->path, log->path_len);
 
-  buf_new= logname(log, namebuf, sizeof(namebuf), log->rotations);
-  buf_old= log->path;
-  for (i=log->rotations-1; i>0; i--)
-  {
-    logname(log, buf_old, FN_REFLEN, i);
-    if (!access(buf_old, F_OK) &&
-        (result= my_rename(buf_old, buf_new, MYF(0))))
-      goto exit;
-    tmp= buf_old;
-    buf_old= buf_new;
-    buf_new= tmp;
+  for (i = log->rotations; i > 0; i--) {
+    new_name = logname(log, new_name_buf, sizeof(new_name_buf), i);
+
+    if (i > 1) {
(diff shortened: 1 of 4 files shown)


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/575/

plugin/audit_log/file_logger.c
```diff
@@ -218,38 +218,45 @@ static char *logname(LOGGER_HANDLE *log, char *buf, size_t buf_len,
...
+  char *new_name = NULL;
+  char *old_name = NULL;
+  unsigned int i = 0;
 
-  if (log->rotations == 0)
-    return 0;
+  if (log->rotations == 0) return 0;
 
-  memcpy(namebuf, log->path, log->path_len);
+  memcpy(new_name_buf, log->path, log->path_len);
```
percona-ysorokin, line 230: Do we need to put trailing zeros here?

plugin/audit_log/file_logger.c
```diff
@@ -218,38 +218,45 @@ static char *logname(LOGGER_HANDLE *log, char *buf, size_t buf_len,
...
+  memcpy(old_name_buf, log->path, log->path_len);
```
percona-ysorokin, line 231: Likewise

plugin/audit_log/tests/mtr/audit_log_rotate_with_flush.test
```diff
@@ -0,0 +1,26 @@
...
+--disable_result_log
+--disable_query_log
+--source audit_log_events.inc
+--enable_query_log
+--enable_result_log
+
+SET GLOBAL audit_log_flush=1;
+
+let $MYSQLD_DATADIR= `select @@datadir`;
+let MYSQLD_DATADIR= $MYSQLD_DATADIR;
```
percona-ysorokin: Declaring just
```
let MYSQLD_DATADIR = `select @@datadir`;
```
is enough here. Environment variables can still be accessed from MTR via `$MYSQLD_DATADIR`

plugin/audit_log/file_logger.c
```diff
@@ -218,38 +218,45 @@ static char *logname(LOGGER_HANDLE *log, char *buf, size_t buf_len,
...
+    if (i > 1) {
+      old_name = logname(log, old_name_buf, sizeof(new_name_buf), i - 1);
+    }
+    else {
+      old_name = old_name_buf;
+      old_name_buf[log->path_len] = 0;
+    }
+
+    if (0 == access(old_name, F_OK) &&
+        0 != my_rename(old_name, new_name, MYF(0))) {
```
percona-ysorokin, line 245: Is it OK that we rename currently open file? (when `i==1`)

oleksandr-kachan: Yes, server will continue writing into this renamed file until it gets closed below.

plugin/audit_log/file_logger.c
oleksandr-kachan: Will initialize buffers

plugin/audit_log/tests/mtr/audit_log_rotate_with_flush.test
oleksandr-kachan: Done

percona-ysorokin (approved):
LGTM

PR #4902, opened 2022-10-13
PS-1098: Fix audit_log_flush behavior when used with auto log rotation (8.0)

oleksandr-kachan:
https://jira.percona.com/browse/PS-1098

In case auto log rotation is enabled for audit log and 'SET GLOBAL audit_log_flush=1' is used right after log file is rotated it leads to just rotated file being reopened. Setting audit_log_flush=1 should do nothing except reopening current audit log file.

The reason of wrong behavior is that do_rotate() method uses data structures of logger handler to store temporary data while checking and renaming existing log files. Once it is done with this task logger handler may be not returned to its correct state in some scenarios.

Rewritten do_rotate() method not to use logger handler as temporary data storage.

### DIFF
--- a/plugin/audit_log/file_logger.c
+++ b/plugin/audit_log/file_logger.c
@@ -218,38 +218,45 @@ static char *logname(LOGGER_HANDLE *log, char *buf, size_t buf_len,
 }
 
 
-static int do_rotate(LOGGER_HANDLE *log)
-{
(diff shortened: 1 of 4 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param/1260/

percona-ysorokin (approved):
LGTM

PR #4903, opened 2022-10-15
PS-8428 fix: ALTER TABLE t ADD FULLTEXT crashes the server when --innodb_encrypt_online_alter_logs=ON

percona-ysorokin:
https://jira.percona.com/browse/PS-8428

Fixed problem with 'ALTERR TABLE ... ADD FULLTEXT ...' statement crashing the server when 'innodb_encrypt_online_alter_logs' is set to 'ON'.

It turned out that the root cause of this problem is Percona-specific 'my_aes_crypt()' function not working as expected with OpenSSL 3.0.x. Basically, this function cannot properly decrypt data encrypted previously with the same key/iv. Internally it relies on the value in the Cipher Context internal buffer ('EVP_CIPHER_CTX_buf_noconst()') after partial AES update. OpenSSL never claimed that after partial AES update this value will point to the remainder of the original message and this behaviour changed in OpenSSL 3.0.x.

Completely removed "my_crypt.h" / "my_crypt.cc" for the following reasons:
- Our code uses 'my_aes_crypt()' only in either ECB 256-bit mode or in EBC 256-bit nopad with custom remainder encryption.
- AES CTR / GCM were never used in the code.
- AES ECB functionality already exists in "my_aes.h" ('my_aes_encrypt()' / 'my_aes_decrypt'),
- 'my_random_bytes()' function functionality already exists in the 'my_rand_buffer()' inside "my_rnd.h". Instead, "my_aes.h" extended with two new functions 'my_legacy_aes_256_cbc_nopad_encrypt()' and
'my_legacy_aes_256_cbc_nopad_decrypt()' implemented on a single internal static 'my_legacy_aes_256_cbc_nopad_crypt()'. This 'my_legacy_aes_256_cbc_nopad_decrypt()' does not rely on the value of the 'EVP_CIPHER_CTX_buf_noconst()' and works as expected with both OpenSSL '1.0.2' / '1.1.1' and '3.0.x'.

Removed 'using' declarations from the 'my_aes.h' and "my_kdf.h" header files that were polluting global namespace.

Added 'innodb.percona_bug_ps8428' MTR test case which checks 'ALTER TABLE ... ADD FULLTEXT ...' under
'--innodb_encrypt_online_alter_logs=ON'.


mysys/my_aes_openssl.cc
```diff
@@ -259,3 +263,89 @@ bool my_aes_needs_iv(my_aes_opmode opmode) {
...
+  if (!EVP_CipherInit_ex(ctx_raw, cipher, nullptr, key, iv, encrypt ? 1 : 0))
+    return clear_error_helper(MY_AES_BAD_DATA);
+  if (!EVP_CIPHER_CTX_set_padding(ctx_raw, 0))
+    return clear_error_helper(MY_AES_BAD_DATA);
+  if (!EVP_CipherUpdate(ctx_raw, dest, &u_len, source, source_length))
+    return clear_error_helper(MY_AES_BAD_DATA);
+
+  assert(static_cast<int>(source_length) >= u_len);
+  if (static_cast<int>(source_length) > u_len) {
+    std::size_t remainder_len = source_length - u_len;
```
satya-bodapati: should we check remainder_len ==0 and skip rest if it is 0?

percona-ysorokin: Basically, `if (static_cast<int>(source_length) > u_len)` is the same as `if (remainder != 0)`. So, we are skipping the rest if we work with a data block that is of length `16*n`

percona-ysorokin: Oh, I see. Probably
```c++
  assert(static_cast<int>(source_length) >= u_len);
  std::size_t remainder_len = source_length - u_len;
  if (remainder_len > 0) {
```
would be better

include/my_aes.h
```diff
@@ -140,4 +137,40 @@ longlong my_aes_get_size(uint32 source_length, enum my_aes_opmode opmode);
...
+/**
+  Decrypt a buffer encrypted with AES 256-bit CBC with no padding
+
+  @param source         Pointer to data for decryption
+  @param source_length  size of encrypted data
+  @param dest           buffer to place decrypted data (must be large enough)
+  @param key            Key to be used for decryption
+  @param iv             16 bytes initialization vector
+  @param padding        if padding needed.
+  @return size of original data.
```
satya-bodapati: in,out etc are missing

mysql-test/suite/innodb/t/percona_bug_ps8428.test
```diff
@@ -0,0 +1,11 @@
+#
```
satya-bodapati, line 1: we can rename the test without a bug number. percona_encrypt_online_alter?

satya-bodapati (changes requested):

mysql-test/suite/innodb/t/percona_bug_ps8428.test
satya-bodapati: [optional]

percona-ysorokin: There is already `innodb.innodb_row_log_encryption` MTR test case which tests `--innodb_encrypt_online_alter_logs=ON` and originally I wanted to extend it instead of creating a new one.
However, it has different `.opt` file and is shared with another test from `keyring_vault` via `.inc` file.
So, I'd rather leave it as is.

include/my_aes.h
percona-ysorokin: fixed

mysys/my_aes_openssl.cc
percona-ysorokin: fixed

satya-bodapati (approved):

percona-ysorokin:
Jenkins build link
https://ps80.cd.percona.com/job/percona-server-8.0-param/1264/ (--suite=innodb,innodb_fts,innodb_undo,innodb_zip,main + keyring_vault)

PR #4905, opened 2022-10-17
PS-8275: Implement auto_evict_timeout

dutow:
This commit implement a new system variable,
group_replication_auto_evict_timeout, which automatically evicts members if they were in flow_control for more than the specified timeout.

By default this variable is 0, and doesn't change the existing behavior.

If it is set to a non zero value, any node that reaches the flow control threshold self evicts itself flom the cluster.

This change doesn't take into account the majority concept introduced in PS-8276: a single node is considered in flow control, and timeout calculation strarts, even if only a single node reaches the limit.

### DIFF
--- a/plugin/group_replication/include/plugin_variables.h
+++ b/plugin/group_replication/include/plugin_variables.h
@@ -295,6 +295,8 @@ struct plugin_options_variables {
   ulong communication_stack_var;
 
   bool allow_single_leader_var{false};
+
+  uint auto_evict_timeout;
 };
 
 #endif /* PLUGIN_VARIABLES_INCLUDE */
--- a/plugin/group_replication/src/pipeline_stats.cc
+++ b/plugin/group_replication/src/pipeline_stats.cc
@@ -28,6 +28,7 @@
 #include "my_byteorder.h"
 #include "my_dbug.h"
 #include "my_systime.h"
+#include "plugin/group_replication/include/leave_group_on_failure.h"
 #include "plugin/group_replication/include/plugin.h"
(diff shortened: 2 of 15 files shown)


plugin/group_replication/include/plugin_variables.h
```diff
@@ -295,6 +295,8 @@ struct plugin_options_variables {
   ulong communication_stack_var;
 
   bool allow_single_leader_var{false};
+
+  uint auto_evict_timeout;
```
venkatesh-prasad-v, line 299: Can we use uint64 here as well?

plugin/group_replication/src/pipeline_stats.cc
```diff
@@ -742,6 +743,52 @@ void Flow_control_module::flow_control_step(
     Pipeline_stats_member_collector *member) {
   if (--seconds_to_skip > 0) return;
 
+  if (get_auto_evict_timeout() > 0) {
+    auto it = m_info.end();
+    if (group_member_mgr != nullptr) {
+      auto ml = group_member_mgr->get_all_members();
```
venkatesh-prasad-v, line 749: `get_all_members()` uses `new`. We must delete the contents of the vector and delete the vector in the end. Please fix this as well.

venkatesh-prasad-v (changes requested):

venkatesh-prasad-v (approved):

dutow:
https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/313/

PR #4909, opened 2022-10-19
PS-8421: Merge MySQL 5.7.40 (part1)

oleksandr-kachan:
https://jira.percona.com/browse/PS-8421
***
VERSION raised to "5.7.40-43.0".
PERCONA_INNODB_VERSION in univ.i raised to "43.0".
***
ROCKSDB_VALGRIND_RUN fix.

-DROCKSDB_VALGRIND_RUN' preprocessor definition is now also
passed to the 'RocksDB' code when top-level 'WITH_VALGRIND'
CMake option is set to 'ON'.

This allows to eliminate
"==10176== Warning: unimplemented fcntl command: 1036"
Valgrind warnings coming from 'rocksdb/env/io_posix.cc' and
false positives coming from not calling some static object
destructors.

Conditional 'ROCKSDB_UBSAN_RUN' / 'ROCKSDB_VALGRIND_RUN' /
'ROCKSDB_JEMALLOC' RocksDB-specific defines grouped together
in the RocksDB Storage Engine 'CMakeLists.txt' file.
***
Fix valgrind "Invalid read of size 8" in rpl.rpl_parallel_multi_db.

The "Invalid read of size 8" Valgrind error is observed in
rpl.rpl_parallel_multi_db MTR test results.

The https://github.com/percona/percona-server/commit/bdd12ac67a6cb0e1e295457205e9f640c6e9d0ab
change introduced check to prevent dangerous use of rpl writebatch for
rocksdb. In order to find out if the feature is used properly in
Query_log_event::do_apply_event() we check if handlers for all query
tables support the feature. The issue here is that by the time the check
is performed all involved table objects are already released for later
reuse or complettely closed and memory is freed.

To fix the issue corresponding check was changed to be done at
the point where all query table have just been opened.
***
Fix valgrind "Syscall param pwrite64(buf) points to uninitialised byte".

Write to uninitialized buffer is used on purpose but it
leads to "Syscall param pwrite64(buf) points to uninitialised byte"
valgrind error in innodb.check_sector_size MTR test.
Use UNIV_MEM_VALID to suppress valgrind error.
***
Fix unit_tests run with valgrind.

The DartConfiguration.tcl config is missing in buid directory.
The BUILD_TESTING is required to be set for include(CTest) to
generate it properly.


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/580/
https://ps57.cd.percona.com/job/percona-server-5.7-asan-param/58/

percona-ysorokin (approved):
LGTM

PR #4911, opened 2022-10-24
PS-8169: LDAP SASL plugin

dutow:
This commit adds the server side plugin for SASL LDAP authentication. The client side is already present in the sources.

A high level overview of the authentication process is also described in the readme file in the plugin directory.

### DIFF
--- a/plugin/auth_ldap/include/plugin_sasl.h
+++ b/plugin/auth_ldap/include/plugin_sasl.h
@@ -1,0 +1,24 @@
+#ifndef PLUGIN_SIMPLE_MPALDAP_H
+/* Copyright (c) 2022, Percona Inc. All Rights Reserved.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; version 2 of the License.
+
+   This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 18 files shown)


plugin/auth_ldap/include/plugin_sasl.h
```diff
@@ -0,0 +1,24 @@
+#ifndef PLUGIN_SIMPLE_MPALDAP_H
+/* Copyright (c) 2019 Francisco Miguel Biete Banon. All rights reserved.
```
altmannmarcelo: shouldn't we use Percona copyright and mention the contributor as part of it?

altmannmarcelo (approved):
It LGTM apart from the question on copyright.

Do we have LDAP jenkins that we can run this against?

plugin/auth_ldap/include/plugin_sasl.h
dutow: I updated the copyrights to also include Percona.

dutow:
We don't have jenkins tests even for the simple LDAP plugin which is easier to set up, we can only test that it builds there.

dutow:
https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-param-parallel-mtr/16/

PR #4921, opened 2022-10-28
PS-7963 Do not cache trigger definitions in the table cache for SELEC…

dlenev:
…T statements

Avoid caching fully-loaded/parsed trigger bodies in the Table Cache for read-only statements.

This is done by implementing lazy full-loading/parsing of trigger bodies, i.e. we don't parse and finalize their load by default, but only do this if operation that does data change comes. Once such operation completes we return the TABLE instance and associated with it triggers to the Table Cache. Later we try to prefer reusing TABLE instances with fully-loaded triggers for data change operations and TABLE instances sans trigger bodies for read-only operations. To achieve that Table Cache code was changed to keep two lists of unused TABLE objects for each tables instead of one - with fully-loaded triggers and without them.

To ensure that spike in data change workload won't fill up the Table Cache by TABLE objects with fully-loaded triggers soft limit on number of such objects in the cache was introduced. If this limit is exceeded unused TABLE objects with fully-loaded/parsed triggers are evicted from the cache in LRU fashion.

Users can control this soft limit using new global table_open_cache_triggers variable. By default this variable is set to 524288, so eviction of unused TABLE objects with fully-loaded triggers happens as before this patch.

New status Table_open_cache_triggers_hits/misses/overflows variables were introduced to allow users to see how well TABLE objects with fully-loaded triggers are cached and if table_open_cache_triggers value needs to be adjusted.

### DIFF
--- a/mysql-test/include/assert_trigger_cache_increment.inc
+++ b/mysql-test/include/assert_trigger_cache_increment.inc
@@ -1,0 +1,91 @@
+# ==== Purpose ====
+#
+# Check that values of system status variables related to storing
+# triggers in Table Cache (Table_open_cache_misses,
+# Table_open_cache_trigger_hits, Table_open_cache_triggers_misses,
+# Table_open_cache_triggers_overflows) have got expected increments
+# since the last execution of this script, fail with debug info if
+# they have not.
+#
+# ==== Usage ====
+#
+# --let $expected_inc = { "tc_misses_inc": VALUE1, "trg_hits_inc": VALUE2, "trg_misses_inc" : VALUE3, "trg_overflows_inc" : VALUE4 }
+# --source include/assert_trigger_cache_increment.inc
+#
+# $expected_inc
+#   The JSON object which specifies expected increments.
+#
+# At the end of the test, source include/destroy_json_functions.inc
+# to remove all auxiliary .inc files created by this script.
+#
+# ==== Example ====
+#
+# # Run SELECT reading from table which was absent in Table Cache so far.
+# SELECT count(*) FROM t1;
+#
+# --let $expected_inc = { "tc_misses_inc": 1, "trg_hits_inc": 0, "trg_misses_inc" : 0, "trg_overflows_inc" : 0 }
+# --source include/assert_trigger_cache_increment.inc
+#
+# Resulting output:
+# include/assert.inc [Expected table cache misses increment : 1]
+# include/assert.inc [Expected table cache triggers hits increment : 0]
+# include/assert.inc [Expected table cache triggers misses increment : 0]
+# include/assert.inc [Expected table cache triggers overflows increment : 0]
+#
+
+# If missing, create helper scripts to iterate through bellow JSON array.
+--let $has_json_functions = `SELECT INSTR('$json_function_files','sv_check_inc')`
+if (!$has_json_functions) {
+  --let $json_label = sv_check_inc
+  --let $json_keys = paramname, psname, varname, atext
+  --source include/create_json_unpacking_iterator.inc
+}
+
+# Array describing each status variable to be checked, what is the name
+# of the corresponding script parameter with expected increment, what
+# is name of variable to remember its value for future executions, and
+# what is assertion text to be used for this variable.
+
+let $json_array = [
+  {
+    "paramname" : "tc_misses_inc",
+    "psname": "Table_open_cache_misses",
+    "varname": "tc_misses_count",
+    "atext": "Expected table cache misses increment"
+  },
+  {
+    "paramname" : "trg_hits_inc",
+    "psname": "Table_open_cache_triggers_hits",
+    "varname": "trg_hits_count",
+    "atext": "Expected table cache triggers hits increment"
+  },
+  {
+    "paramname" : "trg_misses_inc",
+    "psname": "Table_open_cache_triggers_misses",
+    "varname": "trg_misses_count",
+    "atext": "Expected table cache triggers misses increment"
+  },
+  {
+    "paramname" : "trg_overflows_inc",
+    "psname": "Table_open_cache_triggers_overflows",
+    "varname": "trg_overflows_count",
+    "atext": "Expected table cache triggers overflows increment"
+  }
+];
+
+# Iterate through the above array checking increment for each variable
+# it describes.
+--source $json_sv_check_inc_start
+
+while (!$json_sv_check_inc_done) {
+  --disable_query_log
+  --eval SET @oldvalue = ifnull(@$varname, 0)
+  --eval SELECT variable_value INTO @$varname FROM performance_schema.session_status WHERE variable_name ='$psname'
+  --let $expinc = `SELECT JSON_EXTRACT('$expected_inc', '$.$paramname')`
+  --enable_query_log
+  --let $assert_text = $atext : $expinc
+  --let $assert_cond = [SELECT @$varname - @oldvalue] = $expinc
+  --source include/assert.inc
+
+  --source $json_sv_check_inc_next
+}
--- a/mysql-test/t/table_open_cache_triggers_functionality.test
+++ b/mysql-test/t/table_open_cache_triggers_functionality.test
@@ -1,0 +1,181 @@
+--echo #
+--echo # Test coverage for limit on number of table objects in Table Cache
+--echo # with fully-loaded triggers.
+--echo #
+
+--echo
+--echo # Create some test tables with and without triggers.
+CREATE TABLE t1 (i INT);
+INSERT INTO t1 VALUES (1), (2), (3);
+CREATE TRIGGER bi_t1 BEFORE INSERT ON t1 FOR EACH ROW SET @a = @a + 1;
+CREATE TABLE t2 (j INT);
+INSERT INTO t2 VALUES (1);
+CREATE TRIGGER ai_t2 AFTER INSERT ON t2 FOR EACH ROW SET @b = @b + 1;
+CREATE TABLE t3 (k INT);
+INSERT INTO t3 VALUES (1);
+CREATE TRIGGER bi_t3 BEFORE INSERT ON t3 FOR EACH ROW SET @c = @c + 1;
+CREATE TABLE t4 (l INT);
+INSERT INTO t4 VALUES (1);
+
+--echo
+--echo # Set small soft limit on number of table objects in Table Cache with
+--echo # fully-loaded triggers, so we can observe it kicking in.
+SET @save_table_open_cache_triggers = @@global.table_open_cache_triggers;
+SET GLOBAL table_open_cache_triggers = 2;
+
+--echo # To make further test repeatable reset Table Cache.
+FLUSH TABLES;
+--echo # But Ensure that it contains necessary DD tables.
+--disable_result_log
+--disable_query_log
+CREATE TABLE t0 (h INT);
+INSERT INTO t0 VALUES (1);
+CREATE TRIGGER bi_t0 BEFORE UPDATE ON t0 FOR EACH ROW SET @h = @h + 1;
+SELECT COUNT(*) FROM t0;
+DROP TABLE t0;
+SELECT variable_value FROM performance_schema.session_status;
+--enable_query_log
+--enable_result_log
+--echo # Reset status variables to get 0 as their base value.
+FLUSH STATUS;
+
+--echo
+--echo # Read-only statements should not need table objects with fully-loaded
+--echo # triggers and thus change status variables tracking their usage.
+--echo # Otherwise absence of object in table cache causes TC miss.
+--echo # presence - TC hit. Read-only statement requiring 2 table objects
+--echo # when only 1 is present in TC causes 1 hit and 1 miss.
+SELECT count(*) FROM t1;
+
+--let $expected_inc = { "tc_misses_inc": 1, "trg_hits_inc": 0, "trg_misses_inc" : 0, "trg_overflows_inc" : 0 }
+--source include/assert_trigger_cache_increment.inc
+
+SELECT count(*) FROM t1 AS a, t1 AS b;
+
+--let $expected_inc = { "tc_misses_inc": 1, "trg_hits_inc": 0, "trg_misses_inc" : 0, "trg_overflows_inc" : 0 }
+--source include/assert_trigger_cache_increment.inc
+
+SELECT count(*) FROM t4;
+
+--let $expected_inc = { "tc_misses_inc": 1, "trg_hits_inc": 0, "trg_misses_inc" : 0, "trg_overflows_inc" : 0 }
+--source include/assert_trigger_cache_increment.inc
+
+--echo
+--echo # Updating statement needs fully-loaded triggers, so there will be
+--echo # a table cache hit, but also trigger miss and trigger will be
+--echo # loaded.
+INSERT INTO t1 VALUES (4);
+
+--let $expected_inc = { "tc_misses_inc": 0, "trg_hits_inc": 0, "trg_misses_inc" : 1, "trg_overflows_inc" : 0 }
+--source include/assert_trigger_cache_increment.inc
+
+--echo
+--echo # Repeating the statement will cause table cache and trigger hits.
+INSERT INTO t1 VALUES (5);
+
+--let $expected_inc = { "tc_misses_inc": 0, "trg_hits_inc": 1, "trg_misses_inc" : 0, "trg_overflows_inc" : 0 }
+--source include/assert_trigger_cache_increment.inc
+
+--echo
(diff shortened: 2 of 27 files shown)


sql/table_cache.h
```diff
@@ -157,6 +175,20 @@ class Table_cache {
...
 
+  /**
+    Notify the table cache that for one of its TABLE objects we have
+    finalized loading and parsing triggers.
+  */
+  void notify_triggers_load() {
+    // QQ: Study performance overhead from this. Perhaps it is better to
+    // use atomics instead?
+    lock();
+    m_table_triggers_count++;
```
percona-ysorokin: Completely agree here.
`std::atomic<>` should do the work

sql/table_trigger_dispatcher.h
```diff
@@ -63,7 +68,9 @@ class Table_trigger_dispatcher : public Table_trigger_field_support {
  public:
   static Table_trigger_dispatcher *create(TABLE *subject_table);
 
-  bool check_n_load(THD *thd, const dd::Table &table);
+  bool initial_load(THD *thd, const dd::Table &table);
+
+  bool finalize_load(THD *thd);
```
percona-ysorokin, line 71: Do we have access to the `dd::Table` in the `finalize_load()` calling point so that we could load and parse trigger bodies there?

sql/table_cache.h
```diff
@@ -236,7 +268,14 @@ class Table_cache_element {
       TABLE_list;
 
   TABLE_list used_tables;
-  TABLE_list free_tables;
+  /**
+    List of unused TABLE objects for tables without triggers or unused TABLE
+    objects for which triggers were not fully-loaded, so they can only can be
+    used by read-only statements.
+  */
+  TABLE_list free_tables_slim;
```
percona-ysorokin, line 277: @dlenev This is the part that is the most confusing to me. I still cannot understand why do we need to split original list into two. Is this only for better control over the max number of cached tables with and without triggers?

mysql-test/t/table_open_cache_triggers_functionality.test
```diff
@@ -0,0 +1,216 @@
...
+--echo #
+
+--echo
+--echo # Create helper verify_status_increment() routine to be able to
+--echo # to check easily how values of Table Cache and Table Cache triggers
+--echo # hits/misses/overflows status variables have changed.
+--disable_query_log
+DELIMITER |;
+CREATE PROCEDURE
+verify_status_increment(tc_misses_inc INT, trg_hits_inc INT, trg_misses_inc INT, trg_overflows_inc INT)
```
percona-ysorokin: May be rework this with MTR `include/assert.inc`s.
Relying on recorded values in the `.result` file is not as safe. People tend to simply re-record `.result` files and such re-records in the `.result` files are very often skipped by reviewers.
Something like,
```
--let $expected_tc_misses = 0
--let $expected_trg_hits = 0
--let $expected_trg_misses = 0
--let $expected_trg_overflows = 0
--source include/assert_cache_misses.inc
```
or even
```
--let $expected_cache_misses_pattern = "0100"
--source include/assert_cache_misses.inc
```
Inside `assert_cache_misses.inc` you would do use `--source include/assert.inc` to check if values match.

sql/table_cache.h
dlenev: Well... I suspect that realistically there should not be that much difference. As this is not supposed to be hot code path - i.e. cases when you need to load triggers for TABLE object should be normally rare and trigger loading is going to be fairly expensive in any case.
Also it might be more convenient to use the lock()/unlock() in this case if we will decide that we want to be more active enforcing soft limit on TABLE objects with triggers loaded and that we want to check and expel unused objects if number of such objects is exceeded when "opening" table/loading triggers. Current version of thepatch does this only when table is "closed".

sql/table_trigger_dispatcher.h
dlenev: Not easily. In theory, we can acquire dd::Table object from DD API and pass it to finalize_load(). But this might be a bit expensive scalability and code complexity-wise. Let me see if we can store necessary information associated with TABLE_SHARE, load it once during TABLE_SHARE construction and do not have the same information duplicated for each TABLE object.

sql/table_cache.h
dlenev: The idea is to avoid scanning the whole (ok, half) of free_tables list to find a TABLE instance without triggers loaded when we are doing read-only statement. Splitting free_tables list in two - one with triggers and one without them allows to quickly check if unused TABLE with/without triggers exist and retrieve one for usage.

sql/table_trigger_dispatcher.h
dlenev: Follow-up patch solves this problem by doing what was described above.

mysql-test/t/table_open_cache_triggers_functionality.test
dlenev: Changed code along this suggestion.

sql/table_cache.h
dlenev: Implemented.

mysql-test/include/assert_trigger_cache_increment.inc
```diff
@@ -0,0 +1,91 @@
...
+# include/assert.inc [Expected table cache triggers misses increment : 0]
+# include/assert.inc [Expected table cache triggers overflows increment : 0]
+# 
+
+# If missing, create helper scripts to iterate through bellow JSON array.
+--let $has_json_functions = `SELECT INSTR('$json_function_files','sv_check_inc')`
+if (!$has_json_functions) {
+  --let $json_label = sv_check_inc
+  --let $json_keys = paramname, psname, varname, atext
+  --source include/create_json_unpacking_iterator.inc
```
percona-ysorokin, line 41: Wow! This is really cool. I didn't even know such functionality exist in MTR.

sql/sql_base.cc
```diff
@@ -3499,6 +3502,27 @@ retry_share : {
...
+    */
+    if (table->triggers->has_load_been_finalized()) {
+      // Common case. We've got TABLE instance with fully ready triggers.
+      thd->status_var.table_open_cache_triggers_hits++;
+    } else {
+      /*
+        Rare case. We've got either fresh TABLE object or TABLE object,
+        which was used only by read-only statements so far.
+
+        QQ: This is not quite a miss. Perhaps better call it
```
percona-ysorokin: If apply simple logic like `everything that is not a 'hit' is a 'miss'` then everything is OK here.

sql/sys_vars.cc
```diff
@@ -5296,6 +5296,30 @@ static Sys_var_ulong Sys_table_cache_instances(
...
+    table cache.
+  */
+  table_cache_triggers_per_instance =
+      table_cache_triggers / table_cache_instances;
+  return false;
+}
+
+// QQ: Invent better name and description? See also status variables.
+static Sys_var_ulong Sys_table_cache_triggers(
+    "table_open_cache_triggers",
```
percona-ysorokin, line 5312: To be honest, I don't like even existing `table_open_cache` name. Personally, I would have named it `table_definition_cache_size` / `table_descriptor_cache_size`.
However, as `table_open_cache` has been there for years, it wouldn't be smart to name the new variable completely differently.
So, `table_open_cache_triggers` probably serves its purposes - it looks similar to `table_open_cache` and has the word `triggers` in it.

sql/trigger.h
```diff
@@ -50,10 +50,27 @@ typedef ulonglong sql_mode_t;
...
+        TABLE instances. The set for the former is used to store static
+        information about table's triggers and is directly associated with
+        TABLE_SHARE object. The set for the latter is used primarily for
+        trigger execution, and is asssociated with TABLE object with the help
+        of Table_triggers_dispatcher class. Attributes representing static
+        properties in Trigger instances of the latter type reference
+        attributes/memory belonging to attributes of Trigger objects associated
+        with the TABLE_SHARE.
+
+  QQ: Ideally, there should be two different classes for per-TABLE_SHARE and
```
percona-ysorokin: Let's not do this as a part of this PR.

unittest/gunit/table_cache-t.cc
```diff
@@ -500,6 +823,228 @@ TEST_F(TableCacheSingleCacheTest, CacheGetAndRelease) {
...
+  table_cache->unlock();
+  table_4->triggers->finalize_load(thd);
+  EXPECT_TRUE(table_4->triggers->has_load_been_finalized());
+  table_cache->lock();
+
+  // Number of TABLE instances stays the same.
+  EXPECT_EQ(4U, table_cache->cached_tables());
+  // Number of TABLE instances with triggers increases and crosses
+  // threshold, since loading of triggers doesn't cause freeing
+  // TABLE objects per se (QQ perhaps it should?).
```
percona-ysorokin: To be honest, this is a reasonable concern. I understand that this logic is simpler but it invalidates the invariant of the `loaded_triggers_tables()` not being greater than `table_open_cache_triggers`.
As you have a test case for this, it is probably OK but I am still worried that this may trigger an assertion somewhere.

percona-ysorokin (commented):
LGTM
My local testing on Ubuntu Jammy fas fine:
* ./mtr main.table_open_cache_triggers_functionality
* ./mtr main.table_open_cache_triggers_functionality --repeat=2 (repeat)
* ./mtr main.table_open_cache_triggers_functionality --repeat=8 --parallel=2 (parallel + repeat)
* ./mtr main.table_open_cache_triggers_functionality --repeat=192 --parallel=24 (overprovisioned)

percona-ysorokin:
@dlenev I discovered a problem with the gunit tests under ASan
`./runtime_output_directory/merge_large_tests-t --gtest_filter=TableCacheSingleCacheTest.*`
```
Note: Google Test filter = TableCacheSingleCacheTest.*
[==========] Running 5 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 5 tests from TableCacheSingleCacheTest
[ RUN      ] TableCacheSingleCacheTest.CacheAddAndRemove
[       OK ] TableCacheSingleCacheTest.CacheAddAndRemove (4 ms)
[ RUN      ] TableCacheSingleCacheTest.CacheOverflow
[       OK ] TableCacheSingleCacheTest.CacheOverflow (1 ms)
[ RUN      ] TableCacheSingleCacheTest.CacheGetAndRelease
[       OK ] TableCacheSingleCacheTest.CacheGetAndRelease (1 ms)
[ RUN      ] TableCacheSingleCacheTest.CacheGetAndReleaseWithTriggers
AddressSanitizer:DEADLYSIGNAL
=================================================================
==358944==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f9eba1161c0 bp 0x7ffd00321c30 sp 0x7ffd00321c30 T0)
==358944==The signal is caused by a READ memory access.
==358944==Hint: address points to the zero page.
    #0 0x7f9eba1161c0 in base_list_iterator::next_fast() /home/yura/ws/percona-server/sql/sql_list.h:369
    #1 0x7f9ebac94f58 in List_iterator_fast<Trigger>::operator++(int) /home/yura/ws/percona-server/sql/sql_list.h:605
    #2 0x7f9ebae6a7de in Table_trigger_dispatcher::finalize_load(THD*) /home/yura/ws/percona-server/sql/table_trigger_dispatcher.cc:344
    #3 0x55ff705b78ee in table_cache_unittest::TableCacheSingleCacheTest_CacheGetAndReleaseWithTriggers_Test::TestBody() /home/yura/ws/percona-server/unittest/gunit/table_cache-t.cc:598
    #4 0x55ff70743a60 in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) /home/yura/ws/percona-server/extra/googletest/googletest-release-1.11.0/googletest/src/gtest.cc:2607
    #5 0x55ff7076a643 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) /home/yura/ws/percona-server/extra/googletest/googletest-release-1.11.0/googletest/src/gtest.cc:2643
    #6 0x55ff70738c20 in testing::Test::Run() /home/yura/ws/percona-server/extra/googletest/googletest-release-1.11.0/googletest/src/gtest.cc:2682
    #7 0x55ff70738f63 in testing::TestInfo::Run() /home/yura/ws/percona-server/extra/googletest/googletest-release-1.11.0/googletest/src/gtest.cc:2861
    #8 0x55ff70739232 in testing::TestSuite::Run() /home/yura/ws/percona-server/extra/googletest/googletest-release-1.11.0/googletest/src/gtest.cc:3015
    #9 0x55ff70739db5 in testing::internal::UnitTestImpl::RunAllTests() /home/yura/ws/percona-server/extra/googletest/googletest-release-1.11.0/googletest/src/gtest.cc:5855
    #10 0x55ff70744c4c in bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) /home/yura/ws/percona-server/extra/googletest/googletest-release-1.11.0/googletest/src/gtest.cc:2607
    #11 0x55ff7076b40e in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) /home/yura/ws/percona-server/extra/googletest/googletest-release-1.11.0/googletest/src/gtest.cc:2643
    #12 0x55ff7073a0f7 in testing::UnitTest::Run() /home/yura/ws/percona-server/extra/googletest/googletest-release-1.11.0/googletest/src/gtest.cc:5438
    #13 0x55ff706cdd12 in RUN_ALL_TESTS() /home/yura/ws/percona-server/extra/googletest/googletest-release-1.11.0/googletest/include/gtest/gtest.h:2490
    #14 0x55ff706cdc1b in main /home/yura/ws/percona-server/unittest/gunit/gunit_test_main_server.cc:63
    #15 0x7f9eb2a39d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #16 0x7f9eb2a39e3f in __libc_start_main_impl ../csu/libc-start.c:392
    #17 0x55ff6f91b314 in _start (/home/yura/ws/percona-build-8.0_zenfs-asan_gcc12/runtime_output_directory/merge_large_tests-t+0xa9f314)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /home/yura/ws/percona-server/sql/sql_list.h:369 in base_list_iterator::next_fast()
==358944==ABORTING
```

mysql-test/include/assert_trigger_cache_increment.inc
dlenev: Yeah... This and some other JSON-related helper scripts were added to upstream in 2021.

sql/sql_base.cc
dlenev: OK.

sql/sys_vars.cc
dlenev: OK. Let us keep this name as is then.

sql/trigger.h
dlenev: Agreed then.

unittest/gunit/table_cache-t.cc
dlenev: I am not concerned about assertion failures in this case. After all table_open_cache_triggers is a new variable, so there can't be pre-existing assertions about it, and I am pretty sure I have not added one :)

Regarding breaking invariant. I don't think there is Table_cache::m_table_triggers_count < table_open_cache_triggers invariant - after all the latter **is only a soft limit** which can be exceeded if currently running DML statements need more TABLE instances with triggers. In principle we can introduce something like "m_table_triggers_count < table_open_cache_triggers || m_unused_triggers_lru.is_empty()" as invariant... But it is we authors of this patch who make this decision and I am not sure if it is really worth it to be honest.

So perhaps we should keep things as is here as well? What do you think?

dlenev:
Regarding TableCacheSingleCacheTest unit test crash:

Ooops!!! My bad! Thanks a lot for noticing this!!! Fixed now.

unittest/gunit/table_cache-t.cc
dlenev: We agreed to keep things as is during discussion Slack.

dlenev:
Squashed all commits in the branch into a single one and submitted it for final review.

dlenev:
Here are results of Jenkins runs for this PR:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param-parallel-mtr/19/
I don't see any failures related to this patch there.

percona-ysorokin (approved):
LGTM

PR #4724, opened 2022-04-11
PS-8022 feature: Percona Server for ARM platform

percona-ysorokin:
https://jira.percona.com/browse/PS-8022

Initial support for RocksDB Storage Engine on ARM platform.

'RocksDB' Storage Engine main 'CMakeLists.txt' file extended with additional
logic that detects 'armv8-a+crc+crypto' instruction set support that
allows to enable 'FastCRC32' calculations.

Also, introduced 'ALLOW_NO_ARMV8A_CRC_CRYPTO' CMake option that,
when set to 'ON', allows to build RocksDB on ARM platform without
'+crc+crypto' extensions.

PR #4922, opened 2022-10-28
PS-8449 Change MyRocks cmake files to allow Rasperry Pi builds

peppla:
When you try to build Percona Server For MySQL with MyRocks enabled on a Raspberry PI, the following error is displayed: x86_64 is only platform supported.  aarch64 found.  Not building MyRocks

Fixed by modifying the condition that filters out unsupported cpu architectures. Also added a condition to add flags that only apply to x86_64. If I am correct, similar kind of flags are not required for aarch64 if it supports crc.

### DIFF
--- a/storage/rocksdb/CMakeLists.txt
+++ b/storage/rocksdb/CMakeLists.txt
@@ -22,8 +22,8 @@ ELSE ()
 ENDIF ()
 
 # check platform support, no 32 bit
-IF (NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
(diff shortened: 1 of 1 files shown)


percona-ysorokin:
@peppla A few days ago we merged this PR https://github.com/percona/percona-server/pull/4724 to the `release-8.0.30-22` branch. So the ability to build RocksDB on ARM will be included in the upcoming `Percona Server for MySQL 8.0.30-22` release.
The version we merged does even more - it detects ARM processor capabilities and conditionally enables FAST CRC32 calculations (using extended processor instruction sets).
So, if you don't have any objections, I am going to close this PR.

peppla:
> So, if you don't have any objections, I am going to close this PR.

Yes, please. I made the mistake of searching for Raspberry PI rather than searching for ARM in Jira. Sorry not to realize that this was going to be fixed.

PR #4924, opened 2022-11-02
PS-8330 merge: Merge MySQL 8.0.30 - Q3 2022 (ROCKSDB_VALGRIND_RUN)

percona-ysorokin:
https://jira.percona.com/browse/PS-8330

'-DROCKSDB_VALGRIND_RUN' preprocessor definition is now also passed to the 'RocksDB' code when top-level 'WITH_VALGRIND' CMake option is set to 'ON'.

This allows to eliminate
"==10176== Warning: unimplemented fcntl command: 1036" Valgrind warnings coming from 'rocksdb/env/io_posix.cc' and false positives coming from not calling some static object destructors.


inikep (approved):
LGTM. Did you also try `-DROCKSDB_FULL_VALGRIND_RUN`? I found:
```
	# Some tests are slow when run under valgrind and are only run when
	# explicitly requested via the ROCKSDB_FULL_VALGRIND_RUN compiler flag.
	PLATFORM_CCFLAGS += -DROCKSDB_VALGRIND_RUN -DROCKSDB_FULL_VALGRIND_RUN
	PLATFORM_CXXFLAGS += -DROCKSDB_VALGRIND_RUN -DROCKSDB_FULL_VALGRIND_RUN
```

percona-ysorokin:
> LGTM. Did you also try `-DROCKSDB_FULL_VALGRIND_RUN`? I found:
> 
> ```
> 	# Some tests are slow when run under valgrind and are only run when
> 	# explicitly requested via the ROCKSDB_FULL_VALGRIND_RUN compiler flag.
> 	PLATFORM_CCFLAGS += -DROCKSDB_VALGRIND_RUN -DROCKSDB_FULL_VALGRIND_RUN
> 	PLATFORM_CXXFLAGS += -DROCKSDB_VALGRIND_RUN -DROCKSDB_FULL_VALGRIND_RUN
> ```

Yeah, I saw this as well but it looks like this macro affects only RocksDB unit test code that we do not use

PR #4927, opened 2022-11-04
Fix to the GR garbage collector stalls every minute (PS-8450)

y-trudeau:
This is a patch fixing the stalls of GR every minute caused by the garbage collector.  The garbage collection is broken into chunks and a short sleep time with the locks released is added between the chunks to allow transaction to interleave.  This greatly reduced the latency peaks during these events. If the garbage collection runs too slowly, the chunks size is increased every 5 seconds.  Two new variables are introduced:

group_replication_certification_loop_chunk_size
group_replication_certification_loop_sleep_time


plugin/group_replication/src/certifier.cc
```diff
@@ -1223,8 +1247,16 @@ void Certifier::garbage_collect() {
   }
 #endif
 
-  mysql_mutex_unlock(&LOCK_certification_info);
+    mysql_mutex_unlock(&LOCK_certification_info);
+    if (it == certification_info.end()) {
```
dutow, line 1251: I don't think we should access this variable without holding the LOCK_certification_info mutex.

plugin/group_replication/src/certifier.cc
```diff
@@ -1191,23 +1196,42 @@ void Certifier::garbage_collect() {
...
+    stable_gtid_set_lock->unlock();
+
+    /*
+      We need to update parallel applier indexes since we do not know
+      what write sets were purged, which may cause transactions
+      last committed to be incorrectly computed.
+    */
+    increment_parallel_applier_sequence_number(true);
 
 #if !defined(NDEBUG)
```
dutow, line 1236: Please also fix indentation inside this ifdef

plugin/group_replication/src/certifier.cc
```diff
@@ -1191,23 +1196,42 @@ void Certifier::garbage_collect() {
...
-  }
-  stable_gtid_set_lock->unlock();
 
-  /*
-    We need to update parallel applier indexes since we do not know
-    what write sets were purged, which may cause transactions
-    last committed to be incorrectly computed.
-  */
-  increment_parallel_applier_sequence_number(true);
+  while (1) {
```
dutow, line 1207: Since we are already holding the lock at this point, wouldn't it be better to add the condition here instead of the break later?

```
while(it != certification_info.end()))
```

plugin/group_replication/src/certifier.cc
```diff
@@ -1191,23 +1196,42 @@ void Certifier::garbage_collect() {
...
+    stable_gtid_set_lock->wrlock();
+
+    /* Needs to increase the rate if it takes too long, add a chunk every 5s */
```
dutow, line 1210: Is this hard coded 5 second a good solution? The default sleep time is 200ms, but it can be set up to 1ms - based on its value, we can have either 5, or 50 iterations before reaching this.

Also shouldn't we provide a warning / other feedback to the user that we (temporarily) aren't using the settings specified in the currently active configuration?

plugin/group_replication/src/certifier.cc
venkatesh-prasad-v: @dutow Wouldn't this start the garbage collection process start from beginning? With the current logic, it starts from where it had left?

dutow: No, this would only move the end condition.

Currently we have an "infinite" where loop, with a break inside an if at line 1251, if `it == certification_info.end()`

All I am saying  is that we could move that condition inside the condition of the while loop. It wouldn't change how the loop is executed, because we also have a similar condition at the start of the for loop at line 1214.

plugin/group_replication/src/certifier.cc
```diff
@@ -1223,8 +1247,16 @@ void Certifier::garbage_collect() {
   }
 #endif
 
-  mysql_mutex_unlock(&LOCK_certification_info);
+    mysql_mutex_unlock(&LOCK_certification_info);
+    if (it == certification_info.end()) {
+      break;
+    }
+    /* could add else clause with a short sleep */
+    my_sleep(get_certification_loop_sleep_time_var());
```
dutow, line 1255: Based on the comment from Venkatesh, I just realized another unlikely, but possible issue:

We currently:
* get to a point with this iterator, based on m_certification_info
* release the lock
* wait some time
* lock again
* continue work with the same iterator, assuming that it is still valid

But are we 100% sure that the iterator doesn't get invalidated while we wait? We don't hold the lock, any other method can change the collection.

Currently it's a map, and only `garbage_collect()` and `clear_certification_info()` delete items from it. But:

1. I'm not sure that with the lock relaxed, these can't happen in parallel
2. Will we notice if upstream changes the type of the container, or adds other methods that remove items?

plugin/group_replication/src/certifier.cc
venkatesh-prasad-v: Ok. Got it. Thanks for the clarification.

venkatesh-prasad-v:
Hi @dutow @y-trudeau I have addressed the review comments from Zsolt and have created https://github.com/percona/percona-server/pull/4991. Let's us the other PR for discussions.

PR #4929, opened 2022-11-08
PS-8454: Fido authentication crashes after incorrect alter

dutow:
Issue: Login with fido crashes the server after an incorrect ALTER adding a fido factor.

Fix: the generic MFA code calls the plugin twice after the alter, with incorrectly set up data structures.

This commit adds sanity checks to the fido authentication function which returns an error in this case.

It also restores an unneccessary change in the generic mfa code, added in the original fido PR.

### DIFF
--- a/plugin/authentication_fido/src/plugin_fido.cc
+++ b/plugin/authentication_fido/src/plugin_fido.cc
@@ -257,6 +257,10 @@ bool reconstruct_fido_cred(fido_cred_t *cred, const char *data,
 int fido_authenticate(MYSQL_PLUGIN_VIO *vio, MYSQL_SERVER_AUTH_INFO *info) {
   unsigned char *response;
 
+  if (info == nullptr) {
+    return CR_ERROR;  // PS-8454
+  }
+
   if (info->multi_factor_auth_info &&
       info->multi_factor_auth_info[info->current_auth_factor]
           .is_registration_required) {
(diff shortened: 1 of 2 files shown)


plugin/authentication_fido/src/plugin_fido.cc
```diff
@@ -282,6 +286,9 @@ int fido_authenticate(MYSQL_PLUGIN_VIO *vio, MYSQL_SERVER_AUTH_INFO *info) {
   std::vector<unsigned char> decoded(512);
   auto decoded_len = base64_decode(info->auth_string, info->auth_string_length,
                                    decoded.data(), nullptr, 0);
+  if (decoded_len < 65) {
```
altmannmarcelo: where 65 comes from ?
Do we have a pre-processor macro or constant defined anywhere for this?

altmannmarcelo (changes requested):
just one minor question

plugin/authentication_fido/src/plugin_fido.cc
dutow: Later in the code (the memcpy a few lines below), we read the data starting with the 64th byte. I increased it to 65 so we also report an error if it's zero length.

dutow: I added a const for this.

altmannmarcelo (approved):
LGTM

PR #4933, opened 2022-11-22
PS-8433: Reduce memory overhead from cached triggers by removing fixe…

dlenev:
…d size buffer for error.

https://jira.percona.com/browse/PS-8433

For tables with triggers, each TABLE instance in Table Cache has one Table_trigger_dispatcher sub-object and one or more Trigger sub-object associated with it.
Both these classes contain as members 512-byte buffer to store error message which occurred during trigger parsing. Since these buffers are not used in 99% of cases this looks like pure waste of memory.

This patch improves the situation by replacing these always-allocated fixed-size buffers with variable-size buffers which are allocated on TABLE's memory root on demand, only when the table has a trigger with a parse error.

### DIFF
--- a/sql/table_trigger_dispatcher.cc
+++ b/sql/table_trigger_dispatcher.cc
@@ -88,9 +88,8 @@ Table_trigger_dispatcher::Table_trigger_dispatcher(TABLE *subject_table)
       m_record1_field(nullptr),
       m_new_field(nullptr),
       m_old_field(nullptr),
-      m_has_unparseable_trigger(false) {
+      m_parse_error_message(nullptr) {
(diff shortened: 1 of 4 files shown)


percona-ysorokin (approved):
LGTM

dlenev:
Here are results of Jenkins run for this change: https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param-parallel-mtr/20/

dlenev:
> Here are results of Jenkins run for this change: https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param-parallel-mtr/20/

Per discussion on Slack, issues reported by Jenkins seem not to be related to this patch.

PR #4938, opened 2022-11-25
PS-8502 [5.7]: Building issue with ccache on Azure

inikep:
1. Remove `##vso[task.prependpath]`.
2. Use `-DCMAKE_C_COMPILER_LAUNCHER=ccache` and `-DCMAKE_CXX_COMPILER_LAUNCHER=ccache`.
3. Use `-DCMAKE_C_COMPILER=$SELECTED_CC` and `-DCMAKE_CXX_COMPILER=$SELECTED_CXX`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,7 +1,7 @@
 jobs:
 - job: BiDiScan
   pool:
-    vmImage: 'ubuntu-18.04'
+    vmImage: 'ubuntu-20.04'
 
   steps:
   - checkout: self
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -300,19 +300,22 @@ jobs:
...
+          PACKAGES="$SELECTED_CC $PACKAGES"
         else
-          CXX=g++-$(CompilerVer)
-          PACKAGES="$CXX $PACKAGES"
+          SELECTED_CXX=g++-$(CompilerVer)
+          PACKAGES="$SELECTED_CXX $PACKAGES"
         fi
+      else
+        SELECTED_CC=$(Compiler)
+        SELECTED_CXX=$(Compiler)
```
percona-ysorokin: @inikep I believe here we need `++` for `CXX`
may be just
```suggestion
        SELECTED_CC=gcc
        SELECTED_CXX=g++
```

inikep: I think `clang` and `clang++` is more precise.

percona-ysorokin (approved):
LGTM

PR #4940, opened 2022-11-28
PS-8301 [5.7]: Disable Azure pipelines testing for certain branches

inikep:
Disable Azure pipelines for branches that don't change source code and don't influece compilation.


percona-ysorokin (approved):
LGTM

PR #4941, opened 2022-11-28
PS-8301 [8.0]: Disable Azure pipelines testing for certain branches

inikep:
Disable Azure pipelines for branches that don't change source code and don't influence compilation.


percona-ysorokin (approved):
LGTM

PR #4948, opened 2022-12-02
PS-8442: Fix crash in buf_pool_invalidate_instance()

oleksandr-kachan:
https://jira.percona.com/browse/PS-8442

Revert changes introducing recv_writer_thread logic back in 8.0.30.
This thread is not needed in PS, its job is done by LRU threads.


storage/innobase/log/log0recv.cc
```diff
@@ -1331,17 +1331,26 @@ dberr_t recv_apply_hashed_log_recs(log_t &log, bool allow_ibuf) {
 
     mutex_exit(&recv_sys->mutex);
 
+    /* Stop the recv_writer thread from issuing any LRU
+    flush batches. */
+    mutex_enter(&recv_sys->writer_mutex);
+
```
satya-bodapati: Please check about recv_writer_thread() in previous 8.0 versions. 

I think we made a Merge mistake in 8.0.30. We never rely on recv_writer_thread() to do LRU list flushing.
From 8.0.30, we brought upstream logic as is.

During recovery too, our LRU threads will do the flushing.

Please revert the entire recv_writer_thread logic and this change too and test.

If it still fails, the fix should be something else.

satya-bodapati (changes requested):

storage/innobase/log/log0recv.cc
oleksandr-kachan: Removed  recv_writer_thread() related code

satya-bodapati:
@oleksandr-kachan Can you please paste any result that confirms removal of this rec writer thread solves the buf_pool_validate_instance crash?

oleksandr-kachan:
> @oleksandr-kachan Can you please paste any result that confirms removal of this rec writer thread solves the buf_pool_validate_instance crash?

@satya-bodapati, just run previously failing test on Jenkins https://ps80.cd.percona.com/job/percona-server-8.0-pipeline/40888/

percona-ysorokin (approved):
LGTM

PR #4955, opened 2022-12-13
PS-8073: Replication crash when updating a table with a stored genera…

voodam:
…ted field in the primary key

https://jira.percona.com/browse/PS-8073

Problem:
When updating a table with a stored generated field in a primary key with binlog_row_image=minimal on a master, MySQL can't find the updated record on a replica.

Cause:
In f14a832e464b fix (a commit causing the bug) generated fields are updated in unpack_current_row. If a binlog row image doesn't contain base fields which a generated column depends on, generated fields will be updated incorrectly. Then they won't be found by Innodb.

Solution:
Exclude generated fields from update which base fields are excluded from the binlog row image.

This bug is reproduced in 5.7, but it requires additional commits from 8.0 to be fixed. I think, f14a832e464b can be required. Do we need to fix it in 5.7 anyway?

### DIFF
--- a/mysql-test/suite/rpl/r/rpl_bug_ps_8073.result
+++ b/mysql-test/suite/rpl/r/rpl_bug_ps_8073.result
@@ -1,0 +1,90 @@
+include/master-slave.inc
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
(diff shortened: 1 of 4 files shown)


percona-ysorokin (approved):
LGTM

voodam:
Hi @kamil-holubicki, do you want to review or can I merge?

kamil-holubicki (approved):

PR #4956, opened 2022-12-13
PS-8422: Merge MySQL 8.0.31 (fix percona_bug_ps8106 test)

oleksandr-kachan:

Use replica_parallel_workers = 1 instead 0 to fix warning introduced at https://github.com/mysql/mysql-server/commit/4194cd4fcc92fad94a403753661263e32a028ae3

### DIFF
--- a/mysql-test/suite/rpl/r/percona_bug_ps8106.result
+++ b/mysql-test/suite/rpl/r/percona_bug_ps8106.result
@@ -4,7 +4,7 @@ Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
 Note	####	Storing MySQL user name or password information in the master info repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START SLAVE; see the 'START SLAVE Syntax' in the MySQL Manual for more information.
 [connection master]
 SET @replica_parallel_workers_save = @@global.replica_parallel_workers;
-SET GLOBAL replica_parallel_workers = 0;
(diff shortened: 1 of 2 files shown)


mysql-test/suite/rpl/r/percona_bug_ps8106.result
```diff
@@ -5,6 +5,8 @@ Note	####	Storing MySQL user name or password information in the master info rep
 [connection master]
 SET @replica_parallel_workers_save = @@global.replica_parallel_workers;
 SET GLOBAL replica_parallel_workers = 0;
```
percona-ysorokin: May be set `replica_parallel_workers` to `1` in the `.test`file instead. It looks like `0` and `1` have the same meaning.`1` just doesn't trigger a warning.
https://dev.mysql.com/doc/refman/8.0/en/replication-options-replica.html#sysvar_replica_parallel_workers
```
Beginning with MySQL 8.0.30, setting this variable to 0 is deprecated, and doing so raises a warning; 0 as a permitted value for replica_parallel_workers is subject to removal in a future MySQL release; set it to 1 instead, which has the same effect.
```

percona-ysorokin (changes requested):

mysql-test/suite/rpl/r/percona_bug_ps8106.result
oleksandr-kachan: Done

percona-ysorokin (approved):
LGTM

PR #4959, opened 2022-12-22
Follow-up fix for PS-7507 merge: Merge MySQL 8.0.23 (processlist IS-P…

dlenev:
…FS fix).

Updated result file for long-neglected funcs_1.processlist_priv_ps test.

Apparently, this test was not updated during original PS-7507 merge and earlier changes as it needs to be run with --ps-protocol option.

Now results of this test match results of its sibling non-ps-protocol funcs_1.processlist_priv_no_prot test.


mysql-test/suite/funcs_1/r/processlist_priv_ps.result
```diff
@@ -31,31 +31,34 @@ PROCESSLIST	CREATE TEMPORARY TABLE `PROCESSLIST` (
   `COMMAND` varchar(16) NOT NULL DEFAULT '',
   `TIME` int NOT NULL DEFAULT '0',
   `STATE` varchar(64) DEFAULT NULL,
-  `INFO` longtext
```
percona-ysorokin, line 34: It looks like `funcs_1.processlist_val_ps` has the same problem. I would include it into this commit as well.

percona-ysorokin (changes requested):

mysql-test/suite/funcs_1/r/processlist_priv_ps.result
dlenev: Submitted updated version of patch.

percona-ysorokin (approved):
LGTM

PR #4961, opened 2022-12-22
Follow-up fix for PS-7507 merge: Merge MySQL 8.0.23 (processlist IS-PFS fix)

oleksandr-kachan:
Updated result files for long-neglected funcs_1.processlist_priv_ps and funcs_1.processlist_val_ps tests.

Apparently, these tests were not updated during original PS-7507 merge and earlier changes as they need to be run with --ps-protocol option.

Now results of these tests match results of their sibling non-ps-protocol funcs_1.processlist_priv_no_prot and funcs_1.processlist_val_no_prot tests.

(cherry picked from commit dd335934137d17390d2047e2eedc901b91059377)


percona-ysorokin (approved):
LGTM

PR #4968, opened 2023-01-06
PS-8560 [5.7]: Backport improved detection of CPU instruction sets from 8.0

inikep:
1. Introduce CHECK_VARIABLE_DEFINED cmake function
2. Remove checks for `bmi` and `lzcnt` as they are not used by RocksDB anymore


percona-ysorokin (approved):
LGTM

PR #4969, opened 2023-01-06
PS-8560 [8.0]: Backport improved detection of CPU instruction sets from 8.0

inikep:


inikep:
This is a null-merge from 5.7 to 8.0.

PR #4972, opened 2023-01-09
PS-8337 postfix: remove system key keyring vault unittests

dutow:
These depend on a now removed feature and fail


percona-ysorokin (approved):
LGTM

PR #4975, opened 2023-01-10
Fix fido plugin compiles even -DWITH_PERCONA_AUTHENTICATION_FIDO=OFF

kolya7k:
Fix fido plugin compiles even -DWITH_AUTHENTICATION_FIDO=OFF

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1833,6 +1833,8 @@ MYSQL_CHECK_LDAP_DLLS()
 # Percona LDAP Simple / LDAP SASL authentication plugins
 OPTION(WITH_PERCONA_AUTHENTICATION_LDAP "Build with Percona LDAP Simple / LDAP SASL authentication plugins" ON)
 
+OPTION(WITH_PERCONA_AUTHENTICATION_FIDO "Build with Percona FIDO authentication plugin" ON)
+
 # Add Windows specific jemalloc DLL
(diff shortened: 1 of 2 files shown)


dutow:
Thank you for your  patch!

While it is a good idea to make building the server side FIDO plugin configurable,  based on the variable description it is intended for a different purpose.

This variable is not documented in the MySQL documentation, but please see the documentation of the similar WITH_AUTHENTICATION_LDAP variable:
https://dev.mysql.com/doc/refman/8.0/en/source-configuration-options.html#option_cmake_with_authentication_ldap

Based on this we decided to solve the same issue for the LDAP plugin by introducing a new variable, WITH_PERCONA_AUTHENTICATION_LDAP in this commit: https://github.com/percona/percona-server/commit/d3581e9e2517609fe2eed165a154661f14b58ca4

Could you update this PR with a similar channge, a new WITH_PERCONA_AUTHENTICATION_FIDO instead of reusing the upstream variable?

kolya7k:
You're right, modified patch to use WITH_PERCONA_AUTHENTICATION_FIDO option

dutow (approved):

Aleksa-del:
Hello, @kolya7k ! Thank you for your contributions! We would like to send you a gift from Percona. Please, contact us at community-team@percona.com.

PR #4978, opened 2023-01-16
PS-8422: Merge MySQL 8.0.31 (fix have_fips.inc)

oleksandr-kachan:

The have_fips.inc is used to check if FIPS is supported with current openssl build. Internally it toggle's FIPS mode to see whether it is available.

This approach doesn't work for OSes which come with bundled openssl-3. To fix that test_ssl_fips_mode() method updated to use EVP_default_properties_is_fips_enabled() instead of toggleing current FIPS state.


percona-ysorokin (approved):
LGTM

PR #4979, opened 2023-01-16
PS-8577: ALTER TABLE ... AUTO_INCREMENT=1 does not work as expected.

dlenev:
It was possible to set too low auto-increment value for an InnoDB table by using ALTER TABLE ... AUTO_INCREMENT = ... right after ALTER TABLE that added auto-increment column to the table by modifying existing column. In this case auto-increment value was not adjusted to be greater than already-used values in the column, unlike it is done in case when one tries to set too low auto-increment value for column which was created as an auto-increment from the start. The value is also adjusted if one uses COPY algorithm to executed ALTER TABLE ... AUTO_INCREMENT=...

Not adjusting the value resulted in eventual duplicate errors on attempts to insert new auto-increment values in the table.
Also the fact that behavior of ALTER TABLE ... AUTO_INCREMENT = ... was different depending on whether columns was created as auto-increment from the start or were made auto-increment later, even though both cases seem identical in SHOW CREATE TABLE output, meant that there might have been problems when combination of SQL-dump and binary log was used in backup/ restore scenarios.

The problematic behavior occurred due to fact that after ALTER TABLE which modified existing column to be auto-increment table got 0 as persisted auto-increment value, even in cases when column had some other values in it. As result code of INPLACE ALTER TABLE in InnoDB storeage engine responsible for checking whether new auto-increment value is lower than existing one and needs to be adjusted didn't work properly.

This patch solves the problem by changing code in InnoDB SE handling INPLACE ALTER TABLE ... AUTO_INCREMENT = ... to not to trust 0 persisted auto-increment value and do index check instead. This is similar to what happens when we discover 0 persisted auto-increment value during ha_innobase::open().

Alternative approach, which involves persisting correct auto-increment value during ALTER TABLE which modifies existing column to be auto-increment was considered more intrusive.

### DIFF
--- a/mysql-test/suite/innodb/t/auto_increment.test
+++ b/mysql-test/suite/innodb/t/auto_increment.test
@@ -2306,3 +2306,237 @@ ADD KEY (c);
 DROP TABLE t1;
 
 SET @@SESSION.information_schema_stats_expiry = DEFAULT;
+
+
+--echo #
+--echo # PS-8577: ALTER TABLE ... AUTO_INCREMENT=1 does not work as expected,
+--echo #
+--echo # Test case which is similar to one from the bug report.
+CREATE TABLE t1 (id INT PRIMARY KEY);
+INSERT INTO t1 VALUES (3);
+ALTER TABLE t1 MODIFY id INT NOT NULL AUTO_INCREMENT;
+--echo # Use maximum existing value from auto-increment column + 1 as new
+--echo # auto-increment value if someone tries to set this parameter too low.
+ALTER TABLE t1 AUTO_INCREMENT = 1;
+SHOW CREATE TABLE t1;
+--echo # As result the below INSERT should not cause duplicate key errors
+--echo # (like it has happened before the fix).
+INSERT INTO t1 VALUES (), (), ();
+SELECT * FROM t1;
+DROP TABLE t1;
+
+--echo #
+--echo # Case when column was created as auto-increment from the start
+--echo # worked as expected even before the fix.
+CREATE TABLE t2 (id INT PRIMARY KEY AUTO_INCREMENT);
+INSERT INTO t2 VALUES (3);
+ALTER TABLE t2 AUTO_INCREMENT = 1;
+SHOW CREATE TABLE t2;
+INSERT INTO t2 VALUES (), (), ();
+SELECT * FROM t2;
+DROP TABLE t2;
+
+--echo #
+--echo # Versions of the first test case with explicit ALGORITHM clause.
+--echo #
+--echo # Case with ALGORITHM=INPLACE failed before the fix.
+CREATE TABLE t3 (id INT PRIMARY KEY);
+INSERT INTO t3 VALUES (3);
+ALTER TABLE t3 MODIFY id INT NOT NULL AUTO_INCREMENT;
+ALTER TABLE t3 AUTO_INCREMENT = 1, ALGORITHM=INPLACE;
+SHOW CREATE TABLE t3;
+INSERT INTO t3 VALUES (), (), ();
+SELECT * FROM t3;
+DROP TABLE t3;
+
+--echo #
+--echo # But case with ALGORITHM=COPY worked as expected.
+CREATE TABLE t4 (id INT PRIMARY KEY);
+INSERT INTO t4 VALUES (3);
+ALTER TABLE t4 MODIFY id INT NOT NULL AUTO_INCREMENT;
+ALTER TABLE t4 AUTO_INCREMENT = 1, ALGORITHM=COPY;
+SHOW CREATE TABLE t4;
+INSERT INTO t4 VALUES (), (), ();
+SELECT * FROM t4;
+DROP TABLE t4;
+
+--echo #
+--echo # Test scenarios involving import of tablespace which were also
+--echo # affected by the same problem.
+--let $MYSQLD_DATADIR = `SELECT @@datadir`
+--disable_query_log
+CALL mtr.add_suppression("\\[Warning\\] .* Reading max\\(auto_inc_col\\) = .* for table .*, because there was an IMPORT without cfg file.");
+--enable_query_log
+CREATE TABLE t5 (id INT AUTO_INCREMENT PRIMARY KEY);
+INSERT INTO t5 VALUES (3);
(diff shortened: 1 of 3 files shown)


mysql-test/suite/innodb/t/auto_increment.test
```diff
@@ -2306,3 +2306,56 @@ ADD KEY (c);
...
+INSERT INTO t1 VALUES (3);
+ALTER TABLE t1 MODIFY id INT NOT NULL AUTO_INCREMENT;
+--echo # Use maximum existing value from auto-increment column + 1 as new
+--echo # auto-increment value if someone tries to set this parameter too low.
+ALTER TABLE t1 AUTO_INCREMENT = 1;
+SHOW CREATE TABLE t1;
+--echo # As result the below INSERT should not cause duplicate key errors
+--echo # (like it has happened before the fix).
+INSERT INTO t1 VALUES (), (), ();
+SELECT * FROM t1;
```
percona-ysorokin, line 2325: Add `ORDER BY` for test case stability.

mysql-test/suite/innodb/t/auto_increment.test
```diff
@@ -2306,3 +2306,56 @@ ADD KEY (c);
...
+
+--echo #
+--echo # Case when column was created as auto-increment from the start
+--echo # worked as expected even before the fix.
+CREATE TABLE t2 (id INT PRIMARY KEY AUTO_INCREMENT);
+INSERT INTO t2 VALUES (3);
+ALTER TABLE t2 AUTO_INCREMENT = 1;
+SHOW CREATE TABLE t2;
+INSERT INTO t2 VALUES (), (), ();
+SELECT * FROM t2;
```
percona-ysorokin, line 2336: Likewise

mysql-test/suite/innodb/t/auto_increment.test
```diff
@@ -2306,3 +2306,56 @@ ADD KEY (c);
...
+--echo # Versions of the first test case with explicit ALGORITHM clause.
+--echo #
+--echo # Case with ALGORITHM=INPLACE failed before the fix.
+CREATE TABLE t3 (id INT PRIMARY KEY);
+INSERT INTO t3 VALUES (3);
+ALTER TABLE t3 MODIFY id INT NOT NULL AUTO_INCREMENT;
+ALTER TABLE t3 AUTO_INCREMENT = 1, ALGORITHM=INPLACE;
+SHOW CREATE TABLE t3;
+INSERT INTO t3 VALUES (), (), ();
+SELECT * FROM t3;
```
percona-ysorokin, line 2349: Likewise

mysql-test/suite/innodb/t/auto_increment.test
```diff
@@ -2306,3 +2306,56 @@ ADD KEY (c);
...
+
+--echo #
+--echo # But case with ALGORITHM=COPY worked as expected.
+CREATE TABLE t4 (id INT PRIMARY KEY);
+INSERT INTO t4 VALUES (3);
+ALTER TABLE t4 MODIFY id INT NOT NULL AUTO_INCREMENT;
+ALTER TABLE t4 AUTO_INCREMENT = 1, ALGORITHM=COPY;
+SHOW CREATE TABLE t4;
+INSERT INTO t4 VALUES (), (), ();
+SELECT * FROM t4;
```
percona-ysorokin, line 2360: Likewise

storage/innobase/handler/handler0alter.cc
```diff
@@ -6755,8 +6755,14 @@ static void innobase_rename_or_enlarge_columns_cache(
...
-    in the table and compare. */
-    if (ctx->max_autoinc <= max_value_table) {
+    in the table and compare.
+
+    If persisted auto-increment value is 0, it can't be trusted.
+    It might be an indication that auto-increment column just has
+    been added to the table by modifying existing column, so
+    the real maximum value in it has not been persisted yet.
+    So we do index search in this case as well. */
+    if (!max_value_table || ctx->max_autoinc <= max_value_table) {
```
percona-ysorokin: My internal perfectionist insist on avoiding implicit conversions :)
In my opinion, explicit `max_value_table == 0` is much more readable.

percona-ysorokin (changes requested):

satya-bodapati (commented):
you are right, similar logic exists in innobase_open. When it sees a zero value for auto-inc, it recalculates.

This can happen with import and export testcase (without cfg file specifically)
Can you please add a testcase which does, ALTER AUTO_INCREMENT, then export.

Later import, show create table, and some more DMLs. Please try the import, with & without cfg.

mysql-test/suite/innodb/t/auto_increment.test
dlenev: Could you please clarify what we are trying to achieve by doing this? Making this test storage engine independent perhaps (but it is in innodb suite...)?

The thing is that for InnoDB SE "SELECT * FROM t1;" statement will do a full table scan, which is executed by this SE in the order of PK, so results should be stable already... (and the fact that there are around 1000 uses of "SELECT * FROM t1;" statement in innodb suite only which do not cause problem seem to confirm that).

storage/innobase/handler/handler0alter.cc
dlenev: Will do.

mysql-test/suite/innodb/t/auto_increment.test
percona-ysorokin: For InnoDB case, you are absolutely right.
This is most probably coming from my current work on MTR instabilities in the 8.0.31 release where we have a couple of tests with SELECTS without ORDER BY but those are HEAP and RocksDB.
So, it just caught my eye. Please, ignore.

dlenev: Per agreement keeping this thing as is.

percona-ysorokin (approved):
LGTM

dlenev:
The Jenkins results are here https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param-parallel-mtr/30/. There seem to be no failures related to the patch.

satya-bodapati (approved):

PR #4980, opened 2023-01-17
PS-8503: Fix initialization of innodb_checkpoint_age status vars

oleksandr-kachan:
https://jira.percona.com/browse/PS-8503

After re-introducing innodb_checkpoint_age and innodb_checkpoint_max_age status vars there may be a crash while exporting those status vars in case server is in read only mode. The reason is that redo logs subsystem is not initialized in this case.

Updated srv_export_innodb_status() method to skip checking of checkpoint age when server is run in read only mode.


percona-ysorokin (approved):
LGTM

PR #4988, opened 2023-01-31
PS-8613: Fix lock contention introduced by LDAP authorization

dutow:
Issue: With the LDAP authorization implemented, all connection requests take the ACL_lock to verify external roles. This leads to significantly increased login times with high concurrency.

Fix: only take the lock when it's actually necessary: when the user has external roles configured by the authentication plugin.


percona-ysorokin (approved):
LGTM but not for `release-8.0.31-23`

PR #4990, opened 2023-02-01
PS-7538: Crash with innodb_optimize_fulltext_only (5.7)

dutow:
Issue: innodb_optimize_fulltext_only=ON allows users to use OPTIMIZE TABLE to rebuild fulltext indices.

Internally the fulltext rebuild logic reads innodb_ft_num_word_optimize keys from the index, uses zlib to compress them, and then starts decompressing/iterating them.

There's an error in this decompressing logic: if zlib reports end of stream, the reader method assumes that we correctly read the last entry and returns it.

But it's also possible that we were already at the end of the stream, and we didn't read anything. The variable used for returning the next word is reused from the previous iteration in the optimize loop. If it contains a record that exists in the index, the logic ends up "optimizing" it again. If it contains a value that's not present in the index, it crashes.

Fix: only return success if there was actually something in the zip data.

### DIFF
--- a/storage/innobase/fts/fts0opt.cc
+++ b/storage/innobase/fts/fts0opt.cc
@@ -584,6 +584,7 @@ fts_zip_read_word(
 	void*		null = NULL;
 	byte*		ptr = word->f_str;
 	int		flush = Z_NO_FLUSH;
+	bool		read_something = FALSE;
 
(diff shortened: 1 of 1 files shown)


percona-ysorokin (approved):
LGTM

PR #4994, opened 2023-02-09
PS-8422: Null merge PS 5.7.40 at commit fde2cca41b1 (merge-mysql-5.7.40-part1) into 8.0

oleksandr-kachan:


percona-ysorokin (approved):
LGTM

PR #4997, opened 2023-02-16
PS-8627: Fix statistics calculation in temptable Handler::open()

oleksandr-kachan:
https://jira.percona.com/browse/PS-8627
***
Make sure statistics is updated only in case table is opened successfully.

(cherry picked from commit 236ffd49e481b073538f99c3d5665589212579be)


percona-ysorokin (approved):
LGTM

PR #5000, opened 2023-02-19
PS-8626: Inserting JSON data through LOAD DATA INFILE doesn't show fa…

dlenev:
…iled record.

When LOAD DATA or multi-row INSERT statement tried to store bad data in a JSON column an error message was emitted, which didn't reference number of row which caused the problem. Such behavior was not consistent with how similar problems were reported for columns of other types. It didn't allow user to understand which record among inserted was problematic.

This patch solves the problem by changing message for ER_INVALID_JSON_TEXT error, which is emitted in such cases, to mention number of row being inserted. It is easy to do so, as the information about row being processed is already available at the point where error is emitted from Diagnostics_area::current_row_for_condition() member function.

It makes behavior for JSON type consistent with how warnings and errors are reported for other types.

### DIFF
--- a/mysql-test/r/percona_heap_json.result
+++ b/mysql-test/r/percona_heap_json.result
@@ -5,7 +5,7 @@
 CREATE TABLE t0(utf0k json) ENGINE=MEMORY;
 INSERT INTO t0 values('0');
 INSERT INTO t0 VALUES('DBMS stands for DataBase ...');
-ERROR 22032: Invalid JSON text: "Invalid value." at position 0 in value for column 't0.utf0k'.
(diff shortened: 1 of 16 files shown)


percona-ysorokin (approved):
LGTM
The only comment is that we need to merge this directly int o the 8.0.32. So, please, change the base branch of this PR to `release-8.0.32-24` before merging.

dlenev:
Jenkins result doesn't show any failures related to this patch https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param-parallel-mtr/45/

percona-ysorokin (approved):
LGTM (after 8.0.32 rebase)

PR #5001, opened 2023-02-20
PS-8643: Fixing ndb compilation & check testcase

dutow:
This commit fixes NDB build errors with editline.

It also improves the check testcase processlist reporting to work with ndb tests. This includes:
* Limiting the shown columns to only display information that doesn't change
* Ignoring the thread requesting the processlist, as its id can change


percona-ysorokin:
@dutow Looks like there is now an issue with `clang-15`

inikep:
`clang-15` probably fails because `INVERTED` sets: 
```
            -DWITH_NDB=ON
            -DWITH_NDBCLUSTER=ON
            -DWITH_NDB_JAVA=OFF
            -DWITH_EDITLINE=bundled
```

percona-ysorokin:
Closed in favour of https://github.com/percona/percona-server/pull/5008

PR #5002, opened 2023-02-20
PS-7923 Assign a thread ID to InnoDB's encryption resume thread

percona-ysorokin:
https://jira.percona.com/browse/PS-7923

Problem:
InnoDB creates a thread to resume encryption but does not assign it an ID. By default threads have ID of 0, which is a magic value that changes what Owned_gtids::is_owned_by() actually checks for.

Fix:
Assign an ID to an encryption thread. Add assertions to Owned_gtids::add_gtid_owner() and Owned_gtids::remove_gtid() to catch similar errors earlier where it's easier to reason about.

Re-recorded 'innodb.tablespace_encrypt_9' MTR test case as new named thread shifted connection IDs.

### DIFF
--- a/mysql-test/t/bug_ps7923.test
+++ b/mysql-test/t/bug_ps7923.test
@@ -1,0 +1,30 @@
+--connect (con1,localhost,root,,)
+
+--connection default
+
+DELIMITER //;
+
+CREATE PROCEDURE my_infinite_alter_ts()
+BEGIN
+  WHILE TRUE DO
+    ALTER TABLESPACE mysql ENCRYPTION='Y';
+    ALTER TABLESPACE mysql ENCRYPTION='N';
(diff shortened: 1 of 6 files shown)


mysql-test/t/bug_ps7923.test
```diff
@@ -0,0 +1,30 @@
+--connect (con1,localhost,root,,)
```
inikep, line 1: It would be good to add:
```
```

inikep (approved):
LGTM

mysql-test/t/bug_ps7923.test
percona-ysorokin: Not needed as we are restarting the server with `--source include/kill_and_restart_mysqld.inc`

PR #5006, opened 2023-02-23
PS-8648: Introduce PS compilation testing with Cirrus CI using EC2 spot instances

inikep:
1. Add the following configurations:
```
(x86_64) gcc-12 Debug INVERTED [Ubuntu 22.04 Jammy]
(x86_64) gcc-12 RelWithDebInfo INVERTED [Ubuntu 22.04 Jammy]
```
2. Use prepared `Ubuntu 22.04.1 x86-64` AMI image with the `percona-server` repo and gcc-12 gcc-11 gcc-10 gcc-9 clang-15 clang-14 clang-13 node-12 installed.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,23 +1,24 @@
 env:  # Global defaults
-  CCACHE_SIZE: "500M"
-  CCACHE_DIR: "${MOUNT_POINT}/ccache_dir"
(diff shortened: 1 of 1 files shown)


percona-ysorokin (approved):
LGTM

PR #5019, opened 2023-03-09
PS-7023 Adopt InnoDB AIO layer to io_uring interface

dlenev:
Add io_uring as a possible option for InnoDB AIO.

Draft patch for further discussion and expirements.


dlenev:
We decided not to proceed with this PR for now as usage of URING has not shown any visible performance benefits in our tests.
See https://jira.percona.com/browse/PS-7023 for details.

PR #5021, opened 2023-03-10
PS-8674 Compilation error on ARM AppleClang 14.0

rahulmalik87:
https://jira.percona.com/browse/PS-8674

Disabled the PERCONA_AUTHENTICATION_LDAP on APPLE ARM because of compilation error

undefined symbols for architecture arm64:
  "_ldap_set_urllist_proc", referenced from:
ld: symbol(s) not found for architecture arm64

The problem with MacOS system libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1830,7 +1830,13 @@ MYSQL_CHECK_LDAP()
 MYSQL_CHECK_LDAP_DLLS()
 
 # Percona LDAP Simple / LDAP SASL authentication plugins
-OPTION(WITH_PERCONA_AUTHENTICATION_LDAP "Build with Percona LDAP Simple / LDAP SASL authentication plugins" ON)
+# Problem with  MacOS system libraries and complains for compilation error ldap_set_urllist_proc
+IF(APPLE)
+  SET(WITH_PERCONA_AUTHENTICATION_LDAP_DEFAULT OFF)
+ELSE()
+  SET(WITH_PERCONA_AUTHENTICATION_LDAP_DEFAULT ON)
+ENDIF()
+OPTION(WITH_PERCONA_AUTHENTICATION_LDAP "Build with Percona LDAP Simple / LDAP SASL authentication plugins" ${WITH_PERCONA_AUTHENTICATION_LDAP_DEFAULT})
 
 # Add Windows specific jemalloc DLL
 IF(WIN32)


CMakeLists.txt
```diff
@@ -1830,7 +1830,12 @@ MYSQL_CHECK_LDAP()
 MYSQL_CHECK_LDAP_DLLS()
 
 # Percona LDAP Simple / LDAP SASL authentication plugins
-OPTION(WITH_PERCONA_AUTHENTICATION_LDAP "Build with Percona LDAP Simple / LDAP SASL authentication plugins" ON)
+# Problem with  MacOS system libraries and complains for compilation error ldap_set_urllist_proc
+IF(APPLE_ARM)
```
dutow: I think this affects all APPLE builds, not just ARM.

rahulmalik87: removed the ARM condition

CMakeLists.txt
```diff
@@ -1830,7 +1830,12 @@ MYSQL_CHECK_LDAP()
 MYSQL_CHECK_LDAP_DLLS()
 
 # Percona LDAP Simple / LDAP SASL authentication plugins
-OPTION(WITH_PERCONA_AUTHENTICATION_LDAP "Build with Percona LDAP Simple / LDAP SASL authentication plugins" ON)
+# Problem with  MacOS system libraries and complains for compilation error ldap_set_urllist_proc
```
percona-ysorokin, line 1833: @rahulmalik87 just a stylistic comment. From what I see from Oracle's CMake files, they usually do this as following
```
IF(APPLE)
  SET(WITH_PERCONA_AUTHENTICATION_LDAP_DEFAULT OFF)
ELSE()
  SET(WITH_PERCONA_AUTHENTICATION_LDAP_DEFAULT ON)
ENDIF()
OPTION(WITH_PERCONA_AUTHENTICATION_LDAP "Build with Percona LDAP Simple / LDAP SASL authentication plugins" ${WITH_PERCONA_AUTHENTICATION_LDAP_DEFAULT})
```

rahulmalik87: fixed

CMakeLists.txt
```diff
@@ -1830,7 +1830,13 @@ MYSQL_CHECK_LDAP()
...
+IF(APPLE)
+  OPTION(WITH_PERCONA_AUTHENTICATION_LDAP OFF)
+ELSE()
+  OPTION(WITH_PERCONA_AUTHENTICATION_LDAP ON)
+ENDIF()
+OPTION(WITH_PERCONA_AUTHENTICATION_LDAP "Build with Percona LDAP Simple / LDAP SASL authentication plugins")
```
percona-ysorokin: @rahulmalik87 To be honest I did not find anywhere in the CMake docs that this option redeclaration (first with value and then with name) is a valid construct.

dutow: It's somewhat documented in backward compatibility pages. Essentially OPTION creates a cached variable, and will not override existing variables.

In this sense, redeclaration is not an error, but the second declaration won't override the already existing docstring. With this change, that means the docstring will be empty.

rahulmalik87: I checked the docstring in CMakeCache.txt. It was not empty.

`//Build with Percona LDAP Simple / LDAP SASL authentication plugins
WITH_PERCONA_AUTHENTICATION_LDAP:BOOL=OFF`

percona-ysorokin: OK, fine with me if it works

dutow: @rahulmalik87 I rechecked that, you are right. But that is not what the documentation says: "If <variable> is already set as a normal or cache variable, then the command does nothing" Not sure if this is a cmake bug or a documentation bug, as changing the text is not nothing.

rahulmalik87: thanks Guys, Fixed. now using WITH_PERCONA_AUTHENTICATION_LDAP_DEFAULT

CMakeLists.txt
```diff
@@ -1830,7 +1830,13 @@ MYSQL_CHECK_LDAP()
 MYSQL_CHECK_LDAP_DLLS()
 
 # Percona LDAP Simple / LDAP SASL authentication plugins
-OPTION(WITH_PERCONA_AUTHENTICATION_LDAP "Build with Percona LDAP Simple / LDAP SASL authentication plugins" ON)
+# Problem with  MacOS system libraries and complains for compilation error ldap_set_urllist_proc
+IF(APPLE)
+	OPTION(WITH_PERCONA_AUTHENTICATION_LDAP_DEFAULT OFF)
```
dutow: This should be a SET command, not an option.

rahulmalik87: fixed

percona-ysorokin (approved):
LGTM

PR #5025, opened 2023-03-22
PS-8683: Mutex deadlock (com_statistics, com_change_user, show full p…

dlenev:
…rocesslist)

Starting from 8.0.28 versions of MySQL and Percona Server concurrent execution of COM_STATISTICS, COM_CHANGE_USER commands and SHOW PROCESSLIST statement sometimes led to deadlock. Same problem was observed if COM_STATISTICS was replaced with FLUSH STATUS statement and/or SHOW PROCESSLIST was replaced with SELECT from I_S.PROCESSLIST table.

The deadlock occured because of regression from fix for bug#32320541 "RACE CONDITION ON SECURITY_CONTEXT::M_USER". After this patch COM_STATISTICS/FLUSH STATUS, COM_CHANGE_USER and SHOW PROCESSLIST/ SELECT ... FROM I_S.PROCESSLIST started to acquire same locks in different order.

In particular:
1) Code responsible for changing user for connection started to acquire
   THD::LOCK_thd_security_ctx mutex and the acquired LOCK_status mutex during
   call to THD::cleanup_connection(), without releasing the former.
2) Implementations of COM_STATISTICS and FLUSH STATUS commands acquire
   LOCK_status mutex and then during iteration through all connections
   LOCK_thd_remove mutexes without releasing the former.
3) Finally, SHOW PROCESSLIST/I_S.PROCESSLIST implementation acquired
   LOCK_thd_remove mutexes and then THD::LOCK_thd_security_ctx mutex
   during copying information about particular connection, without
   releasing the former.

Naturally, THD::LOCK_thd_security_ctx -> LOCK_status -> LOCK_thd_remove -> THD::LOCK_thd_security_ctx dependency loop occasionally resulted in deadlock.

This patch solves the problem by reducing scope during which THD::LOCK_thd_security_ctx lock is held in COM_CHANGE_USER implementation. We no longer call THD::cleanup_connection()/lock LOCK_status while holding it, thus breaking dependency loop.

It might be a good idea to limit scope of this lock even more, but it looks like an issue which is better to solve in upstream.

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -1967,13 +1967,6 @@ bool dispatch_command(THD *thd, const COM_DATA *com_data,
       break;
     }
     case COM_CHANGE_USER: {
-      /*
-        LOCK_thd_security_ctx protects the THD's security-context from
-        inspection by SHOW PROCESSLIST while we're updating it. Nested
-        acquiring of LOCK_thd_data is fine (see below).
-      */
-      MUTEX_LOCK(grd_secctx, &thd->LOCK_thd_security_ctx);
-
       int auth_rc;
       thd->status_var.com_other++;
 
@@ -1981,7 +1974,17 @@ bool dispatch_command(THD *thd, const COM_DATA *com_data,
       USER_CONN *save_user_connect =
           const_cast<USER_CONN *>(thd->get_user_connect());
       LEX_CSTRING save_db = thd->db();
+
+      /*
+        LOCK_thd_security_ctx protects the THD's security-context from
+        inspection by SHOW PROCESSLIST while we're updating it. However,
+        there is no need to protect this context while we're reading it,
+        sinceother threads are not supposed to modify it.
+        Nested acquiring of LOCK_thd_data is fine (see below).
+      */
       Security_context save_security_ctx(*(thd->security_context()));
+
+      MUTEX_LOCK(grd_secctx, &thd->LOCK_thd_security_ctx);
 
       auth_rc = acl_authenticate(thd, COM_CHANGE_USER);
       auth_rc |= mysql_audit_notify(
--- a/mysql-test/r/change_user.result
+++ b/mysql-test/r/change_user.result
@@ -91,3 +91,47 @@ IS_USED_LOCK('bug31418')
 NULL
 FLUSH STATUS;
 Value of com_select did not change
+#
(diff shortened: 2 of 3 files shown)


sql/sql_parse.cc
```diff
@@ -1979,6 +1972,13 @@ bool dispatch_command(THD *thd, const COM_DATA *com_data,
       LEX_CSTRING save_db = thd->db();
       Security_context save_security_ctx(*(thd->security_context()));
 
+      /*
+        LOCK_thd_security_ctx protects the THD's security-context from
+        inspection by SHOW PROCESSLIST while we're updating it. Nested
+        acquiring of LOCK_thd_data is fine (see below).
+      */
+      MUTEX_LOCK(grd_secctx, &thd->LOCK_thd_security_ctx);
```
percona-ysorokin, line 1987: Shouldn't we put
```      Security_context save_security_ctx(*(thd->security_context()));```
behind the lock as well. As far as I understood this is a deep copy.

Comment (dlenev) [reply sql/sql_parse.cc:1987]: No, this is not necessary. In the above line we create a local copy from current THD's security context. THD's security context is not supposed to be updated by other threads. So read only operations like this should not require protection with LOCK_thd_security_ctx lock.

Comment (percona-ysorokin) [reply sql/sql_parse.cc:1987]: Oh, I think I got this, although this seems like a very dangerous zone to me.
MULTIPLE reading threads (`SHOW PROCESSLIST`):
```
reading security context under the lock
```

SINGLE modifying thread (`change user`):
```
reading security context without any lock
then
modifying security context under the lock
```
Not a typical locking pattern, I would say. At least in my head.
Although seems to be absolutely legal in this case.

percona-ysorokin: May be it makes sense to add a comment why it is OK to do `Security_context save_security_ctx(*(thd->security_context()));` without a lock.

percona-ysorokin (approved):
LGTM

sql/sql_parse.cc
dlenev: Addressed by adjusting comment.

dlenev:
Here are Jenkins results for the previous version of patch which differs from the final one only in comments - https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr-refactor/117/.

As far as I can see a few test failures which were observed are not related to this patch.

PR #5027, opened 2023-03-23
PS-8647: Too low tmp_table_size cause queries to fail.

dutow:
Issue: tmp_table_size can be set in bytes, but the underlying allocator in the TEMPTABLE engine uses megabytes, and starts allocation with 1 MiB. While there is logic that converts tables that go over this size limit to on-disk tables, it only works after the table was already created. If the initial 1MiB allocation fails, the table creation fails.

Fix: always allow allocating this first 1MiB block, even if tmp_table_size is smaller.


storage/temptable/include/temptable/allocator.h
```diff
@@ -35,6 +35,8 @@ TempTable custom allocator. */
 
 #include "my_dbug.h"
 #include "my_sys.h"
+#include "mysql/components/services/log_builtins.h"
```
percona-ysorokin: @dutow my concern here is that surprisingly, for some reason, logging infrastructure is not used inside the `temptable` code. And I am hesitating to introduce this dependency. Something tells me that there are hidden reasons for that of which we are unaware.
So, although it was me who asked you to add a new error message, I am leaning towards completely removing it. Sorry for that.
However, even if we decide to leave it, these two headers are pretty heavy and should be moved to the `allocator..cc` file along with the `Prefer_RAM_over_MMAP_policy_obeying_per_table_limit::Source block_source()` method definition.

storage/temptable/include/temptable/allocator.h
```diff
@@ -259,8 +261,15 @@ struct Prefer_RAM_over_MMAP_policy_obeying_per_table_limit {
            table_resource_monitor->threshold());
 
     if (table_resource_monitor->consumption() + block_size >
-        table_resource_monitor->threshold())
+        table_resource_monitor->threshold()) {
+      if (table_resource_monitor->consumption() == 0) {
```
percona-ysorokin, line 263: Here, I agree that there is a logical bug. Having `1M` minimal block allocation contradicts with the fact that `@tmp_table_size` can be set to a lower value.
However, the more important question is why SELECTING from these I_S tables does not automatically switch to on-disk InnoDB temp tables.
According to this
https://dev.mysql.com/doc/refman/8.0/en/internal-temporary-tables.html
this is exactly what should happen
```
When the tmp_table_size limit is reached, MySQL automatically converts the in-memory internal
temporary table to an InnoDB on-disk internal temporary table.
```

dutow: It does - but at this point, we are in the middle of table creation, which is handled differently. Later, the code that actually fills the table handles this situation.

(The table creation code wants to allocate a small amount, 1 or 2 rows if I remember correctly, which for one of the example tables, would be 32 bytes. But because the allocator wants to deal with at least 1MiB buffers, it increases the allocation size to 1MiB)

storage/temptable/include/temptable/allocator.h
dutow: We could go back to my initial suggestion, adding a warning message to the system variable instead.

percona-ysorokin: Yes, let's do this

percona-ysorokin (approved):
LGTM

PR #5028, opened 2023-03-23
Fix crashing mysql due to inadequate cleanup of cached queries that use TempTable

lottaquestions:
Description
============
The TempTable storage engine can cause the MySQL engine to crash
if the same query that created a temporary table is run more than
once. The reason for this crash is that the TempTable handler is
cleaned up after execution of the query, but the "keys in use" flag
for the underlying temporary table remains "on" in the cached results.
When the query is rerun, it uses the cached results from the
previous run, and since the "keys in use" flag is set, logic in the
execute stage of the query attempts to access a pointer to what it
expects is an opened temporary table. But since the TempTable
handler had been cleaned up after the first run, a new TempTable
handler is instantiated for the new run, which does not yet have
a pointer to an opened temporary table. When the query rerun
attempts to access the opened temporary table via a null pointer
the system crashes. The crashes only happen on the second
rerun of the query. Subsequent reruns after the second one do
not crash the engine.

This change fixes this issue by ensuring that the "keys in use" flag
is reset (0) during the TempTable handler clean up, after the first
run of a query. This is done because, if the underlying
temporary table has been cleaned up, then by definition there should
not be any "keys in use" on it.

Fixes bug https://bugs.mysql.com/bug.php?id=110152 reported to
Oracle MySQL.

Testing
========
An MTR test is provided with the PR that forces the creation of 
temporary tables and hence causes a crash without the change, but
works correctly with the change.

### DIFF
--- a/mysql-test/r/temptable_repeat_cached_query.result
+++ b/mysql-test/r/temptable_repeat_cached_query.result
@@ -1,0 +1,27 @@
+DROP DATABASE IF EXISTS db1;
+CREATE DATABASE `db1`;
+USE `db1`;
+CREATE TABLE db1.`tb1` (
+`col1` varchar(8) NOT NULL 
+);
+INSERT INTO db1.`tb1`(col1) VALUES ('20211114');
+CREATE PROCEDURE db1.`stored_proc1`()
+BEGIN
+WITH t_dataset_info as (
+SELECT col1 as yyyymmdd FROM tb1 WHERE col1 = (SELECT max(col1) FROM tb1 )
+)
+SELECT a.yyyymmdd as yyyymmdd
+FROM (
+SELECT a.yyyymmdd
+FROM t_dataset_info a
+WHERE a.yyyymmdd = (SELECT max(a.yyyymmdd) FROM t_dataset_info a WHERE a.yyyymmdd <= '20220131')
+) a;
+END|
+SET optimizer_switch="derived_merge=off";
+call db1.stored_proc1();
+yyyymmdd
+20211114
+call db1.stored_proc1();
+yyyymmdd
+20211114
+DROP DATABASE db1;
--- a/mysql-test/t/temptable_repeat_cached_query.test
+++ b/mysql-test/t/temptable_repeat_cached_query.test
@@ -1,0 +1,45 @@
+########################################################################################
+# Check that repeat queries that are cached do not cause MySQL to crash due to
+# usable keys not being reset after completion of processing for each query for
+# the TempTable storage engine.
+########################################################################################
+
+# Disable warnings as they add noise that is not relevant to verifying the
+# behavior that is being tested.
+--disable_warnings
+
+DROP DATABASE IF EXISTS db1;
+CREATE DATABASE `db1`;
+USE `db1`;
+
+CREATE TABLE db1.`tb1` (
+  `col1` varchar(8) NOT NULL 
+);
+
+INSERT INTO db1.`tb1`(col1) VALUES ('20211114');
+
+
+--delimiter |
+
+CREATE PROCEDURE db1.`stored_proc1`()
+BEGIN
+
+    WITH t_dataset_info as (
+        SELECT col1 as yyyymmdd FROM tb1 WHERE col1 = (SELECT max(col1) FROM tb1 )
+    )
+    SELECT a.yyyymmdd as yyyymmdd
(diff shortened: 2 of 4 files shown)


haidong:
Hi Percona fork maintainers,

It looks all CI checks are green. Do you have further questions? If not, could this be merged? By the way, the licensing portion is inside the commit message.

percona-ysorokin:
@lottaquestions @haidong 
Although this looks like a totally valid fix to me, there are some risks that it may introduce regressions in other scenarios and it would be better if this issue was fixed by Oracle.
So, before we start deeper analysis I would like to know if Oracle made any promises regarding fixing this in one of the upstream's upcoming releases.
Unfortunately, as this is a security issue, Oracle made the bug private and only the person who submitted it can still have access to the comments.
So, I would really appreciate it if you could update me on the status of this bug.
https://bugs.mysql.com/bug.php?id=110152

lottaquestions:
> @lottaquestions @haidong Although this looks like a totally valid fix to me, there are some risks that it may introduce regressions in other scenarios and it would be better if this issue was fixed by Oracle. So, before we start deeper analysis I would like to know if Oracle made any promises regarding fixing this in one of the upstream's upcoming releases. **Unfortunately, as this is a security issue, Oracle made the bug private and only the person who submitted it can still have access to the comments. So, I would really appreciate it if you could update me on the status of this bug. https://bugs.mysql.com/bug.php?id=110152**

Unfortunately Oracle have not commented on the bug report, as per the original reporter.

dlenev:
Hello!

Thank you for submitting this PR!

After investigating this issue a bit, I think that this patch tries to solve the problem on the wrong level.

1) The problem seems to be on Optimizer/Executor level and not in the TempTable storage engine. Execution of queries in this case tries to read from temporary table which was not properly open in this scenario. And this won't work for any SE, not only TempTable. Not surprisingly we get similar crash if we force the query in question to use Memory SE for temporary tables instead of TempTable SE by doing "SET internal_tmp_mem_storage_engine=MEMORY;" before running test case.

2) Also note that no other storage engine touches TABLE::keys_in_use_for_query member, so making TempTable special in this regard sounds like bad idea to me (MyISAM is exception in this regard - it changes TABLE::keys_in_use_for_query in a couple of places, but only during execution of special operations related to Keycache and pre-loading of indexes. So it should not serve as an example).

Taking into account the above I think that the problem should be solved on Optimizer/Executor level and not in SE.

vadimtk:
@lottaquestions I spoke with Oracle Community team and they will be happy to review your Pull Request to fix this bug once you submit it to them.

lottaquestions:
> Hello!
> 
> Thank you for submitting this PR!
> 
> After investigating this issue a bit, I think that this patch tries to solve the problem on the wrong level.
> 
>     1. The problem seems to be on Optimizer/Executor level and not in the TempTable storage engine. Execution of queries in this case tries to read from temporary table which was not properly open in this scenario. And this won't work for any SE, not only TempTable. Not surprisingly we get similar crash if we force the query in question to use Memory SE for temporary tables instead of TempTable SE by doing "SET internal_tmp_mem_storage_engine=MEMORY;" before running test case.
> 
>     2. Also note that no other storage engine touches TABLE::keys_in_use_for_query member, so making TempTable special in this regard sounds like bad idea to me (MyISAM is exception in this regard - it changes TABLE::keys_in_use_for_query in a couple of places, but only during execution of special operations related to Keycache and pre-loading of indexes. So it should not serve as an example).
> 
> 
> Taking into account the above I think that the problem should be solved on Optimizer/Executor level and not in SE.

Thanks for the feedback. I am working on an Optimizer/Executor level solution.

lottaquestions:
> @lottaquestions I spoke with Oracle Community team and they will be happy to review your Pull Request to fix this bug once you submit it to them.

We already submitted a PR to Oracle and they closed it: https://github.com/mysql/mysql-server/pull/449

ihustle-hard (approved):

ihustle-hard (approved):

ihustle-hard (approved):

percona-ysorokin:
Closing this PR as the problem was fixed on the server level by Oracle in MySQL Server 8.0.35.
Bug #35105404 "Server crashes with signal 11 and asserts in temptable code" 
(commit mysql/mysql-server@a1d235e)

PR #5032, opened 2023-03-29
PS-8188 : Make the clone_plugin to not force plugins to match on recipient/donor

dutow:
https://jira.percona.com/browse/PS-8188
    
    Problem:
    --------
    Clone enforces donor and recipient have same plugins installed.
    For example, if a donor has an extra authentication plugin or a data masking plugin
    installed but not installed on recipient, it is not possible to clone.
    It will fail with the below error.
    
    'Clone Donor plugin <plugin_name> is not active in Recipient.'
    
    Fix:
    ----
    Introduce new clone option 'clone_exclude_plugins_list' to take list of plugins that
    are not installed on recipient.
    
    This option is used on recipient server. The following plugins are disallowed. They must
    be installed on recipient.
    
    "daemon_keyring_proxy_plugin",
    "binlog",
    "performance_schema",
    "memory",
    "innodb",
    "keyring_file",
    "keyring_vault",
    "group_replication"

### DIFF
--- a/plugin/clone/src/clone_plugin.cc
+++ b/plugin/clone/src/clone_plugin.cc
@@ -33,6 +33,8 @@ Clone Plugin: Plugin interface
 #include "plugin/clone/include/clone_local.h"
 #include "plugin/clone/include/clone_server.h"
 #include "plugin/clone/include/clone_status.h"
+#include "include/sql_string.h"
+#include "sql/sql_error.h"
 
 #include <algorithm>
 #include <cctype>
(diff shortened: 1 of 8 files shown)


plugin/clone/src/clone_plugin.cc
```diff
@@ -85,6 +87,20 @@ uint clone_restart_timeout;
...
+
+/** List of plugins that cannot be excluded by clone_exclude_plugins_list */
+static std::vector<std::string> disallow_list{"daemon_keyring_proxy_plugin",
+                                              "binlog",
+                                              "performance_schema",
+                                              "memory",
+                                              "innodb",
+                                              "keyring_file",
+                                              "keyring_vault",
+                                              "group_replication"};
```
satya-bodapati: I think Simon wants group_replication plugin to NOT be in this disallow_list. 
"
I use native cloning for cloning from non-GR servers (and this works fine) so the requirement of the group_replication plugin to be installed seems to be unnecessary. "

Can we remove the group_replication from this list?

satya-bodapati (changes requested):

satya-bodapati:
@dutow Can u please get Jenkins run?

satya-bodapati:
@dutow LGTM! Please ensure clean Jenkins before merge.

satya-bodapati (approved):

PR #5033, opened 2023-03-30
PXC-4165 FIXES UNSTABLE GROUP REPLICATION TESTS

VarunNagaraju:
https://jira.percona.com/browse/PXC-4165

Some group replication tests were failing because of debug wait sync time out warnings(Result content mismatch) and actual timeout where waiting thread didn't receive the signal to wake up and continue the test. It could be because of the immediate RESET being invoked right after the signal being emitted. So, the DEBUG_SYNC=RESET is moved to the cleanup stage of the test which fixed the issue.

### DIFF
--- a/mysql-test/include/gr_parallel_local_and_remote_transactions.inc
+++ b/mysql-test/include/gr_parallel_local_and_remote_transactions.inc
@@ -136,6 +136,7 @@ BEGIN;
 --echo # group_replication_before_message_broadcast debug sync point.
 --let $rpl_connection_name= $local_server_connection2
 --source include/rpl_connection.inc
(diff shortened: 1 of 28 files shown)


venkatesh-prasad-v (approved):

VarunNagaraju:
Jenkins run(Jammy): https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr-refactor/124/
- Re-recorded the big tests.
- Other tests failed because of Xplugin being turned off. Verfied them locally.

kamil-holubicki (approved):

VarunNagaraju:
Upstream was able to reproduce [Bug #110615](https://bugs.mysql.com/bug.php?id=110615). But, it hasn't been fixed in 8.0.33.

PR #5034, opened 2023-03-31
PS-8704: Fix heap-buffer-overflow in mysql_query_attributes iterator

oleksandr-kachan:
https://jira.percona.com/browse/PS-8704

There was an attempt to access one past last element while going through query attributes list.

### DIFF
--- a/sql/server_component/mysql_query_attributes_imp.cc
+++ b/sql/server_component/mysql_query_attributes_imp.cc
@@ -98,12 +98,14 @@ class iterator {
   }
 
   bool next() {
-    while (ofs < thd->bind_parameter_values_count) {
+    assert(thd->bind_parameter_values_count > 0);
+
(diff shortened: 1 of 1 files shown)


sql/server_component/mysql_query_attributes_imp.cc
```diff
@@ -98,12 +98,12 @@ class iterator {
   }
 
   bool next() {
-    while (ofs < thd->bind_parameter_values_count) {
+    while (ofs < thd->bind_parameter_values_count - 1) {
```
percona-ysorokin, line 103: May be add an assertion `thd->bind_parameter_values_count > 0`?

oleksandr-kachan: Done

percona-ysorokin (approved):
LGTM

PR #5038, opened 2023-04-21
PS-8719: Fix audit_log stall on log flush

oleksandr-kachan:
https://jira.percona.com/browse/PS-8719

The audit_log write buffer may be in one of two states - LOG_RECORD_COMPLETE or LOG_RECORD_INCOMPLETE.
The write buffer is a circular one. It is possible that record gets split so that one its part is written in the end of the buffer and onother part is put at the actual buffer beginning. Once buffer flush is involved it processes such records in two steps. After the first record part is flushed, buffer is put into LOG_RECORD_INCOMPLETE state. In this state log file cannot be rotated. The plugin waits for the buffer to flush second part of the record and switch itself into LOG_RECORD_COMPLETE state.

In some situations it is possible that the full record fits the buffer free space and buffer write_pos is moved to the very buffer end. In cases like this one log flush operation treats the buffer being in LOG_RECORD_INCOMPLETE state despite the fact that there is actually no split record in the buffer. If "SET GLOBAL audit_log_flush=ON" is invoked at this time it tries to pause the buffer but as it is in LOG_RECORD_INCOMPLETE state and has no way to switch to LOG_RECORD_COMPLETE state any more the thread running audit_log_flush operation gets stuck.

To fix this, the way how buffer flushing thread determines new buffer state is changed. It checks buffer write_pos to make sure if there is really split record in the buffer or it just reached actuall buffer end.


oleksandr-kachan:
https://ps57.cd.percona.com/job/percona-server-5.7-param/594/

percona-ysorokin (approved):
LGTM

PR #5040, opened 2023-04-27
PS-8647 post push fix: update assertion

dutow:
It has to take into account that now we might allocate above the threshold up to 1MiB.


inikep (approved):
LGTM

PR #5044, opened 2023-05-11
PS-8723 merge: Merge MySQL 8.0.33 - Q2 2023 (ext::zlib fix)

percona-ysorokin:
https://jira.percona.com/browse/PS-8723

A number of source files either directly or indirectly (via, say, 'sql_class.h') include 'my_checksum.h' header file which in turn includes 'zlib.h' and has one of the inline functions referring to 'crc32_z()' function. The issue araises on old platforms (say, CentOS7) that have pretty old system 'zlib' library in which this function just does not exist. In other words, on such platforms, we should always include 'zlib.h' from the 'bundled' location.
However, because of the refactoring Oracle did in the CMake files in the fix for the Bug #35057542
"Create INTERFACE libraries for bundled/system zlib/zstd/lz4" (mysql/mysql-server@1f2b9d6), bundled 'zlib' include directory is no longer set globally in 'MY_INCLUDE_SYSTEM_DIRECTORIES'. Instead, individual targets are supposed to add 'ext::zlib' to the list of dependencies via 'TARGET_LINK_LIBRARIES()'.

This commit adds missing 'ext::zlib' dependencies to Percona-specific targets.

### DIFF
--- a/components/encryption_udf/CMakeLists.txt
+++ b/components/encryption_udf/CMakeLists.txt
@@ -33,7 +33,7 @@ MYSQL_ADD_COMPONENT(encryption_udf
   encryption_udf_component.cc
   server_helpers.h
   server_helpers.cc
-  LINK_LIBRARIES OpenSSLPP::OpenSSLPP
+  LINK_LIBRARIES OpenSSLPP::OpenSSLPP ext::zlib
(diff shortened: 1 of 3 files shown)


inikep (approved):
LGTM

PR #5053, opened 2023-05-17
PS-3942: Add audit_log MTR test to make sure PING_CMD has correct command_class (8.0)

oleksandr-kachan:
https://jira.percona.com/browse/PS-3942


percona-ysorokin (approved):
LGTM

PR #5062, opened 2023-06-05
PS-8728: Implement data masking component

dutow:
Differences compared to upstream docs:

* Upstream documentation has some contradictions / unclear sentences about character set conversions and optimizations. Sometimes it says that it converts things to utf8mb4, sometimes it claims that it return the same character set as the parameter. Sometimes it mentions both for the same function, which is a contradiction. In our implementation:
  * the return value always uses the same character set as the first string argument, if there's any string arguments
  * if there's no string argument, we use utf8mb4
  * we never do unneccessary internal conversions, we process the string in that character set as specified, and copy data in big chucks when possible
  * if the function uses multiple strings, and they are specified in different character sets, we use the character set of the first parameter, and convert the rest to it
  * masking characters CAN be multibyte characters (it is unclear if upstream allows this or not)
* Upstream claims that specific masking functions (PAN, SSN, etc) only support specific format. They also contradict this by showing examples not in those formats. In our implementation, we only do length checks in these functions: if the argument is the correct length, we do the masking as specified, if it's not, we report an error
* In the case of mask_uk_nin, upstream documentation shows an example that's not the specified length, and yet, they claim that it works. It doesn't in our implementation.
* There's no mention about the privilege requirements for the gen_blacklist and gen_dictionary UDFs, other than one sidecomment that all users can execute gen_dictionary. I'm not sure if that's secure, but our implementation currently does exactly that, and allows access to everyone. To accomplish this, internal queries are ran as 'root'@'localhost', as the documentation doesn't mention adding additional users, only the table, or making the table world readable.
* Upstream doc says that masking_dictionary_term_add should return error if a dictionary doesn't exists... but there is no other way to create a new dictionary. Also, the table structure doesn't allow to create a dictionary without an inital term. Our implementation succeeds, and creates the dictionary.
* gen_rnd_pan is kept as in the plugin: it has no optional number parameter, and generates a random, but valid mastercard/visa/discover number instead of something totally random


components/masking_functions/data_masking.ini
```diff
@@ -0,0 +1,32 @@
...
+
+#
+# Plugin configuration file. Place the following on a separate line:
+#
+# library binary file name (without .so or .dll)
+# component_name
+# [component_name] - additional components in plugin
+#
+
+data_masking
```
percona-ysorokin: The same line `data_masking` twice.
Is this intentional?

components/masking_functions/include/udf/udf_utils.h
```diff
@@ -0,0 +1,64 @@
...
+namespace plugins {
+std::string random_string(unsigned long length, bool letter_start);
+
+std::string random_number(const unsigned int length);
+
+long random_number(const long min, const long max);
+
+std::string random_credit_card();
+
+std::string random_iban(std::string country, int length);
```
percona-ysorokin: const std::string&

components/masking_functions/include/udf/udf_utils.h
```diff
@@ -0,0 +1,64 @@
...
+#include <mysql/components/services/udf_metadata.h>
+#include <mysql/udf_registration_types.h>
+#include <random>
+#include <string>
+
+extern REQUIRES_SERVICE_PLACEHOLDER(mysql_udf_metadata);
+
+namespace mysql {
+namespace plugins {
+std::string random_string(unsigned long length, bool letter_start);
```
percona-ysorokin: Here and everywhere
Consider using `std::size_t` for size-like / position-like parameters

components/masking_functions/include/udf/udf_utils.h
percona-ysorokin: signed type for length

components/masking_functions/include/udf/udf_utils.h
```diff
@@ -0,0 +1,64 @@
...
+std::string random_uuid();
+
+std::string random_uk_nin();
+
+std::string random_us_phone();
+
+bool get_arg_character_set(UDF_ARGS *args, size_t index, std::string &charset);
+
+// 0 = variable
+std::size_t get_character_set_width(std::string const &charset);
```
percona-ysorokin: Here and everywhere
Consider `std::string_view` for input string parameters

components/masking_functions/src/component.cc
```diff
@@ -0,0 +1,140 @@
...
+  DBUG_TRACE;
+
+  // reg_srv = mysql_plugin_registry_acquire();
+
+  // if (mysql::plugins::Charset_service::init(reg_srv)) return 1;
+
+  sql_print_information(
+      "Masking Functions Component: Initializing component");
+
+  mysql_service_dynamic_privilege_register->register_privilege(
```
percona-ysorokin: Need to validate return value here and early return from the `component_init()`

components/masking_functions/src/component.cc
```diff
@@ -0,0 +1,140 @@
...
+      privilege_name.data(), privilege_name.size());
+  register_udfs();
```
percona-ysorokin: Need to validate return value of every single function registration and early return from the `component_init()`

components/masking_functions/src/component.cc
```diff
@@ -0,0 +1,140 @@
...
+
+  return 0;
+}
+
+static mysql_service_status_t component_deinit() {
+  DBUG_TRACE;
+
+  sql_print_information("Masking Function Component: Deinitializing component");
+
+  unregister_udfs();
```
percona-ysorokin: Likewise, return code validation

components/masking_functions/src/component.cc
```diff
@@ -0,0 +1,140 @@
...
+
+  mysql_service_dynamic_privilege_register->register_privilege(
```
percona-ysorokin: Likewise, return code validation

components/masking_functions/src/udf/udf_gen_blocklist.cc
```diff
@@ -0,0 +1,124 @@
...
+}
+
+static char *gen_blocklist(UDF_INIT *initid, UDF_ARGS *args, char *,
+                           unsigned long *length, char *is_null, char *) {
+  DBUG_TRACE;
+
+  std::string original_charset;
+  mysql::plugins::get_arg_character_set(args, 0, original_charset);
+  const std::string term = mysql::plugins::convert(
+      {args->args[0], args->lengths[0]}, original_charset, "utf8mb4");
```
percona-ysorokin: Magic value "utf8mb4"
It had better be extracted to a separate constant

components/masking_functions/src/udf/udf_gen_blocklist.cc
```diff
@@ -0,0 +1,124 @@
...
+  const std::string dict_b = mysql::plugins::convert(
+      {args->args[2], args->lengths[2]}, original_charset, "utf8mb4");
+
+  {
+    mysql::components::sql_context sqlc;
+
+    std::string query(
+        "SELECT Term FROM mysql.masking_dictionaries WHERE Dictionary =\"");
+    mysql::components::escape_string_into(query, dict_a);
+    query += "\" AND Term = \"";
```
percona-ysorokin: A dangerous zone for SQL-injections.
Can we use prepared statements here?

components/masking_functions/src/udf/udf_gen_blocklist.cc
```diff
@@ -0,0 +1,124 @@
...
+  if (mysql::plugins::set_return_value_charset_to_match_arg(initid, args, 0)) {
+    std::snprintf(message, MYSQL_ERRMSG_SIZE,
+                  "Unable to set character set service for UDF");
+    return true;
+  }
+
+  initid->maybe_null = 1;
+  initid->const_item =
+      0;  // Non-Deterministic: same arguments will produce different values
+  initid->ptr = NULL;
```
percona-ysorokin: here and everywhere
nullptr instead of NULL

components/masking_functions/src/udf/udf_gen_blocklist.cc
```diff
@@ -0,0 +1,124 @@
...
+    mysql::components::escape_string_into(query, dict_a);
+    query += "\" AND Term = \"";
+    mysql::components::escape_string_into(query, term);
+    query += "\" ORDER BY RAND() LIMIT 1";
+
+    auto sresult = sqlc.query_single_value(query);
+
+    if (!sresult) {
+      *length = term.size();
+      initid->ptr = new char[term.size() + 1];
```
percona-ysorokin: Any UDF can be called in a multi-row contexts. For instance,
SELECT my_udf() FROM SEQUENCE_TABLE(<N>) AS tt;
In this case, the sequence of internal implementation function calls will be the following:
my_udf_init()
my_udf(); // 1
my_udf(); // 2
...
my_udf(); // <N>
my_udf_deinit();

And here you are allocation memory for initid->ptr  inside my_ufd() (main function), which would cause memory leaks as you free initid->ptr only in the my_udf_deinit() function

percona-ysorokin: The typical usage is the following:
initid->ptr is allocated inside my_udf_init()
initid->ptr is used several times in my_udf() (the size of the allocated buffer may be increased increased if necessary)
initid->ptr is deallocated inside my_udf_deinit()

percona-ysorokin: Alternatively, you may introduce a context struct
struct context {
  std::string buffer;
};
and put an instance of new context{} into initid->ptr

percona-ysorokin: Also, consider using udf_wrappers headers
https://github.com/percona/percona-server/tree/8.0/include/mysqlpp
they were designed exactly to eliminate any manual memory management and reporting SQL errors by throwing exception.
See 'encryption_udf' and 'binlog_udf' for usage examples.

components/masking_functions/include/udf/udf_utils.h
```diff
@@ -0,0 +1,64 @@
...
+
+extern REQUIRES_SERVICE_PLACEHOLDER(mysql_udf_metadata);
+
+namespace mysql {
+namespace plugins {
+std::string random_string(unsigned long length, bool letter_start);
+
+std::string random_number(const unsigned int length);
+
+long random_number(const long min, const long max);
```
percona-ysorokin: If these min and max are taken from the UDF arguments, then their type must be long long

components/masking_functions/src/udf/udf_gen_rnd_iban.cc
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  if (args->arg_count >= 1 && args->lengths[1] != 2) {
+    std::snprintf(message, MYSQL_ERRMSG_SIZE,
+                  "Wrong argument: country code must be two characters");
+    return true;
+  }
+
+  if (args->arg_count == 2) {
+    int len = *(int *)args->args[1];
```
percona-ysorokin: Integer values should be extracted by 'long log *'.
current code won't work on big-endian architectures like ARM.

components/masking_functions/src/udf/udf_gen_rnd_iban.cc
```diff
@@ -0,0 +1,106 @@
...
+    if (len < 15 || len > 34) {
```
percona-ysorokin: magic numbers

components/masking_functions/src/udf/udf_gen_rnd_iban.cc
```diff
@@ -0,0 +1,106 @@
...
+                          char *result [[maybe_unused]], unsigned long *length,
+                          char *, char *) {
+  DBUG_TRACE;
+
+  std::string country("ZZ");
+  if (args->arg_count >= 1) {
+    country.assign(static_cast<const char *>(args->args[0]), args->lengths[0]);
+  }
+
+  unsigned int len = 16;
```
percona-ysorokin: magic number

components/masking_functions/src/udf/udf_registration.cc
```diff
@@ -0,0 +1,119 @@
...
+#include "components/masking_functions/include/udf/udf_mask_uuid.h"
+#include "components/masking_functions/include/udf/udf_masking_dictionary_remove.h"
+#include "components/masking_functions/include/udf/udf_masking_dictionary_term_add.h"
+#include "components/masking_functions/include/udf/udf_masking_dictionary_term_remove.h"
+#include "components/masking_functions/include/udf/udf_registration.h"
+
+extern REQUIRES_SERVICE_PLACEHOLDER(udf_registration);
+
+/* The UDFs we will register. */
+static std::array<udf_descriptor, 21> udfs = {{
```
percona-ysorokin: In c++17 you can rely on Class template argument deduction (CTAD)
basically
static std::array udfs = {...};
should also work

components/masking_functions/src/udf/udf_registration.cc
```diff
@@ -0,0 +1,119 @@
...
+  return error;
+}
+
+bool unregister_udfs() {
+  bool error = false;
+
+  int was_present;
+  for (udf_descriptor const &udf : udfs) {
+    // Don't care about the functions not being there.
+    error = error || mysql_service_udf_registration->udf_unregister(
```
percona-ysorokin: Sometimes, some UDFs may be still in use while you try to uninstall the component, so its better to unregister as many functions as possible here instead of stopping at the first failed.
Typical usage scenario:
UNINSTALL COMPONENT ...; // fails
// we wait a few sect and try again until succeed
UNINSTALL COMPONENT ...; // succeeds

Moreover, to get meaningful result from the subsequent UNINSTALL COMPONENT calls, you need to have a global state of currently registered UDFs from the list, so that on next attempt you would only try to uninstall those, that has not been uninstalled yet.

components/masking_functions/src/udf/udf_utils.cc
```diff
@@ -0,0 +1,217 @@
...
+#include <cassert>
+
+namespace mysql {
+namespace plugins {
+std::string random_string(unsigned long length, bool letter_start) {
+  auto randchar_a = []() -> char {
+    const std::string charset(
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        "abcdefghijklmnopqrstuvwxyz");
+    std::random_device r;
```
percona-ysorokin: Recreating std::random_device, std::default_random_engine and std::uniform_int_distribution for just a single character is really-really slow.
Those need to be extracted outside of lambdas - as a local variables of the random_string().
Even better, make just a single instance of them but protect with a mutex

components/masking_functions/src/udf/udf_utils_sql.cc
```diff
@@ -0,0 +1,157 @@
...
+  if (mysql_service_mysql_current_thread_reader->get(&thd)) {
+    throw sql_context_exception("Couldn't query current thd");
+  }
+
+  Security_context_handle sctx;
+  if (thd_security_ctx->get(thd, &sctx)) {
+    throw sql_context_exception("Couldn't query security context");
+  }
+
+  if (cmd_options_srv->set(mysql_h, MYSQL_COMMAND_USER_NAME, "root")) {
```
percona-ysorokin, line 57: Shouldn't we use current user credentials here?

components/masking_functions/src/udf/udf_utils_string.cc
```diff
@@ -0,0 +1,207 @@
...
+
+#include <mysql/components/services/mysql_string.h>
+extern REQUIRES_SERVICE_PLACEHOLDER(mysql_string_converter);
+extern REQUIRES_SERVICE_PLACEHOLDER(mysql_string_character_access);
+extern REQUIRES_SERVICE_PLACEHOLDER(mysql_string_factory);
+
+namespace mysql {
+namespace plugins {
+std::string &ltrim(std::string &s) {
+  s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](unsigned char c) {
```
percona-ysorokin, line 38: there is boost/algorithm/string/trim.hpp
https://www.boost.org/doc/libs/1_82_0/boost/algorithm/string/trim.hpp

components/masking_functions/src/udf/udf_utils_string.cc
```diff
@@ -0,0 +1,207 @@
...
+std::string &rtrim(std::string &s) {
+  s.erase(std::find_if(s.rbegin(), s.rend(),
+                       [](unsigned char c) { return !std::isspace(c); })
+              .base(),
+          s.end());
+  return s;
+}
+
+std::string &tolower(std::string &s) {
+  std::transform(s.begin(), s.end(), s.begin(), ::tolower);
```
percona-ysorokin, line 200: boost/algorithm/string/case_conv.hpp
https://www.boost.org/doc/libs/1_82_0/boost/algorithm/string/case_conv.hpp

mysql-test/suite/component_masking_functions/t/gen_dictionary.test
```diff
@@ -0,0 +1,68 @@
...
+    INDEX dictionary_idx (Dictionary)
+) ENGINE = InnoDB DEFAULT CHARSET=utf8mb4;
+
+INSERT INTO mysql.masking_dictionaries VALUES("us_cities", "city1");
+INSERT INTO mysql.masking_dictionaries VALUES("us_cities", "city2");
+INSERT INTO mysql.masking_dictionaries VALUES("us_cities", "city3");
+INSERT INTO mysql.masking_dictionaries VALUES("us_cities", "city4");
+
+CREATE USER udftest;
+--connect(con1,localhost,udftest,,)
```
percona-ysorokin, line 23: count_sessions.inc for tests using additional connections

components/masking_functions/src/udf/udf_utils_sql.cc
dutow: No, the documentation states that all users have to be able to use the read only functions, and that the modifications functions only depend on the dynamic privilege created by the component, regardless of access permissions to the internal table.

components/masking_functions/src/udf/udf_gen_blocklist.cc
dutow: No, the API doesn't support it, but there is a string escape function I've used, that prevents injection.

components/masking_functions/data_masking.ini
dutow: No, this file is a leftover from the plugin, removed.

components/masking_functions/src/udf/udf_gen_blocklist.cc
dutow: I reworked it using udf_wrappers. I also had to make a few modifications to the wrappers, I've done that in separate commits.

components/masking_functions/CMakeLists.txt
```diff
@@ -46,6 +46,12 @@ SET(DATAMASKING_SOURCES
 
 DISABLE_MISSING_PROFILE_WARNING()
 
+INCLUDE_DIRECTORIES(
```
percona-ysorokin: Instead of polluting global include list, use modern target-based approach
```
target_include_directories(component_masking_functions SYSTEM PRIVATE ${CMAKE_CURRENT_SOURCE_DIR} ${BOOST_PATCHES_DIR} ${BOOST_INCLUDE_DIR})
```

components/masking_functions/CMakeLists.txt
```diff
@@ -46,6 +46,12 @@ SET(DATAMASKING_SOURCES
...
+  ${CMAKE_CURRENT_SOURCE_DIR}
+  ${BOOST_PATCHES_DIR}
+  ${BOOST_INCLUDE_DIR}
+)
+
 if(NOT DEFINED CMAKE_CXX_VISIBILITY_PRESET AND NOT DEFINED CMAKE_VISIBILITY_INLINES_HIDDEN)
```
percona-ysorokin, line 33: The same for global properties
```
set_target_properties(component_masking_functions PROPERTIES
  CXX_VISIBILITY_PRESET hidden
  VISIBILITY_INLINES_HIDDEN YES
)

Comment (percona-ysorokin) [reply components/masking_functions/CMakeLists.txt:33]: And, "yes" this should also be improved in the `encryption_udfs` and `opensslpp` :)

File: components/masking_functions/include/udf/udf_utils.h [hunk 26]
```diff
@@ -25,19 +25,23 @@
...
 
-long random_number(const long min, const long max);
+std::string random_number(std::size_t length);
+
+std::size_t random_number(std::size_t min, std::size_t max);
 
 std::string random_credit_card();
 
-std::string random_iban(std::string country, int length);
+std::string random_iban(std::string_view const& country, std::size_t length);
```
percona-ysorokin: Not a big issue, but std::string_view is recommended to be passed by value.
It has just 2 members: a pointer and a std::size_t.
When such small structs are passed by value in c++17 the ABI allows to split them and pass the value in 2 registers.
https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/

components/masking_functions/include/udf/udf_utils.h
```diff
@@ -47,16 +51,13 @@ std::string random_uk_nin();
...
 
-bool get_arg_character_set(UDF_ARGS *args, size_t index, std::string &charset);
-
-// 0 = variable
-std::size_t get_character_set_width(std::string const &charset);
+bool get_arg_character_set(UDF_ARGS *args, std::size_t index, std::string& charset);
 
 bool set_return_value_charset(UDF_INIT *initid,
-                              const std::string &charset = "utf8mb4");
+                              std::string_view const& charset = "utf8mb4");
```
percona-ysorokin: Probably, `= default_charset `

include/mysqlpp/udf_context.hpp
```diff
@@ -102,6 +102,31 @@ class udf_context {
...
+    if (mysql_service_mysql_udf_metadata->result_set(initid_, "charset", cs))
+      throw std::runtime_error{"Couldn't set character set"};
+  }
+
+  std::string_view get_arg_charset(std::size_t index) const {
+    void *output = nullptr;
+    if (args_->arg_type[index] != STRING_RESULT)
+      throw std::runtime_error{"Argument is not a string"};
+
+    if (mysql_service_mysql_udf_metadata->argument_get(args_, "charset", index,
```
percona-ysorokin, line 120: Does this function allocate memory for the "output" parameter?

include/mysqlpp/udf_context.hpp
```diff
@@ -102,6 +102,31 @@ class udf_context {
     initid_->decimals = DECIMAL_NOT_SPECIFIED;
   }
 
+#ifdef MYSQLPP_CHARSET_SUPPORT
+  void set_return_value_charset(std::string_view const &charset) {
```
percona-ysorokin: There is a potential problem here with "charset" (passed as a string_view) may not contain terminating NUL character. And its value is later used in the c-style interface ("result_set(()") expecting a NUL-terminated string.

include/mysqlpp/udf_context.hpp
```diff
@@ -102,6 +102,31 @@ class udf_context {
...
+    void *cs = const_cast<char *>(charset.data());
+    if (mysql_service_mysql_udf_metadata->result_set(initid_, "charset", cs))
```
percona-ysorokin, line 111: BTW, just an idea how to get rid of mysql_service_mysql_udf_metadata dependency - we can just pass it as an input parameter. I am pretty sure mysql_service_mysql_udf_metadata will be available in places where this set_return_value_charset() is used.

components/masking_functions/src/udf/udf_utils_string.cc
```diff
@@ -41,8 +41,8 @@ std::string &ltrim(std::string &s) {
   return s;
 }
 
-std::string convert(std::string_view const &src, std::string const &src_cs,
-                    std::string const &dst_cs) {
+std::string convert(std::string_view const &src, std::string_view const &src_cs,
```
percona-ysorokin: src_cs may not be null terminated as it is passed by string_view - it is not safe to use .data() in c-style APIs

components/masking_functions/src/udf/udf_utils_string.cc
```diff
@@ -88,10 +88,10 @@ std::string convert(std::string_view const &src, std::string const &src_cs,
  * of the margin values is larger than the argument length, no masking occurs
  * and the argument is returned unchanged.
  */
-std::string mask_inner(const char *str, const long str_length,
-                       const int margin1, const int margin2,
-                       std::string const &original_charset,
-                       std::string const &mask_char) {
+std::string mask_inner(const char *str, std::size_t str_length,
+                       std::size_t margin1, std::size_t margin2,
+                       std::string_view const &original_charset,
```
percona-ysorokin: Likewise, original_charset may not be null terminated as it is passed by string_view - it is not safe to use .data() in c-style APIs

include/mysqlpp/udf_wrappers.hpp
```diff
@@ -215,6 +222,53 @@ class generic_udf<ImplType, INT_RESULT>
   }
 };
 
+struct udf_info {
```
percona-ysorokin: I really like the idea of extracting registration / unregistration logic into a reusable file.
The only thing I would  like to change is to put this new code into a separate file, say 'mysqlpp/udf_registration.hpp' so that we won't make every UDF wrapper user depend on MySQL services. This new functionality will still be available but only when user explicitly requests it (includes this 'mysqlpp/udf_registration.hpp' file).

components/masking_functions/src/udf/udf_registration.cc
```diff
@@ -0,0 +1,900 @@
...
+      ctx.set_arg_type(0, INT_RESULT);
+    }
+
+    if (ctx.get_number_of_args() >= 2) {
+      ctx.mark_arg_nullable(1, false);
+      ctx.set_arg_type(1, STRING_RESULT);
+    }
+
+    if (ctx.get_number_of_args() >= 2) {
+      ctx.set_return_value_charset(ctx.get_arg_charset(1));
```
percona-ysorokin: Didn't you created a helper function for this?

components/masking_functions/include/udf/udf_utils.h
dutow: I changed these to const char*, as charset is always null terminated. I only used string_view for the == operator :)

include/mysqlpp/udf_context.hpp
dutow: No, it only returns a pointer to existing data.

include/mysqlpp/udf_context.hpp
dutow: For now I left it as is, as it doesn't cause unresolved externals in other components - should I change it?

include/mysqlpp/udf_wrappers.hpp
dutow: Done.

components/masking_functions/src/udf/udf_registration.cc
dutow: I did, I changed it to use it where possible.

mysql-test/suite/component_masking_functions/r/mask_canada_sin.result
```diff
@@ -0,0 +1,42 @@
...
+ERROR HY000: mask_canada_sin<string> UDF failed; Wrong argument: must be between 9 and 11 characters
+SELECT mask_canada_sin("this is not an canada_sin");
+ERROR HY000: mask_canada_sin<string> UDF failed; Wrong argument: must be between 9 and 11 characters
+SELECT mask_canada_sin("123-456-789", "!");
+mask_canada_sin("123-456-789", "!")
+!!!-!!!-!!!
+SELECT mask_canada_sin("123456789", "!");
+mask_canada_sin("123456789", "!")
+!!!!!!!!!
+# check charset of mask_pan_relaxed()
```
eleo007: Hi, seems like missprint,  mask_canada_sin should be instead of mask_pan_relaxed.

percona-ysorokin:
Closed in favour of https://github.com/percona/percona-server/pull/5114

PR #5065, opened 2023-06-14
PS-8648 [8.0]: Update AMI with Ubuntu 22.04 x86-64 for Cirrus CI

inikep:
1. Update gcc-11 and clang-15
2. Add gcc-13
3. Increase size of gp2 storage to 40 GB


percona-ysorokin (approved):
LGTM

PR #5066, opened 2023-06-14
Set variables to PARENT_SCOPE in CMake functions

keeper:
8e3cf5e5 replaces macros with functions, but in CMake functions have different scope and the variables set in the functions are not visible to the caller. The change breaks Percona/MySQL building with `WITH_ZLIB="system"` for the reason above (ZLIB_FOUND is not visible when returning from `FIND_SYSTEM_ZLIB()`).

Setting the necessary variables to PARENT_SCOPE (equal to returning variables) solves this issue.


robinnewhouse:
Could we request a review of this PR? Thank you.

inikep:
Thanks for your PR. The patch was accepted by upstream at https://github.com/percona/percona-server/commit/486fb151d050200cae8bb3a3a65b273306ede386 and is included in [Percona-Server-8.0.35-27](https://github.com/percona/percona-server/releases/tag/Percona-Server-8.0.35-27) and newer.

PR #5069, opened 2023-06-15
PS-8811: Linking issues with protobuf with macOS 12 on Azure Pipelines

inikep:
Fix by using `brew install protobuf@21` which is compatible with MySQL 8.0.33.


percona-ysorokin (approved):
LGTM

PR #5072, opened 2023-06-19
PS-8816 fix: main.execution_constants main.sp main.sp-error fail on binaries built with clang-16 / gcc-13

percona-ysorokin:
https://jira.percona.com/browse/PS-8816

Stack overrun checks didn't work correctly for server built with
'clang-16' / 'gcc-13' causing failures of some MTR test cases
'main.sp', 'main.sp-error', 'main.execution_constants', etc.).

The problem occurred because the CMake check that defines stack
growth direction constant ('STACK_DIRECTION'), used in stack overrun
checking function 'check_stack_overrun()', didn't work correctly.
Helper program used by this CMake check ('cmake/stack_direction.c')
failed to compile due to 'incompatible-function-pointer-types' warning
being promoted to an error with newer compilers and this compilation
failure was ignored and wrongly interpreted as an indication of wrong
stack growth direction instead.

Fixed 'ptr_f' function pointer declaration in 'cmake/stack_direction.c'.
Instead of declaring a pointer to a function that returns 'volatile int'
we make the pointer to the function itself 'volatile'.

Also fixed the problem in 'configure.cmake' with
'TRY_RUN(... cmake/stack_direction.c ...)' ignoring compilation
result.

As under ASan (with 'detect_stack_use_after_return' enabled) stack
growth direction is just a wrong concept (in this case stack frames are
allocated on heap with pretty random addresses) 'check_stack_overrun()'
function modified so that under ASan it always returns 'false' (meaning
no stack overrun detected).

The following MTR test cases marked with 'not_asan.inc':
* main.alter_debug
* main.execution_constants
* main.sp-error
* main.sp
* main.view_debug

### DIFF
--- a/configure.cmake
+++ b/configure.cmake
@@ -423,19 +423,42 @@ int main()
 IF(NOT STACK_DIRECTION)
   IF(CMAKE_CROSSCOMPILING)
    MESSAGE(FATAL_ERROR 
-   "STACK_DIRECTION is not defined.  Please specify -DSTACK_DIRECTION=1 "
-   "or -DSTACK_DIRECTION=-1 when calling cmake.")
+   "STACK_DIRECTION is not defined.  Please specify -DSTACK_DIRECTION=1, "
+   "-DSTACK_DIRECTION=-1 or -DSTACK_DIRECTION=0 (disables stack overrun "
+   "checking, should be used for ASan builds) when calling cmake.")
   ELSE()
-    TRY_RUN(STACKDIR_RUN_RESULT STACKDIR_COMPILE_RESULT    
-     ${CMAKE_BINARY_DIR} 
-     ${CMAKE_SOURCE_DIR}/cmake/stack_direction.c
-     )
-     # Test program returns 0 (down) or 1 (up).
-     # Convert to -1 or 1
-     IF(STACKDIR_RUN_RESULT EQUAL 0)
-       SET(STACK_DIRECTION -1 CACHE INTERNAL "Stack grows direction")
-     ELSE()
-       SET(STACK_DIRECTION 1 CACHE INTERNAL "Stack grows direction")
+    # Under Address Sanitizer when "detect_stack_use_after_return" runtime
+    # option is enabled (either via ASAN_OPTIONS environment variable or via
+    # "__asan_default_options()" function), there is no such concept as stack
+    # growth direction as stack frames are allocated with
+    # "__asan_stack_malloc()" special function that can return random
+    # addresses.
+    # Here, we just detect that the binaries are compiled under ASan
+    # and set "STACK_DIRECTION" to a special value "0" which will be
+    # processed by the #cmakedefine in the "config.h.cmake" into an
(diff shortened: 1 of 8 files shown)


percona-ysorokin:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-param-parallel-mtr/5/

inikep (approved):
LGTM

configure.cmake
```diff
@@ -423,19 +423,42 @@ int main()
 IF(NOT STACK_DIRECTION)
   IF(CMAKE_CROSSCOMPILING)
    MESSAGE(FATAL_ERROR 
-   "STACK_DIRECTION is not defined.  Please specify -DSTACK_DIRECTION=1 "
-   "or -DSTACK_DIRECTION=-1 when calling cmake.")
+   "STACK_DIRECTION is not defined.  Please specify -DSTACK_DIRECTION=1, "
+   "-DSTACK_DIRECTION=-1 or -DSTACK_DIRECTION=-0 (inapplicable under ASan) "
```
dlenev: Perhaps it is better to replace "inapplicable under ASan" with something like "Disables stack overrun checking. Should be used for ASan builds." ?

configure.cmake
```diff
@@ -423,19 +423,42 @@ int main()
...
+    # "__asan_default_options()" function), there is no such concept as stack
+    # growth direction as stack frames are allocated with
+    # "__asan_stack_malloc()" special function that can return random
+    # addresses.
+    # Here, we just detect that the binaries are compiled under ASan
+    # and set "STACK_DIRECTION" to a special value "0" which will be
+    # processed by the #cmakedefine in the "config.h.cmake" into an
+    # /* #undef STACK_DIRECTION */
+    IF(WITH_ASAN)
+       SET(STACK_DIRECTION 0 CACHE INTERNAL "Stack grows direction")
```
dlenev: Since you are updating those lines anyway, perhaps we should do here and later: s/grows/growth ?

dlenev (commented):
Hello Yura!

I have a few minor comments/suggestions about your patch:

First of all I think it is a good idea to explicitly describe in the changeset comment itself what the problem was and what was the cause of it (so people don't need to go to Jira ticket for this information). For example we can say something like:

"
Stack overrun checks didn't work correctly for server built with clang-16/gcc-13 causing
failures of some tests. 

The problem occurred because check that determines stack growth direction used in stack overrun checking didn't work correctly. Helper program used by this code has failed to compile
due to warnings being promoted to errors with newer compilers and this failure was ignored 
and interpreted as indication of wrong stack direction instead.
"
What do you think?

configure.cmake
percona-ysorokin: fixed

configure.cmake
percona-ysorokin: fixed

dlenev:
Looks great to me!

dlenev (approved):

PR #5073, opened 2023-06-20
PS-8772: Stall of instance after INSERT (...) SELECT.

dlenev:
https://jira.percona.com/browse/PS-8772

Some queries that used a big intrinsic temporary InnoDB table might have
led to InnoDB SE stall (especially when InnoDB buffer pool was small).
Also concurrent execution of those queries could have led to the same
problem or degraded performance.

The problem occured for queries that scanned through such intrinsic tables
using auto-created secondary key rather than clustered key.

What happened was that in such cases some of the pages belonging to
intrinsic table and accessed by the query stayed fixed in buffer pool
until the end of statement (and not until the end of operation on
specific row as it happens for normal InnoDB tables or only while they
are inspected as it happens for scans of intrinsic tables over clustered
key). As result these buffer fixed pages might have consumed the whole
buffer pool at some point and further statement progress was blocked.

Such behavior was result of combination of the following factors:
1) operations on intrinsic temporary tables are optimized and use single
    mini-transaction for all row operations in the statement (while
    for normal tables commit of mini-transaction happens after
    processing of each record.
2) normally, all buffer fixed pages are unfixed from memory only at
    commit of mini-transaction.
3) process of scanning over a key is optimized to immediately unfix pages
    once the cursor leaves them. However, when scan happens over a secondary
    key additional lookups on the clustered key are performed. And buffer
    pages which are fixed by these additional lookups stay fixed until
    the end of mini-transaction.

This fix solves the problem by ensuring that we commit mini-transaction
which is used for access to initrinsic temporary table after performing
operations on each 100 rows. This unfixes buffer pages touched by
mini-transaction making them available for further use by other or this
statements.

This patch might lead to 1-3% performance reduction in some of scenarios
which heavily use intrinsic temporary tables. OTOH we also have seen
significant performance improvements in other scenarios where big
intrinsic tables were used. So we consider this acceptable.
    
A better fix for this problem might be possible, but it is likely
to be more complex, and more appropriate for upstream in our
opinion.

### DIFF
--- a/mysql-test/suite/innodb/t/optimizer_temporary_table_2.test
+++ b/mysql-test/suite/innodb/t/optimizer_temporary_table_2.test
@@ -1,0 +1,50 @@
+--echo #
+--echo # Test for PS-8772: "Stall of instance after INSERT (...) SELECT".
+--echo #
+
+--echo # Check that InnoDB buffer pool size is set to minimum/5Mb and thus is
+--echo # small enough to trigger original problem. If it ever changes the size
+--echo # of source table below needs to be changed proportionately.
+--let $innodb_buffer_pool_size = `SELECT @@global.innodb_buffer_pool_size`
+--let $assert_text = InnoDB Buffer Pool has minimal size (5Mb).
+--let $assert_cond = $innodb_buffer_pool_size = 5242880
+--source include/assert.inc
+
+CREATE TABLE source (id INT PRIMARY KEY AUTO_INCREMENT, a VARCHAR(100), b VARCHAR(100), c VARCHAR(100), d VARCHAR(100), e VARCHAR(100), hdl_source_id INT, hdl_created_date DATETIME);
+CREATE TABLE dest LIKE source;
+
+--echo # Populate source table with 64K records.
+INSERT INTO source (a, b, c, d, e, hdl_source_id, hdl_created_date)
+  SELECT 'aaaaaaaaa', 'bbbbbbbbbb', 'cccccccccccccccc', 'dddddddddddddddddddddddddddd', 'eeeeeeeeeeeeeeeeee', value, NOW() FROM SEQUENCE_TABLE(65536) AS tt;
+SELECT count(*) FROM source;
+
+--echo # Run query that creates big intrinsic temporary table. Prior to the
+--echo # fix it has exhausted buffer pool by keeping all its pages buffer
+--echo # fixed in memory, so query stalled.
+--echo #
+--echo # Note that to trigger problematic scenario query must perform scan
+--echo # over <auto_key0> secondary key on intrinsic temporary table
+--echo # (and auxiliary lookups to confirm record status on clustered key).
+--echo # In this case the secondary key scan fixes and unfixes pages
+--echo # through which it iterates, while pages accessed during auxiliary
+--echo # lookups on clustered key stay buffer fixed until mtr commit.
+--echo #
+--echo # In cases when intrinsic temporary table is read using clustered
+--echo # key only the problem didn't occur, as similary to previous case
+--echo # key scan fixes and unfixes pages through which it iterates, but
+--echo # there are no auxiliary lookups to buffer fix additional pages.
+--let $query = INSERT INTO dest SELECT id,a,b,c,d,e,hdl_source_id,hdl_created_date FROM (SELECT *, ROW_NUMBER() OVER(PARTITION BY id ORDER BY hdl_source_id DESC, hdl_created_date DESC) rn FROM source) a WHERE rn=1
+--replace_column 10 #
+--eval EXPLAIN $query
+--eval $query
+
+--echo # Queries that used several cursors over intrinsic temporary tables
+--echo # were also affected by the same problem. Again problem was visible
+--echo # only if scan over secondary key was used.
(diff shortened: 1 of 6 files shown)


dlenev:
Jenkins results:https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-parallel-mtr/119/

storage/innobase/row/row0ins.cc
```diff
@@ -2636,9 +2636,15 @@ static dberr_t row_ins_sorted_clust_index_entry(ulint mode, dict_index_t *index,
...
+  is not cached. Also commit mtr and do the search if we have inserted
+  more than MAX_INTRINSIC_MTR_RECORDS within this mtr. The latter is
+  done to unfix buffer pages used by the mtr. Otherwise statements
+  (even single statement in extreme case) using intrinsic tables can
+  consume the whole buffer pool by their buffer fixed pages causing
+  performance problems or even stalls. */
   if (mode == BTR_MODIFY_TREE || index->last_ins_cur->rec == nullptr ||
-      index->last_ins_cur->disable_caching) {
+      index->last_ins_cur->disable_caching ||
+      index->last_ins_cur->mtr_records > MAX_INTRINSIC_MTR_RECORDS) {
```
satya-bodapati, line 2647: Can we use this new condition and set index->last_ins_cur->disable_caching?

dlenev: Hello Satya!

In theory we can. But I think we should not do this due to following reasons:

1) Currently semantics of disable_caching assumes that it stays active even after mtr-commit (until end
    of statement?)... It indicates condition related to table/query structure rather than something specific
    to the row being processed. If we are to change this we will have to add code which will clear this flag
    at least in some cases on mtr-commit and as result we will end up in more complex code.
2) code in row0sel.cc doesn't check this condition, so either we will get discrepancy between insert
    and select cases, or we will have to update row0sel.cc as well (and code for its clearing on mtr-commit
    too). Again the code will become more complex.

What do you think?

satya-bodapati: Ok, the intention of the suggestion is to make it simpler :) OK to keep it as is.

storage/innobase/row/row0ins.cc
```diff
@@ -2710,6 +2716,7 @@ static dberr_t row_ins_sorted_clust_index_entry(ulint mode, dict_index_t *index,
         index->last_ins_cur->rec = insert_rec;
 
         index->last_ins_cur->block = cursor.page_cur.block;
+        index->last_ins_cur->mtr_records++;
```
satya-bodapati, line 2719: When do we reset it to 0?

satya-bodapati: ignore, on cursor release()

storage/innobase/row/row0sel.cc
```diff
@@ -4009,6 +4009,7 @@ dberr_t row_search_no_mvcc(byte *buf, page_cur_mode_t mode,
   cursor is open to return all the records. */
   if (direction != 0) {
     if (prebuilt->m_temp_read_shared && !prebuilt->m_temp_tree_modified) {
+      // QQ: Should we check mtr_records on this branch as well?
```
satya-bodapati: tree is not modified, will we use this cached cursor and keep mtr active here?

I think its not required but please check a bit

dlenev: As discussed on Slack it turns out we need to periodically commit mtr in this case as well.
Added test case for this scenario.

mysql-test/suite/innodb/t/optimizer_temporary_table_2.test
```diff
@@ -0,0 +1,57 @@
...
+CREATE TABLE source (id int primary key auto_increment, a varchar(100), b varchar(100), c varchar(100), d varchar(100), e varchar(100), hdl_source_id int, hdl_created_date datetime);
+CREATE TABLE dest LIKE source;
+
+--echo # Populate source table with 64K records.
+--disable_query_log
+INSERT INTO source (a, b, c, d, e, hdl_source_id, hdl_created_date) values ('aaaaaaaaa', 'bbbbbbbbbb', 'cccccccccccccccc', 'dddddddddddddddddddddddddddd', 'eeeeeeeeeeeeeeeeee', 1, NOW());
+--let $i= 0
+while ($i < 16)
+{
+  INSERT INTO source (a, b, c, d, e, hdl_source_id, hdl_created_date) SELECT a, b, c, d, e, hdl_source_id + 1, NOW() FROM source;
```
percona-ysorokin: Just a suggestion here.
Percona Server has the `SEQUENCE_TABLE()` table-level function (like `JSON_TABLE()`).
https://docs.percona.com/percona-server/8.0/flexibility/sequence_table.html
Basically you can do something like
```sql
INSERT INTO source (a, b, c, d, e, hdl_source_id, hdl_created_date)
  SELECT 'aaaaaaaaa', 'bbbbbbbbbb', 'cccccccccccccccc', 'dddddddddddddddddddddddddddd', 'eeeeeeeeeeeeeeeeee', value, NOW()
    FROM SEQUENCE_TABLE(65536) AS tt;
```

satya-bodapati (approved):
LGTM! Please test on Jenkins before merging the fix

mysql-test/suite/innodb/t/optimizer_temporary_table_2.test
dlenev: Done.

dlenev:
Here are Jenins result for this patch https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-parallel-mtr/136/ . They seem OK to me.

PR #5076, opened 2023-06-24
PS-8820 [5.7]: Add gcc-13 and clang-16 to Azure Pipelines and fix gcc-13 compilation issues

inikep:
1. Add gcc-13 and clang-16 to Azure Pipelines
2. Update RocksDB submodule pointer


percona-ysorokin (approved):
LGTM

PR #5077, opened 2023-06-26
PS-8624: Expose number of TABLE objects with triggers in Table Cache

dlenev:
https://jira.percona.com/browse/PS-8624

Follow-up to PS-7963: "Do not cache trigger definitions in the table cache for SELECT statements".

Improve observability of the Table Cache by adding "Open_tables_with_triggers" status variable which shows current number of TABLE instances with fully-loaded triggers in the Table Cache.

The new status variable in addition to existing "Table_open_cache_triggers_hits" and "Table_open_cache_triggers_misses" status variables should allow users better understand how well caching for TABLE objects with fully-loaded triggers is working and if value of option --table_open_cache_triggers, which limits number of such objects in the cache, needs to be adjusted.

### DIFF
--- a/mysql-test/include/assert_trigger_cache_increment.inc
+++ b/mysql-test/include/assert_trigger_cache_increment.inc
@@ -3,13 +3,13 @@
 # Check that values of system status variables related to storing
 # triggers in Table Cache (Table_open_cache_misses,
 # Table_open_cache_trigger_hits, Table_open_cache_triggers_misses,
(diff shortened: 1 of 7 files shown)


percona-ysorokin (approved):
LGTM

dlenev:
Jenkins result seems to be OK: https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/127/

PR #5079, opened 2023-06-27
PS-8810: FEDERATED Update on Table can silently fail when Table contains DOUBLE Columns

kamil-holubicki:
https://jira.percona.com/browse/PS-8810

Problem:
If the table contains DOUBLE column with specified precision, update of the row, even if DOUBLE column is not specified in WHERE clause, reports that row has been updated, but the row is not updated.

Cause:
Federated SE constructs the query in a way that specifies all columns in WHERE clause even if the original query didn't contain all queries.

If the row contains DOUBLE(M,D) column, the value in WHERE clause is rounded to D digits after the decimal point compared to the underlying, actually stored value. This is because of the field definition. This rounded value goes to the server side and appears there as Item_decimal. It is converted to double in the following way:
  1. convert decimal to a string by decimal2string()
  2. convert string to double by my_strtod() Then analyze_field_constant() detects that the result would be rounded for comparison with the field having precision (M,D), so it is decided that row matching is not safe and query execution does not continue.

Solution:
Force server side to pass stored value through string to double conversion which results in the same number as the conversion of provided constant to double (both go through the same conversion path).

### DIFF
--- a/storage/federated/ha_federated.cc
+++ b/storage/federated/ha_federated.cc
@@ -1958,6 +1958,74 @@ int ha_federated::repair(THD *, HA_CHECK_OPT *check_opt) {
 }
 
 /*
+  We need to handle DOUBLE type in a special way.
+  This is because Federated SE does the update in the following way:
+  1. Query the remote server for the row(s) to be updated
+  2. Call update_row() for each row
+  3. Construct update query together with WHERE clause which contains all
+  columns (even if they were not specified in the original query)
+  4. Query the remote server
+
+  If the row contains DOUBLE(M,D) column, the value in WHERE clause is rounded
+  to D digits after the decimal point comparing to the underlying, actually
+  stored value. This is because of field definition. This rounded value goes to
+  the server side and appears there as Item_decimal. It is converted to double
+  in the following way:
+  1. convert decimal to string by decimal2string()
+  2. convert string to double by my_strtod()
+  Then analyze_field_constant() detects that the result would be rounded for
+  comparison with field having precision (M,D), so it is decided that row
+  matching is not safe and query execution does not continue.
+
+  This is desired behavior because if we have table:
+  CREATE TABLE t1 (id INT PRIMARY KEY AUTO_INCREMENT, i INT, d DOUBLE(8,6));
+  INSERT INTO f1(i, d) VALUES (0, 8.730550);
+  and we execute:
+  UPDATE t1 set i=1 where d=8.7305501;
+  we do not expect row to be updated.
+
+  Note that there is the same issue even if we use a single server.
+  If DOUBLE(M,D) column is specified in WHERE clause, the column is row is
+  not found. Like in the above example: UPDATE t1 set i=1 where d=8.730550;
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-pipeline-parallel-mtr/126/

storage/federated/ha_federated.cc
```diff
@@ -2054,31 +2107,38 @@ int ha_federated::update_row(const uchar *old_data, uchar *) {
...
     if (bitmap_is_set(table->read_set, (*field)->field_index())) {
-      size_t field_name_length = strlen((*field)->field_name);
-      append_ident(&where_string, (*field)->field_name, field_name_length,
-                   ident_quote_char);
-      if ((*field)->is_null_in_record(old_data))
-        where_string.append(STRING_WITH_LEN(" IS NULL "));
-      else {
-        bool needs_quote = (*field)->str_needs_quotes();
-        where_string.append(STRING_WITH_LEN(" = "));
+      if ((*field)->type() == MYSQL_TYPE_DOUBLE &&
```
percona-ysorokin: Do we need to do the same for `MYSQL_TYPE_FLOAT`?
Also `MYSQL_TYPE_DECIMAL` and `MYSQL_TYPE_NEWDECIMAL`

storage/federated/ha_federated.cc
```diff
@@ -1957,6 +1957,59 @@ int ha_federated::repair(THD *, HA_CHECK_OPT *check_opt) {
...
+
+  This is desired behavior because if we have table:
+  CREATE TABLE t1 (id INT PRIMARY KEY AUTO_INCREMENT, i INT, d DOUBLE(8,6));
+  INSERT INTO f1(i, d) VALUES (0, 8.730550);
+  and we execute:
+  UPDATE t1 set i=1 where d=8.7305501;
+  we do not expect row to be updated.
+
+  Note that there is the same issue even if we use a single server.
+  If we DOUBLE(M,D) column is specified in WHERE clause, the column is row is
```
percona-ysorokin:
```suggestion
  If DOUBLE(M,D) column is specified in WHERE clause, the column is
```

storage/federated/ha_federated.cc
kamil-holubicki: There is special handling of float type in analyze_real_field_constant(). It explicitly downcasts double value to float and compares floats. We may have a precision mismatch for double comparison, but as we downcast to float, values are equal.

For DECIMAL type if the constant provided in WHERE clause fits the field precision, it is represented internally as Item_decimal 'as is' without any conversions and roundups.

storage/federated/ha_federated.cc
kamil-holubicki: fixed

kamil-holubicki:
As we discussed on Slack:
> may be we should use your new handle_double_type() ONLY if the field has this (M, D) specification
I've limited the fix only to the case where precision is specified explicitly. I've also extended the MTR test.

percona-ysorokin (approved):
LGTM

PR #5085, opened 2023-07-27
PS-8592: XCom connection stalled forever in read() syscall over network

venkatesh-prasad-v:
https://jira.percona.com/browse/PS-8592

Description
-----------
GR suffered from problems caused by the security probes and network scanner processes connecting to the group replication communication port. This usually is not a problem, but poses a serious threat when another member tries to join the cluster by initialting a connection to the member which is affected by external processes using the port dedicated for group communication for longer durations.

On such activites by external processes, the SSL enabled server stalled forever on the SSL_accept() call waiting for handshake data. Below is the stacktrace:

    Thread 55 (Thread 0x7f7bb77ff700 (LWP 2198598)):
    #0 in read ()
    #1 in sock_read ()
    #2 in BIO_read ()
    #3 in ssl23_read_bytes ()
    #4 in ssl23_get_client_hello ()
    #5 in ssl23_accept ()
    #6 in xcom_tcp_server_startup(Xcom_network_provider*) ()

When the server stalled in the above path forever, it prohibited other members to join the cluster resulting in the following messages on the joiner server's logs.

    [ERROR] [MY-011640] [Repl] Plugin group_replication reported: 'Timeout on wait for view after joining group'
    [ERROR] [MY-011735] [Repl] Plugin group_replication reported: '[GCS] The member is already leaving or joining a group.'

Solution
--------
This patch adds two new variables

1. group_replication_xcom_ssl_socket_timeout

   It is a file-descriptor level timeout in seconds for both accept() and SSL_accept() calls when group replication is listening on the xcom port. When set to a valid value, say for example 5 seconds, both accept() and SSL_accept() return after 5 seconds. The default value has been set to 0 (waits infinitely) for backward compatibility. This variable is effective only when GR is configred with SSL.

2. group_replication_xcom_ssl_accept_retries

   It defines the number of retries to be performed before closing the socket. For each retry the server thread calls SSL_accept()  with timeout defined by the group_replication_xcom_ssl_socket_timeout for the SSL handshake process once the connection has been accepted by the first accept() call. The default value has been set to 10. This variable is effective only when GR is configred with SSL.

Note:
- Both of the above variables are dynamically configurable, but will become effective only on START GROUP_REPLICATION.

Testing Done
----
Date: 27th July 2023
Jenkins: https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-param-parallel-mtr/9/

Failing tests: None
| Sl.No | Test name | Remarks |
|--------|--------|--------|
| | | | |

Manual testing:

Test 1: To show that external connections are closed upon timeout

1. Start group_replication on node1 with socket timeout enabled. (timeout=3 seconds and retries= 3 attempts). Uses 5555 as the port for the group communication.
2. Connect a mysql client to 5555
3. Expected behavior: connection is closed by the server after 12 seconds (initial attempt of 3 seconds + retry attempts of 9 seconds)

https://github.com/percona/percona-server/assets/63641998/1c87b1a6-681e-4011-8fbb-6e9ef958c917

Test 2: To show that join request from other nodes are honored after the external conenction has been closed.
1. Start group_replication on node1 with socket timeout enabled. (timeout=3 seconds and retries= 3 attempts). Uses 5555 as the port for the group communication.
2. Connect a mysql client to 5555
3. Try to join node2 into the cluster
4. Expected behavior: connection is closed by the server after 12 seconds (initial attempt of 3 seconds + retry attempts of 9 seconds) and node2 should be able to join the cluster.

https://github.com/percona/percona-server/assets/63641998/53be271b-d1b8-4a9a-ac85-db3d21a0132e

### DIFF
--- a/plugin/group_replication/src/plugin.cc
+++ b/plugin/group_replication/src/plugin.cc
@@ -312,6 +312,14 @@ ulong get_certification_loop_chunk_size_var() {
   return ov.certification_loop_chunk_size_var;
 }
 
+ulong get_xcom_ssl_socket_timeout_var() {
+  return ov.xcom_ssl_socket_timeout_var;
+}
+
+ulong get_xcom_ssl_accept_retries_var() {
+  return ov.xcom_ssl_accept_retries_var;
+}
+
 long get_flow_control_certifier_threshold_var() {
   return ov.flow_control_certifier_threshold_var;
 }
@@ -2524,6 +2532,9 @@ int build_gcs_parameters(Gcs_interface_parameters &gcs_module_parameters) {
     std::string ssl_crlpath("");
     std::string tls_version("");
     std::string ssl_fips_mode("");
+    std::string xcom_ssl_socket_timeout("");
+    std::string xcom_ssl_accept_retries("");
+
     if (xcom_comm_protocol == XCOM_PROTOCOL) {
       ssl_key.append(sv.ssl_key ? sv.ssl_key : "");
       ssl_cert.append(sv.ssl_cert ? sv.ssl_cert : "");
@@ -2534,6 +2545,16 @@ int build_gcs_parameters(Gcs_interface_parameters &gcs_module_parameters) {
       ssl_crlpath.append(sv.ssl_crlpath ? sv.ssl_crlpath : "");
       tls_version.append(sv.tls_version ? sv.tls_version : "");
       ssl_fips_mode.append(ov.ssl_fips_mode_values[sv.ssl_fips_mode]);
+
+      if (ov.xcom_ssl_socket_timeout_var > 0) {
+        xcom_ssl_socket_timeout.append(
+            std::to_string(ov.xcom_ssl_socket_timeout_var));
+      }
+
+      if (ov.xcom_ssl_accept_retries_var > 0) {
+        xcom_ssl_accept_retries.append(
+            std::to_string(ov.xcom_ssl_accept_retries_var));
+      }
     } else if (xcom_comm_protocol == MYSQL_PROTOCOL) {
       ssl_key.append(ov.recovery_ssl_key_var ? ov.recovery_ssl_key_var : "");
       ssl_cert.append(ov.recovery_ssl_cert_var ? ov.recovery_ssl_cert_var : "");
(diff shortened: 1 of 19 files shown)


plugin/group_replication/src/plugin.cc
```diff
@@ -2534,6 +2545,18 @@ int build_gcs_parameters(Gcs_interface_parameters &gcs_module_parameters) {
       ssl_crlpath.append(sv.ssl_crlpath ? sv.ssl_crlpath : "");
       tls_version.append(sv.tls_version ? sv.tls_version : "");
       ssl_fips_mode.append(ov.ssl_fips_mode_values[sv.ssl_fips_mode]);
+
+      if (ov.xcom_ssl_socket_timeout_var > 0) {
+        std::stringstream ss;
+        ss << ov.xcom_ssl_socket_timeout_var;
+        xcom_ssl_socket_timeout.append(ss.str());
```
percona-ysorokin:
```suggestion
        xcom_ssl_socket_timeout.append(std::to_string(ov.xcom_ssl_socket_timeout_var));
```

plugin/group_replication/src/plugin.cc
```diff
@@ -2534,6 +2545,18 @@ int build_gcs_parameters(Gcs_interface_parameters &gcs_module_parameters) {
...
+      }
+
+      if (ov.xcom_ssl_accept_retries_var > 0) {
+        std::stringstream ss;
+        ss << ov.xcom_ssl_accept_retries_var;
+        xcom_ssl_accept_retries.append(ss.str());
```
percona-ysorokin: Likewise

percona-ysorokin (approved):
LGTM

(minor stylistic comments - can be ignored)

plugin/group_replication/src/plugin.cc
venkatesh-prasad-v: Done

plugin/group_replication/src/plugin.cc
venkatesh-prasad-v: Done

satya-bodapati:
@venkatesh-prasad-v can the same happen without GR? ie on the server listening port?

satya-bodapati (approved):

satya-bodapati:
> @venkatesh-prasad-v can the same happen without GR? ie on the server listening port?

reply from slack:
Server uses a different kind of mechanism to handle incoming connections and we dont see this issue there.

PR #5086, opened 2023-07-27
PS-8848: Make connection control plugin more verbose

oleksandr-kachan:
https://jira.percona.com/browse/PS-8848

The Connection Control plugin is designed to introduce random delay for a user connection attempt in case number of failing attempts reaches some threshold. At the moment plugin adds this delay silently so you cannot notice the plugin is actually doing something by looking into server's logs.

Improved plugin to write an information message to server's error log in case number of failed connection attempts crosses the threshold. To control number of those messages printed to log in cases of a big number of failing connection attempts for different accounts, the limit of 10 messages per minute is hardcoded.

### DIFF
--- a/plugin/connection_control/log_rate_limiter.cc
+++ b/plugin/connection_control/log_rate_limiter.cc
@@ -1,0 +1,87 @@
+/* Copyright (c) 2023 Percona LLC and/or its affiliates. All rights reserved.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; version 2 of the License.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA */
+
+#include "log_rate_limiter.h"
+
+#include "mysqld_error.h"
+
+#include <chrono>
+
+namespace {
+
+constexpr uint64_t RATE_LIMIT_INTERVAL_SECONDS = 60;
+constexpr uint64_t RATE_LIMIT_MESSAGES_PER_INTERVAL = 2;
+
+}  // namespace
+
+namespace connection_control {
+
+LogRateLimiter::LogRateLimiter()
+    : m_sent_messages_per_interval_count{0},
+      m_delayed_connections_per_interval_count{0},
+      m_interval_start_timestamp{static_cast<uint64_t>(
+          std::chrono::duration_cast<std::chrono::seconds>(
+              std::chrono::system_clock::now().time_since_epoch())
+              .count())} {}
+
+void LogRateLimiter::report_delayed_connection(
+    const Sql_string &user_host, bool is_threshold_crossed,
+    Error_handler *error_handler) noexcept {
+  const uint64_t timestamp_now =
+      std::chrono::duration_cast<std::chrono::seconds>(
+          std::chrono::system_clock::now().time_since_epoch())
+          .count();
+
+  if (timestamp_now - m_interval_start_timestamp >
+      RATE_LIMIT_INTERVAL_SECONDS) {
+    if (m_delayed_connections_per_interval_count > 0 &&
+        m_delayed_connections_per_interval_count >
+            m_sent_messages_per_interval_count) {
+      error_handler->handle_info(
+          ER_CONN_CONTROL_DELAYED_CONN_STATS,
+          static_cast<unsigned long long>(
+              m_delayed_connections_per_interval_count),
+          static_cast<unsigned long long>(timestamp_now -
+                                          m_interval_start_timestamp));
+    }
+
+    m_interval_start_timestamp = timestamp_now;
+    m_sent_messages_per_interval_count = 0;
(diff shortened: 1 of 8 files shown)


plugin/connection_control/log_rate_limiter.cc
```diff
@@ -0,0 +1,66 @@
...
+    Error_handler *error_handler) noexcept {
+  const uint64_t timestamp_now =
+      std::chrono::duration_cast<std::chrono::seconds>(
+          std::chrono::system_clock::now().time_since_epoch())
+          .count();
+
+  if (timestamp_now > m_interval_start_timestamp &&
+      timestamp_now - m_interval_start_timestamp >
+          RATE_LIMIT_INTERVAL_SECONDS) {
+    m_interval_start_timestamp = timestamp_now;
```
kamil-holubicki, line 60: Maybe if the new interval starts and something was delayed within the previous interval, we could write the summary of the previous interval.
The current implementation will write 10 messages in 5 seconds, and we will lost information about what happened in the remaining 55 seconds of the interval (how many delayed connections occured, etc)

plugin/connection_control/log_rate_limiter.cc
```diff
@@ -0,0 +1,66 @@
...
+    m_sent_messages_per_interval_count = 0;
+    m_delayed_connections_per_interval_count = 0;
+  }
+
+  m_delayed_connections_per_interval_count++;
+
+  if (is_threshold_crossed &&
+      m_sent_messages_per_interval_count < RATE_LIMIT_MESSAGES_PER_INTERVAL) {
+    m_sent_messages_per_interval_count++;
+    error_handler->handle_info(
```
kamil-holubicki: If it happens in first 5 seconds of interval, we will know nothing about what happened in the remaining 55

plugin/connection_control/log_rate_limiter.cc
```diff
@@ -0,0 +1,66 @@
...
+
+#include "log_rate_limiter.h"
+
+#include "mysqld_error.h"
+
+#include <chrono>
+
+namespace {
+
+const int RATE_LIMIT_INTERVAL_SECONDS = 60;
```
percona-ysorokin: `constexpr uint64_t`

plugin/connection_control/log_rate_limiter.cc
```diff
@@ -0,0 +1,66 @@
...
+const int RATE_LIMIT_MESSAGES_PER_INTERVAL = 10;
```
percona-ysorokin: likewise

share/messages_to_error_log.txt
```diff
@@ -3914,6 +3914,9 @@ ER_CONN_CONTROL_FAILED_TO_SET_CONN_DELAY
 ER_CONN_CONTROL_FAILED_TO_UPDATE_CONN_DELAY_HASH
   eng "Failed to update connection delay hash for account : %s"
 
+ER_CONN_CONTROL_FAILED_CONN_THRESHOLD_REACHED
+  eng "Failing connection attempts threshold reached for %s, %i delayed connection attempts during last %i seconds"
```
percona-ysorokin: `%i` is an incorrect formatter for uint64_t
it should be `PRIu64`
```
"Failing connection attempts threshold reached for %s, %" PRIu64 " delayed connection attempts during last %" PRIu64 " seconds"
```
not sure though if such lines can be inserted into `messages_to_error_log.txt`

plugin/connection_control/log_rate_limiter.cc
```diff
@@ -0,0 +1,66 @@
...
+LogRateLimiter::LogRateLimiter()
+    : m_sent_messages_per_interval_count{0},
+      m_delayed_connections_per_interval_count{0},
+      m_interval_start_timestamp{static_cast<uint64_t>(
+          std::chrono::duration_cast<std::chrono::seconds>(
+              std::chrono::system_clock::now().time_since_epoch())
+              .count())} {}
+
+void LogRateLimiter::report_delayed_connection(
+    const Sql_string &user_host, const bool is_threshold_crossed,
```
percona-ysorokin:
```suggestion
    const Sql_string &user_host, bool is_threshold_crossed,
```

percona-ysorokin (changes requested):

plugin/connection_control/log_rate_limiter.cc
oleksandr-kachan: Done

plugin/connection_control/log_rate_limiter.cc
oleksandr-kachan: Done

plugin/connection_control/log_rate_limiter.cc
oleksandr-kachan: Done

share/messages_to_error_log.txt
oleksandr-kachan: %llu will be used in message template

satya-bodapati: We should never introduce error messages in middle of ER_. it should be always last. Otherwise, other codes will see a shift in number.

satya-bodapati (changes requested):

plugin/connection_control/log_rate_limiter.cc
oleksandr-kachan: Done

share/messages_to_error_log.txt
oleksandr-kachan: Moved messages to the end

kamil-holubicki (approved):
lgtm

percona-ysorokin (approved):
LGTM

satya-bodapati (approved):

PR #5087, opened 2023-07-28
PS-8844: Fix compiler version checks in Azure Pipelines

inikep:
1. Use `$REAL_COMPILER_VER` instead of `$(CompilerVer)`.
2. The method using `bc <<< "$(CompilerVer)` works only with 2 digits (e.g. `7.4`). It doesn't work with 3+ digits (e.g. `7.4.1`).


percona-ysorokin (approved):
LGTM

PR #5088, opened 2023-07-28
PS-8859: Fix audit_log_filter build issue owith GCC-8 or older

oleksandr-kachan:
https://jira.percona.com/browse/PS-8859


percona-ysorokin (approved):
LGTM

PR #5089, opened 2023-08-03
PS-8849 : Unexpected hitting Assertion failure: row0mysql.cc:218:len …

dlenev:
…< 256 * 256 thread.

https://jira.percona.com/browse/PS-8849

Concurrent execution of repeatable-read transactions with queries to a partitioned InnoDB table which used multi-value keys and statements which updated or deleted rows in this table sometimes led to assertion failure.

This assertion failure occurred when transaction with such a query tried to reconstruct older version of row from undo log in order to adhere transaction isolation level rules. For partitioned tables such reconstruction also always involved restoring from undo log value of virtual column for multi-value key. The latter was enforced for all virtual keys by InnoDB partitioning handler code in order to be able to sort records coming from different partitions according to key value. However, for multi-value keys such reconstruction from undo is not done correctly, which resulted in assertion failure.

However, for multi-value keys such reconstruction forced by partitioning doesn't make sense as they can't be used for sorting. So this patch solves the problem by disabling such reconstruction in case of multi-value keys.

### DIFF
--- a/mysql-test/suite/innodb/r/multi_value_basic.result
+++ b/mysql-test/suite/innodb/r/multi_value_basic.result
@@ -1222,3 +1222,27 @@ INSERT INTO t1 VALUES (1);
 INSERT INTO t2 (data) VALUES ('{\"parentID\":1,\"idArray\":[1]}');
 DELETE FROM t1 WHERE id = 1;
 DROP TABLE t2,t1;
+#
+# PS-8849 'Unexpected hitting Assertion failure: row0mysql.cc:218:len < 256 * 256 thread'
+#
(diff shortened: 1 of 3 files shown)


satya-bodapati (approved):
Just for completeness, please the add the same test case without partitions

dlenev:
Hello!

I am closing this PR as we have decided to go with upstream's version of the fix for this problem, see commit 05c1f53be7cb8d9b527eb28e809284b0722c4106.

However, since the latter commit doesn't contain public test case, we will be adding test case from this PR to Percona Sever 8.0.35-27.

PR #5091, opened 2023-08-07
PS-8848: Make connection control plugin more verbose

oleksandr-kachan:
https://jira.percona.com/browse/PS-8848

The Connection Control plugin is designed to introduce random delay for a user connection attempt in case number of failing attempts reaches some threshold. At the moment plugin adds this delay silently so you cannot notice the plugin is actually doing something by looking into server's logs.

Improved plugin to write an information message to server's error log in case number of failed connection attempts crosses the threshold. To control number of those messages printed to log in cases of a big number of failing connection attempts for different accounts, the limit of 10 messages per minute is hardcoded.
In case messages for some connection attempts were suppressed during the interval, there will be a message reporting overall number of delayed connection attempts during this period in the end of each interval.

(cherry picked from commit b00cf3af0fc2f617afb44e679f580ed478426940)

### DIFF
--- a/plugin/connection_control/CMakeLists.txt
+++ b/plugin/connection_control/CMakeLists.txt
@@ -25,6 +25,7 @@ MYSQL_ADD_PLUGIN(connection_control
   connection_control.cc
   security_context_wrapper.cc
   connection_delay.cc
+  log_rate_limiter.cc
   MODULE_ONLY
   LINK_LIBRARIES extra::rapidjson
   )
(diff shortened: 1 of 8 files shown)


inikep:
I already imported all commits from the current percona/8.0 :
https://github.com/percona/percona-server/commits/release-8.0.34-26

oleksandr-kachan:
> I already imported all commits from the current percona/8.0 : https://github.com/percona/percona-server/commits/release-8.0.34-26

Thanks, I'll close this one

PR #5092, opened 2023-08-07
PS-8042: Rename Audit_log_filter_buffer_bypassing_writes status var (8.0)

oleksandr-kachan:

Upstream added similar status var Audit_log_direct_writes. Renaming
Audit_log_filter_buffer_bypassing_writes to Audit_log_filter_direct_writes
to have similar names.


inikep (changes requested):
What is a reason that we change the name of this variable? Please also update a commit message.

oleksandr-kachan:
> What is a reason that we change the name of this variable? Please also update a commit message.

The reason is to sync with upstream. They added similar status var https://dev.mysql.com/doc/refman/8.1/en/audit-log-reference.html#statvar_Audit_log_direct_writes

inikep (approved):
LGTM

PR #5094, opened 2023-08-08
PS-8844: ASAN errors handling improvements

oleksandr-kachan:
***
PS-8844: Merge MySQL 8.0.34 - Improve detect_stack_use_after_return ASAN option handling

In https://github.com/mysql/mysql-server/commit/75a1d41b2802 ASAN option
detect_stack_use_after_return was disabled for all tests by setting it
in mysql-test-run.pl as detect_stack_use_after_return=false. While this
fixes issues with tests which expect ER_STACK_OVERRUN_NEED_MORE error
to happen this also disabled AddressSanitizerUseAfterReturn check for
all other tests where it could be usefull.
    
To improve this:
- removed detect_stack_use_after_return=false from mysql-test-run.pl leaving it set to its default value;
- skipped MTR tests expecting ER_STACK_OVERRUN_NEED_MORE error when run with ASAN.
    
This is based on the following changes:
- https://github.com/percona/percona-server/pull/5072
- https://github.com/percona/percona-server/commit/cc359ef873de
***
PS-8844: Merge MySQL 8.0.34 - Remove error 42 suppression for mtr tests using 'grep'

In https://github.com/mysql/mysql-server/commit/75a1d41b2802 there was
added error 42 as expected to some tests using 'grep'.
    
This is already handled via ASAN suppressions. Corresponding rule was added
in https://github.com/percona/percona-server/commit/fe198afeda0b26ad885f20f37659b38a00a2e375.
    
Removed error 42 from expected in this change.

### DIFF
--- a/mysql-test/include/have_grep.inc
+++ b/mysql-test/include/have_grep.inc
@@ -4,7 +4,7 @@
 123
 456
 EOF
---error 0,1,2,42,127,512
+--error 0,1,2,127,512
 --exec grep -v -e "123" $MYSQL_TMP_DIR/grep_check
 let $status= $__error;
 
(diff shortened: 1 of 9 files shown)


inikep:
LGTM if MTR tests are fine

inikep (approved):

PR #5096, opened 2023-08-10
PS-8042: Improve audit_log_filter.log_prune_seconds test

oleksandr-kachan:

The audit_log_filter.log_prune_seconds MTR tests expects the plugin to generate a set of rotated log files. Log rotation timestamp at the moment is taken from current system time. When test is run on fast hardware, log rotation happens too soon and with the same timestamp for different logs. As a result, rotated logs get rewritten and expected files set is not generated.

To fix that, modified the test to use DEBUG='+d,audit_log_filter_debug_timestamp'. This ensures each next rotated log file has unique timestamp in its name.


inikep (approved):
LGTM

PR #5098, opened 2023-08-14
PS-8871 : RESOURCE GROUP hint fails when used within a prepared state…

dlenev:
…ment

https://jira.percona.com/browse/PS-8871

RESOURCE_GROUP hints specified in the text of prepared statement didn't apply during its execution.

The problem occurred because information about RESOURCE_GROUP hint (the resource group to be used for statement execution) was associated with connection context (a.k.a. THD) and not with prepared statement context. Moreover, to avoid situation when hint specified for a statement affected execution of further statements in connection, it was reset after execution of the statement. Which is not re-execution safe technique.

This patch solves the problem by associating information about this hint with LEX object which is part of prepared statement context. The problem with a hint specified for a statement affecting execution of further statements in the connection is solved automatically in this case, since LEX is re-inited before parsing each statement.

### DIFF
--- a/mysql-test/r/resource_group_bugs.result
+++ b/mysql-test/r/resource_group_bugs.result
@@ -21,3 +21,60 @@ CREATE RESOURCE GROUP CAFE TYPE=USER VCPU=1-3 THREAD_PRIORITY=5;
 ERROR HY000: Resource Group 'CAFE' exists
 DROP RESOURCE GROUP CaFé;
 SET NAMES default;
+#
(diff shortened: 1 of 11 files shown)


oleksandr-kachan (approved):
LGTM but lets try get it tested on Jenkins

dlenev:
A few tests which fail in Jenkins do not seem to be related to this patch to me. See: https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/199/ So I will proceed with merging.

PR #5103, opened 2023-08-16
PS-8042: Fix std::filesystem link error with gcc 8

oleksandr-kachan:

Uing <filesystem> with gcc-8 requires linking with -lstdc++fs.


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param-parallel-mtr/16/

inikep:
What gcc version and OS fails?
gcc-8.4.0 seems to work fine with Azure and `audit_log_filter`:
https://dev.azure.com/inikep/mysql-server/_build/results?buildId=1989&view=logs&j=1dcbf5ef-0d66-5d5f-acaf-a3b832a85095&t=6b0d68b3-4ebe-54fa-ed19-0d350f11be2f

percona-ysorokin (approved):
LGTM

inikep:
I found also https://en.cppreference.com/w/cpp/filesystem#Notes
```
Using this library may require additional compiler/linker options. GNU implementation prior to 9.1 requires linking with -lstdc++fs and LLVM implementation prior to LLVM 9.0 requires linking with -lc++fs.
```

oleksandr-kachan:
> I found also https://en.cppreference.com/w/cpp/filesystem#Notes

Thanks, will recheck Clang separately

PR #5104, opened 2023-08-16
PS-8869: Semi-sync master waits for ack when semi-sync replica is dow…

venkatesh-prasad-v:
PS-8869: semi-sync master waits for ack when semi-sync replica is down & when
             we replace it with async
    
    https://jira.percona.com/browse/PS-8869
    
    Description
    -----------
    This is the regression introduced in the commit 650d2f7
    (Bug#33534218 : replication_sender_observe_commit_only=ON leads to break semisync replication)
    which made the before_send_hook and after_send_hook to be never called if the
    event is in the excluded group.
    
    Prior to 8.0.31, dump thread performed below operations
    1. Read events from binlog
    2. For every event
       2.1. Call before send hook
       2.2. If the event is skipped
            2.2.1. Send hearbeat event if necessary with the latest position
            2.2.2. Update the exclude_end_pos to mark the position in binlog till
                   where the replica receiver is synchronized.
       2.3. If the event is not skipped
            2.3.1. Send heartbeat event if necessary with the last
                   exclude_group_end_pos.
            2.3.2. Call send_packet() to send events to replica.
       2.4. Call after send hook
    
    This way it ensured that before_send and after_send hooks are executed
    irrespective of whether the event is skipped or not.
    
    In 8.0.31 upto 8.0.34,
    1. Read events from binlog
    2. For every event
       2.1. If the event is skipped
            2.2.1. Send hearbeat event if necessary with the latest position
            2.2.2. Update the exclude_end_pos to mark the position in binlog till
                   where the replica receiver is synchronized.
       2.2. If the event is not skipped
            2.3.1. Send heartbeat event if necessary with the last
                   exclude_group_end_pos.
            2.3.2. Call before send hook
            2.3.3. Call send_packet() to send events to replica.
            2.3.4. Call after send hook
    
    Here, the above upstream commit made the before_send_hook and after_send_hook
    to be called only when an event is not skipped.
    
    However this change caused the client connection thread to wait for the ACK
    from semi-sync replica even after it was added back as a semi-sync replica, as
    this change never reads the acknowledgement from the replica (after send hook
    is not executed).
    
    Solution
    --------
    Make the before_send and after_send server hooks to be called even for
    heartbeat events, preserving the behavior before to 8.0.31, by also preserving
    the original bugfix made in the commit 650d2f7, i.e, Observe_transmission_guard
    is created just before calling before_send_hook, so that reserve_header_hook is
    always called when sending a heartbeat event and magic number is properly
    encoded.

Testing Done:
Jenkins: https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-param-parallel-mtr/11/testReport/

Failing tests are unrelated to the current fix.


sql/rpl_binlog_sender.cc
```diff
@@ -654,29 +655,30 @@ int Binlog_sender::send_events(File_reader &reader, my_off_t end_pos) {
         m_packet.copy(tmp);
         m_packet.length(tmp.length());
       }
+      needs_send_packet = true;
```
satya-bodapati: what about if (send_heartbeat_event(log_pos)) at line 632? do we need to set needs_send_packet=true for that case?

venkatesh-prasad-v: Heartbeat events are signals sent by source to replica only to let the replica know that its source(master) is up and running so that replica does not terminate the connection on absence of data for longer period. These are sent only if there are no updates on the binary log for the duration specified by SOURCE_HEARTBEAT_PERIOD (default 60 seconds) clause in the CHANGE MASTER command.

As these events do not actually contain transaction specific data, we can ignore calling before_send and after_send hooks for heartbeat events.

satya-bodapati: Did you mean ignore calling send_packet()? Becuase after your change, we go back to pre 8.0.31 behaviour and the before/after_send_hook()s are executed independently of send_packet(). Isnt it?

venkatesh-prasad-v: I meant avoiding the call to before_send and after_send hooks for `send_heartbeat_event()`. After this patch we will be calling those hooks only when `send_packet()` is called.

venkatesh-prasad-v: Hi! I am sorry for the confusion and please ignore my earlier comments. 

We must call before_send and after_send hooks for heartbeat events to fix this bug.

So, I have further simplified the patch so that the behavior before 8.0.31 is preserved, and upstream bug fix is also honored.

New logic is
```
if (events need to be skipped) {
  before_send_hook()
  send_heartbeat()
  after_send_hook()
} else {
  Create Observe_transmission_guard
  before_send_hook()
  send_packet()
  after_send_hook()
}
```

satya-bodapati: Thanks!

satya-bodapati (approved):
LGTM!

PR #5106, opened 2023-08-17
PS-8854: Add main.percona_udf MTR test

inikep:
Add a test to check FNV1A_64, FNV_64, and MURMUR_HASH user-defined functions.


percona-ysorokin (approved):
LGTM

PR #5108, opened 2023-08-18
PS-8877 : Selecting from performance_schema.innodb_redo_log_files cra…

dlenev:
…shes MySQL

https://jira.percona.com/browse/PS-8877

Concurrent queries on performance_schema.innodb_redo_log_files sometimes caused assertion failure.

Failure occurred due to trivial race condition when two threads tried to execute Log_files_pfs_table::rnd_next() method concurrently for the same object. This happened because Log_files_pfs_table was made a singleton object, while in theory in PFS plugin API objects which represent table handle/PSI_table_handle are supposed to be per usage in statement and per connection, to be able to store state of a cursor for particular table usage in the query in particular connection.

This patch solves this problem by making Log_files_pfs_table non-singleton object which is allocated on table open and deleted on table close. It also introduces auxiliary singleton Log_files_pfs_table::Share class to which shared information from Log_files_pfs_table has been moved.

### DIFF
--- a/mysql-test/suite/innodb/r/percona_bug_ps8877.result
+++ b/mysql-test/suite/innodb/r/percona_bug_ps8877.result
@@ -1,0 +1,8 @@
+#
+# PS-8877 : Selecting from performance_schema.innodb_redo_log_files crashes MySQL
+#
+CREATE DATABASE mysqlslap;
+# Use 'mysqlslap' to run "SELECT * FROM P_S.innodb_redo_log_files" query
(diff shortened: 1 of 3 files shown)


dlenev:
Here are Jenkins results for these changes https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/258/. All test failures seem to be unrelated to this change to me.

percona-ysorokin (approved):
LGTM

PR #5111, opened 2023-08-22
PS-8844: Fix the failing main.mysqldump_gtid_purged

venkatesh-prasad-v:

This patch fixes the test failure of main.mysqldump_gtid_purged that failed due to the uninitialized variable $redirect_stderr in the start_proc_in_background.inc.


inikep (approved):
LGTM

PR #5112, opened 2023-08-22
PS-8844: Use AWS to test arm64 builds with Cirrus CI

inikep:
Cirrus CI community arm64 workers are always busy so we need to use our own AWS jobs.


percona-ysorokin (approved):
LGTM

PR #5116, opened 2023-08-31
PS-8042: Link audit_log_filter with stdc++fs when built with GCC version less than 9.1 (8.0.34)

oleksandr-kachan:


percona-ysorokin (approved):
LGTM

PR #5118, opened 2023-09-08
PS-8907: Empty vio->signal_mask in the x plugin causes MySQL cannot h…

venkatesh-prasad-v:
…andle signal

https://jira.percona.com/browse/PS-8907

Problem
-------
MySQL server abruptly terminates without performing a shutdown on SIGTERM when it has open connections from X plugin due to the unset signal_mask at the vio layer.

Solution
--------
Set the signal_mask at the time of creation of vio object so that signals are handled properly.

Testing Done
---
Jenkins : https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/275/

Failing tests:
1. x.connection - Fixed.


percona-ysorokin (approved):
LGTM

PR #5120, opened 2023-09-10
PS-8836 fix: MyRocks serialization violation? Read after write is failing

percona-ysorokin:
https://jira.percona.com/browse/PS-8836

This is a cherry-pick of the
"Fix for snapshot conflict on SK due to concurrent update" (commit facebook/mysql-5.6@98da68c)

Summary: The issue that was being hit has two different connections to the MySQL server trying to update two different rows via a secondary index scan with RC isolation level set. During the SK index scan, either one of the iterators may end up returning the value updated by the other connection. This leads to the iterator range condition failing, and the iterator returns an early EOF back, and the update fails to apply on the remaining rows that were actually valid candidates for update based on the WHERE conditions.

Differential Revision: D48740767

fbshipit-source-id: f9b58edc249b6deace936dbafc70f79456eb49ca

### DIFF
--- a/mysql-test/suite/rocksdb/t/sk_rc_concurrent_point_update.test
+++ b/mysql-test/suite/rocksdb/t/sk_rc_concurrent_point_update.test
@@ -1,0 +1,91 @@
+--source include/have_rocksdb.inc
+--source include/have_debug_sync.inc
+
+--source include/count_sessions.inc
+
+--connect(con0, localhost, root,, test)
+--connection default
+
+let $i = 1;
+let $define_row_key = row_key BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,;
+let $define_pk = PRIMARY KEY (row_key),;
+
+while ($i <= 2) {
+
+--echo Conn A creating table
+eval CREATE TABLE table1 (
+  $define_row_key
+  val1 TINYINT NOT NULL,
+  val2 VARCHAR(128) NOT NULL,
+  $define_pk
+  KEY idx_val1 (val1)
+) ENGINE=RocksDB;
+
+# Conn A: first, create a table with a few rows, keeping the val1 column all the same
+INSERT INTO table1 (val1, val2) VALUES (14, 'Alfa'), (14, 'Bravo'), (14, 'Charlie'), (14, 'Delta');
+--echo Conn A: `table1` created with 4 rows
+
+--echo Conn A: Table before
+SELECT * FROM table1;
+
+# next, start a transaction from Conn A to do a locking read for the rows based on a SK condition
+# to be specific, select rows matching val1 = 14
+# the update command should be `sent` after activating a DEBUG_SYNC point, so that the
+# update gets frozen
+
+SET TRANSACTION ISOLATION LEVEL READ COMMITTED;
+START TRANSACTION;
+--echo Conn A: Started TRANSACTION A (SELECT .. FOR UPDATE )
+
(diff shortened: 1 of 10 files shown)


mysql-test/suite/rocksdb/t/sk_rc_concurrent_point_update.test
```diff
@@ -0,0 +1,84 @@
+--connect(con0, localhost, root,, test)
```
inikep, line 6: Please add 
```
```
and
```
```

mysql-test/suite/rocksdb/t/sk_rc_concurrent_update.test
```diff
@@ -0,0 +1,77 @@
+--connect(con0, localhost, root,, test)
```
inikep, line 6: Please add 
```
```
and
```
```

mysql-test/suite/rocksdb/t/sk_rc_read_set.test
```diff
@@ -0,0 +1,63 @@
+--connection default
```
inikep: Please add 
```
```

mysql-test/suite/rocksdb/t/sk_rc_concurrent_point_update.test
```diff
@@ -0,0 +1,84 @@
...
+  $define_pk
+  KEY idx_val1 (val1)
+) ENGINE=RocksDB;
+
+# Conn A: first, create a table with a few rows, keeping the val1 column all the same
+INSERT INTO table1 (val1, val2) VALUES (14, 'Alfa'), (14, 'Bravo'), (14, 'Charlie'), (14, 'Delta');
+--echo Conn A: `table1` created with 4 rows
+
+--echo Conn A: Table before
+SELECT * FROM table1;
```
inikep, line 29: We may consider using
```
--let $assert_text = table1 should have 4 rows
--let $assert_cond = [SELECT COUNT(*) from table1] = 4
--source include/assert.inc
```
and
```
--let $assert_text = Alfa should have val1 equal to 14
--let $assert_cond = [SELECT val1 FROM table1 WHERE val2 = 'Alfa'] = 14
--source include/assert.inc
```
But it's big and I'm not convinced.

inikep (changes requested):

mysql-test/suite/rocksdb/t/sk_rc_concurrent_update.test
percona-ysorokin: fixed

mysql-test/suite/rocksdb/t/sk_rc_concurrent_point_update.test
percona-ysorokin: fixed

mysql-test/suite/rocksdb/t/sk_rc_concurrent_point_update.test
percona-ysorokin: The problem is that `assert.inc` executes a lot of additional internal queries (for parsing `$assert_cond`, for instance) and I am not sure how this will interfere with DEBUG_SYNC points already set. So, I'd rather leave this part as is.

inikep (approved):
LGTM

percona-ysorokin:
Jenkins build link (rocksdb suites)
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-param-parallel-mtr/21/#showFailuresLink

PR #5121, opened 2023-09-14
PS-8882 : Crash when performing CREATE TABLE AS SELECT during Alters/…

dlenev:
…Renames

https://jira.percona.com/browse/PS-8882

CREATE TABLE SELECT crashed if it was run when Data-dictionary Table object Cache was full. The latter situation can occur when one runs RENAME TABLES which renames number of tables larger than Data-dictionary Table object Cache capacity (which is equal to --max_connections).

The problem occured because in situation of full dd::Table object cache any release of dd::Table caused immediate deletion of this object. CREATE TABLE SELECT implementation acquired dd::Table object for source table at some point and then prematurely released it (due to wrong scope of helper dd::cache::Auto_releaser object). Later access to this already released and destroyed dd::Table object caused crash.

This patch solves the problem by using wider scope for Auto_releaser in CREATE TABLE SELECT implementation and thus ensuring that dd::Table object for source table is not released too early, before we stop using it.

### DIFF
--- a/mysql-test/r/percona_bug_ps8882.result
+++ b/mysql-test/r/percona_bug_ps8882.result
@@ -1,0 +1,39 @@
+#
+# PS-8882: Crash when performing CREATE TABLE AS SELECT during Alters/Renames
+#
+CREATE TABLE t0 (id INT PRIMARY KEY, v1 VARCHAR(100), v2 VARCHAR(100), v3 VARCHAR(100));
+CREATE DATABASE db1;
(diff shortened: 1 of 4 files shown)


percona-ysorokin (approved):
LGTM

dlenev:
Jenkins results look OK to me: https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/296/

PR #5125, opened 2023-09-18
PS-8844 [8.0]: Set "-fvisibility-inlines-hidden" for all components

inikep:
By default gcc / clang toolchain, when creating a shared library, marks all symbols as 'exported' - `__attribute__((visibility("default")))`. This includes not only the symbols declared explicitly in the library source code but also those linked from the dependent static libraries as well as symbols instantiated from the standard library templated code. This is not a desired behavior for UDF shared objects, MySQL plugins / components that rely on dynamic symbol lookups ('dlsym()' calls). The exported symbols namespace may be extremely polluted in this case, which may lead to slower symbols lookups as well as raising security concerns of being able to load symbols not intended originally to be loadable.
The solution is to switch to `__attribute__((visibility("hidden")))` by default and mark 'exported' only those symbols that are marked with `__attribute__((visibility("default")))` explicitly.

Here, for gcc for instance, `CXX_VISIBILITY_PRESET` CMake target property will be translated into `-fvisibility=hidden` compiler option. `VISIBILITY_INLINES_HIDDEN` as an addition to the previous one will also add `-fvisibility-inlines-hidden` (helps a lot with STL code).
```
        SET_TARGET_PROPERTIES(${target} PROPERTIES
          CXX_VISIBILITY_PRESET hidden
          VISIBILITY_INLINES_HIDDEN YES
        )
```

### DIFF
--- a/cmake/component.cmake
+++ b/cmake/component.cmake
@@ -112,7 +112,29 @@ MACRO(MYSQL_ADD_COMPONENT component_arg)
         SET(COMPONENT_COMPILE_VISIBILITY
           "-fvisibility=hidden" CACHE INTERNAL
           "Use -fvisibility=hidden for components" FORCE)
-        TARGET_COMPILE_OPTIONS(${target} PRIVATE "-fvisibility=hidden")
+
+        # By default gcc / clang toolchain, when creating a shared library,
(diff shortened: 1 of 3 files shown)


percona-ysorokin (approved):
LGTM

PR #5127, opened 2023-09-20
PS-8854 [8.1]: Remove 'include/ext' pre-c++17 compatibility headers

inikep:
During the Masking Functions Component work all the other components / plugins that were using the following headers:
```
	include/ext/optional.hpp
	include/ext/string_view.hpp
```
were refactored to use `std::optional` and `std::string_view` directly so we can safely remove these files.


percona-ysorokin (approved):
LGTM

PR #5132, opened 2023-09-27
PS-8854 [8.1]: Remove support for readline library

inikep:
We can remove our support for `-DWITH_READLINE=system` because:
1. Ctrl+w works on Oracle MySQL client built with -DWITH_EDITLINE=system without any additional /.editrc changes.
2. Ctrl+w works on Oracle MySQL client built with -DWITH_EDITLINE=bundled without any additional /.editrc changes.
3. Oracle builds its packages with -DWITH_EDITLINE=bundled


azure-pipelines.yml
```diff
@@ -494,7 +494,6 @@ jobs:
           CMAKE_OPT+="
             -DWITH_MECAB=system
             -DWITH_NUMA=ON
-            -DWITH_READLINE=system
```
percona-ysorokin, line 497: May be also add `-DWITH_EDITLINE=system` to `INVERTED`?

build-ps/percona-server.spec
```diff
@@ -542,7 +542,7 @@ mkdir debug
            -DWITH_LZ4=bundled \
            -DWITH_ZLIB=bundled \
            -DWITH_ZSTD=bundled \
-           -DWITH_READLINE=system \
+           -DWITH_EDITLINE=bundled \
```
percona-ysorokin, line 545: Don't we need to explicitly specify `-DWITH_EDITLINE=bundled` for debian package specs as well?

percona-ysorokin (commented):
please also check `git diff --check` - there are some new whitespaces introduced

azure-pipelines.yml
inikep: Done but I decided that inverted will be `bundled` and normal config (that uses `-DWITH_SYSTEM_LIBS=ON`) is `system`.

build-ps/percona-server.spec
inikep: I kept `-DWITH_EDITLINE=bundled` because `-DWITH_READLINE=system` was also default option.

Another thing. Above we have `-DWITH_SYSTEM_LIBS=ON` which doesn't have much sense since it probably only sets `-DWITH_CURL=system` and `-DWITH_SSL=system` because the rest (ICU, LIBEVENT, LZ4, PROTOBUF, ZSTD, FIDO) is later changed to `bundled`.

inikep:
> please also check `git diff --check` - there are some new whitespaces introduced

I did `git checkout mysql-8.1.0 cmake/readline.cmake` so we are identical with upstream including wrong whitespaces :)

percona-ysorokin:
It looks like CirrusCI environment is missing `libedit`as well

percona-ysorokin (approved):
LGTM

percona-ysorokin:
@EvgeniyPatlan Please be aware that in 8.1 we no longer need `libreadline` to build / use Percona Server.
Although we updated `ps-build` scripts as part of this commit, Jenkins build tasks may need to be updated as well.

PR #5135, opened 2023-10-05
PS-8949 [8.0]: MyRocks configuration failure on a M1 Mac

inikep:
1. For "arm64" set:
a) `ROCKSDB_BUILD_ARCH=native+crypto` as default (support for "+crypto" is checked on runtime)
b) `ROCKSDB_BUILD_ARCH=armv8.1-a+crypto` when `ROCKSDB_DISABLE_MARCH_NATIVE==1` (for building packages)

2. Fix RocksDB issue:
```
-- Found Git: /usr/bin/git (found version "2.34.1")
/bin/sh: 1: g++: not found
```

### DIFF
--- a/cmake/compiler_features.cmake
+++ b/cmake/compiler_features.cmake
@@ -226,13 +226,60 @@ macro (ROCKSDB_SET_ARM64_DEFINTIONS)
   CHECK_VARIABLE_DEFINED(__ARM_FEATURE_CRYPTO HAVE_ARMV8_CRYPTO)
 
   IF (NOT HAVE_ARMV8_CRC OR NOT HAVE_ARMV8_CRYPTO)
-    IF (ALLOW_NO_ARMV8A_CRC_CRYPTO)
-      MESSAGE(WARNING "No ARMv8-A+crc+crypto support found and ALLOW_NO_ARMV8A_CRC_CRYPTO specified, building MyRocks but without ARMv8-A+crc+crypto/FastCRC32 support")
+    IF (ALLOW_NO_ARMV81A_CRYPTO)
+      MESSAGE(WARNING "No ARMv8.1-A+crypto support found and ALLOW_NO_ARMV81A_CRYPTO specified, building MyRocks but without ARMv8.1-A+crypto/FastCRC32 support")
     ELSE()
-      MESSAGE(FATAL_ERROR "No ARMv8-A+crc+crypto support found. Not building MyRocks. Set ALLOW_NO_ARMV8A_CRC_CRYPTO to build MyRocks with slow CRC32.")
+      MESSAGE(FATAL_ERROR "No ARMv8.1-A+crypto support found. Not building MyRocks. Set ALLOW_NO_ARMV81A_CRYPTO to build MyRocks with slow CRC32.")
     ENDIF()
   ENDIF()
 endmacro (ROCKSDB_SET_ARM64_DEFINTIONS)
+
+
+macro (ROCKSDB_SET_BUILD_ARCHITECTURE)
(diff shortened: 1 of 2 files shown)


cmake/compiler_features.cmake
```diff
@@ -226,15 +226,54 @@ macro (ROCKSDB_SET_ARM64_DEFINTIONS)
...
+      ELSE()
+        SET(ROCKSDB_BUILD_ARCH "native")
+      ENDIF()
+      MESSAGE(STATUS "MyRocks x86_64 build architecture: -march=${ROCKSDB_BUILD_ARCH}")
+    ELSE()
+      MESSAGE(STATUS "MyRocks x86_64 build architecture: -march=${ROCKSDB_BUILD_ARCH}. If the enforced processor architecture is newer than the processor architecture of this build machine, the created binary may crash if executed on this machine.")
+    ENDIF()
+
+    # CMAKE_CXX_FLAGS are used by CHECK_CXX_SOURCE_COMPILES called from ROCKSDB_SET_X86_DEFINTIONS()
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=${ROCKSDB_BUILD_ARCH}")
```
percona-ysorokin, line 254: Shouldn't we do something like
```
<push> CMAKE_CXX_FLAGS
SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=${ROCKSDB_BUILD_ARCH}")
ROCKSDB_SET_X86_DEFINTIONS()
<pop> CMAKE_CXX_FLAGS
```
instead?

cmake/compiler_features.cmake
```diff
@@ -226,15 +226,54 @@ macro (ROCKSDB_SET_ARM64_DEFINTIONS)
...
+        SET(ROCKSDB_BUILD_ARCH "native+crypto")
+      ENDIF()
+      MESSAGE(STATUS "MyRocks arm64 build architecture: -march=${ROCKSDB_BUILD_ARCH}")
+    ELSE()
+      MESSAGE(STATUS "MyRocks arm64 build architecture: -march=${ROCKSDB_BUILD_ARCH}. If the enforced processor architecture is newer than the processor architecture of this build machine, the created binary may crash if executed on this machine.")
+    ENDIF()
+
+    # CMAKE_CXX_FLAGS are used by CHECK_CXX_SOURCE_COMPILES called from ROCKSDB_SET_ARM64_DEFINTIONS()
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=${ROCKSDB_BUILD_ARCH}")
+    ROCKSDB_SET_ARM64_DEFINTIONS()
```
percona-ysorokin, line 280: Likewise

cmake/compiler_features.cmake
inikep: No, we have to build RocksDB wth `-march=X` to use all CPU extensions.
`ROCKSDB_SET_X86_DEFINTIONS` sets only `HAVE_PCLMUL`, `HAVE_SSE42`, etc. and will not work without `-mpclmul -msse4.2` that is set by `-march=X`.
`SET(CMAKE_CXX_FLAGS` has no `PARENT` attribute therefore it concerns only RocksDB.

percona-ysorokin: I see, but honestly this does not look like something that follows CMake "best practices" recommendations.
We basically rely on the fact that ROCKSDB_SET_BUILD_ARCHITECTURE() is called only from RocksDB's CMakeLists.txt while we put this macro into a general purpose top level CMake include `cmake/compiler_features.cmake`.
But this refactoring is definitely not for this PR.

percona-ysorokin (approved):
LGTM

PR #5137, opened 2023-10-12
PS-8964: Fixed hanging keyring_vault tests under ASan

oleksandr-kachan:
https://jira.percona.com/browse/PS-8964

For some reason 'curl' utility that is used to create mounts in Hashicorp Vault server hangs when run under Address Sanitizer (both from GCC and Clang).

Fixed by adding 'LD_PRELOAD= ' to the 'curl' utility invocations in the 'mount_point_service.inc'.


percona-ysorokin (approved):
LGTM

PR #5138, opened 2023-10-13
PS-8965: Fix processing of deleted keys by keyring_vault when Vault_version_v2 used

oleksandr-kachan:
https://jira.percona.com/browse/PS-8965

Vault version Vault_version_v2 keeps records history on backend. In some situations keyring_vault may get records for already deleted keys (with empty 'data' and 'deletion_time' set). At the moment keyring_vault cannot process such data records properly.

To fix the issue data records parser updated to handle such deleted entries.

### DIFF
--- a/plugin/keyring_vault/vault_parser_composer.cc
+++ b/plugin/keyring_vault/vault_parser_composer.cc
@@ -40,12 +40,14 @@
 namespace {
 const char errors_key[] = "errors";
 const char data_key[] = "data";
+const char metadata_key[] = "metadata";
 const char keys_key[] = "keys";
 const char type_key[] = "type";
 const char value_key[] = "value";
 const char max_versions_key[] = "max_versions";
 const char cas_required_key[] = "cas_required";
 const char delete_version_after_key[] = "delete_version_after";
+const char deletion_time_key[] = "deletion_time";
 
 }  // anonymous namespace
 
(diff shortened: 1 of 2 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-param-parallel-mtr/24/

plugin/keyring_vault/vault_parser_composer.cc
```diff
@@ -314,6 +316,20 @@ bool Vault_parser_composer::parse_key_data(const Secure_string &payload,
...
   if (vault_version == Vault_version_v2) {
+    it = data_node->FindMember(metadata_key);
+
+    if (it != data_node->MemberEnd() && it->value.IsObject()) {
+      const auto *metadata_node = &it->value;
+      it = metadata_node->FindMember(deletion_time_key);
+
+      if (it != metadata_node->MemberEnd() && it->value.IsString() &&
+          strlen(it->value.GetString()) != 0) {
+        logger->log(MY_ERROR_LEVEL,
```
percona-ysorokin: This could be a pretty common case. May be `MY_ERROR_LEVEL` is too much. Consider something like WARNING or INFO.

oleksandr-kachan: Done

percona-ysorokin (approved):
LGTM

PR #5148, opened 2023-11-06
PS-8671 : innodb.ddl_kill fails sporadically on Jenkins

dlenev:
https://jira.percona.com/browse/PS-8671

Sporadical failures of innodb.ddl_kill test were caused by two factors:

1) Sometimes parallel ALTER TABLE ADD INDEX implementation executed its
   BTREE_BUILD step not in thread of connection which got ALTER TABLE,
   but in one of auxiliary worker threads. In this case
   'ddl_btree_build_interrupt' debug sync point placed at the start of
   this step and used by the test case was ignored. The problem was that
   DEBUG_SYNC_C_IF_THD() macro, which were used to set this sync point,
   even though gets THD as its first argument still, also, relies on
   'current_thd' thread-local variable being set to non-0 value, which
   was not the case for auxiliary worker threads.

2) The first part of the test case didn't reset properly state of
   debug sync point facility after ALTER TABLE statement waiting on
   sync point was killed. This resulted in non-empty set of debug
   signals set when we entered second part of the test, breaking its
   logic.

This patch solves the former issue by replacing DEBUG_SYNC_C_IF_THD() macro usage by BTREE_BUILD step with simple DEBUG_SYNC() call. This macro provides equivalent functionality in this specific case, while not relying on current_thd value. The latter issue is addressed by resetting state of debug sync facility after execution of the first part of the test case.

### DIFF
--- a/mysql-test/suite/innodb/r/ddl_kill.result
+++ b/mysql-test/suite/innodb/r/ddl_kill.result
@@ -20,6 +20,9 @@ KILL QUERY @id;
 SET DEBUG_SYNC='now SIGNAL interrupt';
 # con1: reap
 ERROR 70100: Query execution was interrupted
+# Cleanup. Since ALTER TABLE was killed while waiting on sync point
+# 'interrupt' signal might have been not received and properly cleared.
+SET DEBUG_SYNC='RESET';
 # con2: Start the DDL
 SET SESSION DEBUG='+d,ddl_btree_build_interrupt';
 SET DEBUG_SYNC='ddl_btree_build_interrupt SIGNAL ready WAIT_FOR interrupt';
--- a/mysql-test/suite/innodb/t/ddl_kill.test
+++ b/mysql-test/suite/innodb/t/ddl_kill.test
@@ -39,6 +39,10 @@ connection con1;
 --error ER_QUERY_INTERRUPTED
 --reap
 
+--echo # Cleanup. Since ALTER TABLE was killed while waiting on sync point
+--echo # 'interrupt' signal might have been not received and properly cleared.
+SET DEBUG_SYNC='RESET';
+
 connect (con2,localhost,root,,);
 connection con2;
 --echo # con2: Start the DDL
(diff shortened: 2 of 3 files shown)


percona-ysorokin (approved):
LGTM

satya-bodapati:
Typically all innodb threads like purge threads, flush threads etc initialized with THD() so that debug sync can be used.

  File             Function                                 Line
0 buf0flu.cc       buf_flush_page_coordinator_thread        3206 THD *thd = create_internal_thd();
1 clone0repl.cc    periodic_write                            566 auto thd = create_internal_thd();
2 dict0stats_bg.cc dict_stats_thread                         357 THD *thd = create_internal_thd();
3 fsp0fsp.cc       fsp_init_resume_alter_encrypt_tablespace 4956 THD *thd = create_internal_thd();
4 fts0opt.cc       fts_optimize_thread                      2813 THD *thd = create_internal_thd();
5 lock0lock.cc     lock_sys_resize                           358 THD *thd = create_internal_thd();
6 log0chkp.cc      log_checkpointer                          982 ut_d(log.m_checkpointer_thd = create_internal_thd());
7 log0write.cc     log_writer                               2230 ut_d(log.m_writer_thd = create_internal_thd());
8 srv0srv.cc       srv_master_thread                        3084 THD *thd = create_internal_thd();
9 srv0srv.cc       srv_worker_thread                        3194 THD *thd = create_internal_thd();
a srv0srv.cc       srv_purge_coordinator_thread             3442 THD *thd = create_internal_thd();
b trx0roll.cc      trx_recovery_rollback_thread              851 THD *thd = create_internal_thd();

May be not a bad idea to initialize THD to these workter threads as well.

dlenev:
> Typically all innodb threads like purge threads, flush threads etc initialized with THD() so that debug sync can be used.
...
> May be not a bad idea to initialize THD to these workter threads as well.

Indeed... But these specific worker threads in ALTER TABLE seem to get by "inheriting" THD for debug_sync purposes from the connection thread which initiated ALTER TABLE... Since this problem is limited to test case only and using DEBUG_SYNC_C_ ... macro from C++ code seems plain wrong to me anyway, I would prefer to stick to current minimalistic patch.

I will contribute it to Upstream and will be happy to accept if they have a different fix for this.

dlenev:
Jenkins run result seem OK to me. See:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/352/
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/353/

satya-bodapati:
Ok with the minimal patch. Please add this suggestion in the upstream bug report

PR #5149, opened 2023-11-07
PS-8927: Create table like doesn't write the charset on binlog if its…

aybek:
… created from a temporary table

https://jira.percona.com/browse/PS-8927

When creating a table with CREATE TABLE ... LIKE ... from a temporary table, the create table isn't written entirely on the binary log, causing an error on the replica, which will create the table with the charset from the default database. Doesn't happen with persistent tables.

This patch solves the problem by forcing the CHARSET to be present into CREATE TABLE statement of the binary log, as the reference table is temporary and may not exist on replica.

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_tmp_table_and_DDL.test
+++ b/mysql-test/suite/rpl/t/rpl_tmp_table_and_DDL.test
@@ -203,4 +203,45 @@ CREATE TABLE t1 LIKE temp_t1;
 --connection master
 DROP TEMPORARY TABLE temp_t1;
 DROP TABLE t1;
+
+--sync_slave_with_master
+#
+# BUG#112364
+# CREATE TABLE LIKE <TEMP_TABLE> does not write the charset on binary log 
+# when using row based replication
+#
+--connection master
+
+# Define temp_t1 character set and collation
+--let $charset_temp_t1= latin1
+--let $collation_temp_t1= latin1_swedish_ci
+
+# Create the temporary tables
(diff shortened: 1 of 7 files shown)


mysql-test/suite/rpl/t/rpl_tmp_table_and_DDL.test
```diff
@@ -199,8 +199,50 @@ CREATE TABLE t1 LIKE temp_t1;
...
+--let $assert_cond= "$collation_t1" = "$collation_temp_t1"
+--let $assert_text= "t1 on slave and temp_t1 have the same character set"
+--source include/assert.inc
+
 # Cleanup
 --connection master
 DROP TEMPORARY TABLE temp_t1;
 DROP TABLE t1;
 --source include/rpl_end.inc
+
```
kamil-holubicki: Empty line. Please remove.

kamil-holubicki (commented):
LGTM. 
Could you paste Jenkins link?

aybek:
Jenkins run: https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-parallel-mtr/349/

kamil-holubicki (approved):

percona-ysorokin:
@aybek one more formatting issue with no new line at the end of file in `mysql-test/suite/rpl/t/rpl_tmp_table_and_DDL.test`.
Sorry
It's kinda good practice to always run `git diff --check` and `git clang-format` before the `git commit` to avoid such issues.
Please merge as soon as this is fixed.
When we merge directly into the `release-xxx` branch it's usually better to do this via `Squash and Merge` instead of `Merge pull request`.

percona-ysorokin (approved):
LGTM with minor formatting comment addressed

PR #5153, opened 2023-11-09
PS-8975 : Mysql freeze (threads waiting on __lll_lock_wait)

dlenev:
https://jira.percona.com/browse/PS-8975

When support for userstat was enabled, concurrent execution of queries to tables, including queries to I_S.TABLE/INDEX_STATISTICS and FLUSH PRIVILEGES command sometimes caused deadlocks between internal locks of Percona Server. As result statements in existing connections were stalled and it was impossible to create new connections to the server.

These deadlocks we caused by the fact that FLUSH PRIVILEGES and queries to I_S.TABLE/INDEX_STATISTICS were acquiring ACL_CACHE lock and LOCK_global_table/index_stats mutexes in different order.

This fix solves this problem by ensuring that we no longer try to acquire ACL_CACHE lock while holding LOCK_global_table/index_stats mutexes. To do this we have changed implementation of I_S.TABLE/ INDEX_STATISTICS to copy statistics records from global container to MEM_ROOT-allocated array under protection of LOCK_global_table/ index_stats mutexes first, and only the check if user has necessary privileges to see copied entries ((which can involve acquisition of ACL_CACHE lock) outside of mutex scope.

We also no longer store information about table engine type in global table statistics container, as this information is not used anywhere.

### DIFF
--- a/mysql-test/r/percona_userstat.result
+++ b/mysql-test/r/percona_userstat.result
@@ -141,4 +141,19 @@ COMMIT_TRANSACTIONS	ROLLBACK_TRANSACTIONS
 17	1
 SET autocommit=1;
 DROP TABLE t1;
+#
+# Bug PS-8975: Mysql freeze (threads waiting on __lll_lock_wait)
+#
+CREATE DATABASE mysqltest1;
+CREATE TABLE mysqltest1.t1 (i INT PRIMARY KEY AUTO_INCREMENT);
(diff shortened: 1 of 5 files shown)


percona-ysorokin (approved):
LGTM

dlenev:
Jenkins result LGTM: https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/354/

PR #5157, opened 2023-11-13
PS-8836 Follow-up for: MyRocks serialization violation? Read after wr…

dlenev:
…ite is failing (#5120)

https://jira.percona.com/browse/PS-8836

This is a cherry-pick of the
"Fix RC SK concurrency testcase for point updates" (commit facebook/mysql-5.6@3c1f09a)

Summary: Without row_key in the SELECT .. FOR UPDATE lookup, the sk_rc_concurrent_point_update.test behaved like an iterator-based scenario, which is covered by another test case. Also, the hidden primary key scenario also doesn't make sense for the point update case.

Differential Revision: D49566243

fbshipit-source-id: 86bec32da3823ee8625ca7e4bfc1ff8dab8b58fd

Authored-by: Saumitr Pathak <saumitr@meta.com>


percona-ysorokin (approved):
LGTM

PR #5158, opened 2023-11-14
PS-8725 : Assertion `m_opened_table != nullptr' failed in temptable::…

dlenev:
…Handler::opened_table_validate()

Test case only.

Original problem was fixed in MySQL Server 8.0.35 by the commit a1d235e83859ed39610b2d39eb1a7642964de72c (Bug#35105404 "Server crashes with signal 11 and asserts in temptable code."). However, this commit doesn't contain a public test case.

There was a GitHub contribution with a fix and test case by Nicholas Othieno, but we have chosen upstream's version of fix and a somewhat different test case.

### DIFF
--- a/mysql-test/r/derived.result
+++ b/mysql-test/r/derived.result
@@ -4734,3 +4734,20 @@ Cat2	5
 Cat2	7
 NULL	NULL
 DROP TABLE t1;
+#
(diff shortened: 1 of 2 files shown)


percona-ysorokin (approved):
LGTM

PR #5159, opened 2023-11-15
(5.7) PS-8879: OOM when alter column to compression

kamil-holubicki:

Problem:
If the large table containing BLOB column is altered in the way that the COLUMN_FORMAT becomes COMPRESSED
(like ALTER TABLE t1 MODIFY a10 BLOB COLUMN_FORMAT COMPRESSED), we observe huge memory utilization during the ALTER. This can lead to OOM situation when mysqld gets killed by the system.

Cause:
Fix for PS-7657 (a943a8fe) removed emptying of compress_heap from before insertion of the row into InnoDB. This was because during the UPDATE, prior to storing the row, the compressed column is read from InnoDB and decompressed. Its decompressed version is kept on prebuilt->compress_heap. Then when we store the row, we need the compress_heap to be still valid to have access to the BLOB column value.
However, if we don't clean the compress_heap, it constantly grows in cases like ALTER ... COLUMN_FORMAT COMPRESSED, because during this process we copy all data from altered table to temp table in a loop.

Solution:
Rework the original fix in the way that compress_heap is emptied after the row is inserted.

### DIFF
--- a/storage/innobase/row/row0mysql.cc
+++ b/storage/innobase/row/row0mysql.cc
@@ -2442,6 +2442,19 @@ row_insert_for_mysql(
 	const byte*		mysql_rec,
 	row_prebuilt_t*		prebuilt)
 {
+	struct CompressHeapCleaner {
+		CompressHeapCleaner(row_prebuilt_t* prebuilt)
+		: prebuilt_(prebuilt) { }
+
+		~CompressHeapCleaner() {
+			if (prebuilt_ && UNIV_LIKELY_NULL(prebuilt_->compress_heap)) {
+				mem_heap_empty(prebuilt_->compress_heap);
+			}
+		}
+
+		row_prebuilt_t *prebuilt_;
+	} compressCheapCleaner(prebuilt);
+
 	/* For intrinsic tables there a lot of restrictions that can be
 	relaxed including locking of table, transaction handling, etc.
 	Use direct cursor interface for inserting to intrinsic tables. */
--- a/storage/innobase/row/row0sel.cc
+++ b/storage/innobase/row/row0sel.cc
@@ -3318,8 +3318,9 @@ row_sel_store_mysql_rec(
 		row_mysql_prebuilt_free_blob_heap(prebuilt);
 	}
 
-	if (UNIV_LIKELY_NULL(prebuilt->compress_heap))
-		row_mysql_prebuilt_free_compress_heap(prebuilt);
+	if (UNIV_LIKELY_NULL(prebuilt->compress_heap)) {
+		mem_heap_empty(prebuilt->compress_heap);
+	}
 
 	if (clust_templ_for_sec) {
 		/* Store all clustered index column of


kamil-holubicki:
https://ps57.cd.percona.com/view/5.7/job/percona-server-5.7-pipeline/34862/

percona-ysorokin (approved):
LGTM

storage/innobase/row/row0sel.cc
```diff
@@ -3318,8 +3318,9 @@ row_sel_store_mysql_rec(
 		row_mysql_prebuilt_free_blob_heap(prebuilt);
 	}
 
-	if (UNIV_LIKELY_NULL(prebuilt->compress_heap))
-		row_mysql_prebuilt_free_compress_heap(prebuilt);
+	if (UNIV_LIKELY_NULL(prebuilt->compress_heap)) {
```
satya-bodapati, line 3321: why can't we use row_mysql_prebuilt_free_compress_heap()? heap empty also would free all the blocks except the first one.

storage/innobase/row/row0mysql.cc
```diff
@@ -2442,6 +2442,19 @@ row_insert_for_mysql(
 	const byte*		mysql_rec,
 	row_prebuilt_t*		prebuilt)
 {
+	struct CompressHeapCleaner {
```
satya-bodapati, line 2445: We followed the same style as prebuilt->blob_heap and I don't understand why we need a special treat for compress heap.

satya-bodapati (commented):
testcase missing.

storage/innobase/row/row0sel.cc
kamil-holubicki: The original way is using empty (before commit 59cb3d5f7da8e88d431f21e93ea60c094efdeb1b)
We can use `row_mysql_prebuilt_free_compress_heap`, it will work as well, sure. But if we are looping over rows, it will cause freeing all blocks and allocating again. If we empty instead of free, it is a good chance that already allocated 1st block is big enough, so we will avoid allocation.

satya-bodapati: I see, an optimization.  

Can you please execute the test case from PS-7657? with and without optimization?

kamil-holubicki: I did it locally. Both versions work fine.

storage/innobase/row/row0mysql.cc
kamil-holubicki: There are two aspects here:
1. updating the row. Here the comments from Nitendra are valid (commit 59cb3d5f7da8e88d431f21e93ea60c094efdeb1b). In simple, we cannot free the uncompressed BLOB before storing it back. So the heap free should be done after storing the row (but that fix removed free completely)
2. Altering table (this ticket case). In this case, we are internally copying data between two tables row by row. As in 1 we removed heap free, we've got OOM.

So to satisfy 1 and 2 we have to free compression heap after row insertion

kamil-holubicki:
> testcase missing.

We've got original testcases from 59cb3d5f7da8e88d431f21e93ea60c094efdeb1b related to UPDATE scenario
xtradb_compressed_columns.test
rpl_update_partitioned_table_with_compressed_column.test

PR #5165, opened 2023-11-28
[WIP] PS FIPS support

oleksandr-kachan:
- Added checkup during server startup if system OpenSSL is configured to use FIPS. Corresponding message will be printed to server error log.
- Added --check-ssl-fips-enabled=OFF|FORCE system variable.
- In case --check-ssl-fips-enabled=FORCE and system OpenSSL is not configured to use FIPS server startup is cancelled.


oleksandr-kachan:
Not actual any more, FIPS support added in a bit different way.

PR #5188, opened 2023-12-19
PS-9039: Reduce combinations used for keyring_vault tests

oleksandr-kachan:
https://jira.percona.com/browse/PS-9039

Run only one keyring_vault.install_keyring_vault test with all 12 combinations. Remaining tests run only with mount-v1 and mount-v2 combinations.


percona-ysorokin (approved):
LGTM

PR #5189, opened 2023-12-20
PS-9041: Fix leaks in testSecureSocket test

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9041


percona-ysorokin (approved):
LGTM

PR #5199, opened 2024-01-08
PS-9057: Fix build with default value used for -DWITH_CURL

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9057

Cmake fails in case -DWITH_CURL is not set directly. In this case it defaults to 'none' but curl is required to build component_keyring_vault.

Change WITH_CURL to be set to 'system' by default.


percona-ysorokin (approved):
LGTM

PR #5200, opened 2024-01-09
PS-9048: Exclude % symbol from ngram generated index tokens

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9048

There is an assertion failure during optimize table execution in case string in indexed column contains % and ngram parser is used. The % is treated in specal way and not expected in index tokens. Improved ngram parser not to include % into parsed tokens.

### DIFF
--- a/mysql-test/suite/innodb_fts/include/percona_ps9048.inc
+++ b/mysql-test/suite/innodb_fts/include/percona_ps9048.inc
@@ -1,0 +1,12 @@
+SET @old_innodb_optimize_fulltext_only = @@innodb_optimize_fulltext_only;
+SET GLOBAL innodb_optimize_fulltext_only = ON;
+
+--eval CREATE TABLE t1 (c TINYTEXT, FULLTEXT KEY (c) $PARSER) ENGINE=InnoDB $CHARSET
+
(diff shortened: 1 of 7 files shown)


satya-bodapati:
@oleksandr-kachan what about the already existing tokens that are generated before the fix?

1. create a data-dir that didn't exclude % with ngram parser
2. upgrade to a server that has this fix

Will it be able to handle the already tokenized words with %?

oleksandr-kachan:
> @oleksandr-kachan what about the already existing tokens that are generated before the fix?
> 1. create a data-dir that didn't exclude % with ngram parser
> 2. upgrade to a server that has this fix
> Will it be able to handle the already tokenized words with %?

@satya-bodapati, no, this fix will not handle it. Need to think what to do with such tokens.

satya-bodapati:
> > @oleksandr-kachan what about the already existing tokens that are generated before the fix?
> > 
> > 1. create a data-dir that didn't exclude % with ngram parser
> > 2. upgrade to a server that has this fix
> >    Will it be able to handle the already tokenized words with %?
> 
> @satya-bodapati, no, this fix will not handle it. Need to think what to do with such tokens.

We have to first check if your fix affects query results.

oleksandr-kachan:
Was fixed in scope of different PR

PR #5201, opened 2024-01-15
PS-9060 Non-pro deps for percona-server-rocksdb-pro on yum

adivinho:


EvgeniyPatlan (approved):

percona-ysorokin:
Are we merging this into "8.0" trunk or this should be done in `release-8.0.35-28`?

inikep:
We need all commits in `8.0` but the proper way should be to merge this PR into `release-8.0.35-27` and then merge again `release-8.0.35-27` to `8.0`.

PR #5205, opened 2024-01-16
PS-9067 Fix MTR test failures when run with --mem

VarunNagaraju:
https://perconadev.atlassian.net/browse/PS-9067

When --mem is used to run MTR tests, thy are run in memory using tmpfs. This causes inconsistencies when recodring test results, since the file paths are not deterministic. So, replace_regex is used to mask the variable paths percona_statement_set.test and bug84640.test.
For percona_changed_page_bmp_flush.test, the generated warning is suppressed.


percona-ysorokin (approved):
LGTM

percona-ysorokin:
@VarunNagaraju Please also check if the problem exist on 8.0 / trunk and port there if necessary

VarunNagaraju:
> @VarunNagaraju Please also check if the problem exist on 8.0 / trunk and port there if necessary

percona_statement_set.test and percona_changed_page_bmp_flush.test doesn't exist in 8.0.
bug84640.test is disabled in 8.0.

PR #5207, opened 2024-01-18
PS-9075: ALTER TABLE ... ALGORITHM=INPLACE fails for table which is n…

kamil-holubicki:
…o longer encrypted

https://perconadev.atlassian.net/browse/PS-9075

Problem:
If we create a table in encrypted tablespace and then remove encryption from this tablespace, the table is rebuilt with encryption disabled. Then if we try to alter this table using INPLACE algorithm, the query fails, complaining that altering encryption attribute is not supported for INPLACE.

Cause:
Commit a478c987 introduced the fix for
DICT_TF2_ENCRYPTION_FILE_PER_TABLE flag setting/checking. Later on, checks have been removed, however the part of the fix in ha_innodb.cc remained.
With the conjunction of the logic from
ha_innobase::check_if_supported_inplace_alter() which used this flag (m_prebuilt->table->flags2), the decision was made that INPLACE is not allowed.

Solution:
Upstream logic of ha_innobase::check_if_supported_inplace_alter() does not rely on m_prebuilt->table->flags2, but relies on TABLE_SHARE::encrypt_type. It is fine, as flags2 in table are not set to DICT_TF2_ENCRYPTION_FILE_PER_TABLE.
However PS logic still used flags2.

1. Fully reverted above-mentioned fix done in a478c987
2. Changed the logic of ha_innobase::check_if_supported_inplace_alter() to be aligned with upstream (use TABLE_SHARE::encrypt_type)


kamil-holubicki:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/433/

satya-bodapati (approved):
LGTM!

Please move this test to percona suite in innovation series. I think you have to merge the fix

percona-ysorokin (commented):
LGTM

PR #5208, opened 2024-01-19
(trunk) PS-9075: ALTER TABLE ... ALGORITHM=INPLACE fails for table which is no longer encrypted

kamil-holubicki:
Merge 8.0 fix into the trunk according to the GCA process.


satya-bodapati (approved):

percona-ysorokin (approved):
LGTM

PR #5219, opened 2024-01-31
PS-9083: server crash when slow query log is on with additional varia…

dlenev:
…bles

https://perconadev.atlassian.net/browse/PS-9083

When slow log was enabled with --long_query_time option set to 0 (or close value) and with --log_slow_verbosity option including "query_info" mode, server might have crashed when a protocol command other than normal statement execution (COM_QUERY or mysql_query() in C API parlance) was executed in a connection which earlier was used to rn some statement. For example, this sometimes happened for COM_QUIT command issue during graceful disconnect.

The crash was caused by attempt to write to slow log list of query tables for such non-COM_QUERY command. To do so we tried to iterate through LEX::query_tables list, which in case of such commands might contain garbage left from execution of earlier statements in the connection.

This fix solves the problem by avoiding writing to slow log list of query tables for server commands other than COM_QUERY (for most of them it makes no sense  or not easily available).

### DIFF
--- a/mysql-test/r/percona_slowlog_query_info.result
+++ b/mysql-test/r/percona_slowlog_query_info.result
@@ -1,4 +1,3 @@
-SET @saved_slow_query_log = @@GLOBAL.slow_query_log;
 SET SESSION log_slow_verbosity = 'query_info';
 SET GLOBAL slow_query_log = 1;
 SET long_query_time = 0;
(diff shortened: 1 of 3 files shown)


dlenev:
Jenkins results for this patch look OK to me: https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/452/

percona-ysorokin (approved):
LGTM

PR #5243, opened 2024-03-01
Percona telemetry

kamil-holubicki:


components/percona_telemetry/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License, version 2.0, for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA
+
+DISABLE_MISSING_PROFILE_WARNING()
+
+IF(WITH_PERCONA_TELEMETRY)
```
percona-ysorokin, line 20: Declare `WITH_PERCONA_TELEMETRY` as a CMake option
```
option(WITH_PERCONA_TELEMETRY "Build Percona Telemetry component" ON)
```

components/percona_telemetry/common.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
```
percona-ysorokin: 1. `#pragma once` is not a standard directive
2. Based on unconfirmed data it confuses `ccache` which we use a lot.

This is definitely a question of personal preferences, but my advice is to never use it - adding 3 lines of standard header guards is not that big overhead.

components/percona_telemetry/component.cc
```diff
@@ -0,0 +1,97 @@
...
+                                command_error_info_srv);
+REQUIRES_SERVICE_PLACEHOLDER_AS(registry, registry_srv);
+REQUIRES_SERVICE_PLACEHOLDER_AS(log_builtins, log_builtins_srv);
+REQUIRES_SERVICE_PLACEHOLDER_AS(log_builtins_string, log_builtins_string_srv);
+REQUIRES_SERVICE_PLACEHOLDER_AS(component_sys_variable_register,
+                                component_sys_variable_register_srv);
+REQUIRES_SERVICE_PLACEHOLDER_AS(component_sys_variable_unregister,
+                                component_sys_variable_unregister_srv);
+
+static std::unique_ptr<PerconaTelemetryComponent> percona_telemetry_component;
```
percona-ysorokin: anonymous namespace can be used for a group of static declarations
```
namespace {
...
} // anonymous namespace
```

components/percona_telemetry/component.cc
```diff
@@ -0,0 +1,97 @@
...
+  services->command_options_service = command_options_srv;
+  services->command_query_service = command_query_srv;
+  services->command_query_result_service = command_query_result_srv;
+  services->command_field_info_service = command_field_info_srv;
+  services->log_builtins_service = log_builtins_srv;
+  services->log_builtins_string = log_builtins_string_srv;
+  services->var_register_service = component_sys_variable_register_srv;
+  services->var_unregister_service = component_sys_variable_unregister_srv;
+
+  percona_telemetry_component =
```
percona-ysorokin, line 67: probably just
```
percona_telemetry_component = std::move(services);
```

components/percona_telemetry/component.cc
```diff
@@ -0,0 +1,97 @@
...
+BEGIN_COMPONENT_METADATA(CURRENT_COMPONENT_NAME)
+METADATA("mysql.author", "Percona Corporation"),
+    METADATA("mysql.license", "GPL"), METADATA("mysql.version", "1"),
+    END_COMPONENT_METADATA();
+
+DECLARE_COMPONENT(CURRENT_COMPONENT_NAME, CURRENT_COMPONENT_NAME_STR)
+component_init, component_deinit END_DECLARE_COMPONENT();
+
+DECLARE_LIBRARY_COMPONENTS &COMPONENT_REF(CURRENT_COMPONENT_NAME)
+    END_DECLARE_LIBRARY_COMPONENTS
```
percona-ysorokin: No EOL at the end of file

components/percona_telemetry/config.cc
```diff
@@ -0,0 +1,126 @@
+#include "config.h"
+#include "common.h"
+
+#define VAR_TELEMETRY_ROOT_DIR "telemetry_root_dir"
```
percona-ysorokin: Here and everywhere, prefer `static constexpr ...` declarations instead of macros
For strings, use `std::string_view`
```
static constexpr std::string_view string_constant[] = "xxx";
```
unless you need the string to be null terminated, in this case 
```
static constexpr char string_constant[] = "xxx";
```

components/percona_telemetry/config.cc
```diff
@@ -0,0 +1,126 @@
...
+
+bool Config::init() {
+  STR_CHECK_ARG(str) telemetry_root_dir_arg;
+  telemetry_root_dir_arg.def_val =
+      const_cast<char *>(TELEMETRY_ROOT_DIR_DEFAULT);
+  if (var_register_service_.register_variable(
+          CURRENT_COMPONENT_NAME_STR, VAR_TELEMETRY_ROOT_DIR,
+          PLUGIN_VAR_STR | PLUGIN_VAR_MEMALLOC | PLUGIN_VAR_READONLY,
+          "Root path of the telemetry data for all mysqld servers", nullptr,
+          nullptr, (void *)&telemetry_root_dir_arg,
```
percona-ysorokin: Here and everywhere, every pointer can be implicitly casted to `void *` / `const void *`. So, probably unnecessary `(void*)` cast unless it is needed to mess with constness.

components/percona_telemetry/config.cc
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  return false;
+}
+
+bool Config::deinit() {
+  bool res = false;
+  if (var_unregister_service_.unregister_variable(CURRENT_COMPONENT_NAME_STR,
+                                                  VAR_TELEMETRY_ROOT_DIR)) {
+    res |= true;
```
percona-ysorokin: bool type does not have logical OR operator defined for it (operator |). This line works because of 2 implicit conversions (to int and back).
May be just `res = true`;

components/percona_telemetry/config.cc
```diff
@@ -0,0 +1,126 @@
...
+                                                  VAR_HISTORY_KEEP_INTERVAL)) {
+    res |= true;
+  }
+
+  return res;
+}
+
+const std::string &Config::telemetry_storage_dir_path() const {
+  // it is read-only value
+  static std::string telemetry_root_dir_value_str(telemetry_root_dir_value_);
```
percona-ysorokin, line 127: Check/ assert telemetry_root_dir_value_ for nullness here. Creating a std::string from nullptr is UB.

components/percona_telemetry/config.h
```diff
@@ -0,0 +1,35 @@
...
+
+ private:
+  SERVICE_TYPE(component_sys_variable_register) & var_register_service_;
+  SERVICE_TYPE(component_sys_variable_unregister) & var_unregister_service_;
+
+  char *telemetry_root_dir_value_;
+  int scrape_interval_value_;
+  int grace_interval_value_;
+  int history_keep_interval_value_;
+};
```
percona-ysorokin: No EOL at the end of file

components/percona_telemetry/config.h
```diff
@@ -0,0 +1,35 @@
...
+  int grace_interval() const;
+  int history_keep_interval() const;
+  int unconditional_history_cleanup_interval() const;
+
+ private:
+  SERVICE_TYPE(component_sys_variable_register) & var_register_service_;
+  SERVICE_TYPE(component_sys_variable_unregister) & var_unregister_service_;
+
+  char *telemetry_root_dir_value_;
+  int scrape_interval_value_;
```
percona-ysorokin: I doubt these xxx_interval values could ever be negative, so std::uint32_t (or similar)

components/percona_telemetry/config.h
```diff
@@ -0,0 +1,35 @@
...
+#include <mysql/components/services/component_sys_var_service.h>
+
+class Config {
+ public:
+  Config(SERVICE_TYPE(component_sys_variable_register) & var_register_service,
+         SERVICE_TYPE(component_sys_variable_unregister) &
+             var_unregister_service);
+  ~Config() = default;
+
+  Config(const Config &rhs) = delete;
```
percona-ysorokin, line 34: You should delete other spacial member functions as well.
```
 Config(Config &&rhs) = delete;
 Config& operator (const Config &rhs) = delete;
 Config& operator (Config &&rhs) = delete;
```

components/percona_telemetry/config.h
```diff
@@ -0,0 +1,35 @@
...
+
+  bool init();
+  bool deinit();
+
+  const std::string &telemetry_storage_dir_path() const;
```
percona-ysorokin: Here and everywhere, consider using `noexcept` for simple methods like these.

components/percona_telemetry/data_provider.cc
```diff
@@ -0,0 +1,317 @@
...
+bool DataProvider::do_query(const std::string &query, QueryResult *result) {
+  MYSQL_RES_H mysql_res = nullptr;
+  MYSQL_ROW_H row = nullptr;
+  uint64_t row_count = 0;
+  unsigned int num_column = 0;
+  std::string result_set;
+  MYSQL_H mysql_h = nullptr;
+  bool res = true;
+
+  if (command_factory_service_.init(&mysql_h)) {
```
percona-ysorokin: May be, use some RAII guard like std::unique_ptr with custom deleted for both mysql_h and mysql_res?

components/percona_telemetry/data_provider.cc
```diff
@@ -0,0 +1,317 @@
...
+}
+
+const std::string &DataProvider::get_database_instance_id() {
+  if (!database_instance_id_cache_.length()) {
+    QueryResult result;
+    if (do_query("SELECT @@server_uuid", &result)) {
+      static std::string empty;
+      return empty;
+    }
+    database_instance_id_cache_ = result[0][0];
```
percona-ysorokin, line 202: This could be a race condition if get_database_instance_id() is called from multiple threads.

components/percona_telemetry/data_provider.cc
```diff
@@ -0,0 +1,317 @@
...
+}
+bool DataProvider::collect_product_version_info(rapidjson::Document *document) {
+  // Version doesn't change during the lifetime, so query and cache it
+  if (version_cache_.empty()) {
+    QueryResult result;
+    if (do_query("SELECT @@VERSION, @@VERSION_COMMENT", &result)) {
+      return true;
+    }
+
+    version_cache_ = result[0][0];
```
percona-ysorokin, line 235: version_cache_ could also be damaged by race conditions

components/percona_telemetry/data_provider.cc
```diff
@@ -0,0 +1,317 @@
...
+}
+
+/* Note that this metric is update very X, so it may be inacurate.
+   We could make it accurate but that would need ANALYZE TABLE for every table
+   which would be overkill.*/
+bool DataProvider::collect_dbs_size_info(rapidjson::Document *document) {
+  QueryResult result;
+  if (do_query("SELECT IFNULL(ROUND(SUM(data_length + index_length), 1), '0') "
+               "size_MB FROM information_schema.tables WHERE table_schema NOT "
+               "IN('mysql', 'information_schema', 'performance_schema', 'sys')",
```
percona-ysorokin, line 332: this list of special databased `'mysql', 'information_schema', 'performance_schema', 'sys'` is repeated several times. May be a constant?

components/percona_telemetry/data_provider.cc
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  rapidjson::StringBuffer string_buffer;
+  string_buffer.Clear();
+  RapidJsonWritterType string_writer(string_buffer);
+  JSON.Accept(string_writer);
+  std::string json(string_buffer.GetString(), string_buffer.GetSize());
+
+  return json;
+}
```
percona-ysorokin: No EOL at the end of file

components/percona_telemetry/data_provider.h
```diff
@@ -0,0 +1,74 @@
...
+#include <mysql/components/services/security_context.h>
+
+//>> ugly part start
+#ifdef RAPIDJSON_NO_SIZETYPEDEFINE
+#include "my_rapidjson_size_t.h"
+#endif
+
+#include <rapidjson/document.h>
+
+#if 0
```
percona-ysorokin: May be a define with meaningful name? COMPONENT_PERCONA_TELEMETRY_DEBUG_JSON_PRINTER or something

components/percona_telemetry/data_provider.h
```diff
@@ -0,0 +1,74 @@
...
+  SERVICE_TYPE(mysql_command_query) & command_query_service_;
+  SERVICE_TYPE(mysql_command_query_result) & command_query_result_service_;
+  SERVICE_TYPE(mysql_command_field_info) & command_field_info_service_;
+  SERVICE_TYPE(mysql_command_thread) & command_thread_service_;
+
+  Logger &logger_;
+
+  std::string database_instance_id_cache_;
+  std::string version_cache_;
+};
```
percona-ysorokin: No EOL at the end of file

components/percona_telemetry/logger.cc
```diff
@@ -0,0 +1,52 @@
...
+Logger::Logger(SERVICE_TYPE(log_builtins) & log_builtins_service,
+               SERVICE_TYPE(log_builtins_string) & log_builtins_string_service,
+               loglevel log_level)
+    : log_level_(log_level) {
+  log_bi = &log_builtins_service;
+  log_bs = &log_builtins_string_service;
+}
+
+static void log(loglevel level, const char *format, va_list *args) {
+  constexpr int buffer_len = 8 * 1024;
```
percona-ysorokin: std::size_t

components/percona_telemetry/logger.cc
```diff
@@ -0,0 +1,52 @@
...
+#include "logger.h"
+
+SERVICE_TYPE(log_builtins) * log_bi;
+SERVICE_TYPE(log_builtins_string) * log_bs;
+
+#define RETURN_IF_BLOCKED(level) \
+  if (level > log_level_) return;
+
+static void log(loglevel level, const char *format, va_list *args)
+    __attribute__((format(printf, 2, 0)));
```
percona-ysorokin: Check if there is a MY_XXX macro which would result in compiler specific attribute declaration .

components/percona_telemetry/storage.cc
```diff
@@ -0,0 +1,135 @@
...
+      bool our_file = (file_uuid == uuid_);
+
+      try {
+        long file_timestamp = std::stol(file_timestamp_str);
+        // We need to delete the file in 2 cases:
+        // 1. This is our file and it is older than this instance's
+        // history_keep_interval
+        // 2. This is someone's else file, but it is older than the highest
+        // possible history_keep_interval
+        if (our_file && (file_timestamp < oldest_timestamp)) {
```
percona-ysorokin, line 90: Consider using std::chrono::time_point and std::chrono::duration for time comparison

components/percona_telemetry/storage.cc
```diff
@@ -0,0 +1,135 @@
...
+  }
+}
+
+bool Storage::store_report(const std::string &report) {
+  try {
+    long current_time = (long)std::time(0);
+
+    clean_old_reports(current_time);
+
+    std::string filename = config_.telemetry_storage_dir_path() + "/" +
```
percona-ysorokin: May be use std::filesystem::path operator / instead

components/percona_telemetry/storage.cc
```diff
@@ -0,0 +1,135 @@
...
+                           std::to_string(current_time) + "-" + uuid_;
+    std::string tmp_filename = filename + TMP_FILE_EXTENSION;
+    std::string json_filename = filename + JSON_FILE_EXTENSION;
+    std::ofstream file_stream(tmp_filename);
+    if (!file_stream.is_open()) throw file_stream_exception(strerror(errno));
```
percona-ysorokin, line 147: You can call
file_stream.exceptions(std::ifstream::failbit);
and every failed stream operation will throw automatically.

components/percona_telemetry/worker.cc
```diff
@@ -0,0 +1,91 @@
...
+        cv_.wait_for(lock, std::chrono::seconds(grace_interval))) {
+      // the thread was signaled to exit
+      return;
+    }
+  }
+
+  data_provider_.thread_access_begin();
+
+  // TODO:
+  // The server is probably still initializing if the grace period is low, so
```
percona-ysorokin: This sounds dangerous

mysql-test/suite/innodb/t/tablespace_encrypt_9.test
```diff
@@ -149,7 +149,8 @@ CREATE TABLESPACE temp_ts ADD DATAFILE 'temp_ts.ibd';
 --echo # connection con3
 connect (con3,localhost,root,,);
 connection con3;
-SELECT connection_id();
+# KH: Do we really need this in test's result?
+# SELECT connection_id();
```
percona-ysorokin: I would say "no"
"connection_id()"values should not be recorded

mysql-test/suite/innodb/t/log_first_rec_group.test
```diff
@@ -80,6 +80,9 @@ while ($pass != 2) {
     --echo LSN: $end_lsn
   }
 
+--echo "start_lsn: $start_lsn"
```
percona-ysorokin: This most probably will make the test fail when it is run with `--repeat=2`

components/percona_telemetry/CMakeLists.txt
kamil-holubicki: done

components/percona_telemetry/common.h
kamil-holubicki: done

components/percona_telemetry/component.cc
kamil-holubicki: what is the advantage?

components/percona_telemetry/component.cc
kamil-holubicki: No, we are constructing object of `PerconaTelemetryComponent` class, which has `services` as the constructor parameter

components/percona_telemetry/component.cc
kamil-holubicki: I've adjusted all the files. My understanding is that from the compiler's point of view, it is not needed. It is rather for other tools like diff, cat, etc, right?

components/percona_telemetry/config.cc
kamil-holubicki: done

components/percona_telemetry/config.cc
kamil-holubicki: Right. It was copy-paste from another component. I see all components do the same. Anyway, changed as not needed indeed.

components/percona_telemetry/config.cc
kamil-holubicki: done

components/percona_telemetry/config.h
kamil-holubicki: done

components/percona_telemetry/config.cc
kamil-holubicki: It should not happen, as it is resolved to the default or explicitly configured value during variable registration which happens in init()
I've added assert and handled init() error, so in case variable registration fails, the component initialization fails.

components/percona_telemetry/config.h
kamil-holubicki: done

components/percona_telemetry/config.h
kamil-holubicki: done

components/percona_telemetry/data_provider.cc
kamil-holubicki: I'm not sure if it is an overkill, but let's try this way... ;)

components/percona_telemetry/data_provider.cc
kamil-holubicki: Only one thread. These objects are not thread safe

components/percona_telemetry/data_provider.cc
kamil-holubicki: see above. Only one thread here

components/percona_telemetry/data_provider.cc
kamil-holubicki: There are 2 places. If we extract it to the constant, we will need to construct the query separately. And that will lead to the conclusion that we need to move the whole query to the constant and replace the code with the constant, like (`do_query(collect_dbs_info_query)`). But the query itself is coupled with parsing of the query result, so in my opinion, it will not make it any easier/cleaner.

components/percona_telemetry/data_provider.cc
kamil-holubicki: done

components/percona_telemetry/data_provider.h
kamil-holubicki: done

components/percona_telemetry/data_provider.h
kamil-holubicki: done

components/percona_telemetry/logger.cc
kamil-holubicki: done

components/percona_telemetry/logger.cc
kamil-holubicki: MY_ATTRIBUTE. done

components/percona_telemetry/storage.cc
kamil-holubicki: We compare time now (std::time(0)) (long) with the timestamp recovered from the filename (long), so it is the shortest path that avoids too much conversions.
We also create the file timestamp from std::time(0), which is long. In other way we would need to do
```
const auto now = std::chrono::system_clock::now();
const std::time_t t_c = std::chrono::system_clock::to_time_t(now);
```
which is actually what std::time(0) does

components/percona_telemetry/storage.cc
kamil-holubicki: done

components/percona_telemetry/storage.cc
kamil-holubicki: yup, but we will need to check if it is open (constructor failure) anyway. There is no constructor that allows specifying these flags. So the constructor may fail itself and we ned this `if`. Then we could set flag like 
`file_stream.exceptions(std::ifstream::failbit | std::ifstream::badbit);`
and use it only for the next call `file_stream.write()`. 
We need custom `file_stream_exception` for other cases (fs::rename() is noexcept), so handling `file_stream.write()` in a common way seems to be reasonable.

components/percona_telemetry/worker.cc
kamil-holubicki: Yes, I need to think more about it. In fact this will be executed after the grace period which is 24 hours, so everything should be initialized, but of course we need to handle this corner case.

mysql-test/suite/innodb/t/log_first_rec_group.test
kamil-holubicki: removed

mysql-test/suite/innodb/t/tablespace_encrypt_9.test
kamil-holubicki: removed.

components/percona_telemetry/config.h
kamil-holubicki: `noexcept` is rather a whish or developer's declaration that it does not throw than a real check (would it be even possible), so for simple methods... yes :)

components/percona_telemetry/worker.cc
kamil-holubicki: Implemented graceful handling of this situation.

components/percona_telemetry/component.cc
percona-ysorokin: Just stylistic.
Instead of
```
static int a;
static int b;
...
```
you can write
```
namespace {
  int a;
  int b;
  ...
}
```
Basically, you don't have to put `static` keyword on each declaration.

components/percona_telemetry/component.cc
percona-ysorokin: Sorry, my mistake

components/percona_telemetry/config.cc
```diff
@@ -0,0 +1,124 @@
+#include "config.h"
+#include <assert.h>
```
percona-ysorokin:
```
#include <cassert> 
```
is a proper c++ header.
Also try to group headers by type: system first, user ones second.

components/percona_telemetry/config.cc
```diff
@@ -0,0 +1,124 @@
...
+
+static constexpr char VAR_TELEMETRY_ROOT_DIR[] = "telemetry_root_dir";
+static constexpr char VAR_SCRAPE_INTERVAL[] = "scrape_interval";
+static constexpr char VAR_GRACE_INTERVAL[] = "grace_interval";
+static constexpr char VAR_HISTORY_KEEP_INTERVAL[] = "history_keep_interval";
+
+static constexpr char TELEMETRY_ROOT_DIR_DEFAULT[] =
+    "/usr/local/percona/telemetry/ps";
+
+static constexpr int SCRAPE_INTERVAL_DEFAULT = 60 * 60 * 24 * 1;  // 1 day
```
percona-ysorokin: Any particular reason these constants are defines as signed?

components/percona_telemetry/config.h
```diff
@@ -0,0 +1,41 @@
...
+  Config(const Config &rhs) = delete;
+  Config(Config &&rhs) = delete;
+  Config &operator=(const Config &rhs) = delete;
+  Config &operator=(Config &&rhs) = delete;
+
+  bool init();
+  bool deinit();
+
+  const std::string &telemetry_storage_dir_path() const noexcept;
+  int scrape_interval() const noexcept;
```
percona-ysorokin: members declared as unsigned but these functions return signed values.

components/percona_telemetry/data_provider.cc
```diff
@@ -0,0 +1,511 @@
+#include "data_provider.h"
+#include <mysqld_error.h>
+#include <sstream>
```
percona-ysorokin, line 18: Likewise, order of system / user headers

components/percona_telemetry/data_provider.cc
```diff
@@ -0,0 +1,511 @@
...
+  if (!sstatus)
+    sstatus |=
+        command_options_service_.set(mysql_h, MYSQL_COMMAND_USER_NAME, "root");
+  if (!sstatus)
+    sstatus |=
+        command_options_service_.set(mysql_h, MYSQL_COMMAND_HOST_NAME, nullptr);
+  if (!sstatus) sstatus |= command_factory_service_.connect(mysql_h);
+
+  // starting from this point, if the above succeeded we need to close mysql_h.
+  std::shared_ptr<bool> mysql_h_close_guard(
```
percona-ysorokin: You don't have do have either dummy bool  or shared_ptr here.
```c++
MYSQL_H mysql_h = nullptr;
mysql_service_status_t sstatus = command_factory_service_.init(&mysql_h);
auto mysql_h_deleter = [&srv = command_factory_service_](void *ptr){
  if (ptr != nullptr) {
    srv.close(static_cast< MYSQL_H >(ptr));  
  }
};
std::unique_ptr<void, mysql_h_deleter> guard{mysql_h, std::move(mysql_h_deleter)};

File: components/percona_telemetry/data_provider.cc [hunk 34]
```diff
@@ -0,0 +1,511 @@
...
+      if (command_error_info_service_.sql_errno(mysql_h, err_no)) {
+        logger_.warning("Failed to get the last query error");
+      }
+    }
+    goto err;
+  }
+
+  command_query_result_service_.store_result(mysql_h, &mysql_res);
+  if (mysql_res) {
+    std::shared_ptr<bool> query_result_free_guard(
```
percona-ysorokin: Likewise, use std::unique_ptr<void, custom_deleter>

components/percona_telemetry/data_provider.cc
```diff
@@ -0,0 +1,511 @@
...
+  if (id.length() == 0) {
+    logger_.warning(
+        "Collecting db_instance_id failed. It may be caused by server still "
+        "initializing.");
+    return true;
+  }
+  rapidjson::Document::AllocatorType &allocator = document->GetAllocator();
+  rapidjson::Value instance_id;
+  instance_id.SetString(id.c_str(), allocator);
+  document->AddMember(rapidjson::StringRef("db_instance_id"), instance_id,
```
percona-ysorokin: Here and everywhere, ideally the JSON key names should be defined as constants in a single place.

components/percona_telemetry/logger.h
```diff
@@ -0,0 +1,31 @@
...
+         loglevel log_level = SYSTEM_LEVEL);
+
+  ~Logger() = default;
+
+  Logger(const Logger &rhs) = delete;
+  Logger(Logger &&rhs) = delete;
+  Logger &operator=(const Logger &rhs) = delete;
+  Logger &operator=(Logger &&rhs) = delete;
+
+  void info(const char *format, ...) __attribute__((format(printf, 2, 3)));
```
percona-ysorokin: I believe there is a wrapper macro for `__attribute__ `.
Probably `MY_ATTRIBUTE`

components/percona_telemetry/storage.cc
```diff
@@ -0,0 +1,146 @@
+#include <fstream>
```
percona-ysorokin: Likewise, order of includes

components/percona_telemetry/storage.cc
```diff
@@ -0,0 +1,146 @@
...
+#include <boost/uuid/random_generator.hpp>
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_io.hpp>
+
+namespace fs = std::filesystem;
+static constexpr auto path_separator =
+    std::filesystem::path::preferred_separator;
+
+// Let's keep it as std::string, because there is no way to string + string_vew
+static const std::string TMP_FILE_EXTENSION(".tmp");
```
percona-ysorokin: could be std::string_view

components/percona_telemetry/storage.cc
percona-ysorokin: Up to you but a few comments:
1. std::time_t is an unspecified type, so it may not be long.
2. Even if it is long, this is a SIGNED type which may not be what you expect (as you may have already noticed I don't like signed types :) )
3. For instance a difference between 2 timestamps may be negative (in case of clock change / ntp adjustments).

components/percona_telemetry/storage.cc
kamil-holubicki: Unfortunatley no :(
https://stackoverflow.com/questions/44636549/why-is-there-no-support-for-concatenating-stdstring-and-stdstring-view

components/percona_telemetry/component.cc
kamil-holubicki: I'm still not fully convinced, but let's do it in modern way as you said :)

1. Symbols in unnamed namespaces are exported from the library anyway, but their names are randomly mangled. This can cause the linker to do its job longer.
2. Maybe I'm old-school, but when I see `static`, I exactly know, what it means in all possible contexts. Having an unnamed namespace it is not instantly clear what the function scope is.
3. If there is only one static variable/function, wrapping it with an anonymous namespace seems to be too verbose. Using `static` in such a case seems to be violating rules: we should use one or another way of limiting the scope, not mixing to avoid mess.
4. Looks like C++ committee undepreciated use of `static`

components/percona_telemetry/config.cc
kamil-holubicki: fixed

components/percona_telemetry/config.h
kamil-holubicki: fixed

components/percona_telemetry/logger.h
kamil-holubicki: fixed

components/percona_telemetry/storage.cc
kamil-holubicki: fixed

components/percona_telemetry/data_provider.cc
kamil-holubicki: done

components/percona_telemetry/data_provider.cc
kamil-holubicki: done

components/percona_telemetry/data_provider.cc
kamil-holubicki: OK, but let me go with `shared_ptr` as I find it more user friendly to read :) (just a preference)
With unique_ptr we have
```
  auto mysql_h_close_guard = [&srv = command_factory_service_,
                   do_close = !sstatus](void *ptr) {
        if (do_close && ptr) srv.close(static_cast<MYSQL_H>(ptr));
      };

  std::unique_ptr<void, decltype(mysql_h_close_guard)> guard{mysql_h, std::move(mysql_h_close_guard)};
```
while with shared_ptr:
```
 std::shared_ptr<void> mysql_h_close_guard(
      mysql_h, [&srv = command_factory_service_,
                   do_close = !sstatus](void *ptr) {
        if (do_close && ptr) srv.close(static_cast<MYSQL_H>(ptr));
      });
```

percona-ysorokin (approved):
LGTM

VarunNagaraju (approved):

PR #5247, opened 2024-03-05
PS-9117: Make innodb_interpreter_output sysvar readonly

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9117

The innodb_interpreter_output sysvar used to show result of the latest cmd execution by innodb_interpreter and it points to internal interpreter state. Attempt to set this sysvar to any vaalue leads to server crash.
Make innodb_interpreter_output sysvar readonly. This prevents attempts of sysvar memory handling code to brake internal state of innodb interpreter.


percona-ysorokin (approved):
LGTM

PR #5253, opened 2024-03-08
PS-9125: Fix gtid_next formatting in Gtid_specification::automatic_to_string()

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9125

There was a missing '\0' in the end of buffer containing formatted gtid_next in Gtid_specification::automatic_to_string() which lead to garbage symbols in the end of read @@session.gtid_next sysvar.


percona-ysorokin (approved):
LGTM

PR #5263, opened 2024-03-19
PS-9151: Fix build with WITH_SSL=openssl11 [8.0]

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9151

The opensslpp build fails with "Could NOT find OpenSSL" error in case
WITH_SSL=openssl11 is used.
The authentication_fido plugin build fails because of unmet fido2 dependency.

To fix the issue:
- Remove find_package(OpenSSL REQUIRED) in opensslpp/CMakeLists.txt.
  All openssl related preparations are done inside ssl.cmake.
- Update ssl.cmake to always create OpenSSL::SSL and OpenSSL::Crypto
  imported libs. For all possible WITH_SSL values: system, openssl11,
  /path/to/openssl.
- Unset WITH_PERCONA_AUTHENTICATION_FIDO in case the server is being
  built with openssl version incompatible with fido2 lib.
- Added missing WITH_PERCONA_AUTHENTICATION_FIDO in conditions related
  to fido2 lib handling.
- Use ext::curl interface library in keyring_vault and keyring_kms cmake.


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-param-parallel-mtr/42/

percona-ysorokin (approved):
LGTM

PR #5275, opened 2024-03-27
PS-9148: Add caching of dictionary table for component_masking_functions (Old)

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9148

- Added caching of mysql.masking_dictionaries table content.
- Implemented masking_dictionaries_flush() UDF which flushes data from the masking dictionaries table to the memory cache.
- Added component_masking.dictionaries_flush_interval_seconds system
  variable.
- Added actual flusher thread. It periodically rereads content of
  dictionary table and updates in-memory cache.


components/masking_functions/include/masking_functions/query_cache.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include "masking_functions/sql_context.hpp"
+
+namespace masking_functions {
+
+class query_cache {
+ public:
+  query_cache();
+
+  bool check_term_presence_in_dictionary(const std::string &dictionary_name,
```
percona-ysorokin: Please, revisit these names.
It was OK for the `query_builder` whose purpose was to build queries according to the name of the method, but for this `query_cache` I would go with conventional naming (get, add/create, remove/delete, ...)

components/masking_functions/include/masking_functions/sql_context.hpp
```diff
@@ -30,6 +32,8 @@ namespace masking_functions {
 // construction.
 class sql_context {
  public:
+  using dict_container_type = std::multimap<std::string, std::string>;
```
percona-ysorokin: In my naming convention `_container` is already a type suffix (similarly to `_type`) meaning this is a container(vector, list, map, etc.) of something. In other words, if somebody sees a type whose name ends with `_container`, he can make certain assumptions.
So, here I would name it `dictionary_term_container`.

components/masking_functions/src/masking_functions/query_cache.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include "masking_functions/query_cache.hpp"
+
+#include "masking_functions/command_service_tuple.hpp"
+#include "masking_functions/primitive_singleton.hpp"
+#include "masking_functions/query_builder.hpp"
+
+#include <algorithm>
+
+namespace masking_functions {
+namespace {
```
percona-ysorokin, line 36: You are not declaring any data in the anonymous namespace, just typedefs. So, do not really see any benefit of using it here.

components/masking_functions/src/masking_functions/query_cache.cpp
```diff
@@ -0,0 +1,133 @@
...
+    masking_functions::primitive_singleton<masking_functions::query_builder>;
+
+}  // namespace
+
+query_cache::query_cache() { load_cache(); }
+
+bool query_cache::load_cache() {
+  auto query = global_query_builder::instance().select_all_from_dictionary();
+  auto result =
+      masking_functions::sql_context{global_command_services::instance()}
```
percona-ysorokin: Should we make `sql_context` a member of `query_cache` to reuse it for different calls.
Still an open question, may not be the best decision when we add locks to support multithreading.

components/masking_functions/src/masking_functions/query_cache.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  if (range.first == range.second) {
+    return false;
+  }
+
+  if (term.length() == 0) {
+    return true;
+  }
+
+  return std::find_if(range.first, range.second, [term](const auto &el) {
```
percona-ysorokin: This is linear search, which is not good. So, we should consider different data structure here.

components/masking_functions/src/masking_functions/query_cache.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+sql_context::optional_string query_cache::select_random_term_for_dictionary(
+    const std::string &dictionary_name) const {
+  auto range = m_dict_cache.equal_range(dictionary_name);
+
+  if (range.first == range.second) {
+    return std::nullopt;
+  }
+
+  int random_step = rand() % std::distance(range.first, range.second);
```
percona-ysorokin: One of the UDF is already using a better version of `rand()`, consider using it instead.

components/masking_functions/src/masking_functions/sql_context.cpp
```diff
@@ -116,19 +116,22 @@ sql_context::optional_string sql_context::query_single_value(
...
-    throw std::runtime_error{"Couldn't fetch row"};
+    if ((*get_services().query_result->fetch_row)(mysql_res, &row) != 0)
+      throw std::runtime_error{"Couldn't fetch length"};
+    if ((*get_services().query_result->fetch_lengths)(mysql_res, &length) != 0)
+      throw std::runtime_error{"Couldn't fetch length"};
 
-  ulong *length = nullptr;
-  if ((*get_services().query_result->fetch_lengths)(mysql_res, &length) != 0)
-    throw std::runtime_error{"Couldn't fetch lenghts"};
+    result->emplace(std::make_pair(std::string{row[0], length[0]},
```
percona-ysorokin: Use piecewise_construct here as well.

components/masking_functions/src/masking_functions/query_cache.cpp
```diff
@@ -0,0 +1,133 @@
...
+                                       const std::string &term) {
+  masking_functions::sql_context sql_ctx{global_command_services::instance()};
+  auto query = global_query_builder::instance().insert_ignore_record(
+      dictionary_name, term);
+
+  if (!sql_ctx.execute(query)) {
+    return false;
+  }
+
+  m_dict_cache.insert({dictionary_name, term});
```
percona-ysorokin: emplace()

components/masking_functions/include/masking_functions/query_cache.hpp
oleksandr-kachan: Done

components/masking_functions/include/masking_functions/sql_context.hpp
oleksandr-kachan: Done

components/masking_functions/src/masking_functions/query_cache.cpp
oleksandr-kachan: It now has some additional items in there, not only typedefs.

components/masking_functions/src/masking_functions/query_cache.cpp
oleksandr-kachan: My impression was that this context is something specific to actual query execution (due to how it initializes and uses mysql_command_* services for each specific query) and it isn't a good idea to reuse same instance for different requests. And yes, reusing it may be an issue when multi-threading is involved.

components/masking_functions/src/masking_functions/query_cache.cpp
oleksandr-kachan: Used data structure changed already

components/masking_functions/src/masking_functions/query_cache.cpp
oleksandr-kachan: Done

components/masking_functions/src/masking_functions/query_cache.cpp
oleksandr-kachan: This part was changed already

components/masking_functions/include/masking_functions/dictionary_container.hpp
```diff
@@ -0,0 +1,57 @@
...
+#include <vector>
+
+namespace masking_functions {
+
+using optional_string = std::optional<std::string>;
+
+class dictionary_container {
+  struct term_container {
+    explicit term_container(std::string term)
+        : term_mutex{}, term_list{std::move(term)} {}
```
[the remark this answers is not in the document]
oleksandr-kachan: Fixed

PR #5276, opened 2024-04-02

VarunNagaraju:

When a spatial index is used on a geomtery column which contains values with more precision, the comparison of minimum bounding rectangle(MBR) for the geometric values can result in unexpected errors because of trunation/rounding of data.

Since the underlying data type used for representing the coordinates of MBR is double, it supports upto 15 digits after the decimal place and any data after after those digits is subject to rounding and truncated.

So, the MBR comparison method borrowed from boost libaries is updated to hanlde comparisons upto 20 digits after the decimal place since the secondary index update relies on this comparison.


VarunNagaraju:
Name of the branch was wrong.

PR #5282, opened 2024-04-15
PS-9155: Crash in row_sel_convert_mysql_key_to_innobase

kamil-holubicki:
https://perconadev.atlassian.net/browse/PS-9155

Problem:
Server crashes during execution of the complicated query with 9 CTEs.

Cause:
For CTE queries, the optimizer can request the creation of a temp table for a particular CTE. If such a temp result is used in table joins, indexes are added to such a table, based on join columns. All possible indexes are added at the 1st stage and then the optimizer decides which of them will be used, removing unused ones. If the CTE temp result is used in joins in other CTEs of the same query, the process of finding indexes is repeated several times.

Indexes are created in TABLE::add_tmp_key() as <auto_keyARRAY_INDEX>, where ARRAY_INDEX is the index in TABLE_SHARE::key_names array. Let's say <auto_key0>, <auto_key1>, <auto_key2> are created. Then the query optimizer decides that only <auto_key2> will be used, so <auto_key0> and <auto_key1> are removed. <auto_key2> is shifted to the 1st unused position (TABLE::move_tmp_key()),
so TABLE_SHARE::key_names contains <auto_key2> on position 0.

Then the query optimizer makes a plan for another join, repeating the above steps. It adds new keys to TABLE_SHARE::key_names and it results with <auto_key2>, <auto_key1>, <auto_key2>. So we've got two <auto_key2> keys, having different definitions.

Then the temp table is created in the InnoDB world.

Then the query is executed. We get to ha_innobase::open() which calls dict_table_get_index_on_name() requesting <auto_key2> (the 2nd one). But the function returns the 1st index.
innobase_match_index_columns() is called to check the consistency between MySql and InnoDB index definition and here we get the message:

[ERROR] [MY-010880] [InnoDB] Found index <auto_key2> whose column info does not match that of MySQL. [ERROR] [MY-010882] [InnoDB] Build InnoDB index translation table for Table /var/lib/mysql/tmp/#sql3dad_13_4 failed

Then we get to ha_innobase::index_read() where
row_sel_convert_mysql_key_to_innobase() is called. The function doesn't do many checks, just converts the key. As the definition of MySql key is longer (e.g. 8 + 8 bytes) than InnoDB's (e.g. 8 + 6), we try to read too much, causing a segmentation fault.

The problem is in the way of naming auto keys by the optimizer. The current algorithm does not guarantee the uniqueness of key names, causing name clashes in InnoDB world.

Solution:
Introduce a variable that will serve the unique number for auto_key names causing index names to be unique.

MTR tests re-recorded, because now key ids are generated starting from zero, even if the 1st generated key is on position 1. Note that there can be <auto_distinct_key> on position added before.


kamil-holubicki:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/482/

sql/table.h
```diff
@@ -842,6 +842,7 @@ struct TABLE_SHARE {
   uint fields{0};            /* Number of fields */
   uint rec_buff_length{0};   /* Size of table->record[] buffer */
   uint keys{0};              /* Number of keys defined for the table*/
+  uint temp_table_key_id{0}; /* Serves the unique number for <auto_keyN> */
```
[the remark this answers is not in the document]
kamil-holubicki: It needs to be public. It follows the convention of this struct. The other way would be implementing an access method, but would not follow convention.

sql/table.cc
```diff
@@ -6029,7 +6029,7 @@ bool TABLE::add_tmp_key(Field_map *key_parts, bool invisible,
     s->max_key_length = std::max(s->max_key_length, key_len);
     s->key_parts += key_part_count;
     assert(s->keys < s->max_tmp_keys);
-    sprintf(s->key_names[s->keys].name, "<auto_key%d>", s->keys);
+    sprintf(s->key_names[s->keys].name, "<auto_key%d>", s->temp_table_key_id++);
```
satya-bodapati, line 6032: [optional]: if you want to reduce testcase diff, may be we can use ++s->temp_table_key_id.

satya-bodapati (approved):
LGTM! I have already reviewed as PXC-4340

sql/table.cc
kamil-holubicki: Most probably it would increase the number of tests to be re-recorded because the 1st auto_key will be <auto_key1> always. I see a lot of tests results have <auto_key0> inside

satya-bodapati: Hmm, this is weird. Sometimes keys with 0 and sometimes key with 1. Ignore it.

satya-bodapati: my bad, because of dup key issue, it might be re-using either 0 or 1.

kamil-holubicki: Note that there may be other keys (not <auto_keyN>) already there (like <auto_distinct_key>), so the 1st auto_keyN create had 1 in the name before the fix (this is why re-record was needed - now it has 0 in its name). 
But there can be also a situation when there are no keys, so 1st auto_key would have 0 in name before the change. If we changed it as you proposed, it would have 1 in the name, and it won't be aligned with already recorded test results (other tests)

satya-bodapati: Thanks for explaining the reason for the difference of key IDs 0 and 1. Please leave it as is.

PR #5285, opened 2024-04-17
PS-8963 fix 8.0: SEQUENCE_TABLE Issue

percona-ysorokin:
https://perconadev.atlassian.net/browse/PS-8963

Introduced new 'PERCONA_SEQUENCE_TABLE' reserved keyword which is supposed
to be a safe replacement for 'SEQUENCE_TABLE' which is now considered to
be deprecated.
Currently both have identical behavior except the fact that the latter now generates
a deprecation warning.

As an intermediate step before deprecated 'SEQUENCE_TABLE' is removed the
MySQL parser has been modified so that 'SEQUENCE_TABLE' is no longer a
reserved keyword and can now be used as a regular MySQL identifier (table name,
column name, alias name, etc).

'SEQUENCE_TABLE' token added to the list of the non-reserved keywords
('<ident_keywords_unambiguous>') in the Percona Server grammar file
('sql_yacc.yy').

Added new 'main.percona_sequence_table_keyword' MTR test case that check
various scenarios of 'SEQUENCE_TABLE' used as a simple table name, as an alias,
or a table function.

Re-recorded 'main.information_schema_keywords' MTR test case because
'SEQUENCE_TABLE' is no longer a reserved keyword.

Updated a number of existing MTR test cases with new 'PERCONA_SEQUENCE_TABLE()'
table-level function instead of 'SEQUENCE_TABLE()'.


kamil-holubicki (approved):

percona-ysorokin:
Jenkins build link
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-param-parallel-mtr/44/
(main.information_schema_keywords already re-recorded).

percona-ysorokin:
Jenkins build link
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-param-parallel-mtr/54/
`main.mysqld--help-notwin` MTR test case already re-recorded.

dlenev (approved):
Hello Yura!

The patch LGTM!

PR #5287, opened 2024-04-19
PS-9211: Fix doxygen for function templates

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9211

Build with the clang-18 failed with the following error:

'@tparam' command used in a comment that is not attached to a template declaration

Made affected documentation a regular comment for now as attempts to fix
it for clang-18 brakes it for clang-16/17.


percona-ysorokin (approved):
LGTM

PR #5289, opened 2024-04-23
PS-9213: Fix typo in condition in mysqld.cc::fix_secure_path()

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9213


inikep (approved):
LGTM
It restores code from upstream: https://github.com/mysql/mysql-server/blob/8.0/sql/mysqld.cc#L11531

PR #5292, opened 2024-05-18
 [PS-9222][PoC] Include reordered fields when calculating mlog buffer size 

keeper:
**Note**: this change is intended to demonstrate the fix for a regression filed as PS-9222 found
in upstream MySQL 8.0.37 and should be reworked and merged accordingly only after it
has been imported. The current PR cherry-picks the commit causing the regression ([e6e13a8f](https://github.com/mysql/mysql-server/commit/e6e13a8f64ca1005f0d2d4da8463fdbef3767917)) from upstream as is, then apply the fix upon it.

The root cause is that the upstream commit adds fields to the log in additional conditions, but the mlog buffer size isn't calculated accordingly, then when the buffer is used up it triggers the assertion and crashes the server.

We've tested the fix with upstream MySQL 8.0.37 with the reproduce step in PS-9222, which is also added as an MTR test in this PR. And the MTR test in e6e13a8f also passes so no new regression is observed.

### DIFF
--- a/storage/innobase/mtr/mtr0log.cc
+++ b/storage/innobase/mtr/mtr0log.cc
@@ -514,18 +514,28 @@ constexpr size_t inst_col_info_size = 6;
 @param[in]   is_comp       true if COMP
 @param[in]   is_versioned  if table has row versions
 @param[in]   is_instant    true if table has INSTANT cols
+@param[in]   changed_order array indicating fields changed position
 @param[out]  size_needed   total size needed on REDO LOG */
 static void log_index_get_size_needed(const dict_index_t *index, size_t size,
                                       uint16_t n, bool is_comp,
                                       bool is_versioned, bool is_instant,
+                                      const bool *changed_order,
                                       size_t &size_needed) {
-  auto size_for_versioned_fields = [](const dict_index_t *ind) {
+  auto size_for_versioned_fields = [n, changed_order](const dict_index_t *ind) {
     size_t _size = 0;
     /* 2 bytes for number of columns with version */
     _size += 2;
 
-    size_t n_versioned_fields = ind->table->get_n_instant_add_cols() +
-                                ind->table->get_n_instant_drop_cols();
+    size_t n_versioned_fields = 0;
+    for (size_t i = 0; i < n; i++) {
+      dict_field_t *field = ind->get_field(i);
+      const dict_col_t *col = field->col;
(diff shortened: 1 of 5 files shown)


sleto-it:
Thanks for the contribution @keeper

storage/innobase/mtr/mtr0log.cc
```diff
@@ -798,9 +809,36 @@ bool mlog_open_and_write_index(mtr_t *mtr, const byte *rec,
...
+  /* Ordinal position of an existing field can't be changed with INSTANT
+  algorithm. But when it is combined with ADD/DROP COLUMN, ordinal position
+  of a filed can be changed. This bool array of size #fields in index,
+  represents if ordinal position of an existing filed is changed. */
+  
+  // Move the logic to find columns that changed order before calculating
+  // the buffer size
+  bool *fields_with_changed_order = nullptr;
+  if (is_versioned) {
+    fields_with_changed_order = new bool[n];
```
zhry110, line 821: `fields_with_changed_order` should be freed.
if (!mlog_open(mtr, alloc, log_ptr)) {
  if (is_versioned) {
    delete[] fields_with_changed_order;
  }
}

keeper: Thanks for review! It's a good catch, and I've fixed this leak in latest update.

ottok:
Now with the release of MySQL 8.0.38 we can see that they did the exact same fix with only a variable name changed and tests omitted end of May, a couple weeks _after_ this PR was published:
- https://github.com/percona/percona-server/commit/2f2d043b6351f19de238398414d624287aebe6f1 = https://github.com/mysql/mysql-server/commit/e6248f511b031d76a7cfd7a7b87d73308698e55b
- later they did the exact same memory leak fix that was in the update to this PR: https://github.com/percona/percona-server/commit/5c9aed82d8265ec4446d1db606a385342bcee3be#diff-929c623893c233f9ccadf8e7988fa95e48f63e9a8910ee3fc194f302b4fe34daR850-R852 = https://github.com/mysql/mysql-server/commit/a3561b3aa41bec668961db9b434fd98733b89a1f

Hence this PR is now obsolete.

VarunNagaraju:
Hi @keeper, @ottok, @zhry110,
Thanks for your contribution. The fixes for the reported issue have been included for PS 8.0.37 which will be released in the near future.

PR #5294, opened 2024-05-20
PS-9219: MySQL converts collation of date data type in ibd but data dictionary (8.0) 

venkatesh-prasad-v:
PS-9219: MySQL converts collation of date data type in ibd but data dictionary

https://perconadev.atlassian.net/browse/PS-9219

Problem
=======
Import tablespace operation fails due to the incorrect table definition
in the ibd file if the charset-collation was changed before the backup
if the table had a column of temporal type (date, datetime, timestamp).

Analysis
========
In MySQL temporal types are always stored and compared using
my_charset_latin1 charset.

During the execution of ALTER TABLE CONVERT TO CHARACTER SET, MySQL
changes the charset and collation stored in data-dictionary and SDI
for temporal columns to a different collation_id.
In practice, this new collation_id is ignored when such columns
are stored/compared. Corresponding Field objects are not updated to
use this new collation (Actually, Field objects for temporal types are
hardcoded to use my_charset_latin1).

This new collation is not visible in I_S and SHOW CREATE TABLE output.
It will be ignored by CREATE TABLE LIKE and rewritten by ALTER TABLE
since both these statements use info from Field objects of source table
to produce Create_field objects describing columns of new table/new
version of the table.

For example:

create table a(dt datetime);
$ ./bin/ibd2sdi ./var/mysqld.1/data/test/a.ibd
"name": "dt",
"collation_id": 8

alter table a CONVERT TO CHARACTER SET utf8mb4 collate utf8mb4_unicode_ci;
$ ./bin/ibd2sdi ./var/mysqld.1/data/test/a.ibd
"name": "dt",
"collation_id": 224

alter table a engine = innodb;
$ ./bin/ibd2sdi ./var/mysqld.1/data/test/a.ibd
"name": "dt",
"collation_id": 8

However, this almost invisible incorrect collation_id causes problems when
we try to restore InnoDB table with such a column from the backup created
using Percona eXtraBackup tool.

This tool uses information from DD/SDI to produce .cfg file describing
InnoDB table being restored. Later information from this file is used
by ALTER TABLE IMPORT TABLESPACE that imports restored table.

Particularly, for new temporal types that support fractional seconds
collation_id from DD/SDI affects the InnoDB "precise type" describing
the column and stored in .cfg. Because of this, incorrect collation_id
in DD/SDI for such columns results in incorrect "precise type" in .cfg
file.

As consequence we get Schema mismatch (Column a precise type mismatch.)
error when ALTER TABLE IMPORT TABLESPACE compares the "precise type"
from such .cfg file with "precise type" for column in table version
being imported into. The latter is based on collation_id which ultimately
comes from Field object and always corresponds to my_charset_latin1 for
temporal types.

Note that this problem do not affect scenarios when we import table
with .cfg file that was generated by MySQL's FLUSH TABLE FOR EXPORT
command, as the latter doesn't use DD/SDI to calculate "precise type"
but gets information from Field object in table being exported instead.

Solution
========
This commit changes ALTER TABLE CONVERT TO CHARACTER SET command to
not alter the character set for temporal types stored in data-dictionary/SDI.
In other words, we now force the server to always use my_charset_latin1
in DD/SDI for temporal types.

There will be a separate fix to Percona eXtraBackup tool which will
change code generating .cfg to ignore collation_id stored in DD/SDI for
temporal columns like it is already done for some other types with fixed
collation_id.

PR for trunk: https://github.com/percona/percona-server/pull/5295

Testing Done
========
Jenkins: https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-param-parallel-mtr/46/console

### DIFF
--- a/sql/sql_table.cc
+++ b/sql/sql_table.cc
@@ -4220,6 +4220,12 @@ const CHARSET_INFO *get_sql_field_charset(const Create_field *sql_field,
   if (sql_field->is_array || cs == &my_charset_bin) return cs;
 
   /*
+    Temporal types always use my_charset_numeric.
+    Ensure that it is not changed when altering table charset.
+  */
+  if (is_temporal_real_type(sql_field->sql_type)) return &my_charset_numeric;
+
+  /*
     table_charset is set only in ALTER TABLE t1 CONVERT TO CHARACTER SET csname
     when we want to change character set for all varchar/char columns.
   */
--- a/mysql-test/r/percona_temporal_type_collation.result
+++ b/mysql-test/r/percona_temporal_type_collation.result
@@ -1,0 +1,95 @@
+CREATE TABLE test (
+dt datetime primary key,
+datetxt varchar(10) GENERATED ALWAYS AS (DATE(dt)) STORED,
+timetxt varchar(10) GENERATED ALWAYS AS (TIME(dt)) STORED
+) ENGINE=InnoDB;
+SHOW CREATE TABLE test;
+Table	Create Table
+test	CREATE TABLE `test` (
+  `dt` datetime NOT NULL,
+  `datetxt` varchar(10) GENERATED ALWAYS AS (cast(`dt` as date)) STORED,
+  `timetxt` varchar(10) GENERATED ALWAYS AS (cast(`dt` as time)) STORED,
+  PRIMARY KEY (`dt`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
+FLUSH TABLE test FOR EXPORT;
+UNLOCK TABLES;
+#
+# Assert that the collation_id is my_charset_numericd in the SDI.
+#
+include/assert_grep.inc [Assert that collation_id is 8 in the sdi]
+#
+# Assert that the collation_id does not change after ALTER TABLE
+# ENGINE=InnoDB
+#
+ALTER TABLE test ENGINE=InnoDB;
+include/assert_grep.inc [Assert that collation_id is 8 in the sdi]
+#
+# Assert that the collation_id does not change after ALTER TABLE CONVERT
+# TO CHARACTER SET
+#
+ALTER TABLE test CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci;
+include/assert_grep.inc [Assert that collation_id is 8 in the sdi]
+#
+# Insert some data and verify that comparision with varchar columns are handled properly.
+#
+INSERT INTO test(dt) VALUES("1997-11-30 12:30:45"),("1999-12-13 10:11:53");
+SELECT * FROM test;
+dt	datetxt	timetxt
+1997-11-30 12:30:45	1997-11-30	12:30:45
+1999-12-13 10:11:53	1999-12-13	10:11:53
+#
+# Verify that there are two rows in the table
+#
+SELECT * FROM test WHERE DATE(dt) = datetxt;
+dt	datetxt	timetxt
+1997-11-30 12:30:45	1997-11-30	12:30:45
+1999-12-13 10:11:53	1999-12-13	10:11:53
+SELECT * FROM test WHERE TIME(dt) = timetxt;
+dt	datetxt	timetxt
+1997-11-30 12:30:45	1997-11-30	12:30:45
+1999-12-13 10:11:53	1999-12-13	10:11:53
+#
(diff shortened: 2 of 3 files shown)


sql/sql_table.cc
```diff
@@ -4219,6 +4219,11 @@ const CHARSET_INFO *get_sql_field_charset(const Create_field *sql_field,
   */
   if (sql_field->is_array || cs == &my_charset_bin) return cs;
 
+  /*
+    Always use my_charset_numeric for temporal types.
```
dlenev: Perhaps it is better to elaborate a bit and say something like:

Temporal types always use my_charset_numeric.
Ensure that it is not changed when altering table charset.

dlenev (changes requested):
Hello Venkatesh!

The patch itself looks good to me. I only have some comments about commit message
and comments.

1) First of all, could you please fix typo in commit message title and replace "collection" with "collation" ? Also "tpye" -> "type".

2) I also think that it is worth to rewrite the rest of commit message:

a) To avoid creating false impression that temporal
   columns can have collation different than my_charset_latin1
   - they really can't, it is only collation stored in
   data-dictionary/SDI for such columns that can be wrong.

b) To focus on how this wrong collation stored in DD/SDI
   causes problems during restoring of backup.

For example we can say something like:

Analysis
========
In MySQL temporal types are always stored and compared using
my_charset_latin1 charset.

During the execution of ALTER TABLE CONVERT TO CHARACTER SET, MySQL
changes the charset and collation stored in data-dictionary and SDI
for temporal columns to a different collation_id.
In practice, this new collation_id is ignored when such columns
are stored/compared. Corresponding Field objects are not updated to
use this new collation (Actually, Field objects for temporal types are
hardcoded to use my_charset_latin1).

This new collation is not visible in I_S and SHOW CREATE TABLE output.
It will be ignored by CREATE TABLE LIKE and rewritten by ALTER TABLE
since both these statements use info from Field objects of source table
to produce Create_field objects describing columns of new table/new
version of the table.

For example:

1. create table a(dt datetime);
   $ ./bin/ibd2sdi ./var/mysqld.1/data/test/a.ibd
   "name": "dt",
   "collation_id": 8

2. alter table a CONVERT TO CHARACTER SET utf8mb4 collate utf8mb4_unicode_ci;
   $ ./bin/ibd2sdi ./var/mysqld.1/data/test/a.ibd
   "name": "dt",
   "collation_id": 224

3. alter table a engine = innodb;
   $ ./bin/ibd2sdi ./var/mysqld.1/data/test/a.ibd
   "name": "dt",
   "collation_id": 8

However, this almost invisible incorrect collation_id causes problems when
we try to restore InnoDB table with such a column from the backup created
using Percona eXtraBackup tool.

This tool uses information from DD/SDI to produce .cfg file describing
InnoDB table being restored. Later information from this file is used
by ALTER TABLE IMPORT TABLESPACE that imports restored table.

Particularly, for new temporal types that support fractional seconds
collation_id from DD/SDI affects the InnoDB "precise type" describing
the column and stored in .cfg. Because of this, incorrect collation_id
in DD/SDI for such columns results in incorrect "precise type" in .cfg
file.

As consequence we get `Schema mismatch (Column a precise type mismatch.)`
error when ALTER TABLE IMPORT TABLESPACE compares the "precise type"
from such .cfg file with "precise type" for column in table version
being imported into. The latter is based on collation_id which ultimately
comes from Field object and always corresponds to my_charset_latin1 for
temporal types.

Note that this problem do not affect scenarios when we import table
with .cfg file that was generated by MySQL's FLUSH TABLE FOR EXPORT
command, as the latter doesn't use DD/SDI to calculate "precise type"
but gets information from Field object in table being exported instead.

Solution
========
This commit changes ALTER TABLE CONVERT TO CHARACTER SET command to
not alter the character set for temporal types stored in data-dictionary/SDI.
In other words, we now force the server to always use my_charset_latin1
in DD/SDI for temporal types.

There will be a separate fix to Percona eXtraBackup tool which will
change code generating .cfg to ignore collation_id stored in DD/SDI for
temporal columns like it is already done for some other types with fixed
collation_id.

sql/sql_table.cc
venkatesh-prasad-v: Done

venkatesh-prasad-v:
@dlenev Thank you very much for the detailed and the insightful review. I have updated both the commit message and the code comment as per your suggestions.

dlenev (approved):
Hello Venki!

I think it is OK to push this patch.

PR #5303, opened 2024-05-31
(8.4) PS-9165: Product Usage Tracking - phase 1

kamil-holubicki:

Implementation of Percona Telemetry Component


sql/mysqld.cc
```diff
@@ -1238,6 +1238,9 @@ bool opt_libcoredumper, opt_corefile = 0;
 bool opt_persist_sensitive_variables_in_plaintext{true};
 int argc_cached;
 char **argv_cached;
+#ifdef HAVE_PERCONA_TELEMETRY
+bool opt_percona_telemetry_disable = false;
```
oleksandr-kachan, line 1242: Just to recheck. As I understand it, telemetry is disabled by default. Should this default to true (just in case)?

mysql-test/suite/innodb/t/tablespace_encrypt_9.test
```diff
@@ -149,7 +149,6 @@ CREATE TABLESPACE temp_ts ADD DATAFILE 'temp_ts.ibd';
 --echo # connection con3
 connect (con3,localhost,root,,);
 connection con3;
-SELECT connection_id();
```
oleksandr-kachan, line 152: Is this related to telemetry or was this some other issue?

oleksandr-kachan (approved):
Just a couple of questions but in general looks good.

sql/mysqld.cc
kamil-holubicki: Yes, it should be disabled and not built-in by default. The reason for this is that it makes sense to have telemetry component only with Telemetry Agent package being dependency to PS package. And this is only the case for packages created by Percona. So we will build-in telemetry in our packages, while if someone clones the repo and wants to build PS by himself, there is no point in having telemetry component there.

mysql-test/suite/innodb/t/tablespace_encrypt_9.test
kamil-holubicki: Having connection_id recorderd in the result seems to be not good idea, as it is a random number in general. The reason was internal connection made by telemetry component.

PR #5312, opened 2024-06-10
PS-9218: Merge MySQL 8.4.0 (replace keyring_file plugin with component_keyring_file in MTR tests)

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9218


percona-ysorokin (approved):
LGTM

PR #5318, opened 2024-06-14

VarunNagaraju:

Problem
======
Errors about InnoDB unable "to purge non-delete-marked record in index" occurred in error log when geometry column with a spatial index on it was updated in a such a way that MBRs for old value and new value were close enough (for example, when difference was smaller than 1.0e-16 for values below 1).
Moreover server might abort due to assertion failure if later the updated row was deleted and then its old version was re-inserted again.

Analysis
======
The problem occurs due to combination of several factors:

MySQL uses double floating-point type to store coordinates in Geometry values and do related calculations.
Update code in InnoDB tries to avoid updating spatial index value if MBR of new value doesn't differ from the old one. To compare MBRs for old and new values it uses mbr_equal_cmp() function, which nowadays uses Boost GIS to do this. The latter uses Boost math::equals() method for this comparison which does approximate epsilon comparison for doubles.
Other code in InnoDB, specifically purge code uses exact comparison of MBRs for lookups in spatial index.

As result of the above discrepancy we get problem with purge described above (purge does not associate old MBR in spatial index with new row version/value and tries to delete along with old row version) and later assertion failures.

Note that before the above optimization to update code was introduced, the method mbr_equal_cmp() was doing exact MBR comparison, so the problem didn't exist. This has been changed when support for geographic coordinate systems were added.

Solution
======
This problem is solved by providing another MBR comparison method which does exact comparison and using it in the Update code (and thus effectively revert to old, pre-bug behavior as result).


VarunNagaraju:
Jenkins testrun: https://ps80.cd.percona.com/job/percona-server-8.x-pipeline-parallel-mtr/199/

PR #5323, opened 2024-06-20
Release 8.0.37 29 - Telemetry Ph1 Packaging changes

surbhat1595:


build-ps/debian/extra/apparmor-profile
```diff
@@ -75,6 +75,10 @@
 # Allow access to openssl config
   /etc/ssl/openssl.cnf r,
 
+# Allow access to PS Telemetry directory
```
kamil-holubicki: I think this file is not used. We use the ones from apparmor.d directory

build-ps/debian/percona-server-server.postinst
```diff
@@ -182,6 +201,29 @@ EOF
 			echo "Please use etc/mysql/mysql.conf.d for any custom configuration settings"
 		fi
 		update-alternatives --force --install /etc/mysql/my.cnf my.cnf "/etc/mysql/mysql.cnf" 300
+		if [ -f "/etc/apparmor.d/usr.sbin.mysqld" ]; then
```
kamil-holubicki, line 212: This part seems to be repeated twice. Maybe it would be possible to extract it to a common function?

kamil-holubicki (approved):
For me it is OK. But let's wait for the final approve from @EvgeniyPatlan

EvgeniyPatlan (approved):

PR #5324, opened 2024-06-20
PS-9235: Skip deleted keys in keys list with Vault API v2

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9235

This is a follow up fix for https://github.com/percona/percona-server/commit/7a874b2ec4a8e579408b65fd836c83065da8f3a5 which fixed only one case of deleted keys processing.

During keyring_vault plugin startup it fethces a list of known key names via Vault API and populates key's cache without fetching actual key data at this stage. In case v2 API is used fetched list contains names for already deleted keys (with delition_time set in key metadata).

Later when server attempts to create a key in keyring with some name its existance is checked in local cache first. The plugin finds a name for already deleted key and makes a wrong assumption as for key existance.

To fix the issue cache populating was improved. Now it loads complette key data from the Vault server while populating cache. This lets plugin to check if particular key is still valid. Already deleted keys will not be added to local cache during plugin startup.


plugin/keyring_vault/vault_io.cc
```diff
@@ -75,7 +76,16 @@ bool Vault_io::get_serialized_object(ISerialized_object **serialized_object) {
   }
 
   std::unique_ptr<Vault_keys_list> keys(new Vault_keys_list());
-  if (vault_parser->parse_keys(json_response, keys.get())) {
+  auto key_callback = [&keys, this](std::unique_ptr<Vault_key> key) {
+    // With Vault API v2 server returns keys for already deleted data (with
+    // deletion_time set) when populating cash during plugin initialization.
```
kamil-holubicki: cash -> cache

plugin/keyring_vault/vault_io.cc
```diff
@@ -75,7 +76,16 @@ bool Vault_io::get_serialized_object(ISerialized_object **serialized_object) {
...
+    // Need to fetch actual key and check it has valid version before using it.
+    if (!retrieve_key_type_and_data(key.get())) {
```
kamil-holubicki, line 83: I understand that we now have the cache populated with all key id:value pairs. I'm not sure if this is OK from the security point of view. I think we should keep as less key-related data as possible. Ideally the key should be only in HW cypher module and in keyring server. But that's the ideal case - we have a cache to speed up the key access.
Would it be possible to fetch the key here, validate it, and keep only its ID in the cache as before?

oleksandr-kachan: @kamil-holubicki, in general, yes. I think I could drop actual data after verifying it. But I'm not sure this will change anything. The point is that it anyway will be cached later in Vault_keys_container::fetch_key(). If we drop data here during cache population it still will be fetched a few seconds later. Do you think it still makes sense?

plugin/keyring_vault/vault_io.cc
oleksandr-kachan: Done

plugin/keyring_vault/vault_io.cc
kamil-holubicki: I don't know what is the use scenario. If it is as you say that ALL keys will be fetched anyway in a few seconds, the proposed solution is fine.

percona-ysorokin (approved):
LGTM

PR #5325, opened 2024-06-20
Release-5.7.44-50

adivinho:


VarunNagaraju (approved):
@adivinho The changes look okay. But, the variable name DEBIAN_VERSION is a bit confusing. I would suggest having something like LINUX_VERSION.

PR #5335, opened 2024-06-26
PS-9165 Percona telemetry MTR test post-push fix

VarunNagaraju:

The thread number in the test percona_utility_user is masked since it is different when telemetry is enabled from when it is disabled.

### DIFF
--- a/mysql-test/t/percona_utility_user.test
+++ b/mysql-test/t/percona_utility_user.test
@@ -245,6 +245,7 @@ SELECT COUNT(*) FROM performance_schema.threads where type='FOREGROUND';
 connection default;
 
 --disable_query_log
+--replace_regex /thread [0-9]+/thread X/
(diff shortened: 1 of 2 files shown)


mysql-test/t/percona_utility_user.test
```diff
@@ -245,6 +245,7 @@ SELECT COUNT(*) FROM performance_schema.threads where type='FOREGROUND';
 connection default;
 
 --disable_query_log
+--replace_regex /thread [0-9]*/thread X/
```
kamil-holubicki: Just a suggestion: + instead of *

kamil-holubicki (approved):

mysql-test/t/percona_utility_user.test
VarunNagaraju: Done.

PR #5337, opened 2024-06-27
PKG-26 Create PS packages with telemetry - 8.4

surbhat1595:


kamil-holubicki (commented):
One question: how WITH_PERCONA_TELEMETRY=ON cmake flag is set? Is it passed by Jenkins as a parameter?

surbhat1595:
> One question: how WITH_PERCONA_TELEMETRY=ON cmake flag is set? Is it passed by Jenkins as a parameter?

We already have WITH_PERCONA_TELEMETRY=ON  cmake flag set to ON in our packaging files.

EvgeniyPatlan (approved):

adivinho (approved):

PR #5340, opened 2024-06-27
PS-9165 postfix 8.0: Product Usage Tracking - phase 1 (MTR fixes)

kamil-holubicki:

1. test_session_info.test duplicated with IDs recorded for the case when Percona Telemetry is built-in. It is better than masking IDs in output, because the test relies on real ID values

2. regression.test - ID masked in test output

3. prep_stmt_sundries.test - make assertion value dependant on Percona Telemetry being built-in

### DIFF
--- a/mysql-test/suite/test_service_sql_api/t/test_session_info_percona_telemetry-master.opt
+++ b/mysql-test/suite/test_service_sql_api/t/test_session_info_percona_telemetry-master.opt
@@ -1,0 +1,2 @@
+$TEST_SESSION_INFO_OPT
+--mysqlx=off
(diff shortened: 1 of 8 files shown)


mysql-test/suite/test_service_sql_api/t/test_session_info_percona_telemetry-master.opt
```diff
@@ -0,0 +1,2 @@
+$TEST_SESSION_INFO_OPT
+--mysqlx=off
```
percona-ysorokin: This may help if the server is built without X Plugin
```suggestion
--loose-mysqlx=off
```

percona-ysorokin (approved):
LGTM with single single comment addressed

PR #5341, opened 2024-06-27
PS-9165 postfix 8.0: Product Usage Tracking - phase 1

kamil-holubicki:

When the server is started with --innodb-read-only flag, Percona Telemetry Component can not be installed/uninstalled, because the it is prohibited to add/delete the row in mysql.component table. In such a case
"Skip updating mysql metadata in InnoDB read-only mode." warning was printed into server's error log which caused several MTR tests to fail.

Solution: Do not call dd::check_if_server_ddse_readonly() but do the check directly and print information level message if needed.


percona-ysorokin (approved):
LGTM

PR #5344, opened 2024-07-01
PS-9233 8.4 UUID version 1-7 implementation according to RFC 9562

lukin-oleksiy:
PS-9233 UUID version 1-7 implementation according to RFC 9562

https://perconadev.atlassian.net/browse/PS-9233


percona-ysorokin (approved):
LGTM

PR #5345, opened 2024-07-01
PS-9218: Merge MySQL 8.4.0 (telemetry fixes)

oleksandr-kachan:
***
PS-9165 postfix: Product Usage Tracking - phase 1 (MTR fixes)

'sys_vars.plugin_dir_basic' and  'clone.plugin_mismatch' MTR test cases modified
so that they could be run on a server built both with and without telemetry
component ('-DWITH_PERCONA_TELEMETRY=ON' CMake option).
***
PS-9165 postfix 8.0: Product Usage Tracking - phase 1

When the server is started with --innodb-read-only flag,
Percona Telemetry Component can not be installed/uninstalled,
because the it is prohibited to add/delete the row in mysql.component
table. In such a case
"Skip updating mysql metadata in InnoDB read-only mode."
warning was printed into server's error log which caused several MTR
tests to fail.

Solution: Do not call dd::check_if_server_ddse_readonly()
but do the check directly and print information level message if needed.
***
PS-9165 postfix 8.0: Product Usage Tracking - phase 1 (MTR fixes) (#5340)

1. test_session_info.test duplicated with IDs recorded for the case
when Percona Telemetry is built-in. It is better than masking IDs in
output, because the test relies on real ID values

2. regression.test - ID masked in test output

3. prep_stmt_sundries - make assertion value dependant on Percona
Telemetry being built-in
***
PS-9165 postfix: Product Usage Tracking - phase 1 (MTR KV fixes) (#5342)

'keyring_vault.keyring_udf' MTR test case modified so that it could be run
on a server built both with and without telemetry component
('-DWITH_PERCONA_TELEMETRY=ON' CMake option).

### DIFF
--- a/mysql-test/include/have_percona_telemetry.inc
+++ b/mysql-test/include/have_percona_telemetry.inc
@@ -1,0 +1,3 @@
+if (`SELECT COUNT(*) = 0 FROM mysql.component WHERE component_urn='file://component_percona_telemetry'`) {
+  --skip Test not supported when Percona Telemetry is not built-in
+}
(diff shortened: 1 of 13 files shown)


percona-ysorokin (approved):
LGTM

PR #5348, opened 2024-07-05
PS-9297 Lost a record while scanning in reverse order on cursor

zhry110:
https://perconadev.atlassian.net/browse/PS-9297

Problem
========
If the last user record on the prev leaf is pessimistically updated during move_backward_from_page, the record may be deleted from the prev leaf and reinserted into the current page. Scanning in reverse order will miss this record.

Cause
========
Same as PS-9092, move_backward_from_page cannot sense that records are moved from the prev leaf to the current leaf.

Solution
========
If btr_insert_into_right_sibling causes records to move between pages, prevent the next page from being optimistically accessed

### DIFF
--- a/storage/innobase/btr/btr0cur.cc
+++ b/storage/innobase/btr/btr0cur.cc
@@ -3849,6 +3849,7 @@ dberr_t btr_cur_pessimistic_update(ulint flags, btr_cur_t *cursor,
   dberr_t optim_err;
   roll_ptr_t roll_ptr;
   bool was_first;
+  bool was_before_last = false;
   ulint n_reserved = 0;
   ulint max_ins_size = 0;
   trx_t *const trx = (thr == nullptr) ? nullptr : thr_get_trx(thr);
@@ -4155,6 +4156,10 @@ dberr_t btr_cur_pessimistic_update(ulint flags, btr_cur_t *cursor,
   record on its page? */
   was_first = page_cur_is_before_first(page_cursor);
 
+  was_before_last =
+      !page_rec_is_supremum(btr_cur_get_rec(cursor)) &&
+      page_rec_is_supremum(page_rec_get_next(btr_cur_get_rec(cursor)));
+
   /* Lock checks and undo logging were already performed by
   btr_cur_upd_lock_and_undo(). We do not try
   btr_cur_optimistic_insert() because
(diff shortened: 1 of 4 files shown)


percona-ysorokin:
Thanks for your contribution, @zhry110
Indeed a high quality patch with a non-trivial MTR test case 💪
Unfortunately, at the moment it is too late to include this patch into 8.0.37.
However, we will definitely do this in 8.0.38.

storage/innobase/btr/btr0cur.cc
```diff
@@ -4169,6 +4174,15 @@ dberr_t btr_cur_pessimistic_update(ulint flags, btr_cur_t *cursor,
   ut_ad(rec_offs_validate(rec, cursor->index, *offsets));
   page_cursor->rec = rec;
 
+  if (was_before_last) {
+    /** We may moved the record from the current page to the next page.
+     * Therefore, we need to disable optimistic access to the next page to
+     * ensure that the record will not be mistakenly skipped during reverse
+     * scanning on cursor.
+     */
+    buf_block_modify_clock_inc(btr_cur_get_block(cursor));
```
satya-bodapati, line 4183: While PS-9092 blocked the merge(that caused records movement), this bug shows that records are moved without a merge.

But when a record is moved pessimistically, how can we assume it is transferred to a neighbour? You should limit this check if the movement is neighbour only.

If the record is moved to a non-neighbour, please check this case.
It should be protected by delete marking of record. ie we should read delete marked version of record and the new version that is moved to some far page is not visible/seen. This is OK.

To summarize:
1. record movement due to merges (PS-9092)
2. record movement to neighbour without merge (this PR )
3. record movement to FAR pages (to be verified that delete marked version is read and new version is not SEEN).

satya-bodapati (changes requested):

storage/innobase/btr/btr0cur.cc
zhry110: Thanks, but in our observations, pessimistic updates are limited to the current page and its sibling pages. Whether it is inserted after splitting or inserted directly into its sibling page, the inserted page needs to be disabled for optimistic access, so we do not need stricter judgment here. In addition, the original record has been physically deleted from the original page, and no changes to the delete-mark will be involved.

satya-bodapati: Hmm, that's because for the query `update c set val=repeat('b',6000) where id=2,` the update only changes non-key fields, and the new record should always land on neighbour pages.



But if there are queries that update key, records can move with delete marking. 
For example UPDATE c SET id = 20 WHERE id = 5;

zhry110: Whether it is PS-9092 or PS-9297, the necessary condition to trigger the problem is that the same physical record is moved between adjacent pages. But UPDATE c SET id = 20 WHERE id = 5; actually del-marks the record with id=5 and inserts the record with id=20 within a transaction. These two records are not physically the same record. Therefore I think this scenario is not affected by this problem, the ReadView of the read transaction will ensure that the correct version of the record is visible.

PR #5357, opened 2024-07-18
PS-9218 merge: Merge MySQL 8.4.0 (partition_names fixes)

percona-ysorokin:
https://perconadev.atlassian.net/browse/PS-9218

Reworked  "table_list->partition_names = &alter_info->partition_names;"
part from the WL#15517
"[5] HeatWave Support for InnoDB Partitions - Support for Loading Partitions"
patch introduced in https://github.com/mysql/mysql-server/commit/951ffba.

It turned out that changing this 'partition_names' affects
"ALTER TABLE ... ALGORITHM=INPLACE, DROP PARTITION ..." logic
under '--rocksdb-allow-unsafe-alter=ON'.
In particular, in the following call stack

1. 'mysql_inplace_alter_table()'
2. 'open_table()' from the
/*
  Tell the SE that the changed table in the data-dictionary.
  For engines which don't support atomic DDL this needs to be
  done before trying to rename the table.
*/
3. 'table->part_info->set_partition_bitmaps()' from the
/*
  Position for each partition in the bitmap is read from the Handler_share
  instance of the table. In MYSQL_OPEN_NO_NEW_TABLE_IN_SE mode, table is not
  opened in the SE and Handler_share instance for it is not created. Hence
  skipping partitions bitmap setting in the MYSQL_OPEN_NO_NEW_TABLE_IN_SE
  mode.
*/
4. 'set_read_partitions()'
5. 'add_named_partition()'

if the "table_list->partition_names = &alter_info->partition_names;" assignment
is made, we will get directly into
'my_error(ER_UNKNOWN_PARTITION, MYF(0), part_name, table->alias);'
as 'partition_names' will include the list of partitions mentioned in the
"DROP PARTITION ..." clause and 'find_or_nullptr()' will not be able to find
already dropped partition in the 'partition_name_hash'.

Fixed by saving the old value of 'partition_names' before calling
'secondary_engine_unload_table()' and restoring it afterwards.

Stabilized 'rocksdb.partition' MTR test case by adding '--sorted_result' to
'SHOW TABLES".

### DIFF
--- a/sql/sql_table.cc
+++ b/sql/sql_table.cc
@@ -43,6 +43,7 @@
 #include <string>
 #include <string_view>
 #include <type_traits>
+#include <utility>
 
 /* HAVE_PSI_*_INTERFACE */
 #include "my_psi_config.h"  // IWYU pragma: keep
@@ -17502,10 +17503,13 @@ bool mysql_alter_table(THD *thd, const char *new_db, const char *new_name,
                                   &table_def))
       return true;
 
-    table_list->partition_names = &alter_info->partition_names;
-    if (secondary_engine_unload_table(
-            thd, table_list->db, table_list->table_name, *table_def, false))
-      return true;
+    const auto old_partition_names = std::exchange(
+        table_list->partition_names, &alter_info->partition_names);
+    auto unload_res = secondary_engine_unload_table(
+        thd, table_list->db, table_list->table_name, *table_def, false);
+    table_list->partition_names = old_partition_names;
+
+    if (unload_res) return true;
   }
 
   /*
--- a/mysql-test/suite/rocksdb/t/partition.test
+++ b/mysql-test/suite/rocksdb/t/partition.test
@@ -35,6 +35,7 @@ SELECT PARTITION_NAME FROM information_schema.rocksdb_ddl WHERE TABLE_NAME="VAR_
 
 --enable_query_log
 
+--sorted_result
 SHOW TABLES;
 
 SELECT * FROM t1 ORDER BY i LIMIT 10;


sql/sql_table.cc
```diff
@@ -17502,7 +17502,6 @@ bool mysql_alter_table(THD *thd, const char *new_db, const char *new_name,
                                   &table_def))
       return true;
 
-    table_list->partition_names = &alter_info->partition_names;
```
dlenev, line 17505: Unfortunately, we can't be sure that secondary_engine_unlock_table() doesn't use TABLE_LIST::partition_names because we don't have access to code of Oracle's Heatwave/Rapid SE which is called from this routine. Moreover, it is likely that it does use this info, even though it is rather awkward from SQL-layer/SE-isolation point of view, because I can see TABLE_LIST::partition_names being assigned in other places the code calls secondary load/unload code.

So I think that it is better to keep this line if we want to be compatible with Oracle.
Instead we can save original table_list->partition_names value on the stack and restore it after calling secondary_engine_unload().

dlenev (changes requested):
Hello Yura!

I have a comment about suggested change and alternative proposal for how to fix the problem. What do you think?

sql/sql_table.cc
dlenev: Interesting... the semantics of std::exchange() is not that obvious from its name...

I think it is OK to push the patch now.

dlenev (approved):
Hello Yura!

Thank you for addressing my previous comment. The patch looks OK to me now.

sql/sql_table.cc
percona-ysorokin: Yeah, I had the same impression when I saw this "exchange" name for the first time :)

PR #5364, opened 2024-07-26
PS-9237 feature: Include support for utf8mb4_0900_ai_ci in MySQL 5.7 

percona-ysorokin:


percona-ysorokin:
This needs to be rebased on top of `release-8.0.39-30`

sql/sql_connect.cc
```diff
@@ -670,6 +670,24 @@ void reset_mqh(THD *thd, LEX_USER *lu, bool get_them = false) {
...
+  // if the 8.0 client sets 'MYSQL_SET_CHARSET_NAME' option to 'utf8mb4' or
+  // leaves it empty, basically meaning the same, this function will be called
+  // with 'cs_number' equal to 255 (meaning 'utf8mb4_0900_ai_ci')
+
+  // at the same time, if 'default_collation_for_utf8mb4' is set to something
+  // other than default 'utf8mb4' collation ('utf8mb4_0900_ai_ci', number 255),
+  // we need to fix 'cs_number' here by setting it to the corresponding number
+  // of 'default_collation_for_utf8mb4' (currently only 'utf8mb4_general_ci',
+  // number 45, is supported)
+  const auto *primary_utf8mb4_collation =
```
dlenev: As far as I can see in other places where we check if default_collation_for_utf8mb4 needs to kick in (e.g. see sql_lex.cc) we directly use my_charset_utf8mb4_0900_ai_ci instead of getting access to it though get_charset_by_csname().

IMO it makes sense to be consistent and do the same here... Especially since this code seems to be called for each connect so saving even a few CPU cycles would be nice.

What do you think?

dlenev: Also I think that instead of two ifs you can simply do:
```
if (cs_number == my_charset_utf8mb4_0900_ai_ci.number)
  cs_numer = thd->variables.default_collation_for_utf8mb4->number;
```

What do you think?

dlenev (changes requested):
Hello Yura!

I would like to suggest a few changes to this patch:

sql/sys_vars.cc
```diff
@@ -1869,6 +1869,21 @@ static bool check_charset(sys_var *, THD *thd, set_var *var) {
...
         return true;
       }
+      // if 'default_collation_for_utf8mb4' is set to something other than
+      // default 'utf8mb4' collation ('utf8mb4_0900_ai_ci') and if the value
+      // returned by 'get_charset_by_csname()' is also default 'utf8mb4'
+      // collation ('utf8mb4_0900_ai_ci'), meaning that were requesting for
+      // 'utf8mb4', we need to fix the returned value depending on the value of
+      // 'default_collation_for_utf8mb4' (currently, only 'utf8mb4_general_ci'
+      // is possible)
+      const auto *primary_utf8mb4_collation =
```
dlenev: The same comment as for previous patch:

Perhaps it is better simply do:

```
if (var->save_result.ptr == &my_charset_utf8mb4_0900_ai_ci)
  var->save_result.ptr = thd->variables.default_collation_for_utf8mb4;
``` 
 instead ?
 
 What do you think?

sql/sql_connect.cc
percona-ysorokin: Yes, @dlenev I had exactly the same doubts and was going back and forth with this. From one side I did not want to hardcode `primary_utf8mb4_collation` to be `my_charset_utf8mb4_0900_ai_ci` (because who knows may be in the next version the default collation will change again). On the other hand, I totally agree that establishing the connection is a critical path and we should not add any unnecessary cycles here.
Anyway, if this caught your attention as well, then it is probably more serious than I thought.
Let's wait for the final feedback from the customer and I will add the changes you suggested into the final patch.

sql/sys_vars.cc
```diff
@@ -1869,6 +1869,21 @@ static bool check_charset(sys_var *, THD *thd, set_var *var) {
         my_error(ER_UNKNOWN_CHARACTER_SET, MYF(0), err.ptr());
         return true;
       }
+      // if 'default_collation_for_utf8mb4' is set to something other than
```
dlenev, line 1872: What about similar code in check_collation_not_null() ?
In theory one can do `SET @@global.collation_connection = utf8mb4`.
Perhaps we should handle this case as well?
Or at least add comment why we don't think it is necessary/don't want to do it (indeed it might be non-trivial to distinguish cases when one uses charset name only -  `utf8mb4` vs case when one specifies collation explicitly -`utf8mb4_0900_ai_ci`.

dlenev: After a bit more pondering about it, I think it is a good idea to have test coverage for cases when user can use both charset name and collation name. Like `SET NAMES utf8mb4;` vs `SET NAMES utf8mb4 COLLATE utf8mb4_0900_ai_ci` and `SET @@global.collation_connection = utf8mb4` vs `SET @@global.collation_connection = utf8mb4_0900_ai_ci`.

Do you agree?

dlenev (changes requested):
A bit more comments about second and third parts of your patch.

Otherwise it looks OK to me.

sql/sql_connect.cc
percona-ysorokin: @dlenev I reworked the critical paths code with simplified versions
```c++
  if (thd->variables.default_collation_for_utf8mb4 !=
      &my_charset_utf8mb4_0900_ai_ci) {
    if (client_cs == &my_charset_utf8mb4_0900_ai_ci) {
      client_cs = thd->variables.default_collation_for_utf8mb4;
    }
  }
```
that does not involve charset by name lookup.

sql/sys_vars.cc
percona-ysorokin: Reworked the same way.

sql/sys_vars.cc
percona-ysorokin: @dlenev I extended coverage with the following statements
```sql
SET character_set_client = utf8mb4;
SET NAMES DEFAULT;
SET NAMES utf8mb4;
SET NAMES utf8mb4 COLLATE utf8mb4_general_ci;
SET CHARACTER SET DEFAULT;
SET CHARACTER SET utf8mb4;
SET collation_connection = utf8mb4_general_ci;
```
Note that 'SET NAMES utf8mb4 COLLATE utf8mb4_0900_ai_ci' will generate
`character_set_client=255` in the binary log - I believe this is correct behavior as uses pecifies the collation explicitly.

Likewise, 'SET collation_connection = utf8mb4_0900_ai_ci' will generate
`character_set_client=255` in the binary log.

As for 'SET collation_connection = utf8mb4' (when we try to assign a character set name to a collation variable), there is now problem here as this statatement is considered
syntactically incorrect In other words, no need to change `check_collation_not_null()`.

hulyav:
Not needed anymore

PR #5368, opened 2024-08-02
PS-9314 fix: Signal 11 when using JSON_TABLE

percona-ysorokin:
https://perconadev.atlassian.net/browse/PS-9314

As part of the 8.0.22 merge we added our fix (commit e29ddfe) for Oracle's Bug #101401
"ISNULL continously returns the same value for LOAD_FILE" (https://bugs.mysql.com/bug.php?id=101401)
However, when Oracle fixed the same issue in 8.0.23, Bug #32096341
"ISNULL continuously returns the same value for LOAD_FILE" (commit mysql/mysql-server@de4b3f7), our fix never got reverted, which introduced this regression.

Fixed by reverting our patch.

Extended 'main.func_str' MTR test case with additional checks for this particular regression.

### DIFF
--- a/mysql-test/r/func_str.result
+++ b/mysql-test/r/func_str.result
@@ -5299,6 +5299,11 @@ is null
 DROP PROCEDURE p1;
 DROP TABLE t1;
 #
+# PS-9314: Signal 11 when using JSON_TABLE
+# (https://perconadev.atlassian.net/browse/PS-9314)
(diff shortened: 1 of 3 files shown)


percona-ysorokin:
This PR is created on top of `release-8.0.37-29` for review purposes only. This needs to be changed to `release-8.0.38-30` once this branch is available.

percona-ysorokin:
@kamil-holubicki Just to let you know, as it was your patch originally.

dlenev (approved):
LGTM!

percona-ysorokin:
Re-created as https://github.com/percona/percona-server/pull/5383 with GCA in mind.

PR #5369, opened 2024-08-02
PS-9301 Backport bug fixes from MySQL 8.0.38

VarunNagaraju:
https://perconadev.atlassian.net/browse/PS-9301

- Fixes a buffer overflow error in Optimizer.
- Fixes an assertion failure: dict0dict.ic:1102:col_index < table->n_cols.
- Improves the binlog cache empty check.
- Fixes the table and database name checks in sql/item.cc:Item_ident::print method.
- Disables spatial index for index-only reads.
- Improves the connection control plugin instrumentation to be visible in performance schema.
- PSI_THREAD_INFO of performance_schema.threads table is now updated properly for prepared statements.
- Fixes the double processing of FTS index words issue when innodb_optimize_fulltext_only is enabled.


dlenev (commented):
Hello Varun!

I have looked at Optimizer-related changes in this PR, specifically at:
1) https://github.com/percona/percona-server/pull/5369/commits/194f9f83c697e6fd7d5e7faf10db897bf008f510
2) https://github.com/percona/percona-server/pull/5369/commits/69505e1e98ed35a35463f15b87de065bbe32ee67

Regarding the first change - I think it is not strictly necessary to backport this change as we are not releasing on platforms which support --lower-case-table-names=2 (MacOS X, Windows) by default. OTOH, in theory, this mode can even be used on Linux if special FS is used though this is really unlikely... So please think about whether you really want to backport this commit once again.

In any case I think it is safe to backport this change.

Regarding the second change - it looks OK to me.

storage/innobase/include/dict0dict.ic
```diff
@@ -1813,7 +1813,7 @@ dict_table_get_index_on_first_col(
         ulint                   col_index)	/*!< in: position of column
 						in table */
 {
-	ut_ad(col_index < table->n_cols);
+	ut_a(col_index < table->n_t_cols);
```
dlenev: I think it is better to keep it as debug-only assert in 5.7 (keep it ut_ad()). Introducing new assert in non-debug builds in 5.7 at this stage doesn't sound like a good idea from risk management point of view to me.

dlenev (changes requested):

storage/innobase/include/dict0dict.ic
dlenev: And there is more high-level question about why are we backporting fix for something that affects only debug builds.

dlenev: I am not quite convinced about backporting some other patches either. I guess we need to have a discussion about criteria for such backports.

But patches themselves look OK to me.

VarunNagaraju:
https://ps57.cd.percona.com/view/5.7/job/percona-server-5.7-param/625/

VarunNagaraju:
MTR tests were also run(locally) with ASAN and Valgrind enabled and no new failures were observed.

PR #5374, opened 2024-08-06
PS-8.0 fix bug: when do general_log->set_file() will change the opt_slow_logname's value

MrRobertYuan:
File_query_log::set_rotated_name() will change opt_slow_logname.
But it will be called by File_query_log::set_file, and File_query_log::set_file will not only be called by slow_log but also general_log, which will cause an error to set the opt_slow_logname to general_log's rotate name.
So fix this by compare its m_log_type and do it only when it is a slow_log


oleksandr-kachan:
Dear Mr Robert Yuan,

Thank you for contributing to Percona Server. Your fix will be included into 8.0.39 and 8.4.2 releases.

PR #5382, opened 2024-08-23
PS-8057: Fix initialization of general and slow log file names (8.0.39)

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-8057

In case general and slow log file names are provided in config and slow log rotation is enabled, slow_query_log_file system variable gets initialized with general log name.

The reason of the issue is that code responsible for slow and general logs handling have many in common and during initial log names initialization code responsible for slow log rotation was ivoked while initializing general log name.

To fix the issue there was a checkup added to make sure code related to slow log rotation is not invoked while general log file name is being processed.

### DIFF
--- a/mysql-test/r/percona_slow_general_log_names.result
+++ b/mysql-test/r/percona_slow_general_log_names.result
@@ -1,0 +1,6 @@
+select @@global.slow_query_log_file;
+@@global.slow_query_log_file
+mysqld.slowlog
+select @@global.general_log_file;
+@@global.general_log_file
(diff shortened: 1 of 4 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.x-pipeline-parallel-mtr/259/

lukin-oleksiy (approved):

percona-ysorokin (approved):
LGTM

PR #5385, opened 2024-08-27
Merge fixes for PS-9144 and PS-9214 into 8.0 release tree.

dlenev:
See individual commits for proper change descriptions.

### DIFF
--- a/storage/innobase/row/row0pread.cc
+++ b/storage/innobase/row/row0pread.cc
@@ -214,6 +214,27 @@ class PCursor {
   PCursor(btr_pcur_t *pcur, mtr_t *mtr, size_t read_level)
       : m_mtr(mtr), m_pcur(pcur), m_read_level(read_level) {}
 
+  /**
+    Restore position of cursor created from Scan_ctx::Range object.
+    Adjust it if starting record of range has been purged.
+  */
+  void restore_position_for_range() noexcept {
+    /*
+      Restore position on the record or its predecessor if the record
+      was purged meanwhile. In the latter case we need to adjust position by
+      moving one step forward as predecessor belongs to another range.
+
+      We rely on return value from restore_position() to detect this scenario.
+      This can be done when saved position points to user record which is
+      always the case for saved start of Scan_ctx::Range.
+    */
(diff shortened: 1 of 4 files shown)


satya-bodapati:
@dlenev Thanks for updating the commit messages. I will take a look.

storage/innobase/row/row0pread.cc
```diff
@@ -345,17 +344,53 @@ void PCursor::resume() noexcept {
...
+  merge or simply inserted to our page concurrently).
+  The latter case is detected by caller by doing !is_after_last_on_page()
+  check and instead of doing switch to the next page we continue processing
+  from the restored user record. */
+
+  m_pcur->restore_position(BTR_SEARCH_LEAF, m_mtr, UT_LOCATION_HERE);
+
+  ut_ad(!m_pcur->is_after_last_on_page());
+
+  if (m_supremum_saved) m_pcur->move_to_next_on_page();
```
satya-bodapati, line 390: Should we reset this m_supremum_saved just to be safe?

satya-bodapati (approved):
Both the fixes looks good to me and easy to understand with the new examples.

Thank you. I just left only one small question about resetting the variable.

dlenev:
Jenkins results look OK to me: https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/677/#showFailuresLink

storage/innobase/row/row0pread.cc
dlenev: Good question... I thought about this for some time...

It seems that we never call resume() more than once after calling savepoint() in Parallel_reader code so resetting it here should not cause problems. OTOH it is not strictly necessary because each call to resume() must be preceded by savepoint() which sets this flag. Also it is interesting that btr_pcur_t::restore_position() doesn't reset all those members set by store_position() (it actually takes some measures to keep them up-to-date in pessimistic case), so at least in theory one might use this fact in future and call restore_position() more than once for the same store_position().

Taking into account all the above I would rather keep the above code as is - just to be on the safe side :)

PR #5395, opened 2024-09-03
PS-9328: Merge 8.4.2 - Fix innodb.parallel_read_kill test failures.

dlenev:
After Upstream's fix for bug#35952353 "SELECT COUNT(*) degraded performance on 8.0 compared to 5.7" (https://github.com/mysql/mysql-server/commit/ffad6cf) MySQL stopped using parallel read for COUNT(*) evaluation when secondary index scan is used.

As result innodb.parallel_read_kill test started to fail, as it employed COUNT(*) that used secondary index scan to test InnoDB parallel read killability. The parallel read code is simply no longer invoked causing DEBUG_SYNC time out.

This patch fixes the failure by forcing the test to use COUNT(*) that uses primary key scan, which is still processed using parallel read, instead.


percona-ysorokin (approved):
LGTM
(please, port this fix to 8.0.39 as well)

dlenev:
Will be pushed to both 8.0 and 8.4 release branches using GCA.

PR #5396, opened 2024-09-03
PS-9293, PS-9328: Fix innodb.parallel_read_kill test failures (8.0 version).

dlenev:
After Upstream's fix for bug#35952353 "SELECT COUNT(*) degraded performance on 8.0 compared to 5.7" (https://github.com/mysql/mysql-server/commit/ffad6cf) MySQL stopped using parallel read for COUNT(*) evaluation when secondary index scan is used.

As result innodb.parallel_read_kill test started to fail, as it employed COUNT(*) that used secondary index scan to test InnoDB parallel read killability. The parallel read code is simply no longer invoked causing DEBUG_SYNC time out.

This patch fixes the failure by forcing the test to use COUNT(*) that uses primary key scan, which is still processed using parallel read, instead.


dlenev:
See https://github.com/percona/percona-server/pull/5395 for review/approval.

percona-ysorokin (approved):
LGTM

PR #5398, opened 2024-09-04
PS-9286 KMIP key activation implementation (8.0)

lukin-oleksiy:
https://perconadev.atlassian.net/browse/PS-9286

extra/libkmip head is at baf6832 with activate operation implementation. keyring_kmip/backend/backend.cc updateed to execute op_activate after key registration.


percona-ysorokin (approved):
LGTM

PR #5401, opened 2024-09-05
PS-9293, PS-9328: Revert of fix for PS-7221: Bug #100402 (crash on se… (8.0 version).

dlenev:
…lect Sys_var_tz::session_value_ptr)

Reverting Percona Server's version of the fix which is no longer necessary. Upstream fixed this issue in a different, more complex, but also probably more clean way in their fix for bug#31168097 "SIG 6 ASSERTION `0' IN TIME_ZONE_UTC::GET_NAME AT TZTIME.CC:1182".

See: https://github.com/mysql/mysql-server/commit/1f8c02ec0cf40bc70ba074a1b83882e11228b9f6

The test case for this issue is kept around.


percona-ysorokin (approved):
LGTM

PR #5402, opened 2024-09-05
PS-9293, PS-9328: Revert of fix for PS-7221: Bug #100402 (crash ... (8.4 version)

dlenev:
See individual commit for details.


percona-ysorokin (approved):
LGTM

PR #5407, opened 2024-09-06
PS-9379 Fix problem with "-latomic" on some platforms in uuid_vx (8.0)

lukin-oleksiy:
…nent

https://perconadev.atlassian.net/browse/PS-9379

Fix of problem with -latomic in uuid_vx component of some platofrms

### DIFF
--- a/components/uuid_vx_udf/CMakeLists.txt
+++ b/components/uuid_vx_udf/CMakeLists.txt
@@ -33,6 +33,26 @@ MYSQL_ADD_COMPONENT(uuid_vx_udf
 
 target_include_directories(component_uuid_vx_udf SYSTEM PRIVATE ${BOOST_PATCHES_DIR} ${BOOST_INCLUDE_DIR})
 
-if(NOT APPLE)
-  target_link_libraries(component_uuid_vx_udf PRIVATE atomic)
+# Check if -latomic is required or not
+if (NOT MSVC)
+  include(CheckCXXSourceCompiles)
+  set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+# Adding "-w" compiler option to suppres all possible warnings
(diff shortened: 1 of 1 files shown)


components/uuid_vx_udf/CMakeLists.txt
```diff
@@ -33,6 +33,19 @@ MYSQL_ADD_COMPONENT(uuid_vx_udf
 
 target_include_directories(component_uuid_vx_udf SYSTEM PRIVATE ${BOOST_PATCHES_DIR} ${BOOST_INCLUDE_DIR})
 
-if(NOT APPLE)
-  target_link_libraries(component_uuid_vx_udf PRIVATE atomic)
+# Check if -latomic is required or not
```
percona-ysorokin, line 36: @lukin-oleksiy, I like this approach for detecting the same from `protobuf` more.
https://github.com/percona/percona-server/blob/646a7cf4b6874eb4a9163078febc13c7938328e1/extra/protobuf/protobuf-24.4/CMakeLists.txt#L184
1. They include `CheckCXXSourceCompiles` CMake include for `CHECK_CXX_SOURCE_COMPILES`
   `include(CheckCXXSourceCompiles)`
2. They save and restore `CMAKE_REQUIRED_FLAGS` before and after modifying it
   `set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})`
   ...
   `set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})`
3. they do single line
   `return static_cast<int>(std::atomic<int64_t>{});`
   so, no porential `unused variable` warnings that may affect compilation results.

percona-ysorokin (changes requested):

components/uuid_vx_udf/CMakeLists.txt
lukin-oleksiy: Solution is somewhere in the middle, with probe code from rocsdb

percona-ysorokin (approved):
LGTM

PR #5410, opened 2024-09-09
PS-9328: Merge 8.4.2 : Tentative fix for binlog_mysqlbinlog_4g_start_…

dlenev:
…position test failures.


dlenev:
Decided not to proceed with this approach as even after this change test in question still sometimes fails in Jenkins on some platforms (takes too much time or too much memory).

PR #5412, opened 2024-09-09
PS-9293, PS-9328: Disabled binlog.binlog_mysqlbinlog_4g_start_position (8.0 version)

dlenev:
…n rocksdb_stress.drop_cf_stress tests.

Disabled two tests which create too much noise in Jenkins and which we could not/do not plan to fix immediately.

binlog.binlog_mysqlbinlog_4g_start_position test requires too much memory/ space so often fails in Jenkins even when run alone.

rocksdb_stress.drop_cf_stress test executes concurrently mix of DDL and often fails due to race.


percona-ysorokin (approved):
LGTM

PR #5414, opened 2024-09-10
PS-9306: MySQL 8.0.38, 8.4.1 and 9.0.0 crashes on restart if database…

satya-bodapati:
… has 10K tables or more

Upstream fixed the problem in 8.0.39 release but no testcase is attached to the bug fix.

This commit introduces a mtr testcase. It creates 8192 tables, restarts the server and it should come up

### DIFF
--- a/mysql-test/suite/innodb/t/percona_create_table_8k.test
+++ b/mysql-test/suite/innodb/t/percona_create_table_8k.test
@@ -1,0 +1,88 @@
+--echo #
+--echo # PS-9306 : MySQL 8.0.38, 8.4.1 and 9.0.0 crashes on restart if database has 10K tables or more
+--echo #
+
+CREATE DATABASE test_8k;
+--source include/count_sessions.inc
+
+DELIMITER |;
+CREATE PROCEDURE create_tables(start_index INT, end_index INT)
+BEGIN
+    DECLARE i INT;
+    SET i = start_index;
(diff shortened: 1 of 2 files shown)


mysql-test/suite/innodb/t/percona_create_table_8k.test
```diff
@@ -0,0 +1,88 @@
+--echo #
```
percona-ysorokin, line 1: Please mark this MTR test case as `big-test`

percona-ysorokin (changes requested):

mysql-test/suite/innodb/t/percona_create_table_8k.test
dlenev: Actually it might make sense to *also* mark it as requiring standalone execution - see not_parallel.inc, so it is not run concurrently with other big tests (I assume it requires quite some resources?)

satya-bodapati: Do we run not_parallel tests on jenkins?

satya-bodapati: @dlenev I verified that it indeed scheduled by one last worker in mtr. So we do not need special run initiated by Jenkins.

satya-bodapati:
closing this one. recreating another one GCA: #5418

PR #5415, opened 2024-09-10
PS-9322 fix: With super_read_only=1, undo truncation cannot update DD and leaves orphan truncate log files (8.0)

percona-ysorokin:
https://perconadev.atlassian.net/browse/PS-9322

Backported Oracle fix for Bug #35784192
"tablespace is missing for table innodb_undo_002"
(commit mysql/mysql-server@bda098a and mysql/mysql-server@97f4285) from 8.4 to 8.0.

Bug#35784192 tablespace is missing for table innodb_undo_002

This bug reports following two problems:

Querying the information schema reports undo tablespace is missing -
  while an undo file is present.
Purge threads don't remove the undo_space_number_trunc.log file.

While the first problem is already fixed through Bug#32104924, Bug#32654667

This bug fix aims to address the second problem.

The purge thread attempts to lock dictionary tables, while truncating the undo tablespace. This lock can't be acquired in a read-only MySQL instance, except when the thread has some special flags associated with it. In this case, we use the skip_readonly_check member of the thread descriptor struct, to signal that the readonly check should not be applied to the purge thread. This is done via the set_skip_readonly_check() method.

Change-Id: Iea6255ff6654752db8658bdd422cc3e58e3a8a97

Bug#35784192 tablespace is missing for table innodb_undo_002 [postfix] Remove MySQL versions number from MTR test.

Change-Id: I3973435edbbc8bf442846c617ade4b5b824bb8c2

### DIFF
--- a/mysql-test/suite/innodb_undo/r/purge_on_replica.result
+++ b/mysql-test/suite/innodb_undo/r/purge_on_replica.result
@@ -1,0 +1,49 @@
+include/master-slave.inc
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
+Note	####	Storing MySQL user name or password information in the connection metadata repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START REPLICA; see the 'START REPLICA Syntax' in the MySQL Manual for more information.
+[connection master]
+# 1: Set replica parameters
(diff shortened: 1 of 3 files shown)


satya-bodapati (approved):

percona-ysorokin:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-param-parallel-mtr/63/

PR #5416, opened 2024-09-10
PS-9322 fix: With super_read_only=1, undo truncation cannot update DD and leaves orphan truncate log files (8.4)

percona-ysorokin:


satya-bodapati:
This is NULL merge from 8.0, as the fix is already in upstream 8.4

satya-bodapati (approved):

PR #5418, opened 2024-09-11
PS-9306: MySQL 8.0.38, 8.4.1 and 9.0.0 crashes on restart if database…

satya-bodapati:
… has 10K tables or more

https://perconadev.atlassian.net/browse/PS-9306

Upstream fixed the problem in 8.0.39 release but no testcase is attached to the bug fix.

This commit introduces a mtr testcase. It creates 8192 tables, restarts the server and it should come up


percona-ysorokin (approved):
LGTM

PR #5421, opened 2024-09-12
PS-9328: Disabled rocksdb.non_blocking_manual_compaction test (8.4 version)

dlenev:
Disabled rocksdb.non_blocking_manual_compaction test which often fails in Jenkins due to race in it and which turned out is non-trivial to fix.


percona-ysorokin (approved):
LGTM

PR #5427, opened 2024-09-14
PS-9369: Fix currently processed query comparison in audit_log

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9369

The audit_log uses stack to keep track of table access operations being performed in scope of one query. It compares last known table access query string stored on top of this stack with actual query in audit event being processed at the moment to decide if new record should be pushed to stack or it is time to clean records from the stack.

Currently audit_log simply compares char* variables to decide if this is the same query string. This approach doesn't work. As a result plugin looses control of the stack size and it starts growing with the time consuming memory. This issue is not noticable on short term server connections as memory is freed once connection is closed. At the same time this leads to extra memory consumption for long running server connections.

The following is done to fix the issue:
- Query is sent along with audit event as MYSQL_LEX_CSTRING structure. It is not correct to ignore MYSQL_LEX_CSTRING.length comparison as sometimes MYSQL_LEX_CSTRING.str pointer may be not iniialised properly. Added string length check to make sure structure contains any valid string.
- Used strncmp to compare actual strings instead of comparing char* variables.

### DIFF
--- a/plugin/audit_log/audit_log.cc
+++ b/plugin/audit_log/audit_log.cc
@@ -746,7 +746,7 @@ struct query_stack_frame {
   /* number of accessed databases */
   int databases_accessed;
   /* query */
-  const char *query;
+  MYSQL_LEX_CSTRING query;
 };
 
 struct query_stack {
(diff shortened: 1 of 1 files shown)


oleksandr-kachan:
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-parallel-mtr/692/

percona-ysorokin (approved):
LGTM

PR #5428, opened 2024-09-16
PS-9328: Fix for sporadic index_merge_innodb test failures.

dlenev:
Sporadic index_merge_innodb test failures occurred when InnoDB purge threads were unable to keep up with removing rows which were marked for deletion. The presence of these rows affected row count estimations making EXPLAIN output differ.

Fixed by running OPTIMIZE TABLE to ensure delete-marked rows removal prior to running EXPLAIN that failed.

### DIFF
--- a/mysql-test/include/index_merge1.inc
+++ b/mysql-test/include/index_merge1.inc
@@ -516,7 +516,11 @@ if ($eval_before_analyze)
 {
   eval $eval_before_analyze;
 }
-analyze table t0;
+# We use OPTIMIZE instead of ANALYZE here to handle the case when InnoDB
+# purge threads do not manage to keep up and completely remove records which
+# were marked for deletion earlier, which affects row count estimations in
+# EXPLAIN below.
+optimize table t0;
 -- enable_result_log
 -- enable_query_log
 


mysql-test/include/index_merge1.inc
```diff
@@ -516,7 +516,11 @@ if ($eval_before_analyze)
 {
   eval $eval_before_analyze;
 }
-analyze table t0;
+# We use OPTIMIZE instead of ANALYZE here to handle the case when InnoDB
+# purge threads do not manage to keep up and completely remove records which
+# were marked for deletion earlier, which affects row count estimations in
+# EXPLAIN below.
+optimize table t0;
```
satya-bodapati, line 523: this is a bit of cheating :stuck_out_tongue:  optimize table essentially rebuilds the table in innodb.
May be you should use mysql-test/include/wait_innodb_all_purged.inc?

dlenev: As discussed on Slack - this include file is used for tests for 3 different engines - InnoDB, RocksDB and MyISAM, and I am not sure how simply adding wait_innodb_all_purged.inc will behave in case of the two latter.
So I would stick to OPTIMIZE TABLE. After all the main purpose of this test is to test Optimizer/Execution engine and not the InnoDB. For example, we could have recreated table to gets stable statistics instead...

satya-bodapati: Right. If the intention is not to test the execution or behavior of InnoDB, then rebuilding is OK.

satya-bodapati (approved):

PR #5429, opened 2024-09-16
PS-9328: Fix for sporadic rpl.rpl_json test failures.

dlenev:
Results from JSON_OBJECTAGG aggregate function can depend on the order in which rows are returned by SELECT clause, in cases when it can get duplicate values as its "key" argument. However, such order is not guaranteed in general case and can, for example, depend on the order in which tables are joined.

rpl.rpl_json test has failed sporadically, when one of its queries that uses JSON_OBJECTAGG in such an ambigous context, was executed using different join order.

This fix solves the problem by enforcing fixed join order for this query.


percona-ysorokin (approved):
LGTM

PR #5432, opened 2024-09-17
TEL-26, TEL-29: Add unit tests and MTR tests for Percona Telemetry Component

kamil-holubicki:
https://perconadev.atlassian.net/browse/TEL-26
https://perconadev.atlassian.net/browse/TEL-29

Add Unit tests for mysql plugin
Add MTR tests for telemetry


kamil-holubicki:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/693

mysql-test/suite/component_percona_telemetry/t/delete_obsolete_file.test
```diff
@@ -0,0 +1,40 @@
+# Test that obsolete files from another servers are removed,
+# but files with names not conforming to expected pattern are not.
+
+--source include/have_percona_telemetry.inc
+--source include/force_restart.inc
```
percona-ysorokin, line 5: Do we really need this as we are doing restart with custom telemetry parameters anyway?

mysql-test/suite/component_percona_telemetry/t/invalid_path.test
```diff
@@ -0,0 +1,42 @@
+# Test that Percona Telemetry Component doesn't do any harm to the server when it is not possible
+# to store telemetry file
+
+--source include/have_percona_telemetry.inc
+--source include/force_restart.inc
```
percona-ysorokin, line 5: Likewise

mysql-test/suite/component_percona_telemetry/t/telemetry_disable.test
```diff
@@ -0,0 +1,50 @@
+# Component enable/disable
+
+--source include/have_percona_telemetry.inc
+--source include/force_restart.inc
```
percona-ysorokin, line 4: This one most probably is needed

mysql-test/suite/component_percona_telemetry/t/telemetry_file_creation.test
```diff
@@ -0,0 +1,76 @@
+# Test the telemetry file creation and cleanup
+
+--source include/have_percona_telemetry.inc
+--source include/force_restart.inc
```
percona-ysorokin, line 4: Likewise

unittest/gunit/components/percona_telemetry/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+   
+MY_CHECK_CXX_COMPILER_WARNING("-Wno-unused-local-typedefs" HAS_WARN_FLAG)
+IF(HAS_WARN_FLAG)
+  STRING_APPEND(CMAKE_CXX_FLAGS " ${HAS_WARN_FLAG}")
+ENDIF()
+
+INCLUDE_DIRECTORIES(SYSTEM
```
percona-ysorokin, line 22: Better use `target_include_directories` for individual targets

unittest/gunit/components/percona_telemetry/data_provider-t.cc
```diff
@@ -0,0 +1,439 @@
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#define private public
```
percona-ysorokin, line 4: Dangerous territory...

unittest/gunit/components/percona_telemetry/data_provider-t.cc
```diff
@@ -0,0 +1,439 @@
...
+  Row row{expected_val};
+
+  EXPECT_CALL(dataProvider,
+              do_query(Eq(expected_query), A<QueryResult *>(), _, _))
+      .Times(should_cache_result ? 1 : 2)
+      .WillRepeatedly(DoAll(
+          WithArg<1>(Invoke([&row](QueryResult *qr) { qr->push_back(row); })),
+          Return(false)));
+
+  // KH:
```
percona-ysorokin, line 119: Leftovers

unittest/gunit/components/percona_telemetry/data_provider-t.cc
```diff
@@ -0,0 +1,439 @@
...
+  EXPECT_TRUE(ri_iter->value.HasMember("is_semisync_source"));
+  auto iter = ri_iter->value.FindMember("is_semisync_source");
+  EXPECT_STREQ(iter->value.GetString(), "1");
+
+  EXPECT_TRUE(ri_iter->value.HasMember("is_replica"));
+  iter = ri_iter->value.FindMember("is_replica");
+  EXPECT_STREQ(iter->value.GetString(), "1");
+}
+
+}  // namespace data_provider_unittests
```
percona-ysorokin, line 439: No end of line at EOF

percona-ysorokin (approved):
LGTM
(mostly cosmetic comments apart from "private" -> "public" change)

unittest/gunit/components/percona_telemetry/CMakeLists.txt
kamil-holubicki: I followed the way other unit tests do it. The idea is that each directory of UT can have several target binaries (a loop ath the end of CMakeLists.txt). All of these targets should have common include paths. So they are specified as INCLUDE_DIRECTORIES (for all) instead of TARGET_INCLUDE_DIRECTORIES for every particular target.

We could add `TARGET_INCLUDE_DIRECTORIES(${test}-t PUBLIC ${GMOCK_INCLUDE_DIRS})` in the loop at the end, but it has the same effect as INCLUDE_DIRECTORIES

unittest/gunit/components/percona_telemetry/data_provider-t.cc
kamil-holubicki: Not really. It is only for unit test binary. No danger of propagating it into the real telemetry *.so.
This trick is used in other unit tests as well. Allows testing of private methods. Here it is necessary, as we have only one public method which does everything by calling private one by one. Creating unit test for this public method makes no sense, so better to test private methods one by one.

unittest/gunit/components/percona_telemetry/data_provider-t.cc
kamil-holubicki: done

unittest/gunit/components/percona_telemetry/data_provider-t.cc
kamil-holubicki: done

mysql-test/suite/component_percona_telemetry/t/telemetry_file_creation.test
kamil-holubicki: done

mysql-test/suite/component_percona_telemetry/t/invalid_path.test
kamil-holubicki: done

mysql-test/suite/component_percona_telemetry/t/delete_obsolete_file.test
kamil-holubicki: The idea here was driven by the fact that MTR framework doesn't restart mysqld between tests if not necessary. Here we change the default times for telemetry gathering, so restarting would reset these times to the defaults. But on the other hand having shorter times should not cause any problems.

unittest/gunit/components/percona_telemetry/CMakeLists.txt
percona-ysorokin: My comment mostly comes from the Modern CMake best practices - always use target-based directives. I believe the main idea behind them is to avoid CMake context pollution. For instance, if you do `INCLUDE_DIRECTORIES` here, then every other `CMakeLists.txt` file you directly or indirectly (via subdirs, for instance) include from this one will inherit these new includes (even when they are not necessarily needed). But I agree, in this particular case, we are not doing anything like this and this does not really matter - you can ignore my comment.

PR #5435, opened 2024-09-18
PS-9328 [8.4]: Revert fixes for unsupported gcc and clang versions

inikep:
Percona Server 8.4.x supports only clang-12, gcc-10 or newer.


dlenev (approved):
LGTM!

PR #5448, opened 2024-10-09
TEL-26: Percona Telemetry Component unit tests added

kamil-holubicki:
Post push fix.

1. Added missing Boost dependency. It caused unit tests build to fail if no Boost was installed in system.
2. Removed unnecessary component dependency. It caused unit tests build to fail if configured with WITH_PERCONA_TELEMETRY=NO


dlenev (approved):
LGTM.

PR #5449, opened 2024-10-10
PS-9190: JS Stored Routines: basic functionality

dlenev:
See individual commits for detailed commit descriptions.


mysql-test/suite/component_js_lang/t/js_lang_big.test
```diff
@@ -0,0 +1,97 @@
+#
+# JS language component tests which require lot of memory or CPU.
+#
```
percona-ysorokin, line 3: The test is named "_big" but there is no `--source include/big_test.inc`

mysql-test/include/plugin.defs
```diff
@@ -210,3 +210,4 @@ component_binlog_utils_udf          plugin_output_directory  no  BINLOG_UTILS_UD
 component_encryption_udf            plugin_output_directory  no  ENCRYPTION_UDF_COMPONENT
 component_masking_functions         plugin_output_directory  no  MASKING_FUNCTIONS_COMPONENT
 component_percona_udf               plugin_output_directory  no  PERCONA_UDF_COMPONENT
+component_js_lang                   plugin_output_directory  no  JS_LANG_COMPONENT
```
percona-ysorokin, line 213: Most probably, we need to add `component_js_lang` to the list of the default test suites in `mysql-test-run.pl`

components/js_lang/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+
+MYSQL_ADD_COMPONENT(js_lang
+  component.cc
+  js_lang_core.cc
+  js_lang_param.cc
+  LINK_LIBRARIES v8_monolith
+  MODULE_ONLY
+)
+
+target_include_directories(component_js_lang SYSTEM PRIVATE ${BOOST_PATCHES_DIR} ${BOOST_INCLUDE_DIR} ${V8_INCLUDE_DIR})
```
percona-ysorokin: Use `target_link_libraries(component_js_lang extra::boost)` for Boost directories (patches are handled by the CMake interface library)

percona-ysorokin: Ideally it would be nice to have `v8.cmake` similar to, say, `boost.cmake` where we check all the provided directories (`V8_INCLUDE_DIR` and `V8_LIB_DIR `) and creates an interface library `extra::v8`.

components/js_lang/js_lang_common.h
```diff
@@ -0,0 +1,102 @@
...
+/*
+  Some global constants and defines used through the component.
+*/
+
+// defined as a macro because needed both raw and stringized
+#define CURRENT_COMPONENT_NAME js_lang
+#define CURRENT_COMPONENT_NAME_STR BOOST_PP_STRINGIZE(COMPONENT_NAME)
+
+// Name/Identifier of the language this component implements.
+static constexpr const char *LANGUAGE_NAME = "JS";
```
percona-ysorokin: I prefer `constexpr std::string_view x = "blah-blah";` for everything except places where we need null-terminated strings. In latter case a constexpr char array `constexpr char y[] = "another-blah-blah"`.

components/js_lang/js_lang_common.h
```diff
@@ -0,0 +1,102 @@
...
+extern REQUIRES_SERVICE_PLACEHOLDER(mysql_thd_security_context);
+extern REQUIRES_SERVICE_PLACEHOLDER(mysql_thd_store);
+
+/**
+  Helper function which is used to absorb results of service calls
+  which are not supposed to fail.
+
+  @sa Note about our approach to error handling in  js_lang_core.cc.
+*/
+static void always_ok(bool service_result) { assert(!service_result); }
```
percona-ysorokin: `static` functions in a header file some people consider `bad practice`.
`inline` is better in my opinion.

components/js_lang/js_lang_core.h
```diff
@@ -0,0 +1,497 @@
...
+/**
+  Wrapper class around v8::Isolate which simplifies its lifetime management
+  through reference counting pointers by packaging it with its allocator.
+
+  It also contains static methods for V8 initialization/shutdown as well as
+  global reference counter for isolates, so we can block component shutdown
+  if there are connections around that use isolates and therefore V8.
+*/
+class Js_isolate {
+ public:
```
percona-ysorokin, line 96: Please mark copy constructor and assignment operator as `deleted`.
Implement proper move constructor and move assignment operator if necessary. If not, mark them as `deleted` as well

components/js_lang/js_lang_core.h
```diff
@@ -0,0 +1,497 @@
...
+  Js_isolate(v8::Isolate *iso, v8::ArrayBuffer::Allocator *alloc)
```
percona-ysorokin, line 98: I think this constructor needs to be made `private` if the main way to create this object is via static `create()` function.

percona-ysorokin: Also, I would extract `v8_init()` and `v8_shutdown()` function into a separate class `Js_platform` of something

components/js_lang/js_lang_core.h
```diff
@@ -0,0 +1,497 @@
...
+
+    return &(i->second);
+  }
+
+  /** Checks if connection executes CREATE FUNCTION or PROCEDURE statement. */
+  bool is_create_routine_command() {
+    mysql_cstring_with_length sql_command;
+    always_ok(mysql_service_mysql_thd_attributes->get(m_thd, "sql_command",
+                                                      &sql_command));
+    return (strcmp(sql_command.str, "create_procedure") == 0 ||
```
percona-ysorokin: This does not take into account that `sql_command.length`.
May be `std::string_view(sql_command.str, sql_command.length) == "create_procedure"`

components/js_lang/js_lang_core.h
```diff
@@ -0,0 +1,497 @@
...
+
+    Also we use separate JS context for each connection (and separate contexts
+    for execution of routines under different users within the same connection).
+
+    Routines which are executed under the same user in the same connection
+    share the same JS context.
+
+    TODO: Make it a class once support for KILLability/OOM handling is added.
+  */
+  struct Auth_id_context {
```
percona-ysorokin, line 228: Are all the members public intentional?

components/js_lang/js_lang_core.h
```diff
@@ -0,0 +1,497 @@
...
+  // UNINSTALL -> INSTALL scenario).
+  static std::atomic<int> s_count;
+};
+
+/**
+  Component's per connection context.
+
+  @note Corresponds to and is associated with core's THD object.
+*/
+class Js_thd {
```
percona-ysorokin, line 133: Similarly, mark all special member functions as `deleted` unless you need them.

components/js_lang/js_lang_core.h
```diff
@@ -0,0 +1,497 @@
...
+ public:
+  Js_thd(MYSQL_THD thd) : m_thd(thd) {}
```
percona-ysorokin: explicit

components/js_lang/js_lang_core.h
```diff
@@ -0,0 +1,497 @@
...
+
+  @note Note that since core's sp_head objects are associated with
+        specific connection (unless they are for trigger), each
+        object of this class is also associated with specific
+        connection.
+  @note However, due to the way SQL core implements per-component
+        THD store, Js_sp object can temporarily outlive Js_thd object
+        for its connection!
+*/
+class Js_sp {
```
percona-ysorokin, line 344: Mark special member functions as `deleted`.

components/js_lang/js_lang_core.h
```diff
@@ -0,0 +1,497 @@
...
+  /**
+    Type of function for getting JS values corresponding to SQL values of
+    IN/INOUT parameters of the routine
+
+    These functions take current Isolate and parameter index as arguments.
+    They return handle for JS value in case of success, and empty handle
+    in case of failure (the error has been reported by the function in the
+    latter case).
+  */
+  typedef std::function<v8::Local<v8::Value>(v8::Isolate *, uint)>
```
percona-ysorokin: using

components/js_lang/js_lang_core.h
```diff
@@ -0,0 +1,497 @@
...
+  /**
+    Type of function for converting JS value and setting it as value of the
+    OUT/INOUT parameter or as a returv value of routine.
+
+    These functions take current v8::Context, parameter index and the JS
+    value as as arguments. They return false in case of success, and
+    true in case of failure (the error has been reported by the function
+    in the latter case).
+  */
+  typedef std::function<bool(v8::Local<v8::Context>, uint,
```
percona-ysorokin: using

components/js_lang/js_lang_param.cc
```diff
@@ -0,0 +1,1225 @@
...
+*/
+
+#include "js_lang_core.h"
+
+// Names of important character sets.
+static const char *const UTF8_CS_NAME = "utf8mb4";
+static const char *const BINARY_CS_NAME = "binary";
+
+// Biggest integer which is safe to store as JS Number value.
+static const long long NUMBER_MAX_SAFE_INTEGER = 9007199254740991;  // 2^53-1
```
percona-ysorokin: May be sized type `std::int64_t`?

components/js_lang/js_lang_param.cc
```diff
@@ -0,0 +1,1225 @@
...
+  This file contains methods of Js_sp class and some helpers which are
+  responsible for converting and storing JS values to/from SQL routine
+  parameters and return values.
+*/
+
+#include "js_lang_core.h"
+
+// Names of important character sets.
+static const char *const UTF8_CS_NAME = "utf8mb4";
+static const char *const BINARY_CS_NAME = "binary";
```
percona-ysorokin: May be `constexpr std::string_view`

components/js_lang/js_lang_param.cc
```diff
@@ -0,0 +1,1225 @@
...
+  return value as well as set this value (using mysql_stored_program_*
+  family services).
+
+  @note In cases when methods for getting metadata or setting values can fail
+        they return false in case of success and true in case of failure (the
+        error is reported by the method in the latter case).
+*/
+class Return_value_handler {
+ public:
+  static bool get_param_type(stored_program_handle sql_sp, uint /* idx */,
```
percona-ysorokin: Here and everywhere `std::size_t` for any index.

components/js_lang/js_lang_param.cc
```diff
@@ -0,0 +1,1225 @@
...
+ public:
+  static bool get_param_type(stored_program_handle sql_sp, uint idx,
+                             uint64_t *sql_type) {
+    if (mysql_service_mysql_stored_program_argument_metadata_query->get(
+            sql_sp, idx, "sql_type", sql_type)) {
+      // Getting parameter type can fail if the SQL type was added recently
+      // and Stored Program Service implementation was not updated to handle
+      // it correctly. Since such scenario is not unlikely we try to play
+      // safe in such a case.
+      my_error(ER_LANGUAGE_COMPONENT, MYF(0),
```
percona-ysorokin, line 119: May be emit an error via component services?

components/js_lang/js_lang_param.cc
```diff
@@ -0,0 +1,1225 @@
...
+  }
+  static bool set_param_int(uint /* idx */, int64_t value) {
+    return mysql_service_mysql_stored_program_return_value_int->set(nullptr,
+                                                                    value);
+  }
+  static bool set_param_uint(uint /* idx */, uint64_t value) {
+    return mysql_service_mysql_stored_program_return_value_unsigned_int->set(
+        nullptr, value);
+  }
+  static bool set_param_string(uint /* idx */, const char *string,
```
percona-ysorokin: consider std::string_view instead of  (string, length) pair

components/js_lang/js_lang_param.cc
```diff
@@ -0,0 +1,1225 @@
...
+  my_error(ER_LANGUAGE_COMPONENT, MYF(0),
+           "Can't create JS value from routine parameter value");
+  return v8::Local<v8::Value>();
+}
+
+/**
+  RAII helper which allows to cleanup string objects represented
+  by my_h_string handle.
+*/
+class My_h_string_guard {
```
percona-ysorokin: Special member functions = deleted

components/js_lang/js_lang_param.cc
```diff
@@ -0,0 +1,1225 @@
...
+ public:
+  My_h_string_guard(my_h_string str_h) : m_str_h(str_h) {}
```
percona-ysorokin: explicit

include/mysql/components/services/mysql_string.h
```diff
@@ -222,6 +222,27 @@ convert_from_buffer_v2_t convert_from_buffer;
...
+  @param      src_buffer    Buffer with a source string
+  @param      src_length    Length of the source string (in bytes).
+  @param      src_charset   Handle for source charset.
+  @param      dest_charset  Handle for destination charset.
+  @param[out] errors        Number of conversion errors.
+
+  @retval False - Success.
+  @retval True  - Failure.
+*/
+DECLARE_BOOL_METHOD(copy_convert,
```
percona-ysorokin, line 240: Is this taken from 9.x or this is just a missing piece you needed?

components/js_lang/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+target_link_directories(component_js_lang SYSTEM PRIVATE ${V8_LIB_DIR})
```
percona-ysorokin: At least in my environment (macOS with the latest CMake), I had to remove the `SYSTEM` keyword to make the code build.

components/js_lang/js_lang_core.h
dlenev: Yes. There is TODO comment to make this a proper class (i.e. with members private by default) once KILL/OOM changes are done. The thing is that these patches will change most of the members of Auth_id_context and its usage, so I have opted for postponing of polishing this struct/class until then...

components/js_lang/js_lang_core.h
dlenev: We can't make Js_isolate private yet - it has to be public so it can be used from std::make_shared<Js_isolate>().
However, it will be possible to do this after one of subsequent patches changes code in such a way that shared_ptr is no longer used with Js_isolate. I have added TODO item about this.

mysql-test/suite/component_js_lang/t/js_lang_big.test
dlenev: Fixed.

components/js_lang/CMakeLists.txt
dlenev: As discussed on Slack - instead of using extra::boost I have opted for removing dependency on Boost for now, since It turned out that code only used BOOST_PP_STRINGIZE() and even that turned out to be not really necessary.

components/js_lang/js_lang_common.h
dlenev: Fixed.

components/js_lang/js_lang_core.h
dlenev: Fixed.

components/js_lang/js_lang_core.h
dlenev: Fixed.

components/js_lang/js_lang_core.h
dlenev: Fixed.

components/js_lang/js_lang_core.h
dlenev: Ditto.

components/js_lang/js_lang_core.h
dlenev: Ditto.

components/js_lang/js_lang_param.cc
dlenev: I have removed this class completely and changed the code to use create_scope_guard() helper from include/scope_guard.h instead.

components/js_lang/js_lang_param.cc
dlenev: See above.

components/js_lang/js_lang_core.h
dlenev: Done.

include/mysql/components/services/mysql_string.h
dlenev: I didn't take this from 9.x, there is no such functionality there.
So this is a missing piece I had to implement myself to support non-UTF8 parameters to JS routines.

components/js_lang/CMakeLists.txt
dlenev: Fixed.

components/js_lang/js_lang_param.cc
dlenev: Indeed. Done.

components/js_lang/js_lang_param.cc
dlenev: Done.

components/js_lang/js_lang_param.cc
dlenev: As discussed on Slack we actually emit error using component services above.
mysql_runtime_error_service.h defines my_error as an alias for mysql_error_service_printf if we are in a component. And the latter is a convenience function which calls runtime error service to report an error.

components/js_lang/js_lang_common.h
dlenev: As discussed on Slack I have opted for using `constexpr char LANGUAGE_NAME[] =` here. We comparing this constant with C-strings only (i.e. ending with \0), even if they returned in a LEX_CSTRING struct, so using std::string_view doesn't make much sense in this case, in my opinion.

components/js_lang/js_lang_param.cc
dlenev: Same as above. I have changed this code to use `constexpr char x[] = ...` instead, as we are comparing these constants with C-strings returned from SQL core as a "const char *"/C-string.

mysql-test/include/plugin.defs
dlenev: Done.

components/js_lang/js_lang_core.h
dlenev: Moved v8_init() and other methods/members related to V8 as a whole into separate Js_v8 class.

components/js_lang/CMakeLists.txt
dlenev: Introduced v8.cmake and ext::v8 interface library.

components/js_lang/js_lang_param.cc
dlenev: I considered this and do not think this makes sense in this specific case...

This method is a thin wrapper around SQL-core service, which doesn't take std::string_view, but (string, length) pair as its parameter. And it gets its values from V8 helper object which doesn't produce std::string_view either (but rather char * and length values). So by using std::string_view in this method we will be creating temporary std::string_view object just for the sake of using std::string_view.

components/js_lang/js_lang_core.h
dlenev: Similarly to what have we have discussed on Slack, it is not mandatory to check sql_command.length in this case as SQL core service returns zero-terminated string anyway. However, it might make sense to do this from performance optimization point of view. Adjusted code accordingly.

dlenev:
Hello @percona-ysorokin !

I think I have addressed all your comments/answered all the questions you had so far.
Could you please be so kind to take a look at the updated code?

components/js_lang/js_lang_core.h
percona-ysorokin: Not for this particular case but just to let you know that there is a technique to use `std::shared_ptr` with "kinda-pricvate" constructors.
https://stackoverflow.com/questions/8147027/how-do-i-call-stdmake-shared-on-a-class-with-only-protected-or-private-const

cmake/v8.cmake
```diff
@@ -0,0 +1,45 @@
...
+
+  # Also check that V8_LIB_DIR contains necessary libraries.
+  FIND_LIBRARY(V8_LIB v8_monolith PATHS ${V8_LIB_DIR} NO_DEFAULT_PATH)
+  IF(NOT V8_LIB)
+    MESSAGE(FATAL_ERROR "Bad path to V8 libraries. No v8_monolith library in V8_LIB_DIR.")
+  ENDIF()
+
+  ADD_LIBRARY(v8_interface INTERFACE)
+
+  TARGET_INCLUDE_DIRECTORIES(v8_interface SYSTEM INTERFACE ${V8_INCLUDE_DIR})
```
percona-ysorokin, line 36: Not sure if `SYSTEM INTERFACE` is the correct order here. Could be `INTERFACE SYSTEM`.
I could be wrong though. Just double-check.

dlenev: It is. https://cmake.org/cmake/help/latest/command/target_include_directories.html says:

```
target_include_directories(<target> [SYSTEM] [AFTER|BEFORE]
  <INTERFACE|PUBLIC|PRIVATE> [items1...]
  [<INTERFACE|PUBLIC|PRIVATE> [items2...] ...])
```

dlenev:
The PR was okayed by @percona-ysorokin and pushed to a branch in the Percona Server main repository https://github.com/percona/percona-server/tree/js-lang.

PR #5452, opened 2024-10-11
PS-9453: percona_telemetry causes a long wait on COND_thd_list due to the absence of the root user

kamil-holubicki:

Problem:

Cause:
reason.

Why it hangs during shutdown:

infinitively.

Solution:
1. Handle the error in cssm_begin_connect() and close opened
2. Percona Telemetry Component uses mysql.session user. When Percona
metrics.

Telemetry Component fails.

### DIFF
--- a/components/percona_telemetry/data_provider.cc
+++ b/components/percona_telemetry/data_provider.cc
@@ -23,6 +23,20 @@
 namespace {
 inline const char *b2s(bool val) { return val ? "1" : "0"; }
 
+/*
+  mysql.session user is mostly enough, but it lacks the following privileges:
+
+  1. REPLICATION SLAVE
+  2. REPLICATION CLIENT
+  3. SELECT on mysql.component
+  4. SELECT on performance_schema.replication_group_members
+
+  These privileges are added at server startup in setup_percona_telemetry()
+  if Percona telemetry is enabled.
+*/
+constexpr const char default_command_user_name[] = "mysql.session";
+constexpr const char default_command_host_name[] = "localhost";
+
 namespace JSONKey {
 const char *pillar_version = "pillar_version";
 const char *db_instance_id = "db_instance_id";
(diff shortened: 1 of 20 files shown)


kamil-holubicki:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/743/
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/744/

mysql-test/r/information_schema_ci.result
```diff
@@ -544,7 +544,9 @@ grant select (a) on test.t1 to joe@localhost with grant option;
 select * from INFORMATION_SCHEMA.COLUMN_PRIVILEGES WHERE table_schema != 'sys';
```
percona-ysorokin, line 544: For some reason permissions changed for this file `100644 → 100755`

mysql-test/suite/component_percona_telemetry/t/no_user.test
```diff
@@ -0,0 +1,70 @@
...
+
+# Wait a few cycles and ensure that SHOW PROCESSLIST does not contain rows related to orphaned sessions.
+--let $timeout = `select $grace_interval + 3 * $scrape_interval`
+--sleep $timeout
+
+--let $assert_text = No orphaned sessions expected in processlist
+--let $assert_cond = [SELECT COUNT(*) as Result FROM performance_schema.processlist WHERE user = "mysql.session";, Result, 1] = 0
+--source include/assert.inc
+
+# Hheck that no new telemetry file was created
```
percona-ysorokin: `Check`

sql/server_component/mysql_command_backend.cc
```diff
@@ -75,6 +75,21 @@ mysql_state_machine_status cssm_begin_connect(mysql_async_connect *ctx) {
...
   MYSQL_SESSION mysql_session = nullptr;
+
+  /* We need to handle the failure in this function.
+     Setting mcs_extn->session_svc right after session open is not enough
+     to handle user lookup errors (and following errors as well)
+     because in case of this function returns error, mysql->extension is
+     cleaned up immediately by the caller. The caller does not take care of
+     session_svc, because it is not aware of this structure.
+  */
+  std::shared_ptr<MYSQL_SESSION> mysql_session_close_guard(
```
percona-ysorokin, line 86: I was struggling with the same double indirection problem some time ago.
`MYSQL_SESSION` is by itself a pointer to an opaque structure. So, I did this trick with
`std::remove_pointer_t<>`;
https://github.com/percona/percona-server/blob/8.0/components/masking_functions/src/masking_functions/sql_context.cpp#L105
In my opinion this make the code more readable.
The only problem I see if you go my way, you won't be able to do `mysql_session = nullptr`. You will need
to call something like `mysql_session_close_guard.release()` but for this, in turns you will need `std::unique_ptr<>` with more boilerplate instead of `std::shared_ptr<>`.
In any case this is just a butification comment.

components/percona_telemetry/data_provider.cc
```diff
@@ -120,25 +134,29 @@ bool DataProvider::do_query(const std::string &query, QueryResult *result,
...
+    We need to call close() always.
+    Even if init() fails, becaues it doesn't allocate anything, calling close()
+    is safe, because internally it checks if provided pointer is valid
+  */
+  std::shared_ptr<MYSQL_H> mysql_h_close_guard(
+      &mysql_h, [&srv = command_factory_service_](MYSQL_H *ptr) {
+        srv.close(*ptr);
+      });
+
   mysql_service_status_t sstatus = command_factory_service_.init(&mysql_h);
```
percona-ysorokin, line 147: Another improvement I discovered recently, which may help simplify things a bit.
Oracle recently introduced a new `MYSQL_COMMAND_LOCAL_THD_HANDLE` option for `mysql_command_service`.
Basically, then you set this option with `nullptr` value, it will create an internal THD object, initialize its session vars, calls my_thread_init() / my_thread_end(), sets `current_thd`,  etc. This THD object will be destroyed when you call `mysql_command_service::close()`.
https://github.com/percona-ysorokin/percona-server/commit/c9a98d3ceacf37449d14b229a9e4e9e2f94138e0#diff-3706ffbbf71661ebd0414b16d27fd6a1cc8f6960488d43d5ad51d9b15fe9ac00R51

percona-ysorokin (approved):
LGTM

file permissions and typos need to be fixed, everything else - just suggestions for improvements.

mysql-test/r/information_schema_ci.result
kamil-holubicki: done

mysql-test/suite/component_percona_telemetry/t/no_user.test
kamil-holubicki: done

PR #5456, opened 2024-10-16
(8.4) PS-9453: percona_telemetry causes a long wait on COND_thd_list due to the absence of the root user

kamil-holubicki:

Problem:

Cause:
reason.

Why it hangs during shutdown:

infinitively.

Solution:
1. Handle the error in cssm_begin_connect() and close opened
2. Percona Telemetry Component uses mysql.session user. When Percona
metrics.

Telemetry Component fails.

### DIFF
--- a/components/percona_telemetry/data_provider.cc
+++ b/components/percona_telemetry/data_provider.cc
@@ -23,6 +23,20 @@
 namespace {
 inline const char *b2s(bool val) { return val ? "1" : "0"; }
 
+/*
+  mysql.session user is mostly enough, but it lacks the following privileges:
+
+  1. REPLICATION SLAVE
+  2. REPLICATION CLIENT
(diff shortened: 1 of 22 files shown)


kamil-holubicki:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.x-pipeline-parallel-mtr/412/
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.x-pipeline-parallel-mtr/413/

percona-ysorokin (approved):
LGTM

kamil-holubicki:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.x-pipeline-parallel-mtr/417/

PR #5460, opened 2024-10-23
(8.0) PS-9453: percona_telemetry causes a long wait on COND_thd_list due to the absence of the root user

kamil-holubicki:

Post push fix.

During the installation of Percona Telemetry Component we grant
'mysql.session' user with additional privileges.
It happens during the server startup, so servers started during the test
will have related timestamps different in general.
Some tests (e.g. clone plugin related) restore the clone instance state
by cloning the donor. In such a case restored instance will have
different timestamps at the beginning and the end of the test and MTR
check will complain because of different tables checksums.

Workaround this problem by excluding mysql.tables_priv from checksum
calculation. Instead, dump all tables privileges but without TIMESTAMP
column.

This is the same approach as for INFORMATION_SCHEMA.ROUTINES.

### DIFF
--- a/mysql-test/include/mtr_check.sql
+++ b/mysql-test/include/mtr_check.sql
@@ -210,6 +210,17 @@ BEGIN
       AND USER NOT IN ('unauthenticated user','mysql.session', 'event_scheduler')
         ORDER BY COMMAND;
 
+  -- During the installation of Percona Telemetry Component we grant 'mysql.session' user with additional privileges.
+  -- It happens during the server startup, so servers started during the test will have related timestamps different in general.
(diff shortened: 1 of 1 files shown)


kamil-holubicki:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/753/
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/754/

dlenev (approved):
LGTM.

PR #5465, opened 2024-10-28
PS-9470 : Code refresh for PS 8.4.3 - post merge MTR fix (fixed auth_…

dlenev:
…sec.acl_tables_row_locking test).

auth_sec.acl_tables_row_locking test failed due to different order of rows returned by queries on P_S.data_locks, which was caused by Upstream's change to P_S.data_locks implementation as part of fix for Bug#36302624 "performance_schema.data_lock_waits is generated with O(N^3) complexity".

See: https://github.com/mysql/mysql-server/commit/96ccd8dc9f9795ff4a671ef0cbf952732461b8a7

The probelm is not specific to PS and affects Upstream as well.

This patch fixes the problem by adding ORDER BY clause to query on P_S.data_locks run by this test, ensuring stable order of result rows, similarly how it is done in some other tests querying this P_S table.


percona-ysorokin (approved):
LGTM

PR #5466, opened 2024-10-30
PS-9470 : Code refresh for PS 8.4.3 - post merge MTR fix (rpl_gtids_t…

dlenev:
…able_disable_binlog_on_slave test improvement).

Greately reduced probability of sporadic failures in rpl_gtid.rpl_gtids_table_disable_binlog_on_slave test.

rpl_gtid.rpl_gtids_table_disable_binlog_on_slave test needs to wait for GTID sets in mysql.gtid_executed table being compressed in a few places. Since this compression happens asynchronously in a compressor or persister threads the test at those points set a global debug flag requesting notification about compression event through debug sync signals, then initiated operation which triggered compression, and then waited for a debug sync point signal indicating that compression has been done. However, sometimes it got this signal too early, because thread performing compression was busy with previous iteration, when the global debug flag requesting notification was set and so we got notified about the end of previous compression iteration. As result mysql.gtid_executed state was different from expected.

This patch solves this problem by replacing this waiting for a signal about compression completion with waiting for expected mysql.gtid_executed state using wait_condition.inc helper script.

Note that the test still can fail occasionally after this patch, but this happens more rarely. This looks like a different problem - a race in GTID code which causes mysql.gtid_executed contents with intersecting GTID sets, which cannot be compressed correctly by current compressor, and which needs to be fixed separately.

### DIFF
--- a/mysql-test/suite/rpl_gtid/r/rpl_gtids_table_disable_binlog_on_slave.result
+++ b/mysql-test/suite/rpl_gtid/r/rpl_gtids_table_disable_binlog_on_slave.result
@@ -69,15 +69,12 @@ include/assert.inc [committed gtids SOURCE_UUID:1-6:17-19]
 #
 # Verify that the gtid_executed table is compressed.
 #
-SET @debug_save= @@GLOBAL.DEBUG;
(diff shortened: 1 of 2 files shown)


percona-ysorokin (approved):
LGTM

PR #5467, opened 2024-10-30
PS-9522 [8.0]: Change deprecated macOS-12 to macOS-14 for Azure Pipelines

inikep:
Microsoft is going to remove `macOS-12` from Azure Pipelines.


percona-ysorokin (approved):
LGTM

PR #5470, opened 2024-10-31
PS-9453: percona_telemetry causes a long wait on COND_thd_list due to the absence of the root user

kamil-holubicki:

Post push fix-2.

Revoked commits 9078f2d8, 60a65b35.

Percona Telemetry Component uses a dedicated user percona.telemetry.


kamil-holubicki:
-DWITH_PERCONA_TELEMETRY=ON
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-parallel-mtr/814/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-parallel-mtr/815/

-DWITH_PERCONA_TELEMETRY=OFF
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-parallel-mtr/830/
https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-parallel-mtr/831/

mysql-test/include/mtr_check.sql
```diff
@@ -206,21 +206,28 @@ BEGIN
   -- not give consistent result either.
   --
   SELECT /*+SET_VAR(use_secondary_engine=OFF)*/ USER, HOST, DB, COMMAND, INFO FROM INFORMATION_SCHEMA.PROCESSLIST
-    WHERE COMMAND NOT IN ('Sleep', 'Daemon', 'Killed')
+    WHERE COMMAND NOT IN ('Sleep', 'Daemon')
```
dlenev: It seems that "Killed" state was added to this condition by Upstream's fix. Should we really revert this?

sql/dd/impl/upgrade/server.cc
```diff
@@ -986,21 +984,24 @@ static const char *percona_telemetry_install[] = {
...
-    "'performance_schema', 'mysql.session', 'replication_group_members', "
-    "'root@localhost', NOW(), 'Select', '');\n",
+    "INSERT IGNORE INTO mysql.user VALUES "
+    "('localhost','percona.telemetry','N','N','N','N','N','N','"
+    "N','N','N','N','N','N','N','N','N','N','N','N','N','N','N','N','N','N','N'"
+    ",'N','N','N','N','','','','',0,0,0,0,'caching_sha2_password','$A$005$"
+    "THISISACOMBINATIONOFINVALIDSALTANDPASSWORDTHATMUSTNEVERBRBEUSED','N',"
+    "CURRENT_TIMESTAMP,NULL,'Y', 'N', 'N', NULL, NULL, NULL, NULL);\n",
+    "UPDATE mysql.user SET Select_priv = 'Y', Repl_slave_priv = 'Y', "
+    "Repl_client_priv = 'Y' WHERE User = 'percona.telemetry' AND Host = 'localhost';\n",
```
dlenev, line 1003: Could you please elaborate why do we need UPDATE as well? Perhaps extend the comment to explain this?

dlenev (changes requested):
Hello Kamil!

Overall, your changes look OK to me! However, I have a couple of questions/comments about the patch, which I would like to get answer before approving the patch.

sql/dd/impl/upgrade/server.cc
dlenev: Ok. I guess this is to handle the case when user manually revokes the privileges from percona.telemtry. IMO still worth commenting.

mysql-test/include/mtr_check.sql
kamil-holubicki: Good catch. I have changed it by mistake. Fixed.

sql/dd/impl/upgrade/server.cc
kamil-holubicki: I've added more clarifications in the code comment.

dlenev (approved):
LGTM.

PR #5471, opened 2024-11-01
(8.4) PS-9453: percona_telemetry causes a long wait on COND_thd_list due to the absence of the root user

kamil-holubicki:
https://perconadev.atlassian.net/browse/PS-9453

Cherry pick from 8.0.

Post push fix-2.

Revoked commits 9078f2d8, 60a65b35.

Percona Telemetry Component uses a dedicated user percona.telemetry.

Conflicts:
	mysql-test/suite/auth_sec/r/atomic_alter_user.result #	mysql-test/t/mysqlpump_extended.test

PR #5474, opened 2024-11-06
(trunk) PS-9453: percona_telemetry causes a long wait on COND_thd_list due to the absence of the root user

inikep:
This PR merges https://github.com/percona/percona-server/pull/5456 into `trunk` using GCA.

Problem:

Cause:
reason.

Why it hangs during shutdown:

infinitively.

Solution:

Handle the error in cssm_begin_connect() and close opened
Percona Telemetry Component uses mysql.session user. When Percona
metrics.
Telemetry Component fails.


kamil-holubicki (changes requested):
Most of these changes were reworked by
https://github.com/percona/percona-server/pull/5470
and
https://github.com/percona/percona-server/pull/5471

new user (percona.telemetry) was introduced.

percona-ysorokin (approved):
LGTM

inikep:
@kamil-holubicki I'm just porting your PRs from `8.4` to `trunk`. It's not possible to add/merge changes now. We will catch your improvements later.

PR #5479, opened 2024-11-07
PS-9470 : Code refresh for PS 8.4.3 - post merge MTR fix (percona.thr…

dlenev:
…eadpool_debug sanitizer error).

Fixed memory leak reported by LeakSanitizer for percona.threadpool_debug test.

The memory leak was caused by missing release of VIO object in one of OOM scenarios tested by threadpool_debug test.

Normally VIO and NET objects for connection are released by THD dtor or THD::release_resources() method call. However, this didn't happen when THD was destroyed on one of OOM-handling paths in Thread_pool_connection_handler::add_connection() method. Instead, we did manual cleanup of NET object, which allowed to avoid closing socket associated with VIO object, and enabled error reporting through this socket even after THD destruction. Unfortunately, this also meant that VIO object was not released properly.

This patch fixes the problem by changing the code to let THD dtor do its job and release both VIO and NET objects associated with connection on this OOM-handling path. But before calling THD dtor, we mark VIO as inactive, to avoid closing the socket and to enable error reporting through it after THD destruction.

### DIFF
--- a/sql/threadpool_unix.cc
+++ b/sql/threadpool_unix.cc
@@ -1204,7 +1204,10 @@ bool Thread_pool_connection_handler::add_connection(
   connection_t *const connection = alloc_connection(thd);
 
   if (unlikely(!connection)) {
-    thd->get_protocol_classic()->end_net();
+    // Tell THD dtor not to close the socket/channel in VIO, but only free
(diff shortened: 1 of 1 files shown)


percona-ysorokin (approved):
LGTM

PR #5480, opened 2024-11-07
PS-9470 : Code refresh for PS 8.4.3 - post merge MTR fix (main.mysqlt…

dlenev:
…est test).

Fixed sporadic failures of main.mysqltest test in Jenkins which occurred when it was executed after some test which created and then dropped events which might have been executed eventually (e.g. main.events_scheduling test).

The failures were caused by difference in processlist dump results which was performed by one of subtests in main.mysqltest test. The expected state of event scheduler thread in the processlist dump was "Waiting on empty queue", which is its normal state when event queue is empty/there are no events to be scheduled (e.g. it can be observed after event scheduler start up with empty mysql.events table). However, in some cases when event queue and mysql.events table are empty the state of event scheduler thread might also be "Waiting for next activation". This happens when event queue was initially non-empty, so event scheduler managed to enter waiting-for-event-activation state, but later became empty due to events from it being dropped. The reason, why this happens is that, unlike event creation or update, event deletion do not wake up event scheduler and cause change of its state.

This patch solves the problem with failures due to such difference by replacing "Waiting for next activation" state with a "Waiting on empty queue" state in processlist dump results.

### DIFF
--- a/mysql-test/t/mysqltest.test
+++ b/mysql-test/t/mysqltest.test
@@ -3044,7 +3044,7 @@ connection default;
 --echo #
 --echo # Test that a query failed with ER_NO_SUCH_THREAD causes mysqltest to dump the processlist before dying
 --echo #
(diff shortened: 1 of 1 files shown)


percona-ysorokin (approved):
LGTM

PR #5481, opened 2024-11-08
PS-9470 : Code refresh for PS 8.4.3 - post merge MTR fix (disabled pe…

dlenev:
…rcona.backup_safe_binlog_info test)

Disabled percona.backup_safe_binlog_info test which failed sporadically in Jenkins.

This test provides coverage for storing current binlog position in InnoDB redo log by LOCK TABLES FOR BACKUP. Investigation has shown that code responsible for storing binlog position in InnoDB redo log no longer works correctly in general case in modern InnoDB. OTOH our backup tools no longer rely on it.

See: https://perconadev.atlassian.net/browse/PS-8897


percona-ysorokin (approved):
LGTM

PR #5489, opened 2024-11-18
PS-9504 [8.0]: Replace `MY_NODISCARD` with `[[nodiscard]]`

inikep:
We require at least gcc 7.1 or clang 5 for compilation which both support `[[nodiscard]]`. `MY_NODISCARD` is no longer needed.


percona-ysorokin (approved):
LGTM

PR #5510, opened 2024-12-03
PS-9416 better kmip error messages (8.0)

lukin-oleksiy:


percona-ysorokin (approved):
LGTM

satya-bodapati:
@lukin-oleksiy Do you think we will see something in Server's error log?

With my ER_ stderr, I could see something in mysqld error log. This helped me to understand that the KMIP server certificate has expired.

PR #5515, opened 2024-12-04
PS-9533 [8.4]: Introduce cmake option to enable probuilds

inikep:
Add `WITH_PERCONA_PROBUILD` cmake option that enables pro features. 
The`PROBUILD` cmake option is an alias for `WITH_PERCONA_PROBUILD`. 
From source code we have to use `#ifdef HAVE_PERCONA_PROBUILD` for pro features.


percona-ysorokin (approved):
LGTM

PR #5516, opened 2024-12-04
PS-9192: JS Stored Routines: advanced error reporting

dlenev:
https://perconadev.atlassian.net/browse/PS-9192

Added a few UDFs which allow to get basic and extended information about last JS error that was reported for the current connection and user. The idea is that these UDFs should simplify debugging of JS stored routines.

They are:

- JS_GET_LAST_ERROR() returns error message for last JS error for the current connection and user.
- JS_GET_LAST_ERROR_INFO() returns an extended information about last JS error such as stored program and line where error was emitted, as well as stack trace if it is available.
- JS_CLEAR_LAST_ERROR() clears last JS error state for the connection/user.

Added test coverage for these UDFs.

### DIFF
--- a/components/js_lang/CMakeLists.txt
+++ b/components/js_lang/CMakeLists.txt
@@ -33,6 +33,7 @@ MYSQL_ADD_COMPONENT(js_lang
   component.cc
   js_lang_core.cc
   js_lang_param.cc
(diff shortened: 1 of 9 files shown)


percona-ysorokin (approved):
LGTM

PR #5520, opened 2024-12-09
PS-9148 feature: Dictionary caching for Masking Functions Component (8.0 with broken sql escaping)

percona-ysorokin:
https://perconadev.atlassian.net/browse/PS-9148


components/masking_functions/src/masking_functions/registration_routines.cpp
```diff
@@ -960,41 +958,35 @@ class gen_blocklist_impl {
...
-      auto query =
-          global_query_builder::instance().check_term_presence_in_dictionary(
-              cs_dict_a, cs_term);
-      auto sresult = sql_ctx.query_single_value(query);
+      auto sresult = global_query_cache::instance()->contains(cs_dict_a_escaped,
+                                                              cs_term_escaped);
 
       if (!sresult) {
-        return {std::string{cs_term.get_buffer()}};
+        return cs_term_escaped;
```
dlenev, line 972: So if term is not present in the dictionary our UDF returns its escaped version? And if it is present, we return some non-escaped random replacement term from the dictionary? This difference escaped - non-escaped sounds suspicious to me. Is this intentional change or a bug?

components/masking_functions/include/masking_functions/basic_sql_context_builder.hpp
```diff
@@ -0,0 +1,54 @@
...
+#define MASKING_FUNCTIONS_BASIC_SQL_CONTEXT_BUILDER_HPP
+
+#include "masking_functions/basic_sql_context_builder_fwd.hpp"
+
+#include "masking_functions/command_service_tuple_fwd.hpp"
+#include "masking_functions/sql_context_fwd.hpp"
+
+namespace masking_functions {
+
+class basic_sql_context_builder {
```
dlenev, line 26: It took me some time to figure out class hierarchy for basic_sql_context_builder and its descendants. Perhaps it would be easier if use more standard "abstract_sql_context_builder" for this class, and perhaps add comments describing this class as well as its descendants? What do you think?

components/masking_functions/src/masking_functions/registration_routines.cpp
```diff
@@ -1025,20 +1017,53 @@ class gen_dictionary_impl {
...
+      throw std::invalid_argument{
+          "Function requires " +
+          std::string(masking_dictionaries_privilege_name) + " privilege"};
     }
+
+    if (ctx.get_number_of_args() > 0)
+      throw std::invalid_argument{
+          "Wrong argument list: masking_dictionaries_flush()"};
+
+    ctx.mark_result_nullable(true);
```
dlenev, line 1050: I think it is worth to add a comment that will explain why this function is nullable/in which cases it can return NULL.

components/masking_functions/src/masking_functions/registration_routines.cpp
```diff
@@ -960,41 +958,35 @@ class gen_blocklist_impl {
     if (ctx.is_arg_null(0)) return std::nullopt;
 
     const auto cs_term = make_charset_string_from_arg(ctx, 0);
-    const auto cs_dict_a = make_charset_string_from_arg(ctx, 1);
-    const auto cs_dict_b = make_charset_string_from_arg(ctx, 2);
+    const auto cs_term_escaped = escape_string(cs_term);
```
dlenev, line 961: After a bit more thinking. IMO it is conceptually wrong to do escaping of dictionary and term names on this level (i.e. in UDF implementation). It is something which belongs to the level interacting with SQL. For example, if we will decide to change the implementation and start using prepared statements on the level below, such escaping on a high-level will become unnecessary and will only harm. So I'd like to propose move escaping back to query_builder.cpp layer.
What do you think?

dlenev (changes requested):
Hello Yura!

Here are some comments about part of the patch I have reviewed so far.

components/masking_functions/include/masking_functions/bookshelf.hpp
```diff
@@ -0,0 +1,58 @@
...
+  bookshelf &operator=(const bookshelf &) = delete;
+  bookshelf &operator=(bookshelf &&) = delete;
+  ~bookshelf();
+
+  bool contains(const std::string &dictionary_name,
+                const std::string &term) const noexcept;
+  // returns empty std::string_view if no such dictionary exist
+  std::string_view get_random(
+      const std::string &dictionary_name) const noexcept;
+  bool remove(const std::string &dictionary_name) noexcept;
```
dlenev, line 43: I think it is worth adding comment to mention what do return values for remove() and insert() mean.
For contains() the name is more or less self-explanatory, but for remove() and insert() it is not obvious, especially in MySQL context.

components/masking_functions/include/masking_functions/component_sys_variable_service_tuple.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include <mysql/components/service.h>
+
+#include <mysql/components/services/component_sys_var_service.h>
+
+#include "masking_functions/component_sys_variable_service_tuple_fwd.hpp"
+
+namespace masking_functions {
+
+// A set of MySQL query services required to perform system variable
```
dlenev, line 27: Perhaps remove word "query" from this description?

components/masking_functions/include/masking_functions/default_sql_context_builder.hpp
```diff
@@ -0,0 +1,41 @@
...
+#define MASKING_FUNCTIONS_DEFAULT_SQL_CONTEXT_BUILDER_HPP
+
+#include "masking_functions/basic_sql_context_builder.hpp"
+
+#include "masking_functions/command_service_tuple_fwd.hpp"
+#include "masking_functions/sql_context_fwd.hpp"
+
+namespace masking_functions {
+
+class default_sql_context_builder : public basic_sql_context_builder {
```
dlenev, line 26: Perhaps it is worth to add a comment describing this class and its difference in comparison to static_sql_context_builder ?

components/masking_functions/include/masking_functions/dictionary.hpp
```diff
@@ -0,0 +1,57 @@
...
+  dictionary &operator=(dictionary &&) = delete;
+
+  ~dictionary() = default;
+
+  bool is_empty() const noexcept { return terms_.empty(); }
+
+  bool contains(const std::string &term) const noexcept;
+  // returns empty std::string_view if the dictionary is empty
+  std::string_view get_random() const noexcept;
+  bool insert(const std::string &term);
```
dlenev, line 44: Again IMO it would be nice to document meaning of return values for insert and remove() as it is not obvious.

components/masking_functions/include/masking_functions/bookshelf.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include <string>
+#include <string_view>
+#include <unordered_map>
+
+#include "masking_functions/dictionary_fwd.hpp"
+
+namespace masking_functions {
+
+class bookshelf {
```
dlenev, line 29: IMO it would be nice to add a comment describing what bookshelf class represents. So one won't have to look into its implementation to figure it out. Indeed bookshelf is place where books (in this case dictionaries) are stored, but still IMO it is worth to spell it out explicitly (e.g. we can say something like: "Bookshelf is a collection of objects representing dictionaries used for data-masking with dictionary name as an index").

components/masking_functions/include/masking_functions/dictionary.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "masking_functions/dictionary_fwd.hpp"
+
+#include <string>
+#include <string_view>
+#include <unordered_set>
+
+namespace masking_functions {
+
+class dictionary {
```
dlenev, line 27: Again, in my opinion, since you are adding new class representing data-masking dictionary it would be nice to add short comment explaining what it is (e.g. collection of terms) and how it is used by data masking functionality on a high-level.

components/masking_functions/include/masking_functions/dictionary_flusher_thread.hpp
```diff
@@ -0,0 +1,120 @@
...
+#include "masking_functions/query_cache_core_fwd.hpp"
+
+namespace masking_functions {
+
+// Facts that impacted 'dictionary_flusher_thread' design.
+// - 'mysql_command_xxx' services cannot be used inside component 'init()' /
+//   'deinit()' handlers as that may mess up with the 'THD' object of the
+//   connection that executes 'INSTALL COMPONENT' / 'UNINSTALL COMPONENT'.
+//   Therefore, 'INSTALL COMPONENT' cannot immediately reload dictionary
+//   cache. It can only spawn a background thread that will do this later.
```
dlenev, line 35: Great comments! However, IMO, it would help if you have mentioned explicitly here that reload of dictionary will be done by this background thread OR by one of data masking UDFs trying to access it, whoever gets to the cache first.

In the end I have found this comment in another file, but having this information in this place would helped me a lot.

components/masking_functions/include/masking_functions/query_cache.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <string>
+
+#include "masking_functions/basic_sql_context_builder_fwd.hpp"
+#include "masking_functions/query_builder_fwd.hpp"
+#include "masking_functions/query_cache_core_fwd.hpp"
+
+namespace masking_functions {
+
+class query_cache {
```
dlenev, line 29: Hmm... "Query Cache" has pretty specific meaning in MySQL context, so using this name here might create some confusion, in my opinion. And technically, you are not caching queries in it or their results, but rather the whole dictionary.

So perhaps it is a good idea to use some other name for this and  query_cache_core classes? dictionaries_cache perhaps? Or masking_dict_cache ? or something else...

What do you think?

components/masking_functions/include/masking_functions/query_cache.hpp
```diff
@@ -0,0 +1,60 @@
...
+  query_cache &operator=(const query_cache &other) = delete;
+  query_cache &operator=(query_cache &&other) = delete;
+  ~query_cache();
+
+  bool contains(const std::string &dictionary_name,
+                const std::string &term) const;
+  // returns a copy of the string to avoid race conditions
+  // an empty string is returned if the dictionary does not exist
+  std::string get_random(const std::string &dictionary_name) const;
+  bool remove(const std::string &dictionary_name);
```
dlenev, line 46: As I have said earlier I think it is good idea to document return values from remove() and insert() methods, since they are non-obvious.

components/masking_functions/include/masking_functions/query_cache.hpp
dlenev: After a bit more thinking... IMO it is definitely worth adding comment explaining relation between query_cache and query_cache_core, otherwise it is really not obvious why we have primitive singleton which stores pointer to query_cache, which stores pointer to query_cache_core, which stores pointer to bookshelf, where we cache things for real (too much indirection for my taste...).

It might help if we come up with a better names for these classes, after all, AFAICS query_cache class not only contains cache, but also sql_context_builder which controls the way UDFs and background thread work with data masking tables.

components/masking_functions/include/masking_functions/query_cache.hpp
```diff
@@ -0,0 +1,60 @@
...
+  bool remove(const std::string &dictionary_name, const std::string &term);
+  bool insert(const std::string &dictionary_name, const std::string &term);
+
+  void reload_cache();
+
+ private:
+  query_cache_core_ptr core_;
+  basic_sql_context_builder_ptr sql_ctx_builder_;
+  query_builder_ptr sql_query_builder_;
```
dlenev, line 55: As far as I can see we have only one instance of query_builder class in the component.
Maybe instead of passing pointer to it around as member function parameter, it makes sense to make it global singleton or simply static member in some class? What do you think?

dlenev (changes requested):
Hello Yura!

Here are bit more comments/questions from me about the PR.

I am continuing my review meanwhile.

mysql-test/suite/component_masking_functions/t/wait_for_component_uninstall.inc
```diff
@@ -0,0 +1,35 @@
+# When flusher thread is enabled in masking functions component,
```
dlenev, line 1: Such approach looks OK to me for case when explicit UNINSTALL COMPONENT is issued. But I wonder what happens if component deinit happens implicitly as result of shutdown. I.e. what will happen if server starts up, inits component and creates background thread, and then immediately tries to shutdown (e.g. due to some other component init failing).

Will the component deinit work correctly in this case?

mysql-test/suite/component_masking_functions/r/dictionary_operations.result
```diff
@@ -21,59 +28,71 @@ ERROR HY000: Can't initialize function 'gen_blocklist'; Wrong argument list: gen
...
+#
 # argument nullness checks for functions not requiring MASKING_DICTIONARIES_ADMIN
 SELECT gen_dictionary(NULL);
 ERROR HY000: gen_dictionary<string> UDF failed; argument 1 cannot be null
 SELECT gen_blocklist('Berlin', NULL, 'us_cities');
 ERROR HY000: gen_blocklist<string> UDF failed; argument 2 cannot be null
 SELECT gen_blocklist('Berlin', 'de_cities', NULL);
 ERROR HY000: gen_blocklist<string> UDF failed; argument 3 cannot be null
+#
 # checking the case when there is no mysql.masking_dictionaries table
```
dlenev, line 57: I think, that after changes in this PR you no longer test case here case when mysql.masking_dictionaries is missing, but instead case when mysql_session@localhost user has insufficient privileges on it... Perhaps the test needs to be extended to test both these scenarios, or at least comment should be adjusted?

mysql-test/suite/component_masking_functions/t/flusher_thread_immediate_restart.test
```diff
@@ -0,0 +1,22 @@
...
+CREATE TABLE mysql.masking_dictionaries(
+    Dictionary VARCHAR(256) NOT NULL,
+    Term VARCHAR(256) NOT NULL,
+    UNIQUE INDEX dictionary_term_idx (Dictionary, Term)
+) ENGINE = InnoDB DEFAULT CHARSET=utf8mb4;
+GRANT SELECT, INSERT, UPDATE, DELETE ON mysql.masking_dictionaries TO 'mysql.session'@'localhost';
+INSERT INTO mysql.masking_dictionaries VALUES('dict1', 'term1');
+
+
+INSTALL COMPONENT 'file://component_masking_functions';
```
dlenev, line 15: OK. I guess this test will occasionally trigger scenario with immediate shutdown and component deinit failure I have mentioned... But I wonder how often it will happen... Perhaps we can make situation more reproducible by adding debug sleep at various stages of flusher thread initialization? What do you think?

mysql-test/suite/component_masking_functions/t/flusher_thread_suspend_resume.test
```diff
@@ -0,0 +1,69 @@
...
+--let $assert_cond = gen_dictionary("dict1") = "term1"
+--let $assert_text = The very first call to gen_dictionary("dict1") should trigger inplace cache reload and return "term1"
+--source include/assert.inc
+
+INSERT INTO mysql.masking_dictionaries VALUES('dict2', 'term2');
+--let $assert_cond = gen_dictionary("dict2") IS NULL
+--let $assert_text = The very first call to gen_dictionary("dict2") should not trigger inplace cache reload and return NULL
+--source include/assert.inc
+
+SET debug_sync = 'now SIGNAL masking_functions_before_cache_reload_signal WAIT_FOR masking_functions_after_cache_reload_signal';
```
dlenev, line 53: A good style is to do debug_sync = 'RESET' at the end of part of the test case using debug sync facility. Yes, in your case problems are unlikely due to presence of .opt file for test case, but IMO it is good idea to stick to good style anyway.

mysql-test/suite/component_masking_functions/t/flusher_thread_connection_reuse.test
```diff
@@ -0,0 +1,40 @@
...
+--connection default
+--reap
+
+--disconnect con_priv
+
+SET debug_sync = 'now SIGNAL masking_functions_before_cache_reload_signal WAIT_FOR masking_functions_after_cache_reload_signal';
+
+--source wait_for_component_uninstall.inc
+
+SET GLOBAL debug = '-d,enable_masking_functions_flusher_create_sync,enable_masking_functions_flush_thread_sync,enable_masking_functions_flush_thread_double_pass';
```
dlenev, line 35: I think it is a good idea to reset debug_sync facility at the end of the test as well.

mysql-test/suite/component_masking_functions/t/rpl_dictionaries_flush_interval.test
```diff
@@ -0,0 +1,138 @@
...
+
+--let $assert_cond = gen_dictionary("dict3") = "term3"
+--let $assert_text = The call to gen_dictionary("dict3") on replica after resuming flusher thread must return "term3"
+--source include/assert.inc
+
+#
+# Cleanup
+--source include/rpl_connection_slave.inc
+REVOKE MASKING_DICTIONARIES_ADMIN ON *.* FROM root@localhost;
+SET GLOBAL debug ='-d,enable_masking_functions_flush_thread_sync';
```
dlenev, line 121: Ditto here. I think it is a good style to reset debug_sync facility at the end of the test as well for both master and slave (though we are not likely to see any problems for this specific test due to such resetting missing).

dlenev (changes requested):
Hello Yura!

I have a bit more comments about your PR.

percona-ysorokin:
This PR is suspended. All its comment will be addressed in the new one
https://github.com/percona/percona-server/pull/5535
(rebased on 8.0.40, squashed commits, fixed clang-tidy warnings for `component_masking_functions` code).

components/masking_functions/src/masking_functions/registration_routines.cpp
percona-ysorokin: This layer has been significantly reworked:

1. Dictionary cache now always stores values in ubt8mb4 (no SQL escaping here).
2. If necessary, terms coming from UDF parameters are converted into utf8mb4 for interaction with dictionary cache.
3. SQL escaping happens only inside the `sql_builder` class.

components/masking_functions/include/masking_functions/basic_sql_context_builder.hpp
percona-ysorokin: `basic_sql_context_builder` renamed to `abstract_sql_context_builder`.
Added a comment describing class hierarchy and purpose of each class.

components/masking_functions/src/masking_functions/registration_routines.cpp
percona-ysorokin: No longer nullable

components/masking_functions/src/masking_functions/registration_routines.cpp
percona-ysorokin: SQL escaping on that level was definitely a wrong decision - this has been reworked properly now.

components/masking_functions/include/masking_functions/bookshelf.hpp
percona-ysorokin: Done

components/masking_functions/include/masking_functions/component_sys_variable_service_tuple.hpp
percona-ysorokin: Done

components/masking_functions/include/masking_functions/default_sql_context_builder.hpp
percona-ysorokin: Done

components/masking_functions/include/masking_functions/dictionary.hpp
percona-ysorokin: Done

components/masking_functions/include/masking_functions/bookshelf.hpp
percona-ysorokin: Done

components/masking_functions/include/masking_functions/dictionary.hpp
percona-ysorokin: Done

components/masking_functions/include/masking_functions/dictionary_flusher_thread.hpp
percona-ysorokin: Done

components/masking_functions/include/masking_functions/query_cache.hpp
percona-ysorokin: `query_cache` renamed to `term_cache`
`query_cache_core` renamed to `term_cache_core`
Added class descriptions.

components/masking_functions/include/masking_functions/query_cache.hpp
percona-ysorokin: Done

components/masking_functions/include/masking_functions/query_cache.hpp
percona-ysorokin: Good point. Reworked as a global singleton - this simplified a lot of interfaces.

mysql-test/suite/component_masking_functions/t/wait_for_component_uninstall.inc
percona-ysorokin: From what I saw during debugging MySQL server during shutdown completely ignores the statuses of the component `deinit()` functions. So, there is a change that after our `deinit()` function completes (with error code), our background thread will still be alive. However, becase we registered our thread in the session manager, another part of the MySQL Server shutdown logic will iterate over remaining active sessions and will make several attempts of gracefully killing them (by setting `THD->killed` flag). One of the most recent flusher thread modifications made sure thsat we react properly to the `THD->killed`. So, this scenario is covered as well.

mysql-test/suite/component_masking_functions/t/flusher_thread_suspend_resume.test
percona-ysorokin: Done

mysql-test/suite/component_masking_functions/t/flusher_thread_connection_reuse.test
percona-ysorokin: Done

mysql-test/suite/component_masking_functions/t/rpl_dictionaries_flush_interval.test
percona-ysorokin: Done

mysql-test/suite/component_masking_functions/r/dictionary_operations.result
percona-ysorokin: See previous response
https://github.com/percona/percona-server/pull/5520?timeline_per_page=10#discussion_r1913972030

mysql-test/suite/component_masking_functions/t/flusher_thread_immediate_restart.test
percona-ysorokin: See previous response

mysql-test/suite/component_masking_functions/r/dictionary_operations.result
percona-ysorokin: MTR test case extended with both scenarios.

PR #5522, opened 2024-12-10
PS-9416 (8.4) better kmip error messages

lukin-oleksiy:


percona-ysorokin (approved):
LGTM

PR #5523, opened 2024-12-11
PS-9416 better kmip error messages (9.x )

lukin-oleksiy:


percona-ysorokin (approved):
LGTM

PR #5528, opened 2024-12-26
PS-9391 Fixes replication break error because of HASH_SCAN 

VarunNagaraju:
https://perconadev.atlassian.net/browse/PS-9391

Problem
=======
When a replica's slave_rows_search_algorithms is set to HASH_SCAN, the
replication may break with HA_ERR_KEY_NOT_FOUND.

Analysis
========
When a replica's slave_rows_search_algorithms is set to HASH_SCAN,
it prepares a unique key list for all the rows in a particular Row_event.
The same unique key list will later be used to retrieve all tuples associated
to each key in the list from storage engine. In the case of multiple updates
targeted at the same row like how it is shown in the testcase, it may happen
that this unique key list filled with entries which don't exist yet in the table.
This is a problem when there is an intermediate update which changes the value
of the index column to a lesser value than the original entry and that changed
value is used in another update as shown in the second part of the testcase.
It is an issue because the unique key list is a std::set which internally
sorts it's entries. When this sorting happens, the first entry of the list
could potentially be a value which doesn't exist in the table and the when
it is searched in next_record_scan() method, it fails returning
HA_ERR_KEY_NOT_FOUND error.

Solution
========
Instead of using std::set to store the distinct keys, a combination of
unordered_set and a list is used to preserve the original order of
updates and avoid duplicates at the same time which prevents the side
effects of sorting.

### DIFF
--- a/sql/log_event.cc
+++ b/sql/log_event.cc
@@ -7870,7 +7870,7 @@ Rows_log_event::Rows_log_event(THD *thd_arg, TABLE *tbl_arg,
       m_curr_row_end(nullptr),
       m_key(nullptr),
       m_key_info(nullptr),
-      m_distinct_keys(Key_compare(&m_key_info)),
+      m_distinct_keys(0, Key_hash(&m_key_info), Key_equal(&m_key_info)),
       m_distinct_key_spare_buf(nullptr) {
   DBUG_TRACE;
   common_header->type_code = event_type;
@@ -7961,7 +7961,7 @@ Rows_log_event::Rows_log_event(
       m_curr_row_end(nullptr),
       m_key(nullptr),
       m_key_info(nullptr),
-      m_distinct_keys(Key_compare(&m_key_info)),
+      m_distinct_keys(0, Key_hash(&m_key_info), Key_equal(&m_key_info)),
       m_distinct_key_spare_buf(nullptr)
 #endif
 {
@@ -9052,9 +9052,9 @@ int Rows_log_event::next_record_scan(bool first_read) {
           marker according to the next key value that we have in the list.
          */
         if (error) {
-          if (m_itr != m_distinct_keys.end()) {
-            m_key = *m_itr;
-            m_itr++;
+          if (m_distinct_key_idx != m_distinct_keys_original_order.size()) {
+            m_key = m_distinct_keys_original_order[m_distinct_key_idx];
+            m_distinct_key_idx++;
             first_read = true;
           } else {
             if (!is_trx_retryable_upon_engine_error(error))
@@ -9088,14 +9088,13 @@ int Rows_log_event::open_record_scan() {
 
   if (m_key_index < MAX_KEY) {
     if (m_rows_lookup_algorithm == ROW_LOOKUP_HASH_SCAN) {
-      /* initialize the iterator over the list of distinct keys that we have */
-      m_itr = m_distinct_keys.begin();
-
-      /* get the first element from the list of keys and increment the
-         iterator
+      /* Initialize the index to go over the vector of distinct keys */
+      m_distinct_key_idx = 0;
+      /* get the first element from the vector of keys and increment the
+         index
        */
-      m_key = *m_itr;
-      m_itr++;
+      m_key = m_distinct_keys_original_order[m_distinct_key_idx];
+      m_distinct_key_idx++;
     } else {
       /* this is an INDEX_SCAN we need to store the key in m_key */
       assert((m_rows_lookup_algorithm == ROW_LOOKUP_INDEX_SCAN) && m_key);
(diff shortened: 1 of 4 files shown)


VarunNagaraju:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-pipeline-parallel-mtr/847/

sql/log_event.h
```diff
@@ -2898,14 +2898,34 @@ class Rows_log_event : public virtual binary_log::Rows_event, public Log_event {
     Key_compare(KEY **ki = nullptr) : m_key_info(ki) {}
     bool operator()(uchar *k1, uchar *k2) const {
       return key_cmp2((*m_key_info)->key_part, k1, (*m_key_info)->key_length,
-                      k2, (*m_key_info)->key_length) < 0;
+                      k2, (*m_key_info)->key_length) == 0;
```
percona-ysorokin, line 2901: Rename this helper class to `Key_equal`.
`Compare` may be misleading as it may mean `comparator` and in your scenario this is no longer the case. All you do now is checking for equality.

sql/log_event.h
```diff
@@ -2898,14 +2898,34 @@ class Rows_log_event : public virtual binary_log::Rows_event, public Log_event {
...
 
    private:
     KEY **m_key_info;
   };
-  std::set<uchar *, Key_compare> m_distinct_keys;
-  std::set<uchar *, Key_compare>::iterator m_itr;
+    class Key_hash {
+    public:
+      Key_hash(KEY **ki = nullptr) : m_key_info(ki) {}
+      size_t operator()(uchar* ptr) const {
```
percona-ysorokin: May be just create a `std::string_view` from `ptr` and `(*m_key_info)->key_length` and calculate a hash of this string_view using standard hash function.
```c++
std::string_view sv{ptr, (*m_key_info)->key_length};
return std::hash(sv);
```

sql/log_event.cc
```diff
@@ -9142,9 +9141,10 @@ int Rows_log_event::add_key_to_distinct_keyset() {
   DBUG_TRACE;
   assert(m_key_index < MAX_KEY);
   key_copy(m_distinct_key_spare_buf, m_table->record[0], m_key_info, 0);
-  std::pair<std::set<uchar *, Key_compare>::iterator, bool> ret =
+  std::pair<std::unordered_set<uchar *, Key_hash, Key_compare>::iterator, bool> ret =
```
percona-ysorokin: May be just `auto`?

sql/log_event.cc
```diff
@@ -9088,14 +9090,11 @@ int Rows_log_event::open_record_scan() {
 
   if (m_key_index < MAX_KEY) {
     if (m_rows_lookup_algorithm == ROW_LOOKUP_HASH_SCAN) {
-      /* initialize the iterator over the list of distinct keys that we have */
-      m_itr = m_distinct_keys.begin();
```
percona-ysorokin, line 9092: Shouldn't we change this to `m_distinct_key_id = 0`?

sql/log_event.cc
```diff
@@ -9068,8 +9068,10 @@ int Rows_log_event::next_record_scan(bool first_read) {
       if ((error = table->file->ha_index_read_map(
                table->record[0], m_key, HA_WHOLE_KEY, HA_READ_KEY_EXACT))) {
         DBUG_PRINT("info", ("no record matching the key found in the table"));
-        if (!is_trx_retryable_upon_engine_error(error))
+        if (!is_trx_retryable_upon_engine_error(error)) {
+          sleep(10);
```
percona-ysorokin: Is this intentional or just for debugging?

percona-ysorokin (changes requested):

sql/log_event.h
VarunNagaraju: Done

sql/log_event.cc
VarunNagaraju: It was just for debugging. Removed it.

sql/log_event.cc
VarunNagaraju: m_distinct_key_idx is already set to 0 when it is declared.

sql/log_event.cc
VarunNagaraju: Done

sql/log_event.h
VarunNagaraju: Thanks for the suggestion. Done.

VarunNagaraju:
https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-pipeline-parallel-mtr/849/

sql/log_event.cc
percona-ysorokin: @VarunNagaraju my point here is that you do not have this initial initialization logic in the constructor of the `Rows_log_event` class but instead have a special function `open_record_scan()` in the interface of this class.
Which means that theoretically the user of this class is allowed to do
```
open_record_scan();
while(...) {
  next_record_scan();
  ...  
}
// and then start iteration for the second time
open_record_scan();
while(...) {
  next_record_scan();
  ...  
}
```
I understand that currently this is not happening - just want to be "future changes"-proof 🙂 here.
In other words, I would still put that `m_distinct_key_id = 0` here.

percona-ysorokin (approved):
LGTM with one `m_distinct_key_id = 0` comment addressed

VarunNagaraju:
https://ps80.cd.percona.com/job/percona-server-8.0-param-parallel-mtr/78/

PR #5529, opened 2025-01-02
(8.0) PS-9621: Initialization of std::atomic_flag caller_active_ in Percona Telemetry component is not compliant with the standard

kamil-holubicki:
https://perconadev.atlassian.net/browse/PS-9621

Initialization of std::atomic_flag in the constructor's initializer list is not compliant with the standard. Gcc allows it, however clang issues a compilation warning which turns into error in maintainer mode build.
(error: braces around scalar initializer [-Werror,-Wbraced-scalar-init])


percona-ysorokin (approved):
LGTM

PR #5531, opened 2025-01-02
(trunk) PS-9621: Initialization of std::atomic_flag caller_active_ in Percona Telemetry component is not compliant with the standard

kamil-holubicki:


percona-ysorokin (approved):
LGTM

PR #5532, opened 2025-01-08
PS-9612 [8.0]: Percona Server build fails on 128+ threads machines

inikep:
Change `MAX_THREAD_GROUPS` from 128 to 1024 to support recent CPUs in thread pool.


percona-ysorokin (approved):
LGTM

